PR #18119, opened 2025-06-11
Add projection data to SegmentAnalysis

cecemei:
## New Feature
`SegmentAnalysis` should return `AggregateProjectionMetadata` in the segment if `AnalysisType.PROJECTIONS` is included in `SegmentMetadataQuery`. If schema doesn't match for the same projection, the specific schema won't be included in the result. 

## Testing & Refactor
- Added `SegmentAnalysisBuilder` class.
- Use parameterized tests in JUnit 5 to simplify tests in `SegmentMetadataQueryQueryToolChestTest`.
- In `IndexMergerV9`, move `sortOrder` must include time column check in projections out of `makeIndexFiles`, to its call site. Also, `Metadata.merge(metadataList, combiningMetricAggs)` would never take null inputs, this refactor makes it more obvious. 
- In `SegmentMetadataQueryQueryToolChest.mergeAnalyses`, throw defensive exception if no data source in the query. 
 
<hr>

 * `SegmentMetadataQuery`
 * `SegmentAnalysis`
 * `SegmentMetadataQueryQueryToolChestTest` 
   * majority is refactoring tests via using the builder class and parameterized tests
 * `IndexMergerV9`

<hr>

This PR has:

- [x] been self-reviewed.


processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
```diff
@@ -119,6 +122,34 @@ public String toString()
...
 
+  /**
+   * Merge two maps of {@link AggregateProjectionMetadata} objects, returning a new map with the same keys and merged
+   * metadata. If the schemas do not match, the metadata is not merged and the key is not included in the result.
+   *
+   * @param projections1 first map of projections to merge
+   * @param projections2 second map of projections to merge
+   * @return merged map of projections
+   */
+  public static Map<String, AggregateProjectionMetadata> merge(
```
clintropolis: I think this logic should be part of the segment metadata query stuff instead of living here since it is specific to that query, so it isn't unintentionally used for other purposes. Similarly, the `Metadata` class has its own logic that it uses when combining the `Metadata` during segment creation, though with different logic that is validating the projections are all the same: https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/Metadata.java#L344

processing/src/main/java/org/apache/druid/query/metadata/metadata/SegmentMetadataQuery.java
```diff
@@ -59,7 +59,8 @@ public enum AnalysisType implements Cacheable
     MINMAX,
     TIMESTAMPSPEC,
     QUERYGRANULARITY,
-    ROLLUP;
+    ROLLUP,
+    PROJECTIONS;
```
clintropolis, line 63: will need to update the docs to include this new analysis type too at some point, https://github.com/apache/druid/blob/master/docs/querying/segmentmetadataquery.md#analysistypes, and also include the fact that when merging results that any mismatched projections will be omitted. It would be fine to do this as a follow-up or part of this PR is fine too

processing/src/main/java/org/apache/druid/segment/IndexMergerV9.java
```diff
@@ -302,20 +270,21 @@ private File makeIndexFiles(
...
+                            segmentMetadata.getProjections(),
+                            adapters,
+                            indexSpec,
+                            segmentWriteOutMedium,
+                            progress,
+                            outDir,
+                            closer,
+                            mergersMap,
+                            segmentMetadata
+          );
```
clintropolis: nit: this is less ugly as an `if` statement

processing/src/test/java/org/apache/druid/query/metadata/SegmentAnalysisBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Helper class to build {@link SegmentAnalysis} objects for testing purposes.
+ */
+public class SegmentAnalysisBuilder
```
clintropolis: nit: we commonly put builders inside of the class they build as an internal class just named like `Builder`, which i think helps with visibility, so people already looking at a class know it has a builder, and just have a static method like `SegmentAnalysis.builder().addFoo(..).build()`

cecemei: I moved it to `SegmentAnalysis`. I put it in test folder initially because it looks like it's mostly needed in tests, not in production code, but i guess it's no harm to put it as internal static class as well. It might also help to unify when we use null, v.s. empty collections.

processing/src/main/java/org/apache/druid/segment/IndexMergerV9.java
cecemei: updated

processing/src/main/java/org/apache/druid/query/metadata/metadata/SegmentMetadataQuery.java
cecemei: updated the doc

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
cecemei: moved it out to `SegmentMetadataQueryQueryToolChest`.

clintropolis (approved):
:metal:

PR #18121, opened 2025-06-11
Add more metric reporting to MSQ

adarshsanjeev:
Adds reporting of additional metrics and dimensions to MSQ task and Dart. 

**Summary**
Currently, MSQ reports the a few metrics. (`ingest/tombstones/count`, `ingest/segments/count` and `ingest/input/bytes`).

This PR adds a few new metrics:
- `query/time`: Reported by controller and worker at the end of the query.
- `query/cpu/time`: Reported by each worker at the end of the query.

Additionally, adds some more dimensions to the metrics emitted. Dimensions added to metrics:
- `queryId`
- `sqlQueryId`
- `engine`: Denotes the engine used for the query, `msq-dart` or `msq-task`.
- `dartQueryId`: (Dart only)
- `type`: Always `msq`
- `dataSource`
- `interval`
- `duration`
- `success`

<hr>

This PR has:

- [ ] been self-reviewed.


extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
```diff
@@ -376,6 +377,7 @@ private MSQTaskReportPayload runInternal(final QueryListener queryListener, fina
 
     final TaskState taskStateForReport;
     final MSQErrorReport errorForReport;
+    final Stopwatch stopwatch = Stopwatch.createStarted();
```
gianm: This should start counting from the moment the query is accepted, which would align with how the native query metric works. You should use the same start time that is used for calculating timeouts.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerContext.java
```diff
@@ -108,8 +113,12 @@ public DartControllerContext(
     this.serverView = serverView;
     this.memoryIntrospector = memoryIntrospector;
     this.context = context;
-    this.metricBuilder = new ServiceMetricEvent.Builder();
     this.emitter = emitter;
+
+    // Set up metric dimensions
+    this.metricBuilder = new ServiceMetricEvent.Builder();
+    MSQMetricUtils.setDartQueryIdDimensions(this.metricBuilder, context);
+    this.metricBuilder.setDimension("engine", DartSqlEngine.NAME);
```
gianm: This line is redundant, I think, since it's already being done in `MSQMetricUtils.setDartQueryIdDimensions`.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartWorkerContext.java
```diff
@@ -120,6 +127,12 @@ public class DartWorkerContext implements WorkerContext
     this.outbox = outbox;
     this.tempDir = tempDir;
     this.queryContext = Preconditions.checkNotNull(queryContext, "queryContext");
+    this.emitter = emitter;
+
+    // Set up metric dimensions
+    this.metricBuilder = new ServiceMetricEvent.Builder();
+    MSQMetricUtils.setDartQueryIdDimensions(this.metricBuilder, queryContext);
+    this.metricBuilder.setDimension("engine", DartSqlEngine.NAME);
```
gianm: Redundant, I think.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
```diff
@@ -553,11 +555,17 @@ private MSQTaskReportPayload runInternal(final QueryListener queryListener, fina
...
     );
+    emitQueryMetrics(stopwatch);
     // Emit summary metrics
     emitSummaryMetrics(msqTaskReportPayload, querySpec);
     return msqTaskReportPayload;
   }
 
+  private void emitQueryMetrics(Stopwatch stopwatch)
+  {
+    context.emitMetric("query/time", stopwatch.millisElapsed());
```
gianm: This should include the major query metric dimensions. Thoughts about what we should do for each of the current ones, in this patch:

- **`dataSource`** - Take all the datasource names from all `TableInputSpec` involved in the query, and report them in the same style as the native engine would report them. Let's ignore other `InputSpec` types for now
- **`type`** - No analog of this in MSQ, so let's just have this say `msq`
- **`interval`** - Take all the intervals from all `TableInputSpec` involved in the query. Let's ignore other `InputSpec` types for now
- **`hasFilters`** - Do not include
- **`duration`** - Compute this the same way that native metrics would work: add up the duration of all the `intervals` in the manner of `BaseQuery#getDuration`
- **`queryId`** - Looks like this is already included
- **`subQueryId`** - Do not include
- **`sqlQueryId`** - Looks like this is already included
- **`context`** - Do not include
- **`success`** - `true` for successful queries, `false` for failures

gianm: A note on this: there is a class `QueryMetrics` that native queries use to report metrics such as `query/time`. It has standard handling for each of these fields. There are also some query-type-specific fields added by subclasses such as `DefaultGroupByQueryMetrics`.

At this time I think we should *not* use `QueryMetrics` here to emit metrics, because it's very tied to the native execution stack: it uses `Query` objects directly and it assumes data servers are only handling a single query type at a time. And I don't think now is the right time to be making big changes to `QueryMetrics`. But there may be some opportunities to share code, such as by breaking out the implementation of computation of specific dimensions (such as `intervals` and `duration`, for example) into helper functions that both can use. We may also revisit harmonizing this more in the future.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerImpl.java
```diff
@@ -220,9 +221,19 @@ public void run()
...
     }
   }
 
+  private void reportCpuMetrics()
+  {
+    long cpuTimeNs = 0L;
+    for (final CounterTracker tracker : stageCounters.values()) {
+      cpuTimeNs += tracker.totalCpu();
+    }
+    context.emitMetric("query/cpu/time", TimeUnit.NANOSECONDS.toMicros(cpuTimeNs));
```
gianm: This one should also have all the same metric dimensions that `query/time` has.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerImpl.java
```diff
@@ -349,6 +361,8 @@ && handleResultsReady(kernelHolder, controllerClient)) {
       }
     }
 
+    context.emitMetric("query/time", stopwatch.millisElapsed());
```
gianm: This one should also have the same metric dimensions as the controller `query/time`.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerImpl.java
adarshsanjeev: Changed.
I assume that for worker metrics, would the datasource, interval, duration etc be based on the stages for which it has actually received work orders. Is that right?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
adarshsanjeev: I wanted to call something out, the metrics that were already being reported already had a `dataSource` dimension, which is the `task.getDataSource()`. To keep it compatible, should these new query specific dimensions like `dataSource` and `interval` be only for the new metrics added by this PR?

gianm: Yes, the `dataSource` metric as I've defined it here should only be attached to the `query/time` and `query/cpu/time` metrics. All other metrics emitted by the task should use `task.getDataSource()`. In the documentation we should call this out.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerImpl.java
gianm: For ease of correlating metrics across workers, let's have all workers compute these dimensions based on the overall query, even if they didn't process all stages. Later on we will want to add stage-specific metrics with stage-specific dimensions.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerContext.java
```diff
@@ -147,6 +149,15 @@ public ControllerQueryKernelConfig queryKernelConfig(
...
     );
 
+    Map<String, Object> indexerContext = IndexerControllerContext.makeWorkerContextMap(
+        querySpec,
+        false,
+        maxConcurrentStages
+    );
+    final ImmutableMap.Builder<String, Object> dartContextBuilder = ImmutableMap.builder();
+    dartContextBuilder.putAll(indexerContext);
+    dartContextBuilder.put(QueryContexts.CTX_DART_QUERY_ID, querySpec.getContext().get(QueryContexts.CTX_DART_QUERY_ID));
```
kgyrtkirk: I feel like you could add another `if` inside `makeWorkerContextMap` to copy over `CTX_DART_QUERY_ID` and skip the rebuild here.

however...I feel like we already have too many contexts - it would be better to not use this.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerContext.java
```diff
@@ -172,8 +183,11 @@ public Injector injector()
   }
 
   @Override
-  public void emitMetric(final String metric, final Number value)
+  public void emitMetric(final String metric, Map<String, Object> overrideDimension, final Number value)
```
kgyrtkirk: I think this method shouldn't belong into this class

the context right now kinda customizes how the `ControllerImpl` will  work....and the `overrideDimension` argument feels like a sign of something is off.

I wonder why not expose the `emitter` and use something else for these?

after some digging it seems to me that passing a correctly pre-initialized `ServiceMetricEvent.Builder` to the constructor of the `ControllerImpl` would be kinda the same....you would need a `copy` or `clone` method for the build as it was adding a few dimensions only in some cases.

not sure how complicated this would be - but if its complicated you could hold the `ServiceMetricEvent.Builder` skeleton in the context and use that before emitting

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/counters/CounterTracker.java
```diff
@@ -112,6 +112,16 @@ public <T extends QueryCounter> T counter(final String counterName, final Suppli
     return (T) countersMap.computeIfAbsent(counterName, ignored -> newCounterFn.get());
   }
 
+  public long totalCpu()
+  {
+    long cpuTimeNs = 0;
+    CpuCounters cpuCounters = (CpuCounters) countersMap.getOrDefault(CounterNames.cpu(), new CpuCounters());
```
kgyrtkirk: nit: it feels odd to use `getOrDefault` here...
if it would return the default - don't think that would have any `counters` in it - as nobody would call `forName` to create them

gianm: Yes the default would not have any counters in it. It would be equivalent to do a `get` and return `0` if `null` comes back.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerContext.java
gianm: It's definitely possible to avoid using `CTX_DART_QUERY_ID` for the metrics.

In Dart, the `CTX_DART_QUERY_ID` is the same as the MSQ query id. The worker already knows this from the `WorkerContext#queryId()` method, so it doesn't need it in the `workerContextMap`. I think the manipulations here could all be dropped if the worker code is adjusted to use `queryId()` instead of trying to pull it from the context map.

On the controller, btw, the `ControllerContext` doesn't currently know the query id. But this could change. I think it would likely clean things up a bit to make this change as well, i.e.:

- remove `queryId` from `ControllerImpl` constructor
- add a `queryId()` method to `ControllerContext`
- remove `queryId` from `ControllerContext#queryKernelConfig` (as the context would now know its query id, no need to provide it)

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/MSQMetricUtils.java
```diff
@@ -0,0 +1,131 @@
...
+  public static Map<String, Object> createQueryMetricDimensions(
+      final Set<String> datasources,
+      final Collection<Interval> intervals,
+      final boolean success
+  )
+  {
+    Map<String, Object> dims = new HashMap<>();
+    dims.put(DruidMetrics.DATASOURCE, DefaultQueryMetrics.getTableNamesAsString(datasources));
+    dims.put(DruidMetrics.INTERVAL, DefaultQueryMetrics.getIntervalsAsStringArray(intervals));
+    dims.put("duration", BaseQuery.calculateDuration(intervals));
```
gianm: Add constants to `DruidMetrics` for `duration` and `success` too.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerContext.java
gianm: IMO, this is OK. I will describe why I feel this way.

The main design problem is that dimensions from multiple places have to come together:

- For tasks, something needs to call `IndexTaskUtils.setTaskDimensions` to provide dimensions like `taskId`, `taskType`, etc for all metrics. These need to be provided somehow by the context, because the sites that emit metrics do not, and should not "know" what task they are running in. The context is the thing that knows.
- For query metrics, something needs to provide dimensions like `success`, `type`, etc. Some are knowable before the query runs (like `type`) but some are only knowable after it completes (like `success`). Because of the latter ones that require the query to actually finish, these dimensions can't be fully provided by the context.

So, there must necessarily be some mix of dimensions provided by the context, and dimensions not provided by the context. I think there's only really two ways to organize the code: either the context provides its builtin dimensions to callers (along with an emitter so the caller can emit itself), _or_ callers provide their additional dimensions to the context, and the context emits. The second way is what's currently implemented in this PR and IMO it is OK.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerImpl.java
```diff
@@ -155,6 +157,16 @@ public class WorkerImpl implements Worker
    */
   private volatile boolean controllerAlive = true;
 
+  /**
+   * Stores list of datasources queried for metric dimensions.
+   */
+  private final Set<String> datasources = new HashSet<>();
```
gianm: Let's just have `datasources` and `intervals` combined into `Map<String, Object> queryMetricDimensions`. It's simpler and it avoids having "tempting" fields lying around like `datasources` and `intervals`. We don't want these used for any purpose other than metric emission.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/MSQMetricUtils.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Utility methods for setting up dimensions on metrics.
+ */
+public class MSQMetricUtils
+{
+  private static void setQueryIdDimensions(
```
gianm: Javadocs for all of these `public static` methods please. They should say what dimensions they're setting.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/MSQMetricUtils.java
```diff
@@ -0,0 +1,131 @@
...
+      final Task task,
+      final QueryContext queryContext
+  )
+  {
+    setQueryIdDimensions(metricBuilder, queryContext);
+    IndexTaskUtils.setTaskDimensions(metricBuilder, task); // Setup common non-MSQ task dimensions
+    metricBuilder.setDimension(DruidMetrics.ENGINE, MSQTaskSqlEngine.NAME);
+  }
+
+  public static void populateDatasourcesAndInterval(
```
gianm: Rather than having out-parameters, IMO it's better to have two methods, one that returns datasources and one that returns intervals. Out-parameters should generally be a last resort, since they aren't normally how code is written and can cause it to be more difficult to read through quickly.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerContext.java
adarshsanjeev: Reworked this

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
```diff
@@ -553,11 +554,31 @@ private MSQTaskReportPayload runInternal(final QueryListener queryListener, fina
...
+        datasources.addAll(MSQMetricUtils.getDatasources(stageDefinition));
+        intervals.addAll(MSQMetricUtils.getIntervals(stageDefinition));
+      }
+    }
+
+    long startTime = DateTimeUtils.getInstantMillis(MultiStageQueryContext.getStartTime(querySpec.getContext()));
+    context.emitMetric(
+        "query/time",
+        MSQMetricUtils.createQueryMetricDimensions(datasources, intervals, success),
+        startTime - System.currentTimeMillis()
```
gianm: This looks backwards.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/MSQMetricUtils.java
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * Sets dimensions for Dart queries, as well as some common dimensions. Sets the following dimensions:
+   * <ul>
+   * <li>queryId</li>
+   * <li>sqlQueryId</li>
+   * <li>type</li>
+   * <li>engine</li>
+   * </ul>
+   */
+  public static void setDartQueryIdDimensions(
```
gianm: The name of this method is a bit odd. It definitely sets more than the "query id dimensions". How about calling it `setDartDimensions`.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/MSQMetricUtils.java
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * Sets dimensions for MSQ task queries, as well as some common dimensions. Sets the following dimensions:
+   * <ul>
+   * <li>queryId</li>
+   * <li>sqlQueryId</li>
+   * <li>type</li>
+   * <li>engine</li>
+   * </ul>
+   */
+  public static void setTaskQueryIdDimensions(
```
gianm: Similar to the other one, how about calling this `setTaskDimensions`? Also, the javadoc isn't quite accurate, it's missing a few dimensions.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerContext.java
```diff
@@ -56,6 +58,12 @@ public interface WorkerContext extends Closeable
   // Using an Injector directly because tasks do not have a way to provide their own Guice modules.
   Injector injector();
 
+  /**
+   * Emit a metric using a {@link ServiceEmitter}. The dimensions passed as an argument are also emitted with the
+   * metrics. These are can override any already existing context specific dimensions.
+   */
+  void emitMetric(String metric, Map<String, Object> overrideDimension, Number value);
```
gianm: `overrideDimensions` (spelling)

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerContext.java
```diff
@@ -33,17 +33,23 @@
...
 /**
  * Context used by multi-stage query controllers. Useful because it allows test fixtures to provide their own
  * implementations.
  */
 public interface ControllerContext
 {
+  /**
+   * Query ID for this context.
+   */
+  String queryId();
```
kgyrtkirk, line 48: we already have a lot of queryId-s this method will just increase the confusion - one implementation of this method returns `DART_QUERY_ID` and the other something else

right now all callsites seem to be inside the implementations of this interface.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
HiveQA: I still feel that the `context#emitMetric` method is odd; it could be something like:

```
    MsqServiceMetricEventBuilder metricBuilder = new MsqServiceMetricEventBuilder(querySpec);

    metricBuilder.setDimension(DruidMetrics.DATASOURCE, DefaultQueryMetrics.getTableNamesAsString(datasources));
    metricBuilder.setDimension(DruidMetrics.INTERVAL, DefaultQueryMetrics.getIntervalsAsStringArray(intervals));
    metricBuilder.setDimension(DruidMetrics.DURATION, BaseQuery.calculateDuration(intervals));
    metricBuilder.setDimension(DruidMetrics.SUCCESS, success);

    metricBuilder.setMetric("query/time", startTime - System.currentTimeMillis());
    context.getEmitter().emit(metricBuilder);

```

probably `MsqServiceMetricEventBuilder` could also provide a more natural home for those static methods which could become instance methods


..or alter the signature of `context#emitrMetric` to  `context#emit(metricBuilder)` ? so that the context could still 
add stuff if it wants - but the caller has the control.

kgyrtkirk: sorry about the above author :D it was me...I was just checking some github features ; which involved switching between 2 users ....

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerContext.java
adarshsanjeev: The context being aware of the queryId does make sense. Now that we have this, we could get the queryId from the context instead of having it in the controller. Would that make it neater?

gianm: > we already have a lot of queryId-s this method will just increase the confusion - one implementation of this method returns DART_QUERY_ID and the other something else

From my point of view both implementations return the same thing: they both return the unique identifier which globally identifies the query. For the task it's the controller task ID and for Dart it's the `dartQueryId`. They don't source the ID from the same place, but that's the point of the context anyway. It abstracts over differences in execution model.

We can explain this better in the javadoc, using something like:

```
/**
 * Globally unique identifier for the query handled by this controller. This is used to set
 * {@link QueryDefinition#getQueryId}. Must be globally unique because this is used for
 * identifying workers, naming temporary files, etc.
 */
String queryId();
```

It might be too disruptive to change the naming now, but I wonder would this concept seem more OK to you if it was called something else, like `jobId`?

> right now all callsites seem to be inside the implementations of this interface.

This is true, although I think it would make sense to have `ControllerImpl` get the query ID from here, rather than have it passed in the constructor.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
gianm: Ah, yes, that would work too. It's kind of equivalent (the `metricBuilder` has a metric name, value, and dimensions in it) but bundled together more nicely. @adarshsanjeev - could you try this approach? It looks like it could be cleaner.

adarshsanjeev: Ah, thanks for catching this!

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/MSQMetricUtils.java
adarshsanjeev: Added the dimensions from `IndexTaskUtils.setTaskDimensions`

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerContext.java
adarshsanjeev: Removed queryId from the controller and using context for that now!

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
adarshsanjeev: @kgyrtkirk Made these changes. The function now takes a service emitter. 

There's an awkward bit in `TaskControllerContext`, where we have to check if the dataSource has already been provided in the metricBuilder before setting the default value, for backward compatibility. I wonder if it is worth adding yet another function of `getDefaultMetricBuilder()` from the context instead, so that the caller can decide what all to override instead of this weird code.

Please let me know your thoughts.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/MSQMetricUtils.java
```diff
@@ -0,0 +1,137 @@
...
+        task.<Map<String, Object>>getContextValue(DruidMetrics.TAGS)
+    );
+    metricBuilder.setDimensionIfNotNull(DruidMetrics.GROUP_ID, task.getGroupId());
+    metricBuilder.setDimension(DruidMetrics.ENGINE, MSQTaskSqlEngine.NAME);
+  }
+
+  /**
+   * Returns a set of all datasources of all {@link TableInputSpec} in the stageDefinition.
+   */
+  public static Set<String> getDatasources(final StageDefinition stageDefinition)
```
kgyrtkirk: nit: seems like this could naturally live as `StageDefintion#getDatasources` ?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/MSQMetricUtils.java
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * Sets dimensions for Dart queries, as well as some common dimensions. Sets the following dimensions:
+   * <ul>
+   * <li>queryId</li>
+   * <li>sqlQueryId</li>
+   * <li>type</li>
+   * <li>engine</li>
+   * </ul>
+   */
+  public static void setDartDimensions(
```
kgyrtkirk: nit: these `setX` methods could live inside some `MSQMetriceEventBuilder` class which should be a sublass of `ServiceMetricEvent.Builder`

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerImpl.java
```diff
@@ -394,6 +433,17 @@ private void handleNewWorkOrder(
     kernelHolder.partitionBoundariesFuture = runWorkOrder.getStagePartitionBoundariesFuture();
   }
 
+  @SuppressWarnings("unchecked")
+  private void updateMetricDimensions(QueryDefinition queryDef)
+  {
+    Set<String> datasourceDim = (Set<String>) queryMetricDimensions.computeIfAbsent(DruidMetrics.DATASOURCE, s -> new HashSet<>());
+    Set<Interval> intervalDim = (Set<Interval>) queryMetricDimensions.computeIfAbsent(DruidMetrics.INTERVAL, s -> new HashSet<>());
+    for (StageDefinition stageDef : queryDef.getStageDefinitions()) {
+      datasourceDim.addAll(MSQMetricUtils.getDatasources(stageDef));
```
kgyrtkirk: nit: is this a one time need - or maybe there should be `queryDef.getDatasources`  ?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerImpl.java
```diff
@@ -394,6 +433,17 @@ private void handleNewWorkOrder(
     kernelHolder.partitionBoundariesFuture = runWorkOrder.getStagePartitionBoundariesFuture();
   }
 
+  @SuppressWarnings("unchecked")
+  private void updateMetricDimensions(QueryDefinition queryDef)
+  {
+    Set<String> datasourceDim = (Set<String>) queryMetricDimensions.computeIfAbsent(DruidMetrics.DATASOURCE, s -> new HashSet<>());
```
kgyrtkirk, line 442: note: I think it's true that all `workOrder`-s will be for the same `queryDef` - is that correct ?

extensions-core/multi-stage-query/src/test/java/org/apache/druid/msq/dart/controller/http/DartSqlResourceTest.java
```diff
@@ -569,6 +578,7 @@ public void test_doPost_regularUser_runtimeError() throws IOException
     assertThat((String) e.get("errorMessage"), CoreMatchers.startsWith("InvalidNullByte: "));
   }
 
+  @Test
```
kgyrtkirk, line 581: good catch :)

extensions-core/multi-stage-query/src/test/java/org/apache/druid/msq/test/MSQTestOverlordServiceClient.java
```diff
@@ -379,4 +386,12 @@ public void closeTask(String taskId)
 
     taskDetailsByTaskId.get(taskId).close();
   }
+
+  /**
+   *
+   */
```
kgyrtkirk: nit: remove?

kgyrtkirk (approved):
thank you for the updates! looks good
left a few notes here and there - the IT failures are unrelated

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerImpl.java
adarshsanjeev: Currently, that is true, yes

extensions-core/multi-stage-query/src/test/java/org/apache/druid/msq/test/MSQTestOverlordServiceClient.java
adarshsanjeev: Added a javadoc

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerImpl.java
adarshsanjeev: Currently, I believe this is a one time need. Now that I have changed getDatasources to a member of stageDefinition, maybe this isn't required.

PR #18123, opened 2025-06-11
Support Sort in decoupled Dart plans

kgyrtkirk:
* 🔁 runs `Sort`-s with the use of a separate stage - which makes this plan not entirely the same as the one executed in the end - but its easier to comprehend and when I've started working on adding real grouping it seems like a good fit
* 📄 the translation to `QueryDefinition` has its own place now ; it right now adds surpulus stages which could be collapsed - but that could be fixed later
* 📦 moves the `*Stage` classes outside of `LogicalStageBuilder` (which was never really a builder)


extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/querykit/scan/ScanQueryFrameProcessorFactory.java
```diff
@@ -58,6 +65,22 @@ public ScanQueryFrameProcessorFactory(@JsonProperty("query") ScanQuery query)
     this.runningCountForLimit = query.isLimited() && query.getOrderBys().isEmpty() ? new AtomicLong() : null;
   }
 
+  public static ScanQueryFrameProcessorFactory makeScanFrameProcessor(
```
adarshsanjeev: nit: makeScanFrameProcessorFactory

adarshsanjeev:
> it right now adds surpulus stages which could be 

These stages would be the extra stages for shuffling and reading inputs?

kgyrtkirk:
>>   it right now adds surpulus stages which could be

> These stages would be the extra stages for shuffling and reading inputs?

yes - I'll leave that to a separate optimization PR; I believe with that addition in some cases it will be better than the old QK based plans.

adarshsanjeev (approved):

kgyrtkirk:
Thank you @adarshsanjeev for taking a look! :)

PR #17838, opened 2025-03-26
Improve revised integration tests for local use by devs

capistrant:
### Description

The main goal of this work is to make running the revised ITs locally easier for devs while not impacting the robustness of ITs already successfully running in Github actions CI pipeline

#### Improved flakiness of cluster startup caused by race condition in container startup (druid vs MySQL)

MySQL has to be up accepting connections before coordinator/overlord containers start, otherwise their is a risk of them trying to create tables and indexes before MySQL is healthy. When this happened on my local it led to undesired behavior (services starting up, failing to create metadata tables and indexes, and then complaining about them not existing for remainder of cluster uptime).

To accomplish this, I made two main changes: 
1. I employed a docker-compose healthcheck for MySQL that waits for the ability to connect to the Druid database before determining itself healthy. I've been tinkering with the intervals, timeouts, etc trying to find a balance between waiting to long in a hopeless scenario and quitting just before success.
2. I modified the docker-compose files generated to be more specific in defining dependency services. Dependencies now come with a condition that must be met before starting. In the case of services that rely on MySQL, they are forced to wait until it is healthy before their containers start.

#### Improved flakiness of Security IT suite

Locally I was still seeing flakiness that seemed to be caused by authz policies not syncing in time for each test. I saw that one test had it's "policy sync sleep" increased to 4 seconds with the others left at 1. I centralized sleep time with a static variable so we don't run into drift like this again. I also upped the sleep to 10 seconds, but am not wed to that number if we want to try to squeeze it down some. After this change, I saw much more consistent success running the Security suite locally.

#### Improved UX for folks using machines with Apple Silicon

In the past the docker-compose template had  a commented out `platform` key that told Apple Silicon users to uncomment it. This will inevitably lead to folks using Apple chips to become fatigued with remembering this bit of information. Instead I a modified the template.py script that generates the docker-compose file to key off of the host platform and set the proper override if we are running on `arm`. 

#### Migrated Query and MultiStageQueryWithMM 

Query and MultiStageQueryWithMM were using their own static docker-compose files. This goes against the pattern of generating the compose file dynamically in the other IT categories. I am suspicious that it is a case of the tests not fully migrating from the "old" IT pattern to the "revised" pattern. After migrating them to the new generated compose file pattern, I had no issues locally. Interested to see them run in CI.

These categories rely on Kafka, which pushed me to add Kafka as a core dependency generated in the compose file by `template.py`. I updated the categories that don't need kafka to exclude it from their compose files going forward.

#### Release note

[...]

PR #18125, opened 2025-06-11
Wait for minio to be healthy before creating buckets in s3 integration-tests-ex

capistrant:
#### Fixed race condition when running `S3DeepStorage` integration-tests-ex locally

Locally, there can be issues with integration-tests-ex that use minio due to the buckets trying to get created before minio is ready. This PR works to resolve that by waiting for minio to be healthy before creating buckets. It should provide more reliability when running integration-tests-ex locally. This is a similar issue to what was seen in mysql locally, and fixed with #17838 

#### Release note

N/A Developer facing

<hr>

 * `integration-tests-ex/cases/cluster/Common/dependencies.yaml`
 * `integration-tests-ex/cases/cluster/S3DeepStorage/docker-compose.py`

<hr>

This PR has:

- [X] been self-reviewed.

### DIFF
--- a/integration-tests-ex/cases/cluster/Common/dependencies.yaml
+++ b/integration-tests-ex/cases/cluster/Common/dependencies.yaml
@@ -120,9 +120,9 @@ services:
       test: [ "CMD", "curl", "-f", "http://localhost:9000/minio/health/ready" ]
       interval: 5s
       timeout: 3s
-      retries: 60
-      start_period: 3m
-      start_interval: 15s
+      retries: 12
+      start_period: 1m
+      start_interval: 5s
 
   create_minio_buckets:
     image: minio/mc


integration-tests-ex/cases/cluster/Common/dependencies.yaml
```diff
@@ -116,6 +116,13 @@ services:
     environment:
       - MINIO_ROOT_USER=${AWS_ACCESS_KEY_ID}
       - MINIO_ROOT_PASSWORD=${AWS_SECRET_ACCESS_KEY}
+    healthcheck:
+      test: [ "CMD", "curl", "-f", "http://localhost:9000/minio/health/ready" ]
+      interval: 5s
+      timeout: 3s
+      retries: 60
+      start_period: 3m
+      start_interval: 15s
```
kgyrtkirk: note: maybe the the intervals could be more agressive; `15s` seems like a lot to wait....

not sure if these matter at all - I was playing with it and startup took `11s` no matter what I've changed...

kgyrtkirk (approved):

integration-tests-ex/cases/cluster/Common/dependencies.yaml
capistrant: I admittedly used the same values that I did with mysql which was much more irregular starting up than minio. Thanks for calling it out, I'll adjust the parameters accordingly

Issue #17628, opened 2025-01-15
Workbench autocomplete should be smater


JRobTS:
### Description

The Druid Console Workbench will auto-complete in the wrong contexts and should be smarter.

I propose the following changes:
* "enter" shouldn't trigger an auto-complete
* Numerals should not be candidate for autocompletion
* Inside of double quote "" blocks, function names should not be candidate for autocompletion: datasource and field names only
* Inside of single quote '' blocks, should not be candidate for autocompletion

### Motivation

Make working with Workbench less annoying.

For example, "1" will auto-complete as "LOG10" on (enter). This is extremely annoying as it's pretty typical to write a query that ends in `GROUP BY 1`; only to have auto-complete change it to `GROUP BY LOG10`.

Other annoyances include when writing inside of a string block ("" or '') the auto-completer will try to fill in function names. `AND subject = 'foo'` will auto-complete to `AND subject = 'FLOOR'`

Datasource and field names should be within context for double quotes "", but not function names. And I can't think of any reason to auto-complete when inside of single quotes ''.

gianm:
I'm not familiar with how the autocomplete is coded, in particular I'm not sure how much control we have over it (vs. it coming in as part of a component we're importing). Perhaps @vogievetsky has a comment on that.

As to the specifics, some 2¢:

- I have also been bit by the autocompleting within single quotes and agree it doesn't seem that useful.
- Perhaps autocompleting on tab only rather than tab and enter would be an improvement. I'd be interested in trying it at least.
- I haven't personally experienced `1` autocompleting as `LOG10` (it doesn't autocomplete at all when I tried it just now on a relatively recent build-- maybe something changed).

JRobTS:
I should note I'm using Apache Druid 29.0.0

I agree with autocompleting on tab only

Tested again and even `SELECT 1` recommends `LOG10`
<img width="428" alt="Image" src="https://github.com/user-attachments/assets/6b5476aa-672a-45f7-a678-4f8d4d37aac5" />

vogievetsky:
Sorry it took me a while to get around to responding to this issue. I somehow missed it. I am planning on upgrading the autocomplete based on the feedback in this issue. Here are my thoughts:

> Numerals should not be candidate for autocompletion

[...]

PR #17353, opened 2024-10-15
Scheduled batch supervisor

abhishekrb19:
This change introduces a scheduled batch supervisor in Druid. The supervisor periodically wakes up to submit an MSQ ingest query, allowing users to automate batch ingestion directly within Druid. Think of it as simple batch task workflows natively integrated into Druid, though it doesn't replace more sophisticated workflow management systems like Apache Airflow. This is an experimental feature.

## Summary of changes:

The `scheduled_batch` supervisor can be configured as follows:

```json
{
    "type": "scheduled_batch",
    "schedulerConfig": {
        "type": "unix",
        "schedule": "*/5 * * * *"
    },
    "spec": {
        "query": "REPLACE INTO foo OVERWRITE ALL SELECT * FROM bar PARTITIONED BY DAY"
    },
    "suspended": false
}
```
The supervisor will submit the `REPLACE` sql query repeatedly every 5 minutes. The supervisor supports two types of cron scheduler configurations:

 1. Unix cron syntax:
     - Type must be set to `unix`.
     - Follows the [standard](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) Unix cron format, e.g., `*/5 * * * *` to schedule the SQL task every 5 minutes.
     - Supports macro expressions such as `@daily`, `@hourly`, `@monthly`, etc.
 2. Quartz cron syntax:
     - Type must be set to `quartz`.
     - Offers more flexibility and [control](https://productresources.collibra.com/docs/collibra/latest/Content/Cron/co_quartz-cron-syntax.htm) for scheduling tasks.
     - Example: `0 0 0 ? 3,6,9,12 MON-FRI` to schedule tasks at midnight on weekdays during March, June, September, and December.

Key points:
- User can specify the `query` along with any context in the `spec`. This structure is identical to what the MSQ task engine [accepts](https://druid.apache.org/docs/latest/multi-stage-query/reference#context-parameters).
- Currently, the batch supervisor will repeatedly submit the DML SQL present in the `spec` as-is on its schedule.
- This can be useful for recurring batch ingestion or reindexing tasks.
- There is no parameterization or "change detection" support for input sources yet, but some simple mechanisms are planned for the future.
- Users can configure multiple scheduled batch supervisors for the same datasource.
- Only DML ingestion queries are currently supported by the scheduled batch supervisor.
- Users can suspend, resume and terminate the scheduled batch supervisor. Suspending the supervisor stops new tasks from being scheduled. 

Some implementation details:
[...]

PR #18126, opened 2025-06-11
Web console: improve SQL autocomplete and add JSON autocomplete 

vogievetsky:
Fixes https://github.com/apache/druid/issues/17628

Made the SQL autocomplete much smarter

image

(note how it is not trying to auto-complete functions and other nonsense)

Also added autocomplete to JSON native queries and to all the JSON entry fields

image

image


web-console/src/views/load-data-view/ingestion-spec-completions.ts
```diff
@@ -0,0 +1,831 @@
...
+      { value: 'rabbit', documentation: 'RabbitMQ streaming ingestion' },
+    ],
+  },
+  // Batch ioConfig properties
+  {
+    path: '$.spec.ioConfig',
+    isObject: true,
+    condition: obj => obj.type === 'index_parallel' || obj.type === 'index',
+    completions: [
+      { value: 'firehose', documentation: 'Legacy data input (deprecated)' },
```
clintropolis: this is gone, not just deprecated

web-console/src/views/load-data-view/ingestion-spec-completions.ts
```diff
@@ -0,0 +1,831 @@
...
+  },
+  // HTTP input source properties
+  {
+    path: '$.spec.ioConfig.inputSource',
+    isObject: true,
+    condition: obj => obj.type === 'http',
+    completions: [
+      { value: 'uris', documentation: 'List of HTTP/HTTPS URIs' },
+      { value: 'httpAuthenticationUsername', documentation: 'HTTP authentication username' },
+      { value: 'httpAuthenticationPassword', documentation: 'HTTP authentication password' },
```
clintropolis, line 507: there are a couple missing ones here, `systemFields` and `requestHeaders` though the latter isn't documented either

web-console/src/druid-models/native-json-query/native-json-query-completions.ts
```diff
@@ -0,0 +1,1098 @@
...
+      { value: 'segmentMetadata', documentation: 'Segment metadata query' },
+      { value: 'dataSourceMetadata', documentation: 'Data source metadata query' },
+    ],
+  },
+
+  // Common properties for most query types
+  {
+    path: '$',
+    isObject: true,
+    condition: obj => obj.queryType !== 'dataSourceMetadata',
```
clintropolis, line 60: afaik, everything has a context and intervals, even `dataSourceMetadata`

web-console/src/druid-models/native-json-query/native-json-query-completions.ts
```diff
@@ -0,0 +1,1098 @@
...
+  {
+    path: '$.filter',
+    isObject: true,
+    completions: [{ value: 'type', documentation: 'Type of filter' }],
+  },
+
+  {
+    path: '$.filter.type',
+    completions: [
+      { value: 'selector', documentation: 'Matches a specific dimension value' },
```
clintropolis, line 472: this is missing the [range filter](https://druid.apache.org/docs/latest/querying/filters/#range-filter) and typed in filter (which i just realized isn't documented yet, from #16039)

web-console/src/druid-models/native-json-query/native-json-query-completions.ts
```diff
@@ -0,0 +1,1098 @@
...
+      { value: 'populateCache', documentation: 'Whether to populate the cache' },
+      {
+        value: 'useResultLevelCache',
+        documentation: 'Whether to use result level cache',
+      },
+      {
+        value: 'populateResultLevelCache',
+        documentation: 'Whether to populate result level cache',
+      },
+      { value: 'bySegment', documentation: 'Return results by segment' },
```
clintropolis, line 721: does this still break the results if you actually set it?

web-console/src/views/load-data-view/ingestion-spec-completions.ts
```diff
@@ -0,0 +1,831 @@
...
+
+import type { JsonCompletionRule } from '../../utils';
+
+/**
+ * Determines if a type is a supervisor type based on the documentation:
+ * "The supervisor type. For streaming ingestion, this can be either kafka, kinesis, or rabbit.
+ * For automatic compaction, set the type to autocompact."
+ */
+function isSupervisorType(type: string): boolean {
+  return type === 'kafka' || type === 'kinesis' || type === 'rabbit' || type === 'autocompact';
```
clintropolis: missing the scheduled batch supervisor, #17353

web-console/src/druid-models/native-json-query/native-json-query-completions.ts
```diff
@@ -0,0 +1,1098 @@
...
+    condition: obj => obj.queryType === 'segmentMetadata',
+    completions: [
+      { value: 'cardinality', documentation: 'Number of unique values in string columns' },
+      { value: 'interval', documentation: 'Time intervals associated with queried segments' },
+      { value: 'minmax', documentation: 'Estimated min/max values for string columns' },
+      { value: 'size', documentation: 'Estimated byte size as if stored in text format' },
+      { value: 'timestampSpec', documentation: 'Timestamp specification of data in segments' },
+      { value: 'queryGranularity', documentation: 'Query granularity of data in segments' },
+      { value: 'aggregators', documentation: 'List of aggregators usable for metric columns' },
+      { value: 'rollup', documentation: 'Whether the segments are rolled up' },
```
clintropolis, line 238: #18119 recently added a `projection` analysis type

web-console/src/dialogs/compaction-config-dialog/compaction-config-completions.ts
```diff
@@ -0,0 +1,401 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { JsonCompletionRule } from '../../utils';
+
+export const COMPACTION_CONFIG_COMPLETIONS: JsonCompletionRule[] = [
+  // Root level properties
+  {
+    path: '$',
```
clintropolis, line 24: this isn't really correct, after #17803 there is a "type" for a datasource compaction config, with the old spec type being "inline" and the one added in that PR is "catalog"

clintropolis (commented):
oh man, that is a lot of API to review, i'll be real here and admit that so far i just scanned it and picked on the things that stuck out.

what are we going for here, is the thinking that mostly correct is better than nothing? or should we try to review this more closely to ensure it is accurate? what is the plan for maintenance, should PR authors update this when they update stuff? I don't have many strong opinions either way yet, just some thoughts that come to mind after looking over this.

All that said, this is pretty cool though 😎

web-console/src/druid-models/native-json-query/native-json-query-completions.ts
vogievetsky: Huh, it is not documented here: https://druid.apache.org/docs/latest/querying/datasourcemetadataquery/ I even double cheched this one. It kind of makes sense for `dataSourceMetadata` not to support `intervals` since it is reporting back the max ingested time.

web-console/src/views/load-data-view/ingestion-spec-completions.ts
vogievetsky: removed

web-console/src/views/load-data-view/ingestion-spec-completions.ts
vogievetsky: I'll add `systemFields` and will add `requestHeaders` when it is documented

web-console/src/druid-models/native-json-query/native-json-query-completions.ts
vogievetsky: I am sticking only to the documented stuff for now

web-console/src/druid-models/native-json-query/native-json-query-completions.ts
vogievetsky: `bySegment: true` should work :-)

web-console/src/views/load-data-view/ingestion-spec-completions.ts
vogievetsky: added

web-console/src/druid-models/native-json-query/native-json-query-completions.ts
vogievetsky: added

web-console/src/dialogs/compaction-config-dialog/compaction-config-completions.ts
vogievetsky: I'll add that when the docs are up

vogievetsky:
Thank you for the review and feedback @clintropolis in general I have taken the approach of only going after the documented properties. I also think of these JSON completions as best effort. It is ok for them to miss stuff. Even if they are totally wrong it is not a blocker (even though it is best to fix) as the completions are purely cosmetic.

As for updating I expect the completions to be updated like like the properties for rendering all the AutoForms. Ideally done by the dev adding a new property in their PR but more often done after the fact once that property is exposed and documented

web-console/lib/keywords.ts
```diff
@@ -19,129 +19,120 @@
...
+  'CUBE',
+  'CURRENT',
+  'DAY',
+  'DAY',
+  'DAYS',
+  'DECADE',
+  'DESC',
   'DISTINCT',
+  'DOW',
+  'DOW',
```
clintropolis: do dupes matter? also this doesn't work in druid (and a bunch of others in this list), though I guess calcite recognizes it since i see it in the parser stuff :shrug:

web-console/src/views/load-data-view/ingestion-spec-completions.ts
```diff
@@ -0,0 +1,904 @@
...
+        documentation: 'Force segments to be sorted by time first',
+      },
+    ],
+  },
+  // useSchemaDiscovery values
+  {
+    path: '$.spec.dataSchema.dimensionsSpec.useSchemaDiscovery',
+    completions: [
+      { value: 'true', documentation: 'Enable automatic type detection and schema discovery' },
+      { value: 'false', documentation: 'Use manual dimension specification (default)' },
```
clintropolis, line 148: this is perhaps misleading since you can still discover dimensions by leaving the list empty OR by specifying `includeAllDimension` as true.. though maybe that makes it even more confusing..

web-console/src/views/load-data-view/ingestion-spec-completions.ts
```diff
@@ -0,0 +1,904 @@
...
+      { value: 'double', documentation: 'Double precision dimension' },
+      { value: 'json', documentation: 'JSON/nested data dimension' },
+    ],
+  },
+  // createBitmapIndex values
+  {
+    path: '$.spec.dataSchema.dimensionsSpec.dimensions.[].createBitmapIndex',
+    completions: [
+      { value: 'true', documentation: 'Create bitmap index (default, faster filtering)' },
+      { value: 'false', documentation: 'No bitmap index (saves storage)' },
```
clintropolis, line 206: should we indicate that this only works for string type dimensions?

web-console/src/views/load-data-view/ingestion-spec-completions.ts
```diff
@@ -0,0 +1,904 @@
...
+    ],
+  },
+  // multiValueHandling values
+  {
+    path: '$.spec.dataSchema.dimensionsSpec.dimensions.[].multiValueHandling',
+    completions: [
+      { value: 'sorted_array', documentation: 'Sort multi-values (default)' },
+      { value: 'sorted_set', documentation: 'Sort and deduplicate multi-values' },
+      { value: 'array', documentation: 'Keep multi-values as-is' },
```
clintropolis, line 215: should we mention this only works for string type dimensions?

web-console/src/druid-models/native-json-query/native-json-query-completions.ts
```diff
@@ -0,0 +1,1102 @@
...
+    isObject: true,
+    completions: [
+      { value: 'type', documentation: 'Type of aggregation' },
+      { value: 'name', documentation: 'Output name for this aggregation' },
+    ],
+  },
+
+  // Aggregation types
+  {
+    path: '$.aggregations.[].type',
```
clintropolis, line 346: i've noticed this repeated quite a lot across the completions (and also some other common components like filters), is there a way to share it perhaps? fine to do as a follow-up, but sharing common elements across query, ingestion, compaction, etc would help make this a ton more maintainable

clintropolis (approved):
approving because i think this does more help than harm, but I still am a bit wary about long term maintainability. In its current state it seems like maybe we just have to like periodically ask claude (or some similar bot) to refresh it to match the docs? It would be really nice to like consolidate the common definitions across the completion files somehow so that human devs could potentially proactively add or adjust stuff here, but not a blocker.

Still, neat overall 🤘

vogievetsky:
I think the feedback all makes sense. I also think this PR makes things a lot better. In a follow up PR I will change this behaviour to use json-schema which is a standard and allows for recursive schema definition

PR #18127, opened 2025-06-11
Quidem: Use PathMatchers for path matching.

gianm:
This is more featureful than the previous approach. It supports wildcards like `*` (does not traverse subdirectories), `**` (traverses subdirectories), and `?` (single char).


sql/src/test/java/org/apache/druid/quidem/DruidQuidemTestBase.java
```diff
@@ -101,7 +102,7 @@ public abstract class DruidQuidemTestBase
   private static final String PROPERTY_FILTER = "quidem.filter";
 
   private final String filterStr;
-  private final List<Pattern> filterPatterns;
+  private final List<String> filterGlobs;
```
kgyrtkirk: I think this field should have the type  `PathMatcher` ; the fact that it might be an `OR` of matchers belongs to the complexity of that `PathMatcher` implementation

gianm: I don't think there is a way to combine two `PathMatcher` with boolean operations- at least I didn't see one. And internally, the `PathMatcher` does not recognize the format `x,y` as being an `OR` of `x` and `y`, which is a feature we currently support for `quidem.filter`. So I think this does have to be a list.

kgyrtkirk: it could be a small local class:
```
static class LocalPathMatcher implements PathMatcher {
    LocalPathMatcher(String pattern) { ... }
    List<PathMatcher> matchers;
}
```

gianm: Ah I see. I was looking for builtin JDK stuff. I made some edits to simplify things using lambdas to create the final `PathMatcher`.

kgyrtkirk:
it did sucked up the dispersing complexity I didn't liked that much - I still see a `LocalX` classes more usefull; they could for sure eat up more real estate on our screens - but they appear as 1st class citizens in the stack traces and debugging something like  `filterMatchers.stream().anyMatch(m -> m.matches(f))` is pretty inconvinient ; this might be just my own preferrence; in this case it for sure doesn't matter at all :D

thank you for the changes!

PR #18136, opened 2025-06-12
Add a preferred tier selection strategy

FrankChen021:
### Description

Use Case

If a Druid cluster is physically deployed in two AZs, we want the queries to be executed within one AZ as possible. 
This means queries from brokers to historicals should be executed on the nodes in the same AZ as possible. 
See the diagram as below:

image

This PR adds a new tier selector at broker side so that we can configure brokers to select historicals based on tier name(in this use case, the tier name can configured as the AZ name).
And if there're no servers in given tier(typically this is the case when historicals in one AZ are under maintance), brokers can still select historicals in another tier(AZ).

And to use this strategy, users also needs to make changes to their load rules to ensure segments are loaded in two tier(AZ)

This PR has:

- [x] been self-reviewed.


abhishekrb19:
@FrankChen021 thanks for the PR! Curious, how is this different from [broker tiering](https://druid.apache.org/docs/latest/operations/mixed-workloads/#broker-tiering)? Wouldn’t it be possible to achieve the same kind of preferred tier selection across AZs by configuring tiers on the brokers?

FrankChen021:
> @FrankChen021 thanks for the PR! Curious, how is this different from [broker tiering](https://druid.apache.org/docs/latest/operations/mixed-workloads/#broker-tiering)? Wouldn’t it be possible to achieve the same kind of preferred tier selection across AZs by configuring tiers on the brokers?

1. broker tiering requires the router to route the queries to different brokers, which introduces extra deployment/dependency. Since router is an optional component, there should be a way that the broker can handle routing.
2. broker tiering requires the client to set some query context parameters for routing. in another word, It's the client that is responsible for routing, which is not the case here
3. broker tiering does not provide a way to achieve fallback routing. If historical nodes in one AZ are down, we need to route queries to historicals nodes in another AZ

docs/configuration/index.md
```diff
@@ -1791,6 +1791,8 @@ These Broker configurations can be defined in the `broker/runtime.properties` fi
...
 |`druid.broker.select.tier.custom.priorities`|An array of integer priorities, such as `[-1, 0, 1, 2]`|Select servers in tiers with a custom priority list.|The config only has effect if `druid.broker.select.tier` is set to `custom`. If `druid.broker.select.tier` is set to `custom` but this config is not specified, the effect is the same as `druid.broker.select.tier` set to `highestPriority`. Any of the integers in this config can be ignored if there's no corresponding tiers with such priorities. Tiers with priorities explicitly specified in this config always have higher priority than those not and those not specified fall back to use `highestPriority` strategy among themselves.|
+|`druid.broker.select.tier.preferred.tier`| The preferred tier name. E.g., `_default_tier` | A non-empty value that specifies the preferred tier in which historical servers will be picked up for queries. If there are no enough historical servers from the preferred tier, servers from other tiers (if there are) will be selected. | null |
```
capistrant:
```suggestion
|`druid.broker.select.tier.preferred.tier`| The preferred tier name. E.g., `_default_tier` | A non-empty value that specifies the preferred tier in which historical servers will be picked up for queries. If there are not enough historical servers from the preferred tier, servers from other tiers (if there are any) will be selected. This config only has effect if `druid.broker.select.tier` is set to `preferred` | null |
```

docs/configuration/index.md
```diff
@@ -1791,6 +1791,8 @@ These Broker configurations can be defined in the `broker/runtime.properties` fi
 |`druid.broker.balancer.type`|`random`, `connectionCount`|Determines how the broker balances connections to Historical processes. `random` choose randomly, `connectionCount` picks the process with the fewest number of active connections to|`random`|
 |`druid.broker.select.tier`|`highestPriority`, `lowestPriority`, `custom`|If segments are cross-replicated across tiers in a cluster, you can tell the broker to prefer to select segments in a tier with a certain priority.|`highestPriority`|
```
capistrant:
```suggestion
|`druid.broker.select.tier`|`highestPriority`, `lowestPriority`, `custom`, `preferred`|If segments are cross-replicated across tiers in a cluster, you can tell the broker to prefer to select segments in a tier with a certain priority.|`highestPriority`|
```

docs/configuration/index.md
```diff
@@ -1791,6 +1791,8 @@ These Broker configurations can be defined in the `broker/runtime.properties` fi
...
+|`druid.broker.select.tier.preferred.tier`| The preferred tier name. E.g., `_default_tier` | A non-empty value that specifies the preferred tier in which historical servers will be picked up for queries. If there are no enough historical servers from the preferred tier, servers from other tiers (if there are) will be selected. | null |
+|`druid.broker.select.tier.preferred.priority`| `highest`, `lowest` | Optional. If there are multiple candidates in a preferred tier, specifies the priority to pick up candidates. By default, the higher priority a historical, the higher chances it will be picked up. | `highest` |
```
capistrant:
```suggestion
|`druid.broker.select.tier.preferred.priority`| `highest`, `lowest` | If there are multiple candidates in a preferred tier, specifies the priority to pick up candidates. By default, the higher priority a historical, the higher chances it will be picked up. This config only has effect if `druid.broker.select.tier` is set to `preferred`| `highest` |
```

server/src/test/java/org/apache/druid/client/selector/TierSelectorStrategyTest.java
```diff
@@ -260,17 +261,214 @@ public void testServerSelectorStrategyDefaults()
...
+        defaultDeprecatedServerSelectorStrategy.pick(servers, EasyMock.createMock(DataSegment.class)),
+        p0
+    );
+    Assert.assertEquals(
+        defaultDeprecatedServerSelectorStrategy.pick(servers, EasyMock.createMock(DataSegment.class), 1)
+                                               .get(0), p0
+    );
+  }
+
+  private void testPreferredTierSelectorStrategy(
```
capistrant, line 300: can you add a javadoc specifying a list of what things are being validated in here?

server/src/main/java/org/apache/druid/client/selector/ServerSelectorStrategy.java
```diff
@@ -38,26 +38,26 @@
 public interface ServerSelectorStrategy
 {
   @Nullable
-  default <T> QueryableDruidServer pick(@Nullable Query<T> query, Set<QueryableDruidServer> servers, DataSegment segment)
+  default <T> QueryableDruidServer pick(@Nullable Query<T> query, Collection<QueryableDruidServer> servers, DataSegment segment)
```
capistrant: It is not clear to me why we are removing the `Set` restriction here? Your PR doesn't seem to introduce a change that requires us to now allow duplicate servers passed to `pick`.. or does it? Maybe I'm missing something.

FrankChen021: in the `PreferredTierSelectorStrategy.pick`, a List is used to save candidate servers. So the `Set` here needs to be changed to `Collection` so that both `List` and original `Set` work.

server/src/test/java/org/apache/druid/client/selector/TierSelectorStrategyTest.java
FrankChen021: added

server/src/test/java/org/apache/druid/client/selector/TierSelectorStrategyTest.java
```diff
@@ -260,17 +261,223 @@ public void testServerSelectorStrategyDefaults()
...
+    );
+    QueryableDruidServer preferredTierHighPriority = new QueryableDruidServer(
+        new DruidServer("test2", "localhost", null, 0, ServerType.HISTORICAL, "preferred", 1),
+        client
+    );
+    QueryableDruidServer anotherTierHighPriority = new QueryableDruidServer(
+        new DruidServer("test3", "localhost", null, 0, ServerType.HISTORICAL, "tier1", 2),
+        client
+    );
+    QueryableDruidServer yetAnotherTierMediumPriority = new QueryableDruidServer(
```
capistrant, line 529: does this introduce randomness to the tests adding a second server of equal priority? That is at least what it seemed like on my local. I got intermittent test failures for this test because the two servers with priority of 1 were flip flopping in actual return order.

server/src/main/java/org/apache/druid/client/selector/PreferredTierSelectorStrategy.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    // Fill with non-preferred servers if we don't have enough preferred servers
+    int fillSize = numServersToPick - preferred.size();
+    for (int i = 0; i < fillSize && i < nonPreferred.size(); i++) {
+      preferred.add(nonPreferred.get(i));
+    }
+
+    return this.serverSelectorStrategy.pick(query, preferred, segment, numServersToPick);
+  }
```
capistrant: Upon further review, I don't think that this is working as you designed. When I check out your changes on my local and run tests in loop, I get lots of failures.

I think the best approach here is to split all of the incoming servers into two Int2ObjectMaps that are sorted by priority. One of them contains all of the preferred servers. The other, the non-preferred servers. You then delegate to the underlying priority selector to do the actual pick for you from the preferred first, and then the non-preferred second (if necessary). I have a potential implementation below in suggestion. I'm not saying this is how it needs to be, but I couldn't think of how to explain what I was thinking in a way that is more clear since it is a little bit goofy 

side effect of this change would be to revert the ServerSelectorStrategy back to using Set instead of less restrictive Collection, since we no longer call it with lists when delegating pick to the priority strategy.

```suggestion
 @Override
  public <T> List<QueryableDruidServer> pick(
      Query<T> query,
      Int2ObjectRBTreeMap<Set<QueryableDruidServer>> prioritizedServers,
      DataSegment segment,
      int numServersToPick
  )
  {
    if (log.isDebugEnabled()) {
      log.debug(
          "Picking [%d] servers from preferred tier [%s] for segment [%s] with priority [%s]",
          numServersToPick, preferredTier, segment.getId(), this.priorityStrategy.getClass().getSimpleName()
      );
    }

    int numPreferredServers = 0;
    Int2ObjectRBTreeMap<Set<QueryableDruidServer>> preferred = new Int2ObjectRBTreeMap<>(priorityStrategy.getComparator());
    Int2ObjectRBTreeMap<Set<QueryableDruidServer>> nonPreferred = new Int2ObjectRBTreeMap<>(priorityStrategy.getComparator());
    for (Set<QueryableDruidServer> priorityServers : prioritizedServers.values()) {
      if (numPreferredServers == numServersToPick) {
        // We have enough preferred servers to satisfy the request, no need to check further
        break;
      }
      for (QueryableDruidServer server : priorityServers) {
        if (preferredTier.equals(server.getServer().getMetadata().getTier())) {
          numPreferredServers++;
          preferred.computeIfAbsent(
              server.getServer().getPriority(),
              p -> new HashSet<>()
          );
          preferred.get(server.getServer().getPriority()).add(server);
          if (numPreferredServers == numServersToPick) {
            // We have enough preferred servers to satisfy the request, no need to check further
            break;
          }
        } else {
          // We have to iterate through all servers even the numbers of the non-preferred servers reach the limit
          // This is because we don't know whether there are preferred servers left in the next priority set
          nonPreferred.computeIfAbsent(
              server.getServer().getPriority(),
              p -> new HashSet<>()
          );
          nonPreferred.get(server.getServer().getPriority()).add(server);
        }
      }
    }
    List<QueryableDruidServer> picks = new ArrayList<>(numServersToPick);
    if (numPreferredServers > 0) {
      // If we have preferred servers, pick them first
      picks.addAll(priorityStrategy.pick(query, preferred, segment, numServersToPick));
    }
    if (picks.size() < numServersToPick) {
      // If we don't have enough preferred servers, pick from the non-preferred ones
      int remaining = numServersToPick - picks.size();
      picks.addAll(priorityStrategy.pick(query, nonPreferred, segment, remaining));
    }
    return picks;
  }
```

capistrant (commented):
After pulling this down to my local and running the UTs you added on repeat, I got lots of test fails and feel like it points to the PR as is not acting completely how you want it to. Please see my comments and let me know if I'm on the right track or if there is some misunderstanding I'm having. Thanks!

server/src/test/java/org/apache/druid/client/selector/TierSelectorStrategyTest.java
FrankChen021: it looks like two servers with same priority bring different results with multiple runs. 
fixed by setting different priorities for this cases to make sure the test result is stable.

server/src/main/java/org/apache/druid/client/selector/PreferredTierSelectorStrategy.java
FrankChen021: the failures should be related to the 'shuffle' method call in the test cases. I didn't see these failure nor did I see failure on the CI, but indeed, with multiple runs, there're failures. 

the implementation you proposed I think is the same as the orignal one but keeps the `ServerSelectorStrategy` unchanged with `Set` parameter.

capistrant: Is it the same? This is the line I'm most worried about in your patch:

`return this.serverSelectorStrategy.pick(query, preferred, segment, numServersToPick);`

That is after you have potentially put non-preferred nodes into a list with preferred nodes. The serverSelector strategy does not care about anything other than either picking a random node or the node with the least conns.

Secondarily it also adds nodes with differing priorites to that list. So even within the preferred nodes only, isn't it giving the server selector the chance do the wrong priority strategy?

FrankChen021: I don't think this is a problem.

Firstly, if you look at the `AbstractTierSelectorStrategy`, it provides an override version of `pick` that limits the `numServersToPick` to 1. And there's no other place in the codebase that call the `pick` method with a explicit `numServersToPick`. This means there won't be preferred+non-preferred cases.

Secondly, even if prefered and non-prefered are returned:
1. since the `preferred` list has a size of `numServersToPick` and it requires `serverSelectorStrategy` to return a list that has size of `numServersToPick`, the  `serverSelectorStrategy`(whether it's the randome or connection count based) returns a list of all given servers in the `prerferred` list without any change.

2. in your version, the `pick` variable also holds both preferred and non-prefered servers, what do you expect the caller to handle these combined servers? It's also the caller that decides how to handle these candidates, the prefered in this case is also not guarateed.

FrankChen021: your comments make me think further there's indeed a problem, not the one you mention but the one we exit the processing too early when the `preferred` size reaches the `numServersToPick`.

capistrant: ~Fair point on the numServersToPick meaning we are always going to pick all of the servers. what about the ordering of the servers returned? in highest priority and lowest priority tier selector it looks like it will always return in an order. is there any language in the docs or code that says this will be an ordered list returned to caller?~

Edit: didn't see the new changes before this comment was written. Not as applicable with the new change

server/src/main/java/org/apache/druid/client/selector/AbstractTierSelectorStrategy.java
```diff
@@ -34,7 +34,7 @@
  */
 public abstract class AbstractTierSelectorStrategy implements TierSelectorStrategy
 {
-  private final ServerSelectorStrategy serverSelectorStrategy;
+  protected final ServerSelectorStrategy serverSelectorStrategy;
```
capistrant, line 37: I think this can be restored to private now that PreferredTierSelectorStrategy is delegating to highest or lowest priority strategy instead of calling server selector direct

capistrant:
> Left one stand-a-lone comment on flipping an instance var back to private instead of protected (matching pre PR state). That would be cool to include before merging, but is a little bit of a nit.

@FrankChen021 if you could submit a 1 line follow up for future releases to include, that would be great

server/src/main/java/org/apache/druid/client/selector/AbstractTierSelectorStrategy.java
FrankChen021: reverted. thanks for pointing out

FrankChen021:
@capistrant You merged this right before I reverted the change you pointed out😄. let me submit in a follow up pr.

cryptoe:
@FrankChen021 . Could you please update the release notes with the following information : 
* What config changes are required to enable this feature
* How does the user benefit from this feature. (this is partly there)
* Known issues/drawbacks.

FrankChen021:
> @FrankChen021 . Could you please update the release notes with the following information :
> 
> * What config changes are required to enable this feature

To enable it, configure `druid.broker.select.tier` to `perferred` at broker side and configure historical nodes into different tiers by setting `druid.server.tier`

> * How does the user benefit from this feature. (this is partly there)

This is useful for across availability zone deployment. Brokers in one AZ select historicals in the same AZ by default but still keeps the ability to select historical nodes in another AZ if historicals in the same AZ are not available.

PR #18144, opened 2025-06-15
MSQ: Enable stages to control their own shuffle logic.

gianm:
This patch pushes shuffle logic from RunWorkOrder into the stage definition. The motivation is to enable optimizations that take advantage of data already being sorted or partitioned in the desired way.

Main changes:

- FrameProcessorFactory is renamed StageProcessor, and its "makeProcessors" method is replaced with a more powerful "execute" method. The new "execute" is expected to fully contain partitioning and shuffling logic.

- RunWorkOrder, the worker class that manages execution, no longer has any shuffle-related processing code. It delegates this to the StageProcessor.

- StandardStageOperations, StandardStageRunner, & StandardStageProcessor are introduced for StageProcessor that want to perform shuffling in the "standard" way that was formerly mandatory.

Other changes:

- Moved FrameContext and related classes from the "kernel" package to the "exec" package.

- Remove "readableChannelUsableWhenWriting" from OutputChannel, to reduce complexity. It was only used by the standard hashPartition implementation, and wasn't strictly needed there.


extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ExecutionContext.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.druid.msq.kernel.ShuffleKind;
+import org.apache.druid.msq.kernel.WorkOrder;
+import org.apache.druid.msq.statistics.ClusterByStatisticsSnapshot;
+
+import javax.annotation.Nullable;
+
+/**
+ * All the things needed for {@link StageProcessor#execute(ExecutionContext)} to run the work for a stage.
+ */
+public interface ExecutionContext
```
kgyrtkirk, line 40: I think this interface might be unnecessary: it declares 12 methods and has only a single implementation - its already seem to have suffered from some bending from that single implementation.

might be better to just drop the interface and introduce generalization when there will be at least 2 implementors

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ExecutionContext.java
```diff
@@ -0,0 +1,107 @@
...
+   * (2) With {@link ShuffleKind#GLOBAL_SORT}, this provides statistics that are used to determine global boundaries.
+   *
+   * Typically called by {@link StandardShuffleOperations#gatherResultKeyStatisticsIfNeeded(ListenableFuture)}.
+   */
+  void onDoneReadingInput(@Nullable ClusterByStatisticsSnapshot snapshot);
+
+  /**
+   * Callback to report a nonfatal warning.
+   */
+  void onWarning(Throwable t);
```
kgyrtkirk, line 113: note: in the scope of an `ExecutionContext` these callback like methods look a little bit like outliers

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageProcessor.java
```diff
@@ -78,23 +83,31 @@ ProcessorsAndChannels<T, R> makeProcessors(
...
+          context.workOrder().getWorkerNumber(),
+          context.workOrder().getInputs(),
+          context.inputSliceReader(),
+          (ExtraInfoType) context.workOrder().getExtraInfo(),
+          stageRunner.workOutputChannelFactory(),
+          context.frameContext(),
+          context.threadCount(),
+          context.counters(),
+          context::onWarning,
+          MultiStageQueryContext.removeNullBytes(context.workOrder().getWorkerContext())
```
kgyrtkirk, line 104: note: around 10 arguments are constructed one way or another from the `context` argument....
I think it would be more straight forward to just pass the `ExecutionContext` and let the other end decide what it needs...

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageRunner.java
```diff
@@ -0,0 +1,224 @@
...
+  @MonotonicNonNull
+  private ListenableFuture<ResultAndChannels<Object>> pipelineFuture;
+
+  public StandardStageRunner(final ExecutionContext executionContext)
+  {
+    this.executionContext = executionContext;
+    this.workOrder = executionContext.workOrder();
+    this.counterTracker = executionContext.counters();
+    this.exec = executionContext.executor();
+    this.cancellationId = executionContext.cancellationId();
```
kgyrtkirk: there are a lot of localized stuff from the `executionContext` here which are only used 1/2 times...I think the field might just add complexity

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageRunner.java
```diff
@@ -0,0 +1,224 @@
...
+      maxOutstandingProcessors =
+          Math.max(1, Math.min(threadCount, processors.getOutputChannels().getAllChannels().size()));
+    }
+
+    final boolean usesProcessingBuffers = workOrder.getStageDefinition().getProcessor().usesProcessingBuffers();
+
+    workResultFuture = exec.runAllFully(
+        counterTracker.trackCpu(processorManager, CpuCounters.LABEL_MAIN),
+        maxOutstandingProcessors,
+        usesProcessingBuffers ? frameContext.processingBuffers().getBouncer() : Bouncer.unlimited(),
```
kgyrtkirk: a method like `executionContext.getBouncer` could have access to everything this line needs

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageRunner.java
```diff
@@ -0,0 +1,224 @@
...
+      // If there are output channels, that acts as a ceiling on the number of processors that can run at once.
+      maxOutstandingProcessors =
+          Math.max(1, Math.min(threadCount, processors.getOutputChannels().getAllChannels().size()));
+    }
+
+    final boolean usesProcessingBuffers = workOrder.getStageDefinition().getProcessor().usesProcessingBuffers();
+
+    workResultFuture = exec.runAllFully(
+        counterTracker.trackCpu(processorManager, CpuCounters.LABEL_MAIN),
+        maxOutstandingProcessors,
```
kgyrtkirk, line 145: a method like `executionContext.getMaxProcessingThreads`  could be usefull here

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageRunner.java
```diff
@@ -0,0 +1,224 @@
...
+
+        if (shuffleSpec.partitionCount() == 1) {
+          // Single partition; no need to write temporary files.
+          hashOutputChannelFactory =
+              new BlockingQueueOutputChannelFactory(frameContext.memoryParameters().getFrameSize());
+          hashOutputBuffered = false;
+        } else {
+          // Multi-partition; write temporary files and then sort each one file-by-file.
+          hashOutputChannelFactory = executionContext.makeIntermediateOutputChannelFactory("hash-parts");
+          hashOutputBuffered = true;
```
kgyrtkirk: note: couldn't the `hashOutputChannelFactory` tell the `hashPartition()` somehow about this fact? the check on `1` and the other `!= MEMORY` is just odd

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/RunWorkOrder.java
```diff
@@ -215,23 +172,11 @@ public void startAsync()
...
-
-      if (stageDef.doesShuffle()) {
-        makeAndRunShuffleProcessors();
-      } else {
-        // No shuffling: work output _is_ stage output. Retain read-only versions to reduce memory footprint.
-        stageOutputChannelsFuture =
-            Futures.immediateFuture(workResultAndOutputChannels.getOutputChannels().readOnly());
-      }
-
+      initInputSliceReader();
```
kgyrtkirk: I think the `ExecutionContext` is created to late; for example this `initInputSliceReader` method is called twice - here and from when that context is created.

I feel like creating it earlier could enable to clean up some of this logic

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/RunWorkOrder.java
```diff
@@ -215,23 +172,11 @@ public void startAsync()
...
+      initGlobalSortPartitionBoundariesIfNeeded();
```
kgyrtkirk, line 177: I feel like something is leaking here...not sure how to fix it - but this and the stuff it moves doesn't seem to belong here...I'll try to dig into this

kgyrtkirk (commented):
I've not finished going thru all parts...so far it was interesting :)
just left some comments here and there :)

Due to the fact that classes were renamed/moved its harder to focus on the core aspects of the changes - since those changes have touched a lot of files I'm a bit worried if the PR enters a conflicted change after some other changes will be merged.

I think it doesn't change fundamentally a lot of things right now - is that right?
but is it on the roadmap to take away the `shuffleSpec` from the `StageDefintion` ?

gianm:
@kgyrtkirk thank you for taking a look.

> Due to the fact that classes were renamed/moved its harder to focus on the core aspects of the changes - since those changes have touched a lot of files I'm a bit worried if the PR enters a conflicted change after some other changes will be merged.

There are a lot of renames, but most of them are just renaming things from `XFrameProcessorFactory` to `XStageProcessor`.

 
> I think it doesn't change fundamentally a lot of things right now - is that right?

Yes, the idea in this patch was to not change anything fundamentally, especially with the existing stage implementations. The main change is that in the future, stages can control their own shuffling logic, rather than being "forced" into having the standard approach applied.

> but is it on the roadmap to take away the `shuffleSpec` from the `StageDefintion` ?

I don't plan to remove `shuffleSpec` from `StageDefinition`. It's still needed for a couple of reasons:

- the MSQ framework still does determine how many workers should be launched, when they should be launched, and which workers should handle which partitions. It needs the `shuffleSpec` to help figure this out.
- for the `GLOBAL_SORT` shuffle, the MSQ framework still determines the partition boundaries based on statistics sent from each worker. It needs the `shuffleSpec` to know it should activate this mode.

For this reason, I have this text in the javadoc for `StageProcessor`:

```
 * For stages that shuffle, i.e. where {@link StageDefinition#doesShuffle()}, the outputs must be partitioned according
 * to the {@link StageDefinition#getShuffleSpec()}.
```

The idea is that the `shuffleSpec` still exists, and the stage processor must respect it. The `StandardStageRunner` and `StandardShuffleOperations` are provided for stages that want to do things the "standard" way. In the future I imagine some stages will have an optimized path for certain cases of `shuffleSpec`, and then fallback to the `StandardShuffleOperations` for other cases.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/RunWorkOrder.java
gianm: Do you mean the `initGlobalSortPartitionBoundariesIfNeeded()`?

The situation with global sort boundaries has some back-and-forth between the worker and controller, which leads to some complexity. The flow of partitioned global sort is like this:

- Each worker gathers statistics from its input as it is read. While doing so it buffers the input.
- Each worker sends those statistics to the controller when done reading input.
- The controller gathers statistics from each worker, uses them to determine good global partition boundaries, then sends those boundaries to all workers. (Crucially each worker gets the same set of boundaries)
- The workers partition data according to the boundaries and sorts within each partition. When each partition is later merged across workers, this yields a total sort of the data.

This means that some stuff related to global partition boundaries is going to leak into the `ExecutionContext`:

- there has to be a way for the worker to send its statistics (`onDoneReadingInput`)
- there has to be a way for the worker to receive the global partition boundaries (`globalClusterByPartitions`)

Coming back to the `initGlobalSortPartitionBoundariesIfNeeded` you highlighted: this is for the case where the controller isn't involved, possibly because there's only one partition. In this case the worker does not send statistics nor does it receive global boundaries from the controller. It can know its boundaries on its own. (It will just be "all data goes to partition 0").

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/RunWorkOrder.java
gianm: Oops, that was a mistake, I only meant for it to be called once. Will fix this.

gianm: I just deleted this call. Only the one inside `makeExecutionContext` was needed.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageProcessor.java
gianm: I did it this way to avoid changing too much code in this PR. Changing the signature of `makeProcessors` would require changes to all the subclasses, which are many. In a future patch I think it would make sense to remove this entire `StandardStageProcessor` class and rewrite the subclasses to directly implement `execute`. I left a note about this in the class-level javadoc:

```
 * This abstract class may be removed someday, in favor of its subclasses using {@link StandardStageRunner} directly.
 * It was introduced mainly to minimize code changes in a refactor.
```

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageRunner.java
gianm: I just ended up removing all of these except for `executionContext`.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageRunner.java
gianm: Added one named `processingBouncer()`.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageRunner.java
gianm: This logic here is specific to the `StandardStageRunner` -- it wouldn't apply to `StageProcessor` implementations that want to do their own thing. They would just need the `threadCount()` that is already there. So I left this as-is.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ExecutionContext.java
gianm: I suppose these could be collapsed. I don't currently have plans to create other implementations. I had a couple reasons for doing an interface:

- I thought it would be easier for developers of `StageProcessor` to understand what they get for `execute` if they could read an interface, rather than a concrete class. The code for a concrete class has a bunch of incidental implementation stuff in it that isn't necessarily helpful to its user.
- I thought it might be useful for testing to have the interface and concrete class be separate things. Although, I didn't end up taking advantage of this. I suppose tests could use mocks too.

I left this as is, because IMO the interface approach is cleaner, but it's kind of borderline so I could go either way.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardStageRunner.java
gianm: I cleaned this up by adding a method `OutputChannelFactory#isBuffered` and using that instead.

gianm:
> I feel like the `result` and the `outpuchannel` should be more closer to eachother...
> there are multiple classes which has an `And` in their name. I feel like some concept is missing...as those are essentially try to repack those into a single object - I suspect that the fact that `Future`-s are used doesn't make it easier either.

I pushed a couple commits resolving merge conflicts, and also adding various javadocs that I hope clarifies the usage of the "result" and "channels" concepts. The main reason for all the future-stuff is that the "result" is available only after all processing is fully complete, whereas the "output channels" may be available for reading prior to processing being complete (when data is streamed between two concurrently-running stages).

Thank you for the review, I'll merge after the tests pass.

PR #18147, opened 2025-06-16
Add embedded Druid cluster tests

kfaraz:
### Summary 

- ⏳ Run integration tests without the need to `mvn build`, create a Docker image or even run startup scripts 
- 🚀 Just write a JUnit `@Test` and run it from your IDE! 🎉 🎉 
- 🔡 Run tests against a fully functioning Druid cluster running in a single JVM
- 🔧 Complete control over all aspects of the Druid services being tested
- 🎩 Load any combination of Druid extensions
- 🐞 Debug the test and the services using debug points and the web-console!
- 📦 Get embedded ZooKeeper + Derby metadata store by default (non embedded versions are supported too)

### Explore

💥 Run the [`EmbeddedKafkaClusterMetricsTest`](https://github.com/apache/druid/blob/5a8dcd5f38decbc1b8a9fa2a9bd8805ed586ca57/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedKafkaClusterMetricsTest.java) to see a simulation in action:
- Start a Druid cluster which uses the `KafkaEmitter` to publish cluster metrics to a topic
- Post a `KafkaSupervisor` to ingest the metrics back into the __same__ Druid cluster! 🐶 🍖
- Put debug points in the test to keep the web-console running and explore the metrics!

### Usage

1.  ✏️ Write a JUnit 5 test class that extends `EmbeddedClusterTestBase`
2. 🔩 Define your cluster in the method `createCluster()`
3. 🧪 Write a `@Test` method
4. 🚀 Run it!

</details>

<details>
<summary><h4>Define a cluster</h4></summary>

```java
EmbeddedDruidCluster cluster = EmbeddedDruidCluster.withEmbeddedDerbyAndZookeeper();
cluster.add(new EmbeddedOverlord());
cluster.add(new EmbeddedIndexer());
```

</details>
<details>
<summary><h4>Add an extension module</h4></summary>

```java
cluster.addExtension(KafkaIndexTaskModule.class);
```
Requirements:
- Module must extend `DruidModule`
- Module (and corresponding `META-INF/services` file) must be on the classpath of the test
   - EITHER add a dependency to the corresponding maven module containing (e.g. maven module for `KafkaIndexTaskModule` is `druid-kafka-indexing-service`)
   - OR add your test in the same maven module

Example test:

https://github.com/apache/druid/blob/5a8dcd5f38decbc1b8a9fa2a9bd8805ed586ca57/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/EmbeddedKafkaSupervisorTest.java#L63-L78

</details>
<details>
<summary><h4>Add common properties</h4></summary>

```java
cluster.addCommonProperty("druid.monitoring.emissionPeriod", "PT1S");
```
These properties correspond to the file `common.runtime.properties` that are applied on all the Druid services.

</details>
<details>
<summary><h4>Add service-specific properties</h4></summary>

```java
EmbeddedOverlord overlord = new EmbeddedOverlord();
overlord.addProperty("druid.manager.segments.useIncrementalCache", "always");
```
These properties override the common runtime properties and correspond to the service-specific `runtime.properties` files.

</details>
<details>
<summary><h4>Add a cluster-managed resource</h4></summary>

```java
cluster.addResource(new EmbeddedZookeeper());
```
Resources are
- anything outside Druid services used by the cluster during its lifecycle.
- started and stopped with the cluster.
- implementations of the interface `EmbeddedResource`.

e.g. `EmbeddedKafkaServer`, `EmbeddedZookeeper`, `InMemoryDerbyResource`

</details>

### Motivation 

Druid ITs can be cumbersome to write, run and debug. They are very resource intensive , require multiple steps even to verify a simple change and are difficult to get right.

Owing to this barrier in writing new tests, Druid IT coverage of certain areas has been low.

This patch aims to address this issue by introducing a new framework to run lightweight integration tests using a fully-functioning Druid cluster running in a single process.

### Goals

Simulation tests aim to:
- 🦸🏻 Increase developer productivity
- 🏗️ Fill the test coverage gap between class-level unit tests and docker-based integration tests.
- ✅ Simplify testing out new features on a full Druid cluster.
- 🍰 Easily define different cluster setups in the test itself without needing to rely on a bunch of config files.
- 🔧 Provide complete control over all aspects of the service(s) being tested.
- 🧵 Test scalability and concurrency handling in critical pieces like `TaskQueue`, `TaskLockbox`, `TaskRunner`, etc.
- 🐞 Debug failures quickly (a typical cluster takes < 1s to start and can then be used for any number of tests).
- 🐎 Discover race conditions due to the fast nature of the test (already discovered some while putting this up 🙂 ).
- 😈 Use real Druid services and no mocked dependencies.

### Non-goals

Embedded cluster tests are not meant to:
- Replace docker-based ITs (standard or revised)
- Test running a multi-process Druid cluster 
- Test Druid docker image

### Design

- Run required Druid services in an embedded mode, that internally launch the corresponding `Cli*` runnable
- Each Druid service is sandboxed and has its own separate Guice injector
- No override of default Guice dependencies provided by the `Cli*` runnables
- No mocks
- Write out task logs and segments to temp directory
- Choose extensions to load via config `druid.extensions.modulesForEmbeddedTest`
- Add custom resources to the Druid cluster (e.g. `EmbeddedKafkaServer`, `EmbeddedZookeeper`)

### Changes to review in core Druid

- Wire up `JettyServerInitializer` properly so that multiple Jetty servers may be bound correctly within the same JVM
  - See [`JettyServerInitUtils.getGuiceFilterHolder`](https://github.com/apache/druid/blob/bdbd0f6d4daa771454d691f23cb42db59201b481/server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerInitUtils.java#L156)
- Add new test-only config `druid.extensions.modulesForEmbeddedTest` which takes in a list of module class names that
should be loaded in a simulation test
- Minor fix in `SeekableStreamTaskRunner` to allow `taskDuration` to be less than 1 minute
- Add method `OverlordClient.postSupervisor` to hit API `POST /druid/indexer/v1/supervisor`
- Update methods in test classes `TestDerbyConnector` and `CuratorTestBase` to have public access

### New classes to review 

- `EmbeddedDruidCluster`: defines a Druid cluster used in an embedded cluster test
- `EmbeddedResource`: defines any resource with a lifecycle that can be used in an `EmbeddedDruidCluster`
- `EmbeddedDruidServer` that can provide a `DruidServerResource` with a lifecycle
- `EmbeddedZookeeper`: uses the existing `CuratorTestBase` to run an embedded zk server
- `EmbeddedKafkaServer`: an embedded Kafka server mostly copying code from `TestBroker` utility class. We can merge these two in a follow up PR.
- JUnit 5 base test class `EmbeddedClusterTestBase`
- `LatchableEmitter` to wait for metric events using countdown latches rather than repeatedly polling Druid APIs

<hr>

This PR has:

- [ ] been self-reviewed.


abhishekagarwal87:
Great work @kfaraz. How do extensions get enabled/disabled in these embedded tests? Are those supported yet?

kfaraz:
Thanks, @abhishekagarwal87 !

**Edit:**
Using extensions is already supported.
- We can load any combination of extension modules as long as they are on the class path. Since these are plain UTs, that simply means adding a maven dependency to the respective extension module.
- Required modules should be specified in the cluster definition. These are then populated in `druid.extensions.modulesForSimulation` (as opposed to production which uses `druid.extensions.loadList`).
- We can also add custom resources to an embedded cluster such as an `EmbeddedKafkaServer`

A test in this PR is already using the Kafka extension as a proof-of-concept.

The syntax would be something like:

```java
EmbeddedKafkaServer kafkaServer = new EmbeddedKafkaServer();
EmbeddedDruidCluster cluster = EmbeddedDruidCluster.withExtensions(List.of(KafkaIndexTaskModule.class))
                                                   .addServer(new EmbeddedCoordinator())
                                                   .addServer(new EmbeddedIndexer())
                                                   .addServer(new EmbeddedOverlord())
                                                   .addResource(kafkaServer);
```

<hr>

To nicely wire up the dependencies and the SDK, I am currently evaluating two alternatives described below.

I am leaning towards B. What are your thoughts?

#### Option A: Keep the test framework in a separate module `simulation-tests` 

- Currently implemented in this PR
- All simulation tests would live in this module itself
- When we want to write a simulation test for an extension, say Kinesis indexing service,
`simulation-tests` module would need to depend on that extension.

Pros:
- The framework, embedded resources for extensions and tests all live in one module.
- Embedded clusters can use any mix of extensions (e.g. say MySQL metadata store + Kafka ingestion)

#### Option B: Keep the test framework in `services/test` package

- This is highest level Druid core module containing the `Cli*` runnables
- Write simulations for core Druid in `services/test` itself
- Extensions can then depend on `services/test` and use the SDK to define their own cluster resources
e.g. Kafka server (embedded or otherwise) and run tests using these resources

Pros:

- More maintainable in the long run
- Keep the simulation tests in the same package as the code being tested
- Simulations will be added organically as a part of feature development
- Simulations count towards Jacoco coverage for extensions
- Clean maven dependencies

Cons:

- Embedded cluster would be able to use only one extension at a time since extensions don't seem
to be able to depend on each other
- __Possible remedy__: Add a module that depends on multiple extensions. Simulation tests that use multiple extensions would have to live in this module. A module which may already serve this purpose is `integration-tests`.

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/DruidServerJunitResource.java
```diff
@@ -0,0 +1,181 @@
...
+import com.google.inject.Module;
+import org.apache.druid.cli.ServerRunnable;
+import org.apache.druid.guice.StartupInjectorBuilder;
+import org.apache.druid.java.util.common.concurrent.Execs;
+import org.apache.druid.java.util.common.lifecycle.Lifecycle;
+import org.apache.druid.java.util.common.logger.Logger;
+import org.apache.druid.metadata.TestDerbyConnector;
+import org.apache.druid.utils.JvmUtils;
+import org.apache.druid.utils.RuntimeInfo;
+import org.junit.rules.ExternalResource;
```
kgyrtkirk: for new frameworks I think it would be best to utilize `junit5`

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/EmbeddedOverlord.java
```diff
@@ -0,0 +1,303 @@
...
+   * Overrides {@link HttpRemoteTaskRunnerFactory} to be able to register the
+   * {@link #taskRunnerListener} on the created {@link HttpRemoteTaskRunner}.
+   */
+  private static class TestHttpRemoteTaskRunnerFactory extends HttpRemoteTaskRunnerFactory
+  {
+    private final TaskRunnerListener listener;
+
+    @Inject
+    public TestHttpRemoteTaskRunnerFactory(
+        @Smile final ObjectMapper smileMapper,
```
kgyrtkirk: maybe use a delegate instead a subclass?

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/EmbeddedDruidServer.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+  }
+
+  /**
+   * Holder for the service client instances that are being used by this server.
+   */
+  private static class ServiceClientHolder
+  {
+    @Inject
+    CoordinatorClient coordinator;
```
kgyrtkirk: I don't really understand why this injection will happen - an instance of it is only present as a private field of this class - could you give a hint? :)

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/EmbeddedBroker.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.druid.query.DruidProcessingConfigTest;
+import org.apache.druid.utils.RuntimeInfo;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Embedded mode of {@link CliBroker} used in simulation tests.
+ */
+public class EmbeddedBroker extends EmbeddedDruidServer
```
kgyrtkirk: this is the framework part: I think these classes should be under `main` of this module

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/EmbeddedOverlord.java
kfaraz: Yeah, I had started out with a delegate but then ended up writing a sub-class. I don't seem to remember why. 😅 
Let me fix it up.

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/EmbeddedDruidServer.java
kfaraz: Guice takes care of since we have bound it using `.toInstance()` in the method `getInitModules()`.

```java
binder.bind(ServiceClientHolder.class).toInstance(clientHolder)
```

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/EmbeddedBroker.java
kfaraz: Yes, I originally meant to do that but now I am wrestling with the options in this comment:
https://github.com/apache/druid/pull/18147#issuecomment-2979598062

What do you think?

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/DruidServerJunitResource.java
kfaraz: Unfortunately, JUnit5 does not have support for `ExternalResource`, which means I won't be able to use `TestDerbyConnector.DerbyConnectorRule` but I guess that can be fixed easily.

Let me see what needs to be done to make this JUnit 5 compatible. Thanks for the tip!

kfaraz:
@kgyrtkirk , thanks for the suggestions! Will try to address them.
Also, please let me know if you have any thoughts on https://github.com/apache/druid/pull/18147#issuecomment-2979598062.

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/DruidServerJunitResource.java
kfaraz: On second thought, it's probably best to make it independent of the JUnit version. We can just expose start/stop methods in our resource that can be called from setup/tearDown in the tests.

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/EmbeddedDruidServer.java
kgyrtkirk: interesting; I wasn't expecting that during a `toInstance` call that will be interpreted; somehow I was thinking `guice` have to be the one creating the original instance...makes sense ; but I still find it a little bit odd :)
I won't forget this trick :D

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/DruidServerJunitResource.java
kgyrtkirk: junit5 could give finer control than `ExternalResource`  ( look at the subclasses of `org.junit.jupiter.api.extension.Extension.class`) and it handles some things much better than 4.

fyi: there is already `TestDerbyConnector$DerbyConnectorRule5` - I've added it when I needed it earlier.

> independent of the JUnit version

sure; might make sense - but I think we should only write new tests in junit5; test writers could have much better control over things when its needed

kgyrtkirk:
regarding [comment](https://github.com/apache/druid/pull/18147#issuecomment-2979598062)

I would really like the separate module - but placing these tests where they should belong will be more in front of someone working on that part...so I think its better to put it inside an existing module - it might be possible to create a separate source set for it; but that would only be a logical isolation (probably with no real benefits)

we could have a module which loads multiple (or all) extensions so that more complicated scenarios could be tested....actually that was the reason I created `quidem-ut` - maybe we could rename that module so that the multi-extension simulation tests could live there along with the qtest stuff?

kfaraz:
> I would really like the separate module - but placing these tests where they should belong will be more in front of someone working on that part...so I think its better to put it inside an existing module - it might be possible to create a separate source set for it; but that would only be a logical isolation (probably with no real benefits)

Thanks for the feedback, @kgyrtkirk ! I feel the same.

> we could have a module which loads multiple (or all) extensions so that more complicated scenarios could be tested....actually that was the reason I created quidem-ut - maybe we could rename that module so that the multi-extension simulation tests could live there along with the qtest stuff?

Yeah, that would be nice! I originally thought we could put it into `integration-tests` module but the quidem UT module is more closely related. We could just call that module `simulation-tests` since quidem UTs are also essentially a type of simulation.

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/EmbeddedOverlord.java
kfaraz: Updated.

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/DruidServerJunitResource.java
kfaraz: Yes, this is what I ended up doing as well. The SDK is essentially independent of JUnit but there is an abstract `DruidClusterTest` based on JUnit 5 that users can extend to write new tests.

simulation-tests/src/test/java/org/apache/druid/testing/simulate/embedded/EmbeddedDruidServer.java
kfaraz: 😃

gianm:
> To nicely wire up the dependencies and the SDK, I am currently evaluating two alternatives described below.
> 
> I am leaning towards B. What are your thoughts?

IMO, tests for a module or an extension should live in that module or extension. This helps us organize the tests better, and also makes it possible for people with their own site-specific extensions to leverage the simulation test system. To me that means option B is preferred. Option A, where the simulation test module actually contains all the test cases, is not desirable because the tests don't get to live where they "should be".

If a test needs to use multiple extensions, at that point it could live in a a test-specific module. Something like:

- module `services` contains the simulation test system in its `test` tree, plus a few basic tests
- other modules contain tests specific to that module, and depend on the `test-jar` for `services` ; for example this may be useful for `druid-basic-security`
- module `simulation-tests` depends on a variety of extensions and contains tests that need to use multiple extensions. This doesn't need to exist until we actually have tests that need multiple extensions

It's kind of a blend of options A and B.

kfaraz:
That makes sense, @gianm . I have done just that except I have kept no tests in services/test right now. There are some in quidem-ut module though. Please let me know if I should move them to services/test instead.

services/src/test/java/org/apache/druid/testing/simulate/junit5/IndexingSimulationTestBase.java
```diff
@@ -0,0 +1,128 @@
...
+{
+  /**
+   * Random test datasource name that is freshly generated for each test method.
+   */
+  protected String dataSource;
+
+  @BeforeEach
+  protected void beforeEachTest()
+  {
+    dataSource = createTestDataourceName();
```
uds5501: Nit: typo - `createTestDatasourceName()`

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/KafkaSupervisorSimTest.java
```diff
@@ -0,0 +1,185 @@
...
+    SupervisorStatus supervisorStatus = getSupervisorStatus(supervisorId);
+    Assertions.assertFalse(supervisorStatus.isSuspended());
+    Assertions.assertTrue(supervisorStatus.isHealthy());
+    Assertions.assertEquals(dataSource, supervisorStatus.getDataSource());
+    Assertions.assertEquals("RUNNING", supervisorStatus.getState());
+    Assertions.assertEquals(topic, supervisorStatus.getSource());
+
+    // Get the task ID
+    List<TaskStatusPlus> taskStatuses = ImmutableList.copyOf(
+        getResult(cluster.leaderOverlord().taskStatuses(null, dataSource, 1))
```
uds5501: Question - For this TC, their are 2 partitions, so number of tasks should be 2 as well right? 

Or we are using a single task for demonstration purpose?

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/KafkaSupervisorSimTest.java
```diff
@@ -0,0 +1,185 @@
...
+    return new KafkaSupervisorTuningConfig(
+        null,
+        null, null, null,
+        1,
+        null, null, null, null, null, null, null, null, null, null,
+        null, null, null, null, null, null, null, null, null, null
+    );
+  }
+
+  private List<ProducerRecord<byte[], byte[]>> generateRecordsForTopic(
```
uds5501: Suggestion: There should be a provision to produce records for each partition

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/KafkaSupervisorSimTest.java
kfaraz: The requirement is `Number of tasks <= Number of partitions`.
A single task can read from any number of partitions.
We launch more tasks to increase ingestion throughput.

services/src/test/java/org/apache/druid/testing/simulate/EmbeddedDruidServer.java
```diff
@@ -0,0 +1,219 @@
...
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * An embedded Druid server used in simulation tests.
+ * This class and most of its methods are kept package protected as they are used
+ * only by the specific server implementations in the same package.
+ */
+public abstract class EmbeddedDruidServer implements ServerReferencesProvider
```
kgyrtkirk: since almost every class is called `EmbeddedX` ; I feel like it would  probably be an option to call these "embedded tests" ?

kfaraz: Yeah, that could work too 👍🏻 

I had originally set out to write a simulation framework (that only emulated the behaviour of a cluster and didn't need to be an actual Druid cluster) but it ended up being a single-process embedded cluster instead.

As for the name, the word "simulation" stuck and it also made for a small and ~easy~ catchy suffix for test names like `XyzSimTest`, so I kept it anyway. But I am not opposed to calling it something else either.

quidem-ut/src/test/java/org/apache/druid/simulate/indexing/IndexTaskSimTest.java
```diff
@@ -0,0 +1,208 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Simulation tests for batch {@link IndexTask} using inline datasources.
+ */
+public class IndexTaskSimTest extends IndexingSimulationTestBase
```
gianm: `quidem-ut` is now not really an appropriate name for this module, as these tests aren't using quidem. I don't think you need to change it in this PR necessarily, as there's already a lot of other changes and I don't want to get into a debate over the name in this PR 😄. A follow up to change the name would be useful.

services/src/test/java/org/apache/druid/testing/simulate/junit5/IndexingSimulationTestBase.java
```diff
@@ -0,0 +1,179 @@
...
+import org.apache.druid.testing.simulate.EmbeddedDruidCluster;
+import org.apache.druid.testing.simulate.EmbeddedOverlord;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Base class for simulation tests related to ingestion and indexing.
```
gianm: IMO would be better to have this be a collection of `static` utility methods. The reason is that if we ever add another kind of "test base" with different kinds of convenience functions, a single test class won't be able to use both sets of convenience functions, due to the impossibility of multiple inheritance. The presence of `runSql` in this class is a hint that something is off (that isn't an indexing API).

indexing-service/src/main/java/org/apache/druid/indexing/overlord/duty/UnusedSegmentsKiller.java
```diff
@@ -191,8 +191,14 @@ public DutySchedule getSchedule()
...
       // Check every hour that the kill queue is being processed normally
-      log.info("Scheduling is enabled to launch embedded kill tasks.");
-      return new DutySchedule(Duration.standardHours(1).getMillis(), Duration.standardMinutes(1).getMillis());
+      log.info(
+          "Unused segment killer is enabled. Kill queue will be rebuilt every [%s] if empty.",
+          killConfig.getDutyPeriod()
+      );
+      return new DutySchedule(
+          killConfig.getDutyPeriod().toStandardDuration().getMillis(),
+          Duration.standardSeconds(10).getMillis()
```
gianm: Is it intentional that this was changed this from 1 minute to 10 seconds?

server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataManager.java
```diff
@@ -679,10 +679,10 @@ private void pollSegments()
 
     stopwatch.stop();
     emitMetric("segment/poll/time", stopwatch.millisElapsed());
-    log.info(
+    /*log.info(
```
gianm: This seems unintentional; please restore the original code if so.

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorReport.java
```diff
@@ -30,7 +31,12 @@ public class SupervisorReport<T>
   private final DateTime generationTime;
   private final T payload;
 
-  public SupervisorReport(String id, DateTime generationTime, T payload)
+  @JsonCreator
```
gianm: I haven't yet found the place where this deserializer is used or tested- where is that? I wanted to check something about its usage: when there's a totally-generic thing like `T` as a property, Jackson doesn't know what `T` to use unless a `TypeReference` is provided. So it's important that the deserialization be done in that specific way.

services/src/test/java/org/apache/druid/testing/simulate/EmbeddedMiddleManager.java
```diff
@@ -0,0 +1,87 @@
...
+   * Creates an {@link EmbeddedMiddleManager}.
+   *
+   * @see EmbeddedMiddleManager
+   * @deprecated Use {@link EmbeddedIndexer} instead.
+   */
+  @Deprecated
+  public EmbeddedMiddleManager()
+  {
+    // Don't sync lookups as cluster might not have a Coordinator
+    // addProperty("druid.lookup.enableLookupSyncOnStartup", "false");
```
gianm: Please don't include commented-out code.

services/src/test/java/org/apache/druid/testing/simulate/EmbeddedDruidServer.java
gianm: my 2¢: to me the word "simulation" is misleading given the final design- I don't believe anything is really being simulated. IMO using "embedded" across the board is better. I suggest searching for "simulate" or "simulation" and adjusting them so we can keep the wording consistent. As for `XyzSimTest`, could be `XyzEmbedTest`.

services/src/test/java/org/apache/druid/testing/simulate/EmbeddedZookeeper.java
```diff
@@ -0,0 +1,55 @@
...
+ * under the License.
+ */
+
+package org.apache.druid.testing.simulate;
+
+import org.apache.druid.curator.CuratorTestBase;
+
+/**
+ * {@link EmbeddedResource} for an embedded zookeeper cluster that can be used
+ * as a JUnit Rule in simulation tests.
```
gianm: Is this comment about "JUnit Rule" correct?

gianm (commented):
Great work. I've always wished for a testing system like this.

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorReport.java
kfaraz: Ah, thanks for catching this. I had intended to use this but it is not being used anymore. Will revert this.

quidem-ut/src/test/java/org/apache/druid/simulate/indexing/IndexTaskSimTest.java
kfaraz: Sounds good. 👍🏻 

I could just put these tests in a separate module, if that seems better?

services/src/test/java/org/apache/druid/testing/simulate/junit5/IndexingSimulationTestBase.java
kfaraz: I was wondering if these methods shouldn't just go into the `EmbeddedDruidCluster` class itself, given that they are always used to interact with a specific cluster.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/duty/UnusedSegmentsKiller.java
kfaraz: Yes, it helps with the tests finishing faster and doesn't really affect a prod setup.
I could also just pass this value as a config too for consistency and let the default be 1 minute.

services/src/test/java/org/apache/druid/testing/simulate/EmbeddedZookeeper.java
kfaraz: No, will clean it up.

services/src/test/java/org/apache/druid/testing/simulate/EmbeddedDruidServer.java
kfaraz: Sounds good, will make the change.

kfaraz:
Thanks for the feedback, @gianm  and for catching the leftover superfluous changes!
I will address your comments and take another pass to ensure that no extra changes have made their way into this PR.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/duty/UnusedSegmentsKiller.java
gianm: If it doesn't really affect a prod setup then it's imo ok to leave it like this.

quidem-ut/src/test/java/org/apache/druid/simulate/indexing/IndexTaskSimTest.java
gianm: Might as well be a separate module named `embedded-tests`. The main reason I say that is I am struggling to think of a good name for the combined quidem + embedded test module.

services/src/test/java/org/apache/druid/testing/simulate/junit5/IndexingSimulationTestBase.java
gianm: That would also work. It would make the calling code look nicer.

The main downside I can think of is that at some point, there may be a lot of functions and the `EmbeddedDruidCluster` class could get pretty long. But if that happens, there are ways to address it. For example we could split the functions out into default methods in interfaces that extend `DruidClusterApi` and have `EmbeddedDruidCluster` implement `DruidClusterApi` and multiple of those other interfaces. Please don't do that now though. I'm just thinking through how it might evolve in the future. For now I think adding the methods directly to `EmbeddedDruidCluster` is good.

quidem-ut/src/test/java/org/apache/druid/simulate/indexing/IndexTaskSimTest.java
kfaraz: That works!

I guess quidem tests are also essentially embedded tests. It's just that they don't use a full cluster but only a Broker-like setup. 

I am not sure how much value it would add, but we could try piping quidem queries through an embedded cluster itself in the future. If we ever do that, we could have just the one module, `embedded-tests`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/duty/UnusedSegmentsKiller.java
kfaraz: Ended up changing this slightly since I realized that it is possible that an Overlord has just come up and has been elected leader but the segment metadata cache has still not warmed up (if operating in mode `ifSynced`).

The unused segment killer performs better with the cache fully synced.

So using the following values which feel good enough for all cases. We can revisit them later if needed.

duty period = 1 hour,
initial delay = duty period / 4 = 15 mins

services/src/test/java/org/apache/druid/testing/simulate/junit5/IndexingSimulationTestBase.java
kfaraz: Thanks for the suggestion! Ended up doing something similar.
- Renamed the class `IndexingSimulationTestBase` to `EmbeddedClusterApis`
- Kept some methods like `createTaskFromPayload()` and `createTestDatasourceName()` as static
- Created an instance of `EmbeddedClusterApis` for each `EmbeddedDruidCluster`
- Tests can now invoke methods as follows:

```java
// general usage pattern
cluster.callApi().onLeaderOverlord(o -> o.runTask(taskId, task));

cluster.callApi().waitForTaskToSucceed(taskId);

// shorthand for SQL
cluster.runSql(...);
```

gianm (approved):
LGTM

kfaraz:
Thanks a lot for the review, @gianm !

PR #18148, opened 2025-06-17
Implement per-segment query timeout on data nodes

jtuglu1:
Druid can struggle with query fairness in the native engine at the data node level (peon, historical).

### Background
There are N primary “resources” which queries contend over during query processing:
1. Processing threads (every query type except window and timeseries)
 - Each query will issue per-segment runners which are then run as individual futures on the processing thread pool.
 - Larger # of segments -> larger # of runners -> larger # of threads	
2. Processing buffers (every query type)
- Each processing thread is assigned a buffer
3. Additional merge Buffers (only GroupBy)
- Merge buffersGroupBy queries require 1 merge buffer per query (2 total buffer), while a nested GroupBy query requires  merge buffers (3 total buffer).

### What's already solved
1. Initial priority-based scheduling of segments
- Each processing thread is held for however long it takes to process one segment, and is released back to the pool after that. When it goes back to the pool, another query can pick it up next, and at that time a higher priority query can take over
- This generally means that as long as two queries have different priorities, the highest priority one will always "take over" relatively quickly at the Historical/Peon level

### Problem: Priority Inversion Issue
1. Automatic query priority assignment isn’t great, so we cannot always rely on the priority assigned being representative of the query’s impact on the cluster.
2. The above breaks down in the following case:

```
Query A – Priority 0 (highest), 10 segment, all take 1s to process
Query B – Priority 0 (lowest), 1000 segment, each take 10s to process
Query B beats A to all the processing threads, now A needs to wait for 10 of B’s segments (wait ~10s) to finish processing.
```

### Description

This table breaks down the current state of processing for each query type:

| Query Type | Location |
|------------|----------|
| dataSourceMetadata | Processing Pool |
| groupBy | Processing Pool |
| scan | Netty |
| search | Processing Pool |
| segmentMetadata | Processing Pool |
| select | Processing Pool |
| timeBoundary | Processing Pool |
| timeseries | Processing Pool |
| topN | Processing Pool |
| windowOperator | Netty |
| union | Processing Pool |

#### Query Processing Pool changes
Since Guava `ListenableFuture` are older than `CompletableFuture`, they do not come with their own self-managed timeout capabilities (meaning there's no `.withTimeout`, etc.). This means that in order to service the timeouts to all futures, I've needed to create a separate threadpool (in order to isolate work from scan queue). This will mean that small timeslices of CPU will be spent on servicing/cancelling timeouts, but this should negligible if configured correctly. The default size of the timeout threadpool is 1, but users can disable it by setting to `druid.processing.numTimeoutThreads=0`.

Adds a `submitRunnerTask(PrioritizedQueryRunnerCallable<T, V> task, long timeout, TimeUnit unit)` method which returns a future with a timeout. This hides the timeout service abstraction from the clients and can be re-used elsewhere if needed.

#### Query Context changes
- Introduces a timeout context parameter `perSegmentTimeout` into the query context which can be used to specify the maximum time each segment scan should take. This parameter operates similar to the `timeout` key, where the default is no timeout, and any value > 0 is treated as a valid timeout.

#### General comments
I'd like to merge this logic at some point into a single runner that interacts with the processing pool (see [here](https://github.com/apache/druid/pull/18593) for a draft of that). But, given that would just be refactoring, I'd prefer to keep that separate from this change.

#### Release note

Implement settable per-segment processing query timeout on data nodes.

This feature can be enabled via setting `perSegmentTimeout` parameter in the query context and by setting `druid.processing.numTimeoutThreads` to a small number (e.g. 1-3) on all data nodes where you want this feature enabled. 

<hr>

This PR has:

- [x] been self-reviewed.


abhishekagarwal87:
I think, this solves the problem only a narrow class of problems and adds another parameter that may not see a broader adoption. I do however agree that query scheduling at data level is an area worth exploring and tinkering with. Though I wonder if there are better ways to solve this problem. One solution that comes to my mind is if we can use virtual threads of sort. Right now, we have this processing thread pool that essentially dictates the compute capacity that segment processing threads. But if these threads are doing lot more IO than CPU, that capacity is being wasted. Recently java has gotten the ability of Virtual threads and that could be used to run segment processing instead of directly using OS-level threads. 

A higher-level comment is that we shouldn't just make this change without some confidence that our solution makes lives better for a good number of use cases. You should first build a test setup that can be used to simulate query congestion at data level along with metrics that reflect the degree of the congestion, throughput, fairness. Once such a system is in place, thats when you can craft few strategies and using your test setup to measure what strategy is the best.

jtuglu1:
> I think, this solves the problem only a narrow class of problems and adds another parameter that may not see a broader adoption. I do however agree that query scheduling at data level is an area worth exploring and tinkering with. Though I wonder if there are better ways to solve this problem. One solution that comes to my mind is if we can use virtual threads of sort. Right now, we have this processing thread pool that essentially dictates the compute capacity that segment processing threads. But if these threads are doing lot more IO than CPU, that capacity is being wasted. Recently java has gotten the ability of Virtual threads and that could be used to run segment processing instead of directly using OS-level threads.
> 
> A higher-level comment is that we shouldn't just make this change without some confidence that our solution makes lives better for a good number of use cases. You should first build a test setup that can be used to simulate query congestion at data level along with metrics that reflect the degree of the congestion, throughput, fairness. Once such a system is in place, thats when you can craft few strategies and using your test setup to measure what strategy is the best.

Regarding virtual threads – yes I thought about this as well, however that was introduced in JDK 21, which is currently not supported fully (AFAIK). When this becomes available, I 100% agree that (assuming historical processing threads aren't mostly CPU-bound), switching to userspace thread and increasing the pool size will be improvement.

jtuglu1:
> A higher-level comment is that we shouldn't just make this change without some confidence that our solution makes lives better for a good number of use cases. You should first build a test setup that can be used to simulate query congestion at data level along with metrics that reflect the degree of the congestion, throughput, fairness. Once such a system is in place, thats when you can craft few strategies and using your test setup to measure what strategy is the best.

Yes, of course – we've observed this issue in our largest cluster numerous times (daily) where larger, low-priority queries beat smaller, shorter queries to the processing threads. This causes backup when processing each segment takes a while (on the order of O(seconds/mins) in some extreme cases). I see the point about having too many "unused" parameters (config bloat, etc.). By default this would be turned off, but it gives us an extra lever to turn to reduce the impact of these heavy-hitter queries.

maytasm:
@abhishekagarwal87 
This idea came from my discussion with @gianm and @gianm suggested this change (see: https://apachedruidworkspace.slack.com/archives/C030CMF6B70/p1746578390954639?thread_ts=1745436989.786489&cid=C030CMF6B70)
We have also observed some queries that takes on avg 1-2 minutes to process a single segment (see: https://apachedruidworkspace.slack.com/archives/C030CMF6B70/p1746141850807709?thread_ts=1745436989.786489&cid=C030CMF6B70)

image
image

abhishekagarwal87:
Hmm. Looking at the query, my guess is that the segment processing is spending bulk of its time in CPU and as such might not be interruptible at all.

abhishekagarwal87:
btw I still have the same concerns and we definitely need tests and metrics to test various techniques to improve progress when data nodes have query congestion. 

Re virtual threads and JDK 21 support, we will need to be compatible with it given that its been around for a while and Java 17 itself reached its EOL in Sep 24.

maytasm (approved):

processing/src/main/java/org/apache/druid/query/ForwardingQueryProcessingPool.java
```diff
@@ -33,10 +38,19 @@
 public class ForwardingQueryProcessingPool extends ForwardingListeningExecutorService implements QueryProcessingPool
 {
   private final ListeningExecutorService delegate;
+  private final ScheduledExecutorService timeoutService;
 
+  public ForwardingQueryProcessingPool(ExecutorService executorService, ScheduledExecutorService timeoutService)
```
clintropolis: maybe consider doing this in a bit more of an 'opt-in' manner where we allow `timeoutService` to be nullable. Like the processing module would check for 0 threads in `numTimeoutThreads`, and if 0 pass null into this, and here we could `log.warn` or something if `submitRunnerTask ` while `timeoutService` is null as a way to disable this functionality.

processing/src/main/java/org/apache/druid/query/QueryContexts.java
```diff
@@ -44,6 +44,7 @@ public class QueryContexts
   public static final String PRIORITY_KEY = "priority";
   public static final String LANE_KEY = "lane";
   public static final String TIMEOUT_KEY = "timeout";
+  public static final String PER_SEGMENT_TIMEOUT_KEY = "segmentTimeout";
```
clintropolis: i wonder if the naming of this property should include something about the processing pool since it only applies to queries that run on the pool, not sure what a great name would be though... naming is the worst, and this is probably fine too

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
```diff
@@ -165,8 +180,16 @@ public Iterable<T> call()
             }
             catch (ExecutionException e) {
               GuavaUtils.cancelAll(true, future, futures);
-              Throwables.propagateIfPossible(e.getCause());
-              throw new RuntimeException(e.getCause());
+              Throwable cause = e.getCause();
+              Throwables.propagateIfPossible(cause);
+              if (cause instanceof TimeoutException || cause instanceof QueryTimeoutException) {
```
gianm: I think this can't ever be `QueryTimeoutException`, because that's a `RuntimeException`, and `Throwables.propagateIfPossible(cause)` would have thrown it.

clintropolis (commented):
overall makes sense to me :+1:

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/GroupByMergingQueryRunner.java
```diff
@@ -392,6 +403,14 @@ private void waitForFutureCompletion(
     }
     catch (ExecutionException e) {
       GuavaUtils.cancelAll(true, future, futures);
+      Throwable cause = e.getCause();
+      if (cause instanceof TimeoutException || cause instanceof QueryTimeoutException) {
```
gianm: It's not ideal that the query runners all need special handling for the per-segment timeouts. It seems brittle. What happens if a query runner doesn't implement it properly, or at all?

I wonder if a simpler approach could be to use the same timeout we already have, but shorten it in `SpecificSegmentQueryRunner`. i.e., update `SpecificSegmentQueryRunner` to set the `timeout` in the query context to `min(timeout, perSegmentTimeout)` when it runs the query on the individual segment. I think this should work just as well, but would require fewer/no changes to the query engines.

gianm (commented):
I like the idea, but have some comments on the implementation.

On the overall idea, btw, I think it will be useful in situations where the cluster workload is heterogeneous and we need to make sure queries play nice with each other. It helps with what is currently a core problem with playing nice: individual segment scans are not preemptible.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/GroupByMergingQueryRunner.java
jtuglu1: > I wonder if a simpler approach could be to use the same timeout we already have, but shorten it in `SpecificSegmentQueryRunner`.

I'm not sure this will work. The reason being is that while this does execute per segment, the timeout is not enforced here. The timeout is unfortunately enforced in the caller runners which aggregate lists of per-segment runners into callable futures (e.g. `ChainedExecutionQueryRunner` or `GroupByMergingQueryRunner`). These callers are in higher stack frames whereas the SpecificSegmentQueryRunner is near the bottom of the stack. The thread that is running the SpecificSegmentQueryRunner is the same thread that we'd be asking to service the timer interrupt to cancel the work(which won't work in Java AFAIK since the Futures framework is not truly async). These callers rely on the aggregate query timeout to make one mega future with that value.

jtuglu1: Any thoughts here @gianm?

jtuglu1: If this duplicated logic is a problem, something which I could do is make a new, common runner layer that each of these QueryRunners call into for creating/aggregating/joining on a list of segment query futures. Something like `PerSegmentQueryManagerRunner`. This would likely mean more refactoring, however.

docs/configuration/index.md
```diff
@@ -1371,6 +1371,7 @@ Processing properties set on the Middle Manager are passed through to Peons.
 |`druid.processing.formatString`|Realtime and Historical processes use this format string to name their processing threads.|processing-%s|
 |`druid.processing.numMergeBuffers`|The number of direct memory buffers available for merging query results. The buffers are sized by `druid.processing.buffer.sizeBytes`. This property is effectively a concurrency limit for queries that require merging buffers. If you are using any queries that require merge buffers (currently, just groupBy) then you should have at least two of these.|`max(2, druid.processing.numThreads / 4)`|
 |`druid.processing.numThreads`|The number of processing threads to have available for parallel processing of segments. Our rule of thumb is `num_cores - 1`, which means that even under heavy load there will still be one core available to do background tasks like talking with ZooKeeper and pulling down segments. If only one core is available, this property defaults to the value `1`.|Number of cores - 1 (or 1)|
+|`druid.processing.numTimeoutThreads`|The number of processing threads to have available for handling per-segment query timeouts.|0|
```
maytasm: nit: can you make it clear that 0 will disable per segment query timeout?

docs/configuration/index.md
```diff
@@ -1514,6 +1515,7 @@ Druid uses Jetty to serve HTTP requests.
 |`druid.processing.formatString`|Indexer processes use this format string to name their processing threads.|processing-%s|
 |`druid.processing.numMergeBuffers`|The number of direct memory buffers available for merging query results. The buffers are sized by `druid.processing.buffer.sizeBytes`. This property is effectively a concurrency limit for queries that require merging buffers. If you are using any queries that require merge buffers (currently, just groupBy) then you should have at least two of these.|`max(2, druid.processing.numThreads / 4)`|
 |`druid.processing.numThreads`|The number of processing threads to have available for parallel processing of segments. Our rule of thumb is `num_cores - 1`, which means that even under heavy load there will still be one core available to do background tasks like talking with ZooKeeper and pulling down segments. If only one core is available, this property defaults to the value `1`.|Number of cores - 1 (or 1)|
+|`druid.processing.numTimeoutThreads`|The number of processing threads to have available for handling per-segment query timeouts.|0|
```
maytasm: nit: can you make it clear that 0 will disable per segment query timeout?

docs/configuration/index.md
```diff
@@ -1622,6 +1624,7 @@ Druid uses Jetty to serve HTTP requests.
 |`druid.processing.formatString`|Realtime and Historical processes use this format string to name their processing threads.|processing-%s|
 |`druid.processing.numMergeBuffers`|The number of direct memory buffers available for merging query results. The buffers are sized by `druid.processing.buffer.sizeBytes`. This property is effectively a concurrency limit for queries that require merging buffers. If you are using any queries that require merge buffers (currently, just groupBy) then you should have at least two of these.|`max(2, druid.processing.numThreads / 4)`|
 |`druid.processing.numThreads`|The number of processing threads to have available for parallel processing of segments. Our rule of thumb is `num_cores - 1`, which means that even under heavy load there will still be one core available to do background tasks like talking with ZooKeeper and pulling down segments. If only one core is available, this property defaults to the value `1`.|Number of cores - 1 (or 1)|
+|`druid.processing.numTimeoutThreads`|The number of processing threads to have available for handling per-segment query timeouts.|0|
```
maytasm: nit: can you make it clear that 0 will disable per segment query timeout?

maytasm (approved):
LGTM. Thanks for cleaning up. This design is much cleaner than the previous one

docs/configuration/index.md
abhishekrb19: Also, could you add guidance on how an operator should configure this property? What happens if an operator set this value too high or as low as 1?

server/src/main/java/org/apache/druid/guice/DruidProcessingModule.java
```diff
@@ -157,6 +158,7 @@ public static QueryProcessingPool createProcessingExecutorPool(
             lifecycle,
             config
         ),
+        config.getNumTimeoutThreads() > 0 ? ScheduledExecutors.fixed(config.getNumTimeoutThreads(), "PrioritizedExecutorService-Timeout-%%d") : null,
```
abhishekrb19: Typo?
```suggestion
        config.getNumTimeoutThreads() > 0 ? ScheduledExecutors.fixed(config.getNumTimeoutThreads(), "PrioritizedExecutorService-Timeout-%d") : null,
```

server/src/main/java/org/apache/druid/guice/BrokerProcessingModule.java
```diff
@@ -84,7 +85,7 @@ public QueryProcessingPool getProcessingExecutorPool(
       DruidProcessingConfig config
   )
   {
-    return new ForwardingQueryProcessingPool(Execs.dummy());
+    return new ForwardingQueryProcessingPool(Execs.dummy(), (ScheduledExecutorService) Execs.dummy());
```
abhishekrb19: Any reason to not pass this as null for the Broker module rather than a dummy one?

processing/src/main/java/org/apache/druid/java/util/common/concurrent/DummyExecutorService.java
```diff
@@ -117,4 +120,38 @@ public void execute(Runnable command)
...
+  public @NotNull ScheduledFuture<?> scheduleAtFixedRate(
+      @NotNull Runnable command,
+      long initialDelay,
+      long period,
+      @NotNull TimeUnit unit
+  )
+  {
+    throw new UnsupportedOperationException();
+  }
+
```
abhishekrb19: Why implement `ScheduledExecutorService` to override these methods and throw `UnsupportedOperationException()`?  Looks like this is only called from the `BrokerProcessingModule` - would passing null there suffice instead of this dummy one?

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/GroupByMergingQueryRunner.java
```diff
@@ -176,6 +176,8 @@ public Sequence<ResultRow> run(final QueryPlus<ResultRow> queryPlus, final Respo
     // query processing together.
     final long queryTimeout = queryContext.getTimeout();
     final boolean hasTimeout = queryContext.hasTimeout();
+    final boolean hasPerSegmentTimeout = queryContext.usePerSegmentTimeout();
```
abhishekrb19, line 179: Is there a cluster-level config that an operator can set and forget for all queries (similar to `druid.server.http.defaultQueryTimeout`)?

Also, is it intentional that we've not [documented](https://druid.apache.org/docs/latest/querying/query-context-reference/#general-parameters) this query context?

processing/src/main/java/org/apache/druid/query/QueryContext.java
```diff
@@ -545,6 +544,32 @@ public void verifyMaxQueryTimeout(long maxQueryTimeout)
     }
   }
 
+  public long getPerSegmentTimeout()
+  {
+    return getPerSegmentTimeout(QueryContexts.NO_TIMEOUT);
+  }
+
+  public long getPerSegmentTimeout(long defaultTimeout)
```
abhishekrb19: can be private-scoped

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/GroupByMergingQueryRunner.java
```diff
@@ -176,6 +176,8 @@ public Sequence<ResultRow> run(final QueryPlus<ResultRow> queryPlus, final Respo
...
+    final long perSegmentTimeout = queryContext.getPerSegmentTimeout();
```
abhishekrb19, line 180: I think for a cluster-level config, we’d want to wire up some changes to `SetAndVerifyContextQueryRunner` so it applies to all query runners; it'd apply the query-specific context and override the cluster default

processing/src/main/java/org/apache/druid/query/QueryContext.java
```diff
@@ -545,6 +544,32 @@ public void verifyMaxQueryTimeout(long maxQueryTimeout)
...
+  {
+    final long timeout = getLong(QueryContexts.PER_SEGMENT_TIMEOUT_KEY, defaultTimeout);
+    if (timeout >= 0) {
+      return timeout;
+    }
+
+    throw new BadQueryContextException(
+        StringUtils.format(
+            "Per-segment timeout [%s] must be a non negative value, but was %d",
```
abhishekrb19:
```suggestion
            "Per-segment timeout [%s] must be a non negative value, but was [%d]",
```

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
```diff
@@ -97,42 +101,54 @@ public Iterator<T> make()
...
+                                throw e;
+                              }
+                              catch (Exception e) {
+                                if (query.context().isDebug()) {
+                                  log.error(e, "Exception with one of the sequences!");
+                                } else {
+                                  log.noStackTrace().error(e, "Exception with one of the sequences!");
                                 }
-                              });
+                                Throwables.throwIfUnchecked(e);
```
abhishekrb19, line 137: `Throwables.propagateIfPossible(e);` -> `Throwables.throwIfUnchecked(e);` bc of deprecated usage?

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
```diff
@@ -141,32 +157,33 @@ public Iterable<T> call()
...
               GuavaUtils.cancelAll(true, future, futures);
               throw new QueryInterruptedException(e);
             }
-            catch (CancellationException e) {
-              throw new QueryInterruptedException(e);
-            }
-            catch (TimeoutException e) {
-              log.warn("Query timeout, cancelling pending results for query id [%s]", query.getId());
+            catch (TimeoutException | QueryTimeoutException e) {
+              log.noStackTrace().warn(e, "Query timeout, cancelling pending results for query [%s]", query.getId());
```
abhishekrb19, line 173: Would this timeout occur if a query exceeds `timeout`?

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
```diff
@@ -141,32 +157,33 @@ public Iterable<T> call()
...
               throw new QueryTimeoutException(StringUtils.nonStrictFormat("Query [%s] timed out", query.getId()));
             }
             catch (ExecutionException e) {
+              log.noStackTrace().warn(e, "Query error, cancelling pending results for query [%s]", query.getId());
               GuavaUtils.cancelAll(true, future, futures);
-              Throwables.propagateIfPossible(e.getCause());
-              throw new RuntimeException(e.getCause());
+              Throwable cause = e.getCause();
+              // Nested per-segment future timeout
+              if (cause instanceof TimeoutException) {
```
abhishekrb19, line 182: Would this timeout occur if it exceeds `perSegmentTimeout` specifically? If so, could you update the error message to reflect that so a user/admin can take corrective action? Same for the above timeout excewption message.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/GroupByMergingQueryRunner.java
```diff
@@ -376,22 +387,24 @@ private void waitForFutureCompletion(
...
-      throw new QueryTimeoutException();
+      throw new QueryTimeoutException(StringUtils.nonStrictFormat("Query [%s] timed out", query.getId()));
     }
     catch (ExecutionException e) {
+      log.noStackTrace().warn(e, "Query error, cancelling pending results for query [%s]", query.getId());
       GuavaUtils.cancelAll(true, future, futures);
+      Throwable cause = e.getCause();
+      // Nested per-segment future timeout
+      if (cause instanceof TimeoutException) {
+        throw new QueryTimeoutException(StringUtils.nonStrictFormat("Query [%s] timed out", query.getId()));
```
abhishekrb19: Ditto on the error message here and above

processing/src/main/java/org/apache/druid/query/ForwardingQueryProcessingPool.java
```diff
@@ -19,12 +19,17 @@
...
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
 
+import javax.annotation.Nullable;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Default implementation of {@link QueryProcessingPool} that just forwards operations, including query execution tasks,
```
abhishekrb19, line 35: Could you update the javadoc here to reflect the usage of `timeoutService` in conjunction to this pool?

processing/src/test/java/org/apache/druid/query/groupby/GroupByQueryRunnerFailureTest.java
```diff
@@ -318,4 +330,258 @@ public boolean isSingleThreaded()
...
+        .queryId("fast")
+        .build();
+
+    CountDownLatch slowStart = new CountDownLatch(2);
+    CountDownLatch fastStart = new CountDownLatch(1);
+
+    QueryRunner<ResultRow> signalingSlowRunner = (queryPlus, responseContext) -> {
+      slowStart.countDown();
+      try {
+        Thread.sleep(60_000L);
```
abhishekrb19, line 544: Does this need to sleep for 60s? In the interest of having faster tests, I think you can just do something like 6s since the queries below have lower timeout values

processing/src/test/java/org/apache/druid/query/ChainedExecutionQueryRunnerTest.java
```diff
@@ -344,6 +358,151 @@ public void testSubmittedTaskType()
...
+    }
+    catch (Exception e) {
+      thrown = e;
+    }
+
+    Assert.assertNull("No results expected due to timeout", results);
+    Assert.assertNotNull("Exception should be thrown", thrown);
+    Assert.assertTrue(
+        "Should be QueryTimeoutException or caused by it",
+        Throwables.getRootCause(thrown) instanceof QueryTimeoutException
```
abhishekrb19, line 415: Since `QueryTimeoutException` can be thrown from multiple places depending on what timeout kicks in, could you also assert on the error message so we know which code path triggered it?

processing/src/test/java/org/apache/druid/query/groupby/GroupByQueryRunnerFailureTest.java
```diff
@@ -318,4 +330,258 @@ public boolean isSingleThreaded()
...
+      return Sequences.empty();
+    };
+
+    QueryRunner<ResultRow> mergeRunners = factory.mergeRunners(
+        processingPool,
+        List.of(runner, mockRunner)
+    );
+
+    Assert.assertThrows(
+        QueryTimeoutException.class,
```
abhishekrb19, line 488: Same comment on exception validation

abhishekrb19 (commented):
Thanks, @jtuglu1! Overall looks good to me.

processing/src/test/java/org/apache/druid/query/groupby/GroupByQueryRunnerFailureTest.java
jtuglu1: It doesn't, no. The long sleep is put there on-purpose to actually fail the test. The entire test itself has a shorter timeout (5s) and will complete shorter than that (~1-2s).

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/GroupByMergingQueryRunner.java
jtuglu1: Users should be able to set `druid.query.default.context.perSegmentTimeout` as a default setting as they can for all other context types.

No, Let me add the context docs.

processing/src/main/java/org/apache/druid/java/util/common/concurrent/DummyExecutorService.java
jtuglu1: I can remove

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
jtuglu1: Yes

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
jtuglu1: Yes. This depends on whether the query is being executed in direct mode as well (but that's not really relevant here).

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/GroupByMergingQueryRunner.java
jtuglu1: The cluster level config can be applied as a default cluster query context flag [here](https://druid.apache.org/docs/latest/querying/query-context#runtime-properties) (e.g. `druid.query.default.context.perSegmentTimeout`). I don't think we need to (nor should) set anything in `SetAndVerifyContextQueryRunner`, given this default context functionality exists.

processing/src/main/java/org/apache/druid/query/QueryContext.java
jtuglu1: I left public considering it's a context getter/setter and the equivalent `getTimeout(long defaultTimeout)` is also public.

server/src/main/java/org/apache/druid/guice/DruidProcessingModule.java
jtuglu1: I had left it in for formatting reasons if I recall

docs/configuration/index.md
```diff
@@ -1371,6 +1371,7 @@ Processing properties set on the Middle Manager are passed through to Peons.
...
 |`druid.processing.numThreads`|The number of processing threads to have available for parallel processing of segments. Our rule of thumb is `num_cores - 1`, which means that even under heavy load there will still be one core available to do background tasks like talking with ZooKeeper and pulling down segments. If only one core is available, this property defaults to the value `1`.|Number of cores - 1 (or 1)|
+|`druid.processing.numTimeoutThreads`|The number of processing threads to have available for handling per-segment query timeouts. Setting this value to `0` removes the ability to service per-segment timeouts, irrespective of query-level configs. As these threads are just servicing timers, it's recommended to set this value to some small percent (e.g. 5%) of the total query processing cores available to the peon.|0|
```
abhishekrb19: "irrespective of query-level configs." -> did you intend to say query context like `perSegmentTimeout`? I think it'll be better to be specific here.

docs/querying/query-context-reference.md
```diff
@@ -44,6 +44,7 @@ Unless otherwise noted, the following parameters apply to all query types, and t
 |Parameter          |Default                                 | Description          |
 |-------------------|----------------------------------------|----------------------|
 |`timeout`          | `druid.server.http.defaultQueryTimeout`| Query timeout in millis, beyond which unfinished queries will be cancelled. 0 timeout means `no timeout` (up to the server-side maximum query timeout, `druid.server.http.maxQueryTimeout`). To set the default timeout and maximum timeout, see [Broker configuration](../configuration/index.md#broker) |
+|`perSegmentTimeout`| `null`                                 | Per-segment processing timeout in millis, beyond which unfinished queries will be cancelled. Should be ≤ `timeout`. 0 `perSegmentTimeout` means `no per-segment timeout`. Generally, a standard default should be O(X seconds). A cluster-wide default value for this query context can be specified via `druid.query.default.context.perSegmentTimeout`.|
```
abhishekrb19, line 47:
```suggestion
|`perSegmentTimeout`| `null`                                 | Per-segment processing timeout in millis, beyond which unfinished queries occupying the processing pool will be cancelled. Should be ≤ `timeout`. Setting `perSegmentTimeout` to 0 means there's no per-segment timeout (up to the server-side maximum per segment timeout `druid.query.default.context.perSegmentTimeout`). Generally, a standard default should be in the order of a few seconds (e.g., 5 seconds). A cluster-wide default value for this query context can be specified via `druid.query.default.context.perSegmentTimeout`.|
```

abhishekrb19: I think using -1 as the default to disable per segment timeouts would be clearer. However, it's consistent with `timeout`'s default strategy, so this is okay

abhishekrb19, line 47: > Should be ≤ `timeout`

I don't think there's any validation on the query path that checks for this invariant right? So this is more of a recommendation?

abhishekrb19 (approved):
LGTM!

docs/querying/query-context-reference.md
jtuglu1: Yes, having a validation here would be caught anyways by the overall `Futures.allAsList().get()` timeout.

maytasm:
@jtuglu1 The per-segment timeout starts counting at submit time (ForwardingQueryProcessingPool.submitRunnerTask wraps the future at submit time). So time spent queued in the historical processing pool counts against the per-segment time. This can timeout fast queries if it's blocked behind slow queries (the processing pools were saturated by expensive queries). Is there any reason we don't start counting when segment is actually processing (instead of submitted)

jtuglu1:
@maytasm no, this looks like a bug. I can push a fix later this week or happy to review one as well.

PR #18154, opened 2025-06-18
Update some dependencies

kgyrtkirk:
Noticed that some deps are overriden with lower versions (for ex: `lang3`).

And wanted to enable the enforcer plugins `requireUpperBoundDeps` to  make sure that doesn't happen...however that uncovered a lot of downgrades...here are few which seemed more-or-less okay.

most notable upgrades:
* `guava` is upgraded to `32.1.3-jre`
* `slf4j1->slf4j2` which implied to upgrade to use `log4j-slf4j2` 
* `kotlin-stdlib` upgraded from `1.6.10`  to `1.9.25`


processing/src/test/java/org/apache/druid/query/aggregation/SerializablePairLongStringComplexMetricSerdeTest.java
```diff
@@ -69,7 +69,7 @@ public void testLargeString() throws Exception
     assertExpected(ImmutableList.of(new SerializablePairLongString(
         100L,
         randomStringUtils.randomAlphanumeric(2 * 1024 * 1024)
-    )), 2097182, 2103140);
+    )), 2097182, 2103139);
```
gianm, line 72: What is happening here? It seems the `expectedCompressedSize` is changing- what drives this, and is it backwards compatible with older readers and writers?

kgyrtkirk: I was also worried when I've first saw these tests breaking :)
but its due to the fact that the random algorithm was altered in `commons-lang3:3.15.0`.
The reason that matters is that `RandomStringUtils` uses `org.apache.commons.lang3.RandomStringUtils` internally.

Prior to `<3.15.0` these tests pass with all the other dep updates and require these updates afterwards.

https://commons.apache.org/proper/commons-lang/changes-report.html#a3.15.0
```
Reimplement RandomUtils and RandomStringUtils on top of SecureRandom#getInstanceStrong() #1235. Thanks to Gary Gregory, Henri Yandell, Fabrice Benhamouda.
```

gianm: ah, thank you for looking into this.

gianm (approved):

kgyrtkirk:
thank you @gianm for taking a look and merging it!

PR #18155, opened 2025-06-18
Include _* and $* regex patterns for unit test patterns

abhishekrb19:
Currently, I don’t think there are any test classes in this repo that follow these patterns, but let’s add these regexes for completeness, since Java class names are allowed to start with these characters.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


kgyrtkirk:
```
find . |fgrep -f <(echo _;echo '$')|grep java$
```
I don't see any `java` files containing `_` or `$` ; could we add a sanity check for checkstyle with 
https://checkstyle.sourceforge.io/checks/regexp/regexponfilename.html
to ensure that no such files are created?

kfaraz:
Yes, it would make sense to prevent authors from writing such test names, even if they are allowed by Java.

abhishekrb19:
Done. Updated to prohibit class names that start with  `_` and `$`. We have existing class names that start with lower-case like `package-info.java`, so didn't add that to the rule.

gianm (approved):

PR #18161, opened 2025-06-19
Fix segment unavailability race in coordinator duty

jtuglu-netflix:
There is a race that cause segment unavailability between coordinatory dutygroup runs when a segment is re-balancing.

The race is as follows:

```
Given segment S, and historicals A, B.
Using Epoch=X to indicate run # of historical management duty group.
```

1. [T=1]: Segment S is loaded onto A with 1x replication on that tier.
2. [T=2, Epoch=0]: Epoch 0 starts
3. [T=3, Epoch=0]: Segment S move from A to B is initiated.
4. [T=4, Epoch=1]: Epoch 1 starts
5. [T=5, Epoch=1]: `PrepareBalancerAndLoadQueues` duty starts
6. [T=6, Epoch=1]: `PrepareBalancerAndLoadQueues` initializes its list of `ServerHolder` [here](https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/server/coordinator/duty/PrepareBalancerAndLoadQueues.java#L91). This reads in the in-progress `MOVE_FROM` actions [here](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/server/coordinator/ServerHolder.java#L171) . This does NOT account for the currently-in-flight `DROP` which is about to be fired..
8. [T=7, Epoch=1]: Load to B completes, callback from Epoch=0 load to B fires. This triggers a `DROP` on A, but this is not reflected on the `SegmentLoadQueueManager` accounting, since the drop is issued [here](https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentLoadQueueManager.java#L129) directly to the peon, instead of going through [here](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentLoadQueueManager.java#L70).
9. [T=8, Epoch=1]: `PrepareBalancerAndLoadQueues` ends 
10. [T=9, Epoch=1]: `RunRules` starts
11. [T=10, Epoch=1]: `LoadRules` starts
12. [T=11, Epoch=1]: Seeing the stale value of `SegmentReplicaCount.dropping` from before, it infers that segment S has 2x replication, causing it to issue a randomized drop to either A or B. In the worst case we drop on B, causing segment unavailability until the next Epoch.

### Description

- I've added logic to treat increment the `dropping` counter to include `MOVE_FROM` cases, protecting against the race.
- I've switched to using `SegmentLoadQueueManager::dropSegment` instead of doing on the HttpLoadQueuePeon directly, so that we can get some extra accounting (might not be needed) about what's actually in the queue, since the `HttpLoadQueuePeon`'s queuedSegments is a local copy initialized at construction.

#### Release note

Fixes a race that causes segment unavailability between coordinatory dutygroup runs when a segment is re-balancing.

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentLoadQueueManager.java
+++ b/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentLoadQueueManager.java
@@ -125,8 +125,8 @@ public class SegmentLoadQueueManager
                 && !peonA.getSegmentsToDrop().contains(segment)
                 && (taskMaster.isHttpLoading()
                     || serverInventoryView.isSegmentLoadedByServer(serverNameB, segment))) {
+              peonA.dropSegment(segment, moveFinishCallback);
               peonA.unmarkSegmentToDrop(segment);
-              dropSegment(segment, serverA);
             } else {
               moveFinishCallback.execute(success);
             }
--- a/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentLoadQueueManager.java
+++ b/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentLoadQueueManager.java
@@ -126,7 +126,7 @@ public class SegmentLoadQueueManager
                 && (taskMaster.isHttpLoading()
                     || serverInventoryView.isSegmentLoadedByServer(serverNameB, segment))) {
               peonA.unmarkSegmentToDrop(segment);
-              peonA.dropSegment(segment, moveFinishCallback);
+              dropSegment(segment, serverA);
             } else {
               moveFinishCallback.execute(success);
             }
--- a/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCount.java
+++ b/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCount.java
@@ -66,6 +66,7 @@ public class SegmentReplicaCount
         ++moving;
         break;
       case DROP:
+      case MOVE_FROM:
         ++dropping;
         break;
       default:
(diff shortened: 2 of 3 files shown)


server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentLoadQueueManager.java
```diff
@@ -126,7 +126,7 @@ public boolean moveSegment(
                 && (taskMaster.isHttpLoading()
                     || serverInventoryView.isSegmentLoadedByServer(serverNameB, segment))) {
               peonA.unmarkSegmentToDrop(segment);
-              peonA.dropSegment(segment, moveFinishCallback);
```
bsyk, line 129: Does anything need to call `moveFinishCallback`?

bsyk: Never mind. All that does is `peonA.unmarkSegmentToDrop(segment);` which was done on the line prior 🤷‍♂️

bsyk: So, should the unmark call happen _after_ the drop, or before?

jtuglu1: This just calls `peonA.unmarkSegmentToDrop(segment);` which is already done in the previous line. I think this might be more in the `else` case, where it wants to ensure the segment is marked as dropped.

jtuglu1: Yeah, I think you're right regarding the ordering. I also think there's a reason the internal peon method is called, given it does something with the callback. I'll switch the ordering and leave the `peonA.dropSegment` as-is for now.

kfaraz:
Thanks for finding this issue, @jtuglu-netflix !
It does seem that there is a race condition here after all.
I tried to reproduce it via sim tests and it is reproducible.

Since I had the test anyway, I am putting it up in a PR with a slightly different fix. Hope you don't mind 😅
Could use your review on that PR.

PR #18162, opened 2025-06-20
Fix move race condition, add simulation test

kfaraz:
### Description

Addresses the race condition identified in #18161 
Kudos to @jtuglu-netflix for identifying it!

### Race condition

- Segment is moving from server A to server B
- If it has just finished its MOVE_TO operation on B but the MOVE_FROM operation on A is still pending,
the segment may be chosen to drop from server B causing under-replication
- If it is chosen to drop from server A, that operation would be ignored since the segment already has an ongoing
operation on server A

### Fix

- Use the delta of MOVE_FROM - MOVE_TO to track pending drop callbacks
- Add a simulation test to verify the change. This test fails before the fix.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCount.java
+++ b/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCount.java
@@ -33,7 +33,8 @@ public class SegmentReplicaCount
 
   private int loading;
   private int dropping;
-  private int moving;
+  private int movingTo;
+  private int movingFrom;
 
   /**
    * Increments number of replicas loaded on historical servers.
@@ -63,7 +64,10 @@ public class SegmentReplicaCount
         ++loading;
         break;
       case MOVE_TO:
-        ++moving;
+        ++movingTo;
+        break;
+      case MOVE_FROM:
+        ++movingFrom;
         break;
       case DROP:
         ++dropping;
@@ -110,7 +114,12 @@ public class SegmentReplicaCount
 
   int moving()
   {
-    return moving;
+    return movingTo;
+  }
+
+  int moveCompletedPendingDrop()
+  {
+    return movingFrom - movingTo;
   }
 
   /**
@@ -162,6 +171,6 @@ public class SegmentReplicaCount
 
     this.loading += other.loading;
     this.dropping += other.dropping;
-    this.moving += other.moving;
+    this.movingTo += other.movingTo;
   }
 }
--- a/server/src/test/java/org/apache/druid/server/coordinator/simulate/SegmentBalancingTest.java
+++ b/server/src/test/java/org/apache/druid/server/coordinator/simulate/SegmentBalancingTest.java
@@ -93,6 +93,42 @@ public class SegmentBalancingTest extends CoordinatorSimulationBaseTest
   }
 
   @Test
+  public void testBalancingDoesNotUnderReplicateSegment()
+  {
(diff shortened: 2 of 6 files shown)


server/src/test/java/org/apache/druid/server/coordinator/simulate/SegmentBalancingTest.java
```diff
@@ -92,6 +92,42 @@ private void testBalancingWithAutoSyncInventory(boolean autoSyncInventory)
...
+    // Run another coordinator cycle
+    runCoordinatorCycle();
+    loadQueuedSegmentsSkipCallbacks();
+
+    // Verify that segments have not been dropped from either server since
+    // MOVE_FROM operation is still not complete
+    Assert.assertEquals(10, historicalT11.getTotalSegments());
+    Assert.assertEquals(5, historicalT12.getTotalSegments());
+    verifyNotEmitted(Metric.DROPPED_COUNT);
+    verifyNotEmitted(Metric.MOVED_COUNT);
```
jtuglu1, line 128: I know we are "skipping" running the callbacks to test whether the over-replication will kick in, but it's not a 100% test if we can't confirm the drops do complete even with in-flight callbacks "running." Can we add a check to ensure that afterwards, the drops complete from their respective historicals and the `totalSegments()` are accurate?

server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCount.java
```diff
@@ -110,7 +114,12 @@ int loading()
 
   int moving()
   {
-    return moving;
+    return movingTo;
+  }
+
+  int moveCompletedPendingDrop()
+  {
+    return movingFrom - movingTo;
```
jtuglu1, line 127: Will this ever be < 0? If not we might want to add an assert or something similar

jtuglu-netflix (approved):
LGTM, minor comments. Thanks for the PR!

server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCount.java
kfaraz: It will be never be less than 0.
But we could still do a `Math.max(0, movingFrom - movingTo)` to be on the safe side.
I don't want to add an assert here since it is a critical path.

jtuglu1: Yeah, sounds good. I think it's fair to put a note about this since it's not immediately obvious. Also since we're relying on it being non-negative below:

```java
...
- replicaCountOnTier.moveCompletedPendingDrop();
...
```

having a negative could result in an inflated `projectedReplicas`, which is the behavior this PR is trying to fix.

kfaraz: Added a javadoc and used a `Math.max`.

server/src/test/java/org/apache/druid/server/coordinator/simulate/SegmentBalancingTest.java
kfaraz: Updated, thanks for the suggestion.

kfaraz:
Thanks for the review and finding the issue, @jtuglu-netflix !

PR #18163, opened 2025-06-20
Feat: Intialize ScheduledThreadPoolExecutor with default worker threads

uds5501:
### Description

Updates `workerThread` calculation for `ScheduledThreadPoolExecutor`. It now follows the following logic:
- honor `tuningConfig.workerThreads()` if set.
- if auto scaler config / io config has max task counts specified, set worker threads to `maxTaskCount/4`.
- ensure at least 2 worker threads are spawned if the determined value (`maxTaskCount/4`) is too low.

 * `SeekableStreamSupervisor`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
+++ b/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
@@ -71,12 +71,10 @@ import org.apache.druid.indexing.seekablestream.SeekableStreamIndexTaskTuningCon
 import org.apache.druid.indexing.seekablestream.SeekableStreamStartSequenceNumbers;
 import org.apache.druid.indexing.seekablestream.common.RecordSupplier;
 import org.apache.druid.indexing.seekablestream.supervisor.IdleConfig;
-import org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervisor;
 import org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervisorSpec;
 import org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervisorStateManager;
 import org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervisorTuningConfig;
 import org.apache.druid.indexing.seekablestream.supervisor.TaskReportData;
-import org.apache.druid.indexing.seekablestream.supervisor.autoscaler.AutoScalerConfig;
 import org.apache.druid.indexing.seekablestream.supervisor.autoscaler.LagBasedAutoScalerConfig;
 import org.apache.druid.java.util.common.DateTimes;
 import org.apache.druid.java.util.common.StringUtils;
@@ -5156,77 +5154,6 @@ public class KafkaSupervisorTest extends EasyMockSupport
     EasyMock.replay(differentTaskType);
   }
 
-  @Test
-  public void test_calculateWorkerThreads_shouldHonourWorkerConfig()
-  {
-    final int totalThreads = 5;
-    KafkaSupervisorTuningConfig tuningConfig = getTuningConfig(totalThreads);
-    KafkaSupervisorIOConfig ioConfig = getIOConfig();
-    Assert.assertEquals(totalThreads, SeekableStreamSupervisor.calculateWorkerThreads(tuningConfig, ioConfig));
-  }
-
-  @Test
-  public void test_calculateWorkerThreads_shouldUseDefaultWorkerThreads()
-  {
-    KafkaSupervisorTuningConfig tuningConfig = getTuningConfig();
-    KafkaSupervisorIOConfig ioConfig = getIOConfig();
-    Assert.assertEquals(
-        DEFAULT_WORKER_THREADS,
-        SeekableStreamSupervisor.calculateWorkerThreads(tuningConfig, ioConfig)
-    );
-  }
-
-  @Test
-  public void test_calculateWorkerThreads_shouldUseMinimumWorkerThreadstWithTasks()
-  {
-    KafkaSupervisorTuningConfig tuningConfig = getTuningConfig();
-    KafkaSupervisorIOConfig ioConfig = getIOConfig(7);
-    Assert.assertEquals(
-        DEFAULT_WORKER_THREADS,
-        SeekableStreamSupervisor.calculateWorkerThreads(tuningConfig, ioConfig)
-    );
-  }
-
-  @Test
-  public void test_calculateWorkerThreads_shouldUseFactorOfTaskCount()
-  {
-    KafkaSupervisorTuningConfig tuningConfig = getTuningConfig();
-    KafkaSupervisorIOConfig ioConfig = getIOConfig(18);
-    Assert.assertEquals(
-        4,
-        SeekableStreamSupervisor.calculateWorkerThreads(tuningConfig, ioConfig)
-    );
-  }
-
-  @Test
-  public void test_calculateWorkerThreads_shouldUseAutoScalerConfig()
-  {
-    KafkaSupervisorTuningConfig tuningConfig = getTuningConfig();
-    AutoScalerConfig autoScalerConfig = new LagBasedAutoScalerConfig(
-        null,
-        null,
-        null,
-        null,
-        null,
-        null,
-        null,
-        null,
-        21,
-        null,
-        5,
-        null,
-        null,
-        true,
-        null,
-        null
-    );
-    KafkaSupervisorIOConfig ioConfig = getIOConfig(autoScalerConfig);
-    Assert.assertEquals(
-        5,
-        SeekableStreamSupervisor.calculateWorkerThreads(tuningConfig, ioConfig)
-    );
-  }
-
   private void addSomeEvents(int numEventsPerPartition) throws Exception
   {
     // create topic manually
@@ -5283,97 +5210,6 @@ public class KafkaSupervisorTest extends EasyMockSupport
     }
   }
 
-  private KafkaSupervisorTuningConfig getTuningConfig()
-  {
-    return getTuningConfig(null);
-  }
-
-  /**
-   * Use this method to create a {@link KafkaSupervisorTuningConfig} with custom thread count.
-   *
-   * @param threadCount the number of threads to use, or null to imitate default settings.
-   */
-  private KafkaSupervisorTuningConfig getTuningConfig(@Nullable Integer threadCount)
-  {
-    return new KafkaSupervisorTuningConfig(
-        null,
-        1000,
-        null,
-        null,
-        50000,
-        null,
-        new Period("P1Y"),
-        null,
-        null,
-        null,
-        false,
-        null,
-        false,
-        null,
-        threadCount,
-        TEST_CHAT_RETRIES,
-        TEST_HTTP_TIMEOUT,
-        TEST_SHUTDOWN_TIMEOUT,
-        null,
-        null,
-        null,
-        null,
-        10,
-        null,
-        null
-    );
-  }
-
-  private KafkaSupervisorIOConfig getIOConfig()
-  {
-    return getIOConfig(1, null);
-  }
-
-  private KafkaSupervisorIOConfig getIOConfig(int taskCount)
-  {
-    return getIOConfig(taskCount, null);
-  }
-
-  private KafkaSupervisorIOConfig getIOConfig(AutoScalerConfig autoScalerConfig)
-  {
-    return getIOConfig(1, autoScalerConfig);
-  }
-
-  /**
-   * Use this method to create a {@link KafkaSupervisorIOConfig} with custom properties.
-   */
-  private KafkaSupervisorIOConfig getIOConfig(
-      int taskCount,
-      AutoScalerConfig autoScalerConfig
-  )
-  {
-    final Map<String, Object> consumerProperties = KafkaConsumerConfigs.getConsumerProperties();
-    consumerProperties.put("myCustomKey", "myCustomValue");
-    consumerProperties.put("bootstrap.servers", kafkaHost);
-    return new KafkaSupervisorIOConfig(
-        topic,
-        null,
-        INPUT_FORMAT,
-        1,
-        taskCount,
-        new Period("PT1H"),
-        consumerProperties,
-        autoScalerConfig,
-        null,
-        KafkaSupervisorIOConfig.DEFAULT_POLL_TIMEOUT_MILLIS,
-        new Period("P1D"),
-        new Period("PT30S"),
-        false,
-        new Period("PT30M"),
-        null,
-        null,
-        null,
-        null,
-        null,
-        null,
-        false
-    );
-  }
 
   private TestableKafkaSupervisor getTestableSupervisor(
       int replicas,
--- a/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
+++ b/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
@@ -155,7 +155,6 @@ public class KafkaSupervisorTest extends EasyMockSupport
   private static final long TEST_CHAT_RETRIES = 9L;
   private static final Period TEST_HTTP_TIMEOUT = new Period("PT10S");
   private static final Period TEST_SHUTDOWN_TIMEOUT = new Period("PT80S");
-  private static final int DEFAULT_WORKER_THREADS = 2;
 
   private static TestingCluster zkServer;
   private static TestBroker kafkaServer;
--- a/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
+++ b/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
@@ -71,10 +71,12 @@ import org.apache.druid.indexing.seekablestream.SeekableStreamIndexTaskTuningCon
 import org.apache.druid.indexing.seekablestream.SeekableStreamStartSequenceNumbers;
 import org.apache.druid.indexing.seekablestream.common.RecordSupplier;
 import org.apache.druid.indexing.seekablestream.supervisor.IdleConfig;
+import org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervisor;
 import org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervisorSpec;
 import org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervisorStateManager;
 import org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervisorTuningConfig;
 import org.apache.druid.indexing.seekablestream.supervisor.TaskReportData;
+import org.apache.druid.indexing.seekablestream.supervisor.autoscaler.AutoScalerConfig;
 import org.apache.druid.indexing.seekablestream.supervisor.autoscaler.LagBasedAutoScalerConfig;
 import org.apache.druid.java.util.common.DateTimes;
 import org.apache.druid.java.util.common.StringUtils;
@@ -155,6 +157,7 @@ public class KafkaSupervisorTest extends EasyMockSupport
   private static final long TEST_CHAT_RETRIES = 9L;
   private static final Period TEST_HTTP_TIMEOUT = new Period("PT10S");
   private static final Period TEST_SHUTDOWN_TIMEOUT = new Period("PT80S");
+  private static final int DEFAULT_WORKER_THREADS = 2;
 
   private static TestingCluster zkServer;
   private static TestBroker kafkaServer;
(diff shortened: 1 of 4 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -940,17 +941,8 @@ public SeekableStreamSupervisor(
         spec.isSuspended()
     );
 
-    int workerThreads;
     if (autoScalerConfig != null && autoScalerConfig.getEnableTaskAutoScaler()) {
       log.info("Running Task autoscaler for supervisor[%s] for datasource[%s]", supervisorId, dataSource);
-
-      workerThreads = (this.tuningConfig.getWorkerThreads() != null
```
kfaraz, line 947: I think the number of worker threads should be calculated something like this:
- If `workerThreads` are specified, honor that config.
- If nothing is specified, use a minimum of 2 to 5 (1 seems too aggressive a downscale from the old 10).
- If honoring the task count config (`ioConfig.taskCount` or `autoScale.taskCountMax()`),
core pool size could be some factor of the task count e.g. `taskCount/4` rather than a fixed value.
- For the non-core pool threads, we should perhaps also increase the timeout to a couple of seconds rather than the default of 10 ms.

If these settings continue to do well over a period of time, we can consider reducing them further.

uds5501: > If workerThreads are specified, honor that config.

fair enough, otherwise this PR should've addressed removal of the tuning configs as well. Will accommodate these suggestions.

kfaraz: Thanks, @uds5501 .
Just to clarify, when `workerThreads` is specified, we should use it exactly, without dividing by a factor.
Otherwise, we are changing the semantics of this config which is not desirable and also confusing.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -162,6 +162,9 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
 
   private static final long MAX_RUN_FREQUENCY_MILLIS = 1000;
   private static final int MAX_INITIALIZATION_RETRIES = 20;
+  private static final int MIN_CORE_POOL_SIZE = 2;
```
```suggestion
  private static final int MIN_WORKER_CORE_THREADS = 2;
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -162,6 +162,9 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
...
+  private static final int TASKS_PER_THREAD_FACTOR = 4;
+  private static final int KEEPALIVE_TIME_SECONDS = 2;
```
```suggestion
  private static final int WORKER_THREAD_KEEPALIVE_TIME_SECONDS = 2;
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
kfaraz: We should use the factor only in conjunction with the other two configs, i.e. `taskCount` and `taskCountMax`.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -162,6 +162,9 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
 
   private static final long MAX_RUN_FREQUENCY_MILLIS = 1000;
   private static final int MAX_INITIALIZATION_RETRIES = 20;
+  private static final int MIN_CORE_POOL_SIZE = 2;
+  private static final int TASKS_PER_THREAD_FACTOR = 4;
```
kfaraz: Nit:
```suggestion
  private static final int DEFAULT_TASKS_PER_WORKER_THREAD = 4;
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -970,13 +975,16 @@ public SeekableStreamSupervisor(
...
+        workerThreads,
+        Execs.makeThreadFactory(StringUtils.encodeForFormat(supervisorTag) + "-Worker-%d")
     );
-    log.info("Created worker pool with [%d] threads for supervisor[%s] for dataSource[%s]", workerThreads, this.supervisorId, this.dataSource);
+    executor.setKeepAliveTime(KEEPALIVE_TIME_SECONDS, TimeUnit.SECONDS);
+
+    this.workerExec = MoreExecutors.listeningDecorator(executor);
+    log.info(
+        "Created worker pool with [%d] threads for supervisor[%s] for dataSource[%s]",
+        workerThreads, this.supervisorId, this.dataSource);
```
```suggestion
        "Created worker pool with [%d] threads for supervisor[%s], dataSource[%s].",
        workerThreads, this.supervisorId, this.dataSource
    );
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -970,13 +975,17 @@ public SeekableStreamSupervisor(
...
+        Execs.makeThreadFactory(StringUtils.encodeForFormat(supervisorTag) + "-Worker-%d")
+    );
+    executor.setKeepAliveTime(WORKER_THREAD_KEEPALIVE_TIME_SECONDS, TimeUnit.SECONDS);
+
+    this.workerExec = MoreExecutors.listeningDecorator(executor);
+    log.info(
+        "Created worker pool with [%d] threads for supervisor[%s] for dataSource[%s]",
+        workerThreads, this.supervisorId, this.dataSource
     );
-    log.info("Created worker pool with [%d] threads for supervisor[%s] for dataSource[%s]", workerThreads, this.supervisorId, this.dataSource);
```
cryptoe, line 979: We can rework the log to be more relevant instead of removing it.

uds5501: It's modified and there, kindly take a look

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -970,13 +965,17 @@ public SeekableStreamSupervisor(
...
-        )
+    ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(
+        workerThreads,
+        Execs.makeThreadFactory(StringUtils.encodeForFormat(supervisorTag) + "-Worker-%d")
+    );
+    executor.setKeepAliveTime(WORKER_THREAD_KEEPALIVE_TIME_SECONDS, TimeUnit.SECONDS);
+
+    this.workerExec = MoreExecutors.listeningDecorator(executor);
+    log.info(
+        "Created worker pool with [%d] threads for supervisor[%s] for dataSource[%s]",
```
```suggestion
        "Created worker pool with [%d] threads for supervisor[%s], dataSource[%s].",
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -970,13 +965,17 @@ public SeekableStreamSupervisor(
...
-    this.workerExec = MoreExecutors.listeningDecorator(
-        ScheduledExecutors.fixed(
-            workerThreads,
-            StringUtils.encodeForFormat(supervisorTag) + "-Worker-%d"
-        )
+    ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(
+        workerThreads,
+        Execs.makeThreadFactory(StringUtils.encodeForFormat(supervisorTag) + "-Worker-%d")
+    );
+    executor.setKeepAliveTime(WORKER_THREAD_KEEPALIVE_TIME_SECONDS, TimeUnit.SECONDS);
```
kfaraz: Can this be passed in the constructor itself? Is there any drawback in setting this after the object has been constructed?

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
```diff
@@ -5209,6 +5283,97 @@ private void addMoreEvents(int numEventsPerPartition, int num_partitions) throws
     }
   }
 
+  private KafkaSupervisorTuningConfig getTuningConfig()
+  {
+    return getTuningConfig(null);
+  }
+
+  /**
+   * Use this method to create a {@link KafkaSupervisorTuningConfig} with custom thread count.
```
kfaraz:
```suggestion
   * Creates a {@link KafkaSupervisorTuningConfig} with the given number of worker threads.
```

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
```diff
@@ -5209,6 +5283,97 @@ private void addMoreEvents(int numEventsPerPartition, int num_partitions) throws
...
+   *
+   * @param threadCount the number of threads to use, or null to imitate default settings.
+   */
+  private KafkaSupervisorTuningConfig getTuningConfig(@Nullable Integer threadCount)
```
kfaraz:
```suggestion
  private KafkaSupervisorTuningConfig createTuningConfigWithWorkerThreads(@Nullable Integer numWorkerThreads)
```

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
```diff
@@ -5209,6 +5283,97 @@ private void addMoreEvents(int numEventsPerPartition, int num_partitions) throws
...
+
+  private KafkaSupervisorIOConfig getIOConfig(int taskCount)
+  {
+    return getIOConfig(taskCount, null);
+  }
+
+  private KafkaSupervisorIOConfig getIOConfig(AutoScalerConfig autoScalerConfig)
+  {
+    return getIOConfig(1, autoScalerConfig);
+  }
```
kfaraz: I think these are unnecessary, just use the method with two args.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
```diff
@@ -5209,6 +5283,97 @@ private void addMoreEvents(int numEventsPerPartition, int num_partitions) throws
...
+
+  /**
+   * Use this method to create a {@link KafkaSupervisorIOConfig} with custom properties.
+   */
+  private KafkaSupervisorIOConfig getIOConfig(
```
kfaraz:
```suggestion
  private KafkaSupervisorIOConfig createSupervisorIOConfig(
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
uds5501: > Can this be passed in the constructor itself?

unfortunately, no. I explored the option but couldn't find any. There's no harm in setting this after object creation, the executor does nothing before tasks are pushed in its queue.

kfaraz: Cool, thanks for the confirmation.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1009,6 +1008,22 @@ public Optional<TaskStatus> getTaskStatus(String id)
     this.taskClient = taskClientFactory.build(dataSource, taskInfoProvider, this.tuningConfig, workerExec);
   }
 
+  public static int calculateWorkerThreads(
```
kfaraz, line 1021: Please add a short javadoc about what these worker threads are used for and how their number is computed.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
kfaraz: Nit: I think we should put this in a utility method in `ScheduledExecutors` class itself. Essentially, another flavor of the `fixed` method which accepts timeout millis too.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -940,17 +943,9 @@ public SeekableStreamSupervisor(
         spec.isSuspended()
     );
 
-    int workerThreads;
+    final int workerThreads = calculateWorkerThreads(tuningConfig, ioConfig);
     if (autoScalerConfig != null && autoScalerConfig.getEnableTaskAutoScaler()) {
       log.info("Running Task autoscaler for supervisor[%s] for datasource[%s]", supervisorId, dataSource);
```
kfaraz, line 948: Nit: I guess the log line can move to the new calculate method itself.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
```diff
@@ -2563,6 +2565,77 @@ public LagStats computeLagStats()
     EasyMock.verify(executorService, spec);
   }
 
+  @Test
+  public void test_calculateWorkerThreads_shouldHonourWorkerConfig()
+  {
+    final int totalThreads = 5;
```
kfaraz:
```suggestion
    final int numWorkerThreads = 5;
```

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
```diff
@@ -2630,21 +2703,32 @@ private static DataSchema getDataSchema()
                      .build();
   }
 
-  private static SeekableStreamSupervisorIOConfig getIOConfig()
+  private static SeekableStreamSupervisorIOConfig getSupervisorIOConfig()
```
kfaraz: Nit: Please use the prefix `create` instead of `get` for these new methods (you can leave the existing method names unchanged) since they are creating new objects return than getting some field that is already present in this instance.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
```diff
@@ -2630,21 +2703,32 @@ private static DataSchema getDataSchema()
...
+  {
+    return getSupervisorIOConfig(1, OBJECT_MAPPER.convertValue(getProperties(), AutoScalerConfig.class));
+  }
+
+  /**
+   * Use this method to create a {@link SeekableStreamSupervisorIOConfig} with custom properties.
+   */
```
kfaraz: Not really needed if we name the method appropriately.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
```diff
@@ -2675,13 +2759,28 @@ private static Map<String, Object> getProperties()
   }
 
   private static SeekableStreamSupervisorTuningConfig getTuningConfig()
+  {
+    return getTuningConfigWithWorkerThreads(1);
+  }
+
+  private static SeekableStreamSupervisorTuningConfig getTuningConfigWithWorkerThreads()
```
kfaraz: Nit: since we are not really specifying any value for `workerThreads` in this method, it shouldn't be included in the name.

```suggestion
  private static SeekableStreamSupervisorTuningConfig createSupervisorTuningConfig()
```

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
```diff
@@ -2563,6 +2565,77 @@ public LagStats computeLagStats()
...
+    AutoScalerConfig autoScalerConfig = new LagBasedAutoScalerConfig(
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        21,
```
kfaraz: Assign this value to a field `taskCountMax` so that the test is easier to follow.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
```diff
@@ -2563,6 +2565,77 @@ public LagStats computeLagStats()
...
+        5,
+        null,
+        null,
+        true,
+        null,
+        null
+    );
+    SeekableStreamSupervisorIOConfig ioConfig = getSupervisorIOConfig(1, autoScalerConfig);
+    Assert.assertEquals(
+        5,
```
kfaraz: This would then become:

```suggestion
        taskCountMax / defaultTasksPerThread,
```

kfaraz (commented):
Changes look good.

Left a few style, javadoc suggestions.

+1 after CI passes

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
uds5501: @kfaraz logging that in a method meant to be calculating worker threads seems a bit odd no?

kfaraz: Yeah, I guess you are right. You can leave it as it is then.

kfaraz: Yeah, I guess you are right. You can leave it as it is then.

Edit: The log line seemed closely related since it has to do with task count etc. but it is fine either way.

uds5501: Will be keeping it as is for now then.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1009,6 +1010,33 @@ public Optional<TaskStatus> getTaskStatus(String id)
...
+  /**
+   * Calculates the number of worker threads to use for the supervisor tasks.
+   * <p>
+   * If the tuning config explicitly sets the worker thread count, that value is used.
+   * Otherwise, the value is derived from either the auto-scaler config (if enabled) or the ioConfig task count,
+   * divided by the default number of tasks per worker thread, with a minimum of {@code MIN_WORKER_CORE_THREADS}.
+   *
+   * @param tuningConfig the tuning configuration
+   * @param ioConfig the IO configuration
+   * @return the number of worker threads for executor
```
kfaraz: These param descriptions don't really have any new info and can be omitted.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1009,6 +1010,33 @@ public Optional<TaskStatus> getTaskStatus(String id)
     this.taskClient = taskClientFactory.build(dataSource, taskInfoProvider, this.tuningConfig, workerExec);
   }
 
+  /**
+   * Calculates the number of worker threads to use for the supervisor tasks.
```
```suggestion
   * Calculates the number of worker threads to use in a {@link SeekableStreamSupervisor}.
```

kfaraz: There should also be a line on what these worker threads are actually used for.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1009,6 +1010,33 @@ public Optional<TaskStatus> getTaskStatus(String id)
...
+   * <p>
+   * If the tuning config explicitly sets the worker thread count, that value is used.
```
```suggestion
   * If the tuning config explicitly specifies the field {@code workerThreads}, that value is used.
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1009,6 +1010,33 @@ public Optional<TaskStatus> getTaskStatus(String id)
...
+   * Otherwise, the value is derived from either the auto-scaler config (if enabled) or the ioConfig task count,
+   * divided by the default number of tasks per worker thread, with a minimum of {@code MIN_WORKER_CORE_THREADS}.
```
```suggestion
   * divided by the {@link DEFAULT_TASKS_PER_WORKER_THREAD}, with a minimum of {@link MIN_WORKER_CORE_THREADS}.
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -940,17 +944,9 @@ public SeekableStreamSupervisor(
         spec.isSuspended()
     );
 
-    int workerThreads;
+    final int workerThreads = calculateWorkerThreads(tuningConfig, ioConfig);
```
kfaraz: This line should be moved to where the `workerExec` is being initialized.

processing/src/main/java/org/apache/druid/java/util/common/concurrent/ScheduledExecutors.java
```diff
@@ -172,8 +173,30 @@ public static ScheduledExecutorFactory createFactory(final Lifecycle lifecycle)
...
+   * @param corePoolSize the minimum number of threads in the pool
+   * @param nameFormat   the naming format for threads created by the pool
+   * @return a new {@link ScheduledExecutorService} with the specified configuration
+   */
   public static ScheduledExecutorService fixed(int corePoolSize, String nameFormat)
   {
     return Executors.newScheduledThreadPool(corePoolSize, Execs.makeThreadFactory(nameFormat));
   }
+
+  public static ScheduledThreadPoolExecutor fixedWithKeepAliveTime(
```
kfaraz, line 193: Needs a javadoc.

processing/src/main/java/org/apache/druid/java/util/common/concurrent/ScheduledExecutors.java
```diff
@@ -172,8 +173,30 @@ public static ScheduledExecutorFactory createFactory(final Lifecycle lifecycle)
...
+      int corePoolSize,
+      String nameFormat,
+      long keepAliveTime,
+      TimeUnit timeUnit
```
kfaraz: It might be simpler to just accept milliseconds.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
```diff
@@ -2675,13 +2763,28 @@ private static Map<String, Object> getProperties()
...
+  private static SeekableStreamSupervisorTuningConfig createSupervisorTuningConfig()
+  {
+    return createSupervisorTuningConfigWithWorkerThreads(null);
+  }
+
+  /**
+   * Creates a {@link SeekableStreamSupervisorTuningConfig} with the given number of worker threads.
+   *
+   * @param workerThreads the number of threads to use, or null to imitate default settings.
+   */
```
kfaraz: Nit: Not really needed (especially in tests) unless it's adding info that is not obvious from the method name and signature.

uds5501:
Thanks for your review @kfaraz , I have addressed the final review comments as well.

PR #18165, opened 2025-06-20
Document stopTaskCount supervisor property

writer-jill:
Document supervisor config property `stopTaskCount`.

This PR has:
- [x] been self-reviewed.


jtuglu1:
I think it's worth mentioning that, if configured, the task autoscaler will actually ignore this field (even if set) when shutting down tasks in response to a task count change (since it needs to redistribute partitions across tasks, and that requires all tasks being shut down).

writer-jill:
> I think it's worth mentioning that, if configured, the task autoscaler will actually ignore this field (even if set) when shutting down tasks in response to a task count change (since it needs to redistribute partitions across tasks, and that requires all tasks being shut down).

@jtuglu-netflix Thanks - I've updated the description.

johnImply:
"If set to a value less than taskCount, only up to the specified number of tasks still be cycled at a time ... the longest running tasks will be cycled first"

docs/ingestion/supervisor.md
```diff
@@ -64,6 +64,7 @@ For configuration properties specific to Kafka and Kinesis, see [Kafka I/O confi
...
 |`earlyMessageRejectionPeriod`|ISO 8601 period|Configures tasks to reject messages with timestamps later than this period after the task reached its task duration. For example, if this property is set to `PT1H`, the task duration is set to `PT1H` and the supervisor creates a task at `2016-01-01T12:00Z`, Druid drops messages with timestamps later than `2016-01-01T14:00Z`. Tasks sometimes run past their task duration, such as in cases of supervisor failover.|No||
+|`stopTaskCount`|Integer|Limits the number of ingestion tasks Druid can cycle at any given time. If not set, Druid can cycle all tasks at the same time. If set to a value less than `taskCount` your cluster needs fewer available slots to run the supervisor, so you can save costs by scaling down your ingestion tier. See [stopTaskCount](#stoptaskcount) for more information.|No|`taskCount` value|
```
clintropolis: I know this links to a section, but to me it feels worth explicitly mentioning even here that this increases overall cycle time even though it may be obvious, because if you don't read the other section or read between the lines you might miss that reducing ingestion cost isn't free and operators probably shouldn't use this setting unless they know what they are doing. Maybe something like:
```
so you can save costs by scaling down your ingestion tier in exchange for slower cycle time and possible lag.
```

writer-jill: updated.

docs/ingestion/supervisor.md
```diff
@@ -64,6 +64,7 @@ For configuration properties specific to Kafka and Kinesis, see [Kafka I/O confi
...
 |`earlyMessageRejectionPeriod`|ISO 8601 period|Configures tasks to reject messages with timestamps later than this period after the task reached its task duration. For example, if this property is set to `PT1H`, the task duration is set to `PT1H` and the supervisor creates a task at `2016-01-01T12:00Z`, Druid drops messages with timestamps later than `2016-01-01T14:00Z`. Tasks sometimes run past their task duration, such as in cases of supervisor failover.|No||
+|`stopTaskCount`|Integer|Limits the number of ingestion tasks Druid can cycle at any given time. If not set, Druid can cycle all tasks at the same time. If set to a value less than `taskCount` your cluster needs fewer available slots to run the supervisor, so you can save costs by scaling down your ingestion tier in exchange for slower cycle time and possible lag. See [stopTaskCount](#stoptaskcount) for more information.|No|`taskCount` value|
```
317brian:
```suggestion
|`stopTaskCount`|Integer|Limits the number of ingestion tasks Druid can cycle at any given time. If not set, Druid can cycle all tasks at the same time. If set to a value less than `taskCount`, your cluster needs fewer available slots to run the supervisor. You can save costs by scaling down your ingestion tier, but this can lead to slower cycle times and lag. See [stopTaskCount](#stoptaskcount) for more information.|No|`taskCount` value|
```

317brian: Missing comma also a small nit  that makes the negative much more explicit

317brian (approved):
LGTM

clintropolis (approved):

PR #18168, opened 2025-06-23
Fix metric handling in WindowOperatorQuery and QueryLogicCompat toolchests.

gianm:
Both toolchests were returning "new DefaultQueryMetrics" from their "makeMetrics" method. This caused two issues:

1) The "query(query)" method was not called, so standard dimensions like dataSource, type, etc, were not included.

2) Site-specific DefaultQueryMetrics customizations were not respected.

I think this only really has an impact for the `windowOperator` query, because I don't think `QueryLogicCompat` can be a top-level toolchest (so it won't be used for metrics emission). I adjusted both anyway though.


gianm:
Added `jacoco:skip` because these methods are awkward to test- I think the tests would end up being very tightly coupled to the implementation and wouldn't add a ton of value. If anyone disagrees lmk and I can take another look.

PR #18169, opened 2025-06-23
Sort JSON keys to fix comparator inconsistency

### LINKED ISSUE #18156: Occasional IllegalArgumentException when sorting StructuredData objects
We are ingesting a complex json column and are occasionally running into a bug during ingestion where the task fails with: 

java.lang.RuntimeException: java.lang.IllegalArgumentException: Comparison method violates its general contract!
This occurs prior to the intermediate persist with the following stacktrace:
```
	Caused by: java.lang.IllegalArgumentException: Comparison method violates its general contract!
		at java.base/java.util.TimSort.mergeLo(TimSort.java:781) ~[?:?]
		at java.base/java.util.TimSort.mergeAt(TimSort.java:518) ~[?:?]
		at java.base/java.util.TimSort.mergeCollapse(TimSort.java:448) ~[?:?]
		at java.base/java.util.TimSort.sort(TimSort.java:245) ~[?:?]
		at java.base/java.util.Arrays.sort(Arrays.java:1307) ~[?:?]
		at java.base/java.util.ArrayList.sort(ArrayList.java:1721) ~[?:?]
		at java.base/java.util.stream.SortedOps$RefSortingSink.end(SortedOps.java:392) ~[?:?]
		at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:258) ~[?:?]
		at java.base/java.util.stream.StreamSpliterators$AbstractWrappingSpliterator.fillBuffer(StreamSpliterators.java:210) ~[?:?]
		at java.base/java.util.stream.StreamSpliterators$AbstractWrappingSpliterator.doAdvance(StreamSpliterators.java:161) ~[?:?]
		at java.base/java.util.stream.StreamSpliterators$WrappingSpliterator.tryAdvance(StreamSpliterators.java:298) ~[?:?]
		at java.base/java.util.Spliterators$1Adapter.hasNext(Spliterators.java:681) ~[?:?]
		at org.apache.druid.segment.incremental.IncrementalIndexAdapter.processRows(IncrementalIndexAdapter.java:85) ~[druid-processing-32.0.1.jar:32.0.1]
		at org.apache.druid.segment.incremental.IncrementalIndexAdapter.<init>(IncrementalIndexAdapter.java:65) ~[druid-processing-32.0.1.jar:32.0.1]
		at org.apache.druid.segment.IndexMergerV9.persist(IndexMergerV9.java:1086) ~[druid-processing-32.0.1.jar:32.0.1]
		at org.apache.druid.segment.IndexMerger.persist(IndexMerger.java:237) ~[druid-processing-32.0.1.jar:32.0.1]
		at org.apache.druid.segment.realtime.appenderator.StreamAppenderator.persistHydrant(StreamAppenderator.java:1657) ~[druid-server-32.0.1.jar:32.0.1]
		at org.apache.druid.segment.realtime.appenderator.StreamAppenderator$2.call(StreamAppenderator.java:694) ~[druid-server-32.0.1.jar:32.0.1]
```

This issue is more reliably reproduced with the following query:
```
SELECT
  "complex_json_column",
  COUNT(*) AS "Count"
FROM "TABLE"
GROUP BY 1
ORDER BY 2 DESC
```
With the following stacktrace
```
Caused by: java.lang.IllegalArgumentException: Comparison method violates its general contract!
        at java.base/java.util.TimSort.mergeLo(TimSort.java:781) ~[?:?]
        at java.base/java.util.TimSort.mergeAt(TimSort.java:518) ~[?:?]
        at java.base/java.util.TimSort.mergeCollapse(TimSort.java:448) ~[?:?]
        at java.base/java.util.TimSort.sort(TimSort.java:245) ~[?:?]
        at java.base/java.util.Arrays.sort(Arrays.java:1233) ~[?:?]
        at java.base/java.util.List.sort(List.java:510) ~[?:?]
        at java.base/java.util.Collections.sort(Collections.java:179) ~[?:?]
        at org.apache.druid.query.groupby.epinephelinae.BufferHashGrouper.iterator(BufferHashGrouper.java:204) ~[druid-processing-32.0.1.jar:32.0.1]
        at org.apache.druid.query.groupby.epinephelinae.SpillingGrouper.iterator(SpillingGrouper.java:274) ~[druid-processing-32.0.1.jar:32.0.1]
        at org.apache.druid.query.groupby.epinephelinae.ConcurrentGrouper$1.call(ConcurrentGrouper.java:426) ~[druid-processing-32.0.1.jar:32.0.1]
        at org.apache.druid.query.groupby.epinephelinae.ConcurrentGrouper$1.call(ConcurrentGrouper.java:422) ~[druid-processing-32.0.1.jar:32.0.1]
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[?:?]
        at org.apache.druid.query.PrioritizedListenableFutureTask.run(PrioritizedExecutorService.java:259) ~[druid-processing-32.0.1.jar:32.0.1]
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
```
The exception doesn't occur every time. Sometimes it fails and throws the IllegalArgumentException and other times the query completes successfully but the results vary slightly every time. One observation is that when setting `"groupByIsSingleThreaded": true` the query always completes successfully and with consistent results. 

Worth mentioning is that this column has a mix of plain double values (e.g. `16245.0`) and json objects where each field is a double (e.g. `{"k1": 1.0, "k2": 2.0, "k3": 3.0}`)
### Affected Version

32.0.1

aho135:
Fixes #18156

### Description

When sorting StructuredData objects there is scenario that causes an IllegalArgumentException in TimSort. This occurs when there are two equivalent JSON objects with different key ordering being compared with a third object. This is demonstrated in the test case.

I am not sure if we should sort the keys at ingestion time, or just sort them for the purposes of the hash computation in StructuredData.

#### Release note
Sort keys in JSON objects to resolve a transitivity bug in the StructuredData comparator

<hr>

 * `ColumnSerializerUtils`
 * `StructuredDataTest`

<hr>

This PR has:

- [X] been self-reviewed.

### DIFF
--- a/processing/src/test/java/org/apache/druid/segment/nested/StructuredDataTest.java
+++ b/processing/src/test/java/org/apache/druid/segment/nested/StructuredDataTest.java
@@ -96,6 +96,18 @@ public class StructuredDataTest
   }
 
   @Test
+  public void testCompareToWithDifferentJSONOrder()
+  {
+    StructuredData sd0 = new StructuredData(ImmutableMap.of("D", 0.0, "E", 0.0, "F", 0.0, "A", 0.0, "B", 0.0, "C", 0.0));
+    StructuredData sd1 = new StructuredData(ImmutableMap.of("A", 0.0, "B", 0.0, "C", 0.0, "D", 0.0, "E", 0.0, "F", 0.0));
+    StructuredData sd2 = new StructuredData(ImmutableMap.of("A", 0.0, "B", 0.0, "C", 0.0, "D", 34304.0, "E", 34304.0, "F", 34304.0));
+
+    Assert.assertEquals(1, sd0.compareTo(sd2));
+    Assert.assertEquals(1, sd1.compareTo(sd2));
+    Assert.assertEquals(0, sd0.compareTo(sd1));
+  }
+
+  @Test
   public void testEqualsAndHashcode()
   {
     EqualsVerifier.forClass(StructuredData.class).withIgnoredFields("hashInitialized", "hashValue", "hash").usingGetClass().verify();
--- a/processing/src/main/java/org/apache/druid/segment/serde/ColumnSerializerUtils.java
+++ b/processing/src/main/java/org/apache/druid/segment/serde/ColumnSerializerUtils.java
@@ -20,6 +20,7 @@
 package org.apache.druid.segment.serde;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.dataformat.smile.SmileFactory;
 import com.fasterxml.jackson.dataformat.smile.SmileGenerator;
 import org.apache.druid.guice.BuiltInTypesModule;
@@ -58,6 +59,7 @@ public class ColumnSerializerUtils
     final ObjectMapper mapper = new DefaultObjectMapper(smileFactory, null);
     mapper.getFactory().setCodec(mapper);
     mapper.registerModules(BuiltInTypesModule.getJacksonModulesList());
+    mapper.configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);
     SMILE_MAPPER = mapper;
   }
 


processing/src/main/java/org/apache/druid/segment/nested/StructuredData.java
```diff
@@ -46,7 +47,7 @@ public class StructuredData implements Comparable<StructuredData>
   private static long computeHash(StructuredData data)
   {
     try {
-      final byte[] bytes = ColumnSerializerUtils.SMILE_MAPPER.writeValueAsBytes(data.value);
+      final byte[] bytes = ColumnSerializerUtils.SMILE_MAPPER.writer(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS).writeValueAsBytes(data.value);
```
abhishekrb19: Could you define this as a static final constant `ObjectWriter` in this class so it can be reused if needed instead of instantiating a new object? Also, please include a comment on why this option is needed.

processing/src/test/java/org/apache/druid/segment/nested/StructuredDataTest.java
```diff
@@ -95,6 +95,18 @@ public void testCompareTo()
 
   }
 
+  @Test
+  public void testCompareToWithDifferentJSONOrder()
```
abhishekrb19, line 99: Since this change introduces a subtle behavioral difference in how JSON columns are ingested or queried, could we also include a SQL query test? I think one option could be to add some rows with similar values we  have here to a new column in `CalciteNestedDataQueryTest` and run a `GROUP BY` on the added column. Prior to the sorting behavior in this patch, the `GROUP BY` results with "duplicate values" would differ but now the results would be more inline with mysql with sorted keys?

abhishekrb19 (commented):
Great find, thanks for the fix @aho135! I double-checked and confirmed that this behavior aligns with how [MySQL](https://dev.mysql.com/doc/refman/8.4/en/json.html) handles `json` type, as well as how Postgres handles `jsonb`. I think it will also be good to call out this behavior in the [docs](https://druid.apache.org/docs/latest/querying/nested-columns/). For example, MySQL includes the note:
> To make lookups more efficient, MySQL also sorts the keys of a JSON object. You should be aware that the result of this ordering is subject to change and not guaranteed to be consistent across releases.

processing/src/test/java/org/apache/druid/segment/nested/StructuredDataTest.java
aho135: Thanks for the review @abhishekrb19!
This current implementation only sorts the keys prior to hash computation, but key order remains the same. This is the minimal change to fix the transitivity bug. If we want to align with MySQL and Postgres jsonb then I can make these changes on the ColumnSerializerUtils.SMILE_MAPPER

The current behavior does align with Postgres json. What do you think the preferred behavior should be?

abhishekrb19: Thanks for clarifying. IMO it makes sense to left-shift the sorting all the way to ingest time. Primarily for these reasons:

- Deferring sorting later to hash computation can cause confusing results  — for example, querying the raw data vs querying an aggregation on the json column would yield different results. So sorting at ingest time will help with consistent query results.

- Enables better rollup during ingestion on json columns. Also, it would align with the current behavior of MySQL and Postgres's jsonb type. Postgres's json [type](https://www.postgresql.org/docs/current/datatype-json.html) preserves ingest-time order but doesn't support grouping on those columns.

@aho135 @clintropolis  any thoughts?

aho135:
With this implementation there is an unnecessary sort performed during persists: https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/incremental/IncrementalIndexAdapter.java#L93

Planning to make some follow up changes to differentiate between this scenario and adding a row

processing/src/test/java/org/apache/druid/segment/nested/StructuredDataTest.java
clintropolis: I actually think it is fine to only do the key ordering during comparison, and maybe even preferable since it leaves the 'raw' json identical to how it was ingested.

>Enables better rollup during ingestion on json columns. Also, it would align with the current behavior of MySQL and Postgres's jsonb type. Postgres's json [type](https://www.postgresql.org/docs/current/datatype-json.html) preserves ingest-time order but doesn't support grouping on those columns.

I would think the key ordering would have very limited impact on the rollup factor; the sorting here is arbitrary since there isn't really a natural way to sort json, so it just needs to be 'stable' so we don't run into the error this PR aims to fix (good find btw).

aho135: Thanks for the review @clintropolis!

Sounds good, since key ordering during comparison is the minimal change to resolve the transitivity bug I'll revert to that. I do think in a follow up PR it could be worthwhile to have an option to sort keys at ingestion time though. I would imagine that this has some performance benefits at query time, especially for json with a large number of keys.

aho135: One way I was thinking to do this would be to add a new property in NestedDataColumnSchema (e.g. sortMapKeys). Then we can pass this value to StructuredData so it knows whether it needs to sort map keys before the hash computation

abhishekrb19: Yeah, works for me. We can tackle the ingest-time ordering in a follow up if needed.

abhishekrb19 (approved):

PR #18104, opened 2025-06-09
Make existing quidem tests run on Dart as well

weishiuntsai:
### Description

This PR makes the existing quidem tests run on Dart as well.  The last fix https://github.com/apache/druid/pull/18088 makes this possible now.  There are no new queries added; this change just allows the existing queries to run on the Dart engine along with sql-native and MSQ.

This PR also includes a couple of minor cleanups:
- remove `!set dartQueryId 00000000-0000-0000-0000-000000000000`, which is no longer needed.
- merge some identical iq files from different engines back to one using `AllDruidEnginesComponentSupplier`.

This PR has:

- [X] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [ ] been tested in a test Druid cluster.

PR #18171, opened 2025-06-24
Move QTest execution into a separate job during CI checks and some cleanup

kgyrtkirk:
* move QTest execution into a separate job during CI checks
* removed approved/unapproved logic - shouldn't make that much difference as tests are run in around 20 minutes
* rebalanced test splits


.github/workflows/ci.yml
```diff
@@ -73,9 +73,25 @@ jobs:
...
-        pattern: [ "A*,G*,R*", "B*,O*,S*,X*,Y*,Z*", "C*,E*", "D*,J*,K*", "F*,H*,U*", "I*,N*,T*", "L*,Q*,W*", "M*,P*,V*"]
+        pattern: [ "C*", "H*,U*,V*", "N*,Q*,S*", "B*,O*,R*", "G*,J*,K*", "F*,L*,M*", "A*,D*,I*,X*,Y*,Z*", "E*,P*,T*,W*"]
     uses: ./.github/workflows/worker.yml
     with:
-      script: .github/scripts/run-unit-tests.sh -Dtest='${{ matrix.pattern }}' -fae
+      script: .github/scripts/run_unit-tests -Dtest='${{ matrix.pattern }}' -fae
       jdk: ${{ matrix.jdk }}
       key: "test-jdk${{ matrix.jdk }}-[${{ matrix.pattern }}]"
+
+  # this will be running other ITs later; but until that migration happens - run QTest in parallel to other tests
```
capistrant: This comment is worded that you want QTest in parallel with other tests, but right now it depends on test-report which won't run until after `run-unit-tests` has finished, leaving these QTests to run at very end of the `ci.yml` execution. Is that what you want or does the `needs` need to be updated here?

kgyrtkirk: damn....forgot to remove the `needs`...
in the end that should be the layout - but for now we could relax that a bit until all the real `its` will be moved in...

update the pr

capistrant:
Sad times when some ITs fail when the PR doesn't touch the druid code -_- ... I think this looks good though. will give Akshat chance to review during his daytime hours as well. Will check back in during my morning Wednesday

kfaraz:
@kgyrtkirk , test distribution looks good, each of new splits finishes in about 20 minutes. But the `QTest` took 38 minutes to finish, is this expected?

kgyrtkirk:
@kfaraz: unfortunately - its already over 20 minutes :disappointed:  not something I was expecting earlier... 

...and with #18104 it would go up to around `48m`...I also feel like this won't be enough maybe it should be split into smaller testclasses; which then run as separate splits

kfaraz:
> ...and with https://github.com/apache/druid/pull/18104 it would go up to around 48m...I also feel like this won't be enough maybe it should be split into smaller testclasses; which then run as separate splits

I see. Yes, splitting it up might be worth a shot otherwise it will always become the CI bottleneck.

kgyrtkirk:
@kfaraz can we do that in a followup?
I think that should be discussed with @weishiuntsai ; but #18104 was not merged as I wanted to do something to make this a little better - and I've stalled it for over a week or even more...

kfaraz:
Sure, @kgyrtkirk , that works too.

.github/workflows/ci.yml
```diff
@@ -73,9 +73,24 @@ jobs:
...
+  run-its:
```
Akshat-Jain: Why call it `run-its` instead of something like `run-QTest`?

IIUC, when ITs are eventually integrated into this file, they would be in a different job than this one? (as in, we'd have both `run-QTest` and `run-its` separately?)

Also, if that's the case, we could even simplify this `run-QTest` spec to not have matrix, and just specify the key and script values directly?

Let me know if I'm missing something. Thanks!

kgyrtkirk: makes sense...I was coming back from a branch I was experimenting with migrating the `its` - but it does make sense to run these in a separate matrix

Akshat-Jain (approved):

kgyrtkirk:
Thank you @capistrant , @kfaraz  and @Akshat-Jain  for taking a look!

PR #18176, opened 2025-06-26
experimental virtual storage fabric mode for historicals

clintropolis:
### Description
This PR adds a new experimental virtual storage fabric mode for historical servers. This is a foundational piece that focuses on allowing the segment cache to behave like an actual cache, decoupling the size of data that historical servers are responsible for from the size of their actual disk.

The mode can be enabled by setting `druid.segmentCache.virtualStorage=true`, with the basic idea being that `druid.server.maxSize` is set to the size of data which historicals will be responsible for serving, while `druid.segmentCache.locations` is configured with the actual physical disk sizes that the historical can have loaded simulataneously. When operating in this mode, historicals will not download segments upon assignment, but rather download them on demand during queries. Segments will remain in the cache until it is needed for other segments, at which point they will be removed from disk, but still 'served' by the historical.

There is no change to balancing logic, as far as the coordinator is concerned, historicals configured this way are able to serve `druid.server.maxSize` worth of segments.

This mode does have a cost right now, the biggest being that segment metadata queries are purposefully not currently able to trigger a segment download. The reason for this is because by default the broker computes the SQL schema by issuing segment metadata queries for every segment that is announced by historicals, so if allowed to download then every historical would need to download every segment they are responsible for on startup. This means however that historicals running in this mode are not able to compute the SQL schema, so instead to use SQL clusters either need to have a regular tier of historicals which can compute the schema, or use the catalog or centralized schema functionality (where the segment schemas are computed at ingest time and published to metadata and the broker fetches them from the coordinator).

This PR also hasn't yet wired up support for dart queries to use virtual storage fabric mode, will either add to this PR or do as a follow-up depending on if it gets finished before this is merged or not.

Lastly, there are no metrics for this functionality yet, that will be coming in a follow-up PR.

#### `StorageLocation` changes
The first area of change is that `StorageLocation` has been essentially rewritten to use a new `CacheEntry` abstraction for the items it manages. The interface looks like

```
public interface CacheEntry
{
  CacheEntryIdentifier getId();
  long getSize();
  boolean isMounted();
  void mount(StorageLocation location) throws IOException, SegmentLoadingException;
  void unmount();
}
```

where `CacheEntryIdentifier` is just a generic interface with `@SubclassesMustOverrideEqualsAndHashCode` so that it can be used.

To support virtual storage, `StorageLocation` now provides two different types of cache entry management, 'static' and 'weak' entries. Static entries are the equivalent of how `StorageLocation` works prior to this PR, items are reserved and stay in the `StorageLocation` until explicitly removed. The weak entries are a new way that `StorageLocation` can manage cache entries and allows it to behave like an actual cache. Weak entres remain in the `StorageLocation` until it reaches its size limit, at which point adding any new static or weak entries will try to reclaim space by unmounting weak entries until it can successfully store the new entry.

To try to avoid one time use weak entries from evicting commonly used entries, weak entries are chosen for eviction using an algorithm based on [SIEVE](https://www.usenix.org/system/files/nsdi24spring_prepub_zhang-yazhuo.pdf), with additional mechanisms to place temporary holds on weak cache entries. The holds are required to support cases such as if a group of weak cache entries are taking part in a query and must all be loaded simultaneously before processing can continue.

#### `SegmentManager`, `SegmentCacheManager`, and `SegmentMapFunction` changes
The next layer of changes is to `SegmentManager` and `SegmentCacheManager`, which have both been heavily modified to try to simplify logic and also support new functionality of being able to fetch and mount segments on demand as required by the virtual storage fabric mode.

Perhaps the most significant change for callers of stuff is that `SegmentManager` now maintains the timeline as a `VersionedIntervalTimeline<String, DataSegment>>` instead of `VersionedIntervalTimeline<String, ReferenceCountedSegmentProvider>`. The flow is now to get the timeline to find the `DataSegment` to use, then use one of two new methods,

```
public Optional<Segment> acquireSegment(DataSegment dataSegment)
```

which retrieves a segment that is already present in the cache, or 

```
public AcquireSegmentAction acquireSegment(DataSegmentAndDescriptor segmentAndDescriptor) throws SegmentLoadingException
```

with `AcquireSegmentAction` being a closeable wrapper for a future to provide the `Optional<Segment>` to support download and mount of a segment on demand. There are corresponding methods on the `SegmentCacheManager` interface that `SegmentManager` wraps.

So the flow for callers is to get a `Optional<Segment>` returned by either of these methods, do segment stuff with the `Optional<Segment>` if present, and then close it - and for the latter method, the `AcquireSegmentAction` must also be closed to release holds it might have on a `StorageLocation`. `ReferenceCountedSegmentProvider` is now an internal implementation detail of `SegmentLocalCacheManager`, callers no longer need to concern themselves with reference counting or acquiring references from a provider since it is now baked into the interaction through the new `SegmentCacheManager` methods.

The change from `ReferenceCountedSegmentProvider` also means that `SegmentMapFunction` has changed, now it is simply `Function<Optional<Segment>, Optional<Segment>>` since the call to `ReferenceCountedSegmentProvider.acquireReference` is hidden behind the `SegmentCacheManager` methods.

#### `SegmentLocalCacheManager` and virtual storage fabric mode

To account for the changes in `StorageLocation`, `SegmentLocalCacheManager` handles the segments with a `SegmentCacheEntry` implementation of `CacheEntry`, which uses the `SegmentId` as the cache key and abstract the process of downloading and memory mapping the segment to create a `ReferenceCountedSegmentProvider` in the mount method, while unmount will close the reference provider and delete the files from disk.

`SegmentLocalCacheManager` is also where the magic happens when the new `druid.segmentCache.isVirtualStorageFabric` config is set. In this mode, all segment load requests to `SegmentManager` will simply add a `DataSegment` entry to the timeline instead of creating a `SegmentCacheEntry` and mounting it, while the load request on `SegmentLocalCacheManager` will be a no-op.

Later, when `acquireSegment` is called, `SegmentLocalCacheManager` checks all `StorageLocation` if the `DataSegment` is already stored as an entry, and if not, creates a new weak `SegmentCacheEntry` and mounts it on a shared work pool, which becomes the future of the `AcquireSegmentAction`. The hold on the weak entry is tied to closing the `AcquireSegmentAction`. 

#### `ServerManager` and other `SegmentManager`/`SegmentCacheManager` caller changes
`ServerManager` and other callers have been adjusted to deal with the timeline changes of `SegmentManager`, and all query types except for segment metadata have been adjusted to use the new `AcquireSegmentAction` flow, allowing queries to download segments on demand using the new virtual storage fabric mode. To not behave badly with the shared processing pool, `ServerManager` will create an `AcquireSegmentAction` for every segment taking part in the query to ensure that it can load all segments to disk prior to scheduling any processing on the shared processing pool. This means that if a query requires more segments than can actively fit on disk, the query will fail.

#### Release note
_todo: some release_

<hr>
This PR has:

- [ ] been self-reviewed.


jtuglu1:
👍  Might be good to add a link to the original SIEVE paper: https://www.usenix.org/conference/nsdi24/presentation/zhang-yazhuo for those interested

kfaraz:
@clintropolis , I haven't gone through the PR yet but how will this affect segment assignment/balancing on the Coordinator?

- The Coordinator currently assigns segments to a Historical only if it has available disk space.
- The available disk space is calculated in `ImmutableDruidServer` based on the configured `maxSize`
and the sum of the `DataSegment.getSize()` of all the segments currently present on this historical
(as reported by the `HttpServerInventoryView`)
- I suppose the balancing algorithm can remain unchanged as it is already designed to optimize query performance
rather than disk usage.

clintropolis:
>👍 Might be good to add a link to the original SIEVE paper: https://www.usenix.org/conference/nsdi24/presentation/zhang-yazhuo for those interested

yea totally, I have it linked in javadocs for `StorageLocation` and was planning to add it to the PR description once I fill it in when this is closer to ready to review just haven't got to it yet 😅 (still changing quite a few things).

>@clintropolis , I haven't gone through the PR yet but how will this affect segment assignment/balancing on the Coordinator?

This first PR has no changes needed to the coordinator logic. When a historical is set to this mode, the idea is that you set `druid.server.maxSize` to how much data you want it to be responsible for, but set the sizes in `druid.segmentCache.locations` to be the actual disk sizes, and during querying the cache manager will load and drop segments internally as appropriate to stay within the constraints of the `druid.segmentCache.locations` sizes. I'll elaborate more in the PR description once this branch gets closer to review ready.

I do have some ideas for follow-up work of adding a new type of "weak" load rule to allow historicals to use the same load logic it does for all segments when `druid.segmentCache.isVirtualStorageFabric` from this PR is set to true, but also still have regular segment loads. This would allow for finer grained control over how segments are loaded by allowing some segments to be sticky and always present in the disk cache (the cache manager supports this internally), while others would be weak references and load on demand but be eligible to be dropped if new strong or weak loads need the space. This likely does require some adjustments to coordinator balancing to distinguish 'weak' loads from regular loads to ensure the regular loads to exceed actual disk space, but I think the changes would be pretty minor. Was planning to address this too in the PR description (or maybe a linked design proposal issue, not sure, haven't decided yet).

jtuglu1:
@clintropolis now that segments are being cached on-demand on historicals (not all pre-loaded), and intra-AZ requests are much faster than S3 reads (intra-AZ requests are typically XXXµs, whereas S3 is ~500 ms for single RTT not to mention the numerous RTT needed from downloading massive segments), is this moving towards the goal to make historicals more "stateless" and more of a distributed caching tier? i.e historical A can pull column C from segment D from historical B? Basically improve "hit" rate and spread query load by allowing historicals to pass segment data between each other?

gianm:
> @clintropolis now that segments are being cached on-demand on historicals (not all pre-loaded), and intra-AZ requests are much faster than S3 reads (intra-AZ requests are typically XXXµs, whereas S3 is ~500 ms for single RTT not to mention the numerous RTT needed from downloading massive segments), is this moving towards the goal to make historicals more "stateless" and more of a distributed caching tier? i.e historical A can pull column C from segment D from historical B? Basically improve "hit" rate and spread query load by allowing historicals to pass segment data between each other?

Initially our thinking was to focus on partial fetches from S3, so we only have to fetch the columns actually needed for a query. From my PoV (@clintropolis may have his own opinion) from there it would make sense to have the Druid servers able to fetch from each other. It would make it more feasible to burst up compute for a short period of time (~seconds even would make sense).

Btw, I would say Historicals are already "stateless" in that they are just a cache of immutable segments (backed by S3) + some compute. The initial work we're doing with virtual storage is enabling the cache to be populated on demand from S3 during a query, instead of needing to be populated before the query starts. This improves things in two main ways:

- Historicals are immediately usable after being booted up; don't need to populate their cache before they usable for queries
- you can have more data in S3 than available disk on Historicals, if you like

Later on, fetching data from each other would be a way to improve the latency of cache population, which will have the effect of making the servers more responsive immediately following a cold start.

abhishekagarwal87:
@clintropolis - how do segment metadata queries work that touch lots of segments and are system-generated? We wouldn't want to fetch all the segments to build sql schema on broker.

cecemei:
Curious how is `SegmentDescriptor` used in `AcquireSegmentAction`? I thought it's enough to just use `DataSegment`? Also I kinda feel many `SegmentManager` logic could be moved to `DataSourceState` class, which could encapsulate better, e.x. segments must be all joinable or non joinable.

clintropolis:
> Initially our thinking was to focus on partial fetches from S3, so we only have to fetch the columns actually needed for a query. From my PoV (@clintropolis may have his own opinion) from there it would make sense to have the Druid servers able to fetch from each other. It would make it more feasible to burst up compute for a short period of time (~seconds even would make sense).

Thanks for chiming in on this @gianm, I agree the next big area I'm probably going to be looking at is partial downloads, which will build on top of some of the stuff here (with some modifications) and introduce some new machinery to do downloads at a finer granularity than the whole segment that this requires.

After that I agree it makes sense to allow pulling 'locally' from other servers if available instead of deep storage after we get all the foundations in place.

clintropolis:
>@clintropolis - how do segment metadata queries work that touch lots of segments and are system-generated? We wouldn't want to fetch all the segments to build sql schema on broker.

Yea, there is a bit of ugly special handling for segment metadata queries right now https://github.com/apache/druid/pull/18176/files#diff-3d43c1b27316a889f58c28a3373b157f38f633d40d6ab4495d350dece7ce7a44R293 that doesn't allow it to download segments for this reason. This means that to use SQL you currently either need a non-virtual tier, or if only using a virtual storage mode tier then either the catalog or centralized schema stuff must be relied on to supply the SQL schema. 

In the future, I have some segment format changes in mind to support partial downloads that would also allow us to have the schema available without downloading the whole segment, so I think this can be a temporary problem.

clintropolis:
>Curious how is SegmentDescriptor used in AcquireSegmentAction? I thought it's enough to just use DataSegment? 

`SegmentDescriptor` isn't strictly used by the `AcquireSegmentAction`, its more of a bit of tracking convenience for callers, which start out with a list of `SegmentDescriptor`, so that it is easy to associate the segment loading futures with the `SegmentDescriptor` originally requested.

>Also I kinda feel many SegmentManager logic could be moved to DataSourceState class, which could encapsulate better, e.x. segments must be all joinable or non joinable.

I would agree that there are possibly some changes that could be done here, but I'm not too motivated to do them in this PR since enough other stuff is already changing. At least `DataSourceState` is basically internal to `SegmentManager`, so it should be pretty easy to change stuff around later if we want.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -65,35 +74,35 @@ public class SegmentLocalCacheManager implements SegmentCacheManager
...
-   * {@link #createOrGetLock}. And then, all threads compete with each other to get the lock.
-   * Finally, the lock should be released using {@link #unlock}.
+   * {@link #lock(DataSegment)}. And then, all threads compete with each other to get the lock.
+   * Finally, the lock should be released using {@link #unlock(DataSegment, ReferenceCountingLock)}.
    *
    * An example usage is:
    *
-   * final ReferenceCountingLock lock = createOrGetLock(segment);
+   * final SegmentCacheEntry entry = new SegmentCacheEntry(segment);
+   * final ReferenceCountingLock lock = entry.lock();
```
cecemei: `entry.lock()` is not true? we could maybe use `entry.lock(segmentLocks)`. also this lock map seems to be the guard for concurrent segment operation, might be worth calling that out, since `SegmentCacheEntry` is not thread safe.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -163,12 +193,45 @@ public List<DataSegment> getCachedSegments() throws IOException
...
-        } else if (isSegmentCached(segment)) {
-          cachedSegments.add(segment);
         } else {
+          removeInfo = true;
+          final SegmentCacheEntry cacheEntry = new SegmentCacheEntry(segment);
+          for (StorageLocation location : locations) {
+            // check for migrate from old nested local storage path format
+            final File legacyPath = new File(location.getPath(), DataSegmentPusher.getDefaultStorageDir(segment, false));
+            if (legacyPath.exists()) {
+              FileUtils.mkdirp(new File(location.getPath(), segment.getId().toString()));
```
cecemei: this is the same as `toPotentialLocation`?

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -163,12 +193,45 @@ public List<DataSegment> getCachedSegments() throws IOException
...
+              Files.move(legacyPath.toPath(), cacheEntry.toPotentialLocation(location.getPath()).toPath(), StandardCopyOption.REPLACE_EXISTING);
+              cleanupLegacyCacheLocation(location.getPath(), legacyPath);
+            }
+
+            if (cacheEntry.checkExists(location.getPath())) {
```
cecemei, line 221: nit: this `checkExists` is not quite necessary since we already have `toPotentialLocation`

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -209,382 +272,172 @@ public void removeInfoFile(DataSegment segment)
...
-   * {@link StorageLocation#maybeReserve(String, DataSegment)} for more details.
-   * If the segment files are damaged in any location, they are removed from the location.
-   * @param segment - Segment to check
-   * @return - Path corresponding to segment directory if found, null otherwise.
-   */
-  @Nullable
-  private File findStoragePathIfCached(final DataSegment segment)
+  public Optional<Segment> acquireSegment(final DataSegment dataSegment)
   {
+    final SegmentCacheEntryIdentifier cacheEntryIdentifier = new SegmentCacheEntryIdentifier(dataSegment.getId());
```
cecemei, line 286: Is there any specific reason to create `SegmentCacheEntryIdentifier` class rather than just use a `SegmentId`?

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -209,382 +272,172 @@ public void removeInfoFile(DataSegment segment)
...
-          cleanupCacheFiles(location.getPath(), localStorageDir);
-          location.removeSegmentDir(localStorageDir, segment);
-          break;
-        } else {
-          // Before returning, we also reserve the space. Refer to the StorageLocation#maybeReserve documentation for details.
-          location.maybeReserve(storageDir, segment);
-          return localStorageDir;
-        }
+      if (location.isReserved(cacheEntryIdentifier)) {
+        final SegmentCacheEntry cacheEntry = location.getCacheEntry(cacheEntryIdentifier);
```
cecemei: `getCacheEntry` is only called if it's static reserved? any reason not to visit before `isReserved` check?

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -595,96 +448,107 @@ public void cleanup(DataSegment segment)
...
+            }
+          }
         }
-      }
-    }
+    );
   }
 
-  private ReferenceCountingLock createOrGetLock(DataSegment dataSegment)
+  private ReferenceCountingLock lock(final DataSegment dataSegment)
```
cecemei, line 589: would the value of `segmentLocks` even be null? i mean, why not just do `segmentLocks.computeIfAbsent`?

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
clintropolis: oops, this is stale javadoc that reflects the state of an earlier commit, will fix

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
clintropolis: yea, the flattening of locations was a change in the last couple commits so i could remove a big shared lock, will clean up

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
clintropolis: do you mean we don't need the `checkExists` function and this line should just be 
```if (cacheEntry.toPotentialLocation(location.getPath()).exists())```?

File: server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java (hunk 4 above)
Comment (clintropolis) [reply server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java:286]: are you suggesting changing `SegmentId` to implement `CacheEntryIdentifier`? I did consider this, and it would probably be ok, though the main reason I didn't is because I wasn't sure where to stop. For this to work, I would need to move `CacheEntryIdentifier` into `druid-processing`, and if that is there then also maybe `CacheEntry` since it would be a bit weird to have the identifier there without the thing that uses it. At that point, do we also move `StorageLocation` since `CacheEntry` and the identifier has no other uses currently beyond that... anyway, that is why I haven't done it yet.

I actually do think it would probably be ok to move all 3 of those things to druid-processing, but was sort of waiting for this discussion :shrug:

File: server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java (hunk 6 above)
Comment (clintropolis) [reply server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java:589]: we remove the locks when the count reduces to 1 on release so that we don't have to keep a lock per segment in memory if nothing is locking it

File: server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java (hunk 5 above)
Comment (clintropolis) [reply server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java]: its not really safe to use a weak entry this way without going through a `ReservationHold`, otherwise the `StorageLocation` could remove and unmount it before we attempt to grab a reference to the segment, so this method only works with static entries on purpose, while the underlying `getCacheEntry` is less restrictive because it is shared with the bootstrapping code that runs on startup. `getCachedSegments` checks for the files and does the reservations, but does not call mount so that it can run the mounts in a threadpool instead, so the bootstrapper iterates over those and then calls boostrap which calls `getCacheEntry` and mount if the entry exists (but does not create any new reservations).

File: server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java (hunk 4 above)
Comment (cecemei) [reply server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java:286]: gotcha - i was just wondering maybe we could simplify the class

File: server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java (hunk 6 above)
Comment (cecemei) [reply server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java:589]: ah i see, the increment/decrement has to be the same atomic operation, sg

File: server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java (hunk 3 above)
Comment (cecemei) [reply server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java:221]: yea, seems like one less method unless there's more in other possible locations.

File: server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java (hunk 5 above)
Comment (cecemei) [reply server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java]: ah i see, so the part of visit weak cache is not really reachable here, i wonder maybe we can add a getReserved which returns null if it's not in static cache.

File: embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageFabricTest.java [hunk 7]
```diff
@@ -0,0 +1,261 @@
...
+  private final EmbeddedCoordinator coordinator = new EmbeddedCoordinator();
+  private final EmbeddedRouter router = new EmbeddedRouter();
+  private final MinIOStorageResource storageResource = new MinIOStorageResource();
+
+  private EmbeddedMSQApis msqApis;
+
+  @Override
+  public EmbeddedDruidCluster createCluster()
+  {
+    historical.addProperty("druid.segmentCache.isVirtualStorageFabric", "true")
```
gianm: the property seems oddly verbose to me. how about `druid.segmentCache.vsf` or `druid.segmentCache.virtualStorage`. same goes for `minVirtualStorageFabricLoadThreads`, I'd suggest like `minVsfLoadThreads` or `minVirtualStorageLoadThreads`.

processing/src/main/java/org/apache/druid/java/util/common/concurrent/Execs.java
```diff
@@ -173,6 +175,97 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor executor)
     );
   }
 
+  public static ExecutorService newBlockingCached(
```
gianm: How about folding these parameters into the existing `newBlockingThreaded`? It's only used internally by `Execs` so it should be fine to change it.

processing/src/main/java/org/apache/druid/java/util/common/concurrent/Execs.java
```diff
@@ -173,6 +175,97 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor executor)
...
+          try {
+            executor.getQueue().put(r);
+          }
+          catch (InterruptedException e) {
+            throw new RejectedExecutionException("Got Interrupted while adding to the Queue", e);
+          }
+        }
+    )
+    {
+      private int running = 0;
```
gianm: `@GuardedBy("this")`

processing/src/main/java/org/apache/druid/java/util/common/concurrent/Execs.java
```diff
@@ -173,6 +175,97 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor executor)
...
+
+      @Override
+      public void execute(Runnable command)
+      {
+        synchronized (this) {
+          running++;
+          growIfNeeded();
```
gianm: what's all this for? Don't the builtin caching thread pools handle this sort of thing on their own?

processing/src/main/java/org/apache/druid/query/metadata/SegmentAnalyzer.java
```diff
@@ -364,4 +374,85 @@ private ColumnAnalysis analyzeArrayColumn(final ColumnCapabilities capabilities)
   {
     return ColumnAnalysis.builder().withCapabilities(capabilities).build();
   }
+
+  public static Segment makeVirtualPlaceholderSegment(DataSegment dataSegment)
```
gianm: IMO, even though this is only needed by `SegmentAnalyzer`, it would be best to move it to its own file. It's only tangentially related to the main functionality of `SegmentAnalyzer`.

processing/src/main/java/org/apache/druid/segment/SegmentMapFunction.java
```diff
@@ -23,24 +23,22 @@
...
+ * Functional interface that captures the process of transforming a {@link Segment} to another {@link Segment} if
+ * possible.
  * <p>
- * The {@link Segment} returned by this method, if present, must always be closed to return the reference to the base
- * {@link ReferenceCountedObjectProvider<Segment>} that it came from.
+ * The {@link Segment} returned by this method, if present, must always be closed by the caller.
  */
 @FunctionalInterface
-public interface SegmentMapFunction extends Function<ReferenceCountedObjectProvider<Segment>, Optional<Segment>>
+public interface SegmentMapFunction extends Function<Optional<Segment>, Optional<Segment>>
```
gianm, line 32: when can the input be absent? Is there anything smart a SegmentMapFunction can do with absent input other than return absent output? I'm asking because I'm wondering if this should be `Function<Segment, Optional<Segment>>` and the caller would just not call it for absent input.

server/src/main/java/org/apache/druid/segment/loading/CacheEntry.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Size in bytes of the cache entry
+   */
+  long getSize();
+
+  boolean isMounted();
+
+  /**
+   * Materializes the cache entry into the assigned {@link StorageLocation}. If a cache entry is already mounted in the
```
gianm, line 42: Does this arrange for it to be materialized in the background, or does it block?

server/src/main/java/org/apache/druid/segment/loading/CacheEntry.java
```diff
@@ -0,0 +1,53 @@
...
+   * location, calling this method should be a no-op. If the cache entry is mounted in a different location, this method
+   * will unmount the item from the other location and mount in the new location.
+   */
+  void mount(File location) throws IOException, SegmentLoadingException;
+
+  /**
+   * Removes the physical artifacts of a cache entry from the location it is currently mounted
```
gianm, line 49: Does this arrange for background removal, or does it block?

server/src/main/java/org/apache/druid/segment/loading/SegmentCacheManager.java
```diff
@@ -39,116 +43,96 @@ public interface SegmentCacheManager
...
-   * @see SegmentCacheManager#getSegment(DataSegment)
+   * @throws SegmentLoadingException - If there is an error in loading the segment or insufficient storage space
+   * @see SegmentCacheManager#load(DataSegment)
+   * @see SegmentCacheManager#shutdownBootstrap()
    */
-  ReferenceCountedSegmentProvider getBootstrapSegment(
-      DataSegment segment,
-      SegmentLazyLoadFailCallback loadFailed
-  ) throws SegmentLoadingException;
+  void bootstrap(DataSegment segment, SegmentLazyLoadFailCallback loadFailed) throws SegmentLoadingException;
```
gianm, line 91: `loadOnBootstrap`? I was thinking having both start with `load` emphasizes the similarities.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -209,382 +278,173 @@ public void removeInfoFile(DataSegment segment)
...
-   * @return - True if segment was downloaded successfully, false otherwise.
-   */
-  private boolean loadInLocationWithStartMarkerQuietly(StorageLocation loc, DataSegment segment, File storageDir, boolean releaseLocation)
+  @Override
+  public void load(final DataSegment dataSegment) throws SegmentLoadingException
   {
-    try {
-      loadInLocationWithStartMarker(segment, storageDir);
-      return true;
+    if (config.isVirtualStorageFabric()) {
```
gianm: Make it an error to enable vsf and also set `numThreadsToLoadSegmentsIntoPageCacheOnDownload`? Seems like in vsf mode, that config is ignored.

server/src/main/java/org/apache/druid/segment/loading/CacheEntry.java
```diff
@@ -0,0 +1,53 @@
...
+   * Size in bytes of the cache entry
+   */
+  long getSize();
+
+  boolean isMounted();
+
+  /**
+   * Materializes the cache entry into the assigned {@link StorageLocation}. If a cache entry is already mounted in the
+   * location, calling this method should be a no-op. If the cache entry is mounted in a different location, this method
+   * will unmount the item from the other location and mount in the new location.
```
gianm: > If the cache entry is mounted in a different location, this method will unmount the item from the other location and mount in the new location

This seems interesting, I'm wondering what's the use case for it?

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
```diff
@@ -0,0 +1,148 @@
...
+  {
+    final Closer safetyNet = Closer.create();
+    Throwable failure = null;
+
+    // getting the future kicks off any background action, so materialize them all to a list to get things started
+    final List<Future<Optional<Segment>>> futures = new ArrayList<>(acquireSegmentActions.size());
+    for (AcquireSegmentAction acquireSegmentAction : acquireSegmentActions) {
+      safetyNet.register(acquireSegmentAction);
+      // if we haven't failed yet, keep collecing futures (we always want to collect the actions themselves though
+      // to close
```
gianm: comment appears to be cut off

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
```diff
@@ -0,0 +1,148 @@
...
+  )
+  {
+    final Closer safetyNet = Closer.create();
+    Throwable failure = null;
+
+    // getting the future kicks off any background action, so materialize them all to a list to get things started
+    final List<Future<Optional<Segment>>> futures = new ArrayList<>(acquireSegmentActions.size());
+    for (AcquireSegmentAction acquireSegmentAction : acquireSegmentActions) {
+      safetyNet.register(acquireSegmentAction);
+      // if we haven't failed yet, keep collecing futures (we always want to collect the actions themselves though
```
gianm: collecting (spelling)

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
```diff
@@ -0,0 +1,148 @@
...
+
+
+    final List<SegmentReference> segmentReferences = new ArrayList<>(acquireSegmentActions.size());
+    for (int i = 0; i < acquireSegmentActions.size(); i++) {
+      // if anything fails, want to ignore it initially so we can collect all additional futures to properly clean up
+      // all references before rethrowing the error
+      try {
+        final AcquireSegmentAction action = acquireSegmentActions.get(i);
+        final Future<Optional<Segment>> future = futures.get(i);
+        final Optional<Segment> segment = future.get(timeoutAt - System.currentTimeMillis(), TimeUnit.MILLISECONDS)
```
gianm: It's possible that the `Segment` gets created but this `future.get` throws an error (maybe the `Segment` gets created a split second after the `future.get` times out; maybe the `future.get` is interrupted; etc). In this case the `Segment` won't be registered with the `safetyNet`. Is that bad?

If it is bad, it could be fixed by catching timeout exceptions from `future.get` and, in that case, adding a listener to the `future` that closes the segment if it ever does end up being generated.

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
```diff
@@ -0,0 +1,148 @@
...
+                                                .map(safetyNet::register);
+        segmentReferences.add(
+            new SegmentReference(
+                action.getDescriptor(),
+                segmentMapFunction.apply(segment),
+                action
+            )
+        );
+      }
+      catch (Throwable t) {
```
gianm: Can one of these `Throwable` be an `InterruptedException`? If that's possible, should check for it and re-set the interrupt flag. Same question for the other `catch (Throwable t)` higher up in the method.

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  public SegmentDescriptor getDescriptor()
+  {
+    return segmentDescriptor;
+  }
+
+  public ListenableFuture<Optional<Segment>> getSegmentFuture()
+  {
+    return segmentFutureSupplier.get();
```
gianm, line 81: Seems odd that the supplier isn't memoized, since it means any time someone calls `getSegmentFuture()` on an `AcquireSegmentAction`, this thing will get resubmitted. For a method named `get` I would expect repeated calls to return the same future.

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  @Override
+  public void close() throws IOException
+  {
+    loadCleanup.close();
```
gianm: Any concern about accidental double-close? Could put in something defensive here, if so.

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
```diff
@@ -0,0 +1,148 @@
...
+                segmentMapFunction.apply(segment),
+                action
+            )
+        );
+      }
+      catch (Throwable t) {
+        if (failure == null) {
+          failure = t;
+        } else {
+          failure.addSuppressed(t);
```
gianm: This could end up being quite a lot of exceptions if 100k segments fail to fetch. Limit the number?

server/src/main/java/org/apache/druid/server/ServerManager.java
```diff
@@ -233,23 +287,37 @@ protected <T> FunctionalIterable<QueryRunner<T>> getQueryRunnersForSegments(
       final Closer closer
   )
   {
+    final List<SegmentReference> segmentReferences;
+    // this is kind of gross, but otherwise we're going to be loading weak assignments more or less as soon as they
```
gianm, line 429: Will need to be documented.

server/src/main/java/org/apache/druid/server/ServerManager.java
```diff
@@ -233,23 +287,37 @@ protected <T> FunctionalIterable<QueryRunner<T>> getQueryRunnersForSegments(
...
-                                segment,
-                                factory,
-                                toolChest,
-                                cpuTimeAccumulator,
-                                cacheKeyPrefix
-                            )
-                   ).orElse(
-                       new ReportTimelineMissingSegmentQueryRunner<>(ref.getSegmentDescriptor())
-                   )
+                closer.register(ref)
```
gianm: I think this doesn't register `ref` with `closer` until the query runner for `ref` is actually created. This might not happen immediately, it depends on what `QueryRunnerFactory#mergeRunners` does. Also, if runner creation for some `ref` throws an error for some reason, future ones won't get registered at all. Or if runner creation is intentionally skipped, which I think is possible for `ScanQueryRunnerFactory` with `limit` (IIRC, it will skip creation of runners after reaching `limit`).

Does any of this cause problems?

processing/src/main/java/org/apache/druid/segment/SegmentReference.java
```diff
@@ -0,0 +1,73 @@
...
+
+  public SegmentReference(
+      SegmentDescriptor segmentDescriptor,
+      Optional<Segment> segmentReference,
+      Closeable cleanupHold
+  )
+  {
+    this.segmentDescriptor = segmentDescriptor;
+    this.segmentReference = segmentReference.map(closer::register);
+    closer.register(cleanupHold);
```
gianm, line 55: This will close `cleanupHold` first, then `segmentReference`. Is that the right order? Seems possibly backwards.

server/src/main/java/org/apache/druid/segment/loading/SegmentCacheManager.java
```diff
@@ -39,116 +43,96 @@ public interface SegmentCacheManager
...
-   *
-   * @see SegmentCacheManager#loadSegmentIntoPageCache(DataSegment)
+   * Applies a {@link SegmentMapFunction} to a {@link Segment} if it is available in the cache. If not present in any
+   * storage location, this method will not attempt to download the {@link DataSegment} from deep storage. The
+   * {@link Segment} returned by this method is considered an open reference, cache implementations must not allow it
+   * to be dropped until it has been closed. As such, the returned {@link Segment} must be closed when the caller is
+   * finished doing segment things.
    */
-  void loadSegmentIntoPageCacheOnBootstrap(DataSegment segment);
+  Optional<Segment> acquireSegment(DataSegment dataSegment);
```
gianm: The name of these two methods being the same is confusing, esp. since one may fetch and the other won't. How about calling them `acquireSegmentIfCached` and `acquireSegment`?

Btw, looking at the implementation of `SegmentLocalCacheManager`, it seems like it doesn't quite do what this says. It looks like it returns segments with static references only. It doesn't look like it returns weakly held segments, even if they are currently in the cache. Is that intentional? It doesn't match the javadoc here.

server/src/main/java/org/apache/druid/segment/loading/SegmentCacheManager.java
```diff
@@ -39,116 +43,96 @@ public interface SegmentCacheManager
...
+   * {@link Segment} returned by this method is considered an open reference, cache implementations must not allow it
+   * to be dropped until it has been closed. As such, the returned {@link Segment} must be closed when the caller is
+   * finished doing segment things.
    */
-  void shutdownBootstrap();
-
-  boolean reserve(DataSegment segment);
+  AcquireSegmentAction acquireSegment(
+      DataSegment dataSegment,
+      SegmentDescriptor descriptor
```
gianm: It's kind of weird that this method takes a `SegmentDescriptor`. All it does with it is return it back to the caller in the `AcquireSegmentAction`, so the caller can have it bundled together with the segment future. I suppose it's fine but I wonder if it would make things better to remove it.

server/src/main/java/org/apache/druid/segment/loading/StorageLocation.java
```diff
@@ -47,17 +85,30 @@ public class StorageLocation
...
   /**
-   * Set of files stored under the {@link #path}.
+   * Current total size of files in bytes, including weak entries.
    */
-  @GuardedBy("this")
-  private final Set<File> files = new HashSet<>();
+  private final AtomicLong currSizeBytes = new AtomicLong(0);
+
+  @GuardedBy("lock")
+  private long currWeakSizeBytes = 0;
```
gianm: appears unused? is this for later?

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -209,382 +278,173 @@ public void removeInfoFile(DataSegment segment)
...
-    // Try the already reserved location. If location has been reserved outside, then we do not release the location
-    // here and simply delete any downloaded files. That is, we revert anything we do in this function and nothing else.
-    for (StorageLocation loc : locations) {
-      if (loc.isReserved(segmentDir)) {
-        File storageDir = loc.segmentDirectoryAsFile(segmentDir);
-        boolean success = loadInLocationWithStartMarkerQuietly(loc, segment, storageDir, false);
-        if (!success) {
-          throw new SegmentLoadingException(
-              "Failed to load segment[%s] in reserved location[%s]", segment.getId(), loc.getPath().getAbsolutePath()
+    final Iterator<StorageLocation> iterator = strategy.getLocations();
```
gianm: Is it possible two callers start loading the same segment in two different locations? Presumably `strategy.getLocations()` might give different locations to different callers. And I don't see locking here.

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
```diff
@@ -0,0 +1,148 @@
...
+    this.segmentFutureSupplier = segmentFutureSupplier;
+    this.loadCleanup = loadCleanup;
+  }
+
+  public SegmentDescriptor getDescriptor()
+  {
+    return segmentDescriptor;
+  }
+
+  public ListenableFuture<Optional<Segment>> getSegmentFuture()
```
gianm: Seems that in some cases, calling this method actually triggers the download to start. That isn't totally obvious so it should be called out in javadoc here.

server/src/main/java/org/apache/druid/segment/loading/SegmentCacheEntryIdentifier.java
```diff
@@ -0,0 +1,59 @@
...
+package org.apache.druid.segment.loading;
+
+import org.apache.druid.timeline.SegmentId;
+
+import java.util.Objects;
+
+/**
+ * Use a {@link SegmentId} as a {@link CacheEntryIdentifier}
+ */
+public final class SegmentCacheEntryIdentifier implements CacheEntryIdentifier
```
gianm, line 29: Wondering how partial fetches could work in the future. Will the `CacheEntryIdentifier` become `SegmentId` + projection or `SegmentId` + projection + column?

processing/src/main/java/org/apache/druid/java/util/common/concurrent/Execs.java
clintropolis: I thought this too initially 😅 , but when I dug into it, these pools actually only grow when the queue feeding them is full, preferring to wait in a queue over the overhead of making new threads (see https://docs.oracle.com/en/java/javase/21/docs//api/java.base/java/util/concurrent/ThreadPoolExecutor.html). So a pool with an unbounded queue will only ever have as many threads as the minimum count.

TBH I don't actually know if we need a pool that can grow and shrink, a fixed size pool would probably be fine. ideally I think I would do this all with virtual threads in java 21 - I did some tests and they works pretty well and then we wouldn't need to define a thread pool here at all.

processing/src/main/java/org/apache/druid/segment/SegmentMapFunction.java
clintropolis: right now at the lowest level the input is `ReferenceCountedSegmentProvider.acquireReference()` which spits out a `Optional<Segment>`, so i found it a little easier to compose stuff with segment map functions with it being symmetrical like this. Other stuff like joins can also possibly not return a segment if like they cannot acquire references on their joinables. 

I'm not sure if there would ever be a use case for doing something if the input is missing.

server/src/main/java/org/apache/druid/segment/loading/CacheEntry.java
clintropolis: block, the caller decides to do this in the background or not

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
clintropolis: good call, i'll try to write a test to explore this and make sure we don't leak anything

processing/src/main/java/org/apache/druid/java/util/common/concurrent/Execs.java
gianm: hmm, weird way for cached pools to work, but OK. IMO a fixed size pool is going to be simpler and fine, so I suggest using that. The pool probably won't be big enough that it causes problems.

For virtual threads, were you thinking the processing threads themselves would be virtual, and we'd do "blocking" fetches in those virtual threads? Or did you have something else in mind?

server/src/main/java/org/apache/druid/segment/loading/SegmentCacheManager.java
clintropolis: not sure of a really clean way to get rid of the descriptors, the issue is that the ‘missing segment’ retry stuff works on descriptors (so it can request again with the missing descriptors). the only way i can think to do it is for caller to keep a map of `DataSegment` to  list of `SegmentDescriptor` - i believe it needs to be a list because of wonky overshadowing stuff with mixed segment granularities, so that `AcquireSegmentAction` has a `DataSegment` instead of a descriptor. I'm not sure it feels any better to do this instead, and i think we would want to have some validation checks to make sure that we actually have the same number of missings as descriptors we are reporting (i don't think it would really be possible for only some of the descriptors associated with a segment to be missing at least if we are grabbing them all at once)

gianm: Making it a generic `T` would help, that way it's clearer that it's just being carried around and isn't affecting the logic.

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
clintropolis: added javadoc

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
clintropolis: good catch, i've repurposed the segment locks in `SegmentLocalCacheManager` to be used just for assigning `StorageLocation`, pushing all other locking down to the `StorageLocation`, which simplifies things quite a bit i think

server/src/main/java/org/apache/druid/segment/loading/StorageLocation.java
clintropolis: yea, i have some metrics in mind to emit in the future involving this and the `Stats` class i have added since your last review, but for now is just used in tests

server/src/main/java/org/apache/druid/segment/loading/SegmentCacheManager.java
clintropolis: I ended up just removing `SegmentDescriptor` and tracking it by list position in the caller since it ended up being less complicated than making the `T`

server/src/main/java/org/apache/druid/segment/loading/CacheEntry.java
clintropolis: i dont think there is a use case actually, i think mounting something already mounted in another place should be an error condition, will change this

processing/src/main/java/org/apache/druid/query/metadata/SegmentAnalyzer.java
clintropolis: moved to `VirtualPlaceholderSegment` which just lives in `org.apache.druid.segment.loading` in `druid-server` since its only used by `ServerManager`

processing/src/main/java/org/apache/druid/java/util/common/concurrent/Execs.java
clintropolis: deleted this in favor of always using a fixed pool

re: virtual threads, I guess I was just imagining this thread pool would be replaced with virtual threads, i'm less certain if processing threads in general should be virtual threads since they are expected to be doing lots of work, but it would be interesting to experiment with

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageFabricTest.java
clintropolis: hmm, i felt using something shortened like `vsf` wasn't super clear unless you already know what it is, and i wanted to be clear that if the config is set that the `segmentCache` operates in a completely different mode. 

In the latest changes there is only `isVirtualStorageFabric` and `virtualStorageFabricLoadThreads`, and really `isVirtualStorageFabric` is the only one operators should have to set once we tune out what the proper size is for the thread pool.

I could shorten it to `druid.segmentCache.isVirtualStorage` or `druid.segmentCache.useVirtualStorage` or? idk naming is hard

server/src/main/java/org/apache/druid/segment/loading/SegmentCacheManager.java
clintropolis: renamed to `acquireCachedSegment` and fixed it up so that when pointing to a weakly held reference it will correctly return an `Optional<Segment>` if it is already loaded in teh cache, including using a reservation hold that gets released when closing the segment

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
clintropolis: added defensive check, also relaxed the double close error on the references returned by `ReferenceCountedSegmentProvider` since while I didn't figure out how to trigger it, i think some of these safety closers have the potential to result in the segment references being closed more than once

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
clintropolis: done

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
clintropolis: this logic got moved to `ServerManager`, but i think now catches `InterruptedException`

processing/src/main/java/org/apache/druid/segment/loading/AcquireSegmentAction.java
clintropolis: added some tests to cover timeouts of various sorts which led to fixing some stuff :+1:

processing/src/main/java/org/apache/druid/segment/SegmentReference.java
clintropolis: i don't think it mattered much, but flipped

server/src/main/java/org/apache/druid/server/ServerManager.java
clintropolis: pulled registering with closer out of this to prevent any of ^ from happening

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageFabricTest.java
gianm: I prefer it being `druid.segmentCache.virtualStorage = true`. I would also accept `druid.segmentCache.useVirtualStorage = true`. The `isVirtualStorage` form sounds odd to me. Definitely would drop the `Fabric`.

gianm (approved):
Thanks for the changes. Please consider shortening the config names a bit.

PR #18139, opened 2025-06-12
Update to Kafka 3.9.1.

gianm:
There are a couple of CVE out for 3.9.0.

PR #18178, opened 2025-06-30
Bump kafka client dep to 3.9.1 and mitigate CVE-2025-27817

capistrant:
### Description

Builds on #18139 which is in regards to addressing a [kafka cve](https://www.cve.org/CVERecord?id=CVE-2025-27817)

Ideally we'd jump to kafka 4.x which has a strict deny all default instead of kafka 3.9.1 which defaults to allowAll. Jumping to 4.x would allow us to avoid the static block in our code slipping in the denyAll default. However, we have a lot of test code that relies on being able to run KafkaServer embedded and this is now gone in 4.x. Initially I was looking into making the jump to 4.x anyways. I was going to migrate as much of the tests that rely on KafkaServer to MockConsumer as I could and then disable the rest of the tests in place of relying more on the kafka ITs. However, I then realized that the new Druid Simulations introduced by @kfaraz also rely on kafka 3.x KafkaServer. So jumping to 4 right now would knee-cap that new and really nice test functionality. At this point, I decided to propose the change in this PR. It is an idea suggested by @kgyrtkirk where we can use 3.9.1 and some static blocks on on our side to protect Druid from this CVE. Operators could still follow the Kafka patch nodes to set the system property with a legitimate allow list for the two consumer configs called out under the CVE. This approach buys us time to move to kafka 4 and find long term replacements to using KafkaServer in our tests to run Kafka embedded.

#### Release note

<hr>

 * `KafkaLookupExtractorFactory`
 * `KafkaREcordSupplier`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaRecordSupplier.java
+++ b/extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaRecordSupplier.java
@@ -63,6 +63,17 @@ import java.util.stream.Collectors;
 
 public class KafkaRecordSupplier implements RecordSupplier<KafkaTopicPartition, Long, KafkaRecordEntity>
 {
+  // by default, we reject all URLs for OAuthBearer authentication
+  // CVE ref: https://www.cve.org/CVERecord?id=CVE-2025-27817
+  // Upgrade kafka dependencies to 4.x to remove the need for this static block
+  static {
+    final String allowedSaslOauthbearerUrlsConfig = "org.apache.kafka.sasl.oauthbearer.allowed.urls";
+    String allowedUrlsProp = System.getProperty(allowedSaslOauthbearerUrlsConfig);
+    if (allowedUrlsProp == null) {
+      System.setProperty(allowedSaslOauthbearerUrlsConfig, "notallowed");
+    }
+  }
+
   private final KafkaConsumer<byte[], byte[]> consumer;
   private final KafkaConsumerMonitor monitor;
   private boolean closed;
--- a/extensions-core/kafka-extraction-namespace/src/main/java/org/apache/druid/query/lookup/KafkaLookupExtractorFactory.java
+++ b/extensions-core/kafka-extraction-namespace/src/main/java/org/apache/druid/query/lookup/KafkaLookupExtractorFactory.java
@@ -64,6 +64,17 @@ import java.util.concurrent.atomic.AtomicReference;
 @JsonTypeName("kafka")
 public class KafkaLookupExtractorFactory implements LookupExtractorFactory
 {
+  // by default, we reject all URLs for OAuthBearer authentication
+  // CVE ref: https://www.cve.org/CVERecord?id=CVE-2025-27817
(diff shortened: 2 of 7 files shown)


capistrant:
I'll add a more full description on 6/30, but I wanted to get it up rather than wait.

I need to try this out in a local cluster and see what the UX is like too. Ideally we could get a better DruidException with meaningful messaging in front of the person submitting a Kafka supervisor

capistrant:
Tested in a Druid cluster and confirmed that it appears to work for creating kafka supervisors, but that the UX will suck as is. Overlord logs contain a good message, but it is not making it's way into the error reporting exposed in the console for the supervisor. Will investigate if there is easy way to propagate the root cause up

capistrant:
> Tested in a Druid cluster and confirmed that it appears to work for creating kafka supervisors, but that the UX will suck as is. Overlord logs contain a good message, but it is not making it's way into the error reporting exposed in the console for the supervisor. Will investigate if there is easy way to propagate the root cause up

Idk if there is much we can do: [kafka code](https://github.com/apache/kafka/blob/3.9.1/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerDelegateCreator.java#L56)

This wraps the original Exception that has the good log about why creating the consumer failed due to a disallowed url. The supervisor UI exposes the generic KafkaException under recent errors, but to see what caused it, you need to look in the Overlord log. A pretty poor experience, especially for clusters with operators who make the supervisor API self serve to users who may not know or be able to reference the underlying overlord logs.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaRecordSupplier.java
```diff
@@ -63,6 +63,17 @@
...
+  // by default, we reject all URLs for OAuthBearer authentication
+  // CVE ref: https://www.cve.org/CVERecord?id=CVE-2025-27817
+  // Upgrade kafka dependencies to 4.x to remove the need for this static block
+  static {
+    final String allowedSaslOauthbearerUrlsConfig = "org.apache.kafka.sasl.oauthbearer.allowed.urls";
+    String allowedUrlsProp = System.getProperty(allowedSaslOauthbearerUrlsConfig);
+    if (allowedUrlsProp == null) {
+      System.setProperty(allowedSaslOauthbearerUrlsConfig, "notallowed");
+    }
+  }
```
kfaraz, line 75: I wonder if we should just put this in `AuthorizationUtils` even though it doesn't pertain to core Druid and only extensions, just so we can have it in a common place with adequate javadocs.

capistrant: Do you mean as a static method in AuthorizationUtils and then the static blocks in the two classes who create the KafkaConsumer would call that method?

kfaraz: Yes. Not a blocker for getting this merged though.

capistrant:
Thank you for the review @kfaraz!

Issue #13951, opened 2023-03-20
Discrepancy with null while sorting based on null values between MSQ and native


adarshsanjeev:
While resolving some test failures, I noticed that there is a small discrepancy in the sort order for null values. This only occurs if druid.generic.useDefaultValueForNull is false and there are null values present in the segment queried, with DESC used as the sort order.

For a query like `SELECT l1, COUNT(*) FROM druid.numfoo GROUP BY l1 ORDER BY l1 DESC LIMIT 10`. (taken from the unit tests)
MSQ queries return the following:
```
  325323L, 1L
  7L, 1L
  0L, 1L
  null, 3L
```
Native queries return the following:
```        
  null, 3L
  325323L, 1L
  7L, 1L
  0L, 1L
```
MSQ treats null as a minimum value, while native always seems to return it at the start, irrespective of ASC/DESC. Ideally, we should return the same results regardless of engine. I am not sure, however, which of these should be changed. The MSQ handling seems more appropriate to me(treating it as the smallest key).

gianm:
The ordering the description calls "native order" is really something specific to the `topN` query. The one labeled "MSQ ordering" is more consistent with how other queries like `groupBy` and `windowOperator` (specifically the `naiveSort` operator) behave. In general most null-aware comparison code I've seen considers `null` as "smaller" than any non-null value, so nulls would appear last in descending order and first in ascending order.

The underlying reason for the native/MSQ discrepancy on this SQL query is that MSQ does not include the `topN` query type, so it always uses the "equivalent" `groupBy`. You can get the same behavior out of native by doing `SET useApproximateTopN = FALSE; SET useLexicographicTopN = FALSE;`

Btw, even `topN` isn't really internally consistent. When ordering by the dimension it puts `null` first regardless of whether the query runs in ascending or descending order. But when ordering by the metric, it puts `null` last, also regardless of asc/desc order. (I just learned this today when experimenting.)

Another note- in #18074 planning changed to use `groupBy` rather than `topN` by default for queries like the one in the original description, so that query would start returning a resultset with `null` last.

Given all this, IMO the best fix for this issue would be to adjust `topN` so it behaves like the other query types: `null` should appear first in `ASC` order, last in `DESC` order.

gianm:
Renamed the issue to "Discrepancy with null while sorting based on null values between topN and groupBy", which is a better name.

PR #18181, opened 2025-06-30
Fix ordering for certain floats in row-based frames.

gianm:
Row-based frames had an incorrect sort order for certain float and double values. The problem was an incorrect transformation in TransformUtils from natural form to bits-sortable form. The now-legacy transformation compares the sign and exponents properly for negative numbers, but compares significands backwards, so if two negative floats have the same exponent they sort backwards. It is fixed by flipping all bits for negative numbers, instead of just the sign bit and exponent bits.

The main change is in TransformUtils, where the following methods gain a `FrameType` parameter: `transformFromDouble`, `detransformToDouble`, `transformFromFloat`, and `detransformToFloat`. With frame type `ROW_BASED_V1`, the legacy transformation is used; with `ROW_BASED_V2` the correct transformation is used.

All other changes are to support rolling updates. Because older code and newer code may coexist during a job, a context parameter `rowBasedFrameType` (default V1) is added. It's set to V2 on the Broker, because the Broker is updated after all data servers. By the time the Broker updates, it is safe to use the new type. Users do not need to set this property directly, nor should they.


kgyrtkirk:
I wonder if `ROW_BASED_V2` should the default after say 2 releases ? 
I was thinking about this earlier....there are things which can't be done on-the-spot; but should be (much) later....is there an annotation or some other marking to ensure that will be considered? (this is now kinda out-of-scope)

In the recent days I've bumped into a [similar issue](https://github.com/apache/druid/issues/13951) for which I feel like the fix might be closely related; if a `row_version` is added - it might be great to also cover that in the same version bump...what do you think? I think that could probably be fixed similarily inside `StringFieldReader` / `StringFieldWriter`

gianm:
> I wonder if `ROW_BASED_V2` should the default after say 2 releases ? I was thinking about this earlier....there are things which can't be done on-the-spot; but should be (much) later....is there an annotation or some other marking to ensure that will be considered? (this is now kinda out-of-scope)

It should be the default at some point, although I was thinking more like 4 releases. That would mean the Broker no longer needs to set `rowBasedFrameType`, which would clean things up a bit. However, making this change would mean that rolling upgrade from a version < 34 isn't possible, and I tend to favor longer compatibility windows, since it simplifies upgrading when people don't do it every single release.

There isn't an annotation really to help find this stuff. Could be useful. In the case of this PR, to help with remembering to change things later, I've tried to leave plenty of comments about what the purpose of V1 is, and why the default is currently V1.

> In the recent days I've bumped into a [similar issue](https://github.com/apache/druid/issues/13951) for which I feel like the fix might be closely related; if a `row_version` is added - it might be great to also cover that in the same version bump...what do you think? I think that could probably be fixed similarily inside `StringFieldReader` / `StringFieldWriter`

I just commented on this one; IMO if the discrepancy is to be fixed, the best place is in the `topN` engine (meaning no frame format change would be needed).

kgyrtkirk:
oh..I see now... thank you for the explanation!

clintropolis (approved):

PR #18182, opened 2025-07-01
Do not retry task actions when a lock is revoked

kfaraz:
### Description

When a task lock has been revoked, the corresponding task action should not be retried since the task
revoking the lock is likely to be still in progress. It is better to just fail the task with the revoked locks since
new locks will need to be acquired anyway.

### Changes

- Throw a DruidException of forbidden category if task locks are revoked. The `ServiceClient` does not retry these errors.
- Reduce the default number of retries in the remote task action client to 13 (about 10 minutes worth of retries).

---

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/TaskLocks.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/TaskLocks.java
@@ -59,7 +59,7 @@ public class TaskLocks
   {
     if (!isLockCoversSegments(task, taskLockbox, segments)) {
       throw DruidException.forPersona(DruidException.Persona.OPERATOR)
-                          .ofCategory(DruidException.Category.FORBIDDEN)
+                          .ofCategory(DruidException.Category.CONFLICT)
                           .build(
                               "Segment IDs[%s] are not covered by locks[%s] for task[%s]",
                               segments.stream().map(DataSegment::getId).collect(Collectors.toSet()),


indexing-service/src/test/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertActionTest.java
```diff
@@ -217,10 +218,11 @@ public void test_fail_badVersion() throws Exception
     actionTestKit.getTaskLockbox().add(task);
     acquireTimeChunkLock(TaskLockType.EXCLUSIVE, task, INTERVAL, 5000);
 
-    IllegalStateException exception = Assert.assertThrows(
-        IllegalStateException.class,
+    DruidException exception = Assert.assertThrows(
+        DruidException.class,
         () -> action.perform(task, actionTestKit.getTaskActionToolbox())
     );
+    Assert.assertEquals(DruidException.Category.FORBIDDEN, exception.getCategory());
```
imply-cheddar: Why not use the `DruidExceptionMatcher`?

docs/configuration/index.md
```diff
@@ -1475,7 +1475,7 @@ For most types of tasks, `SegmentWriteOutMediumFactory` can be configured per-ta
 |`druid.indexer.task.storeEmptyColumns`|Boolean value for whether or not to store empty columns during ingestion. When set to true, Druid stores every column specified in the [`dimensionsSpec`](../ingestion/ingestion-spec.md#dimensionsspec). <br/><br/>If you set `storeEmptyColumns` to false, Druid SQL queries referencing empty columns will fail. If you intend to leave `storeEmptyColumns` disabled, you should either ingest placeholder data for empty columns or else not query on empty columns.<br/><br/>You can overwrite this configuration by setting `storeEmptyColumns` in the [task context](../ingestion/tasks.md#context-parameters).|true|
 |`druid.peon.taskActionClient.retry.minWait`|The minimum retry time to communicate with Overlord.|`PT5S`|
 |`druid.peon.taskActionClient.retry.maxWait`|The maximum retry time to communicate with Overlord.|`PT1M`|
-|`druid.peon.taskActionClient.retry.maxRetryCount`|The maximum number of retries to communicate with Overlord.|60|
+|`druid.peon.taskActionClient.retry.maxRetryCount`|The maximum number of retries to communicate with Overlord.|20 (about 10 minutes of retrying)|
```
imply-cheddar: 20 is 17.5 minutes of retries, not 10 minutes.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/TaskLocks.java
```diff
@@ -57,7 +58,7 @@ static void checkLockCoversSegments(
   )
   {
     if (!isLockCoversSegments(task, taskLockbox, segments)) {
-      throw new ISE(
+      throw Forbidden.exception(
```
gianm: This is kind of hacky, since "Forbidden" leads to `403 Forbidden`, which is meant for situations where an action is unauthorized based on security configuration. IMO it would be better to return a different code, like `400 Bad Request` or `409 Conflict`. Those are both considered nonretryable by the standard retry policy.

kfaraz: `403` made some sense to me in that this task is forbidden or not authorized to act on the interval in question since it doesn't hold the required locks.

But I guess it does become ambiguous with a real forbidden error where the __user__ making the request is not authorized to perform some action.

kfaraz: Updated.

docs/configuration/index.md
kfaraz: Thanks for catching this! I was testing out with the default values in `StandardRetryPolicy.unlimited()`.
Fixed this up and added a test for future reference.

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertActionTest.java
kfaraz: Updated.

gianm (approved):

kfaraz:
Thanks a lot for the reviews, @gianm , @imply-cheddar  !

PR #18183, opened 2025-07-01
Use 'trunc' to make certain tests more robust.

gianm:
Some queries involving trigonometric, exponential, or logarithmic functions produce different results on different platforms. This change uses 'trunc' to harmonize results across platforms.


gianm:
> interesting that result order changes in some cases - but its not ordered; so it doesn't matter....

I thought that was odd too, and tracked it down to a feature in Quidem. It sorts resultsets when it believes that the SQL query does not have an `ORDER BY` (I say "believes" because it does not really have a true parser, it's a simple function `Quidem#isProbablyDeterministic`). The idea is that when the SQL query is not explicitly ordered, the underlying database might return rows in any order, so Quidem wants to ignore the order. It uses a simple lexicographic sort that includes whitespace, so whitespace differences can cause the results to end up in a different order.

kgyrtkirk:
oh...really? that again...I've also bumped into that a while back...sorry for the trouble....
it should really be fixed

PR #18142, opened 2025-06-13
Use CoordinatorClient for fetching tier lookup

uds5501:
### Description

This PR is part of the effort to phase out `DruidLeaderClient` and utilize `CoordinatorClient` to interact with Coordinators. This particular PR focuses on 
- Implementing interface `fetchLookupForTier` for Coordinator Client.
- Replacing `DruidLeaderClient` in `LookupReferencesManager` with `CoordinatorClient`.

<hr>

##### Key changed/added classes in this PR
 * `CoordinatorClientImpl.fetchLookupForTier()`

<hr>

This PR has:

- [x] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [x] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [x] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [ ] been tested in a test Druid cluster.

PR #18184, opened 2025-07-01
Use ManageLifecycle for NamespaceExtractionCacheManager implementations

uds5501:
### Description

It's a patch for managing lifecycle of NamespaceExtractionCacheManager via main lifecycle thread.

Note for reviewers - 
I have modified the elements `CacheSchedulerTest` that checked whether the `executors` have terminated or not because we are not explicitly passing in the lifecycle anymore. I couldn't find a test where we would import LifeCycle module in the UT and run this, so modified for now (open to removal as well)

<hr>

 * `NamespaceExtractionCacheManager.java`
 * `OffHeapNamespaceExtractionCacheManager`
 * `OnHeapNamespaceExtractionCacheManager`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/lookups-cached-global/src/main/java/org/apache/druid/server/lookup/namespace/cache/NamespaceExtractionCacheManager.java
+++ b/extensions-core/lookups-cached-global/src/main/java/org/apache/druid/server/lookup/namespace/cache/NamespaceExtractionCacheManager.java
@@ -45,7 +45,7 @@ public abstract class NamespaceExtractionCacheManager
 {
   private static final Logger log = new Logger(NamespaceExtractionCacheManager.class);
 
-  protected final ScheduledThreadPoolExecutor scheduledExecutorService;
+  private final ScheduledThreadPoolExecutor scheduledExecutorService;
   private final ServiceEmitter serviceEmitter;
 
   public NamespaceExtractionCacheManager(
@@ -65,12 +65,12 @@ public abstract class NamespaceExtractionCacheManager
   }
 
   @LifecycleStart
-  public void initExecutor()
+  public void start()
   {
     this.scheduledExecutorService.scheduleAtFixedRate(
         () -> {
           try {
-            monitor(this.serviceEmitter);
+            monitor(serviceEmitter);
           }
           catch (Exception e) {
             log.error(e, "Error emitting namespace stats");
@@ -83,6 +83,11 @@ public abstract class NamespaceExtractionCacheManager
         10,
         TimeUnit.MINUTES
     );
+  }
+
+  public void stopExecutor()
+  {
+    this.scheduledExecutorService.shutdownNow();
   }
 
   /**
--- a/extensions-core/lookups-cached-global/src/main/java/org/apache/druid/server/lookup/namespace/cache/OnHeapNamespaceExtractionCacheManager.java
+++ b/extensions-core/lookups-cached-global/src/main/java/org/apache/druid/server/lookup/namespace/cache/OnHeapNamespaceExtractionCacheManager.java
@@ -174,6 +174,6 @@ public class OnHeapNamespaceExtractionCacheManager extends NamespaceExtractionCa
   @LifecycleStop
   public void stop()
   {
-    this.scheduledExecutorService.shutdownNow();
+    super.stopExecutor();
   }
 }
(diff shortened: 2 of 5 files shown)


extensions-core/lookups-cached-global/src/main/java/org/apache/druid/server/lookup/namespace/cache/OnHeapNamespaceExtractionCacheManager.java
```diff
@@ -42,6 +43,7 @@
 /**
  *
  */
+@ManageLifecycle
```
kfaraz, line 46: Do not use this annotation. Use `ManageLifecycle` scope in `NamespaceExtractionModule` instead.

extensions-core/lookups-cached-global/src/main/java/org/apache/druid/server/lookup/namespace/cache/NamespaceExtractionCacheManager.java
```diff
@@ -41,14 +40,15 @@
  * cacheHandler.close();
  * }</pre>
  */
+
 public abstract class NamespaceExtractionCacheManager
 {
   private static final Logger log = new Logger(NamespaceExtractionCacheManager.class);
 
-  private final ScheduledThreadPoolExecutor scheduledExecutorService;
+  protected final ScheduledThreadPoolExecutor scheduledExecutorService;
```
kfaraz: We shouldn't need to make this protected.
Instead add a `stop()` method to shutdown the executor service.
The sub-classes can call the stop method from their own stop methods.

extensions-core/lookups-cached-global/src/main/java/org/apache/druid/server/lookup/namespace/cache/NamespaceExtractionCacheManager.java
```diff
@@ -61,11 +61,16 @@ public NamespaceExtractionCacheManager(
             .setPriority(Thread.MIN_PRIORITY)
             .build()
     );
-    ExecutorServices.manageLifecycle(lifecycle, scheduledExecutorService);
-    scheduledExecutorService.scheduleAtFixedRate(
+    this.serviceEmitter = serviceEmitter;
+  }
+
+  @LifecycleStart
+  public void initExecutor()
```
kfaraz:
```suggestion
  public void start()
```

extensions-core/lookups-cached-global/src/main/java/org/apache/druid/server/lookup/namespace/cache/NamespaceExtractionCacheManager.java
```diff
@@ -61,11 +61,16 @@ public NamespaceExtractionCacheManager(
...
+  {
+    this.scheduledExecutorService.scheduleAtFixedRate(
         () -> {
           try {
-            monitor(serviceEmitter);
+            monitor(this.serviceEmitter);
```
kfaraz: The `this` is redundant here.

extensions-core/lookups-cached-global/src/main/java/org/apache/druid/server/lookup/namespace/cache/OnHeapNamespaceExtractionCacheManager.java
uds5501: I did that first, apparently these scopes don't work with abstracted classes (a couple of tests failed)

kfaraz: Really? I should think something like this would work:

- Add a `start`, `stop` method in the abstract class. Do not annotate these methods with `@LifecycleStart` / `@LifecycleStop`
- Override these methods in the sub-classes and annotate with `@LifecycleStart`/`@LifecycleStop`
- Call the `super.start()` and `super.stop()` as needed
- Bind the sub-classes like this:
```
PolyBind
        .optionBinder(binder, Key.get(NamespaceExtractionCacheManager.class))
        .addBinding("offHeap")
        .to(OffHeapNamespaceExtractionCacheManager.class)
        .in(ManageLifecycle.class);
```

kfaraz:
Closing this since the original patch #18142  has been reverted.
The correct fix would be to have `CoordinatorClientImpl.fetchLookupsForTier()` make synchronous requests using `ServiceClient.request()` instead of `ServiceClient.asyncRequest()`.

PR #18185, opened 2025-07-01
Web console: Explore nav fixes (also fix npm audit)

vogievetsky:
This PR improves the navigation of the Explore view by giving it a dedicated header bar (which makes sense as the Explore view is an end state applet for the most part).

image

This PR also ensures that colors are consistent between two time charts if they are faceted by the same dimension.

This PR also makes it so there are no "magic icons" for views floating in the code, instead there is a view -> icon lookup.

Finally this PR fixes an issue flagged by `npm audit` that required the update of webpack-dev-server to v5.2.2


cecemei (approved):
Thanks for the improvement!

PR #18186, opened 2025-07-01
Split QTests out into their own job for all jdks GHA ci.yml runs

capistrant:
### Description

The ci.yml workflow runs tests for "all jdks" on `push` events. These UT runs currently do not break QTest out into its own parallelized jobs like the standard JDK 17 GHA runs do. This results in the Q* UTs running very long for jdk 11 and 21. My PR proposes that we break QTest away from the larger UT GHA job and run as a parallel standalone like we do for jdk17. The QTests for non-jdk17 would spin up only if jdk17 Qtest is successful (This is meant to mirror how UTs for jdk11 and 21 are dependent on jdk17 UT).

##### Alternatives

* Don't run full all jdk test suite on every push event.
* Deal with the existing state

#### Release note

<hr>

 * `ci.yml` GHA workflow

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
         pattern: [ "C*", "H*,U*,V*", "N*,Q*,S*", "B*,O*,R*", "G*,J*,K*", "F*,L*,M*", "A*,D*,I*,X*,Y*,Z*", "E*,P*,T*,W*"]
     uses: ./.github/workflows/worker.yml
     with:
-      script: .github/scripts/run_unit-tests -Dtest='${{ matrix.pattern }}' -fae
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -91,3 +91,18 @@ jobs:
       script: .github/scripts/run_unit-tests -Dtest=QTest -Dquidem.split=${{ matrix.split }} -fae
       key: QTest-${{ matrix.split }}
       jdk: 17
+
+  # this will run only if QTest passes for java17
+  run-qtest-all-jdk:
```
kgyrtkirk, line 96: I think this could probably be relaxed and ignored on other jdks; don't think it would find any interesting issues...

I feel like keeping just the changes of `L79` would be fine

capistrant:
replacing with #18187

PR #18191, opened 2025-07-02
Process Aggregate/Limit in decoupled dart planning

kgyrtkirk:
* also adds some test plumbing to run `Calcite*Test` -s against dart
* cases of `CalciteQueryTest` were enabled


extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/stages/SortStage.java
```diff
@@ -62,4 +67,37 @@ public LogicalStage extendWith(DruidNodeStack stack)
     return null;
   }
 
+
+  public static class OffsetLimitStage extends AbstractFrameProcessorStage
```
cryptoe: This can be a separate stage class no ?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/stages/GroupByStages.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class GroupByStages
+{
+  public static class PreShuffleStage extends ProjectStage
+  {
+    private GroupByQuery gby;
```
cryptoe:
```suggestion
    private GroupByQuery groupByQuery;
```
Small nit.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/stages/GroupByStages.java
```diff
@@ -0,0 +1,122 @@
...
+    PostShuffleStage finalAggStage = new PostShuffleStage(sortStage, gby, grouping.getOutputRowSignature());
+    return finalAggStage;
+  }
+
+  private static GroupByQuery makeGbyQuery(ProjectStage projectStage, Grouping grouping)
+  {
+    GroupByQuery.Builder builder = GroupByQuery.builder();
+    builder.setDimensions(grouping.getDimensionSpecs());
+    builder.setQuerySegmentSpec(QuerySegmentSpec.ETERNITY);
+    builder.setGranularity(Granularities.ALL);
```
cryptoe, line 114: This should be fetched from query context no ? https://github.com/apache/druid/blob/36723f1f012a5fa429fb1b60e14433e4a283778b/extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/querykit/QueryKitUtils.java#L81

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/stages/GroupByStages.java
```diff
@@ -0,0 +1,122 @@
...
+    builder.setAggregatorSpecs(grouping.getAggregatorFactories());
+    builder.setDimFilter(projectStage.getDimFilter());
+    builder.setVirtualColumns(projectStage.getVirtualColumns());
+    builder.setPostAggregatorSpecs(grouping.getPostAggregators());
+    builder.setDataSource(new TableDataSource("DUMMY"));
```
cryptoe: Would be helpful to understand why this was named as dummy and not the actual data source.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/stages/GroupByStages.java
kgyrtkirk: insert is not yet supported - and I think when it will be ; it should be different:

I think the presence of those context keys are there due to the fact that insert needed some tweaks to the final writer - this pattern have added some extra things here and there.
I imagine that adding a `TableModify` or something similar top level element describing the high level `insert` details could provide a more natural home for these details.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/stages/GroupByStages.java
kgyrtkirk: for these stages there is no real `DataSource` - the `query` is only used as an api contract object which could haul a lot of things...
`PreShuffleStage` / `PostShuffleStage` doesn't really care about the `DataSource` itself - but to have a valid `GroupByQuery` it must be set to something - otherwise the query can't be built.

Created a static field instead of creating it here - and added a note to it.

kgyrtkirk:
@cryptoe Thank you for reviewing the changes!

PR #18192, opened 2025-07-02
Add coordinator client implementation for LookupReferenceManager

uds5501:
### Description
Uses synchronous http call to fetch lookups from coordinator.

#### Release note

<hr>

##### Key changed/added classes in this PR
 * `CoordinatorClientImpl.java`
 
<hr>

This PR has:

- [x] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [x] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [ ] been tested in a test Druid cluster.

PR #18193, opened 2025-07-03
Add EmbeddedJdbcLookupTest

kfaraz:
### Description

This patch accompanies #18192 and helps verify the change in that PR.

### Changes

- Add `EmbeddedJdbcLookupTest`
- Access the `SQLMetadataConnector` injected into an `EmbeddedDruidServer` in tests

---

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/embedded-tests/pom.xml
+++ b/embedded-tests/pom.xml
@@ -111,6 +111,12 @@
       <version>${project.parent.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.druid.extensions</groupId>
+      <artifactId>druid-lookups-cached-global</artifactId>
+      <version>${project.parent.version}</version>
+      <scope>test</scope>
+    </dependency>
     <!-- Some embedded tests use the router and the web-console for debugging -->
     <dependency>
       <groupId>org.apache.druid</groupId>
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/lookup/EmbeddedJdbcLookupTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/lookup/EmbeddedJdbcLookupTest.java
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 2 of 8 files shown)


uds5501:
LGTM but the tests seems to be failing

kfaraz:
> LGTM but the tests seems to be failing

Thanks, @uds5501 ! Yeah, the ITs are acting up again.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedIndexTaskTest.java
```diff
@@ -102,7 +102,12 @@ public void test_runIndexTask_forInlineDatasource()
       start = start.plusDays(1);
     }
 
-    // Wait for segments to be queryable
+    // Wait for all segments to be loaded and queryable
```
gianm, line 105: I tried fixing this in #18196 with a different approach, where there's a hook into the Broker to look at its server view. I think it will be more robust, because it ensures the Broker actually sees all segments (in theory, with this inspection of Coordinator metrics, they could be loaded on Historicals but the Broker might not know about them yet).

gianm: Discussed in #18196 that the approach there, while perhaps more robust is also too complex and so we won't do it. IMO, the approach you have here is still an improvement so we might as well do it. It would be even better to add a new Broker metric so we can be sure the Broker saw the segments too, but that can be done separately so this PR isn't changing too much stuff at once.

gianm (approved):

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedIndexTaskTest.java
kfaraz: @gianm , should we proceed with this for now until we add the new Broker monitor for emitting segment count per datasource?

kfaraz: Sounds good, thanks!

PR #18195, opened 2025-07-03
[WIP] Coordinator client impl follow up

uds5501:
Fixes #XXXX.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.


server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
```diff
@@ -112,4 +115,27 @@ public interface CoordinatorClient
...
+   * API: {@code GET /druid/coordinator/v1/rules}
+   */
+  Map<String, List<Rule>> getRulesSync();
+
+  /**
+   * Returns the current leader's host and port.
+   * <p>
+   * API: {@code GET /druid/coordinator/v1/leader}
+   */
+  String findCurrentLeader();
```
kfaraz: The new APIs should all be async.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
```diff
@@ -112,4 +115,27 @@ public interface CoordinatorClient
...
   Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(String tier);
+
+  /**
+   * Returns an iterator over the metadata segments in the cluster.
+   * <p>
+   * API: {@code GET /druid/coordinator/v1/metadata/segments?includeOvershadowedStatus&includeRealtimeSegments}
+   *
+   * @param watchedDataSources Optional datasources to filter the segments by. If null or empty, all segments are returned.
+   */
+  JsonParserIterator<SegmentStatusInCluster> getMetadataSegmentsSync(@Nullable Set<String> watchedDataSources);
```
kfaraz: Return a `ListenableFuture<CloseableIterator<SegmentStatusInCluster>>` here.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -276,6 +290,99 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
+  @Override
+  public JsonParserIterator<SegmentStatusInCluster> getMetadataSegmentsSync(
+      @Nullable Set<String> watchedDataSources
+  )
+  {
+    final StringBuilder pathBuilder = new StringBuilder(
+        "/druid/coordinator/v1/metadata/segments?includeOvershadowedStatus&includeRealtimeSegments");
+    if (watchedDataSources != null && !watchedDataSources.isEmpty()) {
+      for (String dataSource : watchedDataSources) {
+        pathBuilder.append("&dataSource=").append(StringUtils.urlEncode(dataSource));
```
kfaraz: I think the parameter name is `datasources`.

Please add an `EmbeddedCoordinatorClientTest` similar to `EmbeddedOverlordClientTest` so that we can be certain of the correctness of the new methods.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -276,6 +290,99 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
+
+    try {
+      String query = pathBuilder.toString();
+      InputStreamFullResponseHolder responseHolder = client.request(
+          new RequestBuilder(HttpMethod.GET, query),
+          new InputStreamFullResponseHandler()
+      );
+
+      if (responseHolder.getStatus().getCode() != HttpServletResponse.SC_OK) {
+        throw new RE(
```
kfaraz: Why is this needed? Does some calling code or test rely on this specific exception message?

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -276,6 +290,99 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
+          Futures.immediateFuture(responseHolder.getContent()),
+          null,
+          null,
+          null,
+          jsonMapper
+      );
+    }
+    catch (InterruptedException | ExecutionException e) {
+      throw new RuntimeException(e);
+    }
```
kfaraz: Check out the method `OverlordClientImpl.taskStatuses` and see if you can do something similar here to make this code more concise.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -276,6 +290,99 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
+
+    throw new ISE(
+        "Couldn't find leadear, failed response status is[%s], and content[%s]",
+        responseHolder.getStatus().getCode(),
+        responseHolder.getContent()
+    );
```
kfaraz: This seems verbose. Ideally, we should just need this (similar to `OverlordClientImpl.findCurrentLeader()`):

```suggestion
return FutureUtils.transform(
        client.asyncRequest(
            new RequestBuilder(HttpMethod.GET, "/druid/coordinator/v1/leader"),
            new StringFullResponseHandler(StandardCharsets.UTF_8)
        ),
        holder -> {
          try {
            return new URI(holder.getContent());
          }
          catch (URISyntaxException e) {
            throw new RuntimeException(e);
          }
        }
    );
```

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
uds5501: Not required anymore.

> I added it in the first place to emulate the lack of http exception thrown in non 200 responses.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
```diff
@@ -112,4 +116,43 @@ public interface CoordinatorClient
...
+   * API: {@code GET /druid-ext/basic-security/authorization/db/<prefix>/cachedSerializedUserMap}
+   */
+  byte[] getCachedSerializedUserMapSync(String prefix);
+
+  /**
+   * Returns the serialized group mapping and role map for the given prefix stored in coordinator.
+   * <p>
+   * API: {@code GET /druid-ext/basic-security/authorization/db/<prefix>/cachedSerializedGroupMappingMap}
+   */
+  byte[] getCachedSerializedGroupMappingMapSync(String prefix);
```
uds5501: Note for reviewers:

1. These methods are in sync to avoid the lifecycle deadlock that occured in the last [faulty PR](https://github.com/apache/druid/pull/18142). The pattern is similar as `UserAndRoleMap` is a class present in the extension, that will be loaded by the main thread. If the same thread loads the extension and attempts parsing the `UserAndRoleMap`, that won't cause any issue.
2. On a similar note, I am opting to send to `byte[]` to the call point because sending `UserAndRoleMap` was causing a circular dependency (the class is defined in `druid-basic-security` extension). Do I change move the `UserAndRoleMap` and relevant classes to `druid-server` module instead? Need suggestions.

abhishekrb19: On the second point, it seems weird to have these APIs return `byte[]` - it's error prone for callers as they have to understand the decoding semantics.

> I am opting to send to byte[] to the call point because sending UserAndRoleMap was causing a circular dependency (the class is defined in druid-basic-security extension). Do I change move the UserAndRoleMap and relevant classes to druid-server module instead?

IMO it's fine to move the necessary classes from the security extension to the `server` module for reuse and have the APIs return the corresponding concrete POJO classes. It can be somewhere in or close to `org.apache.druid.server.security`

kfaraz: +1, returning the `byte[]` doesn't seem appropriate.
We can consider moving the necessary classes, @uds5501 .

kfaraz: @uds5501 , it does seem that the these methods would need to be called on startup by the main thread.
So let's keep these methods as synchronous for now as you suggest.
We can revisit them later.

uds5501: Have moved the appropriate classes ✅

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
uds5501: I have added the embedded tests for the endpoints touched in this PR and a few more.
I am kind of stuck handling `fetchDataSourceInformation` as I am not able to run the coordinator with appropriate setting required (it's throwing a 404 error and for some reason the `CoordinatorSegmentMetadataCache` is not being initialized (added the cluster in PR), will need help with that.

uds5501:
@kfaraz I have opened this PR for a round of review

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -180,7 +185,9 @@ public static Map<String, Object> deserializeJsonToMap(String payload)
     try {
       return TestHelper.JSON_MAPPER.readValue(
           payload,
-          new TypeReference<>() {}
+          new TypeReference<>()
+          {
+          }
```
kfaraz: Please revert this.
We probably need to update the checkstyle rules / IDEA codestyle to avoid auto-updating this.

sql/src/test/java/org/apache/druid/sql/calcite/util/CalciteTests.java
```diff
@@ -141,7 +143,11 @@ public class CalciteTests
 
   public static final String TEST_SUPERUSER_NAME = "testSuperuser";
   public static final Policy POLICY_NO_RESTRICTION_SUPERUSER = NoRestrictionPolicy.instance();
-  public static final Policy POLICY_RESTRICTION = RowFilterPolicy.from(BaseCalciteQueryTest.equality("m1", 6, ColumnType.LONG));
+  public static final Policy POLICY_RESTRICTION = RowFilterPolicy.from(BaseCalciteQueryTest.equality(
+      "m1",
+      6,
+      ColumnType.LONG
+  ));
```
kfaraz: Easier to read like this:
```suggestion
  public static final Policy POLICY_RESTRICTION = RowFilterPolicy.from(
      BaseCalciteQueryTest.equality("m1", 6, ColumnType.LONG)
  );
```

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
```diff
@@ -119,4 +123,36 @@ public interface CoordinatorClient
...
   Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(String tier);
+
+  /**
+   * Returns an iterator over the metadata segments in the cluster.
+   * <p>
+   * API: {@code GET /druid/coordinator/v1/metadata/segments?includeOvershadowedStatus&includeRealtimeSegments}
+   *
+   * @param watchedDataSources Optional datasources to filter the segments by. If null or empty, all segments are returned.
+   */
+  ListenableFuture<CloseableIterator<SegmentStatusInCluster>> getMetadataSegments(
```
kfaraz:
```suggestion
  ListenableFuture<CloseableIterator<SegmentStatusInCluster>> getAllUsedSegments(
```

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
```diff
@@ -119,4 +123,36 @@ public interface CoordinatorClient
    * @param tier The name of the tier for which the lookup configuration is to be fetched.
    */
   Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(String tier);
+
+  /**
+   * Returns an iterator over the metadata segments in the cluster.
+   * <p>
+   * API: {@code GET /druid/coordinator/v1/metadata/segments?includeOvershadowedStatus&includeRealtimeSegments}
```
kfaraz: We should probably keep `includeOvershadowedSegments` and `includeRealtimeSegments` as arguments of this method, even though their current usage always sets them to `true`.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
```diff
@@ -119,4 +123,36 @@ public interface CoordinatorClient
...
+  ListenableFuture<CloseableIterator<SegmentStatusInCluster>> getMetadataSegments(
+      @Nullable Set<String> watchedDataSources
+  );
+
+  /**
+   * Returns the current snapshot of the rules.
+   * <p>
+   * API: {@code GET /druid/coordinator/v1/rules}
+   */
+  ListenableFuture<Map<String, List<Rule>>> getRules();
```
kfaraz:
```suggestion
  ListenableFuture<Map<String, List<Rule>>> getRulesForAllDatasources();
```

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
```diff
@@ -119,4 +123,36 @@ public interface CoordinatorClient
...
+   * API: {@code GET /druid/coordinator/v1/leader}
+   */
+  ListenableFuture<URI> findCurrentLeader();
+
+  /**
+   * Posts load rules to the coordinator.
+   * <p>
+   * API: {@code POST /druid/coordinator/v1/rules}
+   */
+  ListenableFuture<Void> postLoadRules(String dataSource, List<Rule> rules);
```
kfaraz:
```suggestion
  ListenableFuture<Void> updateRulesForDatasource(String dataSource, List<Rule> rules);
```

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -214,7 +221,7 @@ public ListenableFuture<CompactionStatusResponse> getCompactionSnapshots(@Nullab
   {
     final StringBuilder pathBuilder = new StringBuilder("/druid/coordinator/v1/compaction/status");
     if (dataSource != null && !dataSource.isEmpty()) {
-      pathBuilder.append("?").append("dataSource=").append(StringUtils.urlEncode(dataSource));
+      pathBuilder.append("?").append("datasources=").append(StringUtils.urlEncode(dataSource));
```
kfaraz: This seems like a typo, we need to change the param name in the new API, not this one.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -287,6 +294,76 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
+  @Override
+  public ListenableFuture<CloseableIterator<SegmentStatusInCluster>> getMetadataSegments(
+      @Nullable Set<String> watchedDataSources
+  )
+  {
+    final StringBuilder pathBuilder = new StringBuilder(
+        "/druid/coordinator/v1/metadata/segments?includeOvershadowedStatus&includeRealtimeSegments");
+    if (watchedDataSources != null && !watchedDataSources.isEmpty()) {
+      for (String dataSource : watchedDataSources) {
+        pathBuilder.append("&dataSource=").append(StringUtils.urlEncode(dataSource));
```
kfaraz: Please fix the query param name.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -287,6 +294,76 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
+
+  @Override
+  public ListenableFuture<Map<String, List<Rule>>> getRules()
+  {
+    final String path = "/druid/coordinator/v1/rules";
+    return FutureUtils.transform(
+        client.asyncRequest(
+            new RequestBuilder(HttpMethod.GET, path),
+            new BytesFullResponseHandler()
+        ), holder -> JacksonUtils.readValue(jsonMapper, holder.getContent(), new TypeReference<>() {})
```
kfaraz:
```suggestion
        ),
        holder -> JacksonUtils.readValue(jsonMapper, holder.getContent(), new TypeReference<>() {})
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedCoordinatorClientTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class EmbeddedCoordinatorClientTest extends EmbeddedClusterTestBase
```
kfaraz, line 56: Thanks for adding this!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedCoordinatorClientTest.java
```diff
@@ -0,0 +1,238 @@
...
+  private final EmbeddedIndexer indexer = new EmbeddedIndexer();
+  private final EmbeddedOverlord overlord = new EmbeddedOverlord();
+  private final EmbeddedHistorical historical = new EmbeddedHistorical();
+
+
+  @Override
+  protected EmbeddedDruidCluster createCluster()
+  {
+    indexer.addProperty("druid.segment.handoff.pollDuration", "PT0.1s")
+           .addProperty("druid.worker.capacity", "25");
```
kfaraz: If we are not running more than 1 task concurrently, this can be removed.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedCoordinatorClientTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+    overlord.addProperty("druid.manager.segments.pollDuration", "PT0.1s");
```
kfaraz: Not needed.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedCoordinatorClientTest.java
```diff
@@ -0,0 +1,238 @@
...
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource),
+        agg -> agg.hasSumAtLeast(1)
+    );
+
+    final List<DataSegment> segments = new ArrayList<>(
+        overlord.bindings().segmentsMetadataStorage().retrieveAllUsedSegments(dataSource, null)
+    );
+    List<Interval> intervals = segments.stream()
+                                       .map(DataSegment::getInterval)
+                                       .collect(Collectors.toList());
```
kfaraz: We can skip this. Just use `Intervals.ETERNITY` to fetch the used segments.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedCoordinatorClientTest.java
```diff
@@ -0,0 +1,238 @@
...
+  @Test
+  @Timeout(20)
+  public void test_loadRules()
+  {
+    Rule broadcastRule = new ForeverBroadcastDistributionRule();
+    cluster.callApi().onLeaderCoordinator(
+        c -> c.postLoadRules(dataSource, List.of(broadcastRule))
+    );
+    Map<String, List<Rule>> rules = cluster.callApi().onLeaderCoordinator(
+        c -> c.getRules()
```
kfaraz: Nit: lambda can be simplified.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedCoordinatorClientTest.java
```diff
@@ -0,0 +1,238 @@
...
+  {
+    indexer.addProperty("druid.segment.handoff.pollDuration", "PT0.1s")
+           .addProperty("druid.worker.capacity", "25");
+
+    overlord.addProperty("druid.manager.segments.pollDuration", "PT0.1s");
+    coordinator.addProperty("druid.coordinator.period.metadataStoreManagementPeriod", "PT0.1S")
+               .addProperty("druid.coordinator.period.indexingPeriod", "PT0.1S")
+               .addProperty("druid.coordinator.segmentMetadataCache.enabled", "true")
+               .addProperty("druid.coordinator.segmentMetadataCache.awaitInitializationOnStart", "true")
+               .addProperty("druid.coordinator.segmentMetadataCache.metadataRefreshPeriod", "PT0.1S");
```
kfaraz: Are all of these properties needed in this test?

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
```diff
@@ -428,7 +450,7 @@ public void test_getCompactionSnapshots_nonNullDataSource() throws Exception
         AutoCompactionSnapshot.builder("ds1").build()
     );
     serviceClient.expectAndRespond(
-        new RequestBuilder(HttpMethod.GET, "/druid/coordinator/v1/compaction/status?dataSource=ds1"),
+        new RequestBuilder(HttpMethod.GET, "/druid/coordinator/v1/compaction/status?datasources=ds1"),
```
kfaraz: We shouldn't be changing an existing API or its tests.

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
```diff
@@ -531,4 +553,208 @@ public void test_fetchLookupsForTierSync_detailedEnabled() throws Exception
...
+                      .map(SegmentStatusInCluster::getDataSegment)
+                      .collect(ImmutableList.toImmutableList())
+    );
+  }
+
+  @Test
+  public void test_getRules() throws Exception
+  {
+    final Map<String, List<Rule>> rules = ImmutableMap.of(
+        "xyz", ImmutableList.of(
```
kfaraz:
```suggestion
        "xyz",
        List.of(
```

services/src/main/java/org/apache/druid/server/router/CoordinatorRuleManager.java
```diff
@@ -132,24 +123,24 @@ public boolean isStarted()
   public void poll()
   {
     try {
-      StringFullResponseHolder response = druidLeaderClient.go(
-          druidLeaderClient.makeRequest(HttpMethod.GET, RulesResource.RULES_ENDPOINT)
```
kfaraz, line 136: We can also remove the constant from `RulesResource` now.

services/src/test/java/org/apache/druid/server/router/CoordinatorRuleManagerTest.java
```diff
@@ -85,11 +85,36 @@ public void testAddingToRulesListThrowingError()
...
+        )
+    );
+    EasyMock.replay(client);
+
+    final CoordinatorRuleManager manager = new CoordinatorRuleManager(
+        () -> tieredBrokerConfig,
+        client
+    );
+
+    expectedException.expect(ISE.class);
```
kfaraz: Use `Assert.assertThrows` instead.

services/src/test/java/org/apache/druid/server/router/CoordinatorRuleManagerTest.java
```diff
@@ -130,20 +154,10 @@ private DruidLeaderClient mockClient()
...
-      EasyMock.replay(holder, client);
-      return client;
-    }
-    catch (IOException | InterruptedException e) {
-      throw new RuntimeException(e);
-    }
+    final CoordinatorClient client = EasyMock.niceMock(CoordinatorClient.class);
+    EasyMock.expect(client.getRules())
+            .andReturn(Futures.immediateFuture(rules));
+    EasyMock.replay(client);
```
kfaraz: Use a `NoopCoordinatorClient` instead.

sql/src/test/java/org/apache/druid/sql/calcite/schema/DruidCalciteSchemaModuleTest.java
```diff
@@ -121,6 +124,9 @@ public void setUp()
           binder.bind(FilteredServerInventoryView.class).toInstance(serverInventoryView);
           binder.bind(SegmentManager.class).toInstance(segmentManager);
           binder.bind(DruidOperatorTable.class).toInstance(druidOperatorTable);
+          binder.bind(CoordinatorClient.class)
+                .annotatedWith(Coordinator.class)
```
kfaraz: I don't think `CoordinatorClient` is ever used with an annotation.

sql/src/test/java/org/apache/druid/sql/calcite/schema/DruidCalciteSchemaModuleTest.java
```diff
@@ -89,6 +90,8 @@ public class DruidCalciteSchemaModuleTest extends CalciteTestBase
   @Mock
   private DruidLeaderClient coordinatorDruidLeaderClient;
   @Mock
+  private CoordinatorClientImpl coordinatorClient;
```
kfaraz: Do not use the `*Impl` class here.

sql/src/test/java/org/apache/druid/sql/calcite/util/CalciteTests.java
```diff
@@ -414,17 +424,20 @@ public static SystemSchema createMockSystemSchema(
...
     final DruidNode overlordNode = new DruidNode("test-overlord", "dummy", false, 8090, null, true, false);
 
-    final DruidLeaderClient druidLeaderClient = new DruidLeaderClient(
-        new FakeHttpClient(),
-        provider,
-        NodeRole.COORDINATOR,
-        "/simple/leader"
+    final CoordinatorClient coordinatorClient = new CoordinatorClientImpl(
+        new MockServiceClient(),
+        getJsonMapper()
```
kfaraz: We could also just use a `NoopCoordinatorClient`.

kfaraz (commented):
Overall looks good, left some suggestions.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
uds5501: Isn't this just returning status of the segments? Should it be called `used Segments`?

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
uds5501: 🤦🏿  yeah my bad, updating this

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedCoordinatorClientTest.java
uds5501: I was attempting to enable segmentMetadataCache and write a test for `fetchDataSourceInformation`, skipped writing that for now, will remove these.

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
uds5501: my bad.

sql/src/test/java/org/apache/druid/sql/calcite/schema/DruidCalciteSchemaModuleTest.java
uds5501: fair, it's not used anymore

sql/src/test/java/org/apache/druid/sql/calcite/schema/DruidCalciteSchemaModuleTest.java
uds5501: removed it altogether, not needed

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
uds5501: Sure, we could do that

uds5501:
Thanks for the review @kfaraz , I have pushed the updates, kindly feel free to go through it whenever you get a chance

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
kfaraz: If you look at the implementation of this API on the server-side, the method is called `MetadataResource.getAllUsedSegments` since it returns info of used segments only.

uds5501: Okay will update

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -287,6 +294,93 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
 
+  @Override
+  public ListenableFuture<CloseableIterator<SegmentStatusInCluster>> getAllUsedSegments(
+      @Nullable Set<String> watchedDataSources,
+      Boolean includeOvershadowedStatus,
+      Boolean includeRealtimeSegments
+  )
+  {
+    final StringBuilder pathBuilder = new StringBuilder(
+        "/druid/coordinator/v1/metadata/segments");
```
kfaraz:
```suggestion
    final StringBuilder pathBuilder = new StringBuilder("/druid/coordinator/v1/metadata/segments");
```

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -287,6 +294,93 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
+
+    List<String> params = new ArrayList<>();
```
kfaraz:
```suggestion
    final List<String> params = new ArrayList<>();
```

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -287,6 +294,93 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
+    if (includeOvershadowedStatus != null && includeOvershadowedStatus) {
```
kfaraz: Once we start accepting primitive `boolean` instead of the boxed `Boolean`
```suggestion
    if (includeOvershadowedStatus) {
```

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -287,6 +294,93 @@ public Map<String, LookupExtractorFactoryContainer> fetchLookupsForTierSync(Stri
...
+      params.add("includeOvershadowedStatus");
+    }
+    if (includeRealtimeSegments != null && includeRealtimeSegments) {
+      params.add("includeRealtimeSegments");
+    }
+    if (watchedDataSources != null && !watchedDataSources.isEmpty()) {
+      StringBuilder datasourceParam = new StringBuilder("datasources=");
+      for (String dataSource : watchedDataSources) {
+        datasourceParam.append(StringUtils.urlEncode(dataSource));
+      }
```
kfaraz: I don't think this is correct. If we had multiple watched datasources, this would give us:

```
datasources=ds1ds2ds3
```
when in fact we want
```
datasources=ds1&datasources=ds2&datasources=ds3
```

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
```diff
@@ -531,4 +553,300 @@ public void test_fetchLookupsForTierSync_detailedEnabled() throws Exception
         coordinatorClient.fetchLookupsForTierSync("default_tier")
     );
   }
+
+  @Test
+  public void test_getAllUsedSegments() throws JsonProcessingException
```
kfaraz:
```suggestion
  public void test_getAllUsedSegments_includeOvershadowedAndRealtime() throws JsonProcessingException
```

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
```diff
@@ -531,4 +553,300 @@ public void test_fetchLookupsForTierSync_detailedEnabled() throws Exception
...
+  }
+
+  @Test
+  public void test_getRules_ForAllDatasources_HttpException_throwsError()
+  {
+    serviceClient.expectAndThrow(
+        new RequestBuilder(HttpMethod.GET, "/druid/coordinator/v1/rules"),
+        new HttpResponseException(
+            new StringFullResponseHolder(
+                new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.NOT_FOUND),
```
kfaraz: I don't think this test case is needed since this API will never throw a 404.

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
```diff
@@ -531,4 +553,300 @@ public void test_fetchLookupsForTierSync_detailedEnabled() throws Exception
...
+    Assert.assertEquals(
+        segments,
+        actualSegments.stream()
+                      .map(SegmentStatusInCluster::getDataSegment)
+                      .collect(ImmutableList.toImmutableList())
+    );
+  }
+
+  @Test
+  public void test_getAllUsedSegments_noParams() throws JsonProcessingException
```
kfaraz: We should have a test case for multiple datasources as well.

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
```diff
@@ -531,4 +553,300 @@ public void test_fetchLookupsForTierSync_detailedEnabled() throws Exception
...
+      FutureUtils.getUnchecked(coordinatorClient.findCurrentLeader(), true);
+    }
+    catch (Exception e) {
+      Throwable throwable = Throwables.getRootCause(e);
+      Assert.assertTrue(throwable instanceof HttpResponseException);
+    }
+  }
+
+  @Test
+  public void test_postLoadRules() throws Exception
```
kfaraz:
```suggestion
  public void test_updateRulesForDatasource() throws Exception
```

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -180,7 +185,8 @@ public static Map<String, Object> deserializeJsonToMap(String payload)
     try {
       return TestHelper.JSON_MAPPER.readValue(
           payload,
-          new TypeReference<>() {}
+          new TypeReference<>()
+          {}
```
kfaraz: 😄 
```suggestion
          new TypeReference<>() {}
```

services/src/test/java/org/apache/druid/server/router/CoordinatorRuleManagerTest.java
```diff
@@ -85,11 +87,35 @@ public void testAddingToRulesListThrowingError()
     rules.get(DATASOURCE1).add(new ForeverDropRule());
   }
 
+  @Test
+  public void testThrowingExceptionOnHTTPException()
```
kfaraz:
```suggestion
  public void test_poll_throwsException_ifCoordinatorApiReturnsNotOk()
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedCoordinatorClientTest.java
```diff
@@ -0,0 +1,225 @@
...
+  private final EmbeddedIndexer indexer = new EmbeddedIndexer();
+  private final EmbeddedOverlord overlord = new EmbeddedOverlord();
+  private final EmbeddedHistorical historical = new EmbeddedHistorical();
+
+
+  @Override
+  protected EmbeddedDruidCluster createCluster()
+  {
+    indexer.addProperty("druid.segment.handoff.pollDuration", "PT0.1s");
+    overlord.addProperty("druid.manager.segments.pollDuration", "PT0.1s");
```
kfaraz: This shouldn't be needed either.

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
```diff
@@ -119,4 +123,40 @@ public interface CoordinatorClient
...
+  /**
+   * Returns an iterator over the metadata segments in the cluster.
+   * <p>
+   * API: {@code GET /druid/coordinator/v1/metadata/segments}
+   *
+   * @param watchedDataSources Optional datasources to filter the segments by. If null or empty, all segments are returned.
+   * @param includeOvershadowedStatus If true, includes the overshadowed status of each segment.
+   * @param includeRealtimeSegments If true, includes realtime segments in the result.
+   */
+  ListenableFuture<CloseableIterator<SegmentStatusInCluster>> getAllUsedSegments(
```
kfaraz: @uds5501 , I didn't realize that there is another method already exposed by this interface `fetchUsedSegment`.
But it is used to fetch the segments of a single datasource.

We should do a couple of things here:
- Add a short javadoc to this new method to indicate that it can fetch segments for multiple datasources in one go.
- Also, it turns out that `includeOvershadowedStatus` always has to be true. If it is not true, the server returns a response of a different type which will throw a parsing error on the client. So, let's get rid of the `includeOvershadowedStatus` argument here.
- Rename the method to `fetchAllUsedSegmentsWithOvershadowedStatus` to indicate this fact.
- The `includeRealtimeSegments` argument should be a primitive `boolean`, not a boxed `Boolean`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedCoordinatorClientTest.java
```diff
@@ -0,0 +1,225 @@
...
+    Rule broadcastRule = new ForeverBroadcastDistributionRule();
+    cluster.callApi().onLeaderCoordinator(
+        c -> c.updateRulesForDatasource(dataSource, List.of(broadcastRule))
+    );
+    Map<String, List<Rule>> rules = cluster.callApi().onLeaderCoordinator(CoordinatorClient::getRulesForAllDatasources);
+    Assertions.assertTrue(!rules.isEmpty());
+    Assertions.assertEquals(List.of(broadcastRule), rules.get(dataSource));
+  }
+
+  private void batchIngest()
```
kfaraz:
```suggestion
  private void runIndexTask()
```

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
uds5501: :((

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
uds5501: something's weird with rename, this should've been renamed automatically. will recheck my settings

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
uds5501: gotcha

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
uds5501: added

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClient.java
uds5501: updated this.

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
```diff
@@ -617,51 +617,50 @@ public void test_getAllUsedSegments_noParams() throws JsonProcessingException
...
         ),
         HttpResponseStatus.OK,
         ImmutableMap.of(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON),
-        jsonMapper.writeValueAsBytes(segments)
+        jsonMapper.writeValueAsBytes(ImmutableList.of(SEGMENT3))
     );
 
     CloseableIterator<SegmentStatusInCluster> iterator = FutureUtils.getUnchecked(
-        coordinatorClient.getAllUsedSegments(null, true, false),
+        coordinatorClient.fetchAllUsedSegmentsWithOvershadowedStatus(Collections.singleton("abc"), true),
```
kfaraz: Nit: Use `Set.of()`

server/src/test/java/org/apache/druid/client/coordinator/CoordinatorClientImplTest.java
```diff
@@ -617,51 +617,50 @@ public void test_getAllUsedSegments_noParams() throws JsonProcessingException
...
         true
     );
+
     List<SegmentStatusInCluster> actualSegments = new ArrayList<>();
     while (iterator.hasNext()) {
       actualSegments.add(iterator.next());
     }
     Assert.assertEquals(
-        segments,
+        ImmutableList.of(SEGMENT3),
```
kfaraz: Nit: Use `List.of()`.

kfaraz (approved):
LGTM 🚀 

+1 after CI passes.

PR #18196, opened 2025-07-03
Embedded tests: stricter segment waiting, longer startup timeouts.

gianm:
Two things that make the embedded tests less flaky:

1) Add a "waitForSegmentAvailability" method to EmbeddedBroker, so tests
   can be sure all desired segments are visible. Prior to this change,
   sometimes I would see failures on EmbeddedIndexTaskTest due to not all
   segments being available when the test query runs.

2) Extend the service startup timeout from 10s to 30s. On my machine,
   when it's busy, sometimes it takes more than 10s to start up a service.

### DIFF
--- a/services/src/test/java/org/apache/druid/testing/embedded/EmbeddedBroker.java
+++ b/services/src/test/java/org/apache/druid/testing/embedded/EmbeddedBroker.java
@@ -36,6 +36,7 @@ import org.apache.druid.timeline.SegmentId;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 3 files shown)


services/src/test/java/org/apache/druid/testing/embedded/EmbeddedBroker.java
```diff
@@ -54,6 +79,9 @@ private Broker(LifecycleInitHandler handler)
     @Override
     public Lifecycle initLifecycle(Injector injector)
     {
+      final BrokerServerView serverView = injector.getInstance(BrokerServerView.class);
```
kfaraz, line 83: I had initially done something similar (i.e. registering a listener) for tracking completion of tasks too.
But eventually decided against it as it seemed much cleaner to just wait for the appropriate metric (`task/run/time` in that case).
Waiting for an actual metric event seemed like the best way to wait for an event to happen in the cluster.

I wonder if we shouldn't do something similar here.
We could emit a count metric from Broker for all the datasource + version that it is currently aware of.
While improving the test, it also improves visibility into the actual product.

I also feel that adding listeners in the embedded servers will gradually bloat this class up as we try to add more test cases.

Let me know what you think.

gianm: I think you're right, it's better to add a Broker metric to keep this file simpler. The metric could be a monitor that emits the number of segments available per datasource.

gianm: Based on this thread, I'll close this PR. It also has the 10s -> 30s timeout thing, but I'll raise that again separately if I run into it again.

PR #6742, opened 2018-12-16
Broker: Await initialization before finishing startup.

gianm:
In particular, hold off on announcing the service and starting the
HTTP server until the server view and SQL metadata cache are finished
initializing. This closes a window of time where a Broker could return
partial results shortly after startup.

I decided to do it this way because there are lots of ways people find
Brokers to query, including:

1. Through the Router (which uses the annoucements)
2. Looking at announcements and querying one of the announced servers
3. Through HTTP load balancers hooked up directly to brokers

Delaying startup should play well with all three mechanisms.

As part of this, some simplification of server-lifecycle service
announcements. This helps ensure that the two different kinds of
announcements we do (legacy and new-style) stay in sync.

PR #18200, opened 2025-07-03
Include empty servers in BrokerServerView.

### LINKED ISSUE #18199: Dart: Cannot use Historical with no segments as worker
Historicals with no segments are not detected by `BrokerServerView`, so they are not able to be used as Dart workers. In particular, this means it is not possible to run queries on clusters with no historical data (such as a totally fresh cluster, or a cluster with only realtime data). To fix this we can add such empty Historicals into the server view.

gianm:
Fixes #18199, because this makes empty Historicals visible through getDruidServerMetadatas. It also makes them visible through getDruidServers, causes them to show up in the sys.servers table.

### DIFF
--- a/server/src/main/java/org/apache/druid/client/BatchServerInventoryView.java
+++ b/server/src/main/java/org/apache/druid/client/BatchServerInventoryView.java
@@ -65,7 +65,7 @@ public class BatchServerInventoryView implements ServerInventoryView, FilteredSe
   private final CuratorInventoryManager<DruidServer, Set<DataSegment>> inventoryManager;
   private final AtomicBoolean started = new AtomicBoolean(false);
 
-  private final ConcurrentMap<ServerRemovedCallback, Executor> serverRemovedCallbacks = new ConcurrentHashMap<>();
+  private final ConcurrentMap<ServerCallback, Executor> serverCallbacks = new ConcurrentHashMap<>();
   private final ConcurrentMap<SegmentCallback, Executor> segmentCallbacks = new ConcurrentHashMap<>();
 
   private final ConcurrentMap<String, Set<DataSegment>> zNodes = new ConcurrentHashMap<>();
@@ -127,13 +127,14 @@ public class BatchServerInventoryView implements ServerInventoryView, FilteredSe
           public void newContainer(DruidServer container)
           {
             log.info("New Server[%s]", container);
+            runServerCallbacks(callback -> callback.serverAdded(container));
           }
 
           @Override
           public void deadContainer(DruidServer deadContainer)
           {
             log.info("Server Disappeared[%s]", deadContainer);
-            runServerRemovedCallbacks(deadContainer);
+            runServerCallbacks(callback -> callback.serverRemoved(deadContainer));
           }
 
           @Override
@@ -216,9 +217,9 @@ public class BatchServerInventoryView implements ServerInventoryView, FilteredSe
   }
 
   @Override
-  public void registerServerRemovedCallback(Executor exec, ServerRemovedCallback callback)
+  public void registerServerCallback(Executor exec, ServerCallback callback)
   {
-    serverRemovedCallbacks.put(callback, exec);
+    serverCallbacks.put(callback, exec);
   }
 
   @Override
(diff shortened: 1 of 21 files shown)


server/src/main/java/org/apache/druid/client/BatchServerInventoryView.java
```diff
@@ -243,13 +244,13 @@ protected void runSegmentCallbacks(
     }
   }
 
-  private void runServerRemovedCallbacks(final DruidServer server)
+  private void runServerCallbacks(final Function<ServerCallback, CallbackAction> fn)
   {
-    for (final Map.Entry<ServerRemovedCallback, Executor> entry : serverRemovedCallbacks.entrySet()) {
+    for (final Map.Entry<ServerCallback, Executor> entry : serverCallbacks.entrySet()) {
```
kfaraz, line 249: Nit: Probably cleaner to use `Map.forEach`.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -157,11 +157,24 @@ public CallbackAction segmentSchemasAnnounced(SegmentSchemas segmentSchemas)
...
+    baseView.registerServerCallback(
         exec,
-        server -> {
-          removeServer(server);
-          return CallbackAction.CONTINUE;
+        new ServerCallback() {
+          @Override
+          public CallbackAction serverAdded(DruidServer server)
+          {
+            if (!server.getType().equals(ServerType.BROKER)) {
```
kfaraz, line 167: Please add a 1-liner comment like "Do not keep inventory of other Brokers" or something.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -157,11 +157,24 @@ public CallbackAction segmentSchemasAnnounced(SegmentSchemas segmentSchemas)
...
+              addServer(server);
+            }
+            return CallbackAction.CONTINUE;
```
kfaraz, line 170: Do we ever use the other callback action `UNREGISTER`?
I think it is not used anymore and can probably be removed.

server/src/main/java/org/apache/druid/client/ServerView.java
```diff
@@ -29,7 +29,7 @@
  */
 public interface ServerView
 {
-  void registerServerRemovedCallback(Executor exec, ServerRemovedCallback callback);
+  void registerServerCallback(Executor exec, ServerCallback callback);
```
kfaraz, line 32: Super nit:

To allow for more concise syntax (using lambdas) in the calling code, I wonder if we should have an overloaded method with the signature:

```java
  void registerServerCallback(Executor exec, Consumer<DruidServer> onAdded, Consumer<DruidServer> onRemoved);
```
assuming we get rid of the `CallbackAction` altogether.

`ServerCallback` can then be a concrete class which takes two `Consumer<DruidServer>` in its constructor, default values for which would be no-op.

It might be useful even when we add new methods to `ServerCallback` as callers would need to override only the required methods.

kfaraz (approved):
LGTM, minor non-blocking nitpicks.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
gianm: Added a comment `We don't track brokers in this view.`

server/src/main/java/org/apache/druid/client/BatchServerInventoryView.java
gianm: Personally, I prefer the regular `for` in situations where the `forEach` can't be a one-liner (& even sometimes then) so I kept it as-is.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
gianm: It's not used, & seems like it hasn't been used since #6742. I suppose we could remove it in this PR or in a new one.

server/src/main/java/org/apache/druid/client/ServerView.java
gianm: A builder might be nice too. It would allow keeping together all the `ServerCallback` stuff in the interface. Something like:

```java
ServerCallback.builder()
              .onServerAdded(server -> ...)
              .onServerRemoved(server -> ...)
              .build()
```

or for just doing one:

```java
ServerCallback.onServerRemoved(server -> ...)
```

etc.

Would like to have this be a different PR, so that PR can focus on the thinking about the best way to specify callbacks that only want to listen to one thing.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
kfaraz: Thanks!

PR #18201, opened 2025-07-03
Embedded tests for MSQ realtime queries.

gianm:
This patch adds an `EmbeddedMSQRealtimeQueryTest`, and `EmbeddedMSQApis` to support it.

This patch also moves various usages of RuntimeInfo to regular injection rather than using a static function, and creates a RuntimeInfoModule that is part of the embedded tests and delivers a normal, non-statically-injected RuntimeInfo.

This was necessary because the MSQ tasks, as they start up within the multi-task Indexer, need to see more than 100 MB of total Indexer memory in order to believe that there is sufficient memory to move forwards. Static injection of RuntimeInfo makes this difficult, since the various servers end up clobbering each other, and it isn't guaranteed that the correct RuntimeInfo will be used.


services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
```diff
@@ -86,6 +86,7 @@ private EmbeddedDruidCluster()
   {
     resources.add(testFolder);
     clusterApis = new EmbeddedClusterApis(this);
+    addExtension(RuntimeInfoModule.class);
```
kfaraz, line 92: Please add a list item for this in the class-level javadoc.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -0,0 +1,383 @@
...
+           .addExtension(DartWorkerModule.class)
+           .addExtension(DartControllerMemoryManagementModule.class)
+           .addExtension(DartControllerModule.class)
+           .addExtension(DartWorkerMemoryManagementModule.class)
+           .addExtension(DartWorkerModule.class)
+           .addExtension(IndexerMemoryManagementModule.class)
+           .addExtension(MSQDurableStorageModule.class)
+           .addExtension(MSQIndexingModule.class)
+           .addExtension(MSQSqlModule.class)
+           .addExtension(SqlTaskModule.class)
```
kfaraz, line 136: Awesome! 😂

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    kafka.deleteTopic(topic);
+  }
+
+  @Test
+  @Timeout(60)
+  public void test_selectCount_task_default()
+  {
+    final String sql = StringUtils.format("SELECT COUNT(*) FROM \"%s\"", dataSource);
```
kfaraz, line 220: We probably don't need to use `StringUtils.format()` since `msqApis.runTaskSql()` already accepts formatting arguments.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -0,0 +1,383 @@
...
+        dataSource,
+        null,
+        DataSchema.builder()
+                  .withDataSource(dataSource)
+                  .withTimestamp(new TimestampSpec("__time", "auto", null))
+                  .withGranularity(new UniformGranularitySpec(Granularities.DAY, null, null))
+                  .withDimensions(DimensionsSpec.builder().useSchemaDiscovery(true).build())
+                  .build(),
+        null,
+        new KafkaSupervisorIOConfig(
```
kfaraz, line 344: Should we add a builder or static utility method for creating a `KafkaSupervisor` and `KafkaSupervisorIOConfig` etc since several tests will need to use them?

(best to do it in a follow up PR)

kfaraz (approved):
LGTM. Left some minor suggestions.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
gianm: It's saved into a variable here because later on, the `assertResultsEquals` function being used wants the SQL to include in its error messages.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
gianm: Added this

```
 * <li>{@link RuntimeInfoModule} supplying a {@link RuntimeInfo} with values matching
 * {@link EmbeddedDruidServer#setServerMemory} and {@link EmbeddedDruidServer#setServerDirectMemory}</li>
```

kfaraz: Thanks!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
gianm: Yes, and I feel like some kind of utilities even beyond such builders would help with writing realtime tests. I didn't stop to think about what those might look like exactly though.

kfaraz: Fair enough, will keep it in mind when we start migrating tests from the old IT frameworks.

PR #18207, opened 2025-07-04
Migrate several indexing and compaction integration tests to embedded-tests

kfaraz:
### Summary

- This patch migrates several indexing and compaction integration tests (several of which have been flaking out a lot recently) to the embedded-tests framework.
- Overall, this migration will save about __1 hr 30 mins__ of GitHub runner time.
- The MiddleManager version of the integration tests has currently not been ported as they are slower and currently not desirable in embedded tests (because they launch child processes).
- The MM-based embedded-tests can be easily enabled at any point in the future.

### Changes

- The new tests use SQL queries to verify results instead of native since the SQL syntax is much more concise and makes for more readable unit tests. We can add native queries later or in future tests that are migrated to this framework.
- Add `TaskBuilder` utility to create `Task` objects using fluent syntax
- Add JSON data resource files to `embedded-tests`

### Test migrations

|Old test|New test|
|--------|---------|
|`ITPerfectRollupParallelIndexTest`|`IndexParallelTaskTest`|
|`ITBestEffortRollupParallelIndexTest`|`IndexParallelTaskTest`, added as a new test parameter which uses dynamic partitioning|
|`ITAutoCompactionTest`|`AutoCompactionTest`|
|`ITAutoCompactionLockContentionTest`|`AutoCompactionUpgradeTest`|
|`ITAutoCompactionLockContentionTest`|`KafkaClusterMetricsTest`, method `test_ingestClusterMetrics_compactionSkipsLockedIntervals()`|
|`ITCompactionTaskTest`|`CompactionTaskTest`|
|`ITCompactionSparseColumnTest`|`CompactionSparseColumnTest`|
|`ITOverlordResourceTest`|Already being verified in `OverlordClientTest`|
|`ITOverlordResourceNotFoundTest`|Already being verified in `OverlordClientTest`|

### New nested tests

- `EmbeddedCentralizedSchemaPublishFailureTest` for the group `cds-task-schema-publish-disabled`
- `EmbeddedCentralizedSchemaMetadataQueryDisabledTest` for the group `cds-coordinator-metadata-query-disabled`

### Test run times

|Test|Actual test time|Total time including setup|
|----|-------------|-----------------------------|
|`IndexParallelTaskTest`(indexer)|38 s|__38 s__ |
|`AutoCompactionTest` (indexer)|1 min 20 s|__1 min 20 s__|
|`CompactionSparseColumnTest` (indexer)|12 s|__12 s__|
|Standard `ITPerfectRollupParallelIndexTest`|10 min|15 min|
|Standard `ITPerfectRollupParallelIndexTest`<br/>(Indexer, shuffle deep store test, only 1 config changed)|10 min|15 min|
|Standard `ITPerfectRollupParallelIndexTest`<br/>(MM, shuffle deep store test, only 1 config changed)|10 min|15 min|
|Standard `ITBestEffortRollupParallelIndexTest`|2 min| NA(setup includes other tests too) |
|Revised `ITBestEffortRollupParallelIndexTest`|4 min 10 s| " |
|Standard `ITAutoCompactionTest` (middle manager)|25 mins|"| 
|Standard `ITAutoCompactionTest` (indexer)|15 mins|"|
|Standard `ITCompactionSparseColumnTest` (indexer)|2 min 10 s|"|

---

This PR has:

- [ ] been self-reviewed.


embedded-tests/src/test/resources/data/json/wikipedia_1.json
```diff
@@ -0,0 +1,3 @@
+{"timestamp": "2013-08-31T01:02:33Z", "page": "Gypsy Danger", "language" : "en", "tags": ["t1", "t2"], "user" : "nuclear", "unpatrolled" : "true", "newPage" : "true", "robot": "false", "anonymous": "false", "namespace":"article", "continent":"North America", "country":"United States", "region":"Bay Area", "city":"San Francisco", "added": 57, "deleted": 200, "delta": -143}
```
clintropolis: drive by comment/nit: i know this isn't new, but imo we should fix the problem of referring to this dataset as 'wikipedia' because it is confusing with the quickstart wikipedia data which is also going to be used in some tests, and this stuff only has a vaguely similar schema, maybe `tiny-wikipedia` or something to indicate that its a very small dataset would help clear things up?

kfaraz: Thanks for the suggestion! I will rename these datasets accordingly.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedIndexParallelTaskTest.java
```diff
@@ -0,0 +1,208 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Runs tasks of "index_parallel" type using Indexers.
+ */
+public class EmbeddedIndexParallelTaskTest extends EmbeddedClusterTestBase
```
clintropolis: wondering, do we need to prefix all the test names with `Embedded` since they are all already in `org.apache.druid.testing.embedded.` package?  is this just to be like how we did with the old integration test names?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/EmbeddedAutoCompactionTest.java
```diff
@@ -85,68 +90,177 @@
...
-  private static final String INDEX_TASK_WITH_GRANULARITY_SPEC = "/indexer/wikipedia_index_task_with_granularity_spec.json";
-  private static final String INDEX_TASK_WITH_DIMENSION_SPEC = "/indexer/wikipedia_index_task_with_dimension_spec.json";
-  private static final String INDEX_ROLLUP_QUERIES_RESOURCE = "/indexer/wikipedia_index_rollup_queries.json";
-  private static final String INDEX_ROLLUP_SKETCH_QUERIES_RESOURCE = "/indexer/wikipedia_index_sketch_queries.json";
-  private static final String INDEX_QUERIES_RESOURCE = "/indexer/wikipedia_index_queries.json";
-  private static final String INDEX_TASK_WITH_ROLLUP_FOR_PRESERVE_METRICS = "/indexer/wikipedia_index_rollup_preserve_metric.json";
-  private static final String INDEX_TASK_WITHOUT_ROLLUP_FOR_PRESERVE_METRICS = "/indexer/wikipedia_index_no_rollup_preserve_metric.json";
+  private static final Logger LOG = new Logger(EmbeddedAutoCompactionTest.class);
+  private static final Supplier<TaskPayload> INDEX_TASK =
+      () -> TaskPayload
```
clintropolis: hm, if we're going to build these with code why not just use the real types to make stuff? making maps like this feels error prone, and we lose out on all the nice builders that already exist for some of the stuff

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedIndexParallelTaskTest.java
kfaraz: No, I generally like it mostly because it makes it easy to find and identify the tests and also to segregate them into a separate CI job when needed later (since they can be sprinkled across different modules).

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/EmbeddedAutoCompactionTest.java
kfaraz: Yeah, that's a fair point.

I think I started out with this so that the client side didn't need to add a bunch of modules to the ObjectMapper.
I remember something had failed due to the absence of some types.
However, since we are anyway using the `HttpClient` and `ObjectMapper` from one of the servers itself, I can't recall why it had failed exactly. Probably just didn't get serialized out correctly.

Let me try using some of the existing builders.
Could you point me to the builder (if one exists) for a general `IndexTask` and `ParallelIndexSupervisorTask`?
If one doesn't exist, I will just try to convert the new builder in this PR to suit that purpose.

clintropolis: ah i guess maybe there aren't any builders for the top level tasks, but much of the schema and config stuff at least has builders

kfaraz: Hmm, let me write something up then.
It would make sense to have a clean builder that can be used in other UTs too.

I think this same builder can be repurposed to do that just by returning a concrete object rather than a map.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedIndexParallelTaskTest.java
clintropolis: wondering if we should still call these integration tests, since they really are basically, just using a bit different framework to run them than the other frameworks.

clintropolis: to clarify, none of this discussion is about changes to make in this PR btw, just thinking about stuff in general

kfaraz: > wondering if we should still call these integration tests, since they really are basically, just using a bit different framework to run them than the other frameworks.

Yeah, I don't mind calling a test, say `IndexTaskIntegrationTest` or even any other name really, as long as we follow a single scheme across all modules. (I would just prefer not to retain the existing nomencluture of `IT*Test` as it always felt redundant 😛 ).

I personally liked the ring of `Embedded*Test` and it has the added advantage of denoting that the test uses an embedded cluster (along with the identification and segregation benefits that I mentioned above).
Since it uses an entire cluster, the fact that it is an integration test rather than a plain unit test is also implied.
This name also felt appropriate since the primary module is called `embedded-tests`.

Let me know if you have any particular reservations against the current nomenclature or if you feel `*IntegrationTest` (or other similar names) would offer any specific cognitive advantage.

kfaraz:
Thanks for the feedback, @clintropolis !

I have made the following changes:
- Renamed the json data files to the format `tiny_wiki_x.json`
- Added `TaskBuilder` which creates `Task` objects and uses various existing builders. I have retained some utility methods for shorthand.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedIndexParallelTaskTest.java
kfaraz: As discussed offline, removed the `Embedded*` prefix to simplify the test names.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/AutoCompactionTest.java
```diff
@@ -85,68 +88,141 @@
...
+      Pair.of(Resources.Query.SELECT_APPROX_COUNT_DISTINCT, "5,5"),
+      Pair.of(Resources.Query.SELECT_EARLIEST_LATEST_USER, "nuclear,stringer"),
+      Pair.of(Resources.Query.SELECT_COUNT_OF_CHINESE_PAGES, "Crimson Typhoon,1,905.0,9050.0")
+  );
+  private static final Supplier<TaskBuilder.IndexParallel> INDEX_TASK_WITH_ROLLUP_FOR_PRESERVE_METRICS =
+      () -> TaskBuilder
+          .ofTypeIndexParallel()
+          .jsonInputFormat()
+          .inlineInputSourceWithData(Resources.InlineData.JSON_2_ROWS)
+          .isoTimestampColumn("timestamp")
```
gianm, line 132: I wonder if it would be nicer to have `.timestampColumn("timestamp", "iso")`. Makes it easier to sub in `"auto"` or whatever other format.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/AutoCompactionTest.java
```diff
@@ -1742,84 +1639,85 @@ public void testAutoCompactionDutyWithOverlappingInterval() throws Exception
...
+        StringUtils.format("Query[%s] failed", query)
+    );
   }
 
-  private void verifyQuery(String queryResource, Map<String, Object> keyValueToReplace) throws Exception
+  /**
+   * Verifies the result of a SELECT query
+   *
+   * @param field  Field to select
+   * @param result CSV result with special strings {@code ||} to represent
```
gianm: why do this rather than accept the CSV as-is?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IndexTaskTest.java
```diff
@@ -50,7 +50,7 @@
 /**
  * Simulation tests for batch {@link IndexTask} using inline datasources.
```
clintropolis: nit: unrelated to this PR but did a search and see 2 usages of this terminology in embedded-tests, here and `KafkaClusterMetricsTest` which if i remember correctly was the old name before we settled on embedded (there are another 2 that show up but are related to coordinator simulation stuff which seem legit to call them that)

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionTestBase.java
```diff
@@ -0,0 +1,134 @@
...
+import org.junit.jupiter.api.Assertions;
+
+import java.io.Closeable;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+
+public abstract class CompactionTestBase extends EmbeddedClusterTestBase
```
clintropolis, line 39: other than maybe the cluster config, most of the utility methods on this class don't seem too related to compaction, should they live somewhere more common? Is fine to change this later, just want to avoid copy and paste of these methods down the line or like weird extending of compaction test base for things that aren't doing any compaction

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/AutoCompactionTest.java
```diff
@@ -1989,6 +1884,7 @@ private void forceTriggerAutoCompaction(
       );
 
       // Wait for scheduler to pick up the compaction job
+      // TODO: make this latch-based
```
clintropolis: should either do this or transition from a TODO to a comment about how someone in the future can improve this

gianm: Yes, please don't commit todo comments.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionTestBase.java
```diff
@@ -0,0 +1,134 @@
...
+      overlord.bindings().segmentsMetadataStorage().markAllSegmentsAsUnused(dataSource);
+    };
+  }
+
+  /**
+   * Creates a Task using the given builder and runs it.
+   *
+   * @return ID of the task.
+   */
+  protected String runTask(TaskBuilder<?, ?, ?> taskBuilder, String dataSource)
```
gianm: IMO, `protected` methods on a base class aren't the best way to do utility APIs. Other test cases might want some of these utility methods, but not to be "compaction tests". Also, some tests might want to extend multiple base classes, and this approach makes it impossible.

With MSQ tests we have `EmbeddedMSQApis`, something that collects together utility APIs without using a base class approach. Something similar might work here?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/Resources.java
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.druid.query.aggregation.datasketches.hll.HllSketchBuildAggregatorFactory;
+import org.apache.druid.query.aggregation.datasketches.quantiles.DoublesSketchAggregatorFactory;
+import org.apache.druid.query.aggregation.datasketches.theta.SketchMergeAggregatorFactory;
+
+import java.util.function.Supplier;
+
+/**
+ * Constants and utility methods used in embedded cluster tests.
+ */
+public class Resources
```
gianm: Moving this to `embedded-tests` means that it's now tougher for extensions to write their own tests that pull in `Resources`. Can you split this up? Such as having some `Resources` in `services`, and `MoreResources` in `embedded-tests`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/AutoCompactionTest.java
kfaraz: Yeah, I didn't want to do this initially, but it made for more readable tests as it avoided all the escaping of empty strings and newlines.

```java
verifyScanResult("added", "...||31||...||62");
```
vs
```java
verifyScanResult("added", "\"\"\n31\"\"\n62");
```

Please let me know if this seems hacky and if you feel that it is cleaner to just use the original.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/AutoCompactionTest.java
kfaraz: Sorry, must have missed this in the cleanup.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/AutoCompactionTest.java
kfaraz: I did have that initially. But realized that most tests are just using an ISO timestamp. So used this syntax sugar instead.

For a fully custom timestamp, there is also the option to use.
```java
.dataSchema(d -> d.withTimestamp(new TimestampSpec(...)))
```

Please let me know which one you prefer.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionTestBase.java
kfaraz: Fair enough, I will move the common utility methods to some utility class similar to `EmbeddedMsqApis` as suggested by @gianm [here](https://github.com/apache/druid/pull/18207#discussion_r2237965907)

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionTestBase.java
kfaraz: Makes sense, thanks for the suggestion!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IndexTaskTest.java
kfaraz: Thanks for spotting these!

kfaraz:
> One thing I was thinking about while reviewing is that i believe we are perhaps losing some minor coverage of auth stuff during this transition since I think the base ITs had basic-auth setup, though afaict not much in the way of roles and stuff in most tests (so i think was only authentication that would have been tested except for those that extend AbstractAuthConfigurationTest). I think that is probably fine though as long as we migrate the auth tests over to run on this framework, though they probably don't cast quite a wide of net in terms of APIs being called since those tests are more focused on authorization. I don't think I am suggesting we just bake basic auth into random tests or anything, and really its a bit of a negative of the old frameworks that you have to hunt across several files to determine what configuration is actually active for a given test, but maybe something we should watch out for as we move tests over.

Thanks for calling this out, @clintropolis !

I do plan to migrate the auth tests as well to the embedded framework.
For now, we will migrate the non-auth tests without basic auth enabled.
Once all tests are migrated, we can add test variants which have auth enabled to increase API coverage.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/Resources.java
kfaraz: Updated.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionTestBase.java
kfaraz: Moved most methods to `EmbeddedClusterApis` itself so that all tests can benefit from it.
I have still kept some of the protected methods in `CompactionTestBase` but these mostly just act as syntax sugar over the methods in `EmbeddedClusterApis` to keep the diff in the compaction test classes small.

kfaraz:
@gianm , @clintropolis , thanks for the feedback!
I have updated the PR based on your suggestions.
Please let me know if further changes are required.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/AutoCompactionTest.java
gianm: I see. To me the first form is hard to get used to. I keep wanting to read the `||` as field separators rather than newlines. The second is also weird looking though. I can't think of a good solution immediately. I'm ok with what you think is best.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/AutoCompactionTest.java
gianm: I'm ok with either one.

gianm:
> @gianm , @clintropolis , thanks for the feedback! I have updated the PR based on your suggestions. Please let me know if further changes are required.

Approved, since the only remaining comment (about `verifyScanResult`) is nonblocking.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/AutoCompactionTest.java
kfaraz: I tried a bunch of different symbols, but nothing really conveyed a "newline" well-enough, not even `↵` or `⏎`.
The empty strings are probably the worse of the two though.
Using ellipsis for now, but keeping newline `\n` as is.

So, we would have something like this:

```java
verifyScanResult("added", "...\n31\n...\n62");
```

PR #18209, opened 2025-07-07
Router going oom in revised IT's

cryptoe:
Had some failing IT's where the router was going OOM. Bumped it up.


kfaraz (approved):
@cryptoe , which ITs are the ones going OOM?

cryptoe:
@kfaraz There are optional GCS,Azure storage IT tests which are not enabled by default since we require private creds to enable them.

PR #16849, opened 2024-08-06
Segments sorted by non-time columns.

gianm:
## Summary

Currently, segments are always sorted by `__time`, followed by the sort order provided by the user via `dimensionsSpec` or `CLUSTERED BY`. Sorting by `__time` enables efficient execution of queries involving time-ordering or granularity. Time-ordering is a simple matter of reading the rows in stored order, and granular cursors can be generated in streaming fashion.

However, for various workloads, it's better for storage footprint and query performance to sort by arbitrary orders that do not start with `__time`. With this patch, users can sort segments by such orders.

## API

For spec-based ingestion, users add `forceSegmentSortByTime: false` to dimensionsSpec. The `dimensions` list determines the sort order. To define a sort order that includes `__time`, users explicitly include a dimension named `__time`.

For SQL-based ingestion, users set the context parameter `forceSegmentSortByTime: false`. The CLUSTERED BY clause is then used as the explicit segment sort order.

In both cases, when the new `forceSegmentSortByTime` parameter is `true` (the default), `__time` is implicitly prepended to the sort order, as it always was prior to this patch.

The new parameter is experimental for two main reasons. First, such segments can cause errors when loaded by older servers, due to violating their expectations that timestamps are always monotonically increasing. Second, even on newer servers, not all queries can run on non-time-sorted segments. Scan queries involving time-ordering and any query involving granularity will not run. (To partially mitigate this, a currently-undocumented SQL feature `sqlUseGranularity` is provided. When set to false the SQL planner avoids using `granularity`.)

## Main changes

Changes on the write path:

1) `DimensionsSpec` can now optionally contain a `__time` dimension, which controls the placement of `__time` in the sort order. If not present, `__time` is considered to be first in the sort order, as it has always been.

2) `IncrementalIndex` and `IndexMerger` are updated to sort facts more flexibly; not always by time first.

3) `Metadata` (stored in metadata.drd) gains a `sortOrder` field.

4) MSQ can generate range-based shard specs even when not all columns are singly-valued strings. It merely stops accepting new clustering key fields when it encounters the first one that isn't a singly-valued string. This is useful because it enables range shard specs on `someDim` to be created for clauses like `CLUSTERED BY someDim, __time`.

5) Auto-compaction respects and propagates sort orders that don't start with `__time`.

Changes on the read path:

1) Update cursor holders for `QueryableIndex` and `IncrementalIndex` to return the ordering of the underlying index, so query engines can tell how a segment is sorted.

[...]

PR #18215, opened 2025-07-08
fix issue with projection __time equivalent column mapping

clintropolis:
### Description
Fixes some issues with project `__time` column remapping. The first issue is that a projection with finer granularity would incorrectly replace a coarser granularity `time_floor` function with the projection `__time` column, instead of retaining the `time_floor` operation to further rollup the projection at query time, producing incorrect results. The added test covers this. Fixing this stumbled into another issue with remapping the `__time` column when it was the same as the projection granularity, which would incorrectly map the matching query virtual column to the internal projection time column name instead of to `__time` itself. idk what i was thinking with the previous logic 🤷


processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -1102,6 +1102,61 @@ public void testQueryGranularityFitsProjectionGranularity()
...
 
+  @Test
+  public void testQueryGranularityLargerProjectionGranularity()
+  {
+    final GroupByQuery.Builder queryBuilder =
+        GroupByQuery.builder()
+                    .setDataSource("test")
+                    .setInterval(Intervals.ETERNITY)
+                    .addAggregator(new LongSumAggregatorFactory("c_sum", "c"));
+    if (sortByDim) {
```
cecemei: i thought this meant query has a sortBy at first, but i guess here's just we want to test out time as granularity & virtual column? might worth some comments.

cecemei: also, some projections are named as daily, e.x. abfoo_daily but it's not grouping by DAY right?

cecemei (approved):

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
clintropolis: >i thought this meant query has a sortBy at first, but i guess here's just we want to test out time as granularity & virtual column? might worth some comments.

yea, `sortByDim` is whether or not the segment we are querying is stored sorted first by `__time` or not, its one of the test parameters which controls if we set the `forceSegmentSortByTime` parameter on `DImensionsSpec` added in #16849. When this is set, using a granularity on a query doesn't work because the assumptions the engines make about the segments when using granularity (which acts like an implicit group by time_floor(__time, granularity)) are that the rows of the cursor are ordered by __time and so it can just scan and collect the time bucket group of the granularity as it goes instead of grouping on it properly as if it were a dimension. So for group by, we have to explicitly make a time_floor and group on it instead of using the query granularity. Timeseries queries don't work either if the segment isn't first sorted by __time, so some of the tests also just skip if the segment isn't time sorted. I can rename this parameter to `segmentSortedByTime` or something and it might be clearer?

>also, some projections are named as daily, e.x. abfoo_daily but it's not grouping by DAY right?

ah I guess I could rename them to leave the granularity off of the implicit ones, some of those are projections without a __time equivalent column, but they still live within the constraints of the segment granularity, so its an implicit time_floor of day in this case since that is when the segment interval starts.

PR #16380, opened 2024-05-03
Do not allocate ids conflicting with existing segment ids

AmatyaAvadhanula:
### Bug

Consider a situation where there are:
1) No used segments
2) Unused segments with ids ranging from 1 to 10
3) Pending segments with ids ranging from 1 to 5 due to partial clean up

Previously, new segment allocations would consider only used and pending segments and allocate a segment with id 6.
This is problematic as it conflicts with an existing unused entry in the segments table.

### Cause

The bug was discovered because of the pending segment cleanup introduced recently as part of https://github.com/apache/druid/pull/16144. 
This PR introduced pending segment clean up but it is partial as it cleans pending segments created only after the overlord's upgrade.
Sequence of events:
1) Allocate and commit segments with ids from 1 to 5 
2) Checkout to a commit containing the above PR
3) Allocate and commit segments with ids from 6 to 10. The pending segments would be cleaned up upon completion.
4) Mark all segments from 1 - 10 as unused.
5) Allocate a new segment.

### Changes

After a segment has been allocated, check that it doesn't correspond to any used or unused segment id in druid_segments table.
If it does, try to look at the unused segments and increment the partitionNum to max(unusedSegment.partitionNum) + 1 if needed.
Even though this requires fetching all unused segments for a given (datasource, version, interval) when there is a conflict, the conflict would arise due to unused segments at most once and would not affect batch segment allocation's performance.

### Alternatives
Unused pending segments must be cleaned up or ignored when they are not associated with any existing active lock.

This PR has:

- [x] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [x] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [x] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [x] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [x] been tested in a test Druid cluster.

PR #18216, opened 2025-07-08
Fix concurrent append to interval with only unused segments

kfaraz:
### Bug

Concurrent append uses lock of type APPEND which always uses a lock version of epoch `1970-01-01`.

This can cause data loss in a flow as follows:
- Ingest data using an APPEND task to an empty interval
- Mark all the segments as unused
- Re-run the APPEND task
- Data is not visible since old segment IDs (now unused) are allocated again

### Fix

In segment allocation, do not reuse an old segment ID, used or unused.
This fix was already done for some cases back in #16380 .

An embedded test for this has been included in #18207  [EmbeddedConcurrentAppendReplaceTest](https://github.com/apache/druid/pull/18207/files#diff-86e908136809a8fc70c63b66243416d943864d76071f3aa8edb8a1a856dd053f)

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/test/java/org/apache/druid/indexing/common/task/concurrent/ConcurrentReplaceAndAppendTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/common/task/concurrent/ConcurrentReplaceAndAppendTest.java
@@ -1119,6 +1119,38 @@ public class ConcurrentReplaceAndAppendTest extends IngestionTestBase
     verifyIntervalHasVisibleSegments(JAN_23, segmentV10, segmentV11, segmentV12);
   }
 
+  @Test
+  public void test_concurrentAppend_toIntervalWithUnusedSegments()
+  {
+    // Allocate and commit an APPEND segment
+    final SegmentIdWithShardSpec pendingSegment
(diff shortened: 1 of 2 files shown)


jtuglu1 (approved):
LGTM

imply-cheddar:
This code change will append the new segment to the old set of segments.  If that old set of segments had a non-zero core partition numbers, then the appended segment will carry the same set of core partitions (non-zero) and will publish the new appended segment with the extra appended partition num.

This new segment will never actually be used in queries because the core partitions will never be fully loaded by the system as it is not a complete partition of data (the semantics of the system are that the core partitions must all exist for the segments in that time range to be considered complete, extra partitions *can* exist, but do not have to).  This means that even in the case where this showed up (a segment was dropped and then new data came in), the newly created segment won't actually be queried.

The correct fix for this is to generate a net-new version that is "better" than the previous version, adding more partitions to the unused segments is weird.  IIRC, we used the static 1970 version just as a short-circuit because we weren't sure what the best version to use was in the case where nothing exists yet.  At this point, we now understand why it's important to use a better version and the version of the lock should be "better" than the unused segments.  Please revert this change and adjust to use a better version.

PR #18219, opened 2025-07-08
remove size checks from IncrementalIndex to simplify logic

clintropolis:
### Description
This PR removes the `skipMaxRowsInMemoryCheck` parameter from `IncrementalIndex.add` since it seems to make things overly complicated to reason about. All callers that add rows check `canAppendRow()` (except the sampler, but it should have a small row limit anyway :shrug:), so this should be safe as far as I can tell.

Because it is no longer needed after the changes to the `add` method, this PR also removes `IndexSizeExceededException`.


cecemei (approved):
The remove size checks makes sense. I wonder why do we need the rowsCurrentlyInMemory and bytesCurrentlyInMemory check logic in appenderator? It looks like it's just checking the size change for a single `InputRow`, and this logic seems already covered in `IncrementalIndex.canAppendRow()`

clintropolis:
> wonder why do we need the rowsCurrentlyInMemory and bytesCurrentlyInMemory check logic in appenderator? It looks like it's just checking the size change for a single InputRow, and this logic seems already covered in IncrementalIndex.canAppendRow()

`rowsCurrentlyInMemory` and `bytesCurrentlyInMemory` are across all of the `Sink` an appenderator has, while `canAppendRow` is for a single partial in memory part of a `Sink`.

PR #18220, opened 2025-07-08
MSQ: Fix composing channels losing partitionless frames.

gianm:
The main change is that single-partition ComposingWritableFrameChannels (i.e. those created by `ChannelOutputFactory#openChannel`) now associate all incoming frames with that partition. Previously, frames might have come in with partition set to `NO_PARTITION`, which would cause them to get "lost" by the composing channel.

Fixes a bug introduced in #18144 when composed intermediate stage-internal channels started being used for the output of hash partitioning. Prior to #18144, they were only used for internal channels of the SuperSorter. This bug could cause frames to go missing during sortMerge joins.

This patch also adds an embedded test for various durable storage scenarios, including sortMerge join tests that would have caught the original bug.

Finally, this patch adjusts the way that Calcites escapes string literals, to use the actual characters more often when possible. This helps format the test SQLs generated by the embedded test more nicely.

### DIFF
--- a/embedded-tests/pom.xml
+++ b/embedded-tests/pom.xml
@@ -188,6 +188,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.apache.curator</groupId>
       <artifactId>curator-test</artifactId>
       <scope>test</scope>


embedded-tests/pom.xml
```diff
@@ -182,6 +182,11 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-all</artifactId>
+      <scope>test</scope>
+    </dependency>
     <dependency>
```
cryptoe, line 190:
```suggestion
    <dependency>
      <groupId>org.hamcrest</groupId>
      <artifactId>hamcrest-core</artifactId>
      <scope>test</scope>
    </dependency>
    <dependency>
```

PR #18223, opened 2025-07-09
Add projection names to DataSegment in a backward compatible manner.

cecemei:
### Description

This PR does:
- adds an optional projectionNames field to `DataSegment` to support tracking projection-specific metadata at the segment level
- adds projections field to `sys.segments`, it'd be `DataSegment.getProjections()`

Serialization behavior:
-  If segment is not projection-aware (i.e. `projection==null` in `DataSchema`), segment's projection would be omitted from the serialized JSON. 
- If segment is projection-aware but has no defined projections (i.e. `projection.isEmpty()` in `DataSchema`), its projection would be represented as an empty string `""`.

This behavior aligned with existing serialization/deserialization patterns for projections in classes like `DataSourceMSQDestination`, `DataSchema`, and `CompactionState`.

Additionally, the PR refactors `DataSegment` construction to utilize the existing builder class and deprecates a few constructors.  

<hr>

This PR has:

- [x] been self-reviewed.


processing/src/main/java/org/apache/druid/timeline/DataSegment.java
```diff
@@ -495,20 +502,36 @@ public static class Builder
...
     private ShardSpec shardSpec;
     private CompactionState lastCompactionState;
     private Integer binaryVersion;
     private long size;
 
+    /**
+     * @deprecated use {@link #Builder(SegmentId)} or {@link #Builder(DataSegment)} instead.
+     */
+    @Deprecated
     public Builder()
```
kfaraz: If this is not used anywhere, we can probably just make this constructor private.

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
```diff
@@ -98,6 +99,7 @@ public static class PruneSpecsHolder
   private final Map<String, Object> loadSpec;
   private final List<String> dimensions;
   private final List<String> metrics;
+  private final List<String> projections;
```
kfaraz, line 102: Did you intend to name this field `projectionNames` as mentioned in the PR description?

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
```diff
@@ -230,7 +246,7 @@ public DataSegment(
   }
 
   @Nullable
-  private Map<String, Object> prepareLoadSpec(@Nullable Map<String, Object> loadSpec)
+  private static Map<String, Object> prepareLoadSpec(@Nullable Map<String, Object> loadSpec)
```
clintropolis: nit: if you have to make any additional changes to this PR, maybe move these static methods to be after the instance methods. this is just my personal preference, so not a blocker or anything

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -545,7 +545,7 @@ public void testGetTableMap()
     final RelDataType rowType = segmentsTable.getRowType(new JavaTypeFactoryImpl());
     final List<RelDataTypeField> fields = rowType.getFieldList();
 
-    Assert.assertEquals(19, fields.size());
+    Assert.assertEquals(20, fields.size());
```
clintropolis, line 542: maybe we could add a test for a table with projections?

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -1203,18 +1203,12 @@ private SegmentPublishResult commitAppendSegmentsAndMetadataInTransaction(
...
-                    oldSegment.getDimensions(),
-                    oldSegment.getMetrics(),
-                    pendingSegment.getId().getShardSpec(),
-                    oldSegment.getLastCompactionState(),
-                    oldSegment.getBinaryVersion(),
-                    oldSegment.getSize()
-                )
-            );
+            allSegmentsToInsert.add(DataSegment.builder(oldSegment)
+                                               .dataSource(newVersionSegmentId.getDataSource())
```
kfaraz: This can be removed as the `dataSource` wouldn't change from the `oldSegment`.

server/src/test/java/org/apache/druid/client/ImmutableDruidDataSourceTest.java
```diff
@@ -38,15 +39,21 @@
 
 public class ImmutableDruidDataSourceTest
 {
+  DataSegment TEST_SEGMENT = DataSegment.builder(SegmentId.of("test", Intervals.of("2017/2018"), "version", null))
```
kfaraz: static final

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
```diff
@@ -156,13 +167,19 @@ public DataSegment(
...
-        size
+        size,
+        PruneSpecsHolder.DEFAULT
     );
   }
 
+  /**
+   * @deprecated use {@link #builder(SegmentId)} or {@link #builder(DataSegment)} instead.
+   */
+  @Deprecated
```
kfaraz, line 182: It seems that this PR is removing some usages of these constructors.
If there are no more usages left, we can just remove this constructor or make it private.

kfaraz (commented):
Some minor comments.

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
cecemei: These constructors are used in many places, and I don't believe this PR removes all of them. I'm marking them as deprecated so we can remove them gradually as we encounter them.

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
cecemei: added projection test in segment1 and publishedUncompactedSegment3

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
cecemei: It's somewhat implied that it's just projection names, similar to how metrics and dimensions are listed by name, not by their full definitions.

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
cecemei: true, removed

server/src/test/java/org/apache/druid/client/ImmutableDruidDataSourceTest.java
cecemei: true updated

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
cecemei: true, updated to be private

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
cecemei: updated this

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -583,138 +577,104 @@ public void testSegmentsTable() throws Exception
...
-        0L, //is_overshadowed
-        null, //is_compacted
-        2L  // replication_factor
-    );
-
     // Verify value types.
     verifyTypes(rows, SystemSchema.SEGMENTS_SIGNATURE);
+
+    // segments test1, test2  are published and available.
+    Object[] segment1Expected = new Object[]{
```
clintropolis, line 585: nit: when we have to split declarations across rows we usually do 1 entry per line (except for map key/value pairs)

clintropolis (approved):

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
cecemei: There're 20 elements and it feels noisy and visually bloated when 1 entry per line. Compact formatting looks better for readability in this case IMO.

PR #18227, opened 2025-07-09
supress 2 new CVEs that popped in the druid34 release process

capistrant:
### Description

The mvn cve check turned up two new CVEs for our dependencies. Both of them are unfortunately associated with dependencies that have historically not addressed CVEs and already have suppressions

##### druid-ranger-security 

[CVE-2025-27818](https://nvd.nist.gov/vuln/detail/CVE-2025-27818)
* This dependency is depending on kafka-clients 2.8.x and has resulted in previous supressions by us.
* even the latest version of the ranger dep is still on this old kafka-clients dep.
* I think that it is not exploitable in our env though, unless an audit producer is setup druid side and not just on the ranger servers that this extension relies on for auth? I'm not very familiar with the extension.

##### hadoop-client-runtime

[CVE-2025-52999](https://nvd.nist.gov/vuln/detail/CVE-2025-52999)
* Related to jackson dependency that is shaded into hadoop-client-runtime. We have other supressions related to jackson already. 
* Even upgrading to latest hadoop 3.4.1 (which we can't do yet due to aws clients), wouldn't fix this.

#### Release note

<hr>

 * owasp-dependency-check-supressions.xml

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/owasp-dependency-check-suppressions.xml
+++ b/owasp-dependency-check-suppressions.xml
@@ -142,6 +142,7 @@
     <cve>CVE-2024-47561</cve> <!--  This seems to be a legitimate vulnerability. We would need to go to hadoop-client 3.4 which required aws sdk v2 dependency work to finish -->
     <cve>CVE-2024-29131</cve> <!--  This seems to be a legitimate vulnerability. We would need to go to hadoop-client 3.4 which required aws sdk v2 dependency work to finish -->
     <cve>CVE-2024-22201</cve> <!--  This seems to be a legitimate vulnerability. We would need to go to a hadoop-client which was not yet released  -->
+    <cve>CVE-2025-52999</cve> <!--  This is vulneraability in all versions of hadoop-client-runtime and has not been fixed by hadoop yet -->
   </suppress>
 
   <!-- those are false positives, no other tools report any of those CVEs in the hadoop package -->
@@ -192,6 +193,7 @@
     <cve>CVE-2022-34917</cve>
     <cve>CVE-2023-25194</cve>
     <cve>CVE-2024-31141</cve>
+    <cve>CVE-2025-27818</cve> <!-- not fixed in any version of ranger dependency. I don't think it is exploitable in Druid within this extension -->
   </suppress>
 
   <suppress>


gianm:
Does the ranger extension really use `kafka-clients` or is that getting pulled in extraneously? If it's unnecessary, it's better to exclude `kafka-clients` than to add a CVE suppression.

The `hadoop-client-runtime` suppression seems fine to me. Looking at https://nvd.nist.gov/vuln/detail/CVE-2025-52999 I don't think we'd consider that a meaningful vulnerability with the way we use Hadoop in Druid.

capistrant:
> Does the ranger extension really use `kafka-clients` or is that getting pulled in extraneously? If it's unnecessary, it's better to exclude `kafka-clients` than to add a CVE suppression.


I'll dig into the question on the kafka-clients. I'm skeptical that it actually needs the dependency, but will try to get a definitive answer here. I know ranger has functionality to produce to kafka as an audit mechanism, but I feel like that is more so related to the actual ranger service that this extension is hooking into for auth

abhishekagarwal87:
We should also evaluate if we want to continue keeping the extension around. It's not being actively maintained and comes under discussion only at the release time and there is a CVE to resolve. Let me know if dev mailing list is a better place to start this discussion on.

capistrant:
It does look like you can setup auditing and one of the options is kafka. It is not well documented in the extension page, just briefly mentions providing an audit config. I'm trying to see if the underlying producer allows the operator to supply any set of producer configs they want or just a limited set, potentially excluding the oauth stuff related to this CVE.

capistrant:
> We should also evaluate if we want to continue keeping the extension around. It's not being actively maintained and comes under discussion only at the release time and there is a CVE to resolve. Let me know if dev mailing list is a better place to start this discussion on.

Thanks @abhishekagarwal87 I agree! I have started some adjacent dev threads in apache Druid mailing list. one such is changing our support for contirb extensions in a way that would stop actively scanning for CVEs. The extensions would instead be offered as is and only maintained with best effort. I do agree we should potentially consider processes for removing (archiving?) contrib extensions if they seem not useful/applicable anymore or extremely low usage.

capistrant:
following up on this. For Druid 34 I don't think there is anything we can do but suppress. ranger extension users may legitimately need kafka for audits to keep working so excluding it could break the extension. If we are saying that we are providing contrib extensions as is, suppressing the CVE seems to be the only option.

clintropolis (approved):

PR #18228, opened 2025-07-10
some enhancements to embedded tests for writing query oriented tests

clintropolis:
Some improvements to the embedded test stuff from my experience writing test for #18176, which is a query oriented test.

The primary change is a new `beforeStart` method on `EmbeddedResource` interface, which allows them to perform additional operations when starting up (with access to the `EmbeddedDruidCluster` at this point, so shared resources are available).

I have added usage of this new functionality to clean up how druid service properties are set, specifically `EmbeddedDruidService` has a new `addBeforeStart` method that takes a function add to a list of things to call during its implementation of the new `beforeStart` method. With this, I've cleaned up the `getStartupProperties` method to just copy all the stuff from the map, and moved all of the previous logic into the `beforeStart` hooks (and the historical specific logic into the `EmbeddedHistorical`).

While i was here, I made the durable shuffle test re-use the same ingested data between runs instead of reingesting per test.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDurableShuffleStorageTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDurableShuffleStorageTest.java
@@ -125,7 +125,7 @@ public class EmbeddedDurableShuffleStorageTest extends EmbeddedClusterTestBase
   }
 
   @BeforeAll
-  void setupData() throws IOException
+  final void setupData() throws IOException
   {
     msqApis = new EmbeddedMSQApis(cluster, overlord);
     dataSource = EmbeddedClusterApis.createTestDatasourceName();
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDurableShuffleStorageTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDurableShuffleStorageTest.java
@@ -55,7 +55,7 @@ import org.apache.druid.testing.embedded.minio.MinIOStorageResource;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
 import org.testcontainers.shaded.com.google.common.io.ByteStreams;
@@ -101,17 +101,19 @@ public class EmbeddedDurableShuffleStorageTest extends EmbeddedClusterTestBase
     return EmbeddedDruidCluster
         .withEmbeddedDerbyAndZookeeper()
         .useLatchableEmitter()
-        .addExtension(DartControllerModule.class)
-        .addExtension(DartWorkerModule.class)
-        .addExtension(DartControllerMemoryManagementModule.class)
-        .addExtension(DartWorkerMemoryManagementModule.class)
-        .addExtension(IndexerMemoryManagementModule.class)
-        .addExtension(MSQDurableStorageModule.class)
-        .addExtension(MSQIndexingModule.class)
-        .addExtension(MSQSqlModule.class)
-        .addExtension(SqlTaskModule.class)
-        .addExtension(MSQExternalDataSourceModule.class)
-        .addExtension(S3StorageConnectorModule.class)
+        .addExtensions(
+            DartControllerModule.class,
+            DartWorkerModule.class,
+            DartControllerMemoryManagementModule.class,
+            DartWorkerMemoryManagementModule.class,
+            IndexerMemoryManagementModule.class,
+            MSQDurableStorageModule.class,
+            MSQIndexingModule.class,
+            MSQSqlModule.class,
+            SqlTaskModule.class,
+            MSQExternalDataSourceModule.class,
+            S3StorageConnectorModule.class
+        )
         .addResource(storageResource)
         .addResource(msqStorageResource)
         .addServer(coordinator)
@@ -122,19 +124,25 @@ public class EmbeddedDurableShuffleStorageTest extends EmbeddedClusterTestBase
         .addServer(router);
   }
 
-  @BeforeEach
-  void setUpEach()
+  @BeforeAll
+  void setupData() throws IOException
   {
     msqApis = new EmbeddedMSQApis(cluster, overlord);
     dataSource = EmbeddedClusterApis.createTestDatasourceName();
+    loadWikipediaTable();
+  }
+
+  @Override
+  protected void beforeEachTest()
+  {
+    // do nothing here, the super version of this method generates a new value for dataSource field each time, but
+    // we are setting that in our @BeforeAll where we are just inserting data once and re-using between runs
   }
 
   @Test
   @Timeout(120)
-  public void test_selectFirstPage() throws IOException
+  public void test_selectFirstPage()
   {
-    loadWikipediaTable();
-
     final String sql = StringUtils.format(
         "SET durableShuffleStorage = TRUE;\n"
         + "SET targetPartitionsPerWorker = 3;\n"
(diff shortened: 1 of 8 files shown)


services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
```diff
@@ -166,6 +167,14 @@ public EmbeddedDruidCluster addExtension(Class<? extends DruidModule> moduleClas
     return this;
   }
 
+  @SafeVarargs
+  public final EmbeddedDruidCluster addExtensions(Class<? extends DruidModule>... moduleClasses)
+  {
+    validateNotStarted();
+    extensionModules.addAll(Arrays.asList(moduleClasses));
```
kfaraz: Nit: reads a little better
```suggestion
    extensionModules.addAll(List.of(moduleClasses));
```

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedResource.java
```diff
@@ -39,6 +40,17 @@ public interface EmbeddedResource
    */
   void stop() throws Exception;
 
+
+  /**
+   * Called before {@link #start()} with a pointer to the current cluster. This is primarily useful for any
```
kfaraz, line 45: Sounds more Java-ey?

```suggestion
   * Called before {@link #start()} with a reference to the current cluster. This is primarily useful for any
```

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedServerLifecycle.java
```diff
@@ -151,7 +151,7 @@ private void runServer(ServerRunnable runnable)
     try {
       final Properties serverProperties = new Properties();
       serverProperties.putAll(commonProperties);
-      serverProperties.putAll(server.getStartupProperties(testFolder, zookeeper));
+      serverProperties.putAll(server.getStartupProperties());
```
kfaraz, line 145: Since this was the only usage of `testFolder` and `zookeeper`, we can remove them from the constructor of `EmbeddedServerLifecycle`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDurableShuffleStorageTest.java
```diff
@@ -122,19 +124,25 @@ public EmbeddedDruidCluster createCluster()
         .addServer(router);
   }
 
-  @BeforeEach
-  void setUpEach()
+  @BeforeAll
```
kfaraz, line 127: I am not sure if this `@BeforeAll` method will be run before or after the one defined in `EmbeddedClusterTestBase`, or if the order is even deterministic.

Instead of this, can we just override the `setup` method and do the initialization there?

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
```diff
@@ -166,6 +167,14 @@ public EmbeddedDruidCluster addExtension(Class<? extends DruidModule> moduleClas
     return this;
   }
 
+  @SafeVarargs
+  public final EmbeddedDruidCluster addExtensions(Class<? extends DruidModule>... moduleClasses)
```
kfaraz, line 175: Please add a javadoc here that links to `{@link #addExtension}`.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidServer.java
```diff
@@ -110,6 +150,12 @@ public final T addProperty(String key, String value)
     return (T) this;
   }
 
+  public final T addBeforeStart(BeforeStart hook)
```
kfaraz: Please add a short javadoc to this. It can just link to `EmbeddedResource.beforeStart()`.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidServer.java
```diff
@@ -90,6 +122,14 @@ public void stop() throws Exception
     }
   }
 
+  @Override
+  public void beforeStart(EmbeddedDruidCluster cluster)
```
kfaraz, line 126: Should this method be marked `final`?

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedHistorical.java
```diff
@@ -35,6 +36,20 @@
  */
 public class EmbeddedHistorical extends EmbeddedDruidServer<EmbeddedHistorical>
 {
+  public EmbeddedHistorical()
+  {
+    super();
```
kfaraz: Nit: Is this explicit call needed?

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidServer.java
kfaraz: Nit: On the name, I feel we should either call it `doBeforeStart` or `addBeforeStartHook`.

kfaraz (commented):
Looks good! 🚀 
The startup properties flow seems nicer now.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedResource.java
clintropolis: i just copied the text of `onStarted` :shrug:

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDurableShuffleStorageTest.java
clintropolis: per the docs, https://docs.junit.org/5.0.0/api/org/junit/jupiter/api/BeforeAll.html, the super `@BeforeAll` runs before the subclasses `@BeforeAll` , so this is safe

kfaraz: thanks for the clarification!

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidServer.java
```diff
@@ -66,6 +70,34 @@ public abstract class EmbeddedDruidServer<T extends EmbeddedDruidServer<T>> impl
...
+          final String taskDirectory = cluster.getTestFolder().newFolder().getAbsolutePath();
+          final String storageDirectory = cluster.getTestFolder().newFolder().getAbsolutePath();
+          log.info(
+              "Server[%s] using directories: task directory[%s], logs directory[%s], storage directory[%s].",
+              self.getName(),
+              taskDirectory,
+              logsDirectory,
+              storageDirectory
+          );
+          self.addProperty("druid.host", "localhost");
```
kfaraz, line 86: Just curious, did you run into any issue with this not being hardcoded here?

clintropolis: not currently, but i have had problems with stuff like this in the past so was proactively changing it. The problem i was running into was that some networks DNS services try to be "helpful" and redirect to a search page or the like, so what should be an unresolvable local network hostname ends up being a real address that is definitely not localhost when using `InetAddress.getLocalHost().getCanonicalHostName()`.

kfaraz: I see. That makes sense.
I am currently working on something where I think I might need the non-localhost hostname.
But I will see what I can do to avoid that requirement as it is cleaner to just have it be `localhost`.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidServer.java
clintropolis: not sure, should it? `start` and `stop` are not final either :shrug:

kfaraz: hmm, it's fine either way, I guess.

kfaraz (approved):

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
```diff
@@ -157,7 +157,7 @@ public EmbeddedDruidCluster useLatchableEmitter()
 
   /**
    * Adds an extension to this cluster. The list of extensions is populated in
-   * the common property {@code druid.extensions.modulesForSimulation}.
+   * the common property {@code druid.extensions.modulesForEmbeddedTest}.
```
kfaraz, line 160: Thanks for fixing this!

kfaraz:
UT failure seems unrelated to the change here, @uds5501 is taking a look.

PR #18230, opened 2025-07-10
Fix concurrent append to interval with unused segments

kfaraz:
### Description

This is a better fix for #18216 as suggested by @imply-cheddar  in the [PR comments](https://github.com/apache/druid/pull/18216#issuecomment-3051438524).

#### Note on the original fix

The case of version collision during segment allocation handled in the original PR can happen only if the following conditions are met:
- The task performing the allocation is an append task (`appendToExisting: true`)
   - .Non-append tasks don't use segment allocation anyway and just replace segments that have the same version as their locks.
- The task is using an APPEND lock.
   - Otherwise, the task would have been using an EXCLUSIVE lock which would always have a unique version (due to exclusivity).
- The interval in question has no used segments but it does have some unused segments.
   - If there were no unused segments, there would be no clash.
   - If there were any used segments, we would simply be appending to that version which wouldn't constitute a clash.
- The unused segments in the interval must have been created by a prior task that was also using an APPEND lock.
   - Any other kind of lock would have resulted in a non-1970 version, thus avoiding clash with subsequent APPEND tasks.

__Despite this fact, appending to an already existing version as a side effect is not desirable and can lead to data consistency issues when this version is marked used / unused.__

### Changes

- Use a new segment version rather than allocating new partition numbers for an existing unused version

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/test/java/org/apache/druid/indexing/common/task/concurrent/ConcurrentReplaceAndAppendTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/common/task/concurrent/ConcurrentReplaceAndAppendTest.java
@@ -1141,8 +1141,10 @@ public class ConcurrentReplaceAndAppendTest extends IngestionTestBase
     // Allocate and commit another APPEND segment
     final SegmentIdWithShardSpec pendingSegment2
         = appendTask.allocateSegmentForTimestamp(FIRST_OF_JAN_23.getStart(), Granularities.DAY);
-    Assert.assertEquals(SEGMENT_V0, pendingSegment2.getVersion());
-    Assert.assertEquals(1, pendingSegment2.getShardSpec().getPartitionNum());
+
+    // Verify that the new segment gets a different version
+    Assert.assertEquals(SEGMENT_V0 + "S", pendingSegment2.getVersion());
+    Assert.assertEquals(0, pendingSegment2.getShardSpec().getPartitionNum());
 
     final DataSegment segmentV02 = asSegment(pendingSegment2);
     appendTask.commitAppendSegments(segmentV02);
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -1590,18 +1591,86 @@ private SegmentIdWithShardSpec createNewPendingSegment(
...
+      if (unusedSegmentVersions.contains(candidateVersion.toString())) {
+        candidateVersion.append(PendingSegmentRecord.CONCURRENT_APPEND_VERSION_SUFFIX);
+      } else {
+        foundFreshVersion = true;
+        break;
+      }
+    }
+
+    if (foundFreshVersion) {
+      return new SegmentIdWithShardSpec(
```
imply-cheddar: This is probably such a seldom-done thing that an INFO log indicating that a new version was found and what it is shouldn't flood the logs, but also should provide decent context on what's going on in cases where this ends up getting called/run unexpectedly.

indexing-service/src/test/java/org/apache/druid/indexing/common/task/concurrent/ConcurrentReplaceAndAppendTest.java
```diff
@@ -1141,8 +1141,10 @@ public void test_concurrentAppend_toIntervalWithUnusedSegments()
     // Allocate and commit another APPEND segment
     final SegmentIdWithShardSpec pendingSegment2
         = appendTask.allocateSegmentForTimestamp(FIRST_OF_JAN_23.getStart(), Granularities.DAY);
-    Assert.assertEquals(SEGMENT_V0, pendingSegment2.getVersion());
-    Assert.assertEquals(1, pendingSegment2.getShardSpec().getPartitionNum());
+
+    // Verify that the new segment gets a different version
+    Assert.assertEquals(SEGMENT_V0 + "S", pendingSegment2.getVersion());
+    Assert.assertEquals(0, pendingSegment2.getShardSpec().getPartitionNum());
```
imply-cheddar, line 1150: It would probably also be good to validate the "keeps appending" behavior in the tests as well (i.e. if one is created, marked unused, a new one created, marked unused, created again, marked unused, etc. it should just build mroe and more 'S's)

PR #18232, opened 2025-07-10
Decoupled dart: support unnest/join

kgyrtkirk:
* enables to plan joins and unnest from the decoupled plan.
* `DataSource#segmentmap` usage mostly relies on the native decoupled mode's `SourceDesc` changes
* new tests are run in `DecoupledDart` mode: `CalciteArrayQueryTest` , `CalciteJoinQueryTest` , `CalciteNestedDataQueryTest`
* disabled [FANCY_JOIN_RULES](https://github.com/apache/druid/blob/3413057f3eaff4f135d4452d1def588e6569368f/sql/src/main/java/org/apache/druid/sql/calcite/planner/CalciteRulesManager.java#L225) in decoupled mode - as it was getting into a loop and removing them did not made things worse
* added `UnorderedResultsVerifier` to sort results before comparision in tests
* updated `log4j2.xml`  in msq module to log calcite rule application tries during executiong
* `AbstractDartComponentSupplier` now follows the same delegate pattern as MSQ
* fixed the testbuilder to now reset the properties before `queryContext` calls
* allow `NotYetSupported` to hold multiple reasons - this is necessary in case the same thing fails in different scopes


sql/src/test/java/org/apache/druid/sql/calcite/CalciteNestedDataQueryTest.java
```diff
@@ -883,6 +886,7 @@ public void testGroupByRootSingleTypeString()
     );
   }
 
+  @NotYetSupported(Modes.DART_SECOND_SEGMENT_NOT_SCANNED)
```
gianm: The `DART_SECOND_SEGMENT_NOT_SCANNED` failures are happening due to a problem in `TestTimelineServerView`-- when building a timeline, it exits after adding the first segment. If this is fixed then this `@NotYetSupported` mode won't be needed.

sql/src/test/java/org/apache/druid/sql/calcite/CalciteArraysQueryTest.java
```diff
@@ -2651,6 +2651,7 @@ public void testArrayAggMultiValue()
     );
   }
 
+  @NotYetSupported(Modes.DD_RESULT_MISMATCH_FLOAT_DOUBLE)
```
gianm, line 2667: The result has a `0.1` in it which cannot be represented exactly as a float. I feel like this could be fixed with a results verifier that has some looseness for verifying floating points. Round to 6 decimal places for floats, 15 for doubles?

Or, we could include the rounding in the SQL itself, similar to https://github.com/apache/druid/pull/18183.

sql/src/test/java/org/apache/druid/sql/calcite/CalciteArraysQueryTest.java
```diff
@@ -3875,7 +3876,7 @@ public void testArrayAggGroupByArrayContainsSubquery()
 
   }
 
-  @NotYetSupported(Modes.UNNEST_INLINED)
+  @NotYetSupported({Modes.UNNEST_INLINED, Modes.DD_UNNEST_INLINED})
```
gianm, line 3892: Any idea what all of these various unnest problems are caused by? Seems unchill.

sql/src/test/java/org/apache/druid/sql/calcite/CalciteJoinQueryTest.java
```diff
@@ -5135,6 +5156,7 @@ public void testCountOnSemiJoinSingleColumn(Map<String, Object> queryContext)
     );
   }
 
+  @NotYetSupported(Modes.DD_RESTRICTED_DATASOURCE_SUPPORT2)
```
gianm, line 5149: What is the problem here? The base test is expecting an error about how the rhs of a join can't be restricted, but I think the annotation says that in decoupled-dart, we see "Unauthorized" instead? Why is that?

sql/src/test/java/org/apache/druid/sql/calcite/CalciteJoinQueryTest.java
```diff
@@ -5697,7 +5723,7 @@ public void testPlanWithInFilterMoreThanInSubQueryThreshold()
     );
   }
 
-  @NotYetSupported(Modes.SORT_REMOVE_TROUBLE)
+  @NotYetSupported({Modes.SORT_REMOVE_TROUBLE, Modes.DD_SORT_REMOVE_TROUBLE})
```
gianm, line 5716: The error message is like "Calcite assertion violated", do we have a buggy rule?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/stages/SegmentMapStage.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.apache.druid.msq.exec.StageProcessor;
+import org.apache.druid.msq.logical.LogicalInputSpec;
+import org.apache.druid.msq.logical.StageMaker;
+import org.apache.druid.sql.calcite.planner.querygen.DruidQueryGenerator.DruidNodeStack;
+import org.apache.druid.sql.calcite.planner.querygen.SourceDescProducer.SourceDesc;
+
+import java.util.List;
+
+public class SegmentMapStage extends AbstractFrameProcessorStage
```
gianm: Does this existing mean we are creating a stage specifically for applying segment map fns? It seems inefficient, because they could be folded into the next stage. Are you thinking the folding is going to be done with some future work?

Also, does this stage shuffle? I am not quite seeing where that gets set up.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/StageMaker.java
```diff
@@ -91,13 +95,8 @@ private StageDefinitionBuilder buildStageInternal(LogicalStage stage)
...
-    for (LogicalInputSpec dagInputSpec : inputs) {
-      inputSpecs.add(dagInputSpec.toInputSpec(this));
-    }
+    StageDefinitionBuilder sdb = newStageDefinitionBuilder(inputs);
     StageProcessor<?, ?> stageProcessor = frameProcessorStage.buildStageProcessor(this);
-    StageDefinitionBuilder sdb = newStageDefinitionBuilder();
-    sdb.inputs(inputSpecs);
     sdb.signature(frameProcessorStage.getLogicalRowSignature());
     sdb.processor(stageProcessor);
     sdb.shuffleSpec(MixShuffleSpec.instance());
```
gianm: the Mix spec takes everything down to a single partition, which we want to avoid if there might be a lot of data. Another option is to just have _no_ shuffle spec, which means the output partitions are the same as the input partitions. In principle stages with no shuffle spec could be folded into the next stage, although that doesn't happen yet.

Anyway, maybe that makes more sense for the SegmentMapStage?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/stages/SegmentMapStage.java
kgyrtkirk: these stages doesn't necessarily need to match the physical stages - they do right now; but the `StageMaker` part may optimize it to execute better;
caring about those things at this point wouldn't be easy - but in the `StageMaker` the problem becomes to just identify stages which could be joined....

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/StageMaker.java
kgyrtkirk: yes; its not necessarily efficient - it was always planned to enhance the `StageMaker` a bit to identify stages which could be run together... but it was left for later as that should not change anything but that

sql/src/test/java/org/apache/druid/sql/calcite/CalciteNestedDataQueryTest.java
kgyrtkirk: thank you - its fixed now!

sql/src/test/java/org/apache/druid/sql/calcite/CalciteArraysQueryTest.java
kgyrtkirk: I marked the testcase as I think in this case this is not a rounding error;

I think `ARRAY_AGG` should never make a float become a `double`; so I feel like this is a different kind of problem; it also feels like its outside the scope of this PR

sql/src/test/java/org/apache/druid/sql/calcite/CalciteArraysQueryTest.java
kgyrtkirk: these `DD_UNNEST_INLINED` cases are somewhat unrealistic; in this case the `unnest` is not really there - as it just runs thru a direct array

with unrealistic I mean
```
  SELECT * FROM (select ARRAY[1,2,3] v) t, UNNEST(t.v)
```
would work - which is actually more natural use of `unnest`

never wanted to rewrite these testcases; but might make sense to reject `unnest(ARRAY<>)` expressions; or flatten it at compilation time...

by putting the `@NYS`  on these cases I was just postponing to address that...


`DD_UNNEST_RESULT_MISMATCH` is on the otherhand is a different kind of problem; it detects that there are slight differences in semantics sometimes: 
`MV_TO_ARRAY` maps muyltiple things to the same value 

```sql
SELECT label,l_arr,mv_to_array(mv) from larry
new Object[]{"[1]", "[1]", "[\"1\"]"},
new Object[]{"[2,3]", "[2,3]", "[\"2\",\"3\"]"},
new Object[]{"[]", "[]", null},
new Object[]{"[null]", "[null]", null},
new Object[]{"null", null, null}
```
which makes it impossible to handle it correctly; however the old unnest handling had some tricks which make it work differently:
* decoupled respects the `mv_to_array` call and retains it/executes it
* meanwhile [DruidCorrelateUnnestRel#getRexNodeToUnnest](https://github.com/apache/druid/blob/d64a7d4545c01697360140c32eb0d9bb446db5a0/sql/src/main/java/org/apache/druid/sql/calcite/rel/DruidCorrelateUnnestRel.java#L533)  explicitly unwraps the `mv_to_array` from the expression...which makes it look thru those nulls

I believe the decoupled aprroach is the right in this case; but changing the exisitng engine would be a behavioural change; so it was left as-is for now.

added `testMvToArrayUnnest` to the end of `CalciteArrayQueryTest`

sql/src/test/java/org/apache/druid/sql/calcite/CalciteJoinQueryTest.java
kgyrtkirk: yes - iirc this is a bug in the rule itself ; never really got to fixing it...

kgyrtkirk:
@gianm could you please take another look?

sql/src/test/java/org/apache/druid/sql/calcite/CalciteArraysQueryTest.java
gianm: OK, I think we can leave it for this PR, but it definitely looks like an area that needs some cleanup at some point.

sql/src/test/java/org/apache/druid/sql/calcite/CalciteArraysQueryTest.java
gianm: Currently expression evaluation promotes floats to doubles internally because there isn't specialized float-specific logic. We probably won't add it, since it would require lots of new Java code. It's possible this promotion is contributing to what you're seeing, since `ARRAY_AGG` uses expressions internally.

I don't think this promotion is "bad", btw, it's just something that we should try to have consistency around. Maybe you're saying the same thing with the "result mismatch" tag.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/logical/stages/SegmentMapStage.java
gianm: From what you're saying, I take it the current state of this PR is a stepping-stone on the way towards fusing operations together than could be run in the same stage. Sounds good.

gianm (approved):

gianm:
Thanks for clarifying the questions I had; LGTM.

Issue #18198, opened 2025-07-03
MSQ: Missing join on realtime subquery


gianm:
On the wikipedia dataset, this query yields incorrect results on realtime data:

```sql
SELECT COUNT(*) FROM "wikipedia"
WHERE countryName IN (
  SELECT countryName
  FROM "wikipedia"
  WHERE countryName IS NOT NULL
  GROUP BY 1
  ORDER BY COUNT(*) DESC
  LIMIT 1
)
```

The query results are the same as `SELECT COUNT(*) FROM "wikipedia"`. Under the hood this query runs using a broadcast join. It may be that the `join` datasource is not being correctly passed down to the realtime server.

gianm:
Test cases in #18201: https://github.com/apache/druid/pull/18201/files#diff-4232936da3f4238f5cd8100dbbaa72e37534c88c7b515d5b5b3a9b593dd313d1R278-R320

adarshsanjeev:
I looked into this a bit. 
In a way, it is due to the join not being passed to the server. Since the join can be with a stage output etc, the realtime query might not be able to handle doing these operations in all cases.

The join or unnest information is passed in the `mapSegment` function already, so fix might be to call it on the results of the realtime query.

adarshsanjeev:
Actually, thinking more about it, having the results be wrapped in a segment which is then mapped using the function would fix the issue for the `ScanQueryFrameProcessor`, however, for the `GroupByFrameProcessor`, it would require the mapping to be done before the grouping done on the realtime task, so that wouldn't work.

PR #18235, opened 2025-07-11
Push down datasource to realtime tasks

adarshsanjeev:
Pushes down datasources to realtime tasks in MSQ queries.

Resolves https://github.com/apache/druid/issues/18198

<hr>

This PR has:

- [ ] been self-reviewed.


embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQApis.java
```diff
@@ -78,6 +82,31 @@ public String runDartSql(String sql, Object... args)
     ).trim();
   }
 
+  public void runDartWithError(String sql, Matcher<Throwable> matcher, Object... args)
```
gianm: Instead of this method, use `assertThrows` with `runDartSql` in the test case.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQApis.java
```diff
@@ -130,4 +159,33 @@ public MSQTaskReportPayload runTaskSql(String sql, Object... args)
 
     return taskReportPayload;
   }
+
+  public void expectTaskFailure(String sql, String reason, Object... args)
```
gianm: Instead of this, break out the task submit part from `runTaskSql` and call it `submitTaskSql`. It should return `SqlTaskStatus`. The idea is that generally we want the methods to be composable rather than super-specific to a use case. With `submitTaskSql` broken out, the test case could call that followed by `cluster.callApi().waitForTaskToFail` -- or anything else it wants to do.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -116,6 +116,20 @@ public void waitForTaskToSucceed(String taskId, EmbeddedOverlord overlord)
     verifyTaskHasStatus(taskId, TaskStatus.success(taskId));
   }
 
+  /**
+   * Waits for the given task to fail, with the given errorMsq. If the given
+   * {@link EmbeddedOverlord} is not the leader, this method can only return by
+   * throwing an exception upon timeout.
+   */
+  public void waitForTaskToFail(String taskId, String errorMsg, EmbeddedOverlord overlord)
```
gianm: Better for this to take a `Matcher<String>` as the last argument, so it would be called like:

```java
apis.waitForTaskToFail(taskId, overlord, CoreMatchers.equalTo("xyz"));
```

That way any kind of matching can be done, not just equality.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -141,12 +163,38 @@ public EmbeddedDruidCluster createCluster()
...
-        .addServer(historical)
         .addServer(router);
   }
 
+  @BeforeAll
+  @Override
+  protected void setup() throws Exception
+  {
+    cluster = createCluster();
+    cluster.start();
```
gianm: This appears to be copied from the superclass; you can replace it with `super.setup()`

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/DataServerQueryHandler.java
```diff
@@ -195,7 +200,35 @@ public <RowType, QueryType> DataServerQueryResult<RowType> fetchRowsFromDataServ
...
+  private DataSource transformDatasource(DataSource dataSource)
+  {
+    if (dataSource instanceof InputNumberDataSource) {
+      InputNumberDataSource numberDataSource = (InputNumberDataSource) dataSource;
+      if (numberDataSource.getInputNumber() == inputNumber) {
+        return new TableDataSource(dataSourceName);
+      } else {
+        throw DruidException.forPersona(DruidException.Persona.USER)
+                            .ofCategory(DruidException.Category.INVALID_INPUT)
+                            .build("Unknown InputNumberDataSource datasource with number[%s]. Queries with realtime sources "
```
gianm: This is an end-user-facing error message, it shouldn't start with `Unknown InputNumberDataSource datasource with number[1].` because that doesn't mean much to an end-user. The term `stage outputs` isn't very clear either. Please rewrite it to be more useful to the end user. It should describe concepts they would be familiar with from the SQL side, such as mentioning that broadcast joins are not currently supported with realtime data.

Also, could this situation (an `InputNumberDataSource` for some non-base datasource) happen in any case other than joins? If so the error message should reflect that. Maybe unions?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeUnnestQueryTest.java
```diff
@@ -0,0 +1,312 @@
...
+  {
+    final String sql = StringUtils.format(
+        "SET includeSegmentSource = 'REALTIME';\n"
+        + "SELECT d3 FROM \"%s\" CROSS JOIN UNNEST(MV_TO_ARRAY(\"placementish\")) AS d3\n"
+        + "LIMIT 5",
+        dataSource
+    );
+    final MSQTaskReportPayload payload = msqApis.runTaskSql(sql);
+
+    // By default tasks do not include realtime data; count is zero.
```
gianm: comment seems incorrect

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeUnnestQueryTest.java
```diff
@@ -0,0 +1,312 @@
...
+import org.junit.jupiter.api.Timeout;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+public class EmbeddedMSQRealtimeUnnestQueryTest extends EmbeddedClusterTestBase
```
gianm: Now that there are two of these, we should think about how to simplify the test cases and reduce duplicate code. It seems there's a lot of code copied from `EmbeddedMSQRealtimeQueryTest`, with some small differences, like which dataset is being loaded.

The simplest approach would be to make a `BaseRealtimeQueryTest` that lets subclasses define the dataset, and provides all the other stuff. Another approach could be to add helper functions for loading dataset, setting up supervisors, etc, to `EmbeddedClusterApis`. @kfaraz - wondering if you have thoughts on this?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/DataServerQueryHandler.java
```diff
@@ -195,7 +200,35 @@ public <RowType, QueryType> DataServerQueryResult<RowType> fetchRowsFromDataServ
...
+   */
+  private DataSource transformDatasource(DataSource dataSource)
+  {
+    if (dataSource instanceof InputNumberDataSource) {
+      InputNumberDataSource numberDataSource = (InputNumberDataSource) dataSource;
+      if (numberDataSource.getInputNumber() == inputNumber) {
+        return new TableDataSource(dataSourceName);
+      } else {
+        throw DruidException.forPersona(DruidException.Persona.USER)
+                            .ofCategory(DruidException.Category.INVALID_INPUT)
```
gianm: Category `UNSUPPORTED` would be more appropriate I think.

gianm (commented):
The essential fix looks right to me.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
kfaraz: I think `waitForTaskToFail` feels like an odd condition to check anyway.
`waitForTaskToSucceed` is a much more common (and constructive) condition that is required by many tests.

As suggested by @gianm in other comments that we should make things composable,
this method should probably just be `waitForTaskToFinish`.

The verification of the actual task status and error message can be done in the calling test itself using matchers.
If a similar assertion needs to be done in multiple places in the test, I would still prefer that we add a method `waitForTaskToFail` in that test rather than here.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQApis.java
kfaraz: +1

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
kfaraz: Alternatively, you can add a fresh `@BeforeAll` method named `initializeLookupsAndAddServers()` or similar.
The `@BeforeAll` of the super class will always run before this one.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -141,12 +163,38 @@ public EmbeddedDruidCluster createCluster()
...
+        c -> c.updateAllLookups(EmbeddedClusterApis.deserializeJsonToMap(lookupPayload))
+    );
+
+    // Initialize the broker/data-servers later so that lookups are loaded on startup.
+    cluster.addServer(broker);
+    broker.start();
+    cluster.addServer(indexer);
+    indexer.start();
+    cluster.addServer(historical);
+    historical.start();
```
kfaraz: Nit: You could chain up the `addServer` calls.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeUnnestQueryTest.java
kfaraz: +1, 

We should use a `BaseRealtimeQueryTest` to contain the common setup as well as tests, if any.

I would have liked to move the `createKafkaSupervisor()` method (with adequate args) to `EmbeddedClusterApis`, but the latter is in the `druid-services` module.
So, for the time being, we should move it to `Resources.Kafka` or `Resources.Supervisor` similar to `Resources.Task` being added in #18207.

As we migrate old ITs to the embedded framework, the builders and helper functions will take more concrete shape.
e.g. A `TaskBuilder` is being added in [PR #18207](https://github.com/apache/druid/pull/18207/files#diff-75d04ca7e0b706c5dbbad63c1441cbebb6edbf47c988bbcf1cfd2f7a0f4dc559)

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/DataServerQueryHandler.java
adarshsanjeev: Updated the message

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeUnnestQueryTest.java
adarshsanjeev: I have created a base class and moved the common code there. I have kept `createKafkaSupervisor()` there for now, since it . @kfaraz, please let me know if this is good, or if that and the other functions in that class should be pushed to the Kafka resource instead.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQApis.java
```diff
@@ -110,6 +110,19 @@ public MSQTaskReportPayload runTaskSql(String sql, Object... args)
...
+    return taskStatus;
+  }
+
+  /**
+   * Submits the given SQL query to any of the brokers (using {@code BrokerClient})
+   * of the cluster. Waits for it to complete, then returns the query report.
+   *
+   * @return The result of the SQL as a single CSV string.
+   */
+  public MSQTaskReportPayload runTaskSql(String sql, Object... args)
```
kfaraz: Nit: to distinguish this method with the other one `submitTaskSql` better.
```suggestion
  public MSQTaskReportPayload runTaskSqlAndGetReport(String sql, Object... args)
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/BaseRealtimeQueryTest.java
```diff
@@ -0,0 +1,196 @@
...
+  {
+    kafka = new KafkaResource();
+    return EmbeddedDruidCluster
+        .withEmbeddedDerbyAndZookeeper()
+        .addExtension(KafkaIndexTaskModule.class)
+        .addResource(kafka);
+  }
+
+  @BeforeEach
+  void setUp()
```
kfaraz:
```suggestion
  void setupCreateKafkaTopic()
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -138,54 +133,46 @@ public EmbeddedDruidCluster createCluster()
...
         .addServer(coordinator)
         .addServer(overlord)
-        .addServer(indexer)
-        .addServer(broker)
-        .addServer(historical)
         .addServer(router);
   }
 
+  @BeforeAll
+  @Override
```
kfaraz: You need not do both. Either `@Override` the `setup` method or use the `@BeforeAll` annotation on a fresh method, whichever you prefer.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -138,54 +133,46 @@ public EmbeddedDruidCluster createCluster()
...
-                             TestHelper.JSON_MAPPER.writeValueAsBytes(rowMap)
-                         );
-                       }
-                       catch (JsonProcessingException e) {
-                         throw new RuntimeException(e);
-                       }
-                     })
-                     .toList()
-    );
+    dataSource = EmbeddedClusterApis.createTestDatasourceName();
```
kfaraz: I am not sure if this step is really needed, since the `dataSource` field is anyway randomly assigned before each test.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/DataServerQueryHandler.java
```diff
@@ -195,7 +200,36 @@ public <RowType, QueryType> DataServerQueryResult<RowType> fetchRowsFromDataServ
...
+  }
+
+  /**
+   * Transform {@link InputNumberDataSource}s, which are only understood by MSQ tasks, back into a
+   * {@link TableDataSource}.
+   */
+  private DataSource transformDatasource(DataSource dataSource)
+  {
+    if (dataSource instanceof InputNumberDataSource) {
+      InputNumberDataSource numberDataSource = (InputNumberDataSource) dataSource;
```
gianm: This should also have a branch to check for `RestrictedInputNumberDataSource`. If it sees one, create a `RestrictedDataSource` wrapping a `TableDataSource` using the correct `Policy`. Please include a test for this. You don't need to include an end-to-end test; a unit test for this function specifically is OK. To make it more testable, consider making it `static`.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/DataServerQueryHandler.java
```diff
@@ -195,7 +200,36 @@ public <RowType, QueryType> DataServerQueryResult<RowType> fetchRowsFromDataServ
...
+      if (numberDataSource.getInputNumber() == inputNumber) {
+        return new TableDataSource(dataSourceName);
+      } else {
+        throw DruidException.forPersona(DruidException.Persona.USER)
+                            .ofCategory(DruidException.Category.UNSUPPORTED)
+                            .build(
+                                "Cannot handle joining two sources (like unions or broadcast joins) while "
```
gianm: Suggested clearer message:

> Cannot handle stage with multiple sources while querying realtime data. If using broadcast joins, try setting[sqlJoinAlgorithm] to[sortMerge] in your query context.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/DataServerQueryHandlerUtils.java
```diff
@@ -48,17 +56,65 @@ private DataServerQueryHandlerUtils()
...
-    return query.withDataSource(new TableDataSource(dataSource));
+  /**
+   * Transforms {@link InputNumberDataSource} and {@link RestrictedInputNumberDataSource}, which are only understood
+   * by MSQ tasks, back into {@link TableDataSource} and {@link RestrictedDataSource} recursivly.
+   */
+  static DataSource transformDatasource(
+      final DataSource dataSource,
+      final int inputNumber,
+      final String dataSourceName
+  )
```
kgyrtkirk, line 81: seems like this is pretty late to throw out such exceptions; this should be checked at compilation time; can't you identify the shape which should be rejected?

kgyrtkirk:
the summary of the change is ` Push down datasource to realtime tasks` ; however it seems like its a bugfix ?
could you please update the summary?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/DataServerQueryHandlerUtils.java
adarshsanjeev: Yes, this is not the best way to handle it, however, this check would need to be present as a sanity check, and stops the query from returning any incorrect result.

>identify the shape which should be rejected

This is a bit more difficult to do accurately. To be perfectly accurate the error only needs to be thrown if the datasource being queried actually has realtime segments, and this information is not present at compilation time. The alternate is to fail queries which are querying any datasources if they have any broadcast joins or unions on them, but this has a chance to fail queries that would otherwise pass.

Since this is a Druid 34 blocker, I wanted to prevent any incorrect results without causing any regressions by failing it more eagerly. Is there a way to fail it at compile time accurately?

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
gianm: I am good with `waitForTaskToFinish`, that's even simpler and more generally useful. It should return a task status object and not do any assertions on it. The `waitForTaskToSucceed` method should be rewritten to use `waitForTaskToFinish`.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/DataServerQueryHandlerUtils.java
gianm: I think there is an issue with doing this at SQL planning time, in that the query shape is only disallowed if there is realtime data. The shape is allowed if all data is historical. So, the check needs to be done at the point that this is known. The earliest possible place is at the controller.

I do think it would be good to move this check to the controller, since it would satisfy the principle that validations should be done as early as possible. However, I'm not off the top of my head able to say exactly where it would go. If there is a nice place to put the validation in the controller, let's move it there; but if there isn't then I suppose we can keep it here.

adarshsanjeev: I have gone through the flow, and I am not able to find much improvement in terms of validating earlier. 

The only place to put this check seems to be around `WorkerInputs#create`. Checking if any of the slices has any realtime segments and checking if the stage inputs contain a `StageInputSpec`s should be a validation. This is right after we get the segment information to see if there are any realtime segments.

Unfortunately, this seems to be run at the start of a stage, not the start of the query, so this is not much better in terms of validating earlier, since it would still run the earlier stages. It's bit messy to add a validation here as well.

Since we would need to keep the worker validation anyway as a sanity check, I'm not sure if this change should be made.

gianm: Thanks for checking. Given that, I think it's ok to keep the validation in this place. It should hopefully not be a very long lived validation anyway- we do need to end up supporting broadcast joins with realtime tasks.

gianm (approved):

PR #18246, opened 2025-07-14
remove reset-cluster tool

clintropolis:
### Description
This PR removes the `reset-cluster` tool, which was out of date from missing a few metadata tables. It is mainly for testing, has proven error prone to maintain, and seems like it would be hard to ever completely clean stuff up when extensions are considered, so I think we should just remove it.

#### Release note
The `reset-cluster` dev and testing environement oriented tool has been removed. It was already out of date, and would fail to completely clear out the metadata store (missing "upgradeSegments" and "segmentSchemas" tables added in Druid 28 and 30 respectively). This tool was not intended for production use cases, so there should be no impact for cluster operators.


kfaraz (approved):
LGTM

gianm (approved):

kfaraz:
I think we can merge this off as the only jobs failing are the IT ones, which are unrelated to this patch.

cryptoe:
+1 @kfaraz comment. 
I think we should backport this as well in D34 and call it in the release note.

PR #18247, opened 2025-07-14
indexing-service no longer depends on indexing-hadoop

clintropolis:
### Description
This PR makes it so that the `indexing-hadoop` module is no longer a dependency of `indexing-service`. Instead, `indexing-hadoop` depends on `indexing-service`, and the only references to `indexing-hadoop` is the `services` module. 

To validate the changes I ran `ITHadoopIndexTest`, which .. failed because I think this thing hasn't been updated in some time (i tried but gave up), but all of the hadoop ingestion tasks it ran as part of the tests ran and succeeded, just the queries after failed. I saw the same failures without the changes in this PR too (with the dockerfile and json file fixes from this PR), so I think everything is still ok.

The next step to explore after this PR is moving the `indexing-hadoop` module into a contrib extension.


examples/quickstart/tutorial/hadoop/docker/Dockerfile
```diff
@@ -52,12 +52,12 @@ RUN rpm --import http://repos.azulsystems.com/RPM-GPG-KEY-azulsystems && \
...
     yum -q -y update && \
     yum -q -y upgrade && \
-    yum -q -y install zulu17-jdk && \
+    yum -q -y install zulu11-jdk && \
     yum -q -y install nano net-tools telnet less unzip wget && \
     yum clean all && \
     rm -rf /var/cache/yum zulu-repo_${ZULU_REPO_VER}.noarch.rpm
 
-ENV JAVA_HOME=/usr/lib/jvm/zulu17
+ENV JAVA_HOME=/usr/lib/jvm/zulu11
```
kfaraz, line 60: Why the downgrade?

indexing-service/src/main/java/org/apache/druid/indexing/common/config/TaskConfig.java
```diff
@@ -162,7 +157,6 @@ private TaskConfig(
       String baseDir,
       File baseTaskDir,
       String hadoopWorkingPath,
```
kfaraz: Should `defaultHadoopCoordinates` and `hadoopWorkingPath` be removed too?

indexing-hadoop/src/main/java/org/apache/druid/indexer/HadoopIndexTask.java
```diff
@@ -833,7 +831,7 @@ public String runTask(String[] args) throws Exception
       // can be injected based on the configuration given in config.getSchema().getIOConfig().getMetadataUpdateSpec()
       final SegmentMetadataPublisher maybeHandler;
       if (config.isUpdaterJobSpecSet()) {
-        maybeHandler = new SegmentMetadataPublisher(INJECTOR.getInstance(IndexerMetadataStorageCoordinator.class));
+        maybeHandler = new SegmentMetadataPublisher(HadoopTask.INJECTOR.getInstance(IndexerMetadataStorageCoordinator.class));
```
kfaraz: Just saw that `HadoopTask` has its own static `INJECTOR` and `EXTENSION_LOADER`. 😅 
Didn't go through the whole thing but seems like a really hacky approach to load a bunch of classes.

examples/quickstart/tutorial/hadoop/docker/Dockerfile
clintropolis: this is the hadoop dockerfile, it doesn't support java 17

indexing-service/src/main/java/org/apache/druid/indexing/common/config/TaskConfig.java
clintropolis: oops, yes

indexing-hadoop/src/main/java/org/apache/druid/indexer/HadoopIndexTask.java
clintropolis: yea, this isn't new to this PR 😅. i suppose this line change isn't really needed and was just a side-effect of moving stuff around, but will leave it unless i need to push more commits to this branch

kfaraz (approved):
LGTM!

gianm (commented):
Couple of questions:

- Does the injector and classpath dark magic still work properly with this change? I recall there is some gnarly stuff where we're inspecting the classloader, pulling the jars out of it, and using that to populate the Hadoop classpath. (If the Hadoop ITs pass, probably the answer is yes this still works)
- With this change, if someone removes the `druid-hadoop-indexing` jar, does Druid still work properly? If so that is great.

clintropolis:
>Does the injector and classpath dark magic still work properly with this change? I recall there is some gnarly stuff where we're inspecting the classloader, pulling the jars out of it, and using that to populate the Hadoop classpath. (If the Hadoop ITs pass, probably the answer is yes this still works)

Yes, as far as I can tell everything is still working correctly. The Hadoop ITs don't actually pass for me because the query expectations it is doing afterwards are failing, but the ingestion tasks succeed (and i see the same query expectation failures before this change too, so I think it ok).

>With this change, if someone removes the druid-hadoop-indexing jar, does Druid still work properly? If so that is great.

I haven't actually tried this yet, but I think it should be possible after #18246, since the only remaining references are the `HadoopIndexTaskModule` references in `CliOverlord`, `CliMiddleManager`, `CliPeon`, and `CliIndexer` it should be pretty easy to unhook and leave out of the build. I'm actually going to see if I can just go ahead and get it to work as an actual extension after this gets merged, so might be able to do even better so that it can be left out by default.

gianm (approved):

PR #18248, opened 2025-07-14
Use openjdk:17-slim as base image for ITs

kfaraz:
Debian Buster has reached EOL causing ITs to fail.
Trying out a newer stable image here.

### DIFF
--- a/integration-tests-ex/image/docker/Dockerfile
+++ b/integration-tests-ex/image/docker/Dockerfile
@@ -28,7 +28,7 @@
 # This Dockerfile prefers to use the COPY command over ADD.
 # See: https://phoenixnap.com/kb/docker-add-vs-copy
 
-ARG JDK_VERSION=17-slim-bookworm
+ARG JDK_VERSION=17-slim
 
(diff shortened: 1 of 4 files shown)


kfaraz:
this is odd, it is still picking up the wrong image

```
#5 [druidbase 1/3] FROM docker.io/library/openjdk:17-slim-buster@sha256:2e7658fb62d1c6f319ff9870614deaf8e06038dd41eec3d1ecdcfabd186234fd
```

gianm:
There is this error in one of the runs: `0.492 /bin/sh: 1: /etc/init.d/mysql: not found`. Looking at the image, the file is now in `/etc/init.d/mariadb`. @kfaraz - I tried pushing a change to your branch that updates the scripts to use this new file.

kfaraz:
Thanks, @gianm ! I think @cecemei already has a working fix here, https://github.com/apache/druid/pull/18251.
I think we can merge that change and just close off this PR.

cryptoe:
Yes just merged : #18251

PR #18249, opened 2025-07-14
Fix concurrency issues with StubServiceEmitter

kgyrtkirk:
* concurrency issue with StubServiceEmitter - #18121 have added a few new metrics which have increased the load on it and have caused  random appearances of an issue arising from the fact it used an `ArrayList` under the hood.
* added some catches to shut down queries properly in case some unexpected exceptions occur - this could give better exceptions and reduce time to fix in the future

this should reduce the probability that `QTest` splits remain hanging

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
         <repoOrgUrl>https://repo.maven.apache.org/maven2/</repoOrgUrl>
 
         <!-- Allow the handful of flaky tests with transient failures to pass. -->
-        <surefire.rerunFailingTestsCount>3</surefire.rerunFailingTestsCount>
+        <surefire.rerunFailingTestsCount>0</surefire.rerunFailingTestsCount>
         <surefire.trimStackTrace>false</surefire.trimStackTrace>
 
         <!-- Using -DskipTests or -P skip-tests will skip both the unit tests and
@@ -1094,7 +1094,7 @@
             <dependency>
                 <groupId>org.junit</groupId>
                 <artifactId>junit-bom</artifactId>
-                <version>5.10.2</version>
+                <version>5.13.3</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
         <repoOrgUrl>https://repo.maven.apache.org/maven2/</repoOrgUrl>
 
         <!-- Allow the handful of flaky tests with transient failures to pass. -->
-        <surefire.rerunFailingTestsCount>0</surefire.rerunFailingTestsCount>
+        <surefire.rerunFailingTestsCount>3</surefire.rerunFailingTestsCount>
         <surefire.trimStackTrace>false</surefire.trimStackTrace>
 
         <!-- Using -DskipTests or -P skip-tests will skip both the unit tests and
--- a/processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
+++ b/processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
@@ -24,11 +24,14 @@ import org.apache.druid.java.util.emitter.service.AlertEvent;
 import org.apache.druid.java.util.emitter.service.ServiceEmitter;
 import org.apache.druid.java.util.emitter.service.ServiceMetricEvent;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
 
 /**
  * Test implementation of {@link ServiceEmitter} that collects emitted metrics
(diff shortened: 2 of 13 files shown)


kgyrtkirk:
removed the jfr related enhancements as it was causing some OOM-s for the embedded tests

pom.xml
```diff
@@ -1094,7 +1094,7 @@
             <dependency>
                 <groupId>org.junit</groupId>
                 <artifactId>junit-bom</artifactId>
-                <version>5.10.2</version>
+                <version>5.13.3</version>
```
clintropolis, line 1097: is this related to anything or just opportunistic?

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
```diff
@@ -68,15 +71,15 @@ public void emit(Event event)
...
 
   /**
    * Gets all the metric events emitted since the previous {@link #flush()}.
    *
    * @return Map from metric name to list of events emitted for that metric.
    */
-  public Map<String, List<ServiceMetricEventSnapshot>> getMetricEvents()
+  public Map<String, Queue<ServiceMetricEventSnapshot>> getMetricEvents()
   {
     return metricEvents;
```
kfaraz, line 84: Nit: I guess we can later update this method to also return a map containing read-only lists rather than the `Queue` or maybe just update the method to accept a metric name.

pom.xml
kgyrtkirk: not closely related; just in an earlier version I've tried out a junit 5.12.x feature....then undone it - but retained the bom upgrade as it might still be usefull later (didn't made things worse)

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
kgyrtkirk: yes; some enhancements could be made there - actually in the latchingemitter even after this patch an assimptotically `n^2` computation remained (over time)

I already had to undo a jfr fix in this patch as it was causing some other issues - and I'm afraid this won't fully solve the problem ; sometimes I could still reproduce the issue...with a lower probability....but its still there.

seems like recently it happens on apache/master more frequently - I have to collect more details

kfaraz: > actually in the latchingemitter even after this patch an assimptotically n^2 computation remained (over time)

What is the n^2 computation, is it the evaluation of events against a given WaitCondition?
I had to tried to keep it O(n), might have missed some case.

kgyrtkirk: it `O(n)` but for every new event ; which is technically `n^2` over time

kfaraz: Must be a bug then, because every `WaitCondition` tracks a `processedUntil` index.
And only processes new items after that.

kfaraz: I guess now it might become `n^2` though since we are making a copy of `events`.

kfaraz: If that is the case, we should expose another method in `StubServiceEmitter` to access the original events queue so that a copy is not needed. Alternatively, we can just make `events` field protected so that `LatchableEmitter` can access it.

kgyrtkirk: I don't think it's a good pattern to expose internals. I don't think it should need this readuntil pattern

kfaraz: Well, it is just for tests and not really exposed outside the class, only for use by the sub-class.

The read until is actually very useful to process only the stuff that we need to process.

Otherwise, we would need to pipe every new event through each existing condition either sync or async.
But that would still not handle the case where a condition is registered later.
I guess we could maintain a list of historical events for that purpose.

But all of that just seems overkill for a simple testing use case.

Please let me know if you have some other suggestions to improve this flow.

kgyrtkirk: reading pre-existing events registrations seems dodgy to me; I feel like that's an artifact of the fact that the usage pattern is like:
```
doSomething();
waitSomething();
```
which could have trouble if `doSomething` have already happened before `wait` could even launche....the "read older events" is a try to fix this...however the problem arises from the fact that the pattern itself is flawed...

it should be more like
```
w = registerWait()
doSomething()
wait(w)
```

or possibly with some try-with-resources sugar
```
try( waitForSomething() ) {
  doSomething();
}
```

so that the contract of `waitForSomething` would be:
* only reads messages which are recieved after its registered
* no need to track read point/redispatch old events  -  dispatch only the event caught during the `emit` to the current readers
* it won't even need access to how the events are stored by the `StubEmitter`

not sure if this would apply to all existing use cases - but I hope so...

kfaraz: Yeah, I would have liked it to be something like that. But it would make the embedded tests (where the `LatchableEmitter` is actually used) much less readable.

kfaraz: The original flow actually feels more intuitive to me.
We are essentially registering a consumer for the stream of metric events coming in.
And the last read point is just the offset that we have consumed so far.

(now that the `events` field is a queue, this is even more applicable)

PR #18250, opened 2025-07-14
Fix range filter inconsistency

### LINKED ISSUE #18205: RangeFilter / EqualityFilter match inconsistency for float values when comparision happens in double
For a value like `0.1` the best float / double representation is different.

right now:
* `RangeFilter` doesn't match in these cases
* `EqualityFilter` does match

`Filtration` rewrites single-value `RangeFilter`  to `EqualityFilter` which makes this issue in most cases absent.

`RangeFilter` works like:
```
      assertFilterMatches(
          new RangeFilter("d0", ColumnType.FLOAT, 60.0f, 60.0f, false, false, null),
          ImmutableList.of("4")
      );
      assertFilterMatches(
          new RangeFilter("d0", ColumnType.DOUBLE, 60.0f, 60.0f, false, false, null),
          ImmutableList.of("4")
      );
      assertFilterMatches(
          new RangeFilter("f0", ColumnType.FLOAT, 10.1f, 10.1f, false, false, null),
          ImmutableList.of("1")
      );
      assertFilterMatches(
          new RangeFilter("f0", ColumnType.DOUBLE, 10.1, 10.1, false, false, null),
          ImmutableList.of()
      );
```

meanwhile for `EqualityFilter` matches:
```
      assertFilterMatches(new EqualityFilter("f0", ColumnType.FLOAT, 10.1, null), ImmutableList.of("1"));
      assertFilterMatches(new EqualityFilter("f0", ColumnType.DOUBLE, 10.1, null), ImmutableList.of("1"));
```

nozjkoitop:
Fixes #18205.

### Description

Bitmap index skipped for a closed single-value bound, to trigger the predicate matching where this case is covered now.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/query/filter/RangeFilter.java
+++ b/processing/src/main/java/org/apache/druid/query/filter/RangeFilter.java
@@ -952,7 +952,7 @@ public class RangeFilter extends AbstractOptimizableDimFilter implements Filter
       case CLOSED:
         if (Double.compare(lowerDoubleBound, upperDoubleBound) == 0) {
           final long dblBits = Double.doubleToLongBits(lowerDoubleBound);
-          final int  fltBits = Float.floatToIntBits((float) lowerDoubleBound);
+          final int fltBits = Float.floatToIntBits((float) lowerDoubleBound);
 
           return input -> {
             long inDblBits = Double.doubleToLongBits(input);


processing/src/main/java/org/apache/druid/query/filter/RangeFilter.java
```diff
@@ -926,10 +950,25 @@ public static DruidDoublePredicate makeDoublePredicate(
...
+          final int fltBits = Float.floatToIntBits((float) lowerDoubleBound);
+
+          return input -> {
+            long inDblBits = Double.doubleToLongBits(input);
+            if (inDblBits == dblBits) {
+              return DruidPredicateMatch.TRUE;
+            }
+            return DruidPredicateMatch.of(
+                Float.floatToIntBits((float) input) == fltBits
+            );
```
kgyrtkirk, line 964: does this mean that if `(Double.compare(lowerDoubleBound, upperDoubleBound) == 0` is satisfied ; the predicates is weakened down to `float` level even when `input` is a `double` ?

cc: @clintropolis

kgyrtkirk:
thank you @nozjkoitop for looking into this! :)

processing/src/main/java/org/apache/druid/query/filter/RangeFilter.java
nozjkoitop: The idea is to check the inDblBits == dblBits before, and fallback to the float casting, however it brings up the problem with false positive results for a really precise doubles.

The other solution i see atm, is to try creating a separate flow for the floats

cecemei: Comparing `float` and `double` values is a well known challenge due to their binary representation. The current change appears to be quite limited, it only addresses scenarios where the upper and lower bounds are exactly equal. But what about cases where the upper bound is just slightly greater than the lower bound, such as `new RangeFilter("d0", ColumnType.FLOAT, 120.0244f, 120.0245f, false, false, null)`?

processing/src/test/java/org/apache/druid/segment/filter/RangeFilterTests.java
```diff
@@ -454,15 +454,19 @@ public void testNumericMatchExactlySingleValue()
       );
       assertFilterMatches(
           new RangeFilter("d0", ColumnType.FLOAT, 120.0245f, 120.0245f, false, false, null),
-          ImmutableList.of()
+          ImmutableList.of("3")
```
cecemei, line 457: This test case feels a bit unclear to me. It appears to demonstrate the behavior of comparing `double` and `float` values, but doesn't provide guidance on the correct way to perform such comparisons. When comparing `float` and `double`, it's best to use a tolerance range instead of exact value. For example:
`new RangeFilter("d0", ColumnType.DOUBLE, 120.0243f, 120.0246f, false, false, null)`.

PR #18254, opened 2025-07-15
Fixing deadlock in MSQ worker retry logic.

cryptoe:
The main controller thread in the MSQ controller could block if a task failed while the worker ares launching. 
A dead lock was created where retry event is added in the queue of the controller thread but since the controller thread is blocked, it never really consumes the event. 

This patch adds a failure handler which can be executed whenever a task fails and queue it for retry if applicable. 

The larger fix would be to make contacting the workers, non blocking.

### DIFF
--- a/extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
+++ b/extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
@@ -281,8 +281,8 @@ public class ControllerImpl implements Controller
 
   private WorkerSketchFetcher workerSketchFetcher;
 
-  // WorkerNumber -> WorkOrders which need to be retried and our determined by the controller.
-  // Map is always populated in the main controller thread by addToRetryQueue, and pruned in retryFailedTasks.
+  // WorkerNumber -> WorkOrders which need to be retried and are determined by the controller.
+  // Map is always populated in the main controller thread by addToRetryQueue and pruned in retryFailedTasks.
   private final Map<Integer, Set<WorkOrder>> workOrdersToRetry = new HashMap<>();
 
   // Time at which the query started.
@@ -427,7 +427,7 @@ public class ControllerImpl implements Controller
       exceptionEncountered = e;
     }
 
-    // Fetch final counters in separate try, in case runQueryUntilDone threw an exception.
+    // Fetch final counters in a separate try, in case runQueryUntilDone threw an exception.
     try {
       countersSnapshot = getFinalCountersSnapshot(queryKernel);
     }
@@ -746,13 +746,11 @@ public class ControllerImpl implements Controller
         querySpec,
         queryKernelConfig,
         (failedTask, fault) -> {
-          if (queryKernelConfig.isFaultTolerant() && ControllerQueryKernel.isRetriableFault(fault)) {
-            addToKernelManipulationQueue(kernel -> {
-              addToRetryQueue(kernel, failedTask.getWorkerNumber(), fault);
-            });
-          } else {
-            throw new MSQException(fault);
-          }
+          throwIfNonRetriableFault(fault);
+          // since this is called from the task launcher thread, we need to add it to the kernel manipulation queue so that only the controller thread can manipulate the kernel.
+          addToKernelManipulationQueue(kernel -> {
+            addToRetryQueue(kernel, failedTask.getWorkerNumber(), fault);
+          });
         }
     );
 
@@ -1379,13 +1377,20 @@ public class ControllerImpl implements Controller
       final boolean retryOnFailure
   )
   {
-    // Sorted copy of target worker numbers to ensure consistent iteration order.
+    // Sorted copy of target worker numbers to ensure a consistent iteration order.
     final List<Integer> workersCopy = Ordering.natural().sortedCopy(workers);
     final List<String> workerIds = getWorkerIds();
     final List<ListenableFuture<Void>> workerFutures = new ArrayList<>(workersCopy.size());
 
     try {
-      workerManager.waitForWorkers(workers);
+      workerManager.waitForWorkers(
+          workers,
+          (workerTask, fault) -> {
+            throwIfNonRetriableFault(fault);
+            // no need to add it to the kernel manipulation queue since this is the main controller thread calling this function.
+            addToRetryQueue(queryKernel, workerTask.getWorkerNumber(), fault);
+          }
+      );
     }
     catch (InterruptedException e) {
       Thread.currentThread().interrupt();
@@ -2368,7 +2373,14 @@ public class ControllerImpl implements Controller
       }
 
       // wait till the workers identified above are fully ready
-      workerManager.waitForWorkers(workersNeedToBeFullyStarted);
+      workerManager.waitForWorkers(
+          workersNeedToBeFullyStarted,
+          (workerTask, fault) -> {
+            throwIfNonRetriableFault(fault);
+            // no need to add it to the kernel manipulation queue since this is the main controller thread calling this function.
+            addToRetryQueue(queryKernel, workerTask.getWorkerNumber(), fault);
+          }
+      );
 
       for (Map.Entry<StageId, Map<Integer, WorkOrder>> stageWorkOrders : stageWorkerOrders.entrySet()) {
         contactWorkersForStage(
@@ -2439,7 +2451,7 @@ public class ControllerImpl implements Controller
       workerTaskLauncherFuture.addListener(
           () ->
               addToKernelManipulationQueue(queryKernel -> {
-                // Throw an exception in the main loop, if anything went wrong.
+                // Throw an exception in the main loop if anything went wrong.
                 FutureUtils.getUncheckedImmediately(workerTaskLauncherFuture);
               }),
           Execs.directExecutor()
@@ -2529,7 +2541,7 @@ public class ControllerImpl implements Controller
         );
 
         for (final StageId stageId : newStageIds) {
-          // Allocate segments, if this is the final stage of an ingestion.
+          // Allocate segments if this is the final stage of ingestion.
           if (MSQControllerTask.isIngestion(querySpec)
               && stageId.getStageNumber() == queryDef.getFinalStageDefinition().getStageNumber()
               && (((DataSourceMSQDestination) querySpec.getDestination()).getTerminalStageSpec() instanceof SegmentGenerationStageSpec)) {
@@ -2547,7 +2559,14 @@ public class ControllerImpl implements Controller
               stageDef.doesShuffle() ? stageDef.getShuffleSpec().kind() : "none"
           );
 
-          workerManager.launchWorkersIfNeeded(workerCount);
+          workerManager.launchWorkersIfNeeded(
+              workerCount,
+              (workerTask, fault) -> {
+                throwIfNonRetriableFault(fault);
+                // no need to add it to the kernel manipulation queue since this is the main controller thread calling this function.
+                addToRetryQueue(this.queryKernel, workerTask.getWorkerNumber(), fault);
+              }
+          );
           stageRuntimesForLiveReports.put(stageId.getStageNumber(), new Interval(DateTimes.nowUtc(), DateTimes.MAX));
           startWorkForStage(queryDef, queryKernel, stageId.getStageNumber(), segmentsToGenerate);
         }
@@ -2845,6 +2864,13 @@ public class ControllerImpl implements Controller
           }
         }
       }
+    }
+  }
+
+  private void throwIfNonRetriableFault(MSQFault fault)
+  {
+    if (!queryKernelConfig.isFaultTolerant() || !ControllerQueryKernel.isRetriableFault(fault)) {
+      throw new MSQException(fault);
     }
   }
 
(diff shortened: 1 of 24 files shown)


extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/WorkerManager.java
```diff
@@ -48,21 +49,22 @@ public interface WorkerManager
    * Launch additional workers, if needed, to bring the number of running workers up to {@code workerCount}.
    * Blocks until the requested workers are launched. If enough workers are already running, this method does nothing.
    */
-  void launchWorkersIfNeeded(int workerCount) throws InterruptedException;
+  void launchWorkersIfNeeded(int workerCoun, @Nullable WorkerFailureListener failureListener) throws InterruptedException;
```
adarshsanjeev: nit: workerCount

adarshsanjeev (approved):
Approving it for now, since this is a bugfix that might be required, and the changes do seem correct.

I wonder if there is a better way to do this fix, since passing the function at every callsite seems untidy.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
```diff
@@ -1379,13 +1377,20 @@ private void contactWorkersForStage(
...
     final List<ListenableFuture<Void>> workerFutures = new ArrayList<>(workersCopy.size());
 
     try {
-      workerManager.waitForWorkers(workers);
+      workerManager.waitForWorkers(
+          workers,
+          (workerTask, fault) -> {
+            throwIfNonRetriableFault(fault);
+            // no need to add it to the kernel manipulation queue since this is the main controller thread calling this function.
+            addToRetryQueue(queryKernel, workerTask.getWorkerNumber(), fault);
```
gianm: Passing different listeners to different calls seems really hacky. Is there another way to do it? It would be nicer to use the same `WorkerFailureListener` that was registered on `newWorkerManager`.

Also, I'm not clear on how this helps. It seems the main change is to call `addToRetryQueue` immediately, rather than on the next controller loop. How does that fix the deadlock? (I mean: what was the main controller thread blocked on, and how does calling `addToRetryQueue` immediately help it?)

gianm (commented):
This seems pretty hacky and I am wonder if we can make it less hacky.

With regard to this comment, btw:

> The larger fix would be to make contacting the workers, non blocking.

What are you referring to exactly? The `WorkerClient` is already nonblocking- all calls return futures.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
cryptoe: There are 2 blocking calls in the `workerManager` which is called by the main controller thread : 

```
>  workerManager.launchWorkersIfNeeded(
>  workerManager.waitForWorkers(
```

In case the worker spawned fails for any reason , the worker manager thread calls the failure listener : https://github.com/apache/druid/blob/aa29d43eac515905836a9641eaa83f074e2cc8c0/extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQWorkerTaskLauncher.java#L615 which adds the retry event in the controller kernel queue. 

Since we need access to work orders, we can only access that information via the controller thread due to the single threaded nature of the query kernel. 

Since the main controller thread is blocked on `launchingTheWorker`,` waitingForWorker` to launch, we never really launch the worker since the retry event is never processed. 



Let me try to find a way to pass the failure listener as part of the constructor

cryptoe:
@gianm Was able to clean up the interfaces a bit. 

Now made the failure handler smarter : 

```
    return (failedTask, fault) -> {
      throwIfNonRetriableFault(fault);
      if (Thread.currentThread().getId() == mainThreadId.get()) {
        // this is called from the main controller thread, so we can directly access the kernel.
        addToRetryQueue(controllerQueryKernel, failedTask.getWorkerNumber(), fault);
      } else {
        // since this is called from the task launcher thread, we need to add it to the kernel manipulation queue so that only the controller thread can manipulate the kernel.
        addToKernelManipulationQueue(kernel -> {
          addToRetryQueue(kernel, failedTask.getWorkerNumber(), fault);
        });
      }
    };
```
    
That way it knows the caller and adjusts if we wanna add the failure condition in the retry queue or not.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
gianm: Is the scenario something like this?

- Worker fails at some point during a controller run (while `runInternal` is in the midst of executing)
- `MSQWorkerTaskLauncher` notices the failure in its main loop, which runs in another thread
- `MSQWorkerTaskLauncher` removes the worker from `fullyStartedTasks` and calls `WorkerFailureListener#onFailure`. This queues up `addToRetryQueue` to run in the controller thread
- Controller calls `workerManager.waitForWorkers` during that same `runInternal` run
- The call to `workerManager.waitForWorkers` will never return because the `onFailure` listener has not actually run yet, it's still just queued up. It is waiting for a worker that has already failed and will never be relaunched

And I believe your fix makes it so the controller call to `workerManager.waitForWorkers` actually calls into `addToRetryQueue` (through `WorkerFailureListener#onFailure`). That call to `addToRetryQueue` causes a relaunch of the failed worker, so `workerManager.waitForWorkers` can eventually return normally.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
```diff
@@ -1379,13 +1392,15 @@ private void contactWorkersForStage(
...
   {
-    // Sorted copy of target worker numbers to ensure consistent iteration order.
+    // Sorted copy of target worker numbers to ensure a consistent iteration order.
     final List<Integer> workersCopy = Ordering.natural().sortedCopy(workers);
     final List<String> workerIds = getWorkerIds();
     final List<ListenableFuture<Void>> workerFutures = new ArrayList<>(workersCopy.size());
 
     try {
-      workerManager.waitForWorkers(workers);
+      workerManager.waitForWorkers(
```
gianm: no reason to reformat

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
```diff
@@ -2368,7 +2383,9 @@ private void retryFailedTasks() throws InterruptedException
       }
 
       // wait till the workers identified above are fully ready
-      workerManager.waitForWorkers(workersNeedToBeFullyStarted);
+      workerManager.waitForWorkers(
```
gianm: no reason to reformat

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
```diff
@@ -800,7 +813,7 @@ private void addToRetryQueue(ControllerQueryKernel kernel, int worker, MSQFault
 
     List<WorkOrder> retriableWorkOrders = kernel.getWorkInCaseWorkerEligibleForRetryElseThrow(worker, fault);
     if (!retriableWorkOrders.isEmpty()) {
-      log.info("Submitting worker[%s] for relaunch because of fault[%s]", worker, fault);
+      log.debug("Submitting worker[%s] for relaunch because of fault[%s]", worker, fault);
```
gianm, line 815: Is this intentional in order to reduce log spam?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
```diff
@@ -2547,7 +2564,9 @@ private void startStages() throws IOException, InterruptedException
               stageDef.doesShuffle() ? stageDef.getShuffleSpec().kind() : "none"
           );
 
-          workerManager.launchWorkersIfNeeded(workerCount);
+          workerManager.launchWorkersIfNeeded(
```
gianm: no reason to reformat

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQWorkerTaskLauncher.java
```diff
@@ -167,14 +167,17 @@ public MSQWorkerTaskLauncher(
...
     this.maxTaskStartDelayMillis = maxTaskStartDelayMillis;
     this.config = config;
   }
 
   @Override
-  public ListenableFuture<?> start()
+  public ListenableFuture<?> start(WorkerFailureListener workerFailureListener)
   {
+    if (!this.workerFailureListenerRef.compareAndSet(null, workerFailureListener)) {
+      log.warn("WorkerFailureListener already set. Ignoring re-registration call");
```
gianm: Is this expected to ever happen? Seems sketchy to me and I'm wondering if this should be a defensive exception rather than a warning.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQWorkerTaskLauncher.java
```diff
@@ -301,14 +318,31 @@ public void reportFailedInactiveWorker(int workerNumber)
...
         if (stopFuture.isDone() || stopFuture.isCancelled()) {
           FutureUtils.getUnchecked(stopFuture, false);
           throw new ISE("Stopped");
         }
+
+        WorkerFailureListener workerFailureListener = workerFailureListenerRef.get();
+        if (workerFailureListener != null) {
+          for (TaskTracker taskTracker : taskTrackers.values()) {
+            if (taskTracker.isRetrying() && workerNumbers.contains(taskTracker.workerNumber)) {
+              workerFailureListener.onFailure(
```
gianm: Can this use `invokeFailureListener`?

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQWorkerTaskLauncher.java
```diff
@@ -244,22 +246,37 @@ public List<String> getWorkerIds()
...
           FutureUtils.getUnchecked(stopFuture, false);
           throw new ISE("Stopped");
         }
-
+        // add failed tasks to retry the queue
+        WorkerFailureListener workerFailureListener = workerFailureListenerRef.get();
+        if (workerFailureListener != null) {
+          for (TaskTracker taskTracker : taskTrackers.values()) {
+            if (taskTracker.isRetrying()) {
+              workerFailureListener.onFailure(
```
gianm: Can this use `invokeFailureListener`? It would be nice to run them all through there to limit verbosity.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQWorkerTaskLauncher.java
cryptoe: This is never expected to happen. Will adjust accordingly.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
cryptoe: Yes. Now we try to retry more often, this was causing spam.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQWorkerTaskLauncher.java
cryptoe: I wanted to use the same workerFailureListener for all iterations but since now we are throwing an exception in the start method which mandates only having one failureListener, yes we can use the invokeFailureListener. Adjusted the code.

cryptoe (commented):
Thanks for the review.

extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerImpl.java
cryptoe: Yes this is exactly what is happening.

cryptoe:
> The logic looks good to me (assuming I understand the scenario correctly: [#18254 (comment)](https://github.com/apache/druid/pull/18254#discussion_r2216955916)). However it feels like it's getting overly complicated. IMO, it would be cleaner to restore the original listener behavior, and to deal with the `waitForWorkers` / `launchWorkersIfNeeded` situation a different way:
> 
>     * Return an `IntSet` of failed workers from both
> 
>     * In the controller, if the `IntSet` is nonempty, call `addToRetryQueue` and then re-call the original method.
> 
> 
> Something like this in the controller:
> 
> ```java
> IntSet failedWorkers;
> while (!(failedWorkers = workerManager.waitForWorkers(workers)).isEmpty()) {
>   for (int worker : failedWorkers) {
>     addToRetryQueue(kernel, worker, fault);
>   }
> }
> ```
> 
> That way we won't have to have this thing where there needs to be a same-thread callback in `waitForWorkers` / `launchWorkersIfNeeded`.
> 
> Approving since the current code is acceptable, but do consider the review comments. And if you agree with the alternate structure above, feel free to either implement it or at least add a comment to `getWorkerFailureListener` describing this future possible change.

I can adjust the calling sites to add to retry queue as well like you mentioned. Will raise a follow up PR after this to unblock the release.

PR #17170, opened 2024-09-26
Restore Sink Metric Emission Behaviour: Emit them per-Sink instead of per-FireHydrant

findingrish:
This change https://github.com/apache/druid/pull/15757 to merge FireHydrants flatly for realtime queries for optimising memory usage led to metrics like `query/segment/time` getting emitted per-FireHydrant instead of per-Sink. 

This change restores the metric emission behaviour while keeping the optimisation intact. 

It introduces a new `SinkMetricsEmittingQueryRunner` which accumulates the FireHydrants metrics per-Sink and emits them in the end. 
The emitted metrics are `query/segment/time`, `query/segmentAndCacheTime` & `query/wait/time`. 
`query/wait/time` is the time taken to start processing the first FireHydrant for the sink. 

#### Testing  
- Added UT to verify that each of the above metrics are emitted once per Sink. 
- Locally verified the changes for realtime ingestion. Sample emitted metric, 
```
{"type":"scan","version":"32.0.0-SNAPSHOT","duration":"PT86400S","feed":"metrics","metric":"query/segment/time","hasFilters":"false","service":"druid/middleManager","segment":"kttm_2019-08-25T04:00:00.000Z_2019-08-25T05:00:00.000Z_2024-09-26T15:28:40.091Z","host":"localhost:8100","context":{"defaultTimeout":300000,"finalize":false,"maxQueuedBytes":5242880,"maxScatterGatherBytes":9223372036854775807,"queryFailTime":1727364858115,"queryId":"8dab0179-ec39-4f5f-a9e2-9039da416a5a","queryResourceId":"14c72b21-2dbb-471b-91c0-d7fbe11414b5","scanOutermost":false,"sqlOuterLimit":1001,"sqlQueryId":"8dab0179-ec39-4f5f-a9e2-9039da416a5a","sqlStringifyArrays":false,"timeout":299997},"interval":["2019-08-25T00:00:00.000Z/2019-08-26T00:00:00.000Z"],"id":"8dab0179-ec39-4f5f-a9e2-9039da416a5a","value":0,"dataSource":"kttm","timestamp":"2024-09-26T15:29:18.215Z"}
```

PR #18255, opened 2025-07-15
Minor clean up of LatchableEmitter / StubServiceEmitter

kfaraz:
Follow up to #18249 [discussion](https://github.com/apache/druid/pull/18249#discussion_r2206525247)

### Changes

- Maintain a List of processed events in `LatchableEmitter`.
This is an improvement over the current flow where a copy of events is created upon receiving every new event.
- When a new condition is registered, evaluate all past events upfront, then add it to the set of wait conditions
- Evaluate each new event as it is received

### Other changes

- Hide the internal queue implementation of `StubServiceEmitter` from both tests and the sub-class `LatchableEmitter`
- Reduce the usage of `StubServiceEmitter.getEvents()`. Use the inbuilt `verifyValue` methods instead.


kfaraz:
@kgyrtkirk , I have tried to clean up the `LatchableEmitter` since it seemed wasteful to create a new list every time.

I have also removed the `processedUntil` flag to simplify the flow, while retaining readability.
The `LatchableEmitter` is now agnostic of how `StubServiceEmitter` stores events.

Please let me know if this makes sense.

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
```diff
@@ -74,14 +74,20 @@ public List<Event> getEvents()
...
+
   /**
-   * Gets all the metric events emitted since the previous {@link #flush()}.
+   * Gets all the metric events emitted for the given metric name since the previous {@link #flush()}.
    *
-   * @return Map from metric name to list of events emitted for that metric.
+   * @return List of events emitted for the given metric.
    */
-  public Map<String, Queue<ServiceMetricEventSnapshot>> getMetricEvents()
+  public List<ServiceMetricEventSnapshot> getMetricEvents(String metricName)
```
kgyrtkirk: this is exactly what I should have done in the previous PR....but rushed to get it running....

services/src/test/java/org/apache/druid/server/AsyncQueryForwardingServletTest.java
```diff
@@ -683,7 +683,8 @@ protected void doService(
     }
     catch (NullPointerException ignored) {
     }
-    Assert.assertEquals("query/time", stubServiceEmitter.getEvents().get(0).toMap().get("metric"));
+    // Assert.assertEquals("query/time", stubServiceEmitter.getEvents().get(0).toMap().get("metric"));
```
kgyrtkirk, line 686: nit: commented old code

server/src/test/java/org/apache/druid/server/ClientQuerySegmentWalkerTest.java
```diff
@@ -987,10 +986,11 @@ public void testMetricsWithMaxSubqueryRowsEnabled()
         ImmutableList.of(new Object[]{Intervals.ETERNITY.getStartMillis(), 3L})
     );
 
-    List<Event> events = emitter.getEvents();
+    List<StubServiceEmitter.ServiceMetricEventSnapshot> events =
+        emitter.getMetricEvents(ClientQuerySegmentWalker.ROWS_COUNT_METRIC);
 
-    for (Event event : events) {
-      EventMap map = event.toMap();
+    for (StubServiceEmitter.ServiceMetricEventSnapshot event : events) {
```
kgyrtkirk: nit: right now there is `event.getUserDims()` and `event.getMetricEvent().getUserDims()`  ; I have no clear understanding why this `ServiceMetricEventSnapshot` is necessary ; but if the `userDims` or some other things might be clobbered externally; it would be more straight to just add a copy constructor to `ServiceMetricEvent` -

server/src/test/java/org/apache/druid/server/ClientQuerySegmentWalkerTest.java
```diff
@@ -1038,23 +1038,26 @@ public void testMetricsWithMaxSubqueryBytesEnabled()
         ImmutableList.of(new Object[]{Intervals.ETERNITY.getStartMillis(), 3L})
     );
 
-    List<Event> events = emitter.getEvents();
+    List<StubServiceEmitter.ServiceMetricEventSnapshot> events
+        = emitter.getMetricEvents(ClientQuerySegmentWalker.ROWS_COUNT_METRIC);
```
kgyrtkirk: nit: could be inlined into the `for`

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
```diff
@@ -60,7 +60,13 @@ public class LatchableEmitter extends StubServiceEmitter
...
   private final ScheduledExecutorService conditionEvaluateExecutor;
   private final Set<WaitCondition> waitConditions = new HashSet<>();
-  private final ReentrantReadWriteLock eventReadWriteLock = new ReentrantReadWriteLock(true);
+
+  private final ReentrantLock eventProcessingLock = new ReentrantLock();
+
+  /**
+   * Lists of events that have already been processed by {@link #evaluateWaitConditions(Event)}.
+   */
+  private final List<Event> processedEvents = new ArrayList<>();
```
kgyrtkirk, line 63: no `clear` is called on this field - even during `flush`
could that cause problems?

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
```diff
@@ -158,48 +144,64 @@ public void waitForEventAggregate(
...
+   * {@link #waitConditions}.
+   */
+  private void registerWaitCondition(WaitCondition condition)
+  {
+    eventProcessingLock.lock();
+    try {
+      for (Event event : processedEvents) {
+        if (condition.predicate.test(event)) {
+          condition.countDownLatch.countDown();
+        }
```
kgyrtkirk: nit: shouldn't this be `condition#evaluate` ?

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
```diff
@@ -158,48 +144,64 @@ public void waitForEventAggregate(
...
+      }
+      waitConditions.add(condition);
```
kgyrtkirk: nit: we don't necessarily need to add if the condition is already satisfied

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
```diff
@@ -107,9 +93,9 @@ public void close()
   public void waitForEvent(Predicate<Event> condition, long timeoutMillis)
   {
     final WaitCondition waitCondition = new WaitCondition(condition);
+    registerWaitCondition(waitCondition);
     waitConditions.add(waitCondition);
```
kgyrtkirk: nit: `register` may already have added it to the `waitConditions`

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
```diff
@@ -158,48 +144,64 @@ public void waitForEventAggregate(
...
-            condition.countDownLatch.countDown();
-          }
+        if (condition.predicate.test(event)) {
+          condition.countDownLatch.countDown();
         }
-        condition.processedUntil = currentNumberOfEvents;
       }
     }
     catch (Exception e) {
       log.error(e, "Error while evaluating wait conditions");
```
kgyrtkirk: this is partially unrelated; but I think this should be a serious error - (possibly the predicate is broken?)...repack into `RuntimeException` if necessary?

server/src/test/java/org/apache/druid/server/metrics/ServiceStatusMonitorTest.java
```diff
@@ -47,36 +49,27 @@ public void setUp()
...
     heartbeatTags.put("taskRunner", "http");
-    final StubServiceEmitter emitter = new StubServiceEmitter("service", "host");
     monitor.doMonitor(emitter);
-    Assert.assertEquals(1, emitter.getEvents().size());
-    Assert.assertEquals(1, emitter.getEvents().get(0).toMap().get("leader"));
-    Assert.assertEquals("http", emitter.getEvents().get(0).toMap().get("taskRunner"));
-    Assert.assertEquals(HEARTBEAT_METRIC_KEY, emitter.getEvents().get(0).toMap().get("metric"));
-    Assert.assertEquals(1, emitter.getEvents().get(0).toMap().get("value"));
+    Assert.assertEquals(1, emitter.getNumEmittedEvents());
+    emitter.verifyValue(HEARTBEAT_METRIC_KEY, Map.of("leader", 1, "taskRunner", "http"), 1);
```
kgyrtkirk, line 73: pretty cool how concise this become! :)

services/src/test/java/org/apache/druid/server/AsyncQueryForwardingServletTest.java
kfaraz: Thanks for catching this!

server/src/test/java/org/apache/druid/server/metrics/ServiceStatusMonitorTest.java
kfaraz: 🙂

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
kfaraz: Oops, let me fix this up!

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
kfaraz: Yeah, that's true, let me check this again.

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
kfaraz: I checked this. It seems correct. The condition is evaluated (using `condition.predicate.test()`) against all the events and the latch is counted down if the condition is satisfied at any point.
I think we can probably break the loop once the condition is satisfied.

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
kfaraz: Thanks for catching this, let me clean this up.

server/src/test/java/org/apache/druid/server/ClientQuerySegmentWalkerTest.java
kfaraz: Yeah, I am not sure why this class `ServiceMetricEventSnapshot` was added either.
Let me see what I can do to clean up this part.

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
kfaraz: That would be good but this is running on a separate thread. So the thrown exception will not cause any real failure except for the countdown latch eventually timing out.

I could use futures instead of countdown latches, but the latch gives a cleaner flow.

Let me know if you have any suggestion to handle this.

server/src/test/java/org/apache/druid/server/ClientQuerySegmentWalkerTest.java
kfaraz: Took a look at this. Apparently the `userDims` can change if the builder is reused for emitting another event.
But this is problematic anyway.

If the same event is to be emitted using (one or more) emitters, it is possible that before we even get to emit the event (as emission is always async), the dimensions of the event have changed.

I wonder if we should do this in a separate PR though as it is addressing a specific race condition. Let me know what you think.

kfaraz: Left a comment on the original PR https://github.com/apache/druid/pull/17170#discussion_r2209227633
I have added a `ServiceMetricEvent.copy()` for now as you suggested.

kfaraz:
@kgyrtkirk , thanks a lot for the prompt review and for the suggestions!
I have incorporated your feedback. Please let me know if the changes look good.

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
kgyrtkirk: why the need to put it on a different thread? it doesn't seem like its a lot of computation...

kfaraz: Hmm, yeah, you are right. The original design needed the other thread because some conditions had to be evaluated from the start. But now we evaluate only one event at a time, so I guess we can do it synchronously after all.

Thanks for the suggestion!

kfaraz: I am still not sure if throwing an exception here will help though.
The contract of `Emitter.emit()` says that it shouldn't throw an exception and only log warnings/errors.
An exception thrown here would still not fail the test directly as the `emit()` call might be happening on some service thread that just swallows up the exception.

kgyrtkirk: I'm afraid we can't tell future bugs to not happen - so that apidoc is hard to live up to ; it may suggest to try to avoid that ; but the system invoking this api should fully prepared that something might go wrong even in this method

kfaraz: Fair enough, I will throw a runtime exception here. Hopefully, it helps us catch issues early.

PR #18257, opened 2025-07-15
Disabling parallel builds.

cryptoe:
Disabling parallel builds so that we donot run into : https://github.com/apache/druid/actions/runs/16289799387/job/45996888923

My hunch is that the parallel build is causing some locks to trip.


cryptoe:
Failures unrelated. Going ahead with merge.

PR #18259, opened 2025-07-15
Upgrade pac4j extension

ashibhardwaj:
Fixes CVE-2023-52428.

### Description

CVE-2023-52428 requires upgrading nimbus-jose-jwt to `9.37.2` which is not compatible with pac4j v4. Therefore, this PR does a major version upgrade from pac4j v4.5.7 to v5.7.3. 

#### Version Updates (pom.xml)
- pac4j: 4.5.7 → 5.7.3
- nimbus-jose-jwt: 8.22.1 → 9.37.2
- oauth2-oidc-sdk: 8.22 → 10.8

#### New Dependency Added
- Added pac4j-jee dependency (JEE components were moved to separate module in pac4j 5.x)

#### Import Changes Across Files
- JEEContext and JEEHttpActionAdapter moved from pac4j-core to pac4j-jee
- CallContext removed (no longer exists in pac4j 5.x)
- JavaSerializer from org.pac4j.core.util removed

#### Code Changes
- Pac4jFilter.java
  - Updated constructor to match pac4j 5.x API
  - Removed CallContext usage
  - Updated imports for JEE components
- Pac4jSessionStore.java
  - Replaced pac4j's JavaSerializer with standard Java serialization
- Pac4jAuthenticator.java
  - Updated to use callback path constant
- Test Files
  - Updated Pac4jFilterTest.java and Pac4jSessionStoreTest.java to match new APIs
  - Fixed constructor calls with updated parameters

#### Configuration Files
- Updated licenses.yaml for new dependency versions
- Updated owasp-dependency-check-suppressions.xml for security scanning

### Potential Breaking Changes & Risks
1. Session Serialization Change
Risk: The switch from pac4j's JavaSerializer to standard Java serialization could cause issues with existing user sessions
Impact: Users with active sessions might need to re-authenticate after the upgrade
2. JEE Dependency Separation
Risk: The move of JEE components to pac4j-jee module could affect deployment
Impact: Need to ensure the new dependency is properly included in distribution
3. API Changes
Risk: pac4j 5.x has significant API changes that could affect custom configurations
Impact: Any custom pac4j configurations outside this codebase might break
4. Java 8 Compatibility
Risk: The OWASP suppressions mentioned pac4j 5.7.3 might not support JDK 8
Impact: Could affect environments still running Java 8

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/druid-pac4j/src/main/java/org/apache/druid/security/pac4j/Pac4jFilter.java
+++ b/extensions-core/druid-pac4j/src/main/java/org/apache/druid/security/pac4j/Pac4jFilter.java
@@ -19,19 +19,15 @@
 
 package org.apache.druid.security.pac4j;
 
-import com.google.common.collect.ImmutableMap;
 import org.apache.druid.java.util.common.logger.Logger;
 import org.apache.druid.server.security.AuthConfig;
 import org.apache.druid.server.security.AuthenticationResult;
 import org.pac4j.core.config.Config;
-import org.pac4j.core.context.JEEContext;
-import org.pac4j.core.context.session.SessionStore;
-import org.pac4j.core.engine.CallbackLogic;
 import org.pac4j.core.engine.DefaultCallbackLogic;
 import org.pac4j.core.engine.DefaultSecurityLogic;
-import org.pac4j.core.engine.SecurityLogic;
-import org.pac4j.core.http.adapter.JEEHttpActionAdapter;
-import org.pac4j.core.profile.UserProfile;
+import org.pac4j.core.exception.http.HttpAction;
+import org.pac4j.jee.context.JEEContext;
+import org.pac4j.jee.http.adapter.JEEHttpActionAdapter;
 
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
@@ -42,30 +38,30 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
-import java.util.Collection;
 
 public class Pac4jFilter implements Filter
 {
   private static final Logger LOGGER = new Logger(Pac4jFilter.class);
 
   private final Config pac4jConfig;
-  private final SecurityLogic<Object, JEEContext> securityLogic;
-  private final CallbackLogic<Object, JEEContext> callbackLogic;
-  private final SessionStore<JEEContext> sessionStore;
-
+  private final Pac4jSessionStore sessionStore;
+  private final String callbackPath;
   private final String name;
   private final String authorizerName;
 
-  public Pac4jFilter(String name, String authorizerName, Config pac4jConfig, String cookiePassphrase)
+  public Pac4jFilter(
+          String name,
+          String authorizerName,
+          Config pac4jConfig,
+          String callbackPath,
+          String cookiePassphrase
+  )
   {
     this.pac4jConfig = pac4jConfig;
-    this.securityLogic = new DefaultSecurityLogic<>();
-    this.callbackLogic = new DefaultCallbackLogic<>();
-
+    this.callbackPath = callbackPath;
     this.name = name;
     this.authorizerName = authorizerName;
-
-    this.sessionStore = new Pac4jSessionStore<>(cookiePassphrase);
+    this.sessionStore = new Pac4jSessionStore(cookiePassphrase);
   }
 
   @Override
@@ -73,10 +69,9 @@ public class Pac4jFilter implements Filter
   {
   }
 
-
   @Override
   public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain)
-      throws IOException, ServletException
+          throws IOException, ServletException
   {
     // If there's already an auth result, then we have authenticated already, skip this or else caller
     // could get HTTP redirect even if one of the druid authenticators in chain has successfully authenticated.
@@ -85,38 +80,59 @@ public class Pac4jFilter implements Filter
       return;
     }
 
-    HttpServletRequest httpServletRequest = (HttpServletRequest) servletRequest;
-    HttpServletResponse httpServletResponse = (HttpServletResponse) servletResponse;
-    JEEContext context = new JEEContext(httpServletRequest, httpServletResponse, sessionStore);
+    HttpServletRequest request = (HttpServletRequest) servletRequest;
+    HttpServletResponse response = (HttpServletResponse) servletResponse;
+    JEEContext context = new JEEContext(request, response);
 
-    if (Pac4jCallbackResource.SELF_URL.equals(httpServletRequest.getRequestURI())) {
+    if (request.getRequestURI().equals(callbackPath)) {
+      DefaultCallbackLogic callbackLogic = new DefaultCallbackLogic();
+      String originalUrl = (String) request.getSession().getAttribute("pac4j.originalUrl");
+      String redirectUrl = originalUrl != null ? originalUrl : "/";
+
       callbackLogic.perform(
-          context,
-          pac4jConfig,
-          JEEHttpActionAdapter.INSTANCE,
-          "/",
-          true, false, false, null);
+              context,
+              sessionStore,
+              pac4jConfig,
+              JEEHttpActionAdapter.INSTANCE,
+              redirectUrl,                      // Redirect to original URL or root
+              null,
+              null
+      );
     } else {
-      UserProfile profile = (UserProfile) securityLogic.perform(
-          context,
-          pac4jConfig,
-          (JEEContext ctx, Collection<UserProfile> profiles, Object... parameters) -> {
-            if (profiles.isEmpty()) {
-              LOGGER.warn("No profiles found after OIDC auth.");
-              return null;
-            } else {
-              return profiles.iterator().next();
-            }
-          },
-          JEEHttpActionAdapter.INSTANCE,
-          null, "none", null, null);
-      // Changed the Authorizer from null to "none".
-      // In the older version, if it is null, it simply grant access and returns authorized.
-      // But in the newer pac4j version, it uses CsrfAuthorizer as default, And because of this, It was returning 403 in API calls.
-      if (profile != null && profile.getId() != null) {
-        AuthenticationResult authenticationResult = new AuthenticationResult(profile.getId(), authorizerName, name, ImmutableMap.of("profile", profile));
-        servletRequest.setAttribute(AuthConfig.DRUID_AUTHENTICATION_RESULT, authenticationResult);
-        filterChain.doFilter(servletRequest, servletResponse);
+      DefaultSecurityLogic securityLogic = new DefaultSecurityLogic();
+      try {
+        securityLogic.perform(
+                context,
+                sessionStore,
+                pac4jConfig,
+                (ctx, session, profiles, parameters) -> {
+                  try {
+                    // Extract user ID from pac4j profiles and create AuthenticationResult
+                    if (profiles != null && !profiles.isEmpty()) {
+                      String uid = profiles.iterator().next().getId();
+                      if (uid != null) {
+                        AuthenticationResult authenticationResult = new AuthenticationResult(uid, authorizerName, name, null);
+                        servletRequest.setAttribute(AuthConfig.DRUID_AUTHENTICATION_RESULT, authenticationResult);
+                        filterChain.doFilter(servletRequest, servletResponse);
+                      }
+                    } else {
+                      LOGGER.warn("No profiles found after OIDC auth.");
+                      // Don't continue the filter chain - let pac4j handle the authentication failure
+                    }
+                  }
+                  catch (IOException | ServletException e) {
+                    throw new RuntimeException(e);
+                  }
+                  return null;
+                },
+                JEEHttpActionAdapter.INSTANCE,
+                null,
+                "none",  // Use "none" instead of authorizerName to avoid CSRF issues
+                null
+        );
+      }
+      catch (HttpAction e) {
+        JEEHttpActionAdapter.INSTANCE.adapt(e, context);
       }
     }
   }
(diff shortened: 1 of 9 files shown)


Pankaj260100 (approved):

owasp-dependency-check-suppressions.xml
```diff
@@ -355,7 +355,7 @@
     <!-- However, vulnerability scan still shows this CVE. Pac4j release notes mention 5.3.1 as "fully fixed" version. -->
     <!-- Remove suppression once upgraded to 5.3.1. -->
     <notes><![CDATA[
-   file name: pac4j-core-4.5.7.jar
+   file name: pac4j-core-5.7.3.jar
```
capistrant: per the comment above this line, is this CVE that we are suppressing able to be removed from this file since we are going past 5.3.1

ashibhardwaj: I think it should, let me try removing the suppression.

ashibhardwaj: Upon doing so, I see two tests failing both of which look like transient issues. Can you please re-run both of them?

The third CodeQL check has been failing due to cryptographic algorithm which I have not even changed.

capistrant:
merging master into this branch  should re-trigger CI with latest GHA fixes

ashibhardwaj:
@capistrant I've merged master as recommended and the flaky tests passed. Now only two checks are failing. One due to vulnerabilities in web-console (there already is an [open PR](https://github.com/apache/druid/pull/18310) for its fix) and the other due to use of a potentially broken or risky cryptographic algorithm - but my PR has not updated the algo. We are using the same algo as earlier.

.github/config/codeql-config.yml
```diff
@@ -1,3 +1,5 @@
 query-filters:
   - exclude:
       id: java/stack-trace-exposure
+  - exclude:
```
capistrant: I wonder if it is better to leave this codeql in and just ignore it for this PR if it is false positive vs turn it off completely? committers should still be able to push it through and merge even if there is a codeql fail (I think)

owasp-dependency-check-suppressions.xml
```diff
@@ -129,6 +129,10 @@
     <cve>CVE-2022-26612</cve>
     <!-- this one seems to apply to backend server - https://nvd.nist.gov/vuln/detail/CVE-2023-25613 -->
     <cve>CVE-2023-25613</cve>
+    <cve>CVE-2025-48734</cve>
+    <cve>CVE-2024-13009</cve>
+    <cve>CVE-2023-52428</cve>
+    <cve>CVE-2025-48734</cve>
```
capistrant: this is listed twice

owasp-dependency-check-suppressions.xml
capistrant, line 1: Are all of these additions directly tied to the changes in dependencies in this PR? I don't think we should be modifying suppressions for anything that is not directly caused by this PRs change in dependencies. New CVEs flagged by the cron should be addressed individually with attempts to remediate before suppressing

pom.xml
```diff
@@ -1823,7 +1827,7 @@
             <plugin>
                 <groupId>org.owasp</groupId>
                 <artifactId>dependency-check-maven</artifactId>
-                <version>7.4.4</version>
+                <version>12.1.3</version>
```
capistrant: This is a fairly significant bump for this extension. Are we sure there is no unintended consequences? I agree that it should be brought up to date, but perhaps in another PR dedicated to bumping plugin versions?

pom.xml
```diff
@@ -78,7 +78,7 @@
         <apache.kafka.version>3.9.1</apache.kafka.version>
         <!-- when updating apache ranger, verify the usage of aws-bundle-sdk vs aws-logs-sdk
         and update as needed in extensions-core/druid-ranger-security/pm.xml  -->
-        <apache.ranger.version>2.4.0</apache.ranger.version>
+        <apache.ranger.version>2.6.0</apache.ranger.version>
```
capistrant: Is the ranger stuff in this PR required for pac4j or were you opportunistically trying to update more? same with netty and commons lang3 below? Perhaps it is all tied together, but if not I think this PR should focus in on just changes required for the pac4j extension work.

capistrant (commented):
Beyond the pac4j extension changes themselves, is there a lot of scope creep in this PR updating maven plugin versions, unrelated dependency versions, etc. as well as CVE suppressions? I'm getting a little bit confused on what at it's core is needed for the pac4j change. Since this is a significant change for pac4j I think this PR really should focus only on what is needed to be done to do the upgrade. Any other changes can be grouped in one or more other PRs. Definitely appreciate all the changes in this PR, and will be happy to help review the non-pac4j stuff if you request me on them.

as for the pac4j specific changes, I am continuing to review and will get back hopefully tomorrow.

ashibhardwaj:
@capistrant I had to upgrade those extra dependencies because `Cron Job ITs / security vulnerabilities` kept failing. I can alternatively suppress those. Let me know.

.github/config/codeql-config.yml
ashibhardwaj: I was not aware that we could so I planned to create a follow up PR to enable it back immediately upon merging this one (as mentioned in the description). I can surely remove this exclusion if we can merge the PR even with a failing check.

pom.xml
ashibhardwaj: I actually tried with other versions too but saw some errors. I can fetch them for you from previous commits if you are interested.

owasp-dependency-check-suppressions.xml
ashibhardwaj: All the CVEs being suppressed are due to already existing dependencies.

ashibhardwaj: | CVE Library      | Module             | CVE(s)                                              |
|------------------|-------------------|-----------------------------------------------------|
| netty            | azure extensions  | CVE-2025-55163                                      |
| ranger-plugins   | ranger security   | CVE-2024-45217, CVE-2024-45479, CVE-2024-55532      |
| icu4j            | processing modules| CVE-2025-5222                                       |
| solrj            | ranger security   | CVE-2024-45216                                      |
| kafka-clients    | ranger security   | CVE-2025-27819                                      |
| hadoop-client    | orc-extensions    | CVE-2025-48734, CVE-2024-13009, CVE-2023-52428      |

pom.xml
ashibhardwaj: I have reverted all non-necessary (for pac4j extension) upgrades and added suppressions instead. Replied to your other comment with more details.

capistrant:
> @capistrant I had to upgrade those extra dependencies because `Cron Job ITs / security vulnerabilities` kept failing. I can alternatively suppress those. Let me know.

Since the cron job ITs are not actually a part of a PR workflow in Druid, we can not concern this PR with things that are failing those workflow runs. Therefore, you should not even add suppressions for them to this PR. The only time you should add a suppression as a part of a feature PR is if something in the PR introduces a new CVE that we are both unable to resolve, and are comfortable suppressing if we think it is ok to do so (false positive, not exploitable, risk accepted, etc).

ashibhardwaj:
@capistrant In that case, I'll just remove the suppressions and let the check fail.

extensions-core/druid-pac4j/src/main/java/org/apache/druid/security/pac4j/Pac4jSessionStore.java
```diff
@@ -19,71 +19,77 @@
...
   public Optional<Object> get(WebContext context, String key)
   {
-    final Cookie cookie = ContextHelper.getCookie(context, PAC4J_SESSION_PREFIX + key);
+    final Cookie cookie = getCookie(context, PAC4J_SESSION_PREFIX + key);
     Object value = null;
-    if (cookie != null) {
+    if (cookie != null && cookie.getValue() != null) {
       value = uncompressDecryptBase64(cookie.getValue());
     }
     LOGGER.debug("Get from session: [%s] = [%s]", key, value);
```
capistrant: I'm a little nervous about us logging a decrypted cookie in plaintext, even if it is just for debugging. A misconfig by an operator is all it takes from writing a bunch of these in plaintext. Is it really necessary to ever log this information? Could we put it behind a config even, like `druid.pac4j.insecure.debugLogCookies` (I made up the config name structure here, but just want to make the point of it being some kind of runtime property)

extensions-core/druid-pac4j/src/main/java/org/apache/druid/security/pac4j/Pac4jSessionStore.java
```diff
@@ -96,37 +102,103 @@ public void set(WebContext context, String key, @Nullable Object value)
     Object profile = value;
     Cookie cookie;
 
-    if (value == null) {
-      cookie = new Cookie(PAC4J_SESSION_PREFIX + key, null);
+    // Check if value is null, empty string, or empty collection
+    boolean isEmpty = value == null ||
```
capistrant, line 105: would it be an exception case if the value doesn't match any of the types checked here?

extensions-core/druid-pac4j/src/main/java/org/apache/druid/security/pac4j/Pac4jSessionStore.java
ashibhardwaj: You're absolutely right. I was trying to keep the behavior similar to how it was before but logging decrypted session data is definitely a security risk even if it's just for debugging. 
I'll remove the debug logging from both the get and set methods entirely. If we need debugging for session issues in the future, we can always add safer metadata logging (like just logging the key and whether a value was found) or use debugger breakpoints in development environments.
Thanks for catching this!

extensions-core/druid-pac4j/src/main/java/org/apache/druid/security/pac4j/Pac4jSessionStore.java
ashibhardwaj: No, it wouldn't be an exception case - that's actually the intended behavior. If the value doesn't match any of the "empty-able" types being checked (null, empty String, empty Collection, empty Map), then isEmpty will be false and the code will proceed to store the value normally in the else block.
This makes sense for pac4j sessions because most stored objects are things like CommonProfiles or other custom objects that should be preserved even if they don't fall into those specific empty-checkable categories. The serializeToBytes method can handle any Serializable object, so we only need to explicitly check for the common "empty" cases where we want to clear the cookie instead of storing it.

capistrant: Thanks for the info!

extensions-core/druid-pac4j/src/main/java/org/apache/druid/security/pac4j/Pac4jFilter.java
```diff
@@ -85,38 +80,59 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
     }
 
-    HttpServletRequest httpServletRequest = (HttpServletRequest) servletRequest;
-    HttpServletResponse httpServletResponse = (HttpServletResponse) servletResponse;
-    JEEContext context = new JEEContext(httpServletRequest, httpServletResponse, sessionStore);
+    HttpServletRequest request = (HttpServletRequest) servletRequest;
+    HttpServletResponse response = (HttpServletResponse) servletResponse;
+    JEEContext context = new JEEContext(request, response);
+
+    if (request.getRequestURI().equals(callbackPath)) {
```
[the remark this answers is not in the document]
ashibhardwaj: Good point about path traversal in general, but in this specific case I don't think it's a security concern. This is just determining whether to run OAuth callback logic vs. authentication logic for a fixed, well-known OAuth callback endpoint (/druid-ext/druid-pac4j/callback).
If someone tries path traversal it would:
- Fail the equals() check
- Fall through to the authentication logic instead of callback logic
- Not grant any unauthorized access

The callback logic itself validates OAuth parameters and only processes legitimate callbacks from the identity provider. So path traversal attempts would just result in normal authentication flow rather than any security bypass.
This was the equivalent check we had in pac4j v4 (Pac4jCallbackResource.SELF_URL.equals(httpServletRequest.getRequestURI())), and simple string equality is sufficient for this OAuth callback use case.

extensions-core/druid-pac4j/src/main/java/org/apache/druid/security/pac4j/Pac4jSessionStore.java
```diff
@@ -96,37 +102,103 @@ public void set(WebContext context, String key, @Nullable Object value)
...
-    cookie.setDomain("");
     cookie.setHttpOnly(true);
-    cookie.setSecure(ContextHelper.isHttpsOrSecure(context));
+    // Always set secure flag for authentication cookies to prevent transmission over HTTP
+    // This ensures the cookie is only sent over HTTPS connections
+    boolean isSecure = isHttpsOrSecure(context);
+    if (!isSecure) {
+      LOGGER.warn("Setting authentication cookie over non-HTTPS connection. This is not recommended for production.");
+    }
+    cookie.setSecure(true); // Always set secure flag for authentication cookies
```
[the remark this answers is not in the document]
ashibhardwaj: The current code already logs a warning when the connection isn't HTTPS, so operators are alerted to the security risk.
This prevents accidental exposure in production environments where proxy configurations might not set the secure flag correctly, and it encourages better security practices throughout the development lifecycle.
Authentication systems should be secure by default rather than convenient by default.

ashibhardwaj:
@capistrant thanks for the review! I have addressed all the comments.

capistrant (approved):
Thanks for all the changes and work on getting pac4j upgraded. The last thing is not code related but rather for Druid35 release prep. Can you add a section to the PR description called `release note` and provide a short blurb to potentially be included in the release notes. Can you also add a section called `upgrade note` calling out the potential for re-authentication being needed during a rolling upgrade and why? We will want to make the community aware of the chance for that happening in the upgrade notices. Thanks!

ashibhardwaj:
@capistrant thanks for approving the PR. 
I have added both the sections you suggested in the description.

PR #18261, opened 2025-07-15
[Backport] supress 2 new CVEs that popped in the druid34 release process

capistrant:
Backport of #18227 to 34.0.0.


capistrant:
security IT flaking is something that needs bugfix in maste rand then backport. merging this despite failure.

PR #18262, opened 2025-07-15
Allow time series query to return empty preferred ordering for All Granularity

cecemei:
### Description

Allow time series query to return empty preferred ordering for All Granularity 
 
This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
+++ b/processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
@@ -1426,13 +1426,14 @@ public class CursorFactoryProjectionTest extends InitializedNullHandlingTest
   @Test
   public void testTimeseriesQueryOrderByNotCompatibleWithProjection()
   {
+    // Query has `__time ASC` ordering, but projection has `b ASC` ordering, so no projection can be used.
+    // This is not ideal, since we know query has Granularities.DAY, and segment also has Granularities.DAY, so the __time ordering does not matter.
     final TimeseriesQuery query = Druids.newTimeseriesQueryBuilder()
                                         .dataSource("test")
                                         .intervals(ImmutableList.of(Intervals.ETERNITY))
                                         .granularity(Granularities.DAY)
-                                        .descending(true)
                                         .aggregators(new LongSumAggregatorFactory("c_sum", "c"))
-                                        .context(ImmutableMap.of(QueryContexts.USE_PROJECTION, "c_sum_daily"))
+                                        .context(ImmutableMap.of(QueryContexts.USE_PROJECTION, "b_c_sum"))
                                         .build();
 
     final CursorBuildSpec buildSpec = TimeseriesQueryEngine.makeCursorBuildSpec(query, null);
@@ -1440,7 +1441,7 @@ public class CursorFactoryProjectionTest extends InitializedNullHandlingTest
         DruidException.class,
         () -> projectionsCursorFactory.makeCursorHolder(buildSpec));
     Assert.assertEquals(DruidException.Category.INVALID_INPUT, e.getCategory());
-    Assert.assertEquals("Projection[c_sum_daily] specified, but does not satisfy query", e.getMessage());
+    Assert.assertEquals("Projection[b_c_sum] specified, but does not satisfy query", e.getMessage());
   }
 
   @Test


processing/src/main/java/org/apache/druid/query/timeseries/TimeseriesQueryEngine.java
```diff
@@ -337,7 +338,9 @@ public static CursorBuildSpec makeCursorBuildSpec(TimeseriesQuery query, @Nullab
                        .setAggregators(query.getAggregatorSpecs())
                        .setQueryContext(query.context())
                        .setPreferredOrdering(
-                           query.isDescending() ? Cursors.descendingTimeOrder() : Cursors.ascendingTimeOrder()
+                           query.getGranularity().equals(Granularities.ALL)
+                           ? ImmutableList.of()
```
clintropolis: nit: i think we have been slowly migrating to using stuff like `List.of` instead of `ImmutableList` when possible

clintropolis (commented):
would be nice to add a test unrelated to projections since this impacts all timeseries queries using all granularity and allows them to use non-time ordered segments. You could use `CursorFactoryProjectionTest` which tests both time ordered and non-time ordered segments, and add a test with timeseries that doesn't match any projections (as well as one that matches projections) so we get coverage for both cases.

also, i think the unrelated docker changes probably shouldn't be here?

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
```diff
@@ -175,8 +174,8 @@ public String toString()
   private static ProjectionOrdering computeOrdering(VirtualColumns virtualColumns, List<DimensionSchema> groupingColumns)
   {
     if (groupingColumns.isEmpty()) {
-      // call it time ordered, there is no grouping columns so there is only 1 row for this projection
-      return new ProjectionOrdering(Cursors.ascendingTimeOrder(), null);
+      // no ordering since there is only 1 row for this projection
+      return new ProjectionOrdering(List.of(), null);
```
clintropolis, line 178: why this change? This is effectively time ordered within the constraints of the segment granularity, it should still be able to participate in time order required queries (like when the query granularity is equal to the segment granularity)

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -1343,6 +1351,61 @@ public void testProjectionSelectionMissingColumnOnBaseTableToo()
...
+                                        .aggregators(new LongSumAggregatorFactory("c_sum", "c"))
+                                        .context(ImmutableMap.of(QueryContexts.USE_PROJECTION, "c_sum_daily"))
+                                        .build();
+
+    final CursorBuildSpec buildSpec = TimeseriesQueryEngine.makeCursorBuildSpec(query, null);
+    DruidException e = Assert.assertThrows(
+        DruidException.class,
+        () -> projectionsCursorFactory.makeCursorHolder(buildSpec));
+    Assert.assertEquals(DruidException.Category.INVALID_INPUT, e.getCategory());
+    Assert.assertEquals("Projection[c_sum_daily] specified, but does not satisfy query", e.getMessage());
```
clintropolis: This is the reason why I don't think we should have made that change in `AggregateProjectionSpec` - why shouldn't this be allowed? The segment projection has the same granularity as the query, so it should produce the same results, does it not for some reason?

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -1343,6 +1351,61 @@ public void testProjectionSelectionMissingColumnOnBaseTableToo()
     Assert.assertArrayEquals(new Object[]{"b", null, 12L, 13.2}, results.get(1).getArray());
   }
 
+  @Test
+  public void testTimeseriesQueryAllGranularityCanMatchNonTimeDimProjection()
```
clintropolis, line 1355: it would be nice to have a test for a timeseries query that doesn't match a projection and also doesn't expect any sort of failure (but still runs against a non-time ordered segment since there isn't any coverage for that in `TimeseriesQueryRunnerTest`).

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
clintropolis: oh actually c_sum_daily has a day granularity time column, not missing time column, so is different from the change in `AggregateProjectionSpec` isn't it. Still though, why doesn't this match, is it the direction? since it is ordered by __time it could still work with a descending cursor, maybe the match code needs to allow descending time order if the projection is still ordered by time?

cecemei: This throws because `OrderBy` time doesn't match, `descending(true)` while the projection use `ASC __time`. It's just test coverage and not due to changes in this PR.

clintropolis: ok, then we need to update the match code (it doesn't need to be in this PR). Non-vectorized engines allow descending cursors, so we should special handle time ordering preference to allow ascending time order to match queries that want descending time order

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
clintropolis: i guess some projection match code here needs updated too to consider the data interval of the segment the projection lives in (if the query granularity contains the segment data interval then even an ALL granularity projection is fine to use for the query because it is still at most data interval granularity). This doesn't need to be done in this PR either

cecemei: Not sure I quite understand, do you mean segment granularity check on projection would fail because of this change? I thought this is better because now it's consistent with timeseries query's orderBy

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
cecemei: Added a test for this `testTimeseriesQueryAllGranularitiesAlwaysRuns`

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
clintropolis: I mean that a projection even if it doesn't have a time column it still only has data within the segments data interval. So if the query time column grouping is the same size or coarser than the segment data interval, the projection can be considered time ordered because we would be bucketing the time of the projection rows to the coarser value. Projections with no physical time column still make a selector for __time (because it is required), it is just set to the start of the segments interval, if that helps it make sense.

cecemei: ah do you mean that if a query is bucket at `DAY(__time)`, and segment is bucket at `HOUR(__time)`, then the `OrderBy` in projection doesn't matter since it'd be grouped anyway?

I'm thinking that projection should inherit the granularity from segment, which it seems like it doesn't have that info right now? it should also work with `DESC __time` too if the query bucket is larger?

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
cecemei: true, I also updated the test case to make it more clear about the `OrderBy` mismatch. Would need to add the descending match later.

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
clintropolis: >ah do you mean that if a query is bucket at DAY(__time), and segment is bucket at HOUR(__time), then the OrderBy in projection doesn't matter since it'd be grouped anyway?

Ish, yea, the projection transformation of the __time column must be finer than or equal to the query transformation of the time_column for the projection to be able to compute the same query results, so a time_floor(__time, 'P1D') would group all of the hourly segment times into the same day (even if the projection has no __time transformation because all of the rows would still have a timestamp of the beginning of the hour or whatever).

>I'm thinking that projection should inherit the granularity from segment, which it seems like it doesn't have that info right now? 

A projection doesn't have an explicit granularity, which was a deliberate design decision. It either has a precomputed virtual column with a time_floor expression on __time (or even just __time directly from the base table) which is stored as a physical column in the segment, or it doesn't. The latter case as far as the engines are concerned still has the __time column due to the constraints of the segment interval, it just isn't a physical column and instead is a constant of min timestamp. The projection schema stored in the metadata defines which physical columns are present in the segment for a given projection, so a projection itself shouldn't really inherit a segment granularity directly if that is what you mean. Instead, the match code should consider the interval of the segment to determine if the projection __time transformation matter or not (probably updating `PhysicalColumnChecker` to add a method to get the data interval? though i havent thought too hard about it yet)

In general, we have been downplaying the concept of granularity since it is basically an oddly specific way to define a grouping on time for a query, and the query engine implementations of it only work when a cursor is time ordered (we have some plans to transition this to virtual columns too but haven't got around to it yet).

> it should also work with DESC __time too if the query bucket is larger?

the descending thing isn't really directly related to __time matching, and would handled in a different place in code, i think by just dropping the direction from consideration. A query needing a descending order on __time should work if the base table or a projection is ordered by __time in either direction since the cursor can be done ascending or descending order (vector cursor is not yet, but that isn't this codes concern really). The other constraints of projection __time matching must also be true (such as finer than or equal to __time transformation), but the code that is done via the virtual column comparison because the ordering isn't part of the time_floor expression itself. While we don't have any implementations of it yet, the same would also be true if we had engines that could take advantage of ordering on things other than __time, if we need `order by x desc` then as long as the table is `order by x` in any direction we could use it for the query

cecemei: I see, would it be better if we have some backdoor to ignore time ordering in the query, if that's the only ordering, since the granularity is already checked later?

clintropolis: discussed offline, we can leave this change for now since there is some work to do for matching projections in cases where the __time column floor transformation of the query contains the segment interval

PR #18263, opened 2025-07-16
Enable Dart ResultIterator to consider queryTimeout and enhance CI with stack traces

kgyrtkirk:
* adds back jfr related things which were disabled earlier 
* add a few additional tags for jfr
* starts a background thread in every job which is run by `worker.yml` and collect stacktraces every minute for all java processes
   * I guess this will increase with around a ~`100k` the artifact data for every job
   * this will probably help analyze post-morten the standing-still `ResultIterator` I have seen 
* configures Dart's ResultIterator to timeout reads after querytimeout
   * this could help mitigate the stuck threads being hanging around after the failure I'm still trying to reproduce

### DIFF
--- a/.github/scripts/collect_jstacks
+++ b/.github/scripts/collect_jstacks
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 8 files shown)


kgyrtkirk:
thank you @cryptoe  for reviewing these changes!

PR #18268, opened 2025-07-16
[Backport] Web console: fix maxTask selector with SET statements

capistrant:
Backport of #18221 to 34.0.0.


capistrant:
these are both the flakes in IT that need to be fixed in master and backported to 34

PR #18272, opened 2025-07-17
Make ServiceMetricEvent immutable, do not allow null dimension values

kfaraz:
Follow up to #18255 to handle null dimension values in `ServiceMetricEvent`

Related comment: https://github.com/apache/druid/pull/17170#discussion_r2209227633

`ServiceMetricEvent` should ideally be immutable.
Otherwise, there can be a race if the `ServiceMetricEvent.Builder` is reused while the first event
is still being emitted.

### Changes

- Make `ServiceMetricEvent` immutable. All fields were already immutable. Fix up `userDims` to be immutable too.
- Do not allow null dimension names or values
- Remove `ServiceMetricEvent.copy` as the event is already immutable.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
+++ b/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
@@ -151,20 +151,34 @@ public class ServiceMetricEvent implements Event
       return this;
     }
 
+    /**
+     * Adds a dimension to be emitted with this metric event, only if the given
+     * value is not null.
+     *
+     * @throws IAE if the dimension name is null.
+     */
     public Builder setDimensionIfNotNull(String dim, Object value)
     {
       if (dim == null) {
         throw new IAE("Dimension name cannot be null");
-      } else if (value == null) {
-        throw new IAE("Value of dimension[%s] cannot be null", dim);
       }
 
-      userDims.put(dim, value);
+      if (value != null) {
+        userDims.put(dim, value);
+      }
       return this;
     }
 
+    /**
+     * Adds a dimension to be emitted with this metric event.
+     *
+     * @throws IAE if the dimension name or the given value is null.
+     */
     public Builder setDimension(String dim, Object value)
     {
+      if (value == null) {
+        throw new IAE("Value of dimension[%s] cannot be null", dim);
+      }
       return setDimensionIfNotNull(dim, value);
     }
 
--- a/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
+++ b/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
@@ -47,7 +47,7 @@ public class ServiceMetricEvent implements Event
 
   private final DateTime createdTime;
   private final ImmutableMap<String, String> serviceDims;
-  private final Map<String, Object> userDims;
+  private final ImmutableMap<String, Object> userDims;
   private final String feed;
   private final String metric;
   private final Number value;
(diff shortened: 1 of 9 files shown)


processing/src/test/java/org/apache/druid/query/DefaultQueryMetricsTest.java
```diff
@@ -66,11 +66,11 @@ public void testDefaultQueryMetricsQuery()
         .context(ImmutableMap.of("testKey", "testValue"))
         .build();
     queryMetrics.query(query);
-    queryMetrics.reportQueryTime(0).emit(serviceEmitter);
     // No way to verify this right now since DefaultQueryMetrics implements a no-op for sqlQueryId(String) and queryId(String)
     // This change is done to keep the code coverage tool happy by exercising the implementation
     queryMetrics.sqlQueryId("dummy");
     queryMetrics.queryId("dummy");
+    queryMetrics.reportQueryTime(0).emit(serviceEmitter);
```
Akshat-Jain, line 73: Why was this change needed? 🤔

processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
Akshat-Jain, line 1: `public Builder setDimension(String dim, Object value)` and `public Builder setDimensionIfNotNull(String dim, Object value)` could probably be collapsed into a single method I think? Any reason for keeping them separate? is it to avoid widespread changes across the codebase?

kfaraz: Yes, it keeps the patch small. 
Also, the `setDimensionIfNotNull` is essentially a short hand which does the null check for you, while the other throws exception on nulls.

processing/src/test/java/org/apache/druid/query/DefaultQueryMetricsTest.java
kfaraz: Need to emit the event after all the dimensions are set. Otherwise, this test would fail.

processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
```diff
@@ -153,22 +143,43 @@ public Builder setFeed(String feed)
...
 
+    /**
+     * Adds a dimension to be emitted with this metric event, only if the given
+     * value is not null.
+     *
+     * @throws IAE if the dimension name is null.
+     */
     public Builder setDimensionIfNotNull(String dim, Object value)
     {
+      if (dim == null) {
```
cecemei: nit: as the name `setDimensionIfNotNull` suggests, consider only keep value check in this function, move the dim check and `userDims.put` logic to `setDimension`. Also consider use `StringUtils.isEmpty` to check dim is not empty string as well.

processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
```diff
@@ -63,7 +66,7 @@ private ServiceMetricEvent(
   {
     this.createdTime = createdTime != null ? createdTime : DateTimes.nowUtc();
     this.serviceDims = serviceDims;
-    this.userDims = userDims;
+    this.userDims = userDims == null ? Map.of() : Map.copyOf(userDims);
```
cecemei: nit: `userDims` could also be `ImmutableMap` in line 50.

kfaraz: Yeah, that would make for better symmetry with the other field `serviceDims`, let me do that.

Akshat-Jain (approved):

kfaraz:
I think we can merge this PR since the IT failure is due to some issue with the k8s cluster setup for `ITNestedQueryPushDownTest`, which is unrelated to the changes here.

kfaraz:
Thanks a lot for the reviews, @Akshat-Jain , @cecemei !
(and also @Akshat-Jain for triggering the CI 😃 )

PR #18273, opened 2025-07-17
validate projection granularity is finer than or equal to segment granularity

clintropolis:
### Description
Adds validation to ensure that the granularity of a projection "__time equivalent" virtual column is constrained to be finer than or equal to the segment granularity. Nothing good happens if this is allowed since the __time values of the projection no longer live within the time interval of the segment, so it is now forbidden.

While adding tests, also found a bug with `Granularities.decorateCursorBuildSpec` which could cause problems with segments that are not time ordered because it wasn't ensuring that the preferred ordering of the decorated `CursorBuildSpec` was time ordered (queries using a granularity other than `ALL` must be time ordered to function properly). 

Also cleaned up some other projection validation messages, and cleaned up some tests.

### DIFF
--- a/processing/src/main/java/org/apache/druid/java/util/common/granularity/Granularities.java
+++ b/processing/src/main/java/org/apache/druid/java/util/common/granularity/Granularities.java
@@ -25,10 +25,13 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import org.apache.druid.math.expr.Expr;
 import org.apache.druid.math.expr.ExprMacroTable;
+import org.apache.druid.query.Order;
+import org.apache.druid.query.OrderBy;
 import org.apache.druid.query.Query;
 import org.apache.druid.query.expression.TimestampFloorExprMacro;
 import org.apache.druid.segment.AggregateProjectionMetadata;
 import org.apache.druid.segment.CursorBuildSpec;
+import org.apache.druid.segment.Cursors;
 import org.apache.druid.segment.VirtualColumn;
 import org.apache.druid.segment.VirtualColumns;
 import org.apache.druid.segment.column.ColumnHolder;
@@ -39,6 +42,7 @@ import org.joda.time.chrono.ISOChronology;
 import javax.annotation.Nullable;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 
 /**
@@ -117,6 +121,16 @@ public class Granularities
             () -> Arrays.stream(buildSpec.getVirtualColumns().getVirtualColumns()).iterator()
         )
     );
+    final Order existingTimeOrder = Cursors.getTimeOrdering(buildSpec.getPreferredOrdering());
+    final List<OrderBy> orderBy;
+    if (Order.NONE.equals(existingTimeOrder)) {
+      orderBy = ImmutableList.copyOf(
+          Iterables.concat(Cursors.ascendingTimeOrder(), buildSpec.getPreferredOrdering())
+      );
+    } else {
+      orderBy = buildSpec.getPreferredOrdering();
+    }
+
     final ImmutableList.Builder<String> groupingColumnsBuilder = ImmutableList.builder();
     groupingColumnsBuilder.add(granularityVirtual.getOutputName());
     if (buildSpec.getGroupingColumns() != null) {
@@ -125,6 +139,7 @@ public class Granularities
     return CursorBuildSpec.builder(buildSpec)
                           .setVirtualColumns(virtualColumns)
                           .setGroupingColumns(groupingColumnsBuilder.build())
+                          .setPreferredOrdering(orderBy)
                           .build();
   }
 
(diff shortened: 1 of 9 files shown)


processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
```diff
@@ -182,9 +182,15 @@ public Schema(
         @JsonProperty("ordering") List<OrderBy> ordering
     )
     {
+      if (name == null) {
+        throw DruidException.defensive("projection schema name cannot be null");
```
kfaraz: Should these be defensive or invalid input?

processing/src/test/java/org/apache/druid/segment/AggregateProjectionMetadataTest.java
```diff
@@ -30,19 +30,19 @@
...
 import java.util.SortedSet;
 
-public class AggregateProjectionMetadataTest extends InitializedNullHandlingTest
+class AggregateProjectionMetadataTest extends InitializedNullHandlingTest
 {
   private static final ObjectMapper JSON_MAPPER = TestHelper.makeJsonMapper();
 
   @Test
-  public void testSerde() throws JsonProcessingException
+  void testSerde() throws JsonProcessingException
```
kfaraz, line 47: Just curious, any specific reason for removing the `public` from the class and methods?

kfaraz (commented):
Changes look good, ~but I couldn't find the place that validates the projection granularity vs segment granularity.~

Never mind, looking at the `DataSchema` class now.

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
```diff
@@ -131,13 +134,32 @@ public DataSchema(
     // this validation is not necessarily going to be able to catch everything. It will run again in getDimensionsSpec.
     computeAndValidateOutputFieldNames(this.dimensionsSpec, this.aggregators);
 
+
```
kfaraz, line 142: Nit: extra line.

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
clintropolis: these are defensive because this object is created internally from `AggregateProjectionSpec` which is the user facing object that does invalid input

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
```diff
@@ -131,13 +134,32 @@ public DataSchema(
...
+    if (this.projections != null) {
+      for (AggregateProjectionSpec projection : this.projections) {
+        final AggregateProjectionMetadata.Schema schema = projection.toMetadataSchema();
+        if (schema.getTimeColumnName() == null) {
+          continue;
+        }
+        final Granularity projectionGranularity = schema.getGranularity();
+        if (this.granularitySpec.getSegmentGranularity().isFinerThan(projectionGranularity)) {
+          throw InvalidInput.exception(
+              "Projection[%s] has granularity[%s] must be finer than or equal to segment granularity[%s]",
```
kfaraz: Reads a little odd:

```suggestion
              "Projection[%s] has granularity[%s] which is coarser than segment granularity[%s]",
```

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
```diff
@@ -131,13 +134,32 @@ public DataSchema(
...
+
     if (this.granularitySpec.isRollup() && this.aggregators.length == 0) {
       log.warn(
           "Rollup is enabled for dataSource [%s] but no metricsSpec has been provided. "
           + "Are you sure this is what you want?",
           dataSource
       );
     }
+
+    if (this.projections != null) {
```
kfaraz: nit: may be this part can be put in a separate method?

processing/src/test/java/org/apache/druid/segment/AggregateProjectionMetadataTest.java
clintropolis: not really, just they also aren't needed so just went ahead and did it while changing to use junit5 stuff :shrug:

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
clintropolis: oops, i agree it needs the "which:, but i think "finer than or equal to" makes it more obvious what you need to do as a user since that is the change you need to make to the projection granularity

kfaraz: Yeah, that works too!

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
```diff
@@ -73,9 +73,12 @@ public AggregateProjectionSpec(
       @JsonProperty("aggregators") @Nullable AggregatorFactory[] aggregators
   )
   {
+    if (name == null) {
```
cecemei: nit: maybe `StringUtils.isEmpty` to prevent name being empty string as well?

cecemei (approved):

clintropolis:
CI failure seems unrelated to this PR (seeing it in other places too)

PR #18275, opened 2025-07-17
[Backport] Fix concurrent append to interval with unused segments

capistrant:
Backport of #18230 to 34.0.0.


capistrant:
failing ITs are flakes that also are in master that need fix + backport

PR #18276, opened 2025-07-17
[Backport] remove size checks from IncrementalIndex to simplify logic

capistrant:
Backport of #18219 to 34.0.0.


capistrant:
same thing here, the two ITs also fail/flake in master and will need fix and backport.

PR #18281, opened 2025-07-18
Add MV_FILTER_REGEX SQL function

jtuglu1:
Fixes #XXXX.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/VirtualColumn.java
+++ b/processing/src/main/java/org/apache/druid/segment/VirtualColumn.java
@@ -40,6 +40,7 @@ import org.apache.druid.segment.vector.VectorValueSelector;
 import org.apache.druid.segment.virtual.ExpressionVirtualColumn;
 import org.apache.druid.segment.virtual.FallbackVirtualColumn;
 import org.apache.druid.segment.virtual.ListFilteredVirtualColumn;
+import org.apache.druid.segment.virtual.RegexFilteredVirtualColumn;
 
 import javax.annotation.Nullable;
 import java.util.List;
@@ -56,7 +57,8 @@ import java.util.List;
 @JsonSubTypes(value = {
     @JsonSubTypes.Type(name = "expression", value = ExpressionVirtualColumn.class),
     @JsonSubTypes.Type(name = "fallback", value = FallbackVirtualColumn.class),
-    @JsonSubTypes.Type(name = "mv-filtered", value = ListFilteredVirtualColumn.class)
+    @JsonSubTypes.Type(name = "mv-filtered", value = ListFilteredVirtualColumn.class),
+    @JsonSubTypes.Type(name = "regex-filtered", value = RegexFilteredVirtualColumn.class)
 })
 public interface VirtualColumn extends Cacheable
 {
--- a/processing/src/main/java/org/apache/druid/segment/virtual/RegexFilteredVirtualColumn.java
+++ b/processing/src/main/java/org/apache/druid/segment/virtual/RegexFilteredVirtualColumn.java
@@ -1,0 +1,172 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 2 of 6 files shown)


processing/src/main/java/org/apache/druid/query/expression/RegexpLikeExprMacro.java
```diff
@@ -45,49 +45,92 @@
...
+    if (ExprUtils.isStringLiteral(args.get(1))) {
+      return new RegexpLikeExpr(args);
+    } else {
+      return new RegexpLikeDynamicExpr(args);
+    }
+  }
 
-    final Expr arg = args.get(0);
-    final Expr patternExpr = args.get(1);
+  abstract class BaseRegexpLikeExpr extends ExprMacroTable.BaseScalarMacroFunctionExpr
```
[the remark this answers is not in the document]
jtuglu1: cannot be static

jtuglu1:
@clintropolis I may still cleanup the `RegexpLikeMacro` logic a bit to share more validations/checks, but LMK what you think otherwise.

processing/src/main/java/org/apache/druid/segment/VirtualColumn.java
```diff
@@ -56,7 +58,9 @@
 @JsonSubTypes(value = {
     @JsonSubTypes.Type(name = "expression", value = ExpressionVirtualColumn.class),
     @JsonSubTypes.Type(name = "fallback", value = FallbackVirtualColumn.class),
-    @JsonSubTypes.Type(name = "mv-filtered", value = ListFilteredVirtualColumn.class)
+    @JsonSubTypes.Type(name = "mv-filtered", value = ListFilteredVirtualColumn.class),
+    @JsonSubTypes.Type(name = "regex-filtered", value = RegexFilteredVirtualColumn.class),
+    @JsonSubTypes.Type(name = "prefix-filtered", value = PrefixFilteredVirtualColumn.class)
```
clintropolis: while they probably work ok, i'm not sure it makes sense to use these virtual columns if not using multi-value strings,  so the type names of these probably should include the `mv-` prefix?

clintropolis (approved):
overall lgtm, nice to have these other `DimensionSpec` migrated over to virtual columns. Someday I would like do deprecate `DimensionSpec` other than `DefaultDimensionSpec` and do all of the others as virtual columns, so this aligns well with that.

The virtual columns are missing an implementation of `getIndexSupplier`, so using them for filtering directly is probably sub-optimal, however that is also probably ok because i think the primary use of these functions are so you can tidy up an MVD and not see seemingly unrelated values when using a filter on the column itself due to how filters behave on mvds (if any value in the row matches the whole row matches, and when using grouping the implicit unnest results in the other non-matching row values appearing in results). So like as long as people are just using the `MV_` functions in the select clause and just using the column itself and the regular filter in the where clause it shouldn't be a big deal.

processing/src/main/java/org/apache/druid/segment/VirtualColumn.java
jtuglu1: Sure, let me make that change.

processing/src/main/java/org/apache/druid/query/expression/RegexpLikeExprMacro.java
```diff
@@ -45,49 +45,92 @@ public String name()
...
-
-        if (s == null) {
-          // True nulls do not match anything. Note: this branch only executes in SQL-compatible null handling mode.
-          return ExprEval.ofLong(null);
-        } else {
-          final Matcher matcher = pattern.matcher(s);
-          return ExprEval.ofLongBoolean(matcher.find());
-        }
+      final Expr patternExpr = args.get(1);
+      if (patternExpr.getLiteralValue() == null) {
```
maytasm: nit: use ExprUtils.isStringLiteral() method

PR #18288, opened 2025-07-18
[Backport] validate projection granularity is finer than or equal to segment granularity

capistrant:
Backport of #18273 to 34.0.0.


capistrant:
failing IT is an unrelated issue in master that needs to be fixed.

PR #18239, opened 2025-07-11
By default dont allow index_hadoop tasks to run on a cluster, forcing operators to acknowledge that they are using a deprecated feature

capistrant:
### Description

Druid 34 publicly deprecating `index_hadoop` type tasks. The eventual removal of `index_hadoop` is a big change and will force any users of the task type to make a large change to their cluster operations. This PR aims to ensure no operator is caught off guard by the deprecation and planned removal. It will fail `index_hadoop` tasks with an error stating the why, unless the operator updates their runtime configs to allow `index_hadoop`

#### Alternate approach

I'm sure we could shift the failure left and fail before the overlord even tries to submit the task.

#### Release note

Druid cluster operators must opt-in to using the now deprecated `index_hadoop` task type in their Druid clusters. If you wish to be able to continue submitting `index_hadoop` typed tasks, please set the following runtime property to `true`: `druid.indexer.task.allowHadoopTaskExecution`

Note that this property needs to be set in the local context of your running ingest task. The easiest way to achieve this is to set it in `common.runtime.properties`.

<hr>

##### Key changed/added classes in this PR
 * `TaskConfig`
 * `IndexHadoopTask`

<hr>

This PR has:

- [x] been self-reviewed.
- [x] added documentation for new or modified features or behaviors.
- [x] a release note entry in the PR description.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [x] been tested in a test Druid cluster.

PR #18290, opened 2025-07-18
[Backport] By default dont allow index_hadoop tasks to run on a cluster, forcing operators to acknowledge that they are using a deprecated feature

capistrant:
Backport of #18239 to 34.0.0.


capistrant:
same broken IT that is not blocking this merge.

PR #18291, opened 2025-07-18
Fix race in SegmentLoadDropHandler.

gianm:
SegmentLoadDropHandler has a race that causes dropSegmentDelayMillis to not be respected. It happens in this sequence of events:

- removeSegment is called, which adds the segment to segmentsToDelete and schedules a future drop.
- addSegment is called, which removes the segment from segmentsToDelete and proceeds to load it.
- removeSegment is called again, which adds the segment back to segmentsToDelete, before the originally-scheduled drop executes.
- The originally-scheduled drop executes, sees the segment is in segmentsToDelete, and removes it. This could happen immediately after the second call to removeSegment, defeating dropSegmentDelayMillis.

This can occur in practice when the coordinator period is less than the dropSegmentDelayMillis. By default, it's not (default is 1m Coordinator period and 30s dropSegmentDelayMillis).

This patch addresses it by ensuring that delayed drops only happen if the entry in segmentDropLatches corresponds to the removeSegment call that scheduled the drop.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/coordination/SegmentLoadDropHandler.java
+++ b/server/src/main/java/org/apache/druid/server/coordination/SegmentLoadDropHandler.java
@@ -45,7 +45,7 @@ import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/druid/server/coordination/SegmentLoadDropHandler.java
```diff
@@ -216,21 +208,30 @@ void removeSegment(
...
           }
         }
         catch (Exception e) {
           log.makeAlert(e, "Failed to remove segment! Possible resource leak!")
              .addData("segment", segment)
              .emit();
         }
+        finally {
+          dropLatch.doneDropping();
+          segmentDropLatches.remove(segment, dropLatch);
```
kfaraz, line 233: Based on the changes made to `addSegment` and `removeSegment`, I think we can get away with calling `Map.remove(key)` here instead of `Map.remove(key, value)`.

kfaraz (approved):
Looks good, left minor comment.
Should we add an embedded test to verify the fix?

gianm:
> Should we add an embedded test to verify the fix?

I tried, but had a hard time writing one that actually did lots of moves. I tried adding 3 historicals, but even so the system ended up deciding not to move very much. I am not entirely sure what set of conditions causes that to happen.

kfaraz:
> I tried, but had a hard time writing one that actually did lots of moves. I tried adding 3 historicals, but even so the system ended up deciding not to move very much. I am not entirely sure what set of conditions causes that to happen.

Ah, I see. Since we want to test a drop-add-drop sequence, probably even constantly updating load rules with fast coordinator cycles could do the trick with a single historical.

PR #18292, opened 2025-07-19
some data segment change

cecemei:
### Description

 
This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/task/batch/parallel/PartialSegmentMergeTask.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/task/batch/parallel/PartialSegmentMergeTask.java
@@ -55,6 +55,7 @@ import org.apache.druid.segment.metadata.CentralizedDatasourceSchemaConfig;
 import org.apache.druid.segment.metadata.FingerprintGenerator;
 import org.apache.druid.segment.realtime.appenderator.TaskSegmentSchemaUtil;
 import org.apache.druid.timeline.DataSegment;
+import org.apache.druid.timeline.SegmentId;
 import org.apache.druid.timeline.partition.ShardSpec;
 import org.joda.time.Interval;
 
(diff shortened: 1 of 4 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/common/task/batch/parallel/PartialSegmentMergeTask.java
```diff
@@ -286,24 +287,24 @@ private DataSegmentsWithSchemas mergeAndPushSegments(
...
                                                .map(AggregatorFactory::getName)
                                                .collect(Collectors.toList());
+        SegmentId segmentId = SegmentId.of(
+            getDataSource(),
+            interval,
+            Preconditions.checkNotNull(AbstractBatchIndexTask.findVersion(
+                intervalToVersion,
+                interval
+            ), "version for interval[%s]", interval),
+            null
```
clintropolis: nit: this just calls the version of `of` which takes `partitionNum` with a `0` argument, it might be clearer to do that instead

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
```diff
@@ -113,37 +113,6 @@ public static class PruneSpecsHolder
   private final CompactionState lastCompactionState;
   private final long size;
 
-  /**
-   * @deprecated use {@link #builder(SegmentId)} or {@link #builder(DataSegment)} instead.
-   */
-  @Deprecated
-  public DataSegment(
```
clintropolis, line 120: do we intend to eventually remove the other `@Deprecated` constructor too? i don't expect that in this PR since it is fixing a bug and the other one has quite a lot more callers, just wondering.

clintropolis (approved):

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
cecemei: I think we can fix it as we go along, removing it here since it seems easy here. Definitely no constructor usage for new code.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/batch/parallel/PartialSegmentMergeTask.java
cecemei: this is done

PR #18274, opened 2025-07-17
Address NPE in CloneHistoricals duty

kfaraz:
### Description

Historical cloning can sometimes cause segment downloads to fail due to the load spec being null.
Example:

```
org.apache.druid.segment.loading.SegmentLoadingException: Exception loading segment[xyz]
	at org.apache.druid.server.coordination.SegmentLoadDropHandler.addSegment(SegmentLoadDropHandler.java:177) ~[druid-server-2025.10.0-iap.jar:2025.10.0-iap]
	at org.apache.druid.server.coordination.SegmentLoadDropHandler$1.lambda$addSegment$1(SegmentLoadDropHandler.java:333) ~[druid-server-2025.10.0-iap.jar:2025.10.0-iap]
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) [?:?]
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
	at java.base/java.lang.Thread.run(Thread.java:840) [?:?]
Caused by: java.lang.NullPointerException: Cannot invoke "org.apache.druid.segment.loading.LoadSpec.loadSegment(java.io.File)" because "loadSpec" is null
	at org.apache.druid.segment.loading.SegmentLocalCacheManager$SegmentCacheEntry.loadInLocation(SegmentLocalCacheManager.java:844) ~[druid-server-2025.10.0-iap.jar:2025.10.0-iap]
	at org.apache.druid.segment.loading.SegmentLocalCacheManager$SegmentCacheEntry.loadInLocationWithStartMarker(SegmentLocalCacheManager.java:824) ~[druid-server-2025.10.0-iap.jar:2025.10.0-iap]
	at org.apache.druid.segment.loading.SegmentLocalCacheManager$SegmentCacheEntry.mount(SegmentLocalCacheManager.java:751) ~[druid-server-2025.10.0-iap.jar:2025.10.0-iap]
	at org.apache.druid.segment.loading.SegmentLocalCacheManager.assignLocationAndMount(SegmentLocalCacheManager.java:593) ~[druid-server-2025.10.0-iap.jar:2025.10.0-iap]
[...]

PR #18293, opened 2025-07-19
[Backport] Address NPE in CloneHistoricals duty

capistrant:
Backport of #18274 to 34.0.0.


capistrant:
going to merge for now. we can rew-write 34 history if we want to handle my test changes differently. I **think** I kept the test the same in terms of validating the right thing

kfaraz:
Changes look good.

PR #18294, opened 2025-07-19
[Backport] Fixing deadlock in MSQ worker retry logic.

capistrant:
Backport of #18254 to 34.0.0.


capistrant:
failing IT is due to docker reasons in and will have to be fixed in master

PR #18082, opened 2025-06-05
Allow multiple stream supervisors to ingest into same datasource

jtuglu-netflix:
Fixes https://github.com/apache/druid/issues/18008.

Allows N ≥ 1 supervisors to ingest into the same Druid datasource.

#### Architecture

Instead of choosing to multiplex multiple region consumers onto a single supervisor, I opted to go with a multi-supervisor approach, for the following reasons:

1. Scaling – we can horizontally scale each supervisor's tasks independently with the lag in that region (especially useful for cross-region reads where the consumer poll() tput is much lower).
2. Spec changes – spec changes local to a single region can be done independent of the other regions, preventing unnecessary lag.
3. Flexible design – this could support further relationship between supervisor/datasources in the future, i.e M:N supervisor:datasource vs the current 1:1 (and with this change M:1).

#### Spec/Config Changes

1. `SeekableStreamSupervisorSpec` now has a proper `id` field which is checked for uniqueness within the `SupervisorManager` [here](https://github.com/apache/druid/blob/1e87ec7aa9151b3094ad432fbea80ce07e6f2260/indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java#L491). TBD: this check could further be enhanced by making the PK on the `druid_dataSource` table the new column `supervisorId`. For backwards compatibility, this ID will default to the datasource the supervisor is ingesting into (current behavior). 

2. `SeekableStreamIndexTask` now has a `supervisorId` field in the root json blob which is used to map a task back to its supervisor. [This](https://github.com/apache/druid/blob/1943f5e8bea5c1a828fc176a8f3712c5f745885f/extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisor.java#L161) is helpful in 2 primary cases:

    - When an (in)voluntary restart happens and all existing tasks need to be remapped by the overlord. 
    - When a task runner needs to publish its local segments, and must know the supervisor id of the supervisor it belongs to. 

For backwards compatibility, this field will default to `null`, in which case users of this field will default to the datasource. I believe this should cover the case where task specs are written to DB pre-upgrade, and post upgrade we need to read these tasks back into memory to submit, etc.

#### DB Schema Changes

No physical schema changes, but now, stream supervisors' id is being written to the `druid_dataSource` table's `dataSource` column. For context as to why this decision was made, see [here](https://github.com/apache/druid/pull/18082#issuecomment-2969607160).

[...]

PR #18298, opened 2025-07-21
Fix supervisor to tasks view naviagation in the console

abhishekrb19:
With [apache/druid#18082](https://github.com/apache/druid/pull/18082), the supervisor ID is no longer the same as the datasource name, which breaks the navigation from the supervisors page to the corresponding tasks view.

CleanShot 2025-07-21 at 08 48 38@2x
asks view breaks because currently.

Fix:

Use the `groupId` filter instead of datasource name when navigating from the supervisors to tasks view. This filter is used by other task types too. I think another alternative and cleaner approach would be to add the supervisor ID to the `sys.tasks` table and filter by it explicitly, but that would be a larger server-side change.

CleanShot 2025-07-21 at 08 49 18@2x

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/web-console/src/console-application.tsx
+++ b/web-console/src/console-application.tsx
@@ -408,7 +408,7 @@ export class ConsoleApplication extends React.PureComponent<
         goToDatasource={this.goToDatasources}
         goToQuery={this.goToQuery}
         goToStreamingDataLoader={this.goToStreamingDataLoader}
-        goToTasks={this.goToTasksWithDatasource}
+        goToTasks={this.goToTasksWithTaskGroupId}
         capabilities={capabilities}
       />,
     );
--- a/web-console/src/views/supervisors-view/supervisors-view.tsx
+++ b/web-console/src/views/supervisors-view/supervisors-view.tsx
@@ -782,7 +782,10 @@ export class SupervisorsView extends React.PureComponent<
     switch (supervisor.type) {
       case 'kafka':
       case 'kinesis':
-        goToTasks(`index_${supervisor.type}_${supervisor.supervisor_id}`, `index_${supervisor.type}`);
+        goToTasks(
+          `index_${supervisor.type}_${supervisor.supervisor_id}`,
+          `index_${supervisor.type}`,
+        );
         return;
 
       case 'autocompact':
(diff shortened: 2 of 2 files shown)


abhishekrb19:
Tagging this with the 34.0.0 milestone since it’s a bug introduced with [#18082](https://github.com/apache/druid/pull/18082). The fix is small enough that it’s likely worth backporting.

cc: @capistrant

kfaraz (approved):
LGTM 🚀 
Verified in an embedded test.

cryptoe:
@abhishekrb19 Since https://github.com/apache/druid/pull/18082 is experimental, I think its okay if we do this patch in Druid 35. 
At this point i guess  backports like query correctness, regressions or security hotfixes may interrupt the RC process. 
I will let @capistrant chime in since he is the release manager.

capistrant:
I'm inclined to avoid any more creep into Druid34 wherever possible as I was hoping to move on to RC process ASAP. @abhishekrb19 do you have a feel for how impactful this is in the community? Is it just a UX issue with a work-a-round making a few extra clicks?

abhishekrb19:
> @abhishekrb19 Since https://github.com/apache/druid/pull/18082 is experimental, I think its okay if we do this patch in Druid 35.

The broken "Go to tasks" navigation for running tasks from the supervisors page in the console affects all new supervisors created after [#18082](https://github.com/apache/druid/pull/18082), not just those with multiple supervisors. It's a UX inconvenience, as the default filters won't return any results, but it can be worked around with a few extra clicks/keystrokes in the task view.
>I'm inclined to avoid any more creep into Druid34 wherever possible as I was hoping to move on to RC process ASAP. 

@capistrant I've removed the milestone tag, but please feel free to backport it to 34.0.0 as you see fit.

cryptoe:
Thanks @abhishekrb19 for clarifying. 
So judging from your statement, it effects all new supervisors so a major ux regression.

capistrant:
@abhishekrb19 Sorry I misunderstood the scope as well. Added back the 34.0.0 label since this is not just experimental supervisor facing

PR #18226, opened 2025-07-09
Fail ingestion tasks when OffsetOutOfRangeException is raised

FrankChen021:
Fixes #14344 

### Description

When `OffsetOutOfRangeException` is raised, previously, it waits until task completes. This brings problems for operators that it may take long time to observe such exception is raised.

Now the exception handling is improved if auto reset is not enabled:
1. all ingested messages before the expired offset will be persisted and published
2. task will be marked as FAIL so that operators/users can take actions against problematic topics as early as possible

#### Release note
Tasks that encounters `OffsetOutOfRangeException` exception will fail immediately

####
Follow these steps to test:
1. set up a topic in Kafka
2. config the topic's retention time/retension size to a very small value
    ```bash
    ./kafka-configs.sh --bootstrap-server localhost:9092 --entity-type topics --entity-name a --alter --add-config retention.ms=10000 retention.bytes=16384
    ```
4. set up a kafka ingestion task in Druid
6. use `kafka-producer-perf-test.sh` to generate events to the test topic in very large throughput

```bash
sh kafka-producer-perf-test.sh --topic a --print-metrics --payload-file payload.txt --producer-props bootstrap.servers=localhost:9092 --num-records 1000000000 --throughput 10000000
```

after a while when kafka server deletes offset, we can see from task log, task pushed segments and fail at last

```
2025-07-09T07:53:14,267 INFO [task-runner-0-priority-0] org.apache.druid.segment.realtime.appenderator.BaseAppenderatorDriver - Pushing [1] segments in background
2025-07-09T07:53:14,267 INFO [task-runner-0-priority-0] org.apache.druid.segment.realtime.appenderator.BaseAppenderatorDriver - Pushing segments: [a_2025-05-01T00:00:00.000Z_2025-05-01T01:00:00.000Z_2025-07-08T06:55:43.564Z_7]
2025-07-09T07:53:14,267 INFO [task-runner-0-priority-0] org.apache.druid.segment.realtime.appenderator.StreamAppenderator - Persisted rows[0] and (estimated) bytes[0]
[...]

PR #18301, opened 2025-07-21
Improve offset auto reset

### LINKED ISSUE #18282: Correctly reset kafka offset if auto reset is enabled
This problem has been there for a very long time. See https://github.com/apache/druid/issues/11658

When `OffsetOutOfRangeException` is raised, and if auto reset is not enabled(which is the default one), the behavior is fixed(see https://github.com/apache/druid/pull/18226) so that all data ingested before the exception WILL be handled, and task will FAIL. This is similar to kafka client behavior if 'auto.offset.reset' is not given.

However, when auto reset is enabled, current behaviour does not work as expected.
There're several bugs there:
1. task resets the offset by using the expired offset, which is wrong, instead, the earliest offset should be used
2. overlord kills task after offset is reset, and all ingested data lost
3. overlord can reset the offset when it starts a new task. and this time, overlord will CLEAR(which is wrong) the offset stored in the metadatastore, and throws a exception, which delays the start of task

as we can see that, for the auto reset, both task and overlord are involved in making decision of offset reset which makes it overcomplicated.

The following sequence diagram shows how the task and overlord works.

```mermaid
sequenceDiagram
    participant Task
    participant Overlord
    participant Metdatastorage

    alt Reset Offset from tasks
      Task->>Overlord: 1.Reset Offset with expired offset (WRONG)
      Overlord->>Metdatastorage: 2.Read offset from metadata storage
      Overlord->>Overlord: 3.Check if offset is available
      Overlord->>Overlord: 4.Reset offset if NOT available (with current offset PLUS offset from metadata storage)
      Overlord->>Overlord: 5.Throw exception and wait for next retry if offset is available
      Overlord->>Task: Kill tasks
    end    
    
    alt Starting a new task
      Overlord->>Overlord: 6.Start a new task
      Overlord->>Metdatastorage: 7.Read offset from metadata storage
      Overlord->>Overlord: 8.Check if offset is available
      Overlord->>Overlord: 9.Reset offset if NOT available (with current offset MINUS(which is WRONG) offset from metadata storage)
      Overlord->>Overlord: 10.Throw exception and wait for next retry to start a task(introducing delay)
    end
```

1. Tasks reset the offset by using 'nextOffset' which is currently expired offset. this is wrong
https://github.com/apache/druid/blob/59d257816b85dbeeca336b8e25d341d67bbc5697/extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/IncrementalPublishingKafkaIndexTaskRunner.java#L134-L155

2. overlord throws an exception after reset, which delays the start of a task
https://github.com/apache/druid/blob/ec25d8706cb63672c9da8dec255b79cba46288cc/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java#L3996-L4006

3. overlord clears the offset when reseting offset. 
https://github.com/apache/druid/blob/ec25d8706cb63672c9da8dec255b79cba46288cc/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java#L1900-L1906

in above code block, it call 'minus' method instead of 'plus' method. the `resetOffsetsInternal` used by task correctly calls the `plus` method.
In next turn, overlord find that there's no offset for this partition, it reset the offset to earliest or latest which is based on `useEarliestOffset` property. 
https://github.com/apache/druid/blob/ec25d8706cb63672c9da8dec255b79cba46288cc/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java#L1965-L1975

As a subsquence, when `OffsetOutOfRangeException` is raised, in extreme cases, new tasks are continously forked but all fail.

# Proposed Change
To make it simple and clear, the key point is that, ONLY overlord is responsible for reseting the offset.

1. remove offset reset logic from task. when OffsetOutOfRangeException is raised, it fail but commits all ingested data as fixed by https://github.com/apache/druid/pull/18226
2. when supervisor tries to start a new task, it will know that last stored offset is not available, and tries to reset
    1. use the EARLIEST offset from stream(like Kafka) to reset offset
    2. correctly merge the reset offset and stored offset before update offset in metadata storage. this ensure that the overlord always resets the offset to earliest so that less data are lost.
    3. issue an alert from overlord that offset has been reset
    4. after reset, return the this offset so that supervisor can immediately start task from the reset offset

FrankChen021:
Fixes #18282

### Description

Improve the auto reset logic as proposed in the above issue:

1. only supervisor is responsible for auto reset. task never triggers the auto reset flow
2. task always throws `OffsetOutOfRangeException` if this exception is raised, but will publish ingested message under such case
3. supervisor resets offset to earliest position if it detects offset of a particular partition is unavailable
4. supervisor starts task more quickly if it needs to reset offsets
5. update the obscure error message like `DataSourceMetadata is not found while reset` in task report to make it intuitive when supervisor kills a task because of offset reset

### NOTE
this relies on #18226, which publishes segments when OffsetOutOfRangeException is raised.

This PR has:

- [X] been self-reviewed.

### DIFF
--- a/extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskRunner.java
+++ b/extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskRunner.java
@@ -35,24 +35,18 @@ import org.apache.druid.indexing.seekablestream.common.OrderedPartitionableRecor
 import org.apache.druid.indexing.seekablestream.common.OrderedSequenceNumber;
 import org.apache.druid.indexing.seekablestream.common.RecordSupplier;
 import org.apache.druid.indexing.seekablestream.common.StreamPartition;
-import org.apache.druid.java.util.common.ISE;
 import org.apache.druid.java.util.emitter.EmittingLogger;
-import org.apache.druid.utils.CollectionUtils;
 import org.apache.kafka.clients.consumer.OffsetOutOfRangeException;
-import org.apache.kafka.common.TopicPartition;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.validation.constraints.NotNull;
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.concurrent.TimeUnit;
 
 /**
  * Kafka indexing task runner that supports incremental segment publishing.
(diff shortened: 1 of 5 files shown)


jtuglu1:
@FrankChen021 any thoughts here?

PR #18265, opened 2025-07-16
Add DruidContainer to completely phase out revised and standard ITs

kfaraz:
### Changes

- Add extension `druid-testcontainers` that can be later published as an [individual module](https://testcontainers.com/modules/) for `Testcontainers`
- Add `DruidContainer` - `Testcontainer` impl for running individual Druid services
- Add `DruidContainerResource` to allow use of `DruidContainer` in embedded cluster tests
- Make minor modifications to `EmbeddedDruidServer` and `EmbeddedDruidCluster` to support
running both embedded servers and container-based services in the same cluster
- Add job to run the `*DockerTest`s (2 right now, this number should remain small over time).

Note: __Usage of `DruidContainer` will not be the norm but the exception to test out backward compatibility
and/or Docker related changes only. All other use cases should continue to use `EmbeddedDruidServers`.__

### Pending items

- Populate all the test methods using the builders in #18207 once it is merged

### Advantages

- Can be used to run embedded tests against the distribution Docker image as opposed to the current ITs
which use a custom IT-only image
- Takes us 1 step closer to completely removing the old IT frameworks
- Leverages all the benefits of the embedded test framework while also testing out the Docker containers
- Can be easily used for backward compatibility tests

### Next steps

- Once we have phased out the old ITs, all integration tests will be based on `EmbeddedClusterTestBase`
- Most of these tests will just use `EmbeddedDruidServer` (and required containers only, like `KafkaResource`)
- These tests will use Indexer as worker as they are faster and lighter than MiddleManagers
- There will be only a small number of tests that use `DruidContainer`, some of which will use MiddleManagers

<hr>

This PR has:

- [ ] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
[...]

PR #18302, opened 2025-07-21
Add DruidContainers to run docker tests with embedded-test framework

kfaraz:
This patch is based on the changes originally proposed in #18265.
Since that PR had a very long history of tuning the workflows (sincere thanks to @Akshat-Jain for the assistance!),
I have created a fresh PR instead to simplify reviews.

The old PR had only a couple of review comments which have already been answered.
https://github.com/apache/druid/pull/18265#discussion_r2214466052
https://github.com/apache/druid/pull/18265#discussion_r2214471841

### Summary

- Support running Druid Docker containers in embedded tests
- Use the Druid distribution image for these containers rather than a test-only image as previously done by ITs
- Add a `CliEventCollector` to monitor the containers and write efficient tests
- Run new job `docker-tests` in GHA which runs a few Docker-based tests

### How to run

From Druid root directory:

```php
$ mvn package -Pskip-static-checks -Pskip-tests -Pdist -DskipUTs=true -Dmaven.javadoc.skip=true -T1.0C
$ DRUID_DIST_IMAGE_NAME=apache/druid:34.0.0-rc1 .github/scripts/run_docker-tests
```
(This commands runs successfully for the ongoing release RC. Only 1 test for custom node role fails as it requires a minor change to the `druid.sh` script that is used to start Druid services in the containers)

### Main changes to review

- `DruidContainer`
- `DruidCommand`
- `DruidContainerResource`
- `DockerTestBase`
- `CliEventCollector`

#### Important scripts

- `.github/workflows/docker-tests.yml`
- `.github/scripts/run_docker-tests`

#### New test classes

- `IngestionDockerTest`
- `CustomNodeRoleDockerTest`
- `IngestionBackwardCompatibilityDockerTest`
- `HttpEmitterEventCollectorTest`

<details>
<summary><h3>Change details</h3></summary>

- Add extension `druid-testcontainers` that can be later published as an [individual module](https://testcontainers.com/modules/) for `Testcontainers`
- Add `DruidContainer` - `Testcontainer` impl for running individual Druid services
- Add `DruidContainerResource` to allow use of `DruidContainer` in embedded cluster tests
- Make minor modifications to `EmbeddedDruidServer` and `EmbeddedDruidCluster` to support
running both embedded servers and container-based services in the same cluster
- Remove `distribution-checks.yml` job
- Add a new job to run the the docker tests
- Move `CliCustomNodeRole` to `CliEventCollector`, which is a custom Druid node with command `eventCollector`.
- Remove `ITHighAvailabilityTest` and related files

Note: __Usage of `DruidContainer` will not be the norm but the exception to test out backward compatibility
and/or Docker related changes only. All other use cases should continue to use `EmbeddedDruidServers`.__

#### `CliEventCollector`

Both the standard ITs and revised ITs used a `CliCustomNodeRole` in `ITHighAvailabilityTest`
This patch removes all the code related to these.
`ITHighAvailabilityTest` has already been migrated to the embedded version, `HighAvailabilityTest`.

The `CliCustomNodeRole` has been renamed to `CliEventCollector` and moved to `testing-tools`
This class now serves a dual purpose:
- Verify custom node role behaviour in `CustomNodeRoleDockerTest`
- Run an event collector server which can be used in embedded tests.
   - The event collector receives metrics from other Druid services using the `HttpPostEmitter`.
   - It forwards the metrics to a `LatchableEmitter`.
   - Thus tests can wait on specific metric events being emitted from Druid services.
   - This is particularly useful while running Docker tests.
   - Using a `Cli*` provides a lot of wiring out of the box without having to write new code.

</details>
<details>
<summary><h3>Sample test run</h3></summary>

https://github.com/apache/druid/actions/runs/16416105923/job/46388247532?pr=18302

```java
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.druid.testing.embedded.docker.IngestionDockerTest
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 72.72 s -- in org.apache.druid.testing.embedded.docker.IngestionDockerTest
[INFO] Running org.apache.druid.testing.embedded.docker.IngestionBackwardCompatibilityDockerTest
[INFO] Running org.apache.druid.testing.embedded.docker.IngestionBackwardCompatibilityDockerTest$Apache31
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 82.56 s -- in org.apache.druid.testing.embedded.docker.IngestionBackwardCompatibilityDockerTest$Apache31
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 82.57 s -- in org.apache.druid.testing.embedded.docker.IngestionBackwardCompatibilityDockerTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0
```

</details>

### Advantages

- Can be used to run embedded tests against the distribution Docker image as opposed to the current ITs
which use a custom IT-only image
- Takes us 1 step closer to completely removing the old IT frameworks
- Leverages all the benefits of the embedded test framework while also testing out the Docker containers
- Can be easily used for backward compatibility tests

### Next steps

- As `druid-testcontainers` evolves, we can try to contribute it to the main `Testcontainers` repo
- Add a Druid command to run a single process Druid cluster. This would be particularly beneficial for users of the Druid testcontainer.
- Once we have phased out the old ITs, all integration tests will be based on `EmbeddedClusterTestBase`
- Most of these tests will just use `EmbeddedDruidServer` (and required containers only, like `KafkaResource`)
- These tests will use Indexer as worker as they are faster and lighter than MiddleManagers
- There will be only a small number of tests that use `DruidContainer`, some of which will use MiddleManagers

<hr>

This PR has:

- [ ] been self-reviewed.


.github/scripts/distribution_checks_env.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+export DRUID_DIST_IMAGE_NAME=apache/druid:35.0.0-SNAPSHOT
```
kgyrtkirk: doesn't feel right - you could use the workflow level `env` to declare exports

....or move the full logic into a script and run that...

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         pattern: [ "C*", "H*,U*,V*", "N*,Q*,S*", "B*,O*,R*", "G*,J*,K*", "F*,L*,M*", "A*,D*,I*,X*,Y*,Z*", "E*,P*,T*,W*"]
     uses: ./.github/workflows/worker.yml
     with:
-      script: .github/scripts/run_unit-tests -Dtest=!QTest,'${{ matrix.pattern }}' -Dmaven.test.failure.ignore=true
+      script: .github/scripts/run_unit-tests -Dtest=!QTest,!*DockerTest*,'${{ matrix.pattern }}' -Dmaven.test.failure.ignore=true
```
kgyrtkirk: I'm quilty of not moving the `QTest` into the `validation` phase ; which seem to have given rise to this pattern which could be considered bad practice

this is not really serious...it could be fixed separaetly - but nicely outlines your requirement that you just wanted to add 1 more test....

.github/workflows/distribution-checks.yml
```diff
@@ -38,4 +38,23 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build the Docker image
-        run: DOCKER_BUILDKIT=1 docker build -t apache/druid:tag -f distribution/docker/Dockerfile .
+        run: |
```
kgyrtkirk: I feel like what you are doing in this PR is superior to what this workflow was ever doing.

I think we could just remove this whole workflow...along with the `save/load` of the image and run those dockerized tests right after the image is built!

.github/workflows/docker-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+        if: ${{ failure() && steps.run-it.conclusion == 'failure' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: failure-docker-logs
+          path: docker-logs.tgz
+
+      - name: Collect service logs on failure
+        if: ${{ failure() && steps.run-it.conclusion == 'failure' }}
+        run: |
+          tar cvzf ./service-logs.tgz /tmp/embedded-tests
```
kgyrtkirk: I wonder why these are under `/tmp` and not under a `target` directory?

.github/workflows/docker-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: 17
+          cache: 'maven'
+      - name: Run *DockerTest*
+        id: run-it
+        run: |
+          source .github/scripts/distribution_checks_env.sh
+          .github/scripts/run_unit-tests -Dtest=*DockerTest* -Ddruid.testing.docker.image=$DRUID_DIST_IMAGE_NAME
```
kgyrtkirk: instead of trying to reuse the `run_unit-tests` script; you could possibly make a separate one for this; and run that (you could still invode `run_unit-tests` from there if you want) 

you should preferrably also use the `worker.yml` to run it....that way if you want to test the workflow you could just try to run that script on your own machine!

.github/workflows/docker-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+        timeout-minutes: 60
+
+      - name: Collect docker logs on failure
+        if: ${{ failure() && steps.run-it.conclusion == 'failure' }}
+        run: |
+          mkdir docker-logs
+          for c in $(docker ps -a --format="{{.Names}}")
+          do
+            docker logs $c > ./docker-logs/$c.log
+          done
```
kgyrtkirk, line 61: can you integrate this into the `worker.yml` somehow: put it inside a script and configure the artifact collector to save that as well..

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+  private String containerDirectory;
+
+  private MountedDir indexerLogsDeepStorageDirectory;
+  private MountedDir serviceLogsDirectory;
+  private MountedDir segmentDeepStorageDirectory;
+
+  DruidContainerResource(DruidCommand command)
+  {
+    this.name = StringUtils.format(
+        "container_%s_%d",
```
kgyrtkirk, line 91: not matters much - but I feel like `container_` doesn't really gives a hint what this is...would be better if it would hint that its being created by Druid's embedded test framework....don't have a good suggestion...maybe its not that important

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  @Override
+  public void beforeStart(EmbeddedDruidCluster cluster)
+  {
+    this.cluster = cluster;
+
+    // Mount directories used by the entire cluster (including embedded servers)
+    this.segmentDeepStorageDirectory = new MountedDir(
+        "/tmp/druid/deep-store",
```
kgyrtkirk: why it needs to be under `tmp` if its inside the container? it could even be `/deep-storage`

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+    );
+    this.indexerLogsDeepStorageDirectory = new MountedDir(
+        "/tmp/druid/indexer-logs",
+        cluster.getTestFolder().getOrCreateFolder("indexer-logs").getAbsolutePath()
+    );
+
+    // Mount directories used by this container for easier debugging with service logs
+    this.containerDirectory = cluster.getTestFolder().getOrCreateFolder(name).getAbsolutePath();
+    this.serviceLogsDirectory = new MountedDir(
+        "/opt/druid/log",
```
kgyrtkirk: not sure about this folder layout....at first time I would never look at `/opt/` for logs...

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+  private final String name;
+  private final DruidCommand command;
+  private final Map<String, String> properties = new HashMap<>();
+  private final List<Integer> taskPorts = new ArrayList<>();
+
+  private int servicePort;
+  private DockerImageName imageName;
+  private EmbeddedDruidCluster cluster;
+
+  private String containerDirectory;
```
kgyrtkirk: its a directory - why it needs to be flattened out to be a `String` ?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+  @Override
+  public String toString()
+  {
+    return name;
+  }
+
+  private static class MountedDir
+  {
+    final String hostPath;
+    final String containerPath;
```
kgyrtkirk: these seems like paths - throwing away `File` and `Path` will make it harder to work with them....

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+
+    final DruidContainer container = new DruidContainer(command, imageName)
+        .withExposedPorts(exposedPorts.toArray(new Integer[0]))
+        .withCommonProperties(getCommonProperties())
+        .withServiceProperties(getServerProperties())
+        .withFileSystemBind(
+            segmentDeepStorageDirectory.hostPath,
+            segmentDeepStorageDirectory.containerPath,
+            BindMode.READ_WRITE
+        )
```
kgyrtkirk: you could make the `MountedDir` a first class citizen and not extract its belly...

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+  {
+    final List<Integer> exposedPorts = new ArrayList<>(taskPorts);
+    exposedPorts.add(servicePort);
+    log.info(
+        "Starting Druid container[%s] on port[%d] with mounted directory[%s] and exposed ports[%s].",
+        name, servicePort, containerDirectory, exposedPorts
+    );
+
+    final DruidContainer container = new DruidContainer(command, imageName)
+        .withExposedPorts(exposedPorts.toArray(new Integer[0]))
```
kgyrtkirk: there could be a method which accepts `List` ?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+        )
+        .waitingFor(Wait.forHttp("/status/health").forPort(servicePort));
+
+    // Bind the ports statically (rather than using a mapped port) to the same
+    // value used in `druid.plaintextPort` to make this node discoverable
+    // by other services (both embedded and dockerized).
+    List<String> portBindings = exposedPorts.stream().map(
+        port -> StringUtils.format("%d:%d", port, port)
+    ).collect(Collectors.toList());
+    container.setPortBindings(portBindings);
```
kgyrtkirk: there could be an `exposePorts` which accepts the list...so no `map` and other stuff is at the place its being used

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+
+    return container;
+  }
+
+  /**
+   * Writes the common properties to a file.
+   *
+   * @return Name of the file.
+   */
+  private Properties getCommonProperties()
```
kgyrtkirk: methodname and apidoc mismatch

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+    );
+
+    return commonProperties;
+  }
+
+  /**
+   * Writes the server properties to a file.
+   *
+   * @return Name of the file.
+   */
```
kgyrtkirk: the apidoc doesn't match the function's mission; and its also feels like its irrelevant

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,310 @@
...
+  private static void createLogDirectory(File dir)
+  {
+    try {
+      FileUtils.mkdirp(dir);
+      Files.setPosixFilePermissions(dir.toPath(), PosixFilePermissions.fromString("rwxrwxrwx"));
+    }
+    catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
```
kgyrtkirk, line 225: couldn't the `MountedDir` provide a place for this method? that way if someone needs similar in the future it might got reused...

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainers.java
```diff
@@ -0,0 +1,117 @@
...
+public final class DruidContainers
+{
+  private DruidContainers()
+  {
+    // no instantiation
+  }
+
+  /**
+   * Creates a new {@link DruidContainerResource} to run a Coordinator node.
+   */
```
kgyrtkirk: nit: this apidoc doesn't add any value beyond the function signature:
```
public static DruidContainerResource newCoordinator()
```

this is true for a lot of places....I think adding apidoc like that could be harmfull as they will become stale after some time.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainers.java
```diff
@@ -0,0 +1,117 @@
...
+   * Creates a new {@link DruidContainerResource} to run an Indexer node.
+   */
+  public static DruidContainerResource newIndexer()
+  {
+    return new DruidContainerResource(DruidCommand.INDEXER)
+        .addProperty("druid.lookup.enableLookupSyncOnStartup", "false")
+        .addProperty("druid.processing.buffer.sizeBytes", "50MiB")
+        .addProperty("druid.processing.numMergeBuffers", "2")
+        .addProperty("druid.processing.numThreads", "5");
+  }
```
kgyrtkirk: feels like the `DruidCommand`  was over and underused at the same time...
* it contains some stuff so `new DruidContainerResource(DruidCommand.INDEXER)` could create a valid object
* its underused as a good one will have these 4 properties set...

feels like there should be a `DruidOverlordContainer` / `DruidIndexerContainer` and similar....it could still be an innerclass here...

kgyrtkirk (commented):
left a few comments here and there...some of them might be considered personal preference/etc so not addressing all is ok as well :)

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: This will only ever be used with the embedded test framework. So, I am not sure if included an `embedded` or similar prefix would add much value either.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: Felt unnecessary since this is just used in one place here.
These are methods exposed by `GenericContainer` from `Testcontainers` library, not added by this PR.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: This is a method exposed by `GenericContainer` itself, not added by this PR.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: Since the `MountedDir` class is only needed inside the `DruidContainerResource`, I didn't create it as a top-level class. It is used only to invoke `GenericContainer.withFileSystemBind`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainers.java
kfaraz: Some of these properties might make it into DruidCommand itself if we feel that these are always needed out of the box whenever running a DruidContainer.

I have just kept the minimal set in DruidCommand for now but will double check to confirm if some of these can't be moved there right now.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: This is the location for log files used by the Druid Docker image by default.

.github/workflows/docker-tests.yml
kfaraz: > instead of trying to reuse the run_unit-tests script; you could possibly make a separate one for this; and run that (you could still invode run_unit-tests from there if you want)

Could you please elaborate? What would be the advantage of putting this in a separate script if we would still invoke `run_unit-tests` from there?

.github/workflows/docker-tests.yml
kfaraz: Oh, I think this is not needed anymore. I had used it while debugging the workflow.
Service logs are written inside a fresh `TestFolder` created by the embedded cluster.

kfaraz:
Thanks for the suggestions, @kgyrtkirk !

I have made the following changes:
- Moved all required properties of `DruidCommand` so that all users of `DruidContainer` can benefit from the default set of properties.
- Removed `distribution-checks.yml` workflow and `distribution_checks_env.sh` as you suggested
- Used `File` instead of `String` where applicable
- Used path `/druid/deep-store` on container

I am not very clear on how the usage of `worker.yml`/`run_unit-tests` script in the new workflow.
I am open to suggestions.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainers.java
kfaraz: Updated.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: updated.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: fixed

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: used `File` instead.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: used `File` instead.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: Used `/druid/deep-store` instead.

.github/workflows/docker-tests.yml
kfaraz: removed this step.

.github/workflows/distribution-checks.yml
kfaraz: fixed.

.github/scripts/distribution_checks_env.sh
kfaraz: removed.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainers.java
kfaraz: For now, I have moved all required properties to `DruidCommand` so that all users of `DruidContainer` can get those default properties out of the box.

.github/workflows/docker-tests.yml
kgyrtkirk: you could run it locally to reproduce what's happening on the CI; you could call `mvn` directly; but you will also need to build the image...

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kgyrtkirk: yes - but you are extending that class; `DruidContainer extends GenericContainer<DruidContainer>`
you could make it adapt better for our needs...so that its easier to read...no need to limit to `GenericContainer` api usage

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kgyrtkirk: yes - it is used for that ; but `MountedDir` seemed like a concept you added as it was usefull...
usage of those concepts could be promoted with methods/etc I think that makes it easier to work with it....

kfaraz: Fair enough. I guess we could add a method on `DruidContainer` which uses `MountedDir`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: Updated ,thanks for the suggestion!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: updated.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: added a new method.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionDockerTest.java
```diff
@@ -0,0 +1,296 @@
...
+import java.util.Set;
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * Runs some basic ingestion tests using {@link DruidContainers}.
+ */
+public class IngestionDockerTest extends EmbeddedClusterTestBase
+{
+  static {
+    System.setProperty("druid.testing.docker.image", "apache/druid:35.0.0-SNAPSHOT");
```
gianm: Setting this in a static initializer doesn't seem right, because it will be set when the class is loaded, not necessarily when the test runs. Maybe better to use `@BeforeAll`.

Although, it'd be even better to not have to set the property at all. Like, perhaps the property just sets a default, but tests are able to specify their own when constructing the `DruidContainer`. That's how it works for other kinds of testcontainers.

Also- about the value- do we need to change the value when Druid's version is upgraded? Is it possible to get it programmatically, such as from `DruidContainerResource.class.getPackage().getImplementationVersion()`? Should work when running from a jar, possibly not from an IDE.

Finally, if we are going to keep needing to set this property, better to refer to the name by `DruidContainerResource.PROPERTY_TEST_IMAGE` rather than hardcoding it.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionDockerTest.java
```diff
@@ -0,0 +1,296 @@
...
+        .addResource(broker1)
+        .addServer(overlordFollower)
+        .addServer(broker2)
+        .addServer(new EmbeddedRouter());
+  }
+
+  @Test
+  public void test_runIndexTask_andKillData()
+  {
+
```
gianm: a comment about why this is empty would be helpful (same goes for the others)

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionDockerTest.java
```diff
@@ -0,0 +1,296 @@
...
+    // Get the task statuses
+    List<TaskStatusPlus> taskStatuses = ImmutableList.copyOf(
+        (CloseableIterator<TaskStatusPlus>)
+            cluster.callApi().onLeaderOverlord(o -> o.taskStatuses(null, dataSource, 1))
+    );
+    Assertions.assertEquals(1, taskStatuses.size());
+    Assertions.assertEquals(TaskState.RUNNING, taskStatuses.get(0).getStatusCode());
+
+    // Verify the count of rows ingested into the datasource so far
+    // Assertions.assertEquals("10", cluster.runSql("SELECT COUNT(*) FROM %s", dataSource));
```
gianm: Please don't commit commented-out code.

extensions-core/druid-testcontainers/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+    <version>35.0.0-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
+  </parent>
+
+  <!-- This module should have minimal dependencies so that it can be used independently
+   by libraries and codebases outside Druid to run Testcontainers -->
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.druid</groupId>
+      <artifactId>druid-processing</artifactId>
```
gianm: `druid-processing` is a pretty big dependency with lots of transitive dependencies. It looks like it's just for `StringUtils` and `Logger`. It would be better to avoid those and remove this dependency. For `Logger`, you could either refrain from logging stuff, or you could use `org.slf4j.Logger` directly and depend on `slf4j-api`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,229 @@
...
+    this.name = StringUtils.format(
+        "container_%s_%d",
+        command.getName(),
+        SERVER_ID.incrementAndGet()
+    );
+    this.command = command;
+    addProperty("druid.host", EmbeddedDruidCluster.getDefaultHost());
+  }
+
+  public DruidContainerResource withImage(DockerImageName imageName)
```
gianm: To me a method named `withX` suggests a copy is being made, and the original object is not modified. But here the original object is in fact modified. IMO `setImage` or `useImage` would be clearer.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
```diff
@@ -0,0 +1,229 @@
...
+  {
+    this.imageName = imageName;
+    return this;
+  }
+
+  /**
+   * Uses the Docker test image specified by the system property
+   * {@link #PROPERTY_TEST_IMAGE} for this container.
+   */
+  public DruidContainerResource withTestImage()
```
gianm: Similar comment as above, IMO `usingTestImage()` is clearer.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionDockerTest.java
```diff
@@ -0,0 +1,296 @@
...
+ * Runs some basic ingestion tests using {@link DruidContainers}.
+ */
+public class IngestionDockerTest extends EmbeddedClusterTestBase
+{
+  static {
+    System.setProperty("druid.testing.docker.image", "apache/druid:35.0.0-SNAPSHOT");
+  }
+
+  // Druid Docker containers
+  protected final DruidContainerResource overlordLeader = DruidContainers.newOverlord().withTestImage();
```
gianm: I don't think we can be sure that this one is always the leader. It starts up first, but maybe something happens that causes a leader change. Or maybe for some reason the container version fails to gain leadership due to an error, and then the `overlordFollower` becomes leader.

It would help to add some check in the tests that confirms this one is actually leading. Ideally we check at the both beginning and end of each test.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionDockerTest.java
```diff
@@ -0,0 +1,296 @@
...
+        .addCommonProperty(
+            "druid.extensions.loadList",
+            "[\"druid-s3-extensions\", \"druid-kafka-indexing-service\", \"druid-multi-stage-query\"]"
+        )
+        .addResource(coordinator)
+        .addResource(overlordLeader)
+        .addResource(middleManager)
+        .addResource(historical)
+        .addResource(broker1)
+        .addServer(overlordFollower)
```
gianm: a comment about why we mix containers and noncontainers in this test would be helpful.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
```diff
@@ -231,13 +242,50 @@ public TestFolder getTestFolder()
...
+   * with {@link #getEmbeddedServiceHostname()}. Using the embedded URI ensures
+   * that the underlying service is reachable by both EmbeddedDruidServers and
+   * DruidContainers.
+   */
+  public String getEmbeddedConnectUri(String connectUri)
+  {
+    if (!hasDruidContainers) {
+      return connectUri;
+    } else if (connectUri.contains("localhost")) {
+      return StringUtils.replace(connectUri, "localhost", getEmbeddedServiceHostname());
```
gianm: Seeing URI manipulations like this stresses me out ;P

It's just test code, but it's still better form to manipulate the URL properly. That means parsing it with `new URI(connectUri)` and then creating a replacement `URI` using the 7-arg constructor.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionDockerTest.java
kfaraz: > Setting this in a static initializer doesn't seem right, because it will be set when the class is loaded, not necessarily when the test runs. Maybe better to use @BeforeAll.

Sorry, this was not meant to be committed and was for my local testing only.
Must have crept in in a recent commit.

In CI workflows, this property is always passed as a Java argument `-Ddruid.testing.docker.image`.
So, it needs to be set only when running the test locally.

I chose purposefully not to use a default image so that each test has to set it explicitly.
`KafkaContainer` seems to have deprecated the no-arg constructor too, so that test writers explicitly call out the required image name.

kfaraz: For local testing, what do you think would be the best way to specify this property?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionDockerTest.java
kfaraz: added.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionDockerTest.java
kfaraz: added a validation.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: Fixed.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/DruidContainerResource.java
kfaraz: fixed, thanks for the suggestion!

extensions-core/druid-testcontainers/pom.xml
kfaraz: Thanks for calling this out, I had similar apprehensions.
Removed logger and `StringUtils`.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
kfaraz: fair enough 😂 , I have updated it.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionDockerTest.java
gianm: I think it's fine to require people to set the system property for local testing. Some thought would be needed on part of the person running the tests about what image they want to use. If they want to test against their current branch, they'll need to build a Docker image. Maybe they don't care about that; maybe they just want to test the Docker embedded test itself, in which case it would be fine to run against a released Apache image.

The error message when the property isn't set should spell this out. It should mention that the property is typically set during CI, but isn't set outside CI because we don't know what image you want to use.

gianm: Some thought also needs to be given to how committers should run `mvn test` when they test release builds as part of release voting. Do they need to build a Docker image first in order to run this test? Do they need to specify `-Ddruid.testing.docker.image` to `mvn`? Or is it automatic as part of the `mvn` workflow? I am curious what you think about these items.

kfaraz: Thanks for bringing this up, @gianm !

For release voting, I feel it would be best for committers to download the image built as part of the RC and then running `mvn test -Ddruid.testing.docker.image=apache/druid:rc1 -Dtest=*DockerTest*`.
Since the release voting process is not automated anyway (perhaps by design so that voters are aware of the steps they go through), it is probably okay to require specifying `-Ddruid.testing.docker.image` explicitly.

We can probably add these instructions to release guidelines (and the vote mail) so that users can always copy paste the commands.

What are your thoughts?

kfaraz: > The error message when the property isn't set should spell this out. It should mention that the property is typically set during CI, but isn't set outside CI because we don't know what image you want to use.

Updated the message in `DruidContainerResource`.

gianm: I wonder if the test should be disabled by default, and enabled using a mvn profile. The case for disabling it by default is that it isn't actually testing the current code in the repository. It's testing a image pulled from somewhere else. Having it be its own command makes it more clear what is happening. During voting if committers want to test the docker image too, that can be done with a special `mvn` command separate from the typical test command.

How does this sound to you?

kfaraz: Thanks for the suggestion!
I agree that this would be the cleanest approach.

kfaraz:
@gianm , thanks for the feedback.

I have made the following changes:
- Added a script `run_docker-tests`.
- This script is used by GHA and can also be used to run all docker tests locally
- The script expects an environment variable to be set which gets translated to `-Ddruid.testing.docker.image`. If the env var is not set, tests are not run.
- While voting on release candidates, committers can just invoke this script with the correct image name. If not available, the image would be downloaded by Testcontainers itself.
- Docker tests are now disabled by default and will be triggered only in `mvn verify` given that the required system property `-Ddruid.testing.docker.image` is set.

@kgyrtkirk , I think you had also suggested using `verify` phase and adding a dedicated script.

Example command to run:
```python
DRUID_DIST_IMAGE_NAME=apache/druid:34.0.0-rc1 .github/scripts/run_docker-tests
```
(This commands runs successfully for the ongoing release RC. Only 1 test for custom node role fails as it requires a minor change to the `druid.sh` script that is used to start Druid services in the containers)

Please let me know if these changes make sense.

kfaraz:
@gianm , I also needed to include some other changes to provide visibility into the containers and are very useful in writing these Docker-based tests.

- Converted a `CliCustomNodeRole` to a `CliEventCollector`
- This is a custom test-only Druid service that just collects events over HTTP.
- The events are then forwarded to a `LatchableEmitter` which allows watching for certain events.
- Added a `CustomNodeRoleDockerTest` which supersedes the `ITHighAvailabilityTest`
- Removed code related to the old high availability tests.

I hope these extra changes have not made the PR too bloated.

.github/workflows/ci.yml
kfaraz: Fixed. Please check the latest comments.

.github/workflows/docker-tests.yml
kfaraz: Added script `run_docker-tests`. To run all Docker tests locally,

```
$ DRUID_DIST_IMAGE_NAME=apache/druid:34.0.0-rc1 .github/scripts/run_docker-tests
```

This doesn't include the image building yet though.

server/src/test/java/org/apache/druid/server/metrics/LatchableEmitter.java
```diff
@@ -253,7 +255,7 @@ public EventMatcher hasMetricName(String metricName)
     /**
      * Matches an event only if it has the given metric value.
```
capistrant: nit

```suggestion
     * Matches an event only if it has a value equal to or greater than the given metric value.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/psql/PostgreSQLMetadataResource.java
```diff
@@ -0,0 +1,90 @@
...
+package org.apache.druid.testing.embedded.psql;
+
+import org.apache.druid.java.util.common.StringUtils;
+import org.apache.druid.metadata.storage.postgresql.PostgreSQLMetadataStorageModule;
+import org.apache.druid.testing.embedded.EmbeddedDruidCluster;
+import org.apache.druid.testing.embedded.TestcontainerResource;
+import org.testcontainers.containers.PostgreSQLContainer;
+
+/**
+ * Resource that creates a MariaDB metadata store.
```
capistrant: nit

```suggestion
 * Resource that creates a PostgreSQL metadata store.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/psql/PostgreSQLMetadataStoreTest.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.apache.druid.testing.embedded.psql;
+
+import org.apache.druid.testing.embedded.EmbeddedDruidCluster;
+import org.apache.druid.testing.embedded.EmbeddedRouter;
+import org.apache.druid.testing.embedded.indexing.IndexTaskTest;
+
+/**
+ * Same as {@link IndexTaskTest}, but using a MariaDB metadata store instead of Derby.
```
capistrant: nit

```suggestion
 * Same as {@link IndexTaskTest}, but using a PostgreSQL metadata store instead of Derby.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/CustomNodeRoleDockerTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.druid.testing.DruidCommand;
+import org.apache.druid.testing.cli.CliEventCollector;
+import org.apache.druid.testing.embedded.EmbeddedDruidCluster;
+import org.apache.druid.testing.embedded.EmbeddedRouter;
+import org.apache.druid.testing.embedded.server.HighAvailabilityTest;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+public class CustomNodeRoleDockerTest extends DockerTestBase
```
capistrant: nit: short javadoc relaying what we are testing and why it is valuable test will be helpful for future folks checking in on this

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionBackwardCompatibilityDockerTest.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public class IngestionBackwardCompatibilityDockerTest
+{
+  @Nested
+  public class Apache31 extends IngestionDockerTest
+  {
+    @Override
+    public EmbeddedDruidCluster createCluster()
+    {
+      coordinator.usingImage(DruidContainer.Image.APACHE_31);
```
capistrant: is this something someone is going to have to periodically update and commit to master?

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
```diff
@@ -232,13 +228,21 @@ public TestFolder getTestFolder()
   }
 
   /**
-   * The embedded Zookeeper server used by this cluster, if any.
-   *
-   * @throws NullPointerException if this cluster has no embedded zookeeper.
+   * Uses a container-friendly hostname for all embedded services, Druid as well
+   * as external.
+   */
+  public EmbeddedDruidCluster useContainerFriendlyHostname()
```
capistrant, line 231: what is the reason of having clusters opt into containerFriendlyHostname? Is there a downside or problem of using it if you don't have to, say for an existing embedded test you have written, HighAvailability?

capistrant (commented):
Looking good! few nit comments here and there. My biggest question is the one on container friendly embedded hostnames and why it is not just the default way services are setting up their hostname

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionBackwardCompatibilityDockerTest.java
kfaraz: No, I think we can keep testing against 31 until we decide to make some major change which is not backward compatible with 31 anymore.

I didn't try using an image older than 31, but we could probably go as far back as 27 and still be backward compatible (at least for the things that are being verified in these tests).

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
kfaraz: Fair point!

All the existing embedded tests would still work if we always used the `containerFriendly` hostname. (since the Docker tests already use embedded services and they are able to connect to each other seamlessly)
In fact, that's how it used to be (except we were using `InetAddress.getLocalHost().getCanonicalHostName()` instead of `InetAddress.getLocalHost().getHostAddress()`),
but we changed it in #18228 as there were some apprehensions with using the canonical host name,
ref https://github.com/apache/druid/pull/18228/files#r2197340379.

But now that we are using the `getHostAddress()` which is simply the IP address, I think we can just stick to using the `containerFriendly` one all the time.

cc: @clintropolis

kfaraz:
Thanks for the suggestions, @capistrant ! I have updated the javadocs as suggested and replied to your other comments.
Please let me know what you think.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedDruidCluster.java
capistrant: From an ease of use perspective I think it would be awesome to get to a point where we default to container friendly so people who are writing new tests don't need to have to decide/realize their test requires it. But I don't think it is blocking for this PR

kfaraz: The current default (i.e. localhost) should already be good enough for all (non container) embedded tests.

For Druid container based tests, users writing new tests need not be conscious of the choice as long as they extend `DockerTestBase`.
Also, since Druid containers are not meant to be used frequently except in a couple of smoke tests, the extra step is probably okay?

capistrant: ya, I agree. I'm probably just overthinking it since I'm currently working on the KafkaDataFormat tests which requires an extra docker container that can talk to the kafka container :)

capistrant (approved):
I'm cool with current state of this. Will defer to you @kfaraz if you want to get final word from Gian and/or Zoltan on your changes after their review before you merge

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionBackwardCompatibilityDockerTest.java
gianm: I think this test will get skipped by default since it has `DockerTest` in the name. IMO, it's sensible to always run this particular test. The main reason I thought that Docker tests in general shouldn't all run is that they're mostly intended to run against the "current code". They don't serve their purpose unless the person running the tests actually creates an image matching the current checked out repo. This needs to be done manually so it makes sense that it's a separate command.

But, the backwards compatibility test isn't intended to deploy the "current code". It wants to deploy a specific older version. So, it doesn't have that problem, and it's therefore good to always run it.

Btw, IMO it's nicer to use annotations rather than test names to control what tests run in a standard mvn run. The downside of using names is that it's indirect. It makes it less obvious which tests run and which don't, and also means that if you want to change whether a test runs, you have to rename it.

kfaraz: > Btw, IMO it's nicer to use annotations rather than test names to control what tests run in a standard mvn run. The downside of using names is that it's indirect. It makes it less obvious which tests run and which don't, and also means that if you want to change whether a test runs, you have to rename it.

I agree. In the current code, I am using the `@Tag("docker-tests")` annotation (applied on `DockerTestBase`) to skip the Docker-based tests.
The `*DockerTest*` class name is to tell `mvn verify` to look for these classes while searching for eligible integration tests. Otherwise, it only looks for `*IT` or `IT*` classes by default.

kfaraz: > But, the backwards compatibility test isn't intended to deploy the "current code". It wants to deploy a specific older version. So, it doesn't have that problem, and it's therefore good to always run it.

Thanks for the suggestion!

In that case, I will remove the `BackwardCompatiblityIngestionDockerTest` from the `DockerTestBase` hierarchy. I will also update it to use only Druid 31 containers and embedded servers for it (currently it uses a mix of Druid 31 containers, current code Druid containers and embedded servers).

gianm: > I agree. In the current code, I am using the `@Tag("docker-tests")` annotation (applied on `DockerTestBase`) to skip the Docker-based tests.
> The `*DockerTest*` class name is to tell `mvn verify` to look for these classes while searching for eligible integration tests. Otherwise, it only looks for `*IT` or `IT*` classes by default.

Ah, I see. I missed that when reading the maven stuff.

kfaraz:
Thanks for the reviews, @kgyrtkirk , @capistrant , @gianm !
Thanks for the assist with the CI, @Akshat-Jain !

PR #18305, opened 2025-07-22
Use vector cursors when possible in MSQ

adarshsanjeev:
Add a "shim" cursor implementation, which is an adapter from Cursor to Vector Cursor. This allows vector cursors to be used as normal cursors, by simply reading a vector of values and stepping through it as the cursor advances. 

Shim cursors have been used in MSQ `ScanQueryFrameProcessor` to use vector cursors when possible. 

Also does some minor refactoring around `MSQTestBase`, to make MSQ tests actually use the modules registered by the `getCoreModule()` call in the test framework.

### Main changes
* Added `ShimCursor` to adapt `VectorCursor` to the `Cursor` interface. (`processing/src/main/java/org/apache/druid/segment/shim/ShimCursor.java`)
* Introduced `ShimColumnSelectorFactory` to create appropriate cursors. (`processing/src/main/java/org/apache/druid/segment/shim/ShimColumnSelectorFactory.java`)
* Added specific adapters for vectorized selectors:
  - `ShimMultiValueDimensionSelector` for multi-value dimension vector selectors. (`processing/src/main/java/org/apache/druid/segment/shim/ShimMultiValueDimensionSelector.java`)
  - `ShimNumericColumnValueSelector` for numeric vector selectors. (`processing/src/main/java/org/apache/druid/segment/shim/ShimNumericColumnValueSelector.java`)
  - `ShimObjectColumnValueSelector` for object vector selectors. (`processing/src/main/java/org/apache/druid/segment/shim/ShimObjectColumnValueSelector.java`)

<hr>

This PR has:

- [ ] been self-reviewed.


extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/querykit/scan/ScanQueryFrameProcessor.java
```diff
@@ -287,9 +289,14 @@ protected ReturnOrAwait<Unit> runWithSegment(final SegmentWithDescriptor segment
 
       final Cursor nextCursor;
 
-      // If asCursor() fails, we need to close nextCursorHolder immediately.
+      // If asCursor() or asVectorCursor() fails, we need to close nextCursorHolder immediately.
       try {
-        nextCursor = nextCursorHolder.asCursor();
+        if (nextCursorHolder.canVectorize()) {
```
gianm: Replace this with `query.context().getVectorize().shouldVectorize(nextCursorHolder.canVectorize())`. That way we respect the `vectorize` query parameter (setting `vectorize: false` will disable vectorization).

processing/src/main/java/org/apache/druid/segment/shim/ShimColumnSelectorFactory.java
```diff
@@ -0,0 +1,119 @@
...
+    return dimensionSelectors.computeIfAbsent(
+        dimensionSpec,
+        spec -> {
+          if (spec.mustDecorate()) {
+            throw DruidException.defensive("Only non-decorated dimensions can be vectorized.");
+          }
+          final ColumnCapabilities capabilities = cursor.vectorColumnSelectorFactory
+              .getColumnCapabilities(dimensionSpec.getDimension());
+          if (ColumnProcessors.useDictionaryEncodedSelector(capabilities)) {
+            if (capabilities.hasMultipleValues().isTrue()) {
```
gianm: We need to go down this path if `capabilities.hasMultipleValues().isMaybeTrue()` (if there *may* be multi-values present, we have to be ready to handle them).

processing/src/main/java/org/apache/druid/segment/shim/ShimColumnSelectorFactory.java
```diff
@@ -0,0 +1,119 @@
...
+  @Override
+  public DimensionSelector makeDimensionSelector(DimensionSpec dimensionSpec)
+  {
+    return dimensionSelectors.computeIfAbsent(
+        dimensionSpec,
+        spec -> {
+          if (spec.mustDecorate()) {
+            throw DruidException.defensive("Only non-decorated dimensions can be vectorized.");
+          }
+          final ColumnCapabilities capabilities = cursor.vectorColumnSelectorFactory
```
gianm, line 63: Handle null capabilities too- if the column is missing you can return `DimensionSelector.nilSelector()`.

processing/src/main/java/org/apache/druid/segment/shim/ShimVectorObjectDimSelector.java
```diff
@@ -0,0 +1,148 @@
...
+      objectId = id;
+    }
+  }
+
+  @Override
+  public IndexedInts getRow()
+  {
+    if (hasMultipleValues) {
+      Object object = getObject();
+      ArrayList arrayList = (ArrayList) object;
```
gianm: This is brittle, it assumes that the underlying object is always going to be an `ArrayList`. At least use plain `List`. You should also handle the case here where the underlying object is `null` or `String`. I think this can happen with some underlying selectors.

processing/src/main/java/org/apache/druid/segment/shim/ShimVectorObjectDimSelector.java
```diff
@@ -0,0 +1,148 @@
...
+    return DimensionDictionarySelector.CARDINALITY_UNKNOWN;
+  }
+
+  @Nullable
+  @Override
+  public String lookupName(int id)
+  {
+    Object object = getObject();
+    if (hasMultipleValues) {
+      ArrayList arrayList = (ArrayList) object;
```
gianm: Same as `getRow`, you should handle `null`, `String`, and `List` for the underlying objects.

processing/src/main/java/org/apache/druid/segment/shim/ShimColumnSelectorFactory.java
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+              final SingleValueDimensionVectorSelector vectorSelector =
+                  cursor.vectorColumnSelectorFactory.makeSingleValueDimensionSelector(spec);
+              return new ShimSingleValueDimensionSelector(cursor, vectorSelector);
+            }
+          } else {
+            // Non-dictionary encoded column, like virtual columns.
+            VectorObjectSelector vectorObjectSelector =
+                cursor.vectorColumnSelectorFactory.makeObjectSelector(spec.getDimension());
+            return new ShimVectorObjectDimSelector(cursor, vectorObjectSelector, capabilities.hasMultipleValues().isTrue());
```
gianm: Like above, this should look at `capabilities.hasMultipleValues().isMaybeTrue()` rather than `isTrue()`.

processing/src/main/java/org/apache/druid/segment/shim/ShimColumnSelectorFactory.java
clintropolis: drive by comment: i can't remember if this can actually occur anymore for cursors after #17386, but otoh it doesn't hurt to be safe

processing/src/main/java/org/apache/druid/segment/shim/ShimVectorObjectDimSelector.java
adarshsanjeev: Added branches for these cases. I haven't been able to find a test case that that returns a string here, so I haven't added one yet.

PR #18307, opened 2025-07-22
Handle 1. empty data in CountAggregator; 2. query's required physical column is not in grouping, filtering, ordering, or aggregating

cecemei:
### Description
Fix two bugs in projection:
- Empty data for `CountAggregator`. Prior to this PR, `CountAggregator` throws a null pointer exception since `NullableNumericAggregator` returns null if there's no data. After this PR, `CountAggregator` would return 0 for empty data, since it now asks to use a `forceNonNull` `LongAggregator` for combining.
- Query can ask for physical column that's not in grouping, filtering, ordering, or aggregating, e.x. `select count(*) from unnest(complex_field)`. Prior to this PR, a count aggregator projection would get matched, and exception thrown later since the `complex_field` is not supported by projection cursor. After this PR, count aggregator projection would not be matched.  
  
 * `NullableNumericAggregatorFactory` 
 * `AggregateProjectionMetadata`

<hr>

This PR has:

- [x] been self-reviewed.


processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericAggregatorFactory.java
```diff
@@ -48,31 +50,44 @@
...
     extends AggregatorFactory
 {
+  /**
+   * If this aggregator does not aggregate any values, it will return this value.
+   */
+  @Nullable
+  public Number getDefaultValue()
+  {
+    return null;
+  }
```
clintropolis: since number is not used and really up to the aggregator itself, this should just be `isNullable()` and return a boolean

processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericAggregatorFactory.java
```diff
@@ -48,31 +50,44 @@
...
 
   @Override
-  public final VectorAggregator factorizeVector(VectorColumnSelectorFactory columnSelectorFactory)
+  public VectorAggregator factorizeVector(VectorColumnSelectorFactory columnSelectorFactory)
   {
     Preconditions.checkState(canVectorize(columnSelectorFactory), "Cannot vectorize");
     VectorValueSelector selector = vectorSelector(columnSelectorFactory);
     VectorAggregator aggregator = factorizeVector(columnSelectorFactory, selector);
-    return new NullableNumericVectorAggregator(aggregator, selector);
+    return new NullableNumericVectorAggregator(aggregator, selector, this.getDefaultValue() == null);
```
clintropolis: why can't this just return the `VectorAggregator` like the other factorize methods?

processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericVectorAggregator.java
```diff
@@ -59,17 +59,25 @@ public class NullableNumericVectorAggregator implements VectorAggregator
   @Nullable
   private int[] vAggregationRows = null;
 
-  NullableNumericVectorAggregator(VectorAggregator delegate, VectorValueSelector selector)
+  private final boolean defaultNull;
+
+  NullableNumericVectorAggregator(VectorAggregator delegate, VectorValueSelector selector, boolean defaultNull)
```
clintropolis: these changes are not needed if we just return the base `VectorAggregator`

processing/src/main/java/org/apache/druid/query/aggregation/LongSumAggregatorFactory.java
```diff
@@ -36,12 +36,33 @@ public class LongSumAggregatorFactory extends SimpleLongAggregatorFactory
...
   private final Supplier<byte[]> cacheKey;
 
+  /**
+   * The initValue is the value that will be returned when the aggregator doesn't collect any values.
+   * <p>
+   * For example,
+   * <li> the sum(column) should return null when there's no rows in the segment, thus initValue is null.
+   * <li> the count(column) should return 0 when there's no rows in the segment, thus initValue is 0.
+   */
+  @Nullable
```
clintropolis: i think this should just be `isNullable` and a boolean instead of `initValue` since it isn't used by the aggregator itself

processing/src/main/java/org/apache/druid/query/aggregation/LongSumAggregatorFactory.java
```diff
@@ -36,12 +36,33 @@ public class LongSumAggregatorFactory extends SimpleLongAggregatorFactory
...
+  }
+
   @JsonCreator
   public LongSumAggregatorFactory(
       @JsonProperty("name") String name,
       @JsonProperty("fieldName") final String fieldName,
       @JsonProperty("expression") @Nullable String expression,
-      @JacksonInject ExprMacroTable macroTable
+      @JacksonInject ExprMacroTable macroTable,
+      @JsonProperty("initValue") @Nullable Long initValue
```
clintropolis: i don't think this needs to be on the json at the moment, the count aggregator only calls this internally, and im not sure it is something we want to expose to users yet in the current form, we can always add later if we need it

processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericAggregatorFactory.java
clintropolis: thinking more about this, i guess `isNullable` isn't really the best name, this aggregator factory is for null handling, rather its like some check for if we should bypass checking if the input is nullable or not and so whether or not to use the `NullableNumeric` wrappers...

maybe `isForceNotNullable()` or something better captures the spirit of this flag

processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericAggregatorFactory.java
cecemei: true, updated this

processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericAggregatorFactory.java
```diff
@@ -48,43 +48,65 @@
...
-    return new NullableNumericBufferAggregator(aggregator, nullSelector);
+    if (this.forceNotNullable()) {
+      return aggregator;
+    }
+    return new NullableNumericBufferAggregator(aggregator, makeNullSelector(selector, columnSelectorFactory));
   }
 
   @Override
-  public final VectorAggregator factorizeVector(VectorColumnSelectorFactory columnSelectorFactory)
+  public VectorAggregator factorizeVector(VectorColumnSelectorFactory columnSelectorFactory)
```
clintropolis: why not final?

processing/src/main/java/org/apache/druid/query/aggregation/LongSumAggregatorFactory.java
```diff
@@ -36,12 +36,28 @@ public class LongSumAggregatorFactory extends SimpleLongAggregatorFactory
 {
   private final Supplier<byte[]> cacheKey;
 
-  @JsonCreator
+  /**
+   * If true, the aggregator will return null when there are no values to aggregate. This is the default behavior.
+   * Otherwise, the aggregator must return a default non-null value. This is only used internally and not user-facing.
+   */
+  private final boolean isNullable;
```
clintropolis: i think this should also be `forceNotNullable` like `NullableNumericAggregatorFactory`

processing/src/main/java/org/apache/druid/query/timeseries/TimeseriesQueryEngine.java
```diff
@@ -181,7 +181,8 @@ private Sequence<Result<TimeseriesResultValue>> processVectorized(
               .simple(granularizer.getBucketIterable())
               .map(
                   bucketInterval -> {
-                    // Whether or not the current bucket is empty
+                    // If current bucket is empty and skipEmptyBuckets is true, return null.
+                    // Otherwise, always call aggregators.init, then aggregators.aggregateVector if there's any data.
```
clintropolis: this comment probably seems more useful split up and down where we init the aggs if we haven't already?

processing/src/test/java/org/apache/druid/query/groupby/GroupByQueryRunnerTest.java
```diff
@@ -3399,26 +3399,70 @@ public void testGroupByWithUniquesAndPostAggWithSameName()
   @Test
   public void testGroupByWithCardinality()
   {
-    GroupByQuery query = makeQueryBuilder()
+    // arrange
```
clintropolis: i guess these comments are like to help the reader or something? i haven't really seen this pattern before so i looked it up 😅 . I don't personally think they add much value, but i suppose can leave if you want :shrug:

processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericAggregatorFactory.java
cecemei: updated this

processing/src/main/java/org/apache/druid/query/aggregation/LongSumAggregatorFactory.java
cecemei: updated this

processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericAggregatorFactory.java
cecemei: my mistake, reverted this

processing/src/main/java/org/apache/druid/query/timeseries/TimeseriesQueryEngine.java
cecemei: i'd think this is more like describing what the entire function does, considering that:
- the `init` is being called in two places, so just want to call out early here
- when non-vector, there's no such thing, so thought it might be nice to call out early

processing/src/test/java/org/apache/druid/query/groupby/GroupByQueryRunnerTest.java
cecemei: i feel it's more clear if we separate out the arrange act assert part a bit more, and it really helped for long test functions, maybe less as in shorter ones. calling it out specifically so bringing more awareness to this pattern when writing unit tests, sometimes it's not followed.

processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericAggregatorFactory.java
```diff
@@ -48,22 +48,35 @@
 public abstract class NullableNumericAggregatorFactory<T extends BaseNullableColumnValueSelector>
     extends AggregatorFactory
 {
+  /**
+   * If true, the aggregator will not return null values, even if there's no data to aggregate. In this case, it would
+   * rely on the concrete implementation to return a non-null value.
+   */
```
clintropolis: i think this would more accurately be described as something like "if true, this aggregator will not check for null inputs, instead directly delegating to the underlying aggregator" since it isn't really just specific to just the init value - since we arent wrapping the things with the nullable numeric stuff it means we will call `aggregate` on null rows too, which _should_ result in zeros being read.

To be honest, I'm not completely sure this is generally safe, iirc some column value selectors might explode if we try to get a primitive value from a null row, so it might be worth calling out that this is only known to be safely used to make the count aggregator behave correctly when combining empty results.

processing/src/main/java/org/apache/druid/query/timeseries/TimeseriesQueryEngine.java
clintropolis: >i'd think this is more like describing what the entire function does, considering that:

It feels like its missing some stuff if that is the goal, like how we advance the cursor aggregating values until reaching the end of the cursor, using the granularizer to chunk up the __time values to do the time grouping if applicable. tbh i'm not really sure this block needed a comment at all

>the init is being called in two places, so just want to call out early here

init is only called in the second place to initialize the values if we didn't aggregate anything. The `emptyBucket` variable could also have been called aggsInitialized, or didAggregate or something, since the only reason we call init after the first loop is if we didn't aggregate anything, but emptyBucket makes sense in the context of timeseries

>when non-vector, there's no such thing, so thought it might be nice to call out early

the reason the non-vectorized engine doesn't use have an init is because it uses on heap aggregators, which are constructed initialized... if it used `BufferAggregator` instead it would need to have the same logic to initialize the value in the memory location it is given to store the aggregates like this one does

processing/src/test/java/org/apache/druid/query/groupby/GroupByQueryRunnerTest.java
clintropolis: i suppose we can agree to disagree, i think a line break and clear variable names can achieve the same organizational stuff without being extra text I need to ignore 😅

processing/src/main/java/org/apache/druid/query/aggregation/LongSumAggregatorFactory.java
```diff
@@ -36,25 +36,49 @@ public class LongSumAggregatorFactory extends SimpleLongAggregatorFactory
 {
   private final Supplier<byte[]> cacheKey;
 
+  /**
+   * If false, the aggregator will return null when there are no values to aggregate. This is the default behavior.
+   * Otherwise, the aggregator must return a default non-null value. This is only used internally and not user-facing.
+   */
```
clintropolis: similar to my comment on other javadoc in `NullableNumericAggregator`, this isn't strictly about init, since it also makes it so we call aggregate on null rows (and hopefully get a 0 in all cases)

Also probably worth calling out that is used by count and other callers use at their own risk

processing/src/test/java/org/apache/druid/query/groupby/GroupByQueryRunnerTest.java
kfaraz: My 2 cents: we can either skip the comment altogether as suggested by @clintropolis.
The arrange-act-assert is more preferably captured in the test name itself.

However, for long methods, if we choose to add comments, they should be a little more informative.
Otherwise, they do not add any value and we do just ignore them while reading.

e.g. instead of "assert", it could be something like "Verify that ... whatever you are verifying"
but only something which might not be readily apparent when looking at the code.

kfaraz: Given that the test methods in question are fairly small and straightforward, we can actually skip the comments here.

processing/src/main/java/org/apache/druid/query/aggregation/LongSumAggregatorFactory.java
cecemei: removed this javadoc since it's already covered by `NullableNumericAggregator`

processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericAggregatorFactory.java
cecemei: updated this

processing/src/test/java/org/apache/druid/query/groupby/GroupByQueryRunnerTest.java
cecemei: removed this

PR #18314, opened 2025-07-23
[draft] Start publishing segment load metric from a separate monitor

uds5501:
### Description

Create a `SegmentDiscoveryStatsMonitor` that takes care of segment discovery on broker for data sources. Adds a small POC in embedded tests as well.

<hr>

 * `BrokerServerView`
 * `BrokerSegmentStatsMonitor`

<hr>

This PR has:

- [x] been self-reviewed.


server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsMonitor.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.google.inject.Inject;
+import org.apache.druid.java.util.emitter.service.ServiceEmitter;
+import org.apache.druid.java.util.emitter.service.ServiceMetricEvent;
+import org.apache.druid.java.util.metrics.AbstractMonitor;
+import org.apache.druid.server.coordinator.stats.RowKey;
+
+import java.util.Map;
+
+public class SegmentDiscoveryStatsMonitor extends AbstractMonitor
```
kfaraz: Please add a short javadoc.

kfaraz: A better name would probably be `BrokerSegmentCountStatsMonitor`

server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsMonitor.java
```diff
@@ -0,0 +1,57 @@
...
+  @Inject
+  public SegmentDiscoveryStatsMonitor(SegmentDiscoveryStatsProvider statsProvider)
+  {
+    this.statsProvider = statsProvider;
+  }
+
+  @Override
+  public boolean doMonitor(ServiceEmitter emitter)
+  {
+    emit(emitter, "segment/discover/success", statsProvider.getTotalSuccessfulSegmentLoadCount());
```
kfaraz: From pov of the Broker, servers and segments may come and go. The metric which is useful for tests and otherwise is the total count of segments currently available.

So, we should have a `segment/available/count` metric which gives the current total count rather than a delta count which denotes the number of segments discovered since the last emission window.
I guess that could be useful too but we probably don't need it right now.

server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsProvider.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.druid.timeline.DataSegment;
+
+import java.util.Map;
+
+public interface SegmentDiscoveryStatsProvider
+{
+  /**
+   * Return the number of successful segment loads for each datasource during the emission period.
+   */
+  Map<RowKey, Long> getTotalSuccessfulSegmentLoadCount();
```
kfaraz:
```suggestion
  Map<RowKey, Long> getAvailableSegmentCount();
```

server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsProviderImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.druid.timeline.DataSegment;
+import org.apache.druid.utils.CollectionUtils;
+import org.checkerframework.checker.lock.qual.GuardedBy;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class SegmentDiscoveryStatsProviderImpl implements SegmentDiscoveryStatsProvider
```
kfaraz: Since we just want to emit the current total count for each datasource, I guess you can skip adding this class.

server/src/main/java/org/apache/druid/server/metrics/NoopSegmentDiscoveryStatsProviderImpl.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.apache.druid.server.metrics;
+
+import org.apache.druid.server.coordinator.stats.RowKey;
+import org.apache.druid.timeline.DataSegment;
+
+import java.util.Map;
+
+public class NoopSegmentDiscoveryStatsProviderImpl implements SegmentDiscoveryStatsProvider
```
kfaraz: The Noop impl might not be needed too since the interface really needs just the one method.

server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsProviderImpl.java
```diff
@@ -0,0 +1,85 @@
...
+    );
+    return deltaValues;
+  }
+
+  private static RowKey getMetricKey(final DataSegment segment)
+  {
+    if (segment == null) {
+      return RowKey.empty();
+    }
+    return RowKey.with(Dimension.DATASOURCE, segment.getDataSource()).build();
```
kfaraz: It would be nice to emit counts per datasource-interval-version.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -177,7 +177,8 @@ void setUpEach()
         agg -> agg.hasSumAtLeast(totalRows)
     );
     broker.latchableEmitter().waitForEvent(
-        event -> event.hasDimension(DruidMetrics.DATASOURCE, dataSource)
+        event -> event.hasMetricName("segment/discover/success")
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource)
```
kfaraz, line 183: We should also be checking the value of the metric.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedKafkaClusterMetricsTest.java
```diff
@@ -113,6 +113,7 @@ public void stop()
             .addProperty("druid.manager.segments.killUnused.bufferPeriod", "PT0.1s")
             .addProperty("druid.manager.segments.killUnused.dutyPeriod", "PT1s");
     coordinator.addProperty("druid.manager.segments.useIncrementalCache", "ifSynced");
+    broker.addProperty("druid.monitoring.monitors", "[\"org.apache.druid.server.metrics.SegmentDiscoveryStatsMonitor\"]");
```
kfaraz: If it is not being used in the test, we can skip adding it.

kfaraz (commented):
Thanks for the patch, @uds5501 !
I have left some suggestions.

Edit: Please add docs for the new monitor and metric as well.

server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsMonitor.java
uds5501: fair enough, let me modify it

server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsProviderImpl.java
uds5501: This class serves 2 purpose. 
1. Abstracts away the metrics calculation bloat that I'll have to create in `BrokerServerView` otherwise.
2. Helps me keep jacoco happy by creating new tests and not modifying the old ones a lot.

So, I am inclined to keep that as it is

server/src/main/java/org/apache/druid/server/metrics/NoopSegmentDiscoveryStatsProviderImpl.java
uds5501: Per my reply here - https://github.com/apache/druid/pull/18314#discussion_r2227580619, Iam inclined to keep this class

server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsProviderImpl.java
uds5501: Yeah fair, extending this

server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsMonitor.java
uds5501: added

server/src/main/java/org/apache/druid/server/metrics/SegmentDiscoveryStatsProviderImpl.java
uds5501: Purged this.

server/src/main/java/org/apache/druid/server/metrics/NoopSegmentDiscoveryStatsProviderImpl.java
uds5501: purged

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
uds5501: checking with 10 segments here

server/src/main/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitor.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.inject.Inject;
+import org.apache.druid.java.util.emitter.service.ServiceEmitter;
+import org.apache.druid.java.util.emitter.service.ServiceMetricEvent;
+import org.apache.druid.java.util.metrics.AbstractMonitor;
+import org.apache.druid.server.coordinator.stats.RowKey;
+
+import java.util.Map;
+
+/**
+ * {@Code BrokerSegmentCountStatsMonitor} tracks the currently available segments that broker has discovered.
```
kfaraz:
```suggestion
 * Monitor that tracks the number of segments of a datasource currently queryable by this Broker.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -176,8 +178,10 @@ void setUpEach()
                       .hasDimension(DruidMetrics.DATASOURCE, Collections.singletonList(dataSource)),
         agg -> agg.hasSumAtLeast(totalRows)
     );
-    broker.latchableEmitter().waitForEvent(
-        event -> event.hasDimension(DruidMetrics.DATASOURCE, dataSource)
+    broker.latchableEmitter().waitForEventAggregate(
+        event -> event.hasMetricName("segment/available/count")
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource),
+        agg -> agg.hasCountAtLeast(10)
```
kfaraz: I don't think waiting for the count aggregate is very useful. By invoking `agg.hasCountAtLeast(10)`, we are essentially waiting for the above metric to have been emitted 10 times (with any value or dimensions) since the broker started.

We need to identify all the intervals for which we expect segments to be available, and wait for each of them one by one with the expected count.
Given that this test is using the dataset `wikiticker-2015-09-12-sampled.json.gz` and the Kafka supervisor is set to DAY granularity, I should expect there to be only 1 segment.
But you can double check by putting a debug point in the test and checking the list of segments in the web-console.

So the wait condition might become something like:

```java
broker.latchableEmitter().waitForEvent(
        event -> event.hasMetricName("segment/available/count")
                      .hasDimension(DruidMetrics.INTERVAL, "2015-09-12")
                      .hasValue(1)
);
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedKafkaClusterMetricsTest.java
```diff
@@ -113,6 +113,7 @@ public void stop()
             .addProperty("druid.manager.segments.killUnused.bufferPeriod", "PT0.1s")
             .addProperty("druid.manager.segments.killUnused.dutyPeriod", "PT1s");
     coordinator.addProperty("druid.manager.segments.useIncrementalCache", "ifSynced");
+    broker.addProperty("druid.monitoring.monitors", "[\"org.apache.druid.server.metrics.BrokerSegmentCountStatsMonitor\"]");
```
kfaraz: Is this needed here?

server/src/test/java/org/apache/druid/client/BrokerServerViewTest.java
```diff
@@ -194,6 +212,18 @@ public void testMultipleServerAddedRemovedSegment() throws Exception
             )
         )
     );
+    Map<RowKey, Long> availableSegmentCount = brokerServerView.getAvailableSegmentCount();
+    Map<RowKey, Long> expectedSegmentCount = new HashMap<>();
+    for (DataSegment segment : segments) {
+      expectedSegmentCount.put(
+          RowKey.with(Dimension.DATASOURCE, segment.getDataSource())
+                .with(INTERVAL, segment.getInterval().toString())
```
kfaraz:
```suggestion
                .with(Dimension.INTERVAL, segment.getInterval().toString())
```

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -436,4 +446,42 @@ public List<ImmutableDruidServer> getDruidServers()
...
                   .collect(Collectors.toList());
   }
+
+  @Override
+  public Map<RowKey, Long> getAvailableSegmentCount()
+  {
+    return CollectionUtils.mapValues(segmentAvailableCount, AtomicLong::get);
+  }
+
+  private static long incrementAndGetLong(ConcurrentHashMap<RowKey, AtomicLong> counters, RowKey key)
```
kfaraz: Any specific reason this method is static?

If not, then maybe we can simplify it as follows:
```suggestion
  private void incrementSegmentCount(RowKey key)
```

Same for the decrement method.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -436,4 +446,42 @@ public List<ImmutableDruidServer> getDruidServers()
...
+        counters.remove(key, counter);
+      }
+    }
+    return cnt;
+  }
+
+  private static RowKey getMetricKey(final DataSegment segment)
+  {
+    if (segment == null) {
+      return RowKey.empty();
```
kfaraz: Do we ever need to call this method with a `null` segment?
If not, we should either throw a `DruidException.defensive` or just not have a null check at all.

server/src/main/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsProvider.java
```diff
@@ -0,0 +1,32 @@
...
+package org.apache.druid.server.metrics;
+
+import org.apache.druid.server.coordinator.stats.RowKey;
+
+import java.util.Map;
+
+public interface BrokerSegmentCountStatsProvider
+{
+  /**
+   * Return the number of available segments discovered by broker for a datasource.
```
kfaraz:
```suggestion
   * Return the number of segments queryable by a Broker for a datasource, interval and version.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/EmbeddedHighAvailabilityTest.java
```diff
@@ -119,6 +123,10 @@ public void test_switchLeader_andVerifyUsingSysTables()
     // Run sys queries, switch leaders, repeat
     ServerPair<EmbeddedOverlord> overlordPair = createServerPair(overlord1, overlord2);
     ServerPair<EmbeddedCoordinator> coordinatorPair = createServerPair(coordinator1, coordinator2);
+    broker.latchableEmitter().waitForEvent(
+        event -> event.hasMetricName("segment/available/count")
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource)
+    );
```
kfaraz: Probably cleaner to move this line right after `waitForTaskToSucceed` on line 121.
The total number of segments expected to queryable is 10 across 10 intervals.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -436,4 +446,42 @@ public List<ImmutableDruidServer> getDruidServers()
...
+    return CollectionUtils.mapValues(segmentAvailableCount, AtomicLong::get);
+  }
+
+  private static long incrementAndGetLong(ConcurrentHashMap<RowKey, AtomicLong> counters, RowKey key)
+  {
+    AtomicLong counter = counters.get(key);
+    if (counter == null) {
+      counter = counters.computeIfAbsent(key, k -> new AtomicLong());
+    }
+    return counter.incrementAndGet();
```
kfaraz: Since the map is already a concurrent hash map, we can simplify the code to just be:

```suggestion
    segmentAvailableCount.compute(key, (k, currentValue) -> currentValue == null ? 1 : currentValue + 1);
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedIndexTaskTest.java
```diff
@@ -108,7 +110,8 @@ public void test_runIndexTask_forInlineDatasource()
         agg -> agg.hasSumAtLeast(10)
     );
     broker.latchableEmitter().waitForEvent(
-        event -> event.hasDimension(DruidMetrics.DATASOURCE, dataSource)
+        event -> event.hasMetricName("segment/available/count")
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource)
```
kfaraz: The number of segments expected to be queryable is 10 (across 10 different days).

server/src/test/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitorTest.java
```diff
@@ -0,0 +1,107 @@
...
+                                                          .build();
+  private static final RowKey SEGMENT_METRIC_KEY2 = RowKey.with(Dimension.DATASOURCE, "dataSource2")
+                                                          .with(Dimension.VERSION, "2024-01-02T00:00:00.000Z")
+                                                          .with(Dimension.INTERVAL, "2024-01-02T00:00:00.000Z/2024-01-03T00:00:00.000Z")
+                                                          .build();
+
+  @Before
+  public void setUp()
+  {
+    statsProvider = new BrokerSegmentCountStatsProvider()
```
kfaraz: Can be a lambda.

server/src/test/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitorTest.java
```diff
@@ -0,0 +1,107 @@
...
+    {
+      @Override
+      public Map<RowKey, Long> getAvailableSegmentCount()
+      {
+        return ImmutableMap.of(SEGMENT_METRIC_KEY1, 10L, SEGMENT_METRIC_KEY2, 5L);
```
kfaraz:
```suggestion
        return Map.of(SEGMENT_METRIC_KEY1, 10L, SEGMENT_METRIC_KEY2, 5L);
```

server/src/test/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitorTest.java
```diff
@@ -0,0 +1,107 @@
...
+      }
+    };
+  }
+
+  @Test
+  public void testMonitor()
+  {
+    final BrokerSegmentCountStatsMonitor monitor = new BrokerSegmentCountStatsMonitor(statsProvider);
+    final StubServiceEmitter emitter = new StubServiceEmitter("service", "host");
```
kfaraz:
```suggestion
    final StubServiceEmitter emitter = new StubServiceEmitter();
```

server/src/test/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitorTest.java
```diff
@@ -0,0 +1,107 @@
...
+    final BrokerSegmentCountStatsProvider nullStatsProvider = new BrokerSegmentCountStatsProvider()
+    {
+      @Override
+      public Map<RowKey, Long> getAvailableSegmentCount()
+      {
+        return null;
+      }
+    };
+
+    final BrokerSegmentCountStatsMonitor monitor = new BrokerSegmentCountStatsMonitor(nullStatsProvider);
```
kfaraz:
```suggestion
    final BrokerSegmentCountStatsMonitor monitor = new BrokerSegmentCountStatsMonitor(() -> null);
```

server/src/test/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitorTest.java
```diff
@@ -0,0 +1,107 @@
...
+    final BrokerSegmentCountStatsProvider emptyStatsProvider = new BrokerSegmentCountStatsProvider()
+    {
+      @Override
+      public Map<RowKey, Long> getAvailableSegmentCount()
+      {
+        return ImmutableMap.of();
+      }
+    };
+
+    final BrokerSegmentCountStatsMonitor monitor = new BrokerSegmentCountStatsMonitor(emptyStatsProvider);
```
kfaraz:
```suggestion
    final BrokerSegmentCountStatsMonitor monitor = new BrokerSegmentCountStatsMonitor(() -> Map.of());
```

server/src/test/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitorTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    final BrokerSegmentCountStatsMonitor monitor = new BrokerSegmentCountStatsMonitor(nullStatsProvider);
+    final StubServiceEmitter emitter = new StubServiceEmitter("service", "host");
+    Assert.assertTrue(monitor.doMonitor(emitter));
+
+    Assert.assertEquals(0, emitter.getNumEmittedEvents());
+  }
+
+  @Test
+  public void testMonitorWithEmptyCounts()
```
kfaraz: Nit: method name style for better readability in tests:

```suggestion
  public void test_monitor_withEmptyCounts()
```

server/src/test/java/org/apache/druid/client/BrokerServerViewTest.java
```diff
@@ -64,12 +66,16 @@
...
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
 import java.util.stream.Collectors;
 
+import static org.apache.druid.server.coordinator.stats.Dimension.INTERVAL;
```
kfaraz: Please remove this import.

kfaraz (commented):
Overall looks good. Left some minor suggestions.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
uds5501: that won't work, because `currentValue` needs to be long and `currentValue.incrementAndGet()` will return a `long`

attempted this and it gave error 

```java
segmentAvailableCount.compute(key, (k, currentValue) -> currentValue == null ? new AtomicLong(1) : currentValue.incrementAndGet());
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
uds5501: yeah it was like this, just fixed the interval to `2015-09-12T00:00:00.000Z/2015-09-13T00:00:00.000Z`

server/src/main/java/org/apache/druid/client/BrokerServerView.java
kfaraz: Oh, sorry, I must have clarified. I meant why do we need to keep an `AtomicLong`, we are already using a concurrent hash map. The `segmentAvailableCount` can just be `Map<RowKey, Long>`.

uds5501: fair, used it  now

uds5501:
Note - Something weird is happening with `test_runIndexTask_forInlineDatasource()`. The test when run by itself (until failure) works perfectly, however when ran as part of `EmbeddedMariaDBMetadataStoreTest` it's failing.

kfaraz:
@uds5501 , where does the `EmbeddedMariaDBTest` fail?
Did this start happening after the change or even before?

kfaraz:
@uds5501 , after taking a look at the failure, it seems that the sub-class tests are failing because they override the `createCluster` method. So, they don't have the required `druid.monitoring.monitors` property.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedIndexTaskTest.java
```diff
@@ -61,8 +61,10 @@ public class EmbeddedIndexTaskTest extends EmbeddedClusterTestBase
   @Override
   public EmbeddedDruidCluster createCluster()
   {
+    broker.addProperty("druid.monitoring.monitors", "[\"org.apache.druid.server.metrics.BrokerSegmentCountStatsMonitor\"]");
     return EmbeddedDruidCluster.withEmbeddedDerbyAndZookeeper()
                                .useLatchableEmitter()
+                               .addCommonProperty("druid.monitoring.emissionPeriod", "PT0.1s")
```
kfaraz: Add this property only to the Broker.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedIndexTaskTest.java
```diff
@@ -61,8 +61,10 @@ public class EmbeddedIndexTaskTest extends EmbeddedClusterTestBase
   @Override
   public EmbeddedDruidCluster createCluster()
   {
+    broker.addProperty("druid.monitoring.monitors", "[\"org.apache.druid.server.metrics.BrokerSegmentCountStatsMonitor\"]");
```
kfaraz: Move this property assignment to line 55 (same as being done for the indexer) so that sub-class tests can benefit from it too.

server/src/main/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitor.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.druid.java.util.emitter.service.ServiceMetricEvent;
+import org.apache.druid.java.util.metrics.AbstractMonitor;
+import org.apache.druid.server.coordinator.stats.RowKey;
+
+import java.util.Map;
+
+/**
+ * Monitor that tracks the number of segments of a datasource currently queryable by this Broker.
+ */
+public class BrokerSegmentCountStatsMonitor extends AbstractMonitor
```
kfaraz: Nit: I was wondering if we shouldn't just call it `BrokerSegmentCountMonitor` / `BrokerSegmentStatsMonitor`.
`*CountStats*` feels a little redundant. You can also rename the provider interface accordingly?

What do you think?

kfaraz: Please use `@LoadScope(roles = NodeRole.BROKER_JSON_NAME)` so that this monitor will be loaded only on the Broker after PR #18321 is merged.

server/src/main/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitor.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public boolean doMonitor(ServiceEmitter emitter)
+  {
+    emit(emitter, "segment/available/count", statsProvider.getAvailableSegmentCount());
+    return true;
+  }
+
+  private void emit(ServiceEmitter emitter, String key, Map<RowKey, Long> counts)
```
kfaraz: Should we move this method to the `AbstractMonitor` class? Other monitor classes also seem to have a similar logic.

server/src/main/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitor.java
uds5501: Let's call it `BrokerSegmentStatsMonitor`, might make it easy to extend later farther than just counts.

server/src/main/java/org/apache/druid/server/metrics/BrokerSegmentCountStatsMonitor.java
uds5501: Moving this to `AbstractMonitor` creates a circular dependency with `druid-server`. Though the logic in other monitors is similar, they all have slightly different ways of dealing and building the metric dimensions. Ig we can leave that refactors for later.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -436,4 +443,30 @@ public List<ImmutableDruidServer> getDruidServers()
...
+  {
+    segmentAvailableCount.compute(key, (k, currentValue) -> currentValue == null ? 1 : currentValue + 1);
+  }
+
+  private void decrementSegmentCount(RowKey key)
+  {
+    segmentAvailableCount.computeIfPresent(key, (k, currentValue) -> {
+      long newValue = currentValue - 1;
+      return newValue > 0 ? newValue : null;
+    });
```
kfaraz: Nit: for similarity with the increment method, this can also be:

```suggestion
    segmentAvailableCount.compute(key, (k, currentValue) -> currentValue == null || currentValue <= 1 ? null : currentValue - 1);
```

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -281,6 +286,7 @@ private void serverAddedSegment(final DruidServerMetadata server, final DataSegm
 
           timeline.add(segment.getInterval(), segment.getVersion(), segment.getShardSpec().createChunk(selector));
           selectors.put(segmentId, selector);
+          incrementSegmentCount(getMetricKey(segment));
```
kfaraz: Sorry, I missed this in the original feedback, I think we need to de-duplicate the segment counts as the same segment may be served by multiple processes. So, for every `RowKey` (which already has datasource, interval, version), you can keep a `Set` of partition numbers. The size of this set will give the overall `segment/available/count`.

Also, looking at the embedded tests, I think your original idea of emitting events upon addition/removal might be useful too. You can either add it in this PR or a follow up.

So, you can emit a detailed event `serverview/segment/added`/`serverview/segment/removed` which has `datasource`, `server`, `description` (containing full segment ID). Just track the addition/removal events in a map in `BrokerServerView` and flush the map when `doMonitor` calls `BrokerServerView.getSegmentsUpdated()`.

docs/configuration/index.md
```diff
@@ -1960,6 +1960,7 @@ The following table lists the monitors that are available and the services you c
 |`org.apache.druid.server.metrics.WorkerTaskCountStatsMonitor`|Reports how many ingestion tasks are currently running/pending/waiting, the number of successful/failed tasks, and metrics about task slot usage for the reporting worker, per emission period. |MiddleManager, Indexer|
 |`org.apache.druid.server.metrics.ServiceStatusMonitor`|Reports a heartbeat for the service.|Any|
 |`org.apache.druid.server.metrics.GroupByStatsMonitor`|Report metrics for groupBy queries like disk and merge buffer utilization. |Broker, Historical, Indexer, Peon|
+|`org.apache.druid.server.metrics.BrokerSegmentStatsMonitor`| Report the number of segments of a datasource currently queryable by a Broker.|Broker|
```
kfaraz, line 1963: Please add an entry in `metrics.md` file for the new metric as well.

kfaraz (commented):
LGTM. Only 1 blocking comment to de-duplicate segment counts.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
uds5501: Hmm, dedupe +  two separate kind of events, let me add that as well.

docs/operations/metrics.md
```diff
@@ -93,6 +93,9 @@ Most metric values reset each emission period, as specified in `druid.monitoring
 |`groupBy/spilledBytes`|Number of bytes spilled on the disk by the groupBy queries.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
 |`groupBy/mergeDictionarySize`|Size of on-heap merge dictionary in bytes.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
 
+|`serverview/segment/added`|Number of segments added to the broker|This metric is only available if the `BrokerSegmentStatsMonitor` module is included.|Varies|
```
kfaraz: Please move these next to the `serverview/sync/healthy` metric since they are related to that metric.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/mariadb/EmbeddedMariaDBMetadataStoreTest.java
```diff
@@ -31,6 +32,9 @@ public class EmbeddedMariaDBMetadataStoreTest extends EmbeddedIndexTaskTest
   @Override
   public EmbeddedDruidCluster createCluster()
   {
+    final EmbeddedBroker broker = new EmbeddedBroker()
+        .addProperty("druid.monitoring.monitors", "[\"org.apache.druid.server.metrics.BrokerSegmentStatsMonitor\"]")
+        .addProperty("druid.monitoring.emissionPeriod", "PT0.1s");
```
kfaraz: We probably don't need this change since the `broker` member field in the super class already has the required properties. The `createCluster` method will just use that broker.

docs/operations/metrics.md
```diff
@@ -92,6 +92,8 @@ Most metric values reset each emission period, as specified in `druid.monitoring
 |`groupBy/spilledQueries`|Number of groupBy queries that have spilled onto the disk.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
 |`groupBy/spilledBytes`|Number of bytes spilled on the disk by the groupBy queries.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
 |`groupBy/mergeDictionarySize`|Size of on-heap merge dictionary in bytes.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
+|`serverview/segment/added`|Number of segments added to the broker|This metric is only available if the `BrokerSegmentStatsMonitor` module is included.|Varies|
```
```suggestion
|`serverview/segment/added`|Number of segments discovered and added by the Broker to its server view.|This metric is only available if the `BrokerSegmentStatsMonitor` module is included.|Varies|
```

docs/operations/metrics.md
```diff
@@ -92,6 +92,8 @@ Most metric values reset each emission period, as specified in `druid.monitoring
...
+|`serverview/segment/removed`|Number of segments removed from the broker|This metric is only available if the `BrokerSegmentStatsMonitor` module is included.|Varies|
```
```suggestion
|`serverview/segment/removed`|Number of segments removed by the Broker from its server view.|This metric is only available if the `BrokerSegmentStatsMonitor` module is included.|Varies|
```

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -436,4 +449,55 @@ public List<ImmutableDruidServer> getDruidServers()
...
+        )
+    );
+    return deltaValues;
+  }
+
+  private static RowKey getMetricKey(final DataSegment segment)
+  {
+    return RowKey.with(Dimension.DATASOURCE, segment.getDataSource())
+                 .with(Dimension.INTERVAL, String.valueOf(segment.getInterval()))
+                 .and(Dimension.VERSION, segment.getVersion());
```
kfaraz: Since we are now emitting an added/removed event per segment, rather than interval and version, we can just use the dimensions `dataSource`, `server` and `description` (which contains the full segment ID).

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -76,6 +80,12 @@ public class BrokerServerView implements TimelineServerView
   private final CountDownLatch initialized = new CountDownLatch(1);
   private final FilteredServerInventoryView baseView;
   private final BrokerViewOfCoordinatorConfig brokerViewOfCoordinatorConfig;
+  private final ConcurrentHashMap<RowKey, Long> totalSegmentAddCount = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<RowKey, Long> totalSegmentRemoveCount = new ConcurrentHashMap<>();
+  @GuardedBy("totalSegmentAddCount")
+  private Map<RowKey, Long> previousSegmentAddCount = new HashMap<>();
+  @GuardedBy("totalSegmentRemoveCount")
+  private Map<RowKey, Long> previousSegmentRemoveCount = new HashMap<>();
```
kfaraz: Rather than this, we can just keep a list of events (added, removed).
When `getSegmentsAdded` is called, we just flush out that list.
The count for any given row key will always be 1 if we use segment ID and server in the dimensions.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
uds5501: Are you expecting a full message like `{server_name} service added/removed {segment_id} segment` in the `description`?

server/src/main/java/org/apache/druid/client/BrokerServerView.java
uds5501: For now I have pushed a non event version to verify embedded tests.
I am thinking if a simple `private final ConcurrentHashMap<RowKey, List<DiscoveryEvent>> discoveryEvents` will suffice for now, where `RowKey` has dataSource and server name.

The `DiscoveryEvent` could be a simple enum with `ADDED` / `REMOVED` values. This map will be flushed on each `getSegmentDiscoveryEvents()` fetch.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
kfaraz: No, just the segment id.
`server` should be a separate dimension on the row key

server/src/main/java/org/apache/druid/client/BrokerServerView.java
kfaraz: If you are adding a discovery event class, that can contain the RowKey itself. Then you don't really need the map. You could just have a list of events, which feels more natural.

Drawback is that on the monitor side, you would need to distinguish added and removed events. So you would need to expose this DiscoveryEvent class, which seems unnecessary.

See if you can just make do with a Boolean, or even keeping separate methods and lists for added/removed is fine.

uds5501: Ah, I see, yeah I think I can maintain separate added and removed `RowKeys` for now.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
uds5501: done

server/src/main/java/org/apache/druid/client/BrokerServerView.java
uds5501: done

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
```diff
@@ -177,7 +179,9 @@ void setUpEach()
         agg -> agg.hasSumAtLeast(totalRows)
     );
     broker.latchableEmitter().waitForEvent(
-        event -> event.hasDimension(DruidMetrics.DATASOURCE, dataSource)
+        event -> event.hasMetricName("serverview/segment/added")
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource)
+                      .hasValue(1)
```
kfaraz: Nit: the value of this metric is always 1, so we can remove this condition.

server/src/main/java/org/apache/druid/server/coordinator/stats/Dimension.java
```diff
@@ -31,7 +31,6 @@ public enum Dimension
   DUTY_GROUP("dutyGroup"),
   DESCRIPTION("description"),
   SERVER("server");
-
```
kfaraz: Please revert this change.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -436,4 +445,31 @@ public List<ImmutableDruidServer> getDruidServers()
                   .map(queryableDruidServer -> queryableDruidServer.getServer().toImmutableDruidServer())
                   .collect(Collectors.toList());
   }
+
+  private static RowKey getMetricKey(final DataSegment segment, DruidServerMetadata serverMetadata)
```
kfaraz: Nit: please either keep both args final or both non-final.

kfaraz (approved):
LGTM, left minor suggestions.

kfaraz:
@uds5501 , I think we can close this off for now since the flakiness in the tests has already been addressed.

PR #18317, opened 2025-07-23
Update maven-javadoc-plugin configuration to properly disable javadoc  linting

capistrant:
### Description

Found this issue during druid34 release process. javadoc linting is re-enabled because we updated the plugin version in https://github.com/apache/druid/pull/18138/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L1964 but are still trying to use the old configuration to disable linting. Tested my change locally generating javadocs.

https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html

#### Release note

<hr>

##### Key changed/added classes in this PR
 * `pom.xml`

<hr>

This PR has:

- [X] been self-reviewed.
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [ ] been tested in a test Druid cluster.

PR #18320, opened 2025-07-24
Excercise release build during CI execution

kgyrtkirk:
Adds a workflow to excercise the distribution builds - to ensure that issues like #18317 will be caught before a release is about to happen.

### DIFF
--- a/.github/scripts/validate-dist
+++ b/.github/scripts/validate-dist
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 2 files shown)


.github/scripts/validate-dist
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -x
+
+mvn -B clean install -Papache-release,dist,rat -DskipTests -Ddependency-check.skip -Dgpg.skip
```
cryptoe, line 21: Will it cause the CI to fail in case of CVE's ?

kgyrtkirk: it has `-Ddependency-check.skip` ; so that's not being run...
it would be nice to have an opt-in label to run those on any PR (and on master by default) - but that's a longer story

PR #18321, opened 2025-07-24
Use LoadScope to list services where a monitor should be loaded

kfaraz:
### Description

Defining a load scope safeguards against loading monitors on services where they are not supported.
It also simplifies configuration so that `druid.monitoring.monitors` can always be defined in
`common.runtime.properties` without requiring to override in service-specific `runtime.properties`.

### Changes

- Use `LoadScope` with monitors, same as with `DruidModule`
- Add test in `MetricsModuleTest`
- Use service-specific monitor in `EmbeddedKafkaClusterMetricsTest`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java
+++ b/server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java
@@ -22,7 +22,6 @@ package org.apache.druid.server.metrics;
 import com.google.common.base.Supplier;
 import com.google.common.collect.Iterables;
 import com.google.inject.Binder;
-import com.google.inject.Inject;
 import com.google.inject.Injector;
 import com.google.inject.Key;
 import com.google.inject.Module;
@@ -34,6 +33,7 @@ import org.apache.druid.guice.DruidBinders;
 import org.apache.druid.guice.JsonConfigProvider;
 import org.apache.druid.guice.LazySingleton;
 import org.apache.druid.guice.ManageLifecycle;
+import org.apache.druid.guice.annotations.LoadScope;
 import org.apache.druid.guice.annotations.Self;
 import org.apache.druid.java.util.common.IAE;
 import org.apache.druid.java.util.common.concurrent.Execs;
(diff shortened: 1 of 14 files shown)


server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java
```diff
@@ -218,4 +217,22 @@ public OshiSysMonitor getOshiSysMonitor(
       return new OshiSysMonitor(dimensions, oshiSysConfig);
     }
   }
+
+  /**
+   * Checks if a monitor needs to be loaded on this service based on its node role.
+   */
+  private boolean shouldLoadMonitor(Class<?> monitorClass, Set<NodeRole> nodeRoles)
+  {
+    final LoadScope loadScope = monitorClass.getAnnotation(LoadScope.class);
```
gianm, line 226: Any desire to get fancy here and check superclass annotations too? Probably not necessary but just something that came to mind.

gianm (commented):
The main change LGTM. Can you please also update `docs/configuration/index.md`? The section "Metrics monitors" makes various references to needing service-specific configuration, which shouldn't be necessary anymore. It could be simplified to say that all monitors should be specified in `common.runtime.properties`, and they will only be active on the services that support them.

server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java
kfaraz: Sure, that can be done.

I guess we would need to check the superclass recursively if there is no annotation defined at the current class level. Would that make sense?

kfaraz:
Thanks for the review, @gianm ! I have updated the docs as suggested.

server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java
gianm: Yeah that's what I was imagining.

gianm (approved):

server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java
kfaraz: Thanks! Updated to handle annotations on super-class and added tests.

PR #18328, opened 2025-07-25
Improve projections test coverage

cecemei:
### Description

Improve projections test coverage. Mostly, added a new `ProjectionsTest` which makes the projection-path query on our existing `TestIndex` (`druid.sample.numeric.json`). Along with a few other changes:
- Moved some index related util/ param from `QueryRunnerTestHelper` to `TestIndex`.
- Updated `rtPartialSchemaStringDiscoveryIndex`, somehow this was the same as `rtIndex` before. After updating, `testTimeseriesWithFirstLastAggregator` no longer works for this index, so skip this test.
- Added a new `TestQueryRunnerKit` testing util.

 * `TestQueryRunnerKit`
 * `ProjectionsTest`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/test/java/org/apache/druid/query/timeseries/TimeseriesQueryRunnerTest.java
+++ b/processing/src/test/java/org/apache/druid/query/timeseries/TimeseriesQueryRunnerTest.java
@@ -82,6 +82,7 @@ import org.joda.time.DateTimeZone;
 import org.joda.time.Interval;
 import org.joda.time.Period;
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
@@ -336,6 +337,7 @@ public class TimeseriesQueryRunnerTest extends InitializedNullHandlingTest
   @Test
   public void testTimeseriesProjections()
   {
+    Assume.assumeTrue(QueryRunnerTestHelper.testRunnerHasProjection(runner));
 
     AggregatorFactory maxQuality = new LongMaxAggregatorFactory("maxQuality", "qualityLong");
     Druids.TimeseriesQueryBuilder queryBuilder = Druids.newTimeseriesQueryBuilder()
@@ -367,6 +369,7 @@ public class TimeseriesQueryRunnerTest extends InitializedNullHandlingTest
   @Test
   public void testTimeseriesProjectionsCounts()
   {
+    Assume.assumeTrue(QueryRunnerTestHelper.testRunnerHasProjection(runner));
 
     AggregatorFactory countAgg = new CountAggregatorFactory("count");
     Druids.TimeseriesQueryBuilder queryBuilder = Druids.newTimeseriesQueryBuilder()
@@ -398,6 +401,7 @@ public class TimeseriesQueryRunnerTest extends InitializedNullHandlingTest
   @Test
   public void testTimeseriesNullableLongMax()
   {
+    Assume.assumeTrue(QueryRunnerTestHelper.testRunnerHasProjection(runner));
 
     AggregatorFactory longNullableMax = new LongMaxAggregatorFactory("longNullableMax", "longNumericNull");
     Druids.TimeseriesQueryBuilder queryBuilder = Druids.newTimeseriesQueryBuilder()
@@ -2024,6 +2028,9 @@ public class TimeseriesQueryRunnerTest extends InitializedNullHandlingTest
   @Test
   public void testTimeseriesWithFirstLastAggregator()
   {
+    // some string dimensions are not included, and it changes the results
+    Assume.assumeFalse(runner.toString().equals("rtPartialSchemaStringDiscoveryIndex"));
+
     TimeseriesQuery query = Druids.newTimeseriesQueryBuilder()
                                   .dataSource(QueryRunnerTestHelper.DATA_SOURCE)
                                   .granularity(QueryRunnerTestHelper.MONTH_GRAN)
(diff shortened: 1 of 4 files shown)


processing/src/test/java/org/apache/druid/query/timeseries/TimeseriesQueryRunnerTest.java
```diff
@@ -2024,6 +2028,9 @@ public void testTimeseriesWithMultiValueFilteringJavascriptAggregatorAndAlsoRegu
   @Test
   public void testTimeseriesWithFirstLastAggregator()
   {
+    // some string dimensions are not included, and it changes the results
+    Assume.assumeFalse(runner.toString().equals("rtPartialSchemaStringDiscoveryIndex"));
+
```
clintropolis, line 2033: why is this needed? i think `rtPartialSchemaStringDiscovery` was intended to have `includeAllDimensions` set to true and so should still pick up the dimensions that are not explicitly defined. is this a side-effect of adding some projections to stuff?

processing/src/test/java/org/apache/druid/segment/TestIndex.java
```diff
@@ -221,51 +231,52 @@ public class TestIndex
     ComplexMetrics.registerSerde(HyperUniquesSerde.TYPE_NAME, new HyperUniquesSerde());
   }
 
-  private static Supplier<IncrementalIndex> realtimeIndex = Suppliers.memoize(
+  private static Supplier<IncrementalIndex> rtIndex = Suppliers.memoize(
```
clintropolis, line 234: nit: why the name changes? I guess some are more concise, but more descriptive variable names always feel better to me :shrug:

processing/src/test/java/org/apache/druid/segment/TestIndex.java
```diff
@@ -221,51 +231,52 @@ public class TestIndex
...
-  private static Supplier<IncrementalIndex> noRollupRealtimeIndex = Suppliers.memoize(
+  private static Supplier<IncrementalIndex> noRollupRtIndex = Suppliers.memoize(
       () -> fromJsonResource(SAMPLE_NUMERIC_JSON, false, DIMENSIONS_SPEC)
   );
-  private static Supplier<IncrementalIndex> noBitmapRealtimeIndex = Suppliers.memoize(
+  private static Supplier<IncrementalIndex> noBitmapRtIndex = Suppliers.memoize(
       () -> fromJsonResource(SAMPLE_NUMERIC_JSON, false, DIMENSIONS_SPEC_NO_BITMAPS)
   );
-  private static Supplier<QueryableIndex> mmappedIndex = Suppliers.memoize(
+  private static Supplier<QueryableIndex> mMappedIndex = Suppliers.memoize(
```
clintropolis, line 257: nit: mmap seems more correct than mMap because the sys call is mmap

processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
```diff
@@ -0,0 +1,288 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.mockito.Mockito.times;
```
clintropolis, line 64: in general we've been trying to avoid using Mockito and other mock frameworks unless absolutely necessary... is it really necessary for these tests? a lot of these tests look like they could just be testing the match function directly since they aren't actually scanning a cursor? is this more tedious than i'm imagining?

processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
```diff
@@ -0,0 +1,288 @@
...
+
+  private static List<Arguments> getAllQueryableIndex()
+  {
+    return TestIndex.queryableIndexSupplierMap(true)
+                    .entrySet()
+                    .stream()
+                    .map(entry -> {
+                      // return a mmaped index that can be spied on and its fileMapper can't be closed by the test framework.
+                      QueryableIndex openIndex = Mockito.spy(entry.getValue().get());
+                      Mockito.doNothing().when(openIndex).close();
```
clintropolis, line 80: what is closing stuff? (why is this needed?)

cecemei: It's from junt5, if the parameter implements Closeable, junit5 will automatically call `.close()` on it after the test. More in `ParameterizedInvocationContext.CloseableArgument`. In most of our paramterized test, we're using constructor parameters, the auto closing behavior doesn't apply.

processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
cecemei: The `ArgumentCaptor` can only be used for method on a mocked object. key here is that we don't need to worry about the cursor transformation due to query runner & datasource, as long as we can capture the final cursor that's passed in to the projection match. Also, actually nothing is `mocked` (except the close method), Mockito.spy uses a real object and captures the argument.

processing/src/test/java/org/apache/druid/query/timeseries/TimeseriesQueryRunnerTest.java
cecemei: this is a side-effect of ordering, no related with projection. before this pr, `rtPartialSchemaStringDiscovery` is actually just `rtIndex`, might be a typo. it produces different result because of ordering is now different so doubleLast also changes. it's not explicit but `druid.sample.numeric.json` is actually ordered by its dimensions.

cecemei: i should change my comment here

processing/src/test/java/org/apache/druid/segment/TestIndex.java
cecemei: ah makes sense, although it seems like the second M with upper-case is used more widely, maybe that's something needs to fix

PR #18329, opened 2025-07-25
Add optional realtimeSegmentsOnly in query context

aho135:
### Description

Adds a query context option called realtimeSegmentsOnly that when set to true will only return results from realtime segments. This is useful for our own operations when it is necessary to identify the source of late arriving data.

#### Release note
Adds a query context option called realtimeSegmentsOnly that when set to true will only return results from realtime segments

<hr>

 * `QueryContext`
 * `QueryContexts`
 * `CachingClusteredClient`

<hr>

This PR has:

- [X] been self-reviewed.


aho135:
Thanks for the review @FrankChen021! 

I just fixed merge conflicts

processing/src/main/java/org/apache/druid/query/QueryContexts.java
```diff
@@ -137,6 +137,9 @@ public class QueryContexts
   public static final String NATIVE_QUERY_SQL_PLANNING_MODE_COUPLED = "COUPLED";
   public static final String NATIVE_QUERY_SQL_PLANNING_MODE_DECOUPLED = "DECOUPLED";
 
+  public static final String REALTIME_SEGMENTS_ONLY = "realtimeSegmentsOnly";
```
abhishekrb19, line 140: On the naming, what do you think of something like `queryRealtimeDataOnly`?

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
```diff
@@ -473,6 +473,9 @@ private Set<SegmentServerSelector> computeSegmentsToQuery(
         }
         for (PartitionChunk<ServerSelector> chunk : filteredChunks) {
           ServerSelector server = chunk.getObject();
+          if (query.context().isRealtimeSegmentsOnly() && !server.isRealtimeSegment()) {
```
abhishekrb19: Evaluate and assign `query.context().isRealtimeSegmentsOnly()` once outside the for loops for reuse

abhishekrb19:
Jacaco code coverage check is failing. Looks like most of the sql tests use `TestClusterQuerySegmentWalker` instead of the production class `CachingClusteredClient`. Is it possible to add a test exercising this code path in `CachingClusteredClient` - maybe a quidem test should help?

processing/src/main/java/org/apache/druid/query/QueryContexts.java
aho135: I'm not sure if "query" prefix is needed. Since it's already part of the query context I think that's clear enough

Between realtimeSegmentsOnly and realtimeDataOnly I think both are clear. I'm honestly fine with any of the 3 names though

abhishekrb19: >I'm not sure if "query" prefix is needed. Since it's already part of the query context I think that's clear enough

Ah yes, agreed, "query" does seem redundant.

I was suggesting "data" mostly because realtime data captures both in-memory rows and incrementally persisted segments

aho135: Ahh that's a good point

I thought to use segments because this naming aligns with the `Segments` tab in the console. There is a column named `Is realtime`

aho135:
> Jacaco code coverage check is failing. Looks like most of the sql tests use `TestClusterQuerySegmentWalker` instead of the production class `CachingClusteredClient`. Is it possible to add a test exercising this code path in `CachingClusteredClient` - maybe a quidem test should help?

I was just testing this out, but found that quidem tests are also using TestClusterQuerySegmentWalker. Any other suggestions on how to test this code path? FWIW it does seem like other query context properties are in a similar state

abhishekrb19:
> I was just testing this out, but found that quidem tests are also using TestClusterQuerySegmentWalker. Any other suggestions on how to test this code path? FWIW it does seem like other query context properties are in a similar state

Hmm yeah, I see quidem tests also use the sql test framework's setup in this case. 

Actually I think `CachingClusteredClientTest` would be a good candidate. You could add `ServerType.REALTIME` to the `servers` in the test setup and then run some queries with the desired contexts set to true/false. After the query runs, you can verify that `responseContext.remainingResponsesFromQueryServers` has the correct values populated.

PR #18334, opened 2025-07-28
fix ExpressionVirtualColumn equivalence key to use Expr.stringify instead of raw input string to stabilize comparison

clintropolis:
### Description
Fixes a bug with `ExpressionVirtualColumn` equivalence checking (`VirtualColumn.getEquivalenceKey()`) caused by using the raw input expression string instead of like parsing the expression and using stringify to stabilize it to a homogeneous value. This method is used to match virtual columns in projections with query time virtual columns, so unless the projection was created with the exact same expression string as the query time expression, it would be unable to match correctly (really counter to the purpose of all the equivalence key stuff in the first place, which was added for this purpose 😅)

After the change, `expressionString` is barely needed, but i have retained it so that json serialization and such are not required to parse the expression and restringify just to serialize, and not change behavior of equals/hashcode of the parent `ExpressionVirtualColumn` type (the `EquivalenceKey` is the internal `Expression` type)


processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
```diff
@@ -125,8 +124,11 @@ private ExpressionVirtualColumn(
   )
   {
     this.name = Preconditions.checkNotNull(name, "name");
-    this.expression = new Expression(Preconditions.checkNotNull(expression, "expression"), outputType);
-    this.parsedExpression = parsedExpression;
+    this.expression = new Expression(
```
cecemei, line 127: since `parsedExpression.get()` is already called in line 132, why does `Expression` need a `Supplier<Expr>` instead of just `Expr`? Maybe `Parser.lazyParse` is not needed?

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
```diff
@@ -449,20 +456,21 @@ public boolean equals(Object o)
         return false;
       }
       Expression that = (Expression) o;
-      return Objects.equals(expressionString, that.expressionString) && Objects.equals(outputType, that.outputType);
+      return Objects.equals(parsed.get().stringify(), that.parsed.get().stringify())
```
cecemei, line 459: I spot checked a few `Expr` and they seems to have `equals` implemented, is there any reason to call `stringify` specifically?

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
```diff
@@ -367,13 +369,14 @@ public boolean equals(final Object o)
...
-           Objects.equals(expression, that.expression);
+           Objects.equals(expression.expressionString, that.expression.expressionString) &&
+           Objects.equals(expression.outputType, that.expression.outputType);
   }
 
   @Override
   public int hashCode()
   {
-    return Objects.hash(name, expression);
+    return Objects.hash(name, expression.expressionString, expression.outputType);
```
cecemei, line 379: what's the difference between `expressionString` and `expression.parsed.get().stringify()`?

clintropolis: `expressionString` is the raw expression as input by the user from the json, stringifying the parsed expression homogenizes this so `a+b` and `a + b` and `"a" + "b"` become the same expr after being parsed. (this is the reason the equivalence check was broken, using the raw input strings is not correct because the user could write the same expression in a lot of different ways)

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
clintropolis: its probably ok... for whatever reason I am hesitant to trust equals even tho `Expr` is marked with `SubclassesMustOverrideEqualsAndHashCode` it has a lot less test coverage than stringify which has a lot of coverage to ensure that expr.stringify() makes a string that can be parsed back into the same expr. I guess the solution to my feelings is to just add more coverage using equals and hashcode instead of not using them.

Maybe I'll try to add some coverage and switch to equals.

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
clintropolis: I'm not going to remove `Parser.lazyParse`, this is behind a supplier so we don't spend time parsing expressions unless we actually need to do something with the `Expr` since parsing the expression can add up quite a bit in terms of cpu cost

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
clintropolis: adding some equals/hashcode checks to `FunctionTest` has hit a handful of failures, so my hunch was maybe correct. This is certainly wack that equals/hashcode are not equal, but i think the verdict is that i'll leave this as using stringify until i can fix

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
cecemei: i see, so maybe we should do:
```
this.expressionAnalysis = Suppliers.memoize(() -> parsedExpression.get().analyzeInputs());
```

clintropolis: oops, yea that is a mistake, good catch

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
clintropolis: an update, the failures i've seen so far seem related to typed null constants (like `NullLongExpr` not being equal to `NullDoubleExpr` or a `StringExpr` with a null value), but these all have the same stringified form. Not entirely sure how to best fix it, since they are technically different `Expr`, but they are equivalent... we might need a method like equals but less strict since a given string expression might have multiple possible `Expr` forms depending on stuff like for example whether `asSingleThreaded` is called or not, etc

clintropolis: i suppose more technically we need an actual equivalence method for `Expr` to power all of this stuff, then it could handle more general equivalence like `a + b` and `b + a` being interchangeable, but that's too much for this PR i think

PR #18335, opened 2025-07-29
fix projection json column fields handling

clintropolis:
### Description
changes:
* `AutoTypeColumnMerger` now sets fields list properly on projected json column serializer via new `setFieldsAndOpenWriters` method which sets the fields list (to be shared with parent) and creates and opens field writers for the projected fields
* `NestedDataColumnSerializer` now only writes fields and fieldsinfo during serialization if also writing the dictionaries, so a projected column will not serialize fields and fields info in favor of sharing with the base table column
* added test for projected json column to ensure both with and without projections work correctly

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSerializer.java
+++ b/processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSerializer.java
@@ -271,54 +271,7 @@ public class NestedDataColumnSerializer extends NestedCommonFormatColumnSerializ
       final String fieldFileName = NESTED_FIELD_PREFIX + ctr++;
       fieldsWriter.write(fieldName);
       fieldsInfoWriter.write(field.getValue());
-      final GlobalDictionaryEncodedFieldColumnWriter<?> writer;
-      final ColumnType type = field.getValue().getSingleType();
-      if (type != null) {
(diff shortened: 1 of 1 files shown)


processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSerializer.java
```diff
@@ -486,4 +480,74 @@ public void writeTo(
...
+
+  @Override
+  public void setDictionaryIdLookup(DictionaryIdLookup dictionaryIdLookup)
+  {
+    this.globalDictionaryIdLookup = dictionaryIdLookup;
+    this.writeDictionary = false;
+    this.dictionarySerialized = true;
+  }
+
+  public void setFieldsAndOpenWriters(NestedDataColumnSerializer serializer) throws IOException
```
cecemei, line 445: Seems like this is almost identical to `serializeFields`, maybe can avoid duplicated code?

cecemei (approved):

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSerializer.java
clintropolis: split out method to open the field writers which is the bulk of both loops

PR #18336, opened 2025-07-29
NilStageOutputReader: Fix interruptible static initializer.

gianm:
The "write" method of "Channels.newChannel(baos)" is interruptible, even though "baos" is a local nonblocking stream. This means the static initializer of NilStageOutputReader is also interruptible, which is bad because it might get interrupted if it's running in a processing thread. This leads to the class being unusable with a ExceptionInInitializerError.

The fix is to hard-code the empty frame file. The new static has the same content as the old one.

### DIFF
--- a/extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/shuffle/output/NilStageOutputReader.java
+++ b/extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/shuffle/output/NilStageOutputReader.java
@@ -21,16 +21,11 @@ package org.apache.druid.msq.shuffle.output;
 
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import org.apache.druid.frame.channel.ByteTracker;
 import org.apache.druid.frame.channel.ReadableFrameChannel;
 import org.apache.druid.frame.channel.ReadableNilFrameChannel;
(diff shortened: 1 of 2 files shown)


extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/shuffle/output/NilStageOutputReader.java
```diff
@@ -39,18 +34,11 @@ public class NilStageOutputReader implements StageOutputReader
...
-      EMPTY_FRAME_FILE = baos.toByteArray();
-    }
-    catch (IOException e) {
-      throw new RuntimeException(e);
-    }
-  }
+  /**
+   * Frame file with no frames.
+   */
+  private static final byte[] EMPTY_FRAME_FILE =
```
cryptoe, line 40: if we ever change the format for the frame file this change might get missed. 
Nit: Do you think it makes sense to have this byte array as part of `FrameFile`

gianm: > if we ever change the format for the frame file this change might get missed.

This is a good thing, it will help us realize if we change the frame file format by accident ;P

> Nit: Do you think it makes sense to have this byte array as part of `FrameFile`

I dunno, so far this is the only place it was needed.

PR #18338, opened 2025-07-29
Remove stacktraces for RE failures on 

uds5501:
Removes stack traces for `RE` exception thrown when calling `discover()` on a misconfigured cpu discoverer

### DIFF
--- a/processing/src/main/java/org/apache/druid/java/util/metrics/cgroups/Cpu.java
+++ b/processing/src/main/java/org/apache/druid/java/util/metrics/cgroups/Cpu.java
@@ -77,7 +77,7 @@ public class Cpu
       }
     }
     catch (IOException | RuntimeException ex) {
-      LOG.noStackTrace().error(ex, "Unable to fetch cpu snapshot");
+      LOG.noStackTrace().warn(ex, "Unable to fetch CPU snapshot. Cgroup metrics will not be emitted.");
     }
 
 


processing/src/main/java/org/apache/druid/java/util/metrics/cgroups/Cpu.java
```diff
@@ -77,7 +77,7 @@ public CpuMetrics snapshot()
       }
     }
     catch (IOException | RuntimeException ex) {
-      LOG.error(ex, "Unable to fetch cpu snapshot");
+      LOG.noStackTrace().error(ex, "Unable to fetch cpu snapshot");
```
gianm: This should be `warn`. Error is reserved for serious errors, but this isn't one, it just means that some metrics won't be reported. Btw, the error message should have more detail about the impact. It should mention that this means that `cgroup` metrics will not be emitted. Such as:

```
Unable to fetch CPU snapshot. Cgroup metrics will not be emitted.
```

cryptoe: I think the error message can also mention that try using the `CgroupV2CpuMonitor` ?

uds5501:
Thanks for the review @gianm , I have accommodated the updates.

processing/src/main/java/org/apache/druid/java/util/metrics/cgroups/Cpu.java
gianm: Do you know if it is possible to detect which one is present and switch into the right one automatically? That seems better than trying to write an error message that guides people to make the right config change.

uds5501: I don't think that would work, within `CgroupV2CpuMonitor` is supposed to use `ProcCgroupV2Discoverer` , that itself does not seem to implement the `discover(String cgroup)` completely . The `cgroup` is not being used in the [function](https://github.com/apache/druid/blob/d9f0af84af0d5ac85abd767777aa4c68c8640b23/processing/src/main/java/org/apache/druid/java/util/metrics/cgroups/ProcCgroupV2Discoverer.java#L45)  anywhere so I am not confident that it's complete in itself. 

Additionally  `CgroupV2CpuMonitor` is marked `experimental` so I don't think we should be switching it internally.

cryptoe (approved):

PR #18341, opened 2025-07-30
Making task log storage optional for Kubernetes Runner

uds5501:
### Problem Statement

In current task runner setups, we utilize a common `TaskLog` interface to perform all the operations like log pushes, log streams, payload and report pushes. In future, we want to provide the flexibility to the end user to be able to perform the following things simultaneously:

- Streaming task logs from an external store ( elasticsearch / datadog / deepstorage providers like s3 etc).
- Pushing task logs to separate location that is independent of streaming implementation ( could be stdout / deepstorage [currently supported] etc).
 

### Description

The approach in general is this:

- I am using a `SwitchingTaskLogs` implementation that picks up appropriate options from indexer configuration set in overlord and uses a default type to push and stream reports / statuses etc while uses specific implementations for log push. / log stream if configured.

Property | Default
-- | --
druid.indexer.logs.switching.defaultType | file
druid.indexer.logs.switching.logPushType | use defaultType
druid.indexer.logs.switching.logStreamType | use defaultType
druid.indexer.logs.switching.reportsType | use defaultType

configs as suggested by : @kfaraz ^

- These configs are bound at 3 areas right now:
   -  `CliOverlord.java`
   - `KubernetesOverlordModule.java`
   - `IndexingServiceTaskLogsModule.java`

<hr>

 * `KubernetesPeonLifecycle`
 * `CliOverlord.java`
 * `KubernetesOverlordModule.java`
 * `IndexingServiceTaskLogsModule.java`

#### TODO(s) [notes for author]
- ~Call the binder from all the deep-storage extensions.~
- ~Verify switching setup with K8S Runners.~


indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
```diff
@@ -39,13 +42,22 @@ public class IndexingServiceTaskLogsModule implements Module
   public void configure(Binder binder)
   {
     PolyBind.createChoice(binder, "druid.indexer.logs.type", Key.get(TaskLogs.class), Key.get(FileTaskLogs.class));
+    PolyBind.createChoice(binder, "druid.indexer.logs.delegate.type", Key.get(TaskLogs.class, Names.named("delegate")), Key.get(FileTaskLogs.class));
```
kfaraz: It can be a little confusing to have say these props set on the Overlord:
```properties
druid.indexer.logs.type=external
druid.indexer.logs.delegate.type=file
```

It is unclear why we need a delegate if we just wanted `FileTaskLogs`.

I think the intention is to direct task reports, status and payload to one place and logs to another.
So we need some kind of `switching` or `composite` task logs type, which would give us these props:
```properties
druid.indexer.logs.type=switching
druid.indexer.logs.switching.reportsType=file
druid.indexer.logs.switching.logsType=hdfs
```

With this, we just need a `SwitchingTaskLogs` implementation which takes in a `reportsType` and a `logsType`, both of which would be `TaskLogs` implementations.

kfaraz: To ensure that extensions register themselves correctly as a valid type for say `druid.indexer.logs.switching.logsType`, you would need to add a utility method in `Binders`, say `bindTaskLogs` and invoke it from all the relevant extension modules.

```java
public static <T extends TaskLogs> void bindTaskLogs(Binder binder, String type, Class<T> clazz) {
    // bind to `druid.indexer.logs.type`
    // bind to `druid.indexer.logs.switching.reportsType`
    // bind to `druid.indexer.logs.switching.logsType`
}
```

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/ExternalLogStreamer.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.apache.druid.indexing.common.tasklogs;
+
+import com.google.common.base.Optional;
+import org.apache.druid.tasklogs.TaskLogStreamer;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public class ExternalLogStreamer implements TaskLogStreamer
```
kfaraz: This might not be needed if we go the `switching` route.

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/ExternalTaskLogs.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
+import org.apache.druid.java.util.emitter.EmittingLogger;
+import org.apache.druid.tasklogs.TaskLogs;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class ExternalTaskLogs implements TaskLogs
```
kfaraz: Technically, all implementations of `TaskLogs` are external to Druid anyway.
If we go the `switching` route, this can be renamed to `SwitchingTaskLogs` or equivalent.

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/ExternalTaskLogs.java
```diff
@@ -0,0 +1,104 @@
...
+  @Override
+  public Optional<InputStream> streamTaskStatus(final String taskid) throws IOException
+  {
+    return delegate.streamTaskReports(taskid);
+  }
+
+  @Override
+  public void pushTaskLog(String taskid, File logFile)
+  {
+    log.debug("Skipping task log push for task[%s]", taskid);
```
kfaraz: I think log pushing should be enabled by default.
But there should be a config so that we can turn it off if needed.
The same config could be used for controlling log killing as well.

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
uds5501: > druid.indexer.logs.switching.logsType=hdfs

The options I am trying to provide here is different implementations of `logStorage` vs `logStreaming`, maybe I could just extend this to `druid.indexer.logs.switching.logs.storage` and `druid.indexer.logs.switching.streaming` ?

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/ExternalTaskLogs.java
uds5501: My thought was the base implementation for someone utilizing `external` log push type would not want to use the log pushes (otherwise I think they could use delegate itself?)

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
kfaraz: I see. Yes, the same `switching` concept can be extended for that use case as well.

```properties
druid.indexer.logs.switching.streaming.logsPushType=noop
druid.indexer.logs.switching.streaming.logsStreamType=file
```
where `logsPushType` would bind to a `TaskLogPusher` object.
and `logsStreamType` would bind to a `TaskLogStreamer` object.

Would you still need to distinguish between reports and logs though?
Because I assume reports might need to be pushed to a different place.
If yes, you would still need a

```properties
druid.indexer.logs.switching.streaming.reportsType=noop/file/hdfs
```

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/ExternalTaskLogs.java
kfaraz: Yes, I agree. That is why the `external` + `delegate` combo confused me as well.

I guess once we write out the `switching` implementation, the defaults would become clearer.
But I feel that in any case, log pushing has been a default behaviour of Druid and we should not
disable it by default even with the `switching` impl.

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
uds5501: This sounds good, in my head I am looking to provide the flexibility just for log types for now. The `delegated` / `reportType` could take care of other task responsibilities. So finally, imo it should look like one of these.

### Option 1 [asking for explicit types for each responsibility]

```
druid.indexer.logs.type=switching
druid.indexer.logs.switching.reportsType=file
druid.indexer.logs.switching.payloadManagerType=file
druid.indexer.logs.switching.logsType=switching
druid.indexer.logs.switching.logsType.logsPushType=noop
druid.indexer.logs.switching.logsType.logsStreamType=file
```

translating to

```java
public class SwitchingTaskLogs implements TaskLogs
{
  private final TaskLogs reportDelegate;
  private final TaskLogs payloadDelegate;
  private final TaskLogs logDelegate;

  @Inject
  public SwitchingTaskLogs(
      @Named("report") TaskLogs reportDelegate,
      @Named("payload") TaskLogs payloadDelegate,
      @Named("log") TaskLogs logDelegate
   )
  {

    this.reportDelegate = reportDelegate;
    this.payloadDelegate = payloadDelegate;
    this.logDelegate = logDelegate;
  }
}

public class TaskLogManager implements TaskLogs  // this / some flavour of switching task log will be injected as switching log delegate to handle the task related responsibilities with an appropriate factory placed in front.
{
  private final TaskLogStreamer streamer;
  private final TaskLogPusher pusher; 

  @Inject
  public TaskLogManager(
      @Named("taskLogPusher") TaskLogs pusher,
      @Named("taskLogStreamer") TaskLogs streamer,
   )
  {
    this.pusher = pusher;
    this.streamer = streamer;
  }
}
```

or we could create one like 

### Option 2 [just keeping log handler separate while keeping the rest of the responsibilities with the delegate class]

```
druid.indexer.logs.type=switching
druid.indexer.logs.switching.delegate=file
druid.indexer.logs.switching.logsHandlerType=switching
druid.indexer.logs.switching.logHandler.logsPushType=noop
druid.indexer.logs.switching.logHandler.logsStreamType=file
```

```java
public class SwitchingTaskLogs implements TaskLogs
{
  private final TaskLogs delegate;
  private final TaskLogs switchingLogDelegate; 

  @Inject
  public SwitchingTaskLogs(
      @Named("default") TaskLogs defaultDelegate,
      @Named("log") TaskLogs logDelegate,
   )
  {
    this.defaultDelegate = defaultDelegate;
    this.logDelegate = logDelegate;
  }
}

// the TaskLogManager stays the same.
```

**I am personally leaning towards option 2 given the usecase in hand. what do you folks think?**

<hr> 

EDIT: managed to make it work with a single `SwitchingTaskLogs` with factory. I've added the implementation in this PR for the same.

kfaraz: @uds5501 , I don't think we need two levels of a switching implementation. That would really complicate the configs.

You can provide some decent defaults

When `druid.indexer.logs.type=switching`

|Property|Default|
|---------|-------|
|`druid.indexer.logs.switching.defaultType`|`file`|
|`druid.indexer.logs.switching.logPushType`|use `defaultType`|
|`druid.indexer.logs.switching.logStreamType`|use `defaultType`|
|`druid.indexer.logs.switching.reportsType`|use `defaultType`|

I don't think report and payload will ever need to be different.
If needed, we can add it in the future.

uds5501: Hmm, I see, will use this.

uds5501:
I have verified the `noop` task streaming options in CLI Overlord. 

Used the following configuration in embedded test setup to check the same -

```java
overlord.addProperty("druid.indexer.task.default.context", "{\"useConcurrentLocks\": true}")
            .addProperty("druid.manager.segments.useIncrementalCache", "ifSynced")
            .addProperty("druid.manager.segments.pollDuration", "PT0.1s")
            .addProperty("druid.manager.segments.killUnused.enabled", "true")
            .addProperty("druid.manager.segments.killUnused.bufferPeriod", "PT0.1s")
            .addProperty("druid.manager.segments.killUnused.dutyPeriod", "PT1s")
            .addProperty("druid.indexer.logs.type", "switching")
            .addProperty("druid.indexer.logs.switching.defaultType", "file")
            .addProperty("druid.indexer.logs.switching.streamType", "noop");
```
image

Next test: Using this in conjunction with Kubernetes Task Runners.

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
uds5501: Using this combination noew.

indexing-service/src/test/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogsTest.java
```diff
@@ -0,0 +1,332 @@
...
+
+  @Mock
+  private File testFile;
+
+  private SwitchingTaskLogs taskLogs;
+
+  @Before
+  public void setUp()
+  {
+    MockitoAnnotations.initMocks(this);
```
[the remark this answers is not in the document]
uds5501: Will update the tests to reflect this

kfaraz:
Thanks for the update, @uds5501 ! I will try to take a look at the changes today.

indexing-service/pom.xml
```diff
@@ -267,6 +267,10 @@
             <artifactId>jdbi</artifactId>
             <scope>test</scope>
         </dependency>
+      <dependency>
+        <groupId>org.apache.calcite.avatica</groupId>
+        <artifactId>avatica-core</artifactId>
```
kfaraz: This should not be needed in indexing-service.

uds5501:
Tested locally, the logs streaming is done via s3 as expected, the log pushes + temporary log savings have been shut down (noop). 

Task status - 
```
{
  "id": "index_parallel_kttm1_mbhlngbb_2025-08-04T09:26:54.347Z",
  "groupId": "index_parallel_kttm1_mbhlngbb_2025-08-04T09:26:54.347Z",
  "type": "index_parallel",
  "createdTime": "2025-08-04T09:26:54.359Z",
  "queueInsertionTime": "1970-01-01T00:00:00.000Z",
  "statusCode": "SUCCESS",
  "status": "SUCCESS",
  "runnerStatusCode": "WAITING",
  "duration": 51000,
  "location": {
    "host": "10.244.0.153",
    "port": 8100,
    "tlsPort": -1,
    "k8sPodName": "indexparallelkttm1mbhlngbb2025-d8c190f944e0b8893f24ce3edfdb7q9r"
  },
  "dataSource": "kttm1",
  "errorMsg": null
}
```

report - 

```
{
  "ingestionState": "COMPLETED",
  "unparseableEvents": {},
  "rowStats": {
    "determinePartitions": {
      "processed": 465346,
      "processedBytes": 360464067,
      "processedWithError": 0,
      "thrownAway": 0,
      "unparseable": 0
    },
    "buildSegments": {
      "processed": 465346,
      "processedBytes": 360464067,
      "processedWithError": 0,
      "thrownAway": 0,
      "unparseable": 0
    }
  },
  "errorMsg": null,
  "segmentAvailabilityConfirmed": false,
  "segmentAvailabilityWaitTimeMs": 0,
  "recordsProcessed": {},
  "segmentsPublished": 1
}
```

image


image

processing/src/main/java/org/apache/druid/guice/Binders.java
```diff
@@ -59,4 +60,11 @@ public static MapBinder<String, TaskLogs> taskLogsBinder(Binder binder)
   {
     return PolyBind.optionBinder(binder, Key.get(TaskLogs.class));
   }
+
+  public static <T extends TaskLogs> void bindTaskLogs(Binder binder, String type, Class<T> clazz)
```
kfaraz: Please add a short javadoc to this.

processing/src/main/java/org/apache/druid/guice/Binders.java
```diff
@@ -59,4 +60,11 @@ public static MapBinder<String, TaskLogs> taskLogsBinder(Binder binder)
...
+  {
+    PolyBind.optionBinder(binder, Key.get(TaskLogs.class)).addBinding(type).to(clazz);
+    PolyBind.optionBinder(binder, Key.get(TaskLogs.class, Names.named("defaultType"))).addBinding(type).to(clazz);
```
kfaraz: "defaultType" should be a constant in `TaskLogs` or `SwitchingTaskLogs`.
Use a more qualified name like `switching.defaultType` or something.

processing/src/test/java/org/apache/druid/guice/BindersTest.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.inject.Key;
+import com.google.inject.name.Names;
+import org.apache.druid.tasklogs.NoopTaskLogs;
+import org.apache.druid.tasklogs.TaskLogs;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.Properties;
+
+class BindersTest
```
kfaraz: Thanks for adding this!

processing/src/main/java/org/apache/druid/guice/Binders.java
```diff
@@ -59,4 +60,11 @@ public static MapBinder<String, TaskLogs> taskLogsBinder(Binder binder)
...
+    binder.bind(Key.get(TaskLogs.class, Names.named(type))).to(clazz);
```
kfaraz: Where would we need to refer to a `TaskLogs` object by the name of its implementation?

processing/src/main/java/org/apache/druid/tasklogs/TaskLogPusher.java
```diff
@@ -39,4 +39,9 @@ default void pushTaskReports(String taskid, File reportFile) throws IOException
   default void pushTaskStatus(String taskid, File reportFile) throws IOException
   {
   }
+
+  default boolean logPushEnabled()
```
kfaraz: It seems weird to have a flag to disable the one thing that this interface is meant to do.
Given that we already have a `noop` impl, where do we need this?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordResource.java
```diff
@@ -792,7 +792,7 @@ public Response doGetLog(
         return Response.status(Response.Status.NOT_FOUND)
                        .entity(
                            "No log was found for this task. "
-                           + "The task may not exist, or it may not have begun running yet."
+                           + "The task may not exist, it may not have begun running yet or the taskStream configuration may be faulty, kindly verify the same."
```
kfaraz:
```suggestion
                           + "No logs found for this task. Ensure that the task is running and logging is configured correctly."
```

indexing-service/src/main/java/org/apache/druid/indexing/common/config/FileTaskLogsConfig.java
```diff
@@ -19,11 +19,13 @@
 
 package org.apache.druid.indexing.common.config;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 import javax.validation.constraints.NotNull;
 import java.io.File;
 
+@JsonIgnoreProperties(ignoreUnknown = true)
```
kfaraz: Did you run into any issue while deserializing?
Unknown properties should already be ignored by default.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
```diff
@@ -81,12 +85,16 @@
 public class KubernetesOverlordModule implements DruidModule
 {
 
-  private static final Logger log = new Logger(KubernetesOverlordModule.class);
+  private static final Logger log = new EmittingLogger(KubernetesOverlordModule.class);
```
kfaraz: Nit: Why is this change required?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
```diff
@@ -81,12 +85,16 @@
...
   private static final String K8SANDWORKER_PROPERTIES_PREFIX = IndexingServiceModuleHelper.INDEXER_RUNNER_PROPERTY_PREFIX
                                                                + ".k8sAndWorker";
   private static final String RUNNERSTRATEGY_PROPERTIES_FORMAT_STRING = K8SANDWORKER_PROPERTIES_PREFIX
                                                                         + ".runnerStrategy.%s";
   private static final String HTTPCLIENT_PROPERITES_PREFIX = K8SANDWORKER_PROPERTIES_PREFIX + ".http";
+  private static final String PROPERTY_PREFIX_SWITCHING = "druid.indexer.logs.switching";
+  private static final String PROPERTY_KEY_SWITCHING_PUSH_TYPE = PROPERTY_PREFIX_SWITCHING + ".pushType";
+  private static final String PROPERTY_KEY_SWITCHING_STREAM_TYPE = PROPERTY_PREFIX_SWITCHING + ".streamType";
+  private static final String PROPERTY_KEY_SWITCHING_REPORTS_TYPE = PROPERTY_PREFIX_SWITCHING + ".reportsType";
```
kfaraz: These constants should exist in `SwitchingTaskLogs`.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
```diff
@@ -290,16 +298,83 @@ public RunnerStrategy get()
...
+    Binders.bindTaskLogs(binder, "noop", NoopTaskLogs.class);
+    Binders.bindTaskLogs(binder, "file", FileTaskLogs.class);
+
     binder.bind(NoopTaskLogs.class).in(LazySingleton.class);
     binder.bind(FileTaskLogs.class).in(LazySingleton.class);
+    binder.bind(SwitchingTaskLogs.class).in(LazySingleton.class);
 
     binder.bind(TaskLogPusher.class).to(TaskLogs.class);
     binder.bind(TaskLogKiller.class).to(TaskLogs.class);
+    binder.bind(TaskPayloadManager.class).to(TaskLogs.class);
```
kfaraz: I wonder if these bindings are really needed here. Doesn't the other module `IndexingServiceTaskLogsModule` already perform the same bindings on the Overlord?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
```diff
@@ -290,16 +298,83 @@ public RunnerStrategy get()
...
+  }
+
+  @Provides
+  @Named("streamer")
```
kfaraz: Please move these constants to `SwitchingTaskLogs` and use more consistent names like `switching.logStreamType`.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
```diff
@@ -290,16 +298,83 @@ public RunnerStrategy get()
...
+  public TaskLogs provideStreamer(
```
kfaraz: This logic shouldn't be needed here, `SwitchingTaskLogs` should handle it implicitly.
Same for the other methods.

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,114 @@
...
+
+public class SwitchingTaskLogs implements TaskLogs
+{
+  private final TaskLogs reportTaskLogs;
+  private final TaskLogs logStreamer;
+  private final TaskLogs logPusher;
+
+  @Inject
+  public SwitchingTaskLogs(
+      @Named("reports") TaskLogs reportTaskLogs,
```
kfaraz: There should be an argument for the default type as well.
The other args can be treated as optional bindings.
If they turn out to be not bound (i.e. null), `SwitchingTaskLogs` uses the default one for them.

services/src/main/java/org/apache/druid/cli/CliOverlord.java
```diff
@@ -464,6 +455,38 @@ private void configureOverlordHelpers(Binder binder)
...
+          .toProvider(new ListProvider<TaskLogStreamer>().add(TaskLogs.class))
+          .in(LazySingleton.class);
+
+    binder.bind(TaskLogStreamer.class)
+          .annotatedWith(Names.named("taskstreamer"))
+          .to(TaskRunnerTaskLogStreamer.class)
+          .in(LazySingleton.class);
+  }
+
+  private static void bindSwitchingTaskLogs(Binder binder)
```
kfaraz: The contents of this method can be inlined.

services/src/main/java/org/apache/druid/cli/CliOverlord.java
```diff
@@ -464,6 +455,38 @@ private void configureOverlordHelpers(Binder binder)
...
+  {
+    String logsType = properties.getProperty("druid.indexer.logs.type");
+    if ("switching".equals(logsType)) {
+      bindSwitchingTaskLogs(binder);
+    } else {
+      bindSwitchingTaskLogStreamer(binder);
+    }
+  }
+
+  private static void bindSwitchingTaskLogStreamer(Binder binder)
```
kfaraz: This method can be inlined too since the calling method itself is small enough.

services/src/main/java/org/apache/druid/cli/CliOverlord.java
```diff
@@ -464,6 +455,38 @@ private void configureOverlordHelpers(Binder binder)
     );
   }
 
+  private void bindTaskLogStreamer(Binder binder)
+  {
+    String logsType = properties.getProperty("druid.indexer.logs.type");
+    if ("switching".equals(logsType)) {
+      bindSwitchingTaskLogs(binder);
```
kfaraz: I am not entirely sure if we should be doing this.
For the Overlord, `SwitchingTaskLogs` should still be treated as one of the multiple bound log streamers that the `Overlord` would look at when asked for the logs of a task.

Maybe I am missing some use cases. Could you please explain this change?

kfaraz (commented):
Thanks for the PR, @uds5501 !
I have left some thoughts.

processing/src/main/java/org/apache/druid/tasklogs/TaskLogPusher.java
uds5501: The `TaskLogPusher`  has varied responsibilities internally (reports, logs, statuses etc). This boolean helps in the scenario of K8S runners where it creates a temp file before pushing the logs to deep storage. This method helps skip the entire step

indexing-service/src/main/java/org/apache/druid/indexing/common/config/FileTaskLogsConfig.java
uds5501: Yeah, in unit tests for the modules it seemed to be failing when I configured `switch` types.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
uds5501: These are binding related configs, should they be in the implementation class?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
uds5501: It does however the bindings here ensure you can use the switching implementation independent of the K8S task runnner

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
uds5501: Got it, I'll take care of those handlings internally itself.

indexing-service/src/main/java/org/apache/druid/indexing/common/config/FileTaskLogsConfig.java
kfaraz: Oh, that might be because the `ObjectMapper` you used in the tests was not configured correctly.
Typically, `configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);` does the trick.
Try using the pre-baked `TestHelper.JSON_MAPPER` in the unit test.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
kfaraz: Yeah, they should be in the `SwitchingTaskLogs` class because they are related to that impl only.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
uds5501: Removing these, this will happen directly via `IndexingServiceTaskLogsModule`

processing/src/main/java/org/apache/druid/guice/Binders.java
uds5501: Currently, in the provider implementation. However let me see if i can remove these after creating optional bindings.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
uds5501: my bad, it's not

processing/src/main/java/org/apache/druid/guice/Binders.java
uds5501: Removed.

services/src/main/java/org/apache/druid/cli/CliOverlord.java
uds5501: This has been removed now.

services/src/main/java/org/apache/druid/cli/CliOverlord.java
uds5501: nuked

services/src/main/java/org/apache/druid/cli/CliOverlord.java
uds5501: nuked

indexing-service/src/main/java/org/apache/druid/indexing/common/config/FileTaskLogsConfig.java
uds5501: Done

extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycleTest.java
```diff
@@ -339,7 +339,6 @@ public void test_join_withoutJob_returnsFailedTaskStatus() throws IOException
     EasyMock.expectLastCall().once();
     stateListener.stateChanged(KubernetesPeonLifecycle.State.STOPPED, ID);
     EasyMock.expectLastCall().once();
-
```
kfaraz: The changes to this file seem unintentional. Please revert these.

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
```diff
@@ -20,32 +20,94 @@
...
+
 /**
+ *
  */
 public class IndexingServiceTaskLogsModule implements Module
 {
+  private Properties props;
+
+  @Inject
+  public IndexingServiceTaskLogsModule(Properties props)
```
kfaraz, line 45: This will help you retain the zero arg constructor.

```suggestion
  public setProperties(Properties props)
```

processing/src/main/java/org/apache/druid/guice/Binders.java
```diff
@@ -59,4 +60,16 @@ public static MapBinder<String, TaskLogs> taskLogsBinder(Binder binder)
...
+  /**
+   * Binds implementation of {@link TaskLogs} to the named keys for injection.
+   */
+  public static <T extends TaskLogs> void bindTaskLogs(Binder binder, String type, Class<T> clazz)
+  {
+    PolyBind.optionBinder(binder, Key.get(TaskLogs.class)).addBinding(type).to(clazz);
+    PolyBind.optionBinder(binder, Key.get(TaskLogs.class, Names.named("switching.reportType"))).addBinding(type).to(clazz);
+    PolyBind.optionBinder(binder, Key.get(TaskLogs.class, Names.named("switching.streamType"))).addBinding(type).to(clazz);
+    PolyBind.optionBinder(binder, Key.get(TaskLogs.class, Names.named("switching.pushType"))).addBinding(type).to(clazz);
+    PolyBind.optionBinder(binder, Key.get(TaskLogs.class, Names.named("switching.defaultType"))).addBinding(type).to(clazz);
```
kfaraz: Please use the constants from `SwitchingTaskLogs`.

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class SwitchingTaskLogs implements TaskLogs
+{
+
+  public static final String PROPERTY_PREFIX_SWITCHING = "druid.indexer.logs.switching";
+  public static final String PROPERTY_KEY_SWITCHING_DEFAULT_TYPE = PROPERTY_PREFIX_SWITCHING + ".defaultType";
+  public static final String PROPERTY_KEY_SWITCHING_PUSH_TYPE = PROPERTY_PREFIX_SWITCHING + ".pushType";
```
kfaraz:
```suggestion
  public static final String PROPERTY_KEY_SWITCHING_PUSH_TYPE = PROPERTY_PREFIX_SWITCHING + ".logPushType";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,122 @@
...
+  public static final String PROPERTY_KEY_SWITCHING_STREAM_TYPE = PROPERTY_PREFIX_SWITCHING + ".streamType";
```
kfaraz:
```suggestion
  public static final String PROPERTY_KEY_SWITCHING_STREAM_TYPE = PROPERTY_PREFIX_SWITCHING + ".logStreamType";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,122 @@
...
+  public static final String PROPERTY_KEY_SWITCHING_REPORTS_TYPE = PROPERTY_PREFIX_SWITCHING + ".reportsType";
+
+  public static final String KEY_SWITCHING_REPORTS = "switching.reportType";
+  public static final String KEY_SWITCHING_STREAM = "switching.streamType";
+  public static final String KEY_SWITCHING_PUSH = "switching.pushType";
```
kfaraz: `pushType` seems ambiguous, use `logPushType` instead.

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @Inject
+  public SwitchingTaskLogs(
+      @Nullable @Named(KEY_SWITCHING_DEFAULT) TaskLogs defaultTaskLogs,
+      @Nullable @Named(KEY_SWITCHING_REPORTS) TaskLogs reportTaskLogs,
+      @Nullable @Named(KEY_SWITCHING_STREAM) TaskLogs logStreamer,
+      @Nullable @Named(KEY_SWITCHING_PUSH) TaskLogs logPusher
+  )
+  {
+    this.reportTaskLogs = reportTaskLogs != null ? reportTaskLogs : defaultTaskLogs;
```
kfaraz:
```suggestion
    this.reportTaskLogs = Configs.valueOrDefault(reportTaskLogs, defaultTaskLogs);
```

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
```diff
@@ -20,32 +20,94 @@
...
+      PolyBind.createChoice(
+          binder,
+          SwitchingTaskLogs.PROPERTY_KEY_SWITCHING_STREAM_TYPE,
+          Key.get(TaskLogs.class, Names.named(SwitchingTaskLogs.KEY_SWITCHING_STREAM)),
+          null
+      );
+    } else {
+      binder.bind(Key.get(TaskLogs.class, Names.named(SwitchingTaskLogs.KEY_SWITCHING_STREAM)))
+            .toProvider(() -> null);
+    }
```
kfaraz: Please put this logic in a method so that it can be reused for the 3 cases.

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
```diff
@@ -20,32 +20,94 @@
...
+          SwitchingTaskLogs.PROPERTY_KEY_SWITCHING_REPORTS_TYPE,
+          Key.get(TaskLogs.class, Names.named(SwitchingTaskLogs.KEY_SWITCHING_REPORTS)),
+          null
+      );
+    } else {
+      binder.bind(Key.get(TaskLogs.class, Names.named(SwitchingTaskLogs.KEY_SWITCHING_REPORTS)))
+            .toProvider(() -> null);
+    }
+
+    if (props != null && props.getProperty(SwitchingTaskLogs.PROPERTY_KEY_SWITCHING_STREAM_TYPE) != null) {
```
kfaraz: `props` is never going to be null.

processing/src/main/java/org/apache/druid/guice/Binders.java
```diff
@@ -59,4 +60,16 @@ public static MapBinder<String, TaskLogs> taskLogsBinder(Binder binder)
   {
     return PolyBind.optionBinder(binder, Key.get(TaskLogs.class));
   }
+
+  /**
+   * Binds implementation of {@link TaskLogs} to the named keys for injection.
+   */
+  public static <T extends TaskLogs> void bindTaskLogs(Binder binder, String type, Class<T> clazz)
```
kfaraz:
```suggestion
  public static <T extends TaskLogs> void bindTaskLogs(Binder binder, String type, Class<T> taskLogsType)
```

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,122 @@
...
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
+import org.apache.druid.tasklogs.TaskLogs;
+
+import javax.annotation.Nullable;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class SwitchingTaskLogs implements TaskLogs
```
kfaraz: Please add a javadoc.

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,122 @@
...
+{
+
+  public static final String PROPERTY_PREFIX_SWITCHING = "druid.indexer.logs.switching";
```
kfaraz: Please remove the word `SWITCHING` from all the constants since they already reside in the `SwitchingTaskLogs` class.

```suggestion
  public static final String PROPERTY_PREFIX = "druid.indexer.logs.switching";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,122 @@
...
+  public static final String PROPERTY_KEY_SWITCHING_DEFAULT_TYPE = PROPERTY_PREFIX_SWITCHING + ".defaultType";
```
kfaraz:
```suggestion
  public static final String PROPERTY_DEFAULT_TYPE = PROPERTY_PREFIX_SWITCHING + ".defaultType";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,122 @@
...
+  public static final String PROPERTY_KEY_SWITCHING_PUSH_TYPE = PROPERTY_PREFIX_SWITCHING + ".pushType";
+  public static final String PROPERTY_KEY_SWITCHING_STREAM_TYPE = PROPERTY_PREFIX_SWITCHING + ".streamType";
+  public static final String PROPERTY_KEY_SWITCHING_REPORTS_TYPE = PROPERTY_PREFIX_SWITCHING + ".reportsType";
+
+  public static final String KEY_SWITCHING_REPORTS = "switching.reportType";
```
kfaraz:
```suggestion
  public static final String NAME_REPORTS_TYPE = "switching.reportType";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogs.java
```diff
@@ -0,0 +1,122 @@
...
+  public static final String KEY_SWITCHING_DEFAULT = "switching.defaultType";
+
+  private final TaskLogs reportTaskLogs;
+  private final TaskLogs logStreamer;
+  private final TaskLogs logPusher;
+
+  @Inject
+  public SwitchingTaskLogs(
+      @Nullable @Named(KEY_SWITCHING_DEFAULT) TaskLogs defaultTaskLogs,
+      @Nullable @Named(KEY_SWITCHING_REPORTS) TaskLogs reportTaskLogs,
```
kfaraz: Should we call these `defaultDelegate` and `reportsDelegate` instead?

indexing-service/src/test/java/org/apache/druid/guice/IndexingServiceTaskLogsModuleTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    Injector injector = Guice.createInjector(
+        binder -> {
+          binder.bindScope(LazySingleton.class, Scopes.SINGLETON);
+          binder.bind(Properties.class).toInstance(props);
+          binder.bind(Validator.class).toInstance(new MockValidator());
+        },
+        new IndexingServiceTaskLogsModule(props),
+        new JacksonModule()
+    );
```
kfaraz: This can be put in a method for reuse in all the tests.

indexing-service/src/test/java/org/apache/druid/guice/IndexingServiceTaskLogsModuleTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    TaskLogs pusher = injector.getInstance(Key.get(TaskLogs.class, Names.named("switching.pushType")));
+    Assert.assertTrue(pusher instanceof FileTaskLogs);
+    TaskLogs reports = injector.getInstance(Key.get(TaskLogs.class, Names.named("switching.reportType")));
```
kfaraz: Nit: Please use consistent names, e.g. `logPushType`, `reportsType`, `logStreamType`.

indexing-service/src/test/java/org/apache/druid/guice/IndexingServiceTaskLogsModuleTest.java
```diff
@@ -0,0 +1,142 @@
...
+          binder.bindScope(LazySingleton.class, Scopes.SINGLETON);
+          binder.bind(Properties.class).toInstance(props);
+          binder.bind(Validator.class).toInstance(new MockValidator());
+        },
+        new IndexingServiceTaskLogsModule(props),
+        new JacksonModule()
+    );
+
+    TaskLogs taskLogs = injector.getInstance(TaskLogs.class);
+    Assert.assertTrue(taskLogs instanceof SwitchingTaskLogs);
```
kfaraz, line 58: We should verify that the named types are being defaulted to `file`.

indexing-service/src/test/java/org/apache/druid/guice/IndexingServiceTaskLogsModuleTest.java
```diff
@@ -0,0 +1,142 @@
...
+import javax.validation.metadata.BeanDescriptor;
+import java.util.Collections;
+import java.util.Properties;
+import java.util.Set;
+
+public class IndexingServiceTaskLogsModuleTest
+{
+
+  @Test
+  public void test_switchingTaskLogs_shouldUseDefaultWhenNoPusherConfigured()
```
kfaraz:
```suggestion
  public void test_switchingTaskLogs_usesDefaultType_whenNoPusherConfigured()
```

indexing-service/src/test/java/org/apache/druid/guice/IndexingServiceTaskLogsModuleTest.java
```diff
@@ -0,0 +1,142 @@
...
+        new IndexingServiceTaskLogsModule(props),
+        new JacksonModule()
+    );
+
+    TaskLogs taskLogs = injector.getInstance(TaskLogs.class);
+    Assert.assertTrue(taskLogs instanceof SwitchingTaskLogs);
+  }
+
+  @Test
+  public void test_providePusher_shouldReturnConfiguredImplementaions()
```
kfaraz:
```suggestion
  public void test_switchingTaskLogs_usesConfiguredPusherType()
```

indexing-service/src/test/java/org/apache/druid/guice/IndexingServiceTaskLogsModuleTest.java
```diff
@@ -0,0 +1,142 @@
...
+  {
+    Properties props = new Properties();
+    props.setProperty("druid.indexer.logs.type", "switching");
+    props.setProperty("druid.indexer.logs.switching.defaultType", "file");
+
+    Injector injector = Guice.createInjector(
+        binder -> {
+          binder.bindScope(LazySingleton.class, Scopes.SINGLETON);
+          binder.bind(Properties.class).toInstance(props);
+          binder.bind(Validator.class).toInstance(new MockValidator());
```
kfaraz: Why is this needed?

indexing-service/src/test/java/org/apache/druid/indexing/common/tasklogs/SwitchingTaskLogsTest.java
```diff
@@ -0,0 +1,125 @@
...
+  private SwitchingTaskLogs taskLogs;
+
+  @Before
+  public void setUp()
+  {
+    taskLogs = new SwitchingTaskLogs(defaultTaskLogs, reportTaskLogs, streamerTaskLogs, pusherTaskLogs);
+  }
+
+  @Test
+  public void test_shouldUseIndividualLogs() throws IOException
```
kfaraz: It would be nice to break up this test and perform individual verification for each of the methods.

processing/src/test/java/org/apache/druid/guice/BindersTest.java
```diff
@@ -0,0 +1,58 @@
...
```
kfaraz: Seems cleaner:
```suggestion
public class BindersTest
```

indexing-service/src/test/java/org/apache/druid/guice/IndexingServiceTaskLogsModuleTest.java
uds5501: Not including these makes the test fail with

```
1) [Guice/MissingImplementation]: No implementation for Validator was bound.

Requested by:
1  : JsonConfigurator.<init>(JsonConfigurator.java:84)
      \_ for 2nd parameter
     at JsonConfigProvider.inject(JsonConfigProvider.java:233)
      \_ for 2nd parameter
     at JsonConfigProvider.inject(JsonConfigProvider.java:233)
     at JsonConfigProvider.bind(JsonConfigProvider.java:151)
```

kfaraz: Okay, in that case, include `new ConfigModule()` in the list of modules here.

processing/src/main/java/org/apache/druid/guice/Binders.java
uds5501: Using them here is causing a cyclic dependency.

kfaraz: Oh, I see.

You could just move `SwitchingTaskLogs` to the same package as `TaskLogs`.
There is no reason it cannot be in `processing` module.

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
uds5501: This didn't workout. It seems that `configure()` happens first and then the `@inject` is triggered for Modules.
Keeping it as is for the time being.

processing/src/main/java/org/apache/druid/tasklogs/SwitchingTaskLogs.java
```diff
@@ -17,47 +17,52 @@
...
 {
 
-  public static final String PROPERTY_PREFIX_SWITCHING = "druid.indexer.logs.switching";
-  public static final String PROPERTY_KEY_SWITCHING_DEFAULT_TYPE = PROPERTY_PREFIX_SWITCHING + ".defaultType";
-  public static final String PROPERTY_KEY_SWITCHING_PUSH_TYPE = PROPERTY_PREFIX_SWITCHING + ".pushType";
-  public static final String PROPERTY_KEY_SWITCHING_STREAM_TYPE = PROPERTY_PREFIX_SWITCHING + ".streamType";
-  public static final String PROPERTY_KEY_SWITCHING_REPORTS_TYPE = PROPERTY_PREFIX_SWITCHING + ".reportsType";
+  public static final String PROPERTY_PREFIX = "druid.indexer.logs.switching";
+  public static final String PROPERTY_DEFAULT_TYPE = PROPERTY_PREFIX + ".defaultType";
+  public static final String PROPERTY_KEY_SWITCHING_PUSH_TYPE = PROPERTY_PREFIX + ".logPushType";
```
kfaraz: Similar change for other props:
```suggestion
  public static final String PROPERTY_LOG_PUSH_TYPE = PROPERTY_PREFIX + ".logPushType";
```

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
```diff
@@ -113,4 +79,23 @@ public void configure(Binder binder)
     binder.bind(TaskLogKiller.class).to(TaskLogs.class);
     binder.bind(TaskPayloadManager.class).to(TaskLogs.class);
   }
+
+  private void bindTaskLogImplementation(
+      Binder binder,
+      String propertyKeySwitchingStreamType,
+      String keySwitchingStream
```
kfaraz:
```suggestion
      String typeName
```

indexing-service/src/main/java/org/apache/druid/guice/IndexingServiceTaskLogsModule.java
```diff
@@ -113,4 +79,23 @@ public void configure(Binder binder)
     binder.bind(TaskLogKiller.class).to(TaskLogs.class);
     binder.bind(TaskPayloadManager.class).to(TaskLogs.class);
   }
+
+  private void bindTaskLogImplementation(
+      Binder binder,
+      String propertyKeySwitchingStreamType,
```
kfaraz:
```suggestion
      String propertyKey,
```

processing/src/test/java/org/apache/druid/tasklogs/SwitchingTaskLogsTest.java
```diff
@@ -0,0 +1,211 @@
...
+    EasyMock.expectLastCall();
+    replayAll();
+
+    taskLogs.killAll();
+
+    verifyAll();
+  }
+
+  @Test
+  public void test_killOlderThan_shouldUseReportTaskLogs() throws IOException
```
kfaraz: Nit: `uses` instead of `shouldUse` reads better.
```suggestion
  public void test_killOlderThan_usesReportTaskLogs() throws IOException
```

kfaraz (approved):
Minor nitpicks.

kfaraz:
Thanks for fixing up the docker test job, @uds5501 !

As a follow up to this PR, could you please take up the following:

- Add docs for the new configs added here
- Add a `Release note` section to the description of this PR
- Add embedded tests for task logs

PR #18342, opened 2025-07-30
add support for filtered projections

clintropolis:
### Description
This PR adds a `filter` property to `AggregateProjectionSpec` to allow projections to be pre-filtered to further reduce row counts for views to power queries with commonly used filters. The filter is built into a `ValueMatcher` in `OnHeapAggregateProjection` and rows are checked to match before adding to the facts table. At query time, if a projection has a filter we check to see that the query either contains the exact filter, or is part of an AND filter to ensure that the projection contains all of the rows which would normally match if querying the base table instead. If the query filter matches, the projection filter is removed from the query since it is no longer necessary.

The filter matching is pretty brittle right now and requires an exact match, but this can be improved in follow-up work.

<hr>

This PR has:

- [x] been self-reviewed.


extensions-core/multi-stage-query/src/test/java/org/apache/druid/msq/exec/MSQInsertTest.java
```diff
@@ -553,6 +555,7 @@ public void testInsertOnExternalDataSourceWithCatalogProjections(String contextN
         new AggregateProjectionMetadata.Schema(
```
gianm: add a builder to avoid the need for `null` filter everywhere?

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
```diff
@@ -396,6 +417,45 @@ public Projections.ProjectionMatch matches(
           }
         }
       }
+
+      // if the projection has a filter, the query must contain this filter match
+      if (filter != null) {
```
gianm: IMO, put this logic and `rewriteFilter` in their own dedicated file. The logic will only grow more complex over time.

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
```diff
@@ -554,9 +616,73 @@ public String toString()
...
+            newChildren.add(rewritten);
+            childRewritten = true;
+          } else {
+            newChildren.add(filter);
+          }
+        }
+      }
+      // at least one child must have been rewritten to rewrite the AND
+      if (childRewritten) {
+        if (newChildren.size() > 1) {
```
gianm: can `newChildren` be empty, if all filters under the `AND` of `queryFilter` were part of `projectionFilter`?

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
```diff
@@ -554,9 +616,73 @@ public String toString()
...
+   * Rewrites a query {@link Filter} if possible, removing the {@link Filter} of a projection. To match a projection
+   * filter, the query filter must be equal to the projection filter, or must contain the projection filter as the child
+   * of an AND filter. This method returns null
+   * indicating that a rewrite is impossible with the implication that the query cannot use the projection because the
+   * projection doesn't contain all the rows the query would match if not using the projection.
+   */
+  @Nullable
+  public static Filter rewriteFilter(@Nullable Filter projectionFilter, @Nullable Filter queryFilter)
+  {
+    if (projectionFilter == null || queryFilter == null) {
```
gianm: can `queryFilter` really be null? It seems like this method is only called for nonnull query filters.

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
clintropolis: hmm, this whole method is giant so I started to split all the sub parts out into separate functions to like do the grouping match, aggregator match, and realized there is another place we are doing the query filter match, i think they need to be consolidated to be able to support projections where the filter inputs aren't actually stored in the projection as well as it would be more efficient to look at filter stuff once. I think we would need to shift the filter matching to happen last since it needs the remapped columns the be populated

Will consider splitting into a separate file, though in that case should almost all of the logic be moved out of this file into like `Projections` which already has `findMatchingProjection` which is what calls this matches method? That seems fine to me, just pass in the `AggregateProjectionMetadata.Schema` to most of these methods. Or I guess i could make a utility file per match type, like a `GroupingMatcher` and `FilterMatcher` and `AggregatorMatcher` :shrug:

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
clintropolis: i think not with the current rather weak equals based matching logic - either the parent AND would match a projection AND, or at most a single child of this AND can match. Doing an overlap of AND children would be an improvement, and run into the problem you're asking about I think. I'll see if I can wire that up tho since it seems important for a query with something like `x = 'a' and y = 'b' and z = 'c'` to match a projection with `x = 'a' and y = 'b'`

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
clintropolis: oops, not anymore

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
clintropolis: ok, locally i've moved most of the logic into `Projections` (and moved the `ConstantTimeColumn` stuff in `Projections` to its own file), and it seems nicer so far, more to do tho so havent pushed yet

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
```diff
@@ -219,7 +233,7 @@ public Schema(
       }
       this.timeColumnName = timeColumnName;
       this.timeColumnPosition = foundTimePosition;
-      this.granularity = granularity == null ? Granularities.ALL : granularity;
+      this.effectiveGranularity = granularity == null ? Granularities.ALL : granularity;
```
cecemei, line 235: i think we're expecting the same size/ ordering bwteen groupingColumns and OrderBy between line 223 and 224, this assumption is guaranteed from `ProjectionsSpec` but not obvious here, maybe worth adding some checks.

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
```diff
@@ -146,6 +163,7 @@ public boolean equals(Object o)
     }
     AggregateProjectionSpec that = (AggregateProjectionSpec) o;
     return Objects.equals(name, that.name)
+           && Objects.equals(filter, that.filter)
```
cecemei, line 185: Is it intentional `timeColumnName` is not included in equals/ hashCode/ toString method, but `ordering` is? Both are derived so seems like not needed.

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
cecemei: also it seems like this `effectiveGranularity` and `orderingWithTimeSubstitution` logic is very similar to the `computeOrdering` function, maybe it could be merged?

processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
```diff
@@ -0,0 +1,297 @@
...
+import org.apache.druid.segment.filter.IsBooleanFilter;
+import org.apache.druid.segment.filter.OrFilter;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+class ProjectionsTest
```
gianm, line 43: `public class`? not sure it matters, just odd not to see the keyword.

processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
```diff
@@ -0,0 +1,297 @@
...
+                       .setPhysicalColumns(Set.of("c_sum"))
+                       .setPreferredOrdering(List.of())
+                       .build(),
+        Map.of("c", "c_sum")
+    );
+    Assertions.assertEquals(expected, projectionMatch);
+  }
+
+  @Test
+  void testSchemaMatchFilter()
```
gianm, line 88: We're going to need easier, more concise ways to write tests for filter matching. As the filter matching gets more sophisticated, there will be a lot of variations to test to ensure everything is working properly.

Maybe we can define a few "standard" datasources for the tests, each one with a base table and some projection(s). Then each test would just need to define a `CursorBuildSpec` and the expected `ProjectionMatch`.

processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
```diff
@@ -0,0 +1,297 @@
...
+                                                         )
+                                                     )
+                                                     .build();
+
+    Projections.ProjectionMatch projectionMatch = Projections.matchAggregateProjection(
+        spec.getSchema(),
+        cursorBuildSpec,
+        new RowSignatureChecker(baseTable)
+    );
+    Projections.ProjectionMatch expected = new Projections.ProjectionMatch(
```
gianm: nit- moving `ProjectionMatch` out into its own class would help a little with making the test code more concise.

gianm (approved):
Approved, with some thoughts on the tests.

processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
clintropolis: ah yea, i was writing these tests as mainly synthetic to smoke test stuff as they are not backed by real data. I think for the bulk of tests I would agree we want something like what `CursorFactoryProjectionTest` is doing with a matrix of incremental and immutable segments but with a more cursor + filter focus and make it test that results are the same with and without the projection similar to what the native filter tests are doing

processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
clintropolis: used intellij to generate a test file and saw it without the public, i looked into it and apparently in junit5 it is not necessary so have been rolling with it :shrug: https://docs.junit.org/current/user-guide/#writing-tests-classes-and-methods

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
clintropolis: this is sort of aspirational - I think it would be possible someday for ordering to not be fixed to grouping if we ever open up the comparator used to sort data when building segments. However, no such plans for the time column.

(however the other part you noticed about relation between grouping and ordering maybe sort of breaks this)

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
clintropolis: >i think we're expecting the same size/ ordering bwteen groupingColumns and OrderBy between line 223 and 224, this assumption is guaranteed from ProjectionsSpec but not obvious here, maybe worth adding some checks.

yea, good spot, it is true they unfortunately are coupled right now, though in the future I think i would like to make them more independent rather than consolidate to keep us free to potentially allow user choice in ordering in the future. Given that this is not new in this PR I would rather deal with this in a follow-up

PR #18345, opened 2025-07-31
Web console: build only if needed

vogievetsky:
Make the `script/build` script (the one `mvn` calls) not rebuild if the output is newer than the input. Should save a lot of time on running tests.

### DIFF
--- a/web-console/script/build
+++ b/web-console/script/build
@@ -18,29 +18,98 @@
 
 set -e
 
-# If needed, build SQL docs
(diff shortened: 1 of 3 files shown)


web-console/script/build
```diff
@@ -18,11 +18,98 @@
...
+    if is_newer_than "$config_file" "$CONSOLE_OUTPUT"; then
+      BUILD_CONSOLE=true
+      break
+    fi
+  done
+
+  # Check if any source file is newer than output
+  if [ "$BUILD_CONSOLE" = false ]; then
+    # Find the newest source file
+    NEWEST_SOURCE=$(find src lib -name "*.ts" -o -name "*.tsx" | xargs ls -t 2>/dev/null | head -1)
```
cryptoe, line 102: Neat trick.

PR #18347, opened 2025-07-31
modify NestedFieldVirtualColumn to always use NestedDataComplexColumn instead of CompressedNestedDataComplexColumn

clintropolis:
### Description
This was a mistake, there is no reason `NestedFieldVirtualColumn` needed to be using `CompressedNestedDataComplexColumn` instead of the base `NestedDataComplexColumn` type


clintropolis:
ci failure seems unrelated

PR #18348, opened 2025-08-01
set version of dorny/paths-filter to 3.0.2

clintropolis:
### Description
unit tests seems broken with an error like

```
dorny/paths-filter@v3.0.0 is not allowed to be used in apache/druid. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub, verified in the GitHub Marketplace, or matching the following:
...
```

trying to fix by setting the specific version listed in https://github.com/apache/infrastructure-actions/blob/main/actions.yml#L344

### DIFF
--- a/.github/workflows/revised-its.yml
+++ b/.github/workflows/revised-its.yml
@@ -51,7 +51,7 @@ jobs:
       # the common extension in revised ITs is different from the one in standard ITs
       common-extensions: ${{ steps.filter.outputs.common-extensions }}
     steps:
-      - uses: dorny/paths-filter@v3.0.2
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36
         if: github.event_name == 'pull_request'
         id: filter
         with:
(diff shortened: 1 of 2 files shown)


clintropolis:
hmm, still isn't working
```
dorny/paths-filter@v3.0.2 is not allowed to be used in apache/druid. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub, verified in the GitHub Marketplace, or matching the following: 1Password/load-secrets-action@581a835fb51b8e7ec56b71cf2ffddd7e68bb25e0, AdoptOpenJDK/install-jdk@*, BobAnkh/auto-generate-changelog@*, DavidAnson/markdownlint-cli2-action@992badcdf24e3b8eb7e87ff9287fe931bcb00c6e, DavidAnson/markdownlint-cli2-action@b4c9feab76d8025d1e83
```
we might have to ask infra

clintropolis:
hmm, changing this to ```dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36``` appears to work,I kind of wonder if this was an accident on infra side, because according to https://github.com/apache/infrastructure-actions/blob/main/actions.yml#L341 this should be v3.0.2, however looking at the commit that likely broke our CI it removed 3.0.2 from another approved patterns file: https://github.com/apache/infrastructure-actions/commit/f65b1b9f5c4a50a7559fed2773a00f6361e1fe56#diff-f6b4b5e5fbecf48ae70128dc3cfcc32fdc4e070135904c9607e4e357aabd5c54R93

PR #18349, opened 2025-08-01
Add utility method to EmbeddedClusterApis to run tasks

kfaraz:
### Description

Embedded tests currently run into serialization when submitting a `Task` or `CompactionSupervisorSpec` payload to the cluster using any `OverlordClient`.
This can happen because the `ObjectMapper` instance used by the `OverlordClient` might not have the necessary
Jackson modules loaded.
The hack so far has been to ensure that the first server in an embedded cluster is always an Overlord.
But this can quickly run into issues, e.g. when running a cluster with basic auth enabled where a
Coordinator must be started before all other servers.

### Fix

- Add utility methods in `EmbeddedClusterApis` to `submitTask`, `runTask`, `postSupervisor`
- Accept an `EmbeddedOverlord` in these methods so that the appropriate `OverlordClient` can be used
- The request is still sent over the wire since an Overlord also uses a regular HTTP-based `OverlordClient`,
even if it might end up calling itself.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionTestBase.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionTestBase.java
@@ -70,10 +70,7 @@ public abstract class CompactionTestBase extends EmbeddedClusterTestBase
   protected String runTask(TaskBuilder<?, ?, ?> taskBuilder)
   {
     final String taskId = IdUtils.getRandomId();
-    cluster.callApi().runTask(
-        taskBuilder.dataSource(dataSource).withId(IdUtils.getRandomId()),
-        overlord
-    );
+    cluster.callApi().runTask(taskBuilder.dataSource(dataSource).withId(taskId), overlord);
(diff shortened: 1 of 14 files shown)


gianm (commented):
What is here looks good, but please also update `EmbeddedDruidCluster` methods `leaderOverlord`, `leaderCoordinator`, and `anyBroker` to return properly built clients. As-is, they are foot-guns waiting to happen. Perhaps an approach could be to iterate `servers` to find the right server type and pull the client from that. Or, it could be to pull the `jsonMapper` from the right server type and build a client that is associated with the test framework itself (rather than from a server injector).

kfaraz:
Thanks for the feedback, @gianm !
I have fixed up the methods `leaderOverlord`, `leaderCoordinator` and `anyBroker`.
For now, I am simply using the injected clients from the relevant servers.
In the future, if needed, we can construct a new client using the underlying object mappers.

gianm (approved):

PR #18350, opened 2025-08-01
Add guard rails for Dart around select destination

adarshsanjeev:
Druid currently does not support values of the `selectDestination` other than `taskReport` for Dart. This can cause queries to start failing with cryptic exceptions.

This PR adds a validation when choosing the select destination, to use task report for Dart queries.

I considered having the context be responsible for deciding and returning the destination instead, but kept it in the query context since it:
1. Adds more calls to the context
2. This `getSelectDestination()` method could be called from other places in the future, which causes it to skip this check.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/util/MultiStageQueryContext.java
+++ b/extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/util/MultiStageQueryContext.java
@@ -384,15 +384,23 @@ public class MultiStageQueryContext
 
   public static MSQSelectDestination getSelectDestination(final QueryContext queryContext)
   {
-    if (isDartQuery(queryContext)) {
-      // Dart does not support any other destination.
-      return MSQSelectDestination.TASKREPORT;
-    }
-    return QueryContexts.getAsEnum(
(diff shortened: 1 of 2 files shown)


extensions-core/multi-stage-query/src/main/java/org/apache/druid/msq/util/MultiStageQueryContext.java
```diff
@@ -377,9 +377,17 @@ public static boolean removeNullBytes(final QueryContext queryContext)
...
+  public static boolean isDartQuery(final QueryContext queryContext)
+  {
+    return queryContext.get(QueryContexts.CTX_DART_QUERY_ID) != null;
+  }
 
   public static MSQSelectDestination getSelectDestination(final QueryContext queryContext)
   {
+    if (isDartQuery(queryContext)) {
+      // Dart does not support any other destination.
+      return MSQSelectDestination.TASKREPORT;
```
cryptoe: Lets add a log.warn here that we donot support any other destination than task report in DART.

cryptoe (approved):

PR #18356, opened 2025-08-04
Put errors on warning level for cpu set snapshots

uds5501:
Sending appropriate warnings when cpuset metrics are not available

### DIFF
--- a/processing/src/main/java/org/apache/druid/java/util/metrics/cgroups/CpuSet.java
+++ b/processing/src/main/java/org/apache/druid/java/util/metrics/cgroups/CpuSet.java
@@ -72,7 +72,7 @@ public class CpuSet
       output = lines.stream().map(this::parseStringRangeToIntArray).findFirst().orElse(output);
     }
     catch (RuntimeException | IOException ex) {
-      LOG.error(ex, "Unable to read %s", file);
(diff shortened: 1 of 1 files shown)


processing/src/main/java/org/apache/druid/java/util/metrics/cgroups/CpuSet.java
```diff
@@ -72,7 +72,7 @@ private int[] readCpuSetFile(String file)
       output = lines.stream().map(this::parseStringRangeToIntArray).findFirst().orElse(output);
     }
     catch (RuntimeException | IOException ex) {
-      LOG.error(ex, "Unable to read %s", file);
+      LOG.noStackTrace().warn(ex, "Unable to read %s, these metrics will be skipped", file);
```
cryptoe, line 75: Nit: IMHO, it might be better to add all the `CPUS_FILE`, `EFFECTIVE_CPUS_FILE` etc in an array and then replace `these` with the exact metric name. That would help admins to figure out what happens with `CPUS_FILE` since the cluster admin would most likely grep for the particular string.

uds5501: @cryptoe , The file name will be mentioned here in the log already, just in separate log lines.

cryptoe (approved):

PR #18380, opened 2025-08-07
fix bug causing wrong dimension merger order leading to exceptions or wrong data being used for a projection

clintropolis:
### Description
Fixes accidental use of `HashMap` instead of `LinkedHashMap` causing projection column merger order to be unstable during segment merging. The added test would fail prior to this change (well really the projection added to the base set of projections would cause the test file to fail to initialize all the data since it could no longer persist the segment)

### DIFF
--- a/processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
+++ b/processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
@@ -97,6 +97,7 @@ import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -1777,6 +1785,58 @@ public void testProjectionFilter()
     Assert.assertArrayEquals(new Object[]{"b", 7L, 7.7}, results.get(1).getArray());
   }
 
+  @Test
+  public void testProjectionSelectionTwoVirtual()
+  {
+    // this query can use the projection with 2 dims, which has 7 rows instead of the total of 8
```
cecemei: if projection has 7 rows, why does line 1825 expects 8?

clintropolis: oops, stale comment, i copied another test to write this one 🍝

clintropolis: actually this comment is stale in other places too, will fix those as well

cecemei (approved):

PR #18383, opened 2025-08-08
Kafka dataformat test embedded migration

capistrant:
### Description

Migrate the `druid-integration-tests` `kafka-data-formats` `TestNG` group to Embedded Tests. In order to take advantage of existing utilities to generate serialized event streams for testing, I moved a bunch of test utilities from integration-tests to integration-tests-ex/tools where they can be shared by legacy ITs and EmbeddedTests before eventually being brought fully into EmbeddedTests once legacy ITs are all gone.

Two of the data formats, Avro and Protobuf, required testing with SchemaRegistry. Therefore, I brought in a new resource, ConfluentSchemaRegistry to be used as a testcontainer for these tests.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.


capistrant:
```
[INFO] Running org.apache.druid.testing.embedded.indexing.KafkaDataFormatsTest
[INFO] Tests run: 14, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 234.3 s -- in org.apache.druid.testing.embedded.indexing.KafkaDataFormatsTest
```

well they ran. Kind of a lot of runtime for 14 tests (but I guess way faster than the IT)

capistrant:
I musta broke something in the existing ITs. will look into in monday

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/SchemaRegistryResource.java
```diff
@@ -0,0 +1,48 @@
...
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.kafka.KafkaContainer;
+
+/**
+ * A resource for managing a Schema Registry instance in embedded tests.
+ * This class extends TestcontainerResource to provide a Schema Registry container.
+ * Confluent schema registry is commonly used for managing schemas in Kafka-based applications but not
+ * natively supported by testcontainers.
+ */
+public class SchemaRegistryResource extends TestcontainerResource<GenericContainer<?>>
```
kfaraz: You may name it either `KafkaSchemaRegistryResource` or `ConfluentSchemaRegistryResource` to avoid ambiguity.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/SchemaRegistryResource.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.apache.druid.testing.embedded.TestcontainerResource;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.kafka.KafkaContainer;
+
+/**
+ * A resource for managing a Schema Registry instance in embedded tests.
+ * This class extends TestcontainerResource to provide a Schema Registry container.
+ * Confluent schema registry is commonly used for managing schemas in Kafka-based applications but not
+ * natively supported by testcontainers.
```
kfaraz: These lines may be removed once we rename this class.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/KafkaResource.java
```diff
@@ -128,6 +128,17 @@ public void produceRecordsToTopic(List<ProducerRecord<byte[], byte[]>> records)
...
 
+  /**
+   * Produces records to a topic of this embedded Kafka server.
+   */
+  public void produceRecordsToTopic(List<ProducerRecord<byte[], byte[]>> records)
+  {
+    produceRecordsToTopicWithExtraProperties(
+        records,
+        null
+    );
```
kfaraz: Nit: can probably be in a single line
```suggestion
    produceRecordsToTopicWithExtraProperties(records, null);
```

integration-tests-ex/tools/src/main/java/org/apache/druid/testing/tools/AvroSchemaRegistryEventSerializer.java
```diff
@@ -49,6 +48,22 @@ public class AvroSchemaRegistryEventSerializer extends AvroEventSerializer
...
+    this.config = null;
+    this.client = new CachedSchemaRegistryClient(
+        StringUtils.format("http://%s", schemaRegistryHost),
+        Integer.MAX_VALUE,
+        ImmutableMap.of(
+            "basic.auth.credentials.source", "USER_INFO",
+            "basic.auth.user.info", "druid:diurd"
+        ),
+        ImmutableMap.of()
+    );
```
kfaraz: This snippet can probably be put in a utility method, in say `KafkaUtil`, as it seems to be used in multiple places.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Embedded test of kafka indexing for different data formats.
+ * Formats Included:
+ * <ul>
+ * Avro (with and without schema registry)
+ * CSV
+ * JSON
+ * Protobuf (with and without schema registry)
```
kfaraz: Each of these would need to be put inside `<li></li>` tags.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+import org.junit.jupiter.api.Timeout;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Embedded test of kafka indexing for different data formats.
```
kfaraz:
```suggestion
 * Embedded test of Kafka indexing for different data formats.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+      public void start()
+      {
+        super.start();
+      }
+
+      @Override
+      public void stop()
+      {
+        super.stop();
+      }
```
kfaraz: Not needed.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/SchemaRegistryResource.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  protected GenericContainer<?> createContainer()
+  {
+    return new GenericContainer<>(SCHEMA_REGISTRY_IMAGE)
+        .dependsOn(kafkaContainer)
+        .withExposedPorts(9081)
+        .withEnv("SCHEMA_REGISTRY_HOST_NAME", "schema-registry")
+        .withEnv("SCHEMA_REGISTRY_LISTENERS", "http://0.0.0.0:9081")
+        .withEnv("SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS", kafkaContainer.getBootstrapServers());
```
kfaraz: Instead of depending on `KafkaContainer`, we should use `KafkaResource.getBootstrapServerUrl()`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/SchemaRegistryResource.java
```diff
@@ -0,0 +1,48 @@
...
+ * A resource for managing a Schema Registry instance in embedded tests.
+ * This class extends TestcontainerResource to provide a Schema Registry container.
+ * Confluent schema registry is commonly used for managing schemas in Kafka-based applications but not
+ * natively supported by testcontainers.
+ */
+public class SchemaRegistryResource extends TestcontainerResource<GenericContainer<?>>
+{
+  private static final String SCHEMA_REGISTRY_IMAGE = "confluentinc/cp-schema-registry:latest";
+
+  KafkaContainer kafkaContainer;
```
kfaraz: We should pass `KafkaResource` in the constructor instead.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+    };
+
+    indexer.addProperty("druid.segment.handoff.pollDuration", "PT0.1s")
+           .addProperty("druid.worker.capacity", "10");
+    overlord.addProperty("druid.indexer.task.default.context", "{\"useConcurrentLocks\": true}")
+            .addProperty("druid.manager.segments.useIncrementalCache", "ifSynced")
+            .addProperty("druid.manager.segments.pollDuration", "PT0.1s")
+            .addProperty("druid.manager.segments.killUnused.enabled", "true")
+            .addProperty("druid.manager.segments.killUnused.bufferPeriod", "PT0.1s")
+            .addProperty("druid.manager.segments.killUnused.dutyPeriod", "PT1s");
```
kfaraz: These can be skipped if these tests are not doing any kill testing.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+      @Override
+      public void stop()
+      {
+        super.stop();
+      }
+    };
+
+    indexer.addProperty("druid.segment.handoff.pollDuration", "PT0.1s")
+           .addProperty("druid.worker.capacity", "10");
+    overlord.addProperty("druid.indexer.task.default.context", "{\"useConcurrentLocks\": true}")
```
kfaraz: This can be skipped if we don't plan to run any compaction in these tests.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+      {
+        this.kafkaContainer = kafkaServer.getContainer();
+        super.start();
+      }
+
+      @Override
+      public void stop()
+      {
+        super.stop();
+      }
```
kfaraz: This might not be needed if we pass `KafkaResource` in the constructor.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+    cluster.addExtension(KafkaIndexTaskModule.class)
+           .addExtension(ProtobufExtensionsModule.class)
+           .addExtension(AvroExtensionsModule.class)
+           .useLatchableEmitter()
+           .addCommonProperty("druid.monitoring.emissionPeriod", "PT0.1s")
+           .addCommonProperty(
+               "druid.monitoring.monitors",
+               "[\"org.apache.druid.java.util.metrics.JvmMonitor\","
+               + "\"org.apache.druid.server.metrics.TaskCountStatsMonitor\"]"
+           )
```
kfaraz: We can remove these if monitoring is not needed.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+           .addServer(overlord)
+           .addServer(indexer)
+           .addServer(broker)
+           .addServer(historical);
+
+    return cluster;
+  }
+
+  @Test
+  @Timeout(value = 30, unit = TimeUnit.SECONDS)
```
kfaraz: Nit: shorthand since default unit is seconds.
```suggestion
  @Timeout(30)
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+ *
+ * This tests both InputFormat and Parser. Parser is deprecated for Kafka Ingestion, and those tests will be removed in the future.
+ */
+public class KafkaDataFormatsTest extends EmbeddedClusterTestBase
+{
+  private static final long CYCLE_PADDING_MS = 100;
+  private static final int EVENTS_PER_SECOND = 6;
+
+
+  private static ObjectMapper jsonMapper;
```
kfaraz: This can be a constant or you can use the mapper bindings from one of the servers, say:
```
overlord.bindings().jsonMapper()
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+           .addServer(historical);
+
+    return cluster;
+  }
+
+  @Test
+  @Timeout(value = 30, unit = TimeUnit.SECONDS)
+  public void test_indexKafka_avroDataFormat_withParser() throws IOException
+  {
+    new AvroExtensionsModule().getJacksonModules().forEach(jsonMapper::registerModule);
```
kfaraz: This won't be needed if we use the mapper bound to one of the servers, preferably the `EmbeddedOverlord`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,716 @@
...
+    kafkaServer.createTopicWithPartitions(dataSource, 3);
+    EventSerializer serializer = jsonMapper.readValue("{\"type\": \"avro\"}", EventSerializer.class);
+    int recordCount = generateStreamAndPublishToKafka(dataSource, serializer, false);
+    String jsonString = "{"
```
kfaraz: For all of the JSON strings, please try to use an existing builder or add a new one if necessary.
You may add it in `Resources` class or a new class which sits next to `TaskBuilder`.

embedded-tests/src/test/resources/test-data/avro/input_format.json
```diff
@@ -0,0 +1,81 @@
+{
+  "type": "avro_stream",
```
kfaraz: Since we are using inline JSON strings / builders for the other stuff, can we also use a builder for this schema definition?

kfaraz (commented):
Thanks for the PR, @capistrant !
I have left some thoughts.

kfaraz:
@capistrant , can we also delete the `ITKafkaIndexingServiceDataFormatTest` in this PR?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
capistrant: don't hire me to build a website 😆

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/SchemaRegistryResource.java
capistrant: Do you think there is risk of startup issues removing the dependency? `dependsOn(kafkaContainer)` uses testcontainers api to not start the schema registry until Kafka has has started, making sure that schema registry won't error out if it starts up before the Kafka container is up.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
capistrant: The current implementation does wait on a task count metric, but I removed the jvm monitor.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
capistrant: Lastest commits stop using all the crazy json string --> object mappings. I get why that was a bad idea. Lazy me going from the old ITs which were processing files. But I don't think I fully follow the suggestion on using a builder. Do you have an example you are thinking of from another embedded test?

capistrant:
> @capistrant , can we also delete the `ITKafkaIndexingServiceDataFormatTest` in this PR?

I was about to delete it but then I saw that two other TestNG groups reference it so I got worried about just nuking it since I wasn't considering those cluster configurations for this testing. It is the CDS stuff. Are you more familiar with that and what the downsides would be of me deleting the whole test?

capistrant:
@kfaraz The detailed review is greatly appreciated! I began making changes in response. but I still have not looked into why all the other ITs are failing now so I assume there are more changes to be done yet. Left a clarifying question in your comment about using a builder instead of json strings for creating bits of the supervisor specs. Thanks!

kfaraz:
> I was about to delete it but then I saw that two other TestNG groups reference it so I got worried about just nuking it since I wasn't considering those cluster configurations for this testing. It is the CDS stuff. Are you more familiar with that and what the downsides would be of me deleting the whole test?

Ah, thanks for the clarification. In that case, we can leave it for now.
We will remove it once all the CDS tests have been migrated too.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,722 @@
...
+        new JSONPathSpec(true, null),
+        null,
+        false
+    );
+
+    // Manually create parser map for deprecated withParserMap method
+    Map<String, Object> parserMap = new HashMap<>();
+    parserMap.put("type", "protobuf");
+    parserMap.put("parseSpec", overlord.bindings().jsonMapper().convertValue(parseSpec, Map.class));
+    parserMap.put("protoBytesDecoder", overlord.bindings().jsonMapper().convertValue(protobufBytesDecoder, Map.class));
```
kfaraz: Why convert them? Would it work to simply have this instead:

```suggestion
    parserMap.put("protoBytesDecoder", protobufBytesDecoder);
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,722 @@
...
+          Map.of("schema.registry.url", StringUtils.format("http://%s:%s", cluster.getEmbeddedHostname().toString(), schemaRegistry.getContainer().getMappedPort(9081)))
+      );
+    } else {
+      kafkaServer.produceRecordsToTopic(producerRecords);
+    }
+    return producerRecords.size();
+  }
+
+  // Uses parser which is deprcated for KafkaIndexingService
+  private KafkaSupervisorSpec createDeprectatedKafkaSupervisor(String supervisorId, String topic, Map<String, Object> parserMap)
```
kfaraz: Nit: I don't think the method name should have deprecated in it.
We can leave that info in the javadoc of the method.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,722 @@
...
+        "\t",
+        null,
+        WIKI_DIM_LIST,
+        false,
+        0
+    );
+
+    StringInputRowParser parser = new StringInputRowParser(parseSpec, null);
+
+    Map<String, Object> parserMap = overlord.bindings().jsonMapper().convertValue(parser, Map.class);
```
kfaraz: I am not sure why `DataSchema` accepts a map rather than an actual `InputRowParser`.
It might have something to do with the modules that these classes were added to originally.
But now, we can add a constructor which accepts an `InputRowParser` to avoid the unnecessary conversion.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,722 @@
...
+        new StringDimensionSchema("unpatrolled"),
+        new StringDimensionSchema("newPage"),
+        new StringDimensionSchema("robot"),
+        new StringDimensionSchema("anonymous"),
+        new StringDimensionSchema("namespace"),
+        new StringDimensionSchema("continent"),
+        new StringDimensionSchema("country"),
+        new StringDimensionSchema("region"),
+        new StringDimensionSchema("city")
+    ));
```
kfaraz: For simplicity, you can use 
```java
DimensionsSpec.builder().setDefaultSchemaDimensions(
List.of(
      "page",
      "language", 
      "..."
)
).build();

File: embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java [hunk 23]
```diff
@@ -0,0 +1,722 @@
...
+    EventSerializer serializer = new AvroEventSerializer();
+    int recordCount = generateStreamAndPublishToKafka(dataSource, serializer, false);
+    
+    Map<String, Object> avroSchema = createWikipediaAvroSchemaMap();
+    
+    InlineSchemaAvroBytesDecoder avroBytesDecoder = overlord.bindings().jsonMapper().readValue(
+        StringUtils.format("{\"type\": \"schema_inline\", \"schema\": %s}", 
+                           overlord.bindings().jsonMapper().writeValueAsString(avroSchema)),
+        InlineSchemaAvroBytesDecoder.class
+    );
```
kfaraz: Can't we use constructor of `InlineSchemaAvroBytesDecoder` itself?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
kfaraz: I have left some suggestions that might get us closer to the builder pattern.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
capistrant: ya, sorry. missed these. The next commit will get rid of all the json string parsing

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
capistrant: Next commit will have a take on this. I was having serde issues with an additional constructor that allows an InputRowParser. That is how I understood your suggestion. I instead changed the existing DataSchema constructor to allow InputRowParser, only if the existing standard of parserMap is not present. let me know what you think

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
capistrant: I think the json mapper was broken for this decoder 🤔 my next commit will attempt to fix

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
capistrant: ok, this was a mistake. I'd say don't review latest changes until I change this approach cuz it had unintended consequences

capistrant: I went back to the old InputRowParser --> Map approach (but more consolidated). I know not ideal. I can try some other approaches tomorrow

capistrant:
I am at a bit of a loss about the standard-its. I feel like the druid services can't start cuz the test helper jars moved. I will need to get the standard-its running on my local I guess, to see the actual startup errors.

kfaraz:
Hmm, this is odd. I would have thought that we don't need to include the `druid-it-tools` as a dependency in `distribution/pom.xml` since they are used only by the client side of the integration-tests code. The services shouldn't be using it anyway as it is test-only code.

extensions-core/protobuf-extensions/src/test/java/org/apache/druid/data/input/protobuf/ProtobufInputRowParserTest.java
```diff
@@ -110,6 +113,24 @@ public void setUp()
     decoder = new FileBasedProtobufBytesDecoder("prototest.desc", "ProtoTestEvent");
   }
 
+  @Test
+  public void testSerde() throws Exception
+  {
+
+
```
kfaraz: Typo: extra newlines

integration-tests-ex/tools/src/main/java/org/apache/druid/testing/tools/SleepModule.java
```diff
@@ -20,17 +20,15 @@
...
 
 public class SleepModule implements DruidModule
 {
 
   @Override
   public void configure(Binder binder)
   {
-    SqlBindings.addOperatorConversion(binder, SleepOperatorConversion.class);
-    ExpressionModule.addExprMacro(binder, SleepExprMacro.class);
+    // No-op: Sleep operator is already registered by extensions-core/testing-tools SleepModule
```
kfaraz: In that case, can we just remove this duplicate `SleepModule` class?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaSchemaRegistryResource.java
```diff
@@ -0,0 +1,51 @@
...
+import org.testcontainers.containers.GenericContainer;
+
+/**
+ * A resource for managing a Schema Registry instance in embedded tests.
+ */
+public class KafkaSchemaRegistryResource extends TestcontainerResource<GenericContainer<?>>
+{
+  private static final String SCHEMA_REGISTRY_IMAGE = "confluentinc/cp-schema-registry:latest";
+
+  KafkaResource kafkaResource;
```
kfaraz: private final

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/KafkaResource.java
```diff
@@ -110,12 +110,12 @@ public void deleteTopic(String topicName)
...
   }
 
-  /**
-   * Produces records to a topic of this embedded Kafka server.
-   */
-  public void produceRecordsToTopic(List<ProducerRecord<byte[], byte[]>> records)
+  public void produceRecordsToTopicWithExtraProperties(
+      List<ProducerRecord<byte[], byte[]>> records,
+      Map<String, Object> extraProperties
+  )
```
kfaraz:
```suggestion
  public void produceRecordsToTopic(
      List<ProducerRecord<byte[], byte[]>> records,
      Map<String, Object> overrideProducerProperties
  )
```

integration-tests-ex/tools/src/main/java/org/apache/druid/testing/tools/StreamGenerator.java
```diff
@@ -32,4 +34,8 @@ public interface StreamGenerator
    * Runs and returns the number of messages written.
    */
   long run(String streamTopic, StreamEventWriter streamEventWriter, int totalNumberOfSeconds, DateTime overrrideFirstEventTime);
+
+  List<byte[]> generate(int totalNumberOfSeconds);
+
+  List<byte[]> generate(int totalNumberOfSeconds, DateTime overrideFirstEventTime);
```
kfaraz: Please add short javadocs to these. Rename to `generateRecords()` or `generateEvents()`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+                                                            "continent", "country", "region", "city", "added",
+                                                            "deleted", "delta");
+
+  private final EmbeddedBroker broker = new EmbeddedBroker();
+  private final EmbeddedIndexer indexer = new EmbeddedIndexer();
+  private final EmbeddedOverlord overlord = new EmbeddedOverlord();
+  private final EmbeddedHistorical historical = new EmbeddedHistorical();
+  private final EmbeddedCoordinator coordinator = new EmbeddedCoordinator();
+  private KafkaResource kafkaServer;
+  private KafkaSchemaRegistryResource schemaRegistry;
```
kfaraz: I guess these two can be final too.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+  public EmbeddedDruidCluster createCluster()
+  {
+    final EmbeddedDruidCluster cluster = EmbeddedDruidCluster.withEmbeddedDerbyAndZookeeper().useContainerFriendlyHostname();
+
+    kafkaServer = new KafkaResource();
+
+    schemaRegistry = new KafkaSchemaRegistryResource(kafkaServer);
+
+    indexer.addProperty("druid.segment.handoff.pollDuration", "PT0.1s")
+           .addProperty("druid.worker.capacity", "20");
```
kfaraz: Does this cluster need such a large capacity?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+
+    waitForDataAndVerifyIngestedEvents(dataSource, recordCount);
+    stopSupervisor(dataSource);
+  }
+
+  private void waitForDataAndVerifyIngestedEvents(String dataSource, int expectedCount)
+  {
+    // Wait for the task to succeed
+    overlord.latchableEmitter().waitForEventAggregate(
+        event -> event.hasMetricName("task/success/count")
```
kfaraz: You can wait on `task/run/time` metric instead with a status dimension of SUCCESS.
Then you can avoid adding the `TaskCountStatsMonitor`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource),
+        agg -> agg.hasSumAtLeast(1)
+    );
+    // Wait for the broker to discover the realtime segments
+    broker.latchableEmitter().waitForEvent(
+        event -> event.hasDimension(DruidMetrics.DATASOURCE, dataSource)
+    );
+
+    // Verify the count of rows ingested into the datasource so far
+    Assertions.assertEquals(StringUtils.format("%d", expectedCount), cluster.runSql("SELECT COUNT(*) FROM %s", dataSource));
```
kfaraz:
```suggestion
    Assertions.assertEquals(String.valueOf(expectedCount), cluster.runSql("SELECT COUNT(*) FROM %s", dataSource));
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+  }
+
+  /**
+   * Creates a KafkaSupervisorSpec with a parser instead of an InputFormat.
+   * <p>Parsers are deprecated for kafka indexing, but need to remain tested until they are removed.</p>
+   *
+   * @param supervisorId the ID of the supervisor
+   * @param topic        the Kafka topic to consume from
+   * @param parser       the parser map to use in the supervisor spec
+   * @return a KafkaSupervisorSpec with the provided parser map
```
kfaraz: We can probably remove these.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+        CYCLE_PADDING_MS
+    );
+    List<byte[]> records = streamGenerator.generate(10);
+
+    ArrayList<ProducerRecord<byte[], byte[]>> producerRecords = new ArrayList<>();
+    for (byte[] record : records) {
+      producerRecords.add(new ProducerRecord<>(topic, record));
+    }
+
+
```
kfaraz: Typo: Extra newline

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+    if (useSchemaRegistry) {
+      kafkaServer.produceRecordsToTopicWithExtraProperties(
+          producerRecords,
+          Map.of("schema.registry.url", StringUtils.format("http://%s:%s", cluster.getEmbeddedHostname().toString(), schemaRegistry.getContainer().getMappedPort(9081)))
```
kfaraz: Please expose a method in `KafkaSchemaRegistryResource` which returns the correct URL.
See `PostgreSQLMetadataResource.onStarted(cluster)` method for an example of constructing a URL
with the embedded hostname.

```suggestion
          Map.of("schema.registry.url", schemaRegistry.getRegistryUrl()))
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+    // Wait for the task to succeed
+    overlord.latchableEmitter().waitForEventAggregate(
+        event -> event.hasMetricName("task/success/count")
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource),
+        agg -> agg.hasSumAtLeast(1)
+    );
+    // Wait for the broker to discover the realtime segments
+    broker.latchableEmitter().waitForEvent(
+        event -> event.hasDimension(DruidMetrics.DATASOURCE, dataSource)
+    );
```
kfaraz: You can use the following condition instead to avoid flakiness.
Since we are already waiting for a task to complete, the segments are already published and are not realtime anymore.

```java
broker.latchableEmitter().waitForEvent(
        event -> event.hasMetricName("segment/schemaCache/refresh/count")
    );
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+        false,
+        0,
+        null
+    );
+    KafkaSupervisorSpec supervisorSpec = createKafkaSupervisor(dataSource, dataSource, inputFormat);
+
+    final Map<String, String> startSupervisorResult = cluster.callApi().onLeaderOverlord(
+        o -> o.postSupervisor(supervisorSpec)
+    );
+    Assertions.assertEquals(Map.of("id", dataSource), startSupervisorResult);
```
kfaraz: Use `cluster.callApi().postSupervisor()` for shorthand.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+  public void test_indexKafka_protobufDataFormatWithSchemaRegistry()
+  {
+    kafkaServer.createTopicWithPartitions(dataSource, 3);
+    EventSerializer serializer = new ProtobufSchemaRegistryEventSerializer(StringUtils.format("%s:%s", cluster.getEmbeddedHostname().toString(), schemaRegistry.getContainer().getMappedPort(9081)));
+    serializer.initialize(dataSource);
+    int recordCount = generateStreamAndPublishToKafka(dataSource, serializer, true);
+    SchemaRegistryBasedProtobufBytesDecoder protobufBytesDecoder = new SchemaRegistryBasedProtobufBytesDecoder(
+        null,
+        null,
+        List.of(StringUtils.format("http://%s:%s", cluster.getEmbeddedHostname().toString(), schemaRegistry.getContainer().getMappedPort(9081))),
```
kfaraz: Please use `schemaRegistry.getRegistryUrl()` instead.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+    waitForDataAndVerifyIngestedEvents(dataSource, recordCount);
+    stopSupervisor(dataSource);
+  }
+
+  @Test
+  @Timeout(30)
+  public void test_indexKafka_protobufDataFormatWithSchemaRegistry()
+  {
+    kafkaServer.createTopicWithPartitions(dataSource, 3);
+    EventSerializer serializer = new ProtobufSchemaRegistryEventSerializer(StringUtils.format("%s:%s", cluster.getEmbeddedHostname().toString(), schemaRegistry.getContainer().getMappedPort(9081)));
```
kfaraz: This method can also move to `KafkaSchemaRegistryResource` as `getRegistryHostAndPort()`

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -0,0 +1,727 @@
...
+  @Test
+  public void test_indexKafka_protobufDataFormat()
+  {
+    kafkaServer.createTopicWithPartitions(dataSource, 3);
+    EventSerializer serializer = new ProtobufEventSerializer();
+
+    int recordCount = generateStreamAndPublishToKafka(dataSource, serializer, false);
+
+    FileBasedProtobufBytesDecoder protobufBytesDecoder = new FileBasedProtobufBytesDecoder(
+        "data/protobuf/wikipedia.desc",
```
kfaraz: Please add this as a constant to `MoreResources.ProtobufData`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaDataFormatsTest.java
```diff
@@ -104,20 +104,26 @@ public class KafkaDataFormatsTest extends EmbeddedClusterTestBase
...
-  private KafkaSchemaRegistryResource schemaRegistry;
+  private final KafkaResource kafkaServer = new KafkaResource();
+  private final KafkaSchemaRegistryResource schemaRegistry = new KafkaSchemaRegistryResource(kafkaServer);
+
+  @AfterEach
+  public void afterEach() throws Exception
+  {
+    // I had an issue with tasks piling up, even when I added the supervisor stop at the end of each test.
+    // This is a workaround to ensure that the tasks are cleared before the next test runs.
+    // This is not ideal, but it works for now. Why are tasks not hard stopped when the supervisor is terminated?
```
kfaraz, line 115: Thanks for noting this down, we can address this later.

kfaraz (approved):

PR #18391, opened 2025-08-12
Make multi-stage-query (MSQ engine) a core capability of Druid rather than an extension

kfaraz:
### Description

Druid MSQ engine offers a very convenient method of performing SQL-based batch ingestion, export, etc.

Given the importance and heavy adoption of MSQ, this patch converts MSQ into a core capability of Druid
rather than an extension, as was always intended.

### Changes

- Move all MSQ classes to `services` module without renaming the package
- Move the module loader file `org.apache.druid.guice.DruidModule` to `services/src/main/resources/META-INF/services`
  - This effectively means that the modules listed in this file will always be loaded on all Druid services as desired
- Update dependencies as needed
- Remove mentions of extension from example `common.runtime.properties` files

### Alternative

Instead of moving everything to `services`, the alternative was to simply move the
`multi-stage-query` module into a top-level module but that would
lead to cyclic dependencies between `services` and `multi-stage-query` since:
a) multi-stage-query depends on services for various classes
b) services is the entry point and must depend on all the core modules that need to be on the classpath

### Pending

- Once initial feedback is received, the docs will also be updated to remove mentions of the multi-stage-query extension.

### Future

- Over time, various MSQ classes may be moved out of `services` module into more appropriate modules such as
`processing`, `server`, `indexing-service` or `sql` as applicable.

### Release note

- Multi-stage-query (MSQ engine) is now a core capability of Druid rather than an extension.
- Remove `druid-multi-stage-query` from `druid.extensions.loadList` in `common.runtime.properties`.

### Upgrade notes

While upgrading, the extension `druid-multi-stage-query` must be removed from `druid.extensions.loadList`.
Otherwise, Druid services would fail to start as they would not be able to locate this extension.

<hr>

This PR has:

- [ ] been self-reviewed.


kfaraz:
Closing in favor of #18394

PR #18172, opened 2025-06-24
Web console: Change to use overlord APIs vs coordinator APIs

vogievetsky:
This is the console follow up to:

- https://github.com/apache/druid/pull/17545 (added new Overlord APIs for data-management)
- https://github.com/apache/druid/pull/17935 (completely removes support for updating segments from the Coordinator)
- https://github.com/apache/druid/pull/17999 (document the API changes)

PR #17999, opened 2025-05-13
Docs: Add details of Overlord data management APIs

kfaraz:
### Description

#17545 added new Overlord APIs for data-management,
but it also supported updating segments from the Coordinator as a backup mechanism.
#17935 completely removes support for updating segments from the Coordinator.

All consumers of data-management APIs must migrate to the new paths now.

### Release note

- All of the following APIs, have been moved from the Coordinator to the Overlord.
- The corresponding Coordinator APIs are now deprecated and will be removed in a future release.
- For the time being, they simply redirect the request to the Overlord.
- The request and response payloads are backwards compatible.

|API|Method|Deprecated Coordinator path|New Overlord path|
|---|------|---------------------------|-----------------|
|Mark a single segment as used|`POST`|`/druid/coordinator/v1/datasources/{datasource}/segments/{segmentId}`|`/druid/indexer/v1/datasources/{datasource}/segments/{segmentId}`|
|Mark a single segment as unused|`DELETE`|`/druid/coordinator/v1/datasources/{datasource}/segments/{segmentId}`|`/druid/indexer/v1/datasources/{datasource}/segments/{segmentId}`|
|Mark a group of non-overshadowed segments as used|`POST`|`/druid/coordinator/v1/datasources/{datasource}/markUsed`|`/druid/indexer/v1/datasources/{datasource}/markUsed`|
|Mark a group of segments as unused|`POST`|`/druid/coordinator/v1/datasources/{datasource}/markUnused`|`/druid/indexer/v1/datasources/{datasource}/markUnused`|
|Mark all non-overshadowed segments of a datasource as used|`POST`|`/druid/coordinator/v1/datasources/{datasource}`|`/druid/indexer/v1/datasources/{datasource}`|
|Mark all segments of a datasource as unused|`DELETE`|`/druid/coordinator/v1/datasources/{datasource}`|`/druid/indexer/v1/datasources/{datasource}`|

<hr>

This PR has:

- [ ] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
[...]

PR #18395, opened 2025-08-12
Fix broken kill unused segments endpoint in UI

jtuglu1:
Fixes #XXXX.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.


gianm:
This changed in #18172, along with a bunch of other APIs, as part of the work to move segment metadata management to the Overlord (started in #17545). The change may have been over-zealous, since it looks like this particular API did not move to the Overlord.

gianm:
@kfaraz was it intentional that this API did not move? IMO, it's odd for this one to remain on the Coordinator, given that `GET` and `DELETE` on `/druid/coordinator/v1/datasources/{datasource}` (apparently its parent) did move to the Overlord.

kfaraz:
@gianm , at the time of the original PR #17545 , we only moved APIs that mark segments as used or unused
as they were the only ones needed for consistent behaviour with segment metadata caching enabled on the Overlord.

The list of updated APIs was called out in the original PR as well as the docs PR #17999 .
But I must have missed it during the review of web-console changes in #18172.

#### Move this API?

I did have plans to eventually move this API to the Overlord too as it just calls `overlordClient.runKillTask()`.
But I was undecided on the API path to use and some other details.

> it's odd for this one to remain on the Coordinator, given that GET and DELETE on /druid/coordinator/v1/datasources/{datasource} (apparently its parent) did move to the Overlord.

- `DELETE /coordinator/datasources/{datasource}/intervals/{interval}` launches a `kill` task
- Whereas other `DELETE /druid/coordinator/v1/datasources/...` APIs mark segments as unused
- Using DELETE method is also somewhat incorrect because a successful response here doesn't indicate
successful deletion of any resource, just that a task was submitted to the Overlord.
- Given that the API just translates to calling `/task` on the Overlord with a simple payload as below,
__do we even need this API?__
```json
{
"type": "kill",
"dataSource": "<your-datasource>",
"interval": "<your-kill-interval>"
}
```

__Suggested next steps:__

- Mark this coordinator API as deprecated in Druid 35.
- Delete all of the deprecated coordinator APIs in Druid 36.
- Update web-console to call Overlord `/task` API with the payload suggested above.
- This will be useful in the future, when we expose other `kill` task options on the web-console such as specifying `versions` to kill etc.
- I don't think this warrants a patch release.
- We can call it out as a known issue in the docs and suggest mitigation steps ("Submit a JSON task" option or enable auto-kill on the Coordinator or embedded kill on the Overlord).

@gianm , @jtuglu1 , please let me know what you think.

PR #18398, opened 2025-08-13
Cleanup `cron-job-its` action

amaechler:
### Description

- Address a typo
- Remove the `strategy.fail-fast` directive, which only makes sense when using a `matrix`

<hr>

This PR has:

- [x] been self-reviewed.


abhishekrb19 (approved):
Makes sense, thanks @amaechler! 😄

PR #18402, opened 2025-08-14
Add compaction templates and `CompactionJobQueue`

kfaraz:
### Note

A lot of the changes related to compaction template implementations and persisting templates in Druid catalog were once a part of this PR but have been removed until there is consensus on the best approach.

This PR now deals with only refactoring the `OverlordCompactionScheduler` to use the `CompactionJobQueue` and other related changes.

#### Functionality

- Reset the compaction job queue every 15 minutes by default.
- Whenever a compaction task finishes, check if jobs pending in the queue can be launched.
- When a compaction supervisor is stopped, remove its jobs from the queue.
- When a compaction supervisor is started (or restarted), create jobs for it and add them to the queue.

#### Classes

- Add `BatchIndexingJob` which may contain either a `ClientTaskQuery` or a `ClientSqlQuery` (for MSQ jobs).
- Add `BatchIndexingJobTemplate` that can create jobs for a given source and destination
- Add `CompactionConfigBasedJobTemplate` which implements `CompactionJobTemplate`
- Update `CompactionSupervisor` to create jobs using templates
- Add `CompactionJobQueue` to create and submit compaction jobs to the Overlord

#### Refactor for reuse

- Move common code from `CompactSegments` to `CompactionSlotManager`, `CompactionSnapshotBuilder`
- Update `CompactionStatus`, `CompactionStatusTracker` and `DataSourceCompactibleSegmentIterator`

### Future work

- Have a common `BatchIndexingSupervisor` which uses templates to create jobs.
   - It could be implemented by `ScheduledBatchSupervisor` and `CompactionSupervisor`.
   - This change was originally included in this patch but has been left out to keep the changes small.
- Reset the job queue (if required) when the cluster-level compaction config changes
- Allow force reset of the job queue by calling an API per supervisor (e.g. supervisor reset) and overall
- Cancel tasks which are not aligned with the supervisor spec anymore
- Do not queue up jobs for recently compacted intervals even if the segment timeline has been updated
- Determine if any change was made to a time chunk of the segment timeline in the latest metadata store poll

### Release note

TODO

<hr>

This PR has:

- [ ] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingCompactionTemplate.java
```diff
@@ -0,0 +1,257 @@
...
+  )
+  {
+    final List<CompactionJob> allJobs = new ArrayList<>();
+
+    // Include future dates in the first rule
+    final DateTime currentTime = jobParams.getScheduleStartTime();
+    DateTime previousRuleStartTime = DateTimes.MAX;
+    for (int i = 0; i < rules.size() - 1; ++i) {
+      final CompactionRule rule = rules.get(i);
+      final DateTime ruleStartTime = rule.computeStartTime(currentTime, rules.get(i + 1));
```
uds5501: Q: considering the signature of the method `computeStartTime(DateTime referenceTime, CompactionRule beforeRule)`, adding a `rules.get(i+1)` as `beforeRule` sounds a bit counter-intuitive. 

Does it mean that the list of rules being passed here is meant to be executed in reverse?

uds5501 (commented):
Had a qq -

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingCompactionTemplate.java
kfaraz: Yes, the `beforeRule` name is explained in the javadoc of the `computeStartTime` method.
It comes before the current rule in the segment timeline.
But the rules can be evaluated in any order really, since the time chunks of two rules are mutually exclusive.

The rule with the latest time period comes first in the list since that is how we typically think of compaction,
and also how we define load rules in Druid. It keeps the rule definitions intuitive and straightforward.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingCompactionTemplate.java
```diff
@@ -0,0 +1,257 @@
...
+    );
+
+    // Filter out jobs if they are outside the search interval
+    final List<CompactionJob> validJobs = new ArrayList<>();
+    for (CompactionJob job : allJobs) {
+      final Interval compactionInterval = job.getCandidate().getCompactionInterval();
+      if (searchInterval.contains(compactionInterval)) {
+        validJobs.add(job);
+      }
+    }
```
imply-cheddar: Why would you get jobs outside of the interval?  That seems like a problem with the contract or the specific implementation rather than a concern that everybody who ever calls the method needs to apply?

indexing-service/src/main/java/org/apache/druid/indexing/compact/CatalogCompactionJobTemplate.java
```diff
@@ -0,0 +1,114 @@
...
+    final ResolvedTable resolvedTable = catalog.resolveTable(tableId);
+    if (resolvedTable == null) {
+      return null;
+    }
+
+    final BatchIndexingJobTemplate delegate
+        = resolvedTable.decodeProperty(IndexingTemplateDefn.PROPERTY_PAYLOAD);
+    if (delegate instanceof CompactionJobTemplate) {
+      return (CompactionJobTemplate) delegate;
+    } else {
```
imply-cheddar: Perhaps this is an established pattern.  But I would never have thought that you are supposed to get the table from the catalog and then pass a magical parameter to "decodeProperty" in order to get it to become an object that is a template.

Why isn't there a method that's like `asJobTemplate()` or `as()` or something like that?  What's `PROPERTY_PAYLOAD` and why is it special?  Will a decoded property always generate a `BatchIndexingJobTemplate`?  What does "decode property" have to do with generating a `BatchIndexingJobTemplate`?

Maybe as I read more of the code I'll understand, but only reading the usage-side of this is not very intuitive.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CatalogCompactionJobTemplate.java
```diff
@@ -0,0 +1,114 @@
...
+
+  @Override
+  public List<CompactionJob> createCompactionJobs(
+      DruidInputSource source,
+      CompactionJobParams params
+  )
+  {
+    final CompactionJobTemplate delegate = getDelegate();
+    if (delegate == null) {
+      return List.of();
```
imply-cheddar: This would mean that the table doesn't actually exist right?  Should we offer some sort of indication that there is a reference to a table that doesn't exist?  an exception?  a log line?  Something that can help figure out that there's an issue with how things are setup?

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionConfigBasedJobTemplate.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    return jobs;
+  }
+
+  @Override
+  public String getType()
+  {
+    throw new UnsupportedOperationException("This template type cannot be serialized");
+  }
```
imply-cheddar, line 100: What is someone supposed to do if this exception gets thrown?  Why would it have happened?  Are there any hints that can be provided to the developer who sees this and needs to fix it?  Also, it should probably be either a `DruidException.defensive()` or a `NotYetImplemented` exception.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobParams.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.druid.indexing.template.JobParams;
+import org.apache.druid.server.compaction.CompactionSnapshotBuilder;
+import org.apache.druid.server.coordinator.ClusterCompactionConfig;
+import org.apache.druid.timeline.SegmentTimeline;
+import org.joda.time.DateTime;
+
+/**
+ * Parameters used while creating a {@link CompactionJob} using a {@link CompactionJobTemplate}.
+ */
+public class CompactionJobParams implements JobParams
```
imply-cheddar: In some other code, you are adjsuting the input with a searchInterval in order to limit the time interval seen.  When I initially read that, I wondered "isn't that a param?  why isn't it on the param object".  After seeing this class, I still don't have a good answer, why isn't that a param on the param object?

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobParams.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public SegmentTimeline getTimeline(String dataSource)
+  {
+    return timelineProvider.getTimelineForDataSource(dataSource);
+  }
+
+  public CompactionSnapshotBuilder getSnapshotBuilder()
+  {
+    return snapshotBuilder;
+  }
```
imply-cheddar: Where is the right place to find the contract of what these objects are and what they do and why they exist?  If I'm just a lowly developer trying to create a new CompactionTask thingie and I'm passing in a `CompactionJobParams` how do I go about figuring out what the semantics of the things that were given to me are and what I need to do with them?

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
```diff
@@ -0,0 +1,303 @@
...
+
+import javax.annotation.Nullable;
+import java.util.Map;
+import java.util.PriorityQueue;
+
+/**
+ * Iterates over all eligible compaction jobs in order of their priority.
+ * A fresh instance of this class must be used in every run of the
+ * {@link CompactionScheduler}.
+ */
```
imply-cheddar, line 75: What's the scale of compaction jobs?  Like, how many do we expect this to be iterating at any point in time?

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -107,43 +121,57 @@ public class OverlordCompactionScheduler implements CompactionScheduler
...
+      DruidInputSourceFactory druidInputSourceFactory,
       ScheduledExecutorFactory executorFactory,
+      BrokerClient brokerClient,
       ServiceEmitter emitter,
       ObjectMapper objectMapper
   )
   {
+    final long segmentPollPeriodSeconds =
+        segmentManagerConfig.getPollDuration().toStandardDuration().getMillis();
+    this.schedulePeriodMillis = Math.min(5_000, segmentPollPeriodSeconds);
```
imply-cheddar: A 5 second poll is pretty rapid, what's the logic behind the need for such a rapid poll and why it won't be a significant resource burden?

server/src/main/java/org/apache/druid/indexing/template/JobParams.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.apache.druid.indexing.template;
+
+import org.joda.time.DateTime;
+
+/**
+ * Provides parameters required to create a {@link BatchIndexingJob}.
+ */
+public interface JobParams
```
imply-cheddar: Why interface instead of class?  It seems like this interface is very likely to only ever carry `getters` and it's unclear to me why it's important that different classes can implement this instead of just having a reference to one of these lying around.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CatalogCompactionJobTemplate.java
kfaraz: Yes, this felt hacky to me too.  Unfortunately, the Druid catalog currently only understands a "table" as the top-level object. So, I just stuck to that model for the time being. I have also put in some relevant points in the PR descrioption under "Open questions".

@clintropolis , do you have any suggestions on what would be the preferred approach to store an indexing/compaction template in the catalog?

server/src/main/java/org/apache/druid/indexing/template/JobParams.java
kfaraz: Fixed, made `JobParams` a concrete class.
We would still need to have a `CompactionJobParams` which extends `JobParams` since compaction job templates use some extra stuff like `CompactionSnapshotBuilder` and `ClusterCompactionConfig`.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobParams.java
kfaraz: Added javadocs for these methods. Please let me know if they do not seem adequate.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionConfigBasedJobTemplate.java
kfaraz: Converted to a defensive exception with a better error message.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CatalogCompactionJobTemplate.java
kfaraz: Fixed to throw a `NotFound` exception with the proper error message.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
kfaraz: Each time chunk (using the target segment granularity) of each datasource would translate to a single compaction job.

On a large cluster with say 1 year of hourly segment data for 10 datasources, this number could easily be 365 * 24 * 10 = ~80k.

Every time the queue is reset, we re-create all of the jobs and check which ones are already done and which ones need to be queued up.
Some of the work is wasteful and we may optimize it in follow up PRs to simply not re-create jobs for intervals which we know to be already compacted.

I have added some metrics to easily monitor the size of the job queue.

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
kfaraz: Yes, this value has been carried over from the original iteration of the `OverlordCompactionScheduler`.
I had intended the scheduler to be able to pick up compaction jobs as soon as compaction task slots become available but it is wasteful to recompute the entire queue for that, especially since on large clusters, recomputation of the entire queue may take up to a couple of minutes (seen from `coordinator/time` metric on large clusters).

We can do the following instead:
- Keep the schedule period as 5 minutes (or may be even higher?)
- When the scheduler kicks in, reset the job queue.
- We already receive a callback whenever a task completes.
- When a task completes and slots become available, check if there is any pending job still in the queue from the last scheduled run. If there is a pending job, launch that.

Please let me know if this makes sense.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobParams.java
kfaraz: The search interval seemed like a better fit for the `InputSource` since it defines the time range of the data that is the input for a job template.

It seemed redundant to include it in `CompactionJobParams` too since the passed `DruidInputSource` already contains the search interval.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingCompactionTemplate.java
kfaraz: Thanks for catching this. This should not be needed anymore.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusTracker.java
```diff
@@ -86,50 +75,46 @@ public Set<String> getSubmittedTaskIds()
...
-
-    // Skip intervals that violate max allowed input segment size
-    final long inputSegmentSize = config.getInputSegmentSizeBytes();
-    if (candidate.getTotalBytes() > inputSegmentSize) {
-      return CompactionStatus.skipped(
-          "'inputSegmentSize' exceeded: Total segment size[%d] is larger than allowed inputSegmentSize[%d]",
-          candidate.getTotalBytes(), inputSegmentSize
-      );
-    }
+    final CompactionStatus pendingStatus = CompactionStatus.pending("not compacted yet");
```
capistrant: nit: renaming this status to pending when there is an actual status called pending that may not be the final status we return feels odd. I didn't mind the previous name or maybe something like `computedStatus`, not sure. Big scheme, not a big deal and up to you if you just want to leave it as is.

server/src/main/java/org/apache/druid/server/compaction/CompactionSlotManager.java
```diff
@@ -0,0 +1,358 @@
...
+        .collect(Collectors.toList());
+    final Map<String, List<Interval>> datasourceToLockedIntervals =
+        new HashMap<>(FutureUtils.getUnchecked(overlordClient.findLockedIntervals(lockFilterPolicies), true));
+    log.debug(
+        "Skipping the following intervals for Compaction as they are currently locked: %s",
+        datasourceToLockedIntervals
+    );
+
+    // Skip all the intervals locked by higher priority tasks for each datasource
+    // This must be done after the invalid compaction tasks are cancelled
```
capistrant: should this suggestion/requirement be in the javadoc instead of a normal code comment?

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionSupervisor.java
```diff
@@ -51,6 +54,32 @@
...
+  {
+    return !supervisorSpec.isSuspended();
+  }
+
+  /**
+   * Creates compaction jobs for this supervisor.
+   */
+  public List<CompactionJob> createJobs(
+      DruidInputSource inputSource,
+      DruidDatasourceDestination destination,
```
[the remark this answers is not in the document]
capistrant: bumping this notice for awareness. seems like signature needs refactor

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionStateMatcher.java
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * Target compacted state of segments used to determine if compaction is needed
+ * for an interval. An explicitly defined target state helps avoid superfluous
+ * compaction when only the job definition has changed.
+ * <p>
+ * This class is mostly a duplicate of {@code CompactionState} but is kept
+ * separate to allow:
+ * <ul>
+ * <li>fields to be nullable so that only non-null fields are used for matching</li>
+ * <li>legacy "compaction-incompatible" fields to be removed</li>
```
capistrant: I think this may be me coming late to the party on compaction. For my sake, could you elaborate on this list item to get me up to speed. I guess I'm unclear on what you refer to as compaction-incompatible fields

indexing-service/src/main/java/org/apache/druid/indexing/compact/MSQCompactionJobTemplate.java
```diff
@@ -0,0 +1,191 @@
...
+              candidate,
+              CompactionSlotManager.getMaxTaskSlotsForMSQCompactionTask(sqlTemplate.getContext())
+          )
+      );
+    }
+
+    return jobs;
+  }
+
+  private ClientSqlQuery createQueryForJob(String dataSource, Interval compactionInterval)
```
capistrant: wondering aloud on if there is a way to make the query and formatting more flexible/extensible if folks want to use more than these standard format vars. Perhaps if someone wanted to do some filtering during compaction? Or would you suggest that if a person desires that, that it would be in some custom template type that they roll on their own?

capistrant: I guess since this is just inline, the filter could be defined inline. But if someone changed the query after it existed for some time that filter wouldn't be re-applied to already compacted segments since it is not persisted in the target state.

indexing-service/src/main/java/org/apache/druid/indexing/compact/MSQCompactionJobTemplate.java
```diff
@@ -0,0 +1,191 @@
+/*
```
capistrant: when I was playing with this on my local I accidentally got into a state where I had defined an expected granularity of MONTH, but the query in the template was DAY. It looked like it just went into an infinite compact loop. Could we force the templated query to honor segment granularity from the state matcher by injecting it? I guess this forces everyone creating a rule to select a segment gran though, which is probably not desired.

kfaraz: Allowing the state matcher to be distinct from the query is by design as it allows us to make improvements the query without worrying about all the intervals being recompacted when not desired.

For the time being, it is upto the user to ensure that the state matcher and the query are compatible with each other.
In future versions, we will include some kind of template validation.

indexing-service/src/main/java/org/apache/druid/indexing/compact/MSQCompactionJobTemplate.java
kfaraz: Yes, the filter can simply be included in the SQL.

> But if someone changed the query after it existed for some time that filter wouldn't be re-applied to already compacted segments since it is not persisted in the target state.

Yes, that is by design. We don't want compaction to be triggered as long as the compaction state has not changed.
This ensures that every minor tweak made to the SQL query does not end up recompacting everything.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionStateMatcher.java
kfaraz: Updated the javadoc to clarify this point. But it mostly refers to things like the `transformSpec` which can filter out data or even the `dimensionsSpec` where we may choose to drop some dimensions and thus cause an irreversible aggregation of the data.

(Side note: the more desirable way to achieve this pre-aggregation to improve query perf would be to use projections).

server/src/main/java/org/apache/druid/server/compaction/CompactionSlotManager.java
kfaraz: updated, thanks for the suggestion!

kfaraz:
Since there is a lack of clarity on how the catalog would be used to persist compaction templates, the catalog bits have been removed from this PR for the time being.

For posterity, some possible options for storing compaction templates are:

  a. As a table inside a new schema `index_template`(currently used in this PR)
  b. OR as a table inside the `druid` schema: Currently used for datasources only
  c. OR as a single row inside `sys`.`templates`: probably not preferable since the catalog models everything as tables and their properties, but this would be neither.

Note: In all of the above cases, the template is always physically stored as a single row in `druid_tableDefs` in the metadata store.

  d. A separate metadata table `druid_indexingTemplates`. The schema used to access the contents would still be one of a, b, or c.

gianm (approved):
LGTM

kfaraz:
Thanks for the reviews, @capistrant , @gianm , @uds5501 , @cheddar !

PR #18403, opened 2025-08-14
Consolidate the conversion between Granularity and VirtualColumn, and improve the mapping of granularity usage in projections.

cecemei:
### Description

Consolidate the conversion between `Granularity` and `VirtualColumn`, and improve the mapping of granularity usage in projections.

Before this PR, projection has issues handling granularity with time zones, examples:
- TimeseriesQuery with Pacific time hourly granularity can't use UTC hourly projection. 
- GroupbyQuery with Indian time hourly granularity is matched to UTC hourly projection falsely (it's 5h30m offset)

After this PR, the granularity matches as following:
- TimeseriesQuery with Pacific time hourly granularity can use UTC hourly projection. 
- GroupbyQuery with Indian time hourly granularity can't use UTC hourly projection.

This is done through:
- `VirtualColumn` -> `Granularity` should be able to handle all `TimestampFloorExpr` regardless of the time zone.
- `Granularity` -> `VirtualColumn` should always be able to convert `PeriodGranularity` to `TimestampFloorExpr`, regardless of the time zone.
- `PeriodGranularity` has a new `canBeMappedTo` function, in a simple way, a finer gran can be mapped to a coarser gran. It also considers timezone, origin, month/year, week, day, hour compatible stuff. E.x. 
  - same zone hour can be mapped to day, 2 hour can be mapped to 4 hour, but can not be mapped to 3 hour.
  - different zone sometimes hour can also be mapped to day, like pacific time, but indian time can not 
  - week can not be mapped to month, we simplify this by not allowing week/day/hour period if there's month/year
- query will almost always have a __gran `VirtualColumn`, unless `Granularities.ALL`.
- projections would check its __gran `canBeMappedTo` to query __gan, this is more restrictive than `isFinerThan`.

**Additional restrictions on projection granularity:**
- it has to be in UTC time zone and null origin. This is consistent with the `__time` column. Without this requirement, we can't simply map `timeColumnName` in projection to `__time`.

<hr>

 * `Projections`
 * `Granularities`
 * `PeriodGranularity`
 * `PeriodGranularityTest`
 * `CursorFactoryProjectionTest`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
+++ b/processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
@@ -224,8 +224,7 @@ public class AggregateProjectionSpec
     String timeColumnName = null;
     Granularity granularity = null;
 
-    // determine the granularity and time column name for the projection, based on the first time-like grouping column.
-    // if there are multiple time-like grouping columns, they must be "coarser" than the first time-like grouping column.
+    // determine the granularity and time column name for the projection, based on the finest time-like grouping column.
     for (final DimensionSchema dimension : groupingColumns) {
       ordering.add(OrderBy.ascending(dimension.getName()));
       if (ColumnHolder.TIME_COLUMN_NAME.equals(dimension.getName())) {
--- a/processing/src/main/java/org/apache/druid/segment/projections/Projections.java
+++ b/processing/src/main/java/org/apache/druid/segment/projections/Projections.java
@@ -21,12 +21,12 @@ package org.apache.druid.segment.projections;
 
 import org.apache.druid.data.input.impl.AggregateProjectionSpec;
 import org.apache.druid.error.InvalidInput;
-import org.apache.druid.java.util.common.StringUtils;
 import org.apache.druid.java.util.common.granularity.Granularities;
 import org.apache.druid.java.util.common.granularity.Granularity;
 import org.apache.druid.java.util.common.granularity.PeriodGranularity;
 import org.apache.druid.query.QueryContexts;
 import org.apache.druid.query.aggregation.AggregatorFactory;
+import org.apache.druid.query.cache.CacheKeyBuilder;
 import org.apache.druid.query.filter.Filter;
 import org.apache.druid.segment.AggregateProjectionMetadata;
 import org.apache.druid.segment.CursorBuildSpec;
@@ -70,7 +70,12 @@ public class Projections
         if (name != null && !name.equals(spec.getSchema().getName())) {
           continue;
         }
-        final ProjectionMatch match = matchAggregateProjection(spec.getSchema(), cursorBuildSpec, dataInterval, physicalChecker);
+        final ProjectionMatch match = matchAggregateProjection(
+            spec.getSchema(),
+            cursorBuildSpec,
+            dataInterval,
+            physicalChecker
+        );
         if (match != null) {
           if (cursorBuildSpec.getQueryMetrics() != null) {
             cursorBuildSpec.getQueryMetrics().projection(spec.getSchema().getName());
@@ -400,7 +405,9 @@ public class Projections
                    && projection.getEffectiveGranularity() instanceof PeriodGranularity) {
           PeriodGranularity virtualGran = (PeriodGranularity) virtualGranularity;
           PeriodGranularity projectionGran = (PeriodGranularity) projection.getEffectiveGranularity();
-          byte[] combinedKey = StringUtils.toUtf8(projectionGran + "->" + virtualGran);
+          byte[] combinedKey = new CacheKeyBuilder((byte) 0x0).appendCacheable(projectionGran)
+                                                              .appendCacheable(virtualGran)
+                                                              .build();
           if (PERIOD_GRAN_CACHE.computeIfAbsent(combinedKey, (unused) -> projectionGran.canBeMappedTo(virtualGran))) {
             return matchBuilder.addReferencedPhysicalColumn(ColumnHolder.TIME_COLUMN_NAME);
           }


processing/src/main/java/org/apache/druid/java/util/common/granularity/Granularities.java
```diff
@@ -194,14 +192,24 @@ public static ExpressionVirtualColumn toVirtualColumn(Granularity granularity, S
...
-      }
+      return fromExpr(((ExpressionVirtualColumn) virtualColumn).getParsedExpression().get());
+    }
+    return null;
+  }
+
+  @Nullable
+  private static Granularity fromExpr(Expr expr)
+  {
+    String identifier = expr.getIdentifierIfIdentifier();
```
clintropolis: wny switch to `getIdentifierIfIdentifier`? It doesn't matter really here probably, but `getBindingIfIdentifier` is more correct since the binding refers to the underlying column input instead of the variable name in the expression

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
```diff
@@ -213,16 +214,19 @@ public Schema(
...
       this.orderingWithTimeSubstitution = Lists.newArrayListWithCapacity(ordering.size());
-      Granularity granularity = null;
+      Granularity granularity = Granularities.ALL;
       for (int i = 0; i < ordering.size(); i++) {
         OrderBy orderBy = ordering.get(i);
         if (orderBy.getColumnName().equals(timeColumnName)) {
           orderingWithTimeSubstitution.add(new OrderBy(ColumnHolder.TIME_COLUMN_NAME, orderBy.getOrder()));
+          if (foundTimePosition != -1) {
+            throw DruidException.defensive("projection[%s] has multiple time columns in ordering: %s", name, ordering);
+          }
```
clintropolis: I guess this shouldn't be able to happen right now since we generate the ordering so the same entry shouldn't ever show up

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
```diff
@@ -213,16 +214,19 @@ public Schema(
...
           foundTimePosition = i;
           timeColumnName = groupingColumns.get(foundTimePosition);
           final VirtualColumn vc = this.virtualColumns.getVirtualColumn(groupingColumns.get(foundTimePosition));
           if (vc != null) {
-            granularity = Granularities.fromVirtualColumn(vc);
+            granularity = MoreObjects.firstNonNull(Granularities.fromVirtualColumn(vc), Granularities.ALL);
```
clintropolis: this seems less clear to me than just initializing to null and letting it naturally fill in like it was doing previously

processing/src/main/java/org/apache/druid/java/util/common/granularity/Granularities.java
```diff
@@ -194,14 +192,24 @@ public static ExpressionVirtualColumn toVirtualColumn(Granularity granularity, S
...
+    if (identifier != null) {
+      return identifier.equals(ColumnHolder.TIME_COLUMN_NAME)
+             ? Granularities.NONE
+             : Granularities.ALL;
```
clintropolis, line 209: ALL i guess because this isn't a time grouping?  maybe worth a comment

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
```diff
@@ -230,10 +230,15 @@ private static ProjectionOrdering computeOrdering(VirtualColumns virtualColumns,
...
-        if (granularity == null && maybeGranularity != null) {
+        if (maybeGranularity == null || maybeGranularity.equals(Granularities.ALL)) {
+          // no __time in inputs or not supported, skip
+          continue;
+        }
+        if (granularity == null) {
           granularity = maybeGranularity;
           timeColumnName = dimension.getName();
-        } else if (granularity != null && maybeGranularity != null && maybeGranularity.isFinerThan(granularity)) {
+        } else if (maybeGranularity.isFinerThan(granularity)) {
```
clintropolis: i think this check needs to make sure there is no timezone on the granularity to be considered as the time column since i think __time is assumed to be UTC and so any column on the projection being considered as the substitute __time column must also be utc. This seems worth adding a test for

processing/src/main/java/org/apache/druid/java/util/common/granularity/PeriodGranularity.java
```diff
@@ -216,6 +226,139 @@ public String toString()
...
+   * <li>it has years or months
+   * <li>it has milliseconds
+   */
+  public static Long getStandardSecondsOrThrow(Period period)
+  {
+    Optional<Long> s = getStandardSeconds(period);
+    if (s.isPresent()) {
+      return s.get();
+    } else {
+      throw DruidException.defensive("Period[%s] cannot be converted to standard whole seconds", period);
```
clintropolis: this seems expensive to throw an exception, if there are multiple projections to match against a query which has a timezone, this method is going to be called for every projection we try to match of every segment

processing/src/main/java/org/apache/druid/java/util/common/granularity/PeriodGranularity.java
```diff
@@ -216,6 +226,139 @@ public String toString()
            '}';
   }
 
+  /**
+   * Returns true if this granularity can be mapped to the target granularity. For example:
+   * <li>Period('PT1H') in UTC can be mapped to Period('P1D') in UTC</li>
+   * <li>Period('PT1H') in America/Los_Angeles can be mapped to Period('PT1H') in UTC</li>
+   * <li>Period('P1D') in America/Los_Angeles cannot be mapped to Period('P1D') in UTC</li>
+   */
+  public boolean canBeMappedTo(PeriodGranularity target)
```
clintropolis, line 250: admittedly i'm still digesting how exactly this method works, but it seems kind of expensive to do to match for every projection we consider of every segment when the timezones don't match (at least `getUtcMappablePeriodSecondsOrThrow` seems expensive). Perhaps we should make a cache of these conversions so we can re-use the work we've done since its likely going to be a lot of the same checks over and over?

clintropolis: also these methods feel like they could use some additional comments to make it clearer what is going on and why this works

processing/src/main/java/org/apache/druid/java/util/common/granularity/Granularities.java
cecemei: switched to `getBindingIfIdentifier`

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
cecemei: yes this should not happen, I think I was thinking about the case when there're multiple virtual columns using __time, this is not the best way to handle it, anyway, reverted change in this file.

processing/src/main/java/org/apache/druid/segment/AggregateProjectionMetadata.java
cecemei: reverted change

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
cecemei: added the requirement and test

processing/src/main/java/org/apache/druid/java/util/common/granularity/PeriodGranularity.java
cecemei: true, updated this to return Optional<Long> instead

processing/src/main/java/org/apache/druid/java/util/common/granularity/PeriodGranularity.java
cecemei: added some additional comments. also added `PERIOD_GRAN_CACHE` in `Projections.java`.

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
```diff
@@ -230,12 +234,40 @@ private static ProjectionOrdering computeOrdering(VirtualColumns virtualColumns,
...
+            continue;
+          } else {
+            // we don't allow:
+            // 1. virtual column on __time if there's no grouping on __time
+            // 2. non-UTC or non-epoch origin period granularity
+            throw InvalidInput.exception(
+                "cannot use granularity[%s] on column[%s] as projection time column",
+                maybeGranularity,
+                dimension.getName()
+            );
```
clintropolis: I don't think this needs to be an exception, we just need to not consider it as a time column, the same way as a coarser granularity would not be __time. These columns can still be substituted for query virtual columns, so there is value in being able to pre-compute them, they just cannot serve as a replacement for __time (or time floor on __time) in queries.

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
```diff
@@ -230,12 +234,40 @@ private static ProjectionOrdering computeOrdering(VirtualColumns virtualColumns,
...
+                                                       && ((PeriodGranularity) granularity).canBeMappedTo((PeriodGranularity) maybeGranularity))) {
+          // keep existing granularity
+        } else {
+          throw InvalidInput.exception(
+              "cannot map time granularity[%s] on column[%s] to time granularity[%s] on column[%s], failed to determine projection time column",
+              granularity,
+              timeColumnName,
+              maybeGranularity,
+              dimension.getName()
+          );
```
clintropolis: again, not an error condition imo. a projection doesn't need to necessarily include __time as a grouping column, if there is no __time column then the granularity is effectively the segment granularity, and we use a constant __time column where all values are the start time of the segment.

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
```diff
@@ -218,10 +223,9 @@ private static ProjectionOrdering computeOrdering(VirtualColumns virtualColumns,
 
     String timeColumnName = null;
     Granularity granularity = null;
-    // try to find the __time column equivalent, which might be a time_floor expression to model granularity
-    // bucketing. The time column is decided as the finest granularity on __time detected. If the projection does
-    // not have a time-like column, the granularity will be handled as ALL for the projection and all projection
-    // rows will use a synthetic timestamp of the minimum timestamp of the incremental index
+
+    // determine the granularity and time column name for the projection, based on the first time-like grouping column.
+    // if there are multiple time-like grouping columns, they must be "coarser" than the first time-like grouping column.
```
clintropolis: this is too restrictive, the __time column does not need to be first, we just need to replace any candidate __time column with the finer granularity column, since what we are looking for is the finest granularity column we can use as a substitute for __time in any queries against the base table

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -375,13 +379,21 @@ public static ProjectionMatchBuilder matchQueryVirtualColumn(
...
-        if (virtualGranularity.isFinerThan(projection.getEffectiveGranularity())) {
-          return null;
-        }
         // same granularity, replace virtual column directly by remapping it to the physical column
         if (projection.getEffectiveGranularity().equals(virtualGranularity)) {
           return matchBuilder.remapColumn(queryVirtualColumn.getOutputName(), ColumnHolder.TIME_COLUMN_NAME)
                              .addReferencedPhysicalColumn(ColumnHolder.TIME_COLUMN_NAME);
+        } else if (virtualGranularity.equals(Granularities.NONE)
+                   || projection.getEffectiveGranularity().equals(Granularities.ALL)) {
+          return null;
```
clintropolis: what is this check for none and all separately for? Are there other query virtual column granularities that are ok with an all granularity projection?

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -44,6 +46,8 @@
 
 public class Projections
 {
+  private static final Map<byte[], Boolean> PERIOD_GRAN_CACHE = new HashMap<>();
```
clintropolis: this needs to be a concurrent map or use a lock or something

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
cecemei: ah i was thinking about some cases when there's a time-like virtual column (non-utc) and another time-like virtual column (utc, can be converted to gran), e.x. `PT1H` in Pacific and `P1D` in UTC, if we just ignore, we would just use floor(__time, `P1D`) as the time column, and the virtual column `PT1H` in Pacific would just be computed based on  floor(__time, `P1D`), that would be incorrect. 

it seems simple and safe to just assume the first time-like grouping column must be gran, and the following must be coarser, e.x. `PT1H` in UTC and `P1D` in Pacific is an acceptable config. Another case is `PT2H`, `PT3H`, `PT1H`, it seems a bit more complex (or calculated) if we don't have the FIRST time-like must be gran assumption.

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
cecemei: if query granularity is NONE, and projection gran is not (would already been caught by the first if statement), we just can't map, e.x. format(__time, 'YYYY-MM-DD HH:mm:ss') query can't use hourly projection. This exits early.

but i also reformatted this part to be more readable, so return null should be at bottom now

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
cecemei: updated to use `ConcurrentHashMap`, no lock should be fine since we're only using `computeIfAbsent`

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
cecemei: i should be more specific, previous this is using a simple isFinerThan comparison, which is not safe, because `PT2H` is finer than `PT3H` but the former can't be mapped to the latter. now we're switching to `canBeMappedTo`, but it only applies to period gran, so adding this special handling for `Granularities.NONE` and `Granularities.ALL`.

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
cecemei: `finer` is not completely safe, we're looking for the most compatible gran, e.x. between `PT2H` and `PT3H`, we can find `PT1H` as compatible (but this is not currently supported, i'm just using this as an example). 

and deciding on the most compatible is tricky since we don't want non-UTC time to sneak in, e.x. `PT30M` in Pacific time would sort of invalidate our `PT1H` gran in UTC, handling this just makes things more complicated, thus choose to handle in a strict way here

cecemei: actually the compatible is `PT6H`. i realized projection need to find coarse ones as compatible

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
cecemei: true, updated this

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
cecemei: updated this logic to handle the gran, it should never throw exception now, skips non-utc and non-mappable vc

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
cecemei: actually realized non-UTC vc would just be treated as regular grouping column, for `PT2H` and `PT3H`, it'd just choose the finer one `PT2H`.

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
clintropolis: this comment is stale now since it no longer requires time to be first and additional time groupings to be coarser

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -383,17 +388,30 @@ public static ProjectionMatchBuilder matchQueryVirtualColumn(
...
                              .addReferencedPhysicalColumn(ColumnHolder.TIME_COLUMN_NAME);
+        } else if (Granularities.ALL.equals(virtualGranularity)
+                   || Granularities.NONE.equals(projection.getEffectiveGranularity())) {
+          // if virtual gran is ALL or projection gran is NONE, it's guaranteed that projection gran can be mapped to virtual gran
+          return matchBuilder.addReferencedPhysicalColumn(ColumnHolder.TIME_COLUMN_NAME);
+        } else if (virtualGranularity instanceof PeriodGranularity
+                   && projection.getEffectiveGranularity() instanceof PeriodGranularity) {
+          PeriodGranularity virtualGran = (PeriodGranularity) virtualGranularity;
+          PeriodGranularity projectionGran = (PeriodGranularity) projection.getEffectiveGranularity();
+          byte[] combinedKey = StringUtils.toUtf8(projectionGran + "->" + virtualGran);
```
clintropolis: it might be better to use `getCacheKey()` from the granularities instead of toString conversion, this whole bit could be 
```
byte[] combinedKey = new CacheKeyBuilder(0).appendCacheable(projectionGran).appendCacheable(virtualGran).build()
```

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
cecemei: updated comment

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
cecemei: updated

PR #17937, opened 2025-04-18
Raw SQL Support on HTTP endpoints

FrankChen021:
A part of #17769.

### Description

Allow raw sql in the HTTP body on http endpoints. This helps reduce the client side complexity which requires users to do character escaping when the SQL is placed in the JSON document. The following example demonstrates the usage:

```bash
echo 'SELECT 1' | curl -X POST "http://localhost:8082/druid/v2/sql" --data @-
```

#### Release note

Druid now supports raw text format SQL on `/druid/v2/sql` endpoint. The content in the HTTP body of a HTTP request which does not set `Content-Type` to `application/json` will be treated as a raw sql text.

##### Key changed/added classes in this PR
1. Only HTTP requests with `Content-Type: application/json` will be parsed as traditional JSON format SQL
2. All other HTTP requests including those with empty Content-Type will be parsed as raw SQL
3. Router also supports raw text SQL, but when it forwards queries to brokers, it uses the JSON format for simplicity

<hr>

This PR has:

- [x] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [x] added documentation for new or modified features or behaviors.
- [x] a release note entry in the PR description.
- [x] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [x] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [x] added integration tests.
- [x] been tested in a test Druid cluster.

PR #18404, opened 2025-08-14
Restoring support for parameters in media types for the sql query endpoint

cryptoe:
The PR: https://github.com/apache/druid/pull/17937 added new functionality to the sql query endpoints.

A bug was introduced with this PR where a sql query request going to the router with content type parameters such as ` Content-Type: application/json; text/plain` will always fail.

```
org.apache.druid.server.initialization.jetty.HttpException: Unsupported Content-Type: application/json; charset=UTF-8. Only application/json, text/plain or application/x-www-form-urlencoded is supported.
```

This patch restores the correct original behavior.

### DIFF
--- a/sql/src/main/java/org/apache/druid/sql/http/SqlQuery.java
+++ b/sql/src/main/java/org/apache/druid/sql/http/SqlQuery.java
@@ -277,26 +277,25 @@ public class SqlQuery
       ISqlQueryExtractor<String> rawQueryExtractor
   ) throws HttpException
   {
+    if (contentType == null) {
+      throw new HttpException(Response.Status.BAD_REQUEST, "Missing Content-Type header");
+    }
     try {
-      if (MediaType.APPLICATION_JSON.equals(contentType)) {
+      final MediaType requestMediaType = MediaType.valueOf(contentType);
+      if (MediaType.APPLICATION_JSON_TYPE.isCompatible(requestMediaType)) {
 
         SqlQuery sqlQuery = jsonQueryExtractor.extract();
         if (sqlQuery == null) {
           throw new HttpException(Response.Status.BAD_REQUEST, "Empty query");
         }
         return sqlQuery;
-
-      } else if (MediaType.TEXT_PLAIN.equals(contentType)) {
-
+      } else if (MediaType.TEXT_PLAIN_TYPE.isCompatible(requestMediaType)) {
         String sql = rawQueryExtractor.extract().trim();
         if (sql.isEmpty()) {
           throw new HttpException(Response.Status.BAD_REQUEST, "Empty query");
         }
-
         return new SqlQuery(sql, null, false, false, false, null, null);
-
-      } else if (MediaType.APPLICATION_FORM_URLENCODED.equals(contentType)) {
-
+      } else if (MediaType.APPLICATION_FORM_URLENCODED_TYPE.isCompatible(requestMediaType)) {
         String sql = rawQueryExtractor.extract().trim();
         if (sql.isEmpty()) {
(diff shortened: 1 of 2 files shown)


cryptoe:
@FrankChen021 I think we might need to do a 34.0.1 for this issue since the main sql query endpoint is broken in a backward incompatible way. 
What would happen to users is that the queries would suddenly start failing when they upgrade to 34. 

Is this something you could drive ?

FrankChen021 (approved):
Thanks for the fix

kfaraz:
If we are doing a `34.0.1` release, it would be nice to include this bugfix as well:
https://github.com/apache/druid/pull/18395

cc: @cryptoe , @FrankChen021 , @jtuglu1

cryptoe:
Adding more test cases. Would try to finish soon.

gianm:
> If we are doing a `34.0.1` release, it would be nice to include this bugfix as well: #18395
> 
> cc: @cryptoe , @FrankChen021 , @jtuglu1

Let's discuss this on a dev list thread for 34.0.1, it's off topic for the current PR.

IMO we should do 34.0.1 since this change breaks a relatively common way of specifying the json content-type. I suppose we didn't notice it in testing because our tests all use the `application/json` form rather than setting any options.

> A bug was introduced with this PR where a sql query request going to the router/broker with content type parameters such as  `Content-Type: application/json; text/plain` will always fail.

I believe this example is invalid, I don’t think requests can have multiple content-types. The error message has an example of a valid content-type: `application/json; charset=UTF-8` means `application/json` with the `charset=UTF-8` option.

cryptoe:
> I believe this example is invalid, I don’t think requests can have multiple content-types. The error message has an example of a valid content-type: `application/json; charset=UTF-8` means `application/json` with the `charset=UTF-8` option.

It was a copy issue. Fixed the description. Thanks for catching

kfaraz:
> Let's discuss this on a dev list thread for 34.0.1, it's off topic for the current PR.

Sure, thanks!

sql/src/main/java/org/apache/druid/sql/http/SqlQuery.java
```diff
@@ -343,5 +341,8 @@ private static SqlQuery from(
     catch (IOException e) {
       throw new HttpException(Response.Status.BAD_REQUEST, "Unable to read query from request: " + e.getMessage());
     }
+    catch (IllegalArgumentException e) {
+      throw new HttpException(Response.Status.BAD_REQUEST, "Invalid Content-Type header: " + e.getMessage());
```
gianm: Not sure if this is right for this location. There's a lot going on in the `try` block and maybe something else can throw `IllegalArgumentException` other than an invalid content-type. Why not throw this exception directly in the `else` block of the content-type checking conditional?

gianm (commented):
LGTM other than a question about the exception handling. Thanks.

sql/src/main/java/org/apache/druid/sql/http/SqlQuery.java
cryptoe: Thanks for the catch. Fixed it. The earlier catch location was an attempt to make switch statement work until I dropped that idea in the final cut .

jtuglu1:
Thanks @cryptoe. Looks like some tests are failing – do you want me to fix these up?

cryptoe:
@jtuglu1 Will try to fix them tomorrow.

cryptoe:
Thanks for the review @FrankChen021 , @gianm . 
@gianm Addressed your concerns and merged it. Please feel free to comment if I missed anything. Will raise a follow up.

PR #18406, opened 2025-08-15
Start building out more robust fabric8 retries

capistrant:
### Description

Improve transient fault tolerance of `KubernetesPeonClient#launchPeonJobAndWaitForStart` by adding retry mechanism to two fabric8 calls to external k8s service that will retry in the event of connection closed exceptions.

#### Background

When we switched to vert.x from okhttp for the connection pool underlying fabric8 KubernetesClient, we lost the okhttp feature that sends periodic keepalives on the connections in the k8s connection pool. This feature of okhttp connection pool, proactively kept connections between druid and k8s open. vert.x does not have this keepalive feature, and instead will check connection health on a timed loop, and clean up any connections that are no longer healthy. The default interval of checking these connections is 1 hour. If a connection is closed on the server side in less than an hour, the connection pool is liable to try and use it, which will fail immediately. I have seen this happen in the aforementioned `KubernetesPeonClient#launchPeonJobAndWaitForStart`. This is a mitigation for that issue.

The pattern we see when this happens is usually a prolonged period of heavy traffic between Druid and k8s that expands the vert.x connection pool out to its max allowed size. This is followed by a lull in traffic, where we expect the server is closing idle connections. Finally, a spike in Druid --> k8s traffic results in druid trying to use all of the connections for operations, and any that are closed result in exceptions and hard failures. A retry on this exception should help prevent jobs from failing just because they got a bad connection from the pool.

##### Low level vert.x config exposure

Add more knobs for configuring vert.x connection pool. For instance, have a more aggressive interval to clean up bad connections. Right now we are stuck with mostly all defaults from Vert.x and fabric8 is not making the configs easily modifiable.

##### Refactor of all fabric8 --> k8s interactions in this druid extension

Longer term, cleaning up this extension and standardizing retry mechanisms for all contact with external k8s service is probably needed. I'm sure we can come up with a way to more nicely re-use basic retry code to avoid duplication like my PR + we may want to expose retry logic to configuration somehow (via exception list or something?) or have different modes that dictate how eager druid is to retry vs fail k8s operations.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesPeonClient.java
+++ b/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesPeonClient.java
@@ -88,29 +88,20 @@ public class KubernetesPeonClient
       String jobName = job.getMetadata().getName();
 
       log.info("Submitting job[%s] for task[%s].", jobName, task.getId());
-      client.batch()
-            .v1()
-            .jobs()
-            .inNamespace(namespace)
-            .resource(job)
-            .create();
+      createK8sJobWithRetries(job);
       log.info("Submitted job[%s] for task[%s]. Waiting for POD to launch.", jobName, task.getId());
 
-      // wait until the pod is running or complete or failed, any of those is fine
+      // Wait for the pod to be available
       Pod mainPod = getPeonPodWithRetries(jobName);
-      Pod result = client.pods()
-                         .inNamespace(namespace)
-                         .withName(mainPod.getMetadata().getName())
-                         .waitUntilCondition(pod -> {
-                           if (pod == null) {
-                             return true;
-                           }
-                           return pod.getStatus() != null && pod.getStatus().getPodIP() != null;
-                         }, howLong, timeUnit);
-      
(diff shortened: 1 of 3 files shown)


extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesPeonClient.java
```diff
@@ -353,6 +476,32 @@ private boolean shouldRetryStartingPeonPod(Throwable e)
...
+      Optional<String> messageSubstring = entry.getValue();
+      
+      if (exceptionClass.isInstance(e)) {
+        if (messageSubstring.isPresent()) {
+          return e.getMessage() != null && e.getMessage().contains(messageSubstring.get());
+        } else {
+          return true;
+        }
+      }
+    }
```
abhishekagarwal87: nit: it would be more readable if you remove that constant map and have few if/else conditions here for each exception type. something like 

```
if(exception is httpClientException) ..
if (exception is connectionClosed)...
return false
```

capistrant:
Thanks for the review @abhishekagarwal87! I agree  on your comment. I was getting ahead of myself thinking the map may be re-used in other places. Who knows if that would ever be useful in future so followed your suggestion for now.

PR #18407, opened 2025-08-15
Fail projection filter match early if any child in `AndFilter` failed rewrite

cecemei:
Fail projection filter match early if any child in `AndFilter` failed rewrite. Changes:

- move `rewriteFilter` to `ProjectionFilterMatch`.
- return early if any child from `projectionAndFilter` failed rewrite. Without this change we'd get NPE error from `queryFilter.equals(projectionFilter)`

<hr>

This PR has:

- [x] been self-reviewed.


processing/src/main/java/org/apache/druid/segment/projections/ProjectionFilterMatch.java
```diff
@@ -34,4 +39,86 @@ private ProjectionFilterMatch()
...
+
+  /**
+   * Rewrites a query {@link Filter} if possible, removing the {@link Filter} of a projection. To match a projection
+   * filter, the query filter must be equal to the projection filter, or must contain the projection filter as the child
+   * of an AND filter. This method returns null
+   * indicating that a rewrite is impossible with the implication that the query cannot use the projection because the
+   * projection doesn't contain all the rows the query would match if not using the projection.
+   */
+  @Nullable
+  public static Filter rewriteFilter(Filter projectionFilter, Filter queryFilter)
```
clintropolis, line 51: nit: to me it doesn't quite feel like it belongs here at least since `rewriteFilter` isn't really something that belongs to `ProjectionFilterMatch`, rather `ProjectionFilterMatch` is a tool used by `rewriteFilter`. Between this and where it used to live I think i kind of prefer the old place where this logic to lived with the rest of the projection matching stuff since it all is sort of related. Or if really really want to split out, maybe it should be a standalone `ProjectionFilterRewrite` utility?

Also with this split out, it seems important to indicate that the query filter be rewritten with the projection remapped column names using `rewriteRequiredColumns` before calling this to make sure matching is possible

PR #18413, opened 2025-08-18
Kubernetes embedded tests

uds5501:
Extends K3S setup to integrate druid operator tests with the existing embedded test setup


embedded-tests/pom.xml
```diff
@@ -419,6 +461,17 @@
           <excludedGroups>docker-test</excludedGroups>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
kfaraz: This is not needed here since no other module needs to use the test jar of `embedded-tests`.
If you need to use the test jar of some other module here, please move this block in that module.

extensions-core/druid-testcontainers/src/main/java/org/apache/druid/testing/DruidCommand.java
```diff
@@ -56,6 +60,15 @@ enum Server implements DruidCommand
...
         ),
+        30000,
+        Map.of(
+            "node", "coordinators",
+            "druidServiceType", "coordinator",
+            "healthPath", "/status/health",
+            "readinessProbePath", "/status/health",
+            "sharedStorageDir", "/druid/shared-storage",
+            "metadataName", "test-cluster-coordinator"
+        ),
```
kfaraz: It seems like we can skip these changes.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -36,20 +38,42 @@
...
     command.getDefaultProperties().forEach(properties::setProperty);
   }
 
+  public K3sDruidService governWithOperator()
+  {
+    this.operatorVariablesTemplate = new HashMap<>(command.getOperatorConfiguration());
+    this.isGovernedByOperator = true;
+    this.manifestTemplate = "manifests/druid-service-operator.yaml";
+    return this;
+  }
```
kfaraz: Instead of doing this, pass in the template file to use and try to reuse the template variables such as `${service}` etc.

Edit: If needed, we can add a few extra steps while doing the template replacement in `createManifestYaml`.

extensions-core/druid-testcontainers/src/main/java/org/apache/druid/testing/DruidCommand.java
```diff
@@ -36,6 +36,10 @@ public interface DruidCommand
 
   Map<String, String> getDefaultProperties();
 
+  Integer getExposedOperatorPort();
+
+  Map<String, String> getOperatorConfiguration();
```
kfaraz: We shouldn't need to make any change to `DruidCommand` for the new tests.

extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/simulate/K3SResource.java
```diff
@@ -0,0 +1,331 @@
...
+import java.time.Duration;
+import java.util.Set;
+import java.util.zip.GZIPInputStream;
+
+import static org.testcontainers.containers.BindMode.READ_WRITE;
+
+/**
+ * A K3s container for use in embedded tests.
+ */
+public class K3SResource extends TestcontainerResource<K3sContainer>
```
kfaraz: This class can be removed since we already have `K3sClusterResource`.

uds5501: yeah this will be purged.

extensions-core/druid-testcontainers/src/main/java/org/apache/druid/testing/DruidCommand.java
uds5501: This seemed like a good place for keeping the extra configs and reusing an existing class that I'll require for making operator based setup working, would you recommend I attempt creating a new class for those?

kfaraz: What are the extra configs you need?

uds5501: The   `Integer getExposedOperatorPort();` is required because the current port (ex - 8088) is not supported internally by druid-operator (fails when I attempt to create a `NodePort`)

kfaraz: Since this is a special requirement and not something generic which is always needed when using DruidCommand, you can just pass in the custom port directly to the `K3sDruidService` by exposing a method `usePort` or `setPort`.

uds5501: Alright, will do that.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterResource.java
```diff
@@ -74,18 +74,18 @@ public class K3sClusterResource extends TestcontainerResource<K3sContainer>
...
+  protected final List<File> manifestFiles = new ArrayList<>();
+  protected final List<K3sDruidService> services = new ArrayList<>();
 
-  private KubernetesClient client;
-  private String druidImageName;
+  protected KubernetesClient client;
+  protected String druidImageName;
 
-  private final Closer closer = Closer.create();
+  protected final Closer closer = Closer.create();
```
kfaraz: We should keep these fields private.
We can expose methods if needed.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
```diff
@@ -0,0 +1,313 @@
...
+    this.druidImageName = druidImageName;
+    return this;
+  }
+
+  @Override
+  public K3sClusterWithOperatorResource addService(K3sDruidService service)
+  {
+    services.add(service);
+    return this;
+  }
```
kfaraz: We shouldn't need to override these methods since they are not changing anything from the super class.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
```diff
@@ -0,0 +1,313 @@
...
+    services.forEach(this::waitUntilServiceIsHealthy);
+  }
+
+  private void initializeDruidTestFolders(TestFolder testFolder) {
+    testFolder.getOrCreateFolder("druid-storage");
+    testFolder.getOrCreateFolder("druid-storage/segments");
+    testFolder.getOrCreateFolder("druid-storage/segment-cache");
+    testFolder.getOrCreateFolder("druid-storage/metadata");
+    testFolder.getOrCreateFolder("druid-storage/indexing-logs");
+  }
```
kfaraz: These setups happen by default if we specify the correct configs. We shouldn't need to do this explicitly.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
```diff
@@ -0,0 +1,313 @@
...
+
+  private String prepareCommonPropertiesString(Properties properties, TestFolder testFolder) {
+    StringBuilder sb = new StringBuilder();
+    sb.append("druid.metadata.storage.type=derby\n");
+    sb.append("    druid.metadata.storage.connector.connectURI=jdbc:derby://localhost:1527/var/druid/metadata.db;create=true\n");
+    sb.append("    druid.metadata.storage.connector.host=localhost\n");
+    sb.append("    druid.metadata.storage.connector.port=1527\n");
+    sb.append("    druid.metadata.storage.connector.createTables=true\n");
```
kfaraz: the embedded cluster being used in the new test is already configured to use Psql metadata store. We should remove these configs.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
```diff
@@ -0,0 +1,313 @@
...
+    sb.append("    druid.storage.type=local\n");
+    sb.append("    druid.storage.storageDirectory=" + testFolder.getOrCreateFolder("druid-storage/segments").getAbsolutePath() + "\n");
```
kfaraz: The embedded cluster is already using miniostorage, we shouldn't need these.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
```diff
@@ -0,0 +1,313 @@
...
+    sb.append("    druid.selectors.indexing.serviceName=druid/overlord\n");
+    sb.append("    druid.selectors.coordinator.serviceName=druid/coordinator\n");
```
kfaraz: What are these needed for?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
```diff
@@ -0,0 +1,313 @@
...
+    sb.append("    druid.indexer.logs.type=file\n");
+    sb.append("    druid.indexer.logs.directory=/druid/data/indexing-logs\n");
```
kfaraz: These are also covered by `MinIOStorageResource` IIRC.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -67,22 +91,51 @@ public String createManifestYaml(String druidImage)
...
     }
     catch (Exception e) {
       throw new ISE(e, "Could not create manifest for service[%s]", command);
     }
   }
 
+  private String replaceOperatorVariables(String manifest)
+  {
+    String hostname = EmbeddedHostname.containerFriendly().toString();
+    this.properties.setProperty("druid.host", hostname);
```
kfaraz: This was already being done in the constructor. Let's restore that.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -67,22 +91,51 @@ public String createManifestYaml(String druidImage)
...
+    return manifest;
+  }
+
+  private String prepareNodePropertiesString(Properties nodeProperties)
+  {
+    StringBuilder sb = new StringBuilder();
+    for (String key : nodeProperties.stringPropertyNames()) {
+      sb.append("        ").append(key).append("=").append(nodeProperties.getProperty(key)).append("\n");
+    }
+    return sb.toString().trim(); // Remove trailing newline
```
kfaraz: Might be cleaner to do:

```java
StringWriter writer = new StringWriter();
properties.store(writer, null);
return writer.toString();
```

embedded-tests/pom.xml
```diff
@@ -399,6 +399,48 @@
...
+      <artifactId>docker-java-core</artifactId>
+      <version>3.3.4</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.docker-java</groupId>
+      <artifactId>docker-java-transport-httpclient5</artifactId>
+      <version>3.3.4</version>
+      <scope>test</scope>
+    </dependency>
```
kfaraz: These deps and the fabric8 deps should already be present in this pom.
Please double check and remove duplicates.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterResource.java
uds5501: Fixed.

embedded-tests/pom.xml
uds5501: done, must've missed duplicates while fixing merge conflict.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
uds5501: 1. Using super.addService now.
2. I need to return a different type of resource, hence overrode this.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
uds5501: was needed in http context, don't need them anymore, purged now.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
uds5501: ack, purging and checking.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
uds5501: I agree, however we need to send some common properties to the yaml to ensure that operator loads some pods. I suggest we let these be. (no, empty properties is not working)

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
uds5501: Probably not, I need to make a valid yaml,  I don't see how it'll handle the required indentation. 
Yup. Doesn't work for some reason the runtime bit cries. Used a different builder. (however there i am not indenting and the tests are working so not too sure)

Edit: Nope, indentation was indeed required to get the yaml working.

extensions-core/druid-testcontainers/src/main/java/org/apache/druid/testing/DruidCommand.java
```diff
@@ -164,7 +165,12 @@ enum Server implements DruidCommand
     private final Integer[] exposedPorts;
     private final Map<String, String> defaultProperties;
 
-    Server(String name, String javaOpts, Map<String, String> defaultProperties, Integer... exposedPorts)
+    Server(
+        String name,
+        String javaOpts,
+        Map<String, String> defaultProperties,
+        Integer... exposedPorts
+    )
```
kfaraz: Please revert all the formatting changes to this class as well.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/CustomNodeRoleDockerTest.java
```diff
@@ -89,5 +89,17 @@ public Map<String, String> getDefaultProperties()
...
+    public Integer getExposedOperatorPort()
+    {
+      return 9302;
+    }
+
+    @Override
+    public Map<String, String> getOperatorConfiguration()
+    {
+      return Map.of();
+    }
```
kfaraz: Should not be needed anymore.

uds5501: Yeah I am pushing the final changes, earlier commit was WIP

embedded-tests/pom.xml
uds5501: this too is purged.

uds5501:
Thank you for your patient review @kfaraz , i have accommodated them and replied to a few. Kindly take a look whenever you get a chance.

kfaraz:
Thanks for updating the PR, @uds5501 , it seems much cleaner now!
I will do a final review soon.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterResource.java
```diff
@@ -138,35 +141,61 @@ protected K3sContainer createContainer()
...
+        .withConfig(Config.fromKubeconfig(getContainer().getKubeConfigYaml()))
+        .build();
+    closer.register(client);
 
-    // Wait for all pods to be ready and services to be healthy
-    client.pods().inNamespace(DRUID_NAMESPACE).resources().forEach(this::waitUntilPodIsReady);
-    services.forEach(this::waitUntilServiceIsHealthy);
+    loadLocalDockerImageIntoContainer(druidImageName, cluster.getTestFolder());
+
+    manifestFiles.forEach(this::applyManifest);
```
kfaraz: These 3 seem disjoint -  they should be separate methods:
`initKubernetesClient`
`loadLocalDockerImageIntoContainer()`
`applyManifests()`

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterResource.java
```diff
@@ -138,35 +141,61 @@ protected K3sContainer createContainer()
...
+    services.forEach(this::waitUntilServiceIsHealthy);
+  }
+
+  protected List<K3sDruidService> getServices()
+  {
+    return services;
+  }
 
-    // Create common config map
+  protected Properties injectClusterCommonPropertiesToConfigMap(EmbeddedDruidCluster cluster)
```
kfaraz: If we are already putting the properties in the config map, why do we need to return them too?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -36,20 +36,49 @@
...
 public class K3sDruidService
 {
-  private static final String MANIFEST_TEMPLATE = "manifests/druid-service.yaml";
+  private static final String KEY_COMMON_RUNTIME_PROPERTIES = "${commonRuntimeProperties}";
+  private static final String KEY_NODE_RUNTIME_PROPERTIES = "${nodeRuntimeProperties}";
 
   private final DruidCommand command;
   private final Properties properties;
+  private final Properties commonProperties;
+  private Integer druidPort;
```
kfaraz:
```suggestion
  private int servicePort;
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -36,20 +36,49 @@
  */
 public class K3sDruidService
 {
-  private static final String MANIFEST_TEMPLATE = "manifests/druid-service.yaml";
+  private static final String KEY_COMMON_RUNTIME_PROPERTIES = "${commonRuntimeProperties}";
+  private static final String KEY_NODE_RUNTIME_PROPERTIES = "${nodeRuntimeProperties}";
```
kfaraz: We should inline these two as well since the other params are also inlined.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -67,22 +96,36 @@ public String createManifestYaml(String druidImage)
...
       );
 
       String manifest = StringUtils.replace(template, "${service}", getName());
       manifest = StringUtils.replace(manifest, "${command}", command.getName());
-      manifest = StringUtils.replace(manifest, "${port}", String.valueOf(command.getExposedPorts()[0]));
+      manifest = StringUtils.replace(manifest, "${port}", String.valueOf(druidPort));
       manifest = StringUtils.replace(manifest, "${image}", druidImage);
       manifest = StringUtils.replace(manifest, "${serviceFolder}", getServicePropsFolder());
-
+      manifest = StringUtils.replace(manifest, KEY_COMMON_RUNTIME_PROPERTIES, buildPropertiesString(commonProperties, 4));
```
kfaraz: Rather than passing in the common properties just to have them be substituted in the yaml, we could instead get the yaml _without_ the common properties and then substitute them only where needed.

In fact, we should do that for the service properties too, since it is required only in the special case of Druid operator.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -99,7 +142,7 @@ public String getHealthCheckUrl()
     return StringUtils.format(
         "http://%s:%s/status/health",
         EmbeddedHostname.containerFriendly().toString(),
-        command.getExposedPorts()[0]
+        isGovernedByOperator ? command.getExposedOperatorPort() : command.getExposedPorts()[0]
```
kfaraz: Is this leftover from a previous set of changes?

```suggestion
        servicePort
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -122,4 +165,9 @@ private String getServicePropsFolder()
         throw new IAE("Unsupported command[%s]", server);
     }
   }
+
+  public String getServiceDiscoveryPath()
```
kfaraz: Is this still needed?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
```diff
@@ -0,0 +1,286 @@
...
+    super.usingDruidImage(druidImageName);
+    return this;
+  }
+
+  @Override
+  public K3sClusterWithOperatorResource addService(K3sDruidService service)
+  {
+    super.addService(service);
+    return this;
+  }
```
kfaraz: Why do we need to override these methods?
For the fluent chaining, it's okay if we get back a `K3sClusterResource` (i.e. super class), we can still chain the required methods on it.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
```diff
@@ -0,0 +1,286 @@
...
+
+ private Properties getCombinedCommonProperties(Properties properties)
+  {
+    Properties defaults = new Properties();
+    defaults.setProperty("druid.metadata.storage.type", "derby");
+    defaults.setProperty("druid.metadata.storage.connector.connectURI", "jdbc:derby://localhost:1527/var/druid/metadata.db;create=true");
+    defaults.setProperty("druid.metadata.storage.connector.host", "localhost");
+    defaults.setProperty("druid.metadata.storage.connector.port", "1527");
+    defaults.setProperty("druid.metadata.storage.connector.createTables", "true");
```
kfaraz: Aren't we using PSql DB?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
uds5501: I can't start a druid-operator without common properties, it gives a 422 otherwise.

embedded-tests/src/test/resources/manifests/druid-service-operator.yaml
```diff
@@ -0,0 +1,117 @@
...
+    environment: stage
+    release: alpha
+  podAnnotations:
+    dummy: k8s_extn_needs_atleast_one_annotation
+  volumes:
+    - name: mysqlconnector
+      emptyDir: { }
+  volumeMounts:
+    - name: mysqlconnector
+      mountPath: "/opt/druid/extensions/mysql-connector"
```
kfaraz: We are using PSql in the test setup. Do we need the mysql connector?

extensions-core/kubernetes-overlord-extensions/pom.xml
```diff
@@ -282,6 +306,17 @@
           <skip>true</skip>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>test-jar</goal>
```
kfaraz: Which test classes from this module do we want to use?

embedded-tests/pom.xml
```diff
@@ -399,6 +399,25 @@
       <version>6.2.12</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.druid.extensions</groupId>
+      <artifactId>druid-kubernetes-overlord-extensions</artifactId>
+      <version>35.0.0-SNAPSHOT</version>
```
kfaraz:
```suggestion
      <version>${project.parent.version}</version>
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -36,20 +36,49 @@
...
+    return this;
+  }
+
+  public K3sDruidService withCommonProperties(Properties commonProperties)
+  {
+    this.commonProperties.putAll(commonProperties);
+    return this;
+  }
+
+  public K3sDruidService withDruidPort(Integer druidPort)
```
kfaraz:
```suggestion
  public K3sDruidService usingPort(int servicePort)
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
kfaraz: No, you should use the common properties. But the substitution should happen in `K3sClusterResourceWithOperator` since the set of common properties is already available there.
We shouldn't have to pass in the properties here just to have them be substituted.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/KubernetesClusterWithOperatorDockerTest.java
```diff
@@ -0,0 +1,97 @@
...
+        .useContainerFriendlyHostname()
+        .addResource(k3sCluster)
+        .addServer(overlord)
+        .addServer(broker)
+        .addServer(eventCollector)
+        .addCommonProperty(
+            "druid.extensions.loadList",
+            "[\"druid-s3-extensions\", \"druid-kafka-indexing-service\","
+            + "\"druid-multi-stage-query\", \"postgresql-metadata-storage\","
+            + " \"druid-kubernetes-overlord-extensions\", \"druid-kubernetes-extensions\", \"druid-datasketches\"]"
```
kfaraz: Why do we need these two?
```
\"druid-kubernetes-extensions\", \"druid-datasketches\"
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterResource.java
```diff
@@ -129,6 +129,9 @@ protected K3sContainer createContainer()
         // Druid service is discoverable with the Druid service discovery
         portBindings.add(port + ":" + port);
       }
+      Integer exposedPort = service.getDruidPort();
+      container.addExposedPorts(exposedPort);
```
kfaraz:
```suggestion
      int servicePort = service.getDruidPort();
      container.addExposedPorts(servicePort);
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
```diff
@@ -67,22 +96,36 @@ public String createManifestYaml(String druidImage)
...
     }
     catch (Exception e) {
       throw new ISE(e, "Could not create manifest for service[%s]", command);
     }
   }
 
+  /**
+   * Builds a properties string to be used in the manifest.yaml file supporting a uniform indentation.
+   */
+  private String buildPropertiesString(Properties properties, int indentationSpaces)
```
kfaraz: This method should probably exist in the new class `K3sClusterResourceWithOperator` since that is where we will actually use it for the substitution.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
uds5501: missed pushing it, my bad

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
uds5501: nope

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
uds5501: we are, let me purge this

embedded-tests/src/test/resources/manifests/druid-service-operator.yaml
uds5501: nope, btw it's a stray mount, it doesn't have anythign attached to it because I removed the additional container

extensions-core/kubernetes-overlord-extensions/pom.xml
uds5501: Earlier we wanted to use it in the POC version, I think we can purge this now.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/KubernetesClusterWithOperatorDockerTest.java
uds5501: I was using k8s discovery earlier, don't need these anymore.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterResource.java
uds5501: done

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterResource.java
uds5501: I wanted to utilize the common function to return the common properties that'll be required to be baked in the yaml, this seemed like a decent way to get them. The other way would've been using a `super.getCommonProperties()` I guess. [for the operator context, we don't really need to push the common props to the k8s btw]

kfaraz: The common properties are already available in the cluster object passed in to onStarted. It would be simpler to just get that directly. Then we could make this method `private void`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterWithOperatorResource.java
uds5501: nuked

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
uds5501: yeah replaced.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sClusterResource.java
uds5501: fair, used that

extensions-core/kubernetes-overlord-extensions/pom.xml
uds5501: re-added, its being used in embedded tests

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/K3sDruidService.java
uds5501: sorted now

embedded-tests/pom.xml
```diff
@@ -399,6 +403,25 @@
       <version>6.2.12</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.druid.extensions</groupId>
+      <artifactId>druid-kubernetes-overlord-extensions</artifactId>
+      <version>${project.parent.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
```
kfaraz: This dependency is not needed. Please remove it.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/KubernetesClusterWithOperatorDockerTest.java
```diff
@@ -0,0 +1,82 @@
...
+    final K3sClusterResource k3sCluster = new K3sClusterWithOperatorResource()
+        .usingDruidTestImage()
+        .usingDruidManifestTemplate(MANIFEST_TEMPLATE)
+        .addService(new K3sDruidService(DruidCommand.Server.COORDINATOR).usingPort(30081))
+        .addService(overlordService)
+        .addService(new K3sDruidService(DruidCommand.Server.HISTORICAL).usingPort(30083))
+        .addService(new K3sDruidService(DruidCommand.Server.ROUTER).usingPort(30088))
+        .addService(brokerService);
+
+    // Add an EmbeddedOverlord and EmbeddedBroker to use their client and mapper bindings.
```
kfaraz, line 59: These lines might be needed after all since we are already binding the 8090 port, even though it is not actually being used:

```suggestion
    // Add an EmbeddedOverlord and EmbeddedBroker to use their client and mapper bindings.
    overlord.addProperty("druid.plaintextPort", "7090");
    broker.addProperty("druid.plaintextPort", "7082");
    
```

embedded-tests/pom.xml
```diff
@@ -399,6 +403,25 @@
...
+    </dependency>
```
kfaraz:
```suggestion
```

extensions-core/kubernetes-overlord-extensions/pom.xml
```diff
@@ -282,6 +281,17 @@
...
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
```
kfaraz:
```suggestion
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/KubernetesClusterWithOperatorDockerTest.java
uds5501: Yeah that's fair, adding them

PR #18415, opened 2025-08-18
fix: Fix CacheKeyBuilder to correctly handle empty Cacheable keys

hongkunxu:
**Bug**
CacheKeyBuilder.appendCacheable had a dead-code bug in its null check. The helper method cacheableToByteArray() returns EMPTY_BYTES (not null) when the input Cacheable is null, so the original guard if (cacheableToByteArray == null) could never match that case. Meanwhile, Cacheable.getCacheKey() is declared @Nullable and multiple production classes (e.g., RestrictedDataSource, UnnestDataSource, ExternalDataSource) return null from it to signal "uncacheable." That path happened to work only by coincidence — cacheableToByteArray() returned null implicitly when getCacheKey() returned null, and the == null check caught it.

**Fix**
Refactored cacheableToByteArray to make null handling explicit:

- cacheable == null (absent component) -> returns EMPTY_BYTES, key stays valid. This preserves backward compatibility with production callers that pass nullable fields like query.getDimensionsFilter() or JoinDataSource.leftFilter.
- cacheable.getCacheKey() == null (uncacheable object) -> returns null, which causes appendCacheable to set invalidKey = true and build() returns null.
- cacheable.getCacheKey() returns an empty array -> throws IllegalArgumentException (unchanged behavior).

Also simplified the empty-key check from Arrays.equals(key, EMPTY_BYTES) to key.length > 0.

**Tests**
Added two new test cases in CacheKeyBuilderTest:

- testAppendCacheableWithNullInput — verifies that appendCacheable(null) does not invalidate the key.
- testAppendCacheableWithUncacheableObject — verifies that appending a Cacheable whose getCacheKey() returns null causes build() to return null.

Could some committer help to review this PR?
cc @agile @trotter @mcroydon @StFS

### DIFF
--- a/processing/src/main/java/org/apache/druid/query/cache/CacheKeyBuilder.java
+++ b/processing/src/main/java/org/apache/druid/query/cache/CacheKeyBuilder.java
@@ -33,7 +33,6 @@ import javax.annotation.Nullable;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
@@ -103,15 +102,18 @@ public class CacheKeyBuilder
     return buffer.array();
   }
 
+  @Nullable
   private static byte[] cacheableToByteArray(@Nullable Cacheable cacheable)
   {
     if (cacheable == null) {
       return EMPTY_BYTES;
-    } else {
-      final byte[] key = cacheable.getCacheKey();
-      Preconditions.checkArgument(!Arrays.equals(key, EMPTY_BYTES), "cache key is equal to the empty key");
-      return key;
     }
+    final byte[] key = cacheable.getCacheKey();
+    if (key == null) {
+      return null;
+    }
+    Preconditions.checkArgument(key.length > 0, "cache key is equal to the empty key");
+    return key;
   }
 
   private static byte[] stringCollectionToByteArray(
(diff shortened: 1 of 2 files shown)


FrankChen021 (approved):

hongkunxu:
Hi @FrankChen021 , Could you please tell me why not merge this PR?

FrankChen021:
close and repen to trigger the CI @hongkunxu

hongkunxu:
Hi @FrankChen021 , Could you help move this forward?

GWphua:
@hongkunxu Thanks for the contribution. Please address the problems that were highlighted in the CI.

Notably, 
1. PR title. (prefix title with `fix: `)
2. Checkstyle (Test with `mvn checkstyle:check`)

processing/src/main/java/org/apache/druid/query/cache/CacheKeyBuilder.java
```diff
@@ -269,12 +271,12 @@ public CacheKeyBuilder appendFloatArray(float[] input)
...
   {
-    byte[] cacheableToByteArray = cacheableToByteArray(input);
-    if (cacheableToByteArray == null) {
+    final byte[] bytes = cacheableToByteArray(input);
+    if (bytes == null) {
       invalidKey = true;
       return this;
     }
-    appendItem(CACHEABLE_KEY, cacheableToByteArray);
+    appendItem(CACHEABLE_KEY, bytes);
```
GWphua, line 279: This change is a pure rename right? Can we leave this untouched?

hongkunxu: Since this PR has been open for quite some time and I no longer recall the original context clearly, I revisited it and found the changes are unnecessary. I’ve closed the PR. Thanks @GWphua  for the review.

PR #18416, opened 2025-08-18
Added limit to history in Supervisor

vogievetsky:
This prevents the web console from choking if there is a ton of supervisor history.

Unlike my usual console PRs I am actually modifying the Java API for `supervisor/<id>/history` to have a `count` parameter. I called it `limit` at first but renamed it to count to match other history APIs.

### DIFF
--- a/indexing-service/src/test/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResourceTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResourceTest.java
@@ -1183,12 +1183,12 @@ public class SupervisorResourceTest extends EasyMockSupport
     // Test with limit=0 (should return 400 Bad Request)
     response = supervisorResource.specGetHistory(request, "id1", 0);
     Assert.assertEquals(400, response.getStatus());
-    Assert.assertEquals(ImmutableMap.of("error", "Count must be greater than zero if set"), response.getEntity());
+    Assert.assertEquals(ImmutableMap.of("error", "Count must be greater than zero if set (count was 0)"), response.getEntity());
 
     // Test with negative limit (should return 400 Bad Request)
     response = supervisorResource.specGetHistory(request, "id1", -1);
     Assert.assertEquals(400, response.getStatus());
-    Assert.assertEquals(ImmutableMap.of("error", "Count must be greater than zero if set"), response.getEntity());
+    Assert.assertEquals(ImmutableMap.of("error", "Count must be greater than zero if set (count was -1)"), response.getEntity());
 
     // Test with limit larger than available history
     response = supervisorResource.specGetHistory(request, "id1", 100);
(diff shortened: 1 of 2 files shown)


docs/api-reference/supervisor-api.md
```diff
@@ -1827,6 +1827,12 @@ Retrieves an audit history of specs for a single supervisor.
 
 `GET` `/druid/indexer/v1/supervisor/{supervisorId}/history`
 
+#### Query parameters
+
+* `count` (optional)
+  * Maximum number of history entries to return. If not specified, returns all history entries.
```
cryptoe: I think we might want to call out the sorting order here. No ?

indexing-service/src/test/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResourceTest.java
```diff
@@ -1134,12 +1134,94 @@ public void testSpecGetHistoryWithAuthFailure()
...
+    Assert.assertEquals(200, response.getStatus());
+    Assert.assertEquals(limitedVersions, response.getEntity());
+
+    // Test with limit=0
+    response = supervisorResource.specGetHistory(request, "id1", 0);
+    Assert.assertEquals(200, response.getStatus());
+    Assert.assertEquals(Collections.emptyList(), response.getEntity());
+
+    // Test with negative limit
+    response = supervisorResource.specGetHistory(request, "id1", -1);
```
cryptoe, line 1189: Should count <=0 be a bad request ie 400 ?

cryptoe:
Thanks for taking this up @vogievetsky !!

indexing-service/src/test/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResourceTest.java
vogievetsky: updated!

server/src/main/java/org/apache/druid/metadata/SQLMetadataSupervisorManager.java
```diff
@@ -138,19 +138,29 @@ public Map<String, List<VersionedSupervisorSpec>> getAll()
   }
 
   @Override
-  public List<VersionedSupervisorSpec> getAllForId(String id)
+  public List<VersionedSupervisorSpec> getAllForId(String id, @Nullable Integer limit) throws IllegalArgumentException
   {
+    if (limit != null && limit <= 0) {
+      throw new IllegalArgumentException("Limit must be greater than zero if set");
```
cryptoe, line 144: The end user is setting count but the exception message received by the end user is talking about limit. 
Since there can always be a variable name change, what we could do is to this check in the : https://github.com/apache/druid/blob/8f139bdf5fb5dead3fe02fe36f667faa0b81fb66/indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java#L534 that way we could control the end user message better. 

Nit: We might want to add the limit passed just to aid debugging.

cryptoe (commented):
Minor comment. 
Other than that lgtm.

vogievetsky:
Added validation to supervisor resource. Not sure where to put the limit in the debugging... there isn't an existing log statement

cryptoe:
Thanks for the PR. 🚀

PR #18417, opened 2025-08-19
Fix missing table column marker on multi-stage-query doc page

rasmusselsmark:
### Description

I noticed that under https://druid.apache.org/docs/latest/multi-stage-query/reference/#durable-storage-configurations, the markdown table is not rendered correctly:

image

Adding the missing column marker

This PR has:

- [x] been self-reviewed.


cryptoe (approved):

cryptoe:
Thanks for the PR @rasmusselsmark . Look forward to many more.

PR #18422, opened 2025-08-20
Consolidate ObjectMappers in MSQTestBase.

gianm:
This patch consolidates the two ObjectMappers in MSQTestBase into a single ObjectMapper. The single ObjectMapper is based on the modules from the test class's ComponentSupplier. In addition to simplification, this has the benefit of making it easier to create subclasses that provide additional Jackson modules.

### DIFF
--- a/extensions-core/multi-stage-query/src/test/java/org/apache/druid/msq/test/MSQTestBase.java
+++ b/extensions-core/multi-stage-query/src/test/java/org/apache/druid/msq/test/MSQTestBase.java
@@ -155,6 +155,8 @@ import org.apache.druid.segment.loading.LocalLoadSpec;
 import org.apache.druid.segment.loading.SegmentCacheManager;
 import org.apache.druid.segment.realtime.appenderator.AppenderatorsManager;
 import org.apache.druid.segment.writeout.OffHeapMemorySegmentWriteOutMediumFactory;
+import org.apache.druid.segment.writeout.SegmentWriteOutMediumFactory;
+import org.apache.druid.segment.writeout.TmpFileSegmentWriteOutMediumFactory;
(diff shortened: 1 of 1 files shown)


extensions-core/multi-stage-query/src/test/java/org/apache/druid/msq/test/MSQTestBase.java
```diff
@@ -429,13 +456,10 @@ public void setUp2() throws Exception
...
-    ObjectMapper secondMapper = setupObjectMapper(qf.injector());
-    indexIO = new IndexIO(secondMapper, ColumnConfig.DEFAULT);
-
-    segmentCacheManager = new SegmentCacheManagerFactory(TestIndex.INDEX_IO, secondMapper).manufacturate(newTempFolder("cacheManager"));
-
-    MSQSqlModule sqlModule = new MSQSqlModule();
+    objectMapper = setupObjectMapper(qf.injector());
+    indexIO = new IndexIO(objectMapper, ColumnConfig.DEFAULT);
 
+    segmentCacheManager = new SegmentCacheManagerFactory(TestIndex.INDEX_IO, objectMapper).manufacturate(newTempFolder("cacheManager"));
```
kfaraz: Nit: Could we use the above `indexIO` here instead of `TestIndex.INDEX_IO`?

kfaraz (approved):

extensions-core/multi-stage-query/src/test/java/org/apache/druid/msq/test/MSQTestBase.java
gianm: Yes, just pushed that.

PR #18424, opened 2025-08-21
Jetty12 (Jakarta EE8) upgrade

capistrant:
#### Upgrade to Jetty 12

Upgrades our jetty dependency to jetty 12.0.x (12.0.25 at time of writing). Uses Jakarta EE8 compatible jetty 12 to enable a less intrusive upgrade.

##### Avatica Changes

Calcite Avatica has a hard dependency on jetty9 still and there is no clear timeline on their upgrade plans. To avoid this being a blocker we have reduced our dependency footprint on Avatica in this PR by bringing the Jetty handler implementations for Druid JDBC while retaining compatibility with the Avatica JDBC client.

##### Hadoop Compatibility

Hadoop ecosystem also comes with lots of dependencies on jetty9 and java11 support. This section lays out implications for hadoop support

###### Deep storage

Confirmed functionality using Docker IT env

###### index_hadoop

Confirmed functionality using Docker IT env. We will want more expansive testing here. I tried to slim down the code that sets up the injector, but since the task needs to load all of the extensions, I ended up falling back to an ugly duplication of the service injector creation, just without jetty. I'm sure it can be improved if we want to.

##### Java 11 Support Implications

Jetty12 deps are in java 17 class file version, so running druid services on java11 seems out of the question. If this is the case, java 11 support will have to be dropped in 35.

### Future Work

* Explore jetty 12 features and integrate them where appropriate

#### Release note

Upgrading Druid to Jetty 12 requires dropping Java 11 runtime support for Druid. The one exception to this is that Hadoop Ingestion can continue to run on Java 11 YARN clusters because the Jetty 12 dependency can be excluded from that runtime classpath safely.

A new server configuration option has been added with this PR: `druid.server.http.uriCompliance`. Jetty 12 by default has strict enforcement of `RFC3986` URI format. This is a change from Jetty 9. To retain compatibility with legacy Druid, we default this config to `LEGACY`, which uses the more permissive URI format enforcement that jetty9 used. If the cluster you operate does not require LEGACY compatibility, it is recommended to use the upstream jetty default of `RFC3986` in your Druid deployment. Please see [jetty documentation](https://jetty.org/docs/jetty/12.1/programming-guide/server/compliance.html#uri) for more info.

#### Upgrade Note

Java 11 runtime no longer supported. Java 17 is the only officially supported and recommended runtime environment for Druid 35

<hr>

PR is mostly a lot of tiny changes and tweaks. But some places to get close look could be:
* JDBC related files
    * sql/src/main/java/org/apache/druid/sql/avatica/DruidAvaticaJsonHandler.java
    * sql/src/main/java/org/apache/druid/sql/avatica/DruidAvaticaProtobufHandler.java
* Hadoop Indexing related files
    * indexing-hadoop/src/main/java/org/apache/druid/indexer/HadoopDruidIndexerConfig.java
    * indexing-hadoop/src/main/java/org/apache/druid/indexer/JobHelper.java
    * indexing-hadoop/src/main/java/org/apache/druid/indexer/Utils.java
    * server/src/main/java/org/apache/druid/initialization/CoreInjectorBuilder.java
    * server/src/main/java/org/apache/druid/initialization/Initialization.java
    * server/src/main/java/org/apache/druid/initialization/ServerInjectorBuilder.java
* server/src/main/java/org/apache/druid/guice/http/JettyHttpClientModule.java
* server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerInitUtils.java
* server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerModule.java

<hr>

This PR has:

- [ ] been self-reviewed.


sql/src/main/java/org/apache/druid/sql/avatica/DruidAvaticaJsonHandler.java
```diff
@@ -19,46 +19,127 @@
...
-import javax.servlet.http.HttpServletResponse;
-import java.io.IOException;
+import javax.inject.Inject;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.Objects;
 
-public class DruidAvaticaJsonHandler extends AvaticaJsonHandler
+public class DruidAvaticaJsonHandler extends Handler.Abstract implements MetricsAwareAvaticaHandler
```
kfaraz: I think it would be nice to add a `DruidAvaticaHandler` at this point which can be extended by both the Json handler and the Protobuf handler. It was not possible earlier since both implemented different interfaces.

We can put the common logic (if any) in the base `DruidAvaticaHandler` class.
Even if there is no common logic, it would still be more manageable in the long term.

sql/src/test/java/org/apache/druid/quidem/DruidAvaticaTestDriver.java
```diff
@@ -186,7 +186,7 @@ public void close()
       }
     }
 
-    protected AbstractAvaticaHandler getAvaticaHandler(final DruidMeta druidMeta)
+    protected Handler.Abstract getAvaticaHandler(final DruidMeta druidMeta)
```
kfaraz: This could return the `DruidAvaticaHandler` suggested in the other comment.

docs/configuration/index.md
```diff
@@ -1496,6 +1496,7 @@ Druid uses Jetty to serve HTTP requests.
 |`druid.server.http.enableForwardedRequestCustomizer`|If enabled, adds Jetty ForwardedRequestCustomizer which reads X-Forwarded-* request headers to manipulate servlet request object when Druid is used behind a proxy.|false|
 |`druid.server.http.allowedHttpMethods`|List of HTTP methods that should be allowed in addition to the ones required by Druid APIs. Druid APIs require GET, PUT, POST, and DELETE, which are always allowed. This option is not useful unless you have installed an extension that needs these additional HTTP methods or that adds functionality related to CORS. None of Druid's bundled extensions require these methods.|`[]`|
 |`druid.server.http.contentSecurityPolicy`|Content-Security-Policy header value to set on each non-POST response. Setting this property to an empty string, or omitting it, both result in the default `frame-ancestors: none` being set.|`frame-ancestors 'none'`|
+|`druid.server.http.uriCompliance`|Jetty `UriCompliance` mode for Druid's embedded Jetty servers. To modify, override this config with the string representation of any `UriCompliance` mode that Jetty supports.|LEGACY|
```
kfaraz: Can we also list out the possible values of this config?
Or link to the Jetty page which lists the values.

docs/operations/java.md
```diff
@@ -27,7 +27,7 @@ a Java runtime for Druid.
 
 ## Selecting a Java runtime
 
- The project team recommends Java 17. Although you can use Java 11, support for it is deprecated.
+Druid officially support Java 17.
```
kfaraz:
```suggestion
Druid officially supports Java 17.
```

server/src/main/java/org/apache/druid/server/initialization/jetty/StandardResponseHeaderFilterHolder.java
```diff
@@ -66,16 +66,17 @@ public StandardResponseHeaderFilterHolder(final ServerConfig serverConfig)
...
       final HttpServletResponse proxyResponse,
       final Response serverResponse
   )
   {
     for (final String headerName : StandardResponseHeaderFilterHolder.STANDARD_HEADERS) {
-      if (serverResponse.getHeaders().containsKey(headerName) && proxyResponse.containsHeader(headerName)) {
-        ((org.eclipse.jetty.server.Response) proxyResponse).getHttpFields().remove(headerName);
+      if (serverResponse.getHeaders().contains(headerName) && proxyResponse.containsHeader(headerName)) {
+        // In Jetty 12 EE8, use the standard servlet API method to remove headers
+        proxyResponse.setHeader(headerName, null);
```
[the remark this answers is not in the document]
capistrant: I guess I need to dig deeper into this. I don't think we can do the cast to a jetty response object like we used to. And I don't want to reach into the `ee8` internals either, even if something in there would provide an api for removing a header.

sql/src/main/java/org/apache/druid/sql/avatica/DruidAvaticaJsonHandler.java
```diff
@@ -19,46 +19,127 @@
...
+            jsonResponse = jsonHandler.convertToErrorResponse(e);
+          }
+
+          LOG.trace("response: %s", jsonResponse);
+          response.setStatus(jsonResponse.getStatusCode());
+          response.write(true, ByteBuffer.wrap(jsonResponse.getResponse().getBytes(StandardCharsets.UTF_8)), callback);
+        }
+
+        callback.succeeded();
+        return true;
```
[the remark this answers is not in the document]
capistrant: latest commits should try to address this along with general lack of consistency in request handling between json and protobuf handlers.

indexing-hadoop/src/main/java/org/apache/druid/indexer/JobHelper.java
```diff
@@ -155,7 +155,8 @@ public static void setupClasspath(
     for (String jarFilePath : jarFiles) {
 
       final File jarFile = new File(jarFilePath);
-      if (jarFile.getName().endsWith(".jar")) {
+      // Keep Druid jetty jars out of the classpath. They are not runnable in the < 17 hadoop java runtime
+      if (jarFile.getName().endsWith(".jar") && !jarFile.getName().contains("jetty")) {
```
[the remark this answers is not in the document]
capistrant: I agree that this is suboptimal. I was in the "lets see if we can make it work" mindset. Maybe there is a more clean approach to filtering the jetty jars out.

kfaraz:
Thanks for the PR, @capistrant ! I plan to finish a review today.

server/src/main/java/org/apache/druid/initialization/CoreInjectorBuilder.java
```diff
@@ -137,4 +137,52 @@ public CoreInjectorBuilder forServer()
...
+        new AuthenticatorModule(),
+        new AuthenticatorMapperModule(),
+        new EscalatorModule(),
+        new AuthorizerModule(),
+        new AuthorizerMapperModule(),
+        new StartupLoggingModule(),
+        new ExternalStorageAccessSecurityModule(),
+        new ServiceClientModule(),
+        new StorageConnectorModule(),
+        new CatalogCoreModule()
```
kfaraz, line 145: Rather than duplicating the whole list, let's just call the new method from `forServer` and have an extra `add(new JettyServerModule())` in there.

server/src/main/java/org/apache/druid/initialization/CoreInjectorBuilder.java
```diff
@@ -137,4 +137,52 @@ public CoreInjectorBuilder forServer()
     );
     return this;
   }
+
+  /**
+   * Rough bridge solution for Hadoop indexing that needs server-like Injector but can't run jetty 12
+   */
+  @Deprecated
```
kfaraz: I don't think this should be marked deprecated. It just provides a subset of all the modules.

server/src/main/java/org/apache/druid/initialization/Initialization.java
```diff
@@ -62,4 +62,16 @@ public static Injector makeInjectorWithModules(
   {
     return ServerInjectorBuilder.makeServerInjector(baseInjector, ImmutableSet.of(), modules);
   }
+
+  /**
+   * Rough bridge solution for Hadoop indexing that needs server-like Injector but can't run jetty 12
+   */
+  @Deprecated
+  public static Injector makeInjectorWithoutJettyModules(
```
kfaraz: The javadoc of this class says that we should use `StartupInjectorBuilder` instead.
So, rather than adding a new deprecated method, should we try to use `StartupInjectorBuilder` directly?

server/src/main/java/org/apache/druid/initialization/ServerInjectorBuilder.java
```diff
@@ -73,6 +73,23 @@ public static Injector makeServerInjector(
         .build();
   }
 
+  /**
+   * Rough bridge solution for Hadoop indexing that needs server-like Injector but can't run jetty 12
```
kfaraz: Instead of saying `Rough bridge solution all ...`, can we just say `Needed for ...` in all the places.

server/src/main/java/org/apache/druid/initialization/ServerInjectorBuilder.java
```diff
@@ -118,6 +135,37 @@ public Injector build()
     return new ExtensionInjectorBuilder(serviceBuilder).build();
   }
 
+  /**
+   * Rough bridge solution for Hadoop indexing that needs server-like Injector but can't run jetty 12
+   */
+  public Injector buildWithoutJettyModules()
```
kfaraz, line 116: Let's try to common out the parts of this method with the existing `build` method.

kfaraz (commented):
Done a partial review. Going through the remaining files.

server/src/main/java/org/apache/druid/initialization/CoreInjectorBuilder.java
capistrant: Sure, I will remove. I guess I was trying to discourage anything else from using this instead of just being compatible with jetty. Although, no such thing currently exists.

integration-tests/stop_cluster.sh
```diff
@@ -14,6 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Determine if docker compose is available. If not, assume Docker supports
```
kfaraz: Does this change need to be included in this PR? Or can it be done separately too?

server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerInitUtils.java
```diff
@@ -109,15 +109,6 @@ public static void addFilters(ServletContextHandler handler, Set<? extends Servl
...
   }
 
-  public static Handler getJettyRequestLogHandler()
-  {
-    // Ref: http://www.eclipse.org/jetty/documentation/9.2.6.v20141205/configuring-jetty-request-logs.html
-    RequestLogHandler requestLogHandler = new RequestLogHandler();
-    requestLogHandler.setRequestLog(new JettyRequestLog());
-
-    return requestLogHandler;
-  }
```
clintropolis, line 119: it looks like this is replaced with `Server.setRequestLog` in jetty 12, https://jetty.org/docs/jetty/12.1/programming-guide/migration/11-to-12.html#api-changes-handler-requestlog, do we need to add that somewhere like the server initializers class or something?

integration-tests/script/docker_run_cluster.sh
```diff
@@ -14,6 +14,15 @@
...
 # limitations under the License.
 
+# Determine if docker compose is available. If not, assume Docker supports
+# the compose subcommand
+set +e
+if which docker-compose > /dev/null
+then
+  DOCKER_COMPOSE='docker-compose'
+else
+  DOCKER_COMPOSE='docker compose'
```
clintropolis: shouldn't we prefer `docker compose`? i think `docker-compose` is pretty old now and hasn't been included in docker distribution for a while now

docs/operations/java.md
```diff
@@ -43,26 +43,18 @@ Druid relies on the environment variables `JAVA_HOME` or `DRUID_JAVA_HOME` to fi
...
 uses certain internal JDK APIs, which must be added to `--add-exports` and `--add-opens` on the Java command line.
 
-On Java 11, if these parameters are not included, you will see warnings like the following:
-
-```
-WARNING: An illegal reflective access operation has occurred
-WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
-WARNING: All illegal access operations will be denied in a future release
-```
-
```
clintropolis, line 65: I think you still see these if you don't add the stuff, are we just assuming that people are used to it by now?

capistrant: ehh that was probably me just seeing Java 11 and not Java 11+ and hitting delete. I'll confirm it is still there and add back without the explicit java 11 part

integration-tests/script/docker_run_cluster.sh
capistrant: heh, ya. I probably just need to fix my local setup and not use docker-compose. If nothing else I'll flip it to only fall back to docker-compose.

docs/operations/java.md
capistrant: actually maybe removing them is right if java 17 is actually throwing exceptions without them like we indicate on line 66

server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerInitUtils.java
capistrant: 💯 definitely a miss by not properly migrating this. I think initialization in JettyServerModule is the right spot? Confirmed that it was wired up across all druid services in my Docker env after moving it in latest commits.

server/src/main/java/org/apache/druid/server/initialization/jetty/StandardResponseHeaderFilterHolder.java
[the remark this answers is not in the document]
capistrant: I did some testing here and found that this approach is indeed removing the duplicate headers from proxied requests.. I tested locally with and without this set to null and without it, the headers in the constant are duplicated. I think this is a bit of a gray area in the servlets api, which has the copilot bot on high alert.

capistrant:
@kfaraz Thank you for the initial review! I made some modifications to the initialization code and avatica handler per your review. Let me know if this is going down the lines of what you were thinking.

pom.xml
```diff
@@ -754,26 +754,44 @@
...
+                    </exclusion>
+                </exclusions>
             </dependency>
             <dependency>
-                <groupId>org.eclipse.jetty</groupId>
-                <artifactId>jetty-servlet</artifactId>
+                <groupId>org.eclipse.jetty.ee8</groupId>
+                <artifactId>jetty-ee8-servlet</artifactId>
                 <version>${jetty.version}</version>
+              <exclusions>
```
kfaraz: Nit: indentation seems off here.

indexing-hadoop/src/main/java/org/apache/druid/indexer/JobHelper.java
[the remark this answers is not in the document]
kfaraz: Could we use `jetty-ee8-*` instead?

server/src/main/java/org/apache/druid/initialization/ServerInjectorBuilder.java
```diff
@@ -73,6 +73,22 @@ public static Injector makeServerInjector(
         .build();
   }
 
+  /**
+   * Needed for Hadoop indexing that needs server-like Injector but can't run jetty 12
+   */
+  @VisibleForTesting
```
kfaraz: I think this method is used by `HadoopDruidIndexerConfig` too. If that is the case, let's remove this annotation.

server/src/main/java/org/apache/druid/initialization/ServerInjectorBuilder.java
```diff
@@ -91,6 +107,33 @@ public ServerInjectorBuilder serviceModules(final Iterable<? extends Module> mod
...
+
+  /**
+   * Needed for Hadoop indexing that needs server-like Injector but can't run jetty 12
+   */
+  public Injector buildWithoutJettyModules()
+  {
+    return this.build(false);
+  }
+
+  public static Module registerNodeRoleModule(Set<NodeRole> nodeRoles)
```
kfaraz: It seems that this method is unchanged. Can we move it back to its original position so that we can avoid it from showing up in the diff?

server/src/main/java/org/apache/druid/server/initialization/ServerConfig.java
```diff
@@ -306,6 +312,11 @@ public boolean isEnableQueryRequestsQueuing()
     return enableQueryRequestsQueuing;
   }
 
+  public String getUriCompliance()
```
kfaraz: This should probably return a `UriCompliance` object rather than a String.

server/src/main/java/org/apache/druid/server/initialization/ServerConfig.java
```diff
@@ -185,6 +188,9 @@ public ServerConfig(boolean enableQueryRequestsQueuing)
   @JsonProperty
   private boolean enableHSTS = false;
 
+  @JsonProperty
+  private String uriCompliance = UriCompliance.LEGACY.getName();
```
kfaraz: Would it make sense to have this be of type `UriCompliance` rather than `String`?

server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerModule.java
```diff
@@ -467,25 +468,21 @@ public void stop()
...
         {
-          request.setAttribute(RequestDispatcher.ERROR_EXCEPTION, null);
-          super.handle(target, baseRequest, request, response);
+          baseRequest.setAttribute(RequestDispatcher.ERROR_EXCEPTION, null);
+          return super.handle(baseRequest, response, callback);
         }
       });
     }
 
+    log.info("Configuraing Jetty Request log for server");
```
kfaraz: Is this needed?

server/src/main/java/org/apache/druid/server/initialization/jetty/StandardResponseHeaderFilterHolder.java
```diff
@@ -66,16 +66,17 @@ public StandardResponseHeaderFilterHolder(final ServerConfig serverConfig)
...
   public static void deduplicateHeadersInProxyServlet(
       final HttpServletResponse proxyResponse,
       final Response serverResponse
   )
   {
     for (final String headerName : StandardResponseHeaderFilterHolder.STANDARD_HEADERS) {
-      if (serverResponse.getHeaders().containsKey(headerName) && proxyResponse.containsHeader(headerName)) {
-        ((org.eclipse.jetty.server.Response) proxyResponse).getHttpFields().remove(headerName);
+      if (serverResponse.getHeaders().contains(headerName) && proxyResponse.containsHeader(headerName)) {
+        // In Jetty 12 EE8, use the standard servlet API method to remove headers
```
kfaraz: By "standard servlet API method", are you referring to `proxyResponse.setHeader()` or some other method that we would like to use but are not using here?

server/src/test/java/org/apache/druid/initialization/ServerConfigTest.java
```diff
@@ -79,6 +81,7 @@ public void testSerde() throws Exception
     Assert.assertEquals("my-cool-policy", modifiedConfig.getContentSecurityPolicy());
     Assert.assertEquals("my-cool-policy", modifiedConfig2.getContentSecurityPolicy());
     Assert.assertTrue(modifiedConfig2.isEnableHSTS());
+    Assert.assertEquals(UriCompliance.RFC3986.getName(), modifiedConfig2.getUriCompliance());
```
kfaraz: Please update one of the tests to verify the default value of this config too.

server/src/test/java/org/apache/druid/server/AsyncManagementForwardingServletTest.java
```diff
@@ -534,12 +532,11 @@ public String getCurrentLeader()
...
 
-      final HandlerList handlerList = new HandlerList();
-      handlerList.setHandlers(
-          new Handler[]{JettyServerInitUtils.wrapWithDefaultGzipHandler(
+      final Handler.Sequence handlerList = new Handler.Sequence(
+          JettyServerInitUtils.wrapWithDefaultGzipHandler(
               root,
               ServerConfig.DEFAULT_GZIP_INFLATE_BUFFER_SIZE,
-              Deflater.DEFAULT_COMPRESSION)}
+              Deflater.DEFAULT_COMPRESSION)
```
kfaraz:
```suggestion
              Deflater.DEFAULT_COMPRESSION
          )
```

server/src/test/java/org/apache/druid/server/initialization/jetty/StandardResponseHeaderFilterHolderTest.java
```diff
@@ -130,6 +139,40 @@ public void test_get_invalidContentSecurityPolicy()
...
+  public void test_deduplicateHeadersInProxyServlet_duplicatesExist()
+  {
+    EasyMock.expect(proxyResponse.containsHeader("Cache-Control")).andReturn(true).once();
+    proxyResponse.setHeader("Cache-Control", null);
+    EasyMock.expectLastCall().once();
+    EasyMock.expect(proxyResponse.containsHeader("Strict-Transport-Security")).andReturn(false).once();
+
+    EasyMock.expect(clientResponse.getHeaders()).andReturn(HttpFields.from(new HttpField("Cache-Control", "true"), new HttpField("Strict-Transport-Security", "true"))).times(3);
+
+
```
kfaraz, line 154: Nit: extra newline

server/src/test/java/org/apache/druid/server/initialization/jetty/StandardResponseHeaderFilterHolderTest.java
```diff
@@ -130,6 +139,40 @@ public void test_get_invalidContentSecurityPolicy()
...
 
+  @Test
+  public void test_deduplicateHeadersInProxyServlet_duplicatesExist()
+  {
+    EasyMock.expect(proxyResponse.containsHeader("Cache-Control")).andReturn(true).once();
+    proxyResponse.setHeader("Cache-Control", null);
+    EasyMock.expectLastCall().once();
+    EasyMock.expect(proxyResponse.containsHeader("Strict-Transport-Security")).andReturn(false).once();
+
+    EasyMock.expect(clientResponse.getHeaders()).andReturn(HttpFields.from(new HttpField("Cache-Control", "true"), new HttpField("Strict-Transport-Security", "true"))).times(3);
```
kfaraz: Please break this up into multiple lines in any way you see fit.

server/src/test/java/org/apache/druid/server/initialization/jetty/StandardResponseHeaderFilterHolderTest.java
```diff
@@ -130,6 +139,40 @@ public void test_get_invalidContentSecurityPolicy()
...
+
+
+    replayAllMocks();
+
+    StandardResponseHeaderFilterHolder.deduplicateHeadersInProxyServlet(proxyResponse, clientResponse);
+  }
+
+  @Test
+  public void test_duplicateHeadersInProxyServlet_noDuplicates()
```
kfaraz:
```suggestion
  public void test_deduplicateHeadersInProxyServlet_withNoDuplicates()
```

server/src/test/java/org/apache/druid/server/initialization/jetty/StandardResponseHeaderFilterHolderTest.java
```diff
@@ -130,6 +139,40 @@ public void test_get_invalidContentSecurityPolicy()
     );
   }
 
+  @Test
+  public void test_deduplicateHeadersInProxyServlet_duplicatesExist()
```
kfaraz:
```suggestion
  public void test_deduplicateHeadersInProxyServlet_withDuplicates()
```

server/src/test/java/org/apache/druid/server/mocks/MockHttpServletResponse.java
```diff
@@ -145,9 +149,13 @@ public void addDateHeader(String name, long date)
     throw new UnsupportedOperationException();
   }
 
+  /**
+   * HttpServletResponse 4.0.1 spec dictates that setHeader  overwrites existing values.
```
kfaraz: Nit: Since this must already be specified in the super javadoc, we can just make this a regular (non-javadoc) comment inside the method body.

server/src/test/java/org/apache/druid/server/mocks/MockHttpServletResponse.java
```diff
@@ -337,4 +345,17 @@ public Locale getLocale()
   {
     throw new UnsupportedOperationException();
   }
+
+
```
kfaraz: Nit: extra newline.

sql/src/main/java/org/apache/druid/sql/avatica/DruidAvaticaHandler.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.calcite.avatica.metrics.MetricsSystem;
+import org.apache.calcite.avatica.metrics.Timer;
+import org.apache.calcite.avatica.remote.LocalService;
+import org.apache.calcite.avatica.remote.MetricsHelper;
+import org.apache.calcite.avatica.remote.Service;
+import org.apache.calcite.avatica.server.MetricsAwareAvaticaHandler;
+import org.apache.calcite.avatica.util.UnsynchronizedBuffer;
+import org.eclipse.jetty.server.Handler;
+
+public abstract class DruidAvaticaHandler extends Handler.Abstract implements MetricsAwareAvaticaHandler
```
kfaraz, line 45: Please add a short javadoc, if possible.

sql/src/main/java/org/apache/druid/sql/avatica/DruidAvaticaProtobufHandler.java
```diff
@@ -19,46 +19,100 @@
...
-public class DruidAvaticaProtobufHandler extends AvaticaProtobufHandler
+public class DruidAvaticaProtobufHandler extends DruidAvaticaHandler
 {
+
+  private static final Logger LOG = new Logger(DruidAvaticaProtobufHandler.class);
+
   public static final String AVATICA_PATH_NO_TRAILING_SLASH = "/druid/v2/sql/avatica-protobuf";
   public static final String AVATICA_PATH = AVATICA_PATH_NO_TRAILING_SLASH + "/";
 
+  private final ProtobufHandler pbHandler;
```
kfaraz:
```suggestion
  private final ProtobufHandler protobufHandler;
```

sql/src/main/java/org/apache/druid/sql/avatica/DruidAvaticaProtobufHandler.java
```diff
@@ -19,46 +19,100 @@
...
+          handlerResponse = pbHandler.apply(requestBytes);
+        }
+        catch (Exception e) {
+          LOG.debug(e, "Error invoking request");
+          handlerResponse = pbHandler.convertToErrorResponse(e);
+        }
+
+        response.setStatus(handlerResponse.getStatusCode());
+        response.write(true, ByteBuffer.wrap(handlerResponse.getResponse()), callback);
+        return true;
```
kfaraz, line 104: I wonder if parts of this logic can also live in the super class `DruidAvaticaHandler`. Not a blocker for this PR though.

sql/src/test/java/org/apache/druid/sql/avatica/DruidAvaticaHandlerTest.java
```diff
@@ -255,7 +255,7 @@ protected String getJdbcUrlTail()
   }
 
   // Default implementation is for JSON to allow debugging of tests.
-  protected AbstractAvaticaHandler getAvaticaHandler(final DruidMeta druidMeta)
+  protected Handler.Abstract getAvaticaHandler(final DruidMeta druidMeta)
```
kfaraz:
```suggestion
  protected DruidAvaticaHandler getAvaticaHandler(final DruidMeta druidMeta)
```

sql/src/test/java/org/apache/druid/sql/avatica/DruidAvaticaProtobufHandlerTest.java
```diff
@@ -35,12 +44,49 @@ protected String getJdbcUrlTail()
   }
 
   @Override
-  protected AbstractAvaticaHandler getAvaticaHandler(final DruidMeta druidMeta)
+  protected Handler.Abstract getAvaticaHandler(final DruidMeta druidMeta)
```
kfaraz:
```suggestion
  protected DruidAvaticaHandler getAvaticaHandler(final DruidMeta druidMeta)
```

kfaraz (approved):
Thanks for addressing the feedback, @capistrant !
I have left minor suggestions/queries in some places.
But none of those should block this PR.

integration-tests/stop_cluster.sh
capistrant: I ended up removing it and will not PR it separately. My machine just had a bad setup for docker, I assume most people use `docker compose` > `docker-compose` these days so supporting legacy is not needed. Especially since we are phasing out legacy ITs

server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerModule.java
capistrant: likely not! Was a convenience for me to grep across service logs during some functional testing.

server/src/main/java/org/apache/druid/server/initialization/jetty/StandardResponseHeaderFilterHolder.java
capistrant: eh, I meant method as in "procedure of doing something".. Bad choice of words when writing a comment in Java code 😆 I will clarify

indexing-hadoop/src/main/java/org/apache/druid/indexer/JobHelper.java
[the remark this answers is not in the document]
capistrant: no, the way jetty is packaged now is that their core code that is javaEE agnostic is in `jetty-*.jar` and code specific to the EE that a project needs is in `jetty-eeX-*.jar`.

clintropolis (approved):

PR #18429, opened 2025-08-23
SqlTestFramework: Load datasets using injected ObjectMapper.

gianm:
Previously, datasets were loaded using TestHelper's default object mapper. Using the injected ObjectMapper ensures all the proper modules are loaded to be able to create test datasets.

Also fixes an omission where, previously, `transformSpec` was ignored when loading the test datasets.


gianm:
A bunch of `QTest` are failing. I took a look at a couple of them and I think they are being affected by the fix to `transformSpec`. That means the outcomes might be more correct now than they were before.

gianm:
I added the result of running the QTests with `-Dquidem.overwrite`. I didn't check them all, but the few I did check seem to be more correct now than they were in the past, owing to the fact that `transformSpec` is now being applied.

weishiuntsai (approved):
I looked at the Quidem test update.  That part LGTM.  `transformSpec` now is kicking in properly.

PR #18430, opened 2025-08-23
Handle local "files" in FakeIndexTaskUtil.

gianm:
The LocalInputSource can do "baseDir" + "filter", but can also have a list of "files". Previously, FakeIndexTaskUtil would throw NPE when encountering a LocalInputSource that used "files" instead of "baseDir" and "filter".

### DIFF
--- a/sql/src/test/java/org/apache/druid/sql/calcite/util/FakeIndexTaskUtil.java
+++ b/sql/src/test/java/org/apache/druid/sql/calcite/util/FakeIndexTaskUtil.java
@@ -84,6 +84,7 @@ public class FakeIndexTaskUtil
     );
   }
 
+  @Nullable
   private static File resolveFile(File projectRoot, @Nullable File file)
   {
     if (file == null || file.isAbsolute()) {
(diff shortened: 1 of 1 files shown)


sql/src/test/java/org/apache/druid/sql/calcite/util/FakeIndexTaskUtil.java
```diff
@@ -68,24 +70,29 @@ public static TestDataSet makeDS(ObjectMapper objectMapper, File src)
...
-        newBaseDir,
+        resolveFile(projectRoot, localInputSource.getBaseDir()),
         localInputSource.getFilter(),
-        localInputSource.getFiles(),
+        localInputSource.getFiles().stream().map(f -> resolveFile(projectRoot, f)).collect(Collectors.toList()),
         localInputSource.getSystemFields()
     );
   }
 
+  private static File resolveFile(File projectRoot, @Nullable File file)
```
abhishekrb19, line 88: nit: `@Nullable` annotation

PR #18433, opened 2025-08-23
Add "lines" input format.

gianm:
This is useful for reading entire lines for later processing. For example, it can be used to read newline-delimited JSON into a single nested column, by combining this input format with the `TRY_PARSE_JSON` function.


processing/src/main/java/org/apache/druid/data/input/impl/RegexInputFormat.java
```diff
@@ -96,10 +96,6 @@ public InputEntityReader createReader(InputRowSchema inputRowSchema, InputEntity
   @Override
   public long getWeightedSize(String path, long size)
   {
-    CompressionUtils.Format compressionFormat = CompressionUtils.Format.fromFileName(path);
-    if (CompressionUtils.Format.GZ == compressionFormat) {
-      return size * CompressionUtils.COMPRESSED_TEXT_WEIGHT_FACTOR;
-    }
-    return size;
+    return size * CompressionUtils.estimatedCompressionFactor(CompressionUtils.Format.fromFileName(path));
```
abhishekrb19, line 99: Perhaps we can extend `CompressionUtils.estimatedCompressionFactor()` or add a thin wrapper that accepts the `path` directly and returns the compression factor - internally invoking `CompressionUtils.Format.fromFileName()`

abhishekrb19 (approved):
👍 

It will be good to add web-console support for this feature separately.

processing/src/main/java/org/apache/druid/data/input/impl/RegexInputFormat.java
gianm: IMO these are equivalently useful for what we need them for right now, so I'll keep it the way I had it.

gianm: thanks for reviewing!

docs/ingestion/data-formats.md
```diff
@@ -168,6 +168,28 @@ For example:
...
+
+For example:
+
+```json
+"ioConfig": {
+  "inputFormat": {
+    "type": "lines"
+  },
+  ...
+}
```
zhangyue19921010: This change adds `Lines` to the markdown doc. This is good because users can understand that it will create a single column named `line` containing the entire line as a string and how to use `lines` in the `ioConfig` of this PR, reading each line as UTF-8 text.

PR #18436, opened 2025-08-25
Enable segment interval unlocks on segment handoffs

uds5501:
## Goal
Enable unlocking segment intervals during ingestion at segment handoffs during streaming ingestion if there is no other sink utilizing the same interval.

## Major classes changed.
- `SeekableStreamAppenderatorConfig.java` && `KafkaIndexTaskTuningConfig.java`
- `TaskLockCallbackImpl.java` and `TaskLockCallback.java`
- `StreamAppenderator.java`

### possible todos
- Passing the `shouldReleaseLockOnHandoff` to ingestion tasks from supervisors.
    - to discuss, do we even need these? web console based ingestion tasks should be good enough.


server/src/main/java/org/apache/druid/segment/realtime/appenderator/TaskLockCallback.java
```diff
@@ -0,0 +1,39 @@
...
+ * the segments are handed off.
+ */
+public interface TaskLockCallback
+{
+  /**
+   * Releases the lock for the given interval.
+   *
+   * @param interval interval for which the lock needs to be released
+   */
+  default void releaseLock(Interval interval)
```
kfaraz: This is a new functional interface, we shouldn't need to provide a default implementation for the method.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/TaskLockCallback.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.apache.druid.segment.realtime.appenderator;
+
+import org.joda.time.Interval;
+
+/**
+ * This interface provides a callback mechanism to interact with TaskLockbox for releasing interval locks when
+ * the segments are handed off.
+ */
+public interface TaskLockCallback
```
kfaraz: Maybe mention in the javadoc that this interface is needed only because `TaskLockbox` is in `indexing-service` module.
Maybe rename the interface `TaskIntervalUnlocker` since it is only going to be used for unlocking intervals.
Also, please annotate it `@FunctionalInterface`.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -1559,6 +1567,22 @@ public Void apply(@Nullable Object input)
...
 
+  private void unlockIntervalIfApplicable(Sink abandonedSink)
+  {
+    Interval abandonedInterval = abandonedSink.getInterval();
+    boolean isIntervalActive = sinks.entrySet().stream()
+                                    .anyMatch(entry -> {
+                                      Sink sink = entry.getValue();
+                                      return !Objects.equals(sink, abandonedSink)
+                                             && sink.isWritable()
+                                             && sink.getInterval().equals(abandonedInterval);
```
kfaraz: We should check for `overlaps`.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/TaskLockCallbackImpl.java
```diff
@@ -0,0 +1,44 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.indexing.common.actions;
+
+import org.apache.druid.segment.realtime.appenderator.TaskLockCallback;
+import org.joda.time.Interval;
+
+public class TaskLockCallbackImpl implements TaskLockCallback
```
kfaraz: We shouldn't need a separate class for this implementation. We should just inline this.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamAppenderatorConfig.java
```diff
@@ -57,11 +57,13 @@ public class SeekableStreamAppenderatorConfig implements AppenderatorConfig
 
   private final SeekableStreamIndexTaskTuningConfig tuningConfig;
   private final int maxColumnsToMerge;
+  private final boolean shouldReleaseLockOnHandoff;
```
kfaraz: We probably don't need an extra field or method here, just access it via the tuning config.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskTuningConfig.java
```diff
@@ -146,6 +149,9 @@ public SeekableStreamIndexTaskTuningConfig(
       this.numPersistThreads = Math.max(numPersistThreads, AppenderatorConfig.DEFAULT_NUM_PERSIST_THREADS);
     }
     this.maxColumnsToMerge = maxColumnsToMerge == null ? DEFAULT_MAX_COLUMNS_TO_MERGE : maxColumnsToMerge;
+    this.shouldReleaseLockOnHandoff = shouldReleaseLockOnHandoff == null
+                                      ? DEFAULT_SHOULD_RELEASE_LOCK_ON_HANDOFF
+                                      : shouldReleaseLockOnHandoff;
```
kfaraz: short hand:
```suggestion
    this.shouldReleaseLockOnHandoff = Configs.valueOrDefault(shouldReleaseLockOnHandoff, DEFAULT_SHOULD_RELEASE_LOCK_ON_HANDOFF);
```

kfaraz (commented):
Overall approach looks good, please rename the config to `releaseLocksOnHandoff` for brevity.

Along with unit tests, please add a new test method to `EmbeddedKafkaSupervisorTest` to verify that locks are released properly.

uds5501:
Thanks for the review @kfaraz , I have accommodated the review comments, currently added a hacky 5s delay to verify there are no locks in the end due to a race condition.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/AppenderatorConfig.java
```diff
@@ -38,6 +38,11 @@ public interface AppenderatorConfig extends TuningConfig
 
   boolean isSkipBytesInMemoryOverheadCheck();
 
+  default boolean getReleaseLocksOnHandoff()
```
kfaraz:
```suggestion
  default boolean isReleaseLocksOnHandoff()
```

server/src/main/java/org/apache/druid/segment/realtime/appenderator/AppenderatorConfig.java
```diff
@@ -77,4 +82,6 @@ default int getMaxColumnsToMerge()
   {
     return IndexMerger.UNLIMITED_MAX_COLUMNS_TO_MERGE;
   }
+
+
```
kfaraz: Nit: extra new lines

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/EmbeddedKafkaSupervisorTest.java
```diff
@@ -56,28 +59,33 @@ public class EmbeddedKafkaSupervisorTest extends EmbeddedClusterTestBase
...
   public EmbeddedDruidCluster createCluster()
   {
     final EmbeddedDruidCluster cluster = EmbeddedDruidCluster.withEmbeddedDerbyAndZookeeper();
+    indexer.addProperty("druid.segment.handoff.pollDuration", "PT0.1s");
+    overlord.addProperty("druid.manager.segments.pollDuration", "PT0.1s");
 
     kafkaServer = new KafkaResource();
 
     cluster.addExtension(KafkaIndexTaskModule.class)
+           .addCommonProperty("druid.monitoring.emissionPeriod", "PT0.1s")
```
kfaraz: Why is this needed?

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -374,6 +375,17 @@ public SupervisorStatus getSupervisorStatus(String supervisorId)
     throw new ISE("Could not find supervisor[%s]", supervisorId);
   }
 
+  /**
+   * Fetches the currently locked intervals by tasks.
+   * 
+   * @param lockFilterPolicies List of filters for different datasources.
+   * @return Map from datasource name to list of intervals locked by tasks.
+   */
+  public Map<String, List<Interval>> getLockedIntervals(List<LockFilterPolicy> lockFilterPolicies)
```
kfaraz: Nit: I don't think this warrants a new method. This can be inlined in the one place it is currently being used.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/EmbeddedKafkaSupervisorTest.java
```diff
@@ -119,6 +127,13 @@ public void test_runKafkaSupervisor()
     cluster.callApi().postSupervisor(kafkaSupervisorSpec.createSuspendedSpec());
     supervisorStatus = cluster.callApi().getSupervisorStatus(supervisorId);
     Assertions.assertTrue(supervisorStatus.isSuspended());
+    indexer.latchableEmitter().waitForEventAggregate(
+        event -> event.hasMetricName("ingest/handoff/count")
+                      .hasDimension(DruidMetrics.DATASOURCE, List.of(dataSource)),
+        agg -> agg.hasSumAtLeast(10)
+    );
+    Thread.sleep(5000);
```
kfaraz:
```suggestion
    overlord.latchableEmitter().waitForEventAggregate(
        event -> event.hasMetricName("task/action/run/time")
                      .hasDimension(DruidMetrics.DATASOURCE, dataSource)
                      .hasDimension(DruidMetrics.TASK_ACTION_TYPE, "lockRelease")
        agg -> agg.hasCountAtLeast(10)
    );
```

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -1559,6 +1567,22 @@ public Void apply(@Nullable Object input)
...
   }
 
+  private void unlockIntervalIfApplicable(Sink abandonedSink)
+  {
+    Interval abandonedInterval = abandonedSink.getInterval();
+    boolean isIntervalActive = sinks.entrySet().stream()
+                                    .anyMatch(entry -> {
+                                      Sink sink = entry.getValue();
+                                      return !Objects.equals(sink, abandonedSink)
+                                             && sink.isWritable()
```
kfaraz: What do these two conditions signify?
Please add a javadoc to this method briefly calling out the cases when the interval would be unlocked.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -1559,6 +1567,22 @@ public Void apply(@Nullable Object input)
...
+                                             && sink.getInterval().overlaps(abandonedInterval);
+                                    });
+    if (!isIntervalActive) {
+      taskIntervalUnlocker.releaseLock(abandonedInterval);
+    }
+    log.info("implement this.");
```
kfaraz:
```suggestion
    if (isIntervalActive) {
      log.info("Interval[%s] is still being appended to by sink[%s]", abandonedInterval, sinkId);
    } else {
       log.info("Unlocking interval[%s] as there are no more active sinks for it.", abandonedInterval);
       taskIntervalUnlocker.releaseLock(abandonedInterval);
    }
```

kfaraz: Also, given that this is a new feature, I am not sure if we should throw an exception if the unlock fails.
Can you please check if throwing an exception here will cause the task itself to fail or any other side effect?

Let's do the following instead:
- Add `throws IOException` to `TaskIntervalLocker.releaseLock()`
- Put the actual call to `taskIntervalUnlocker.releaseLock()` inside try/catch
- Emit an alert if unlock fails

In a follow up PR:
Let's also discuss the possibility of some clean up mechanism that gets rid of orphan locks.
Maybe the Overlord could have some kind of duty to perform automated clean up of orphan locks.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaIndexTaskTest.java
```diff
@@ -2911,7 +2911,8 @@ private KafkaIndexTask createTask(
         maxParseExceptions,
         maxSavedParseExceptions,
         null,
-        null
+        null,
+        false
```
kfaraz: Nit: Pass `null` instead so that we fallback to the default, even if it changes in the future.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/EmbeddedKafkaSupervisorTest.java
```diff
@@ -56,28 +59,33 @@ public class EmbeddedKafkaSupervisorTest extends EmbeddedClusterTestBase
...
   private final EmbeddedOverlord overlord = new EmbeddedOverlord();
+  private final EmbeddedHistorical historical = new EmbeddedHistorical();
   private KafkaResource kafkaServer;
 
   @Override
   public EmbeddedDruidCluster createCluster()
   {
     final EmbeddedDruidCluster cluster = EmbeddedDruidCluster.withEmbeddedDerbyAndZookeeper();
+    indexer.addProperty("druid.segment.handoff.pollDuration", "PT0.1s");
+    overlord.addProperty("druid.manager.segments.pollDuration", "PT0.1s");
```
kfaraz: Overlord does not seem to be polling segments in this test. We can remove this.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskTuningConfig.java
```diff
@@ -293,6 +298,12 @@ public int getMaxColumnsToMerge()
     return maxColumnsToMerge;
   }
 
+  @JsonProperty
+  public boolean getReleaseLockOnHandoff()
```
kfaraz: Nit: Primitive boolean getters are named `isXyz()` by convention.
Boxed `Boolean` getters may be named `getXyz()`.

```suggestion
  public boolean isReleaseLockOnHandoff()
```

server/src/main/java/org/apache/druid/segment/realtime/appenderator/AppenderatorConfig.java
kfaraz: Please add a short javadoc.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/TaskIntervalUnlocker.java
```diff
@@ -0,0 +1,38 @@
...
+ * the segments are handed off. We need this interface to avoid cyclic dependencues because the
+ * {@code TaskLockbox} is in druid-indexing-service module
+ */
+@FunctionalInterface
+public interface TaskIntervalUnlocker
+{
+  /**
+   * Releases the lock for the given interval.
+   *
+   * @param interval interval for which the lock needs to be released
```
kfaraz: We can omit this as it doesn't add any new info.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/TaskIntervalUnlocker.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * This interface provides a callback mechanism to interact with TaskLockbox for releasing interval locks when
+ * the segments are handed off. We need this interface to avoid cyclic dependencues because the
+ * {@code TaskLockbox} is in druid-indexing-service module
+ */
+@FunctionalInterface
+public interface TaskIntervalUnlocker
+{
+  /**
+   * Releases the lock for the given interval.
```
kfaraz: We should check the implementation on the Overlord side and call it out here on whether this releases locks for this exact interval or any lock held by this Task that _overlaps_ this interval.

server/src/test/java/org/apache/druid/segment/realtime/appenderator/BatchAppenderatorTester.java
```diff
@@ -174,7 +174,8 @@ public BatchAppenderatorTester(
         0L,
         OffHeapMemorySegmentWriteOutMediumFactory.instance(),
         IndexMerger.UNLIMITED_MAX_COLUMNS_TO_MERGE,
-        basePersistDirectory == null ? createNewBasePersistDirectory() : basePersistDirectory
+        basePersistDirectory == null ? createNewBasePersistDirectory() : basePersistDirectory,
+        false
```
kfaraz: Pass null for default value.

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2449,6 +2450,118 @@ public void testSchemaAnnouncement() throws Exception
     }
   }
 
+
+  @Test
+  public void test_abandonSegment_unlockIntervalWithOverlap() throws Exception
+  {
+    final List<Interval> unlockedIntervals = new ArrayList<>();
+    final TaskIntervalUnlocker mockUnlocker = interval -> {
```
kfaraz:
```suggestion
    final TaskIntervalUnlocker intervalUnlocker = interval -> {
```

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2449,6 +2450,118 @@ public void testSchemaAnnouncement() throws Exception
...
+      );
+
+      appenderator.add(identifier1, row1, Suppliers.ofInstance(Committers.nil()), false);
+      appenderator.add(identifier2, row2, Suppliers.ofInstance(Committers.nil()), false);
+
+      Assert.assertEquals(2, appenderator.getSegments().size());
+
+      synchronized (unlockedIntervals) {
+        unlockedIntervals.clear();
+      }
```
kfaraz: Isn't this a fresh list? Why do we need to clear it?

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2449,6 +2450,118 @@ public void testSchemaAnnouncement() throws Exception
...
+
+      appenderator.drop(identifier1).get();
+
+      synchronized (unlockedIntervals) {
```
kfaraz, line 2492: You can avoid the `synchronized` by using a thread-safe collection like some concurrent map or set.

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2449,6 +2450,118 @@ public void testSchemaAnnouncement() throws Exception
...
+        .basePersistDirectory(temporaryFolder.newFolder())
+        .maxRowsInMemory(2)
+        .releaseLocksOnHandoff(true)
+        .taskIntervalUnlocker(mockUnlocker)
+        .build()) {
+      final Appenderator appenderator = tester.getAppenderator();
+
+      appenderator.startJob();
+
+      final SegmentIdWithShardSpec identifier1 = si("2000-01-01T00:00/2000-01-01T01:00", "version1", 0);
```
kfaraz:
```suggestion
      final SegmentIdWithShardSpec segmentId1 = si("2000-01-01T00:00/2000-01-01T01:00", "version1", 0);
```

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2449,6 +2450,118 @@ public void testSchemaAnnouncement() throws Exception
...
+      final SegmentIdWithShardSpec identifier2 = si("2000-01-01T01:00/2000-01-01T02:00", "version1", 0);
```
kfaraz:
```suggestion
      final SegmentIdWithShardSpec segmentId2 = si("2000-01-01T01:00/2000-01-01T02:00", "version1", 0);
```

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2449,6 +2450,118 @@ public void testSchemaAnnouncement() throws Exception
     }
   }
 
+
+  @Test
+  public void test_abandonSegment_unlockIntervalWithOverlap() throws Exception
```
kfaraz:
```suggestion
  public void test_dropSegment_unlocksInterval() throws Exception
```

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2449,6 +2450,118 @@ public void testSchemaAnnouncement() throws Exception
...
+        Assert.assertEquals(identifier1.getInterval(), unlockedIntervals.get(0));
+      }
+
+      Assert.assertEquals(1, appenderator.getSegments().size());
+      Assert.assertTrue(appenderator.getSegments().contains(identifier2));
+    }
+  }
+
+  @Test
+  public void test_abandonSegment_shouldNotUnlockInterval() throws Exception
```
kfaraz:
```suggestion
  public void test_dropSegment_skipsUnlockInterval_ifOverlappingSinkIsActive() throws Exception
```

kfaraz (commented):
Left some more suggestions.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/EmbeddedKafkaSupervisorTest.java
uds5501: Not needed anymore, was testing race condition, now removed.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
uds5501: Okay, doing so

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/EmbeddedKafkaSupervisorTest.java
uds5501: fair, removing this.

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
uds5501: I did this for sanity in case the appenderator already dropped it (not really required because it's being called later explicitly)

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
uds5501: These 2 conditions mean:
- The other sink should not be the same sink we are abandoning during evaluation
- The other sink should not have finished writing, if it has finished writing, its interval should be available to be unlocked imo.

Added the same to java doc.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
uds5501: > Can you please check if throwing an exception here will cause the task itself to fail or any other side effect?

It should not because it's being called async from`StreamAppenderatorDiver` and the onSuccess / onFailure hooks are updating metric related records, so safe to do so.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
kfaraz: Thanks!

extensions-contrib/rabbit-stream-indexing-service/src/main/java/org/apache/druid/indexing/rabbitstream/RabbitStreamIndexTaskTuningConfig.java
```diff
@@ -100,7 +100,8 @@ public RabbitStreamIndexTaskTuningConfig(
         maxParseExceptions,
         maxSavedParseExceptions,
         numPersistThreads,
-        maxColumnsToMerge
+        maxColumnsToMerge,
+        false
```
kfaraz: Let's pass null here.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/EmbeddedKafkaSupervisorTest.java
kfaraz: Not removed yet.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/test/TestModifiedKafkaIndexTaskTuningConfig.java
```diff
@@ -81,7 +81,8 @@ public TestModifiedKafkaIndexTaskTuningConfig(
         maxParseExceptions,
         maxSavedParseExceptions,
         numPersistThreads,
-        maxColumnsToMerge
+        maxColumnsToMerge,
+        false
```
kfaraz: Pass null here.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskTuningConfig.java
```diff
@@ -41,6 +42,7 @@ public abstract class SeekableStreamIndexTaskTuningConfig implements TuningConfi
   private static final Period DEFAULT_INTERMEDIATE_PERSIST_PERIOD = new Period("PT10M");
   private static final IndexSpec DEFAULT_INDEX_SPEC = IndexSpec.DEFAULT;
   private static final Boolean DEFAULT_REPORT_PARSE_EXCEPTIONS = Boolean.FALSE;
+  private static final Boolean DEFAULT_SHOULD_RELEASE_LOCK_ON_HANDOFF = Boolean.FALSE;
```
kfaraz:
```suggestion
  private static final boolean DEFAULT_RELEASE_LOCKS_ON_HANDOFF = false;
```

server/src/main/java/org/apache/druid/segment/realtime/appenderator/AppenderatorConfig.java
```diff
@@ -38,6 +38,14 @@ public interface AppenderatorConfig extends TuningConfig
 
   boolean isSkipBytesInMemoryOverheadCheck();
 
+  /**
+   * Whether the interval locks be released when handing segments have been handed off to data nodes.
```
kfaraz:
```suggestion
   * @return true if locks should be released after segments have been handed off to Historicals.
```

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
kfaraz: I think we can remove these extra conditions for now.
We should not unlock the interval as long as there is any sink that overlaps the interval.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/TaskIntervalUnlocker.java
kfaraz: This still needs to be addressed.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/TaskIntervalUnlocker.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.druid.segment.realtime.appenderator;
+
+import org.joda.time.Interval;
+
+import java.io.IOException;
+
+/**
+ * This interface provides a callback mechanism to interact with TaskLockbox for releasing interval locks when
+ * the segments are handed off. We need this interface to avoid cyclic dependencues because the
+ * {@code TaskLockbox} is in druid-indexing-service module
```
kfaraz:
```suggestion
 * Used to release task locks after segments have been handed off, typically with long running tasks
 * to avoid holding locks for longer than necessary. This interface is used instead of {@code TaskActionClient}
 * to prevent a cyclic dependency with druid-indexing-service module.
```

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
kfaraz: Not addressed yet.

Tip: It generally helps to resolve comments after the relevant changes have been pushed. 🙂 , so that no updates are missed.

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2449,6 +2450,110 @@ public void testSchemaAnnouncement() throws Exception
...
+
+      appenderator.startJob();
+
+      final SegmentIdWithShardSpec segmentId1 = si("2000-01-01T00:00/2000-01-01T01:00", "version1", 0);
+      final SegmentIdWithShardSpec segmentId2 = si("2000-01-01T01:00/2000-01-01T02:00", "version1", 0);
+
+      final InputRow row1 = new MapBasedInputRow(
+          DateTimes.of("2000"),
+          ImmutableList.of("dim1"),
+          ImmutableMap.of("dim1", "bar", "met1", 1)
```
kfaraz: Prefer using `List.of` and `Map.of`.

kfaraz (commented):
Left some final suggestions, rest looks good.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/TaskIntervalUnlocker.java
uds5501: added.

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
uds5501: done.

kfaraz (approved):
+1. Can be merged once CI is green.

PR #18441, opened 2025-08-27
MSQ: Push down limits in ScanQueryFrameProcessor.

gianm:
When the scan query sorts by something that matches the underlying cursor, stop reading early.


multi-stage-query/src/main/java/org/apache/druid/msq/querykit/scan/ScanQueryFrameProcessor.java
```diff
@@ -497,4 +550,19 @@ private ColumnSelectorFactory wrapColumnSelectorFactoryIfNeeded(final ColumnSele
...
+  /**
+   * Returns the {@link ScanQuery#getScanRowsOffset()} plus {@link ScanQuery#getScanRowsLimit()}, or
+   * {@link Long#MAX_VALUE} if that addition would overflow.
+   */
+  private long getQueryOffsetPlusLimit()
+  {
+    final long sum = query.getScanRowsOffset() + query.getScanRowsLimit();
+    if (sum < 0) {
+      // Overflow
+      return Long.MAX_VALUE;
```
clintropolis, line 563: should this be a failure?

multi-stage-query/src/main/java/org/apache/druid/msq/querykit/scan/ScanQueryFrameProcessor.java
```diff
@@ -479,6 +522,16 @@ private long setNextCursor(
     this.cursorCloser = cursorCloser;
     this.segment = segment;
     this.cursorOffset.reset();
+    this.cursorRowsRead = 0;
+
+    if (query.isLimited()
+        && ordering.size() >= query.getOrderBys().size()
+        && query.getOrderBys().equals(ordering.subList(0, query.getOrderBys().size()))) {
```
clintropolis, line 529: in the future it might be nice to have a 'same but reversed direction' check so that we could make reverse cursors match too

gianm: Wouldn't the holder for a reverse cursor report reverse order in its `getOrdering()`? In that case this way of checking would be OK, I think.

multi-stage-query/src/main/java/org/apache/druid/msq/querykit/scan/ScanQueryFrameProcessor.java
gianm: I was thinking no, because  `query.getScanRowsLimit()` is set to `Long.MAX_VALUE` to indicate "unlimited", and if there is an offset then the addition will overflow. In this case setting the result to `Long.MAX_VALUE` is the right thing to do because it still means "unlimited".

multi-stage-query/src/main/java/org/apache/druid/msq/querykit/scan/ScanQueryFrameProcessor.java
clintropolis: oh right, i forgot how this worked :+1:

clintropolis: well, i suppose its not very sophisticated right now in terms of how it reports ordering when i look at the code for queryable index holder, 
```
    final List<OrderBy> indexOrdering = index.getOrdering();
    if (Cursors.preferDescendingTimeOrdering(cursorBuildSpec)
        && Cursors.getTimeOrdering(indexOrdering) == Order.ASCENDING) {
      this.ordering = Cursors.descendingTimeOrder();
    } else {
      this.ordering = indexOrdering;
    }
```
but probably good enough for now

PR #18444, opened 2025-08-27
Create Kubernetes peon lifecycle task log persist timeout

capistrant:
### Description

Prevents `saveLogs()` from hanging indefinitely when there are fabric8 issues. Read https://github.com/fabric8io/kubernetes-client/issues/7163 for some extra context on how LogWatch processing can be indefinitely blocking situation. Situations like this are suspects in issues around overlord graceful shutdowns being able to complete successfully. The upstream kubernetes-client fix associated with the linked issue is preferable to my implementation, but with that being unreleased, it is not clear when we can integrate the fix into Druid.

#### Release note

Introduce task log saving timeouts for the `kubernetes-overlord-extensions` `mm-less` ingestion framework. Persisting task logs will no longer have the potential to block indefinitely. Instead there is a time limit for persisting logs, that if breached, results in giving up on the persist. The default timeout is 5 minutes, but can be configured by overriding `druid.indexer.runner.logSaveTimeout` with a valid Duration (e.g `PT60S`)

<hr>

 * `KubernetesPeonLifecycle.java`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
+++ b/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
@@ -49,7 +49,11 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
@@ -92,6 +96,7 @@ public class KubernetesPeonLifecycle
   private final ObjectMapper mapper;
   private final TaskStateListener stateListener;
   private final SettableFuture<Boolean> taskStartedSuccessfullyFuture;
+  private final long logSaveTimeoutMs;
 
   @MonotonicNonNull
   private LogWatch logWatch;
@@ -104,7 +109,8 @@ public class KubernetesPeonLifecycle
       KubernetesPeonClient kubernetesClient,
       TaskLogs taskLogs,
       ObjectMapper mapper,
-      TaskStateListener stateListener
+      TaskStateListener stateListener,
+      long logSaveTimeoutMs
   )
   {
     this.task = task;
@@ -114,6 +120,7 @@ public class KubernetesPeonLifecycle
     this.mapper = mapper;
     this.stateListener = stateListener;
     this.taskStartedSuccessfullyFuture = SettableFuture.create();
+    this.logSaveTimeoutMs = logSaveTimeoutMs;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -349,6 +356,24 @@ protected void startWatchingLogs()
   }
 
   protected void saveLogs()
+  {
+    ExecutorService executor = Executors.newSingleThreadExecutor();
```
cryptoe: Does it make sense to expose the timeouts here
https://github.com/fabric8io/kubernetes-client?tab=readme-ov-file#configuring-the-client
rather than doing a timeout via an exec service ?

capistrant: > Does it make sense to expose the timeouts here
> 
> https://github.com/fabric8io/kubernetes-client?tab=readme-ov-file#configuring-the-client
> 
> rather than doing a timeout via an exec service ?
> 
> 
> 
> 

https://github.com/fabric8io/kubernetes-client/issues/7163

From talking with @kgyrtkirk i was under the impression that the reason this log read can hang is because of the above bug. I don't think any existing timeouts can fix that. But we could also expose all the timeouts that fabric8 supplies as part of this PR?

cryptoe: I see thanks for clarifying.

cryptoe: Lets name the thread.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -349,6 +356,24 @@ protected void startWatchingLogs()
...
+    try {
+      Future<?> future = executor.submit(this::doSaveLogs);
+      future.get(logSaveTimeoutMs, TimeUnit.MILLISECONDS);
+    }
+    catch (TimeoutException e) {
+      log.warn("saveLogs() timed out after %d ms for task [%s]", logSaveTimeoutMs, taskId.getOriginalTaskId());
+    }
+    catch (Exception e) {
+      log.error(e, "saveLogs() failed for task [%s]", taskId.getOriginalTaskId());
```
cryptoe: Since tasks logs would not visible this has to be phrased in a manner that the cluster admin can understand. 
`Unable to save logs from the for the task[s] on location[details]. This does not have any impact on the work done by of the task. If this continues to happen, check Kubernetes server logs for potential errors `.

cryptoe (approved):
Minor comments. LGTM otherwise.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
kgyrtkirk: this seems to create a new executor for every call?
that looks a bit odd 

I don't know how many additional patches will be needed to make this fabric8+vertx stuff work correctly.
Would it be hard to add a knob which could restore the okhttp usage?
if we are about to handle the timeouts ourselves - is fabric8 the right choice of library? 
I was taking a quick glance at https://github.com/kubernetes-client/java and it seemed more straightforward...

capistrant: @kgyrtkirk you raise good points. The only reason I'm proposing this is because of the fabric8 bug not being available in any release. I'd like to remove it ASAP once fabric8 does a release or patch release with this fix. maybe we need to take a step back and consider making the client pluggable allowing for operators to migrate between clients. vertx has solved a lot of problems for some use cases but caused other problems.

capistrant: > this seems to create a new executor for every call? that looks a bit odd

new executor per task when this saveLogs is called once when task completes

PR #18448, opened 2025-08-28
Poll from memory before fetching task information from DB

jtuglu1:
### Description

When running high #s of tasks in a cluster, the Overlord can become bottlenecked by the endpoint `/task/{taskid}/status`.
Because `taskQueryTool.getTaskInfo(taskid)` currently only issues a direct I/O to the underlying task storage, this results in heavy I/O + serde from the metadata store to pull the associated task information.

This change stores/updates the `TaskInfo` information inside the `activeTasks` queue's `TaskEntry` type, allowing the requests to fetch from memory if the task entry exists, otherwise falling back to the DB.

Some lingering questions:

#### Code Style/Smell
I didn't really like the idea of shoving the `TaskInfo` into the `TaskEntry`, but it felt like the most convenient way of storing all the necessary information needed. We're really just relying on the `TaskInfo` for the creation time as well as some up-to-date value of the current `TaskStatus`. It might be worth standardizing on `TaskInfo` as `Task` and `TaskStatus` are often paired with each other.

##### `TaskStatus`
Despite updating the task status on `TaskQueue::notifyStatus` callbacks, I wonder whether it may be possible for the respective states in task storage and memory to diverge for a given task's status. 
- In normal state, my initial thought is that the value returned from `TaskQueryTool::getActiveTaskInfo` and `TaskQueryTool::getTaskInfo` should always return state that is *at least* as recent as the latest commit to the DB, since the status updates to the DB in `TaskQueue::notifyStatus` are in the same critical section as those to the `TaskEntry` value. There might be merit in updating the task status in the critical section once it's been committed to the DB, however.
- In a failure scenario, the overlord could receive requests prior to initially (or recently) syncing with DB, resulting in partial/incorrect results. 

##### `TaskQueryTool::getAllActiveTasks`
Prior to this change, the function would use a dummy timestamp for both creation/queue insertion timestamps in the returned payload. Now that we can serve accurate timestamps, I believe we can still use the same timestamp for both fields.

#### Release note

Poll from memory before fetching task information from DB

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/HeapMemoryTaskStorage.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/HeapMemoryTaskStorage.java
@@ -76,7 +76,7 @@ public class HeapMemoryTaskStorage implements TaskStorage
   }
 
   @Override
-  public void insert(Task task, TaskStatus status)
+  public TaskInfo<Task, TaskStatus> insert(Task task, TaskStatus status)
   {
     Preconditions.checkNotNull(task, "task");
     Preconditions.checkNotNull(status, "status");
@@ -94,6 +94,7 @@ public class HeapMemoryTaskStorage implements TaskStorage
     }
 
     log.info("Inserted task[%s] with status[%s]", task.getId(), status);
+    return TaskStuff.toTaskInfo(newTaskStuff);
   }
 
   @Override
@@ -154,6 +155,18 @@ public class HeapMemoryTaskStorage implements TaskStorage
     for (final TaskStuff taskStuff : tasks.values()) {
       if (taskStuff.getStatus().isRunnable()) {
         listBuilder.add(taskStuff.getTask());
+      }
+    }
+    return listBuilder.build();
+  }
+
+  @Override
+  public List<TaskInfo<Task, TaskStatus>> getActiveTaskInfos()
+  {
+    final ImmutableList.Builder<TaskInfo<Task, TaskStatus>> listBuilder = ImmutableList.builder();
+    for (final TaskStuff taskStuff : tasks.values()) {
+      if (taskStuff.getStatus().isRunnable()) {
+        listBuilder.add(TaskStuff.toTaskInfo(taskStuff));
       }
     }
     return listBuilder.build();
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/MetadataTaskStorage.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/MetadataTaskStorage.java
@@ -46,6 +46,7 @@ import org.apache.druid.metadata.MetadataStorageTablesConfig;
 import org.apache.druid.metadata.TaskLookup;
 import org.apache.druid.metadata.TaskLookup.ActiveTaskLookup;
 import org.apache.druid.metadata.TaskLookup.TaskLookupType;
+import org.joda.time.DateTime;
 
 import javax.annotation.Nullable;
 import java.util.Collections;
@@ -111,7 +112,7 @@ public class MetadataTaskStorage implements TaskStorage
   }
 
   @Override
-  public void insert(final Task task, final TaskStatus status)
+  public TaskInfo<Task, TaskStatus> insert(final Task task, final TaskStatus status)
   {
     Preconditions.checkNotNull(task, "task");
     Preconditions.checkNotNull(status, "status");
@@ -123,11 +124,12 @@ public class MetadataTaskStorage implements TaskStorage
     );
 
     log.info("Inserting task [%s] with status [%s].", task.getId(), status);
+    final DateTime insertionTime = DateTimes.nowUtc();
 
     try {
       handler.insert(
           task.getId(),
-          DateTimes.nowUtc(),
+          insertionTime,
           task.getDataSource(),
           task,
           status.isRunnable(),
@@ -142,6 +144,14 @@ public class MetadataTaskStorage implements TaskStorage
     catch (Exception e) {
       throw new RuntimeException(e);
     }
+
+    return new TaskInfo<>(
+        task.getId(),
+        insertionTime,
+        status,
+        task.getDataSource(),
+        task
+    );
   }
 
   @Override
@@ -186,6 +196,15 @@ public class MetadataTaskStorage implements TaskStorage
                   .stream()
                   .filter(taskInfo -> taskInfo.getStatus().isRunnable() && taskInfo.getTask() != null)
                   .map(TaskInfo::getTask)
+                  .collect(Collectors.toList());
+  }
+
+  @Override
+  public List<TaskInfo<Task, TaskStatus>> getActiveTaskInfos()
+  {
+    return handler.getTaskInfos(Collections.singletonMap(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
+                  .stream()
+                  .filter(taskInfo -> taskInfo.getStatus().isRunnable() && taskInfo.getTask() != null)
                   .collect(Collectors.toList());
   }
 
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/MetadataTaskStorage.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/MetadataTaskStorage.java
@@ -192,7 +192,7 @@ public class MetadataTaskStorage implements TaskStorage
   {
     // filter out taskInfo with a null 'task' which should only happen in practice if we are missing a jackson module
     // and don't know what to do with the payload, so we won't be able to make use of it anyway
-    return handler.getTaskInfos(Collections.singletonMap(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
+    return handler.getTaskInfos(Map.of(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
                   .stream()
                   .filter(taskInfo -> taskInfo.getStatus().isRunnable() && taskInfo.getTask() != null)
                   .map(TaskInfo::getTask)
@@ -202,7 +202,7 @@ public class MetadataTaskStorage implements TaskStorage
   @Override
   public List<TaskInfo<Task, TaskStatus>> getActiveTaskInfos()
   {
-    return handler.getTaskInfos(Collections.singletonMap(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
+    return handler.getTaskInfos(Map.of(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
                   .stream()
                   .filter(taskInfo -> taskInfo.getStatus().isRunnable() && taskInfo.getTask() != null)
                   .collect(Collectors.toList());
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/MetadataTaskStorage.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/MetadataTaskStorage.java
@@ -192,11 +192,9 @@ public class MetadataTaskStorage implements TaskStorage
   {
     // filter out taskInfo with a null 'task' which should only happen in practice if we are missing a jackson module
     // and don't know what to do with the payload, so we won't be able to make use of it anyway
-    return handler.getTaskInfos(Map.of(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
-                  .stream()
-                  .filter(taskInfo -> taskInfo.getStatus().isRunnable() && taskInfo.getTask() != null)
-                  .map(TaskInfo::getTask)
-                  .collect(Collectors.toList());
+    return getActiveTaskInfos().stream()
+                               .map(TaskInfo::getTask)
+                               .collect(Collectors.toList());
   }
 
   @Override
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
@@ -426,7 +426,7 @@ public class TaskQueue
   {
     // Don't do anything with tasks that have recently finished; notifyStatus will handle it.
     if (entry != null && !entry.isComplete) {
-      final Task task = entry.taskInfo.getTask();
+      final Task task = entry.getTask();
 
       if (entry.future == null) {
         if (runnerTaskFuture == null) {
@@ -569,7 +569,7 @@ public class TaskQueue
 
     if (added.get()) {
       taskLockbox.add(taskInfo.getTask());
-    } else if (!entry.taskInfo.getTask().equals(taskInfo.getTask())) {
+    } else if (!entry.getTask().equals(taskInfo.getTask())) {
       throw new ISE("Cannot add task[%s] as a different task for the same ID has already been added.", taskInfo.getTask().getId());
     }
   }
@@ -688,7 +688,7 @@ public class TaskQueue
       return;
     }
 
-    final Task task = entry.taskInfo.getTask();
+    final Task task = entry.getTask();
     Preconditions.checkNotNull(task, "task");
     Preconditions.checkNotNull(taskStatus, "status");
     Preconditions.checkState(active, "Queue is not active!");
@@ -925,7 +925,7 @@ public class TaskQueue
   {
     return activeTasks.values().stream()
                       .filter(entry -> !entry.isComplete)
-                      .map(entry -> entry.taskInfo.getTask())
+                      .map(entry -> entry.getTask())
                       .collect(Collectors.toMap(Task::getId, TaskQueue::getMetricKey));
   }
 
@@ -962,7 +962,7 @@ public class TaskQueue
 
     return activeTasks.values().stream()
                       .filter(entry -> !entry.isComplete)
-                      .map(entry -> entry.taskInfo.getTask())
+                      .map(entry -> entry.getTask())
                       .filter(task -> !runnerKnownTaskIds.contains(task.getId()))
                       .collect(Collectors.toMap(TaskQueue::getMetricKey, task -> 1L, Long::sum));
   }
@@ -1061,7 +1061,7 @@ public class TaskQueue
         entry -> !entry.isComplete
                  && entry.taskInfo.getDataSource().equals(datasource)
     ).map(
-        entry -> entry.taskInfo.getTask()
+        entry -> entry.getTask()
     ).collect(
         Collectors.toMap(Task::getId, Function.identity())
     );
@@ -1174,6 +1174,14 @@ public class TaskQueue
     {
       this.taskInfo = taskInfo;
       this.lastUpdatedTime = DateTimes.nowUtc();
+    }
+
+    /**
+     * Returns the task associated with this {@link TaskEntry}
+     */
+    public Task getTask()
+    {
+      return taskInfo.getTask();
     }
 
     /**
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
@@ -251,7 +251,6 @@ public class TaskQueue
                 }
               }
             }
-
           }
       );
       ScheduledExecutors.scheduleAtFixedRate(
@@ -560,12 +559,8 @@ public class TaskQueue
             added.set(true);
             return new TaskEntry(taskInfo);
           } else if (prevEntry.lastUpdatedTime.isBefore(updateTime)) {
-            prevEntry.lastUpdatedTime = updateTime;
-          }
-
-          // Ensure we keep the current status up-to-date
-          if (!prevEntry.taskInfo.getStatus().equals(taskInfo.getStatus())) {
-            prevEntry.taskInfo = taskInfo;
+            // Ensure we keep the current status up-to-date
+            prevEntry.updateStatus(taskInfo.getStatus(), updateTime);
           }
 
           return prevEntry;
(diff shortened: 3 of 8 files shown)


processing/src/main/java/org/apache/druid/indexer/TaskInfo.java
```diff
@@ -76,4 +76,14 @@ public EntryType getTask()
   {
     return task;
   }
+
+  /**
+   * Returns a copy of this TaskInfo object with a new StatusType
+   * @param newStatus
+   * @return a new TaskInfo
```
kfaraz:
```suggestion
   * Returns a copy of this TaskInfo object with the given status.
```

processing/src/main/java/org/apache/druid/indexer/TaskInfo.java
```diff
@@ -76,4 +76,14 @@ public EntryType getTask()
...
+   */
+  public TaskInfo<EntryType, StatusType> withNewStatus(StatusType newStatus)
```
kfaraz: Nit: I think the `new` prefix can be omitted.

```suggestion
  public TaskInfo<EntryType, StatusType> withStatus(StatusType status)
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/MetadataTaskStorage.java
```diff
@@ -189,6 +199,15 @@ public List<Task> getActiveTasks()
                   .collect(Collectors.toList());
   }
 
+  @Override
+  public List<TaskInfo<Task, TaskStatus>> getActiveTaskInfos()
+  {
+    return handler.getTaskInfos(Collections.singletonMap(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
```
kfaraz: Nit: prefer Map.of as it is more concise.
```suggestion
    return handler.getTaskInfos(Map.of(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -1140,15 +1168,15 @@ private void removeTaskLock(Task task)
...
+    private TaskInfo<Task, TaskStatus> taskInfo;
 
     private DateTime lastUpdatedTime;
     private ListenableFuture<TaskStatus> future = null;
     private boolean isComplete = false;
 
-    TaskEntry(Task task)
+    TaskEntry(TaskInfo<Task, TaskStatus> taskInfo)
     {
-      this.task = task;
```
kfaraz, line 1151: We could have retained this field to simplify the patch since the `Task` is going to be final anyway.
Only the status can change:

```
this.task = taskInfo.getTask()
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -1020,24 +1027,45 @@ public Optional<Task> getActiveTask(String id)
   }
 
   /**
-   * List of all active and completed tasks currently being managed by this
-   * TaskQueue.
+   * Polls {@link #activeTasks} for the task with the corresponding {@code taskId}
+   * @param taskId
+   * @return an optional TaskInfo
```
kfaraz:
```suggestion
   * Gets the {@link TaskInfo} for the given {@code taskId} from {@link #activeTasks} if present,
   * otherwise returns an empty optional.
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -835,26 +844,24 @@ void syncFromStorage()
 
     try {
       if (active) {
-        final Map<String, Task> newTasks =
-            CollectionUtils.toMap(taskStorage.getActiveTasks(), Task::getId, Function.identity());
-        final Map<String, Task> oldTasks =
-            CollectionUtils.mapValues(activeTasks, entry -> entry.task);
+        final Map<String, TaskInfo<Task, TaskStatus>> newTasks = CollectionUtils.toMap(taskStorage.getActiveTaskInfos(), (taskInfo) -> taskInfo.getTask().getId(), Function.identity());
```
kfaraz: Style: Might be nicer to put each arg in a separate line.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -835,26 +844,24 @@ void syncFromStorage()
...
-        final Collection<Task> addedTasks = mapDifference.entriesOnlyOnRight().values();
-        final Collection<Task> removedTasks = mapDifference.entriesOnlyOnLeft().values();
+        final MapDifference<String, TaskInfo<Task, TaskStatus>> mapDifference = Maps.difference(oldTasks, newTasks);
+        final Collection<TaskInfo<Task, TaskStatus>> addedTasks = mapDifference.entriesOnlyOnRight().values();
+        final Collection<TaskInfo<Task, TaskStatus>> removedTasks = mapDifference.entriesOnlyOnLeft().values();
 
         // Remove tasks not present in metadata store if their lastUpdatedTime is before syncStartTime
         int numTasksRemoved = 0;
-        for (Task task : removedTasks) {
+        for (TaskInfo<Task, TaskStatus> task : removedTasks) {
```
kfaraz, line 852: Looking at this patch, I wonder if `TaskInfo<Task, TaskStatus>` shouldn't be a separate class of its own.
Probably named `TaskWithStatus`.

(I am not even sure why this class was generic in the first place. The only other possible type I see the generic
param taking is `TaskIdentifier`, which could have been its own class `TaskInfoWithId` or something. But we can address this later).

indexing-service/src/test/java/org/apache/druid/indexing/overlord/http/OverlordResourceTest.java
```diff
@@ -131,7 +131,7 @@ public void setUp()
     provisioningStrategy = EasyMock.createMock(ProvisioningStrategy.class);
     authConfig = EasyMock.createMock(AuthConfig.class);
     overlord = EasyMock.createStrictMock(DruidOverlord.class);
-    taskMaster = EasyMock.createStrictMock(TaskMaster.class);
+    taskMaster = EasyMock.createMock(TaskMaster.class);
```
kfaraz: Why not strict anymore?

kfaraz (commented):
Thanks, @jtuglu1 ! Patch looks good.
I have left some minor suggestions in a couple of places.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
jtuglu1: Yeah. I opted for removing since in many cases, we may overwrite the `TaskInfo`, which already contains a `Task` so it didn't make sense to me to keep both fields.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
jtuglu1: Agreed

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -250,6 +251,7 @@ public void start()
                 }
               }
             }
+
```
kfaraz: Nit: extra newline.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -548,27 +550,32 @@ public boolean add(final Task task)
...
             added.set(true);
-            return new TaskEntry(task);
+            return new TaskEntry(taskInfo);
           } else if (prevEntry.lastUpdatedTime.isBefore(updateTime)) {
             prevEntry.lastUpdatedTime = updateTime;
           }
 
+          // Ensure we keep the current status up-to-date
+          if (!prevEntry.taskInfo.getStatus().equals(taskInfo.getStatus())) {
+            prevEntry.taskInfo = taskInfo;
```
kfaraz: This update should happen inside the `if` clause before this, so that we have a consistent view.
In fact, we should add a method to `TaskEntry.update(TaskInfo info)` which updates both the task info and the last updated time.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
kfaraz: Yeah, that's reasonable.

But it was nice to have the `final Task` field since the task never changes, and it allows for easy access.
We are only ever going to update the status of the task, never the payload itself.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
kfaraz: Alternatively, add a method `updateStatus(TaskStatus status)` which assigns a new taskinfo object inside the `TaskEntry` like `this.taskInfo = taskInfo.withStatus(status)`.
Having this method would clarify that only the status field is going to change.

indexing-service/src/test/java/org/apache/druid/indexing/overlord/http/OverlordResourceTest.java
jtuglu1: I've re-worked the ordering of the mock expectations so this shouldn't be necessary anymore.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
jtuglu1: > This update should happen inside the if clause before this, so that we have a consistent view.

Yeah, this seems right to me – we don't want to assign an old/stale status.

I think pairing these 2 updates inside a `TaskEntry` helper makes sense, as it'll ensure that future updates to the code always pair these updates together. In this case, it's a bit lucky that `notifyStatus` is always wrapped by the `updateTaskEntry`, which ensures the `lastUpdatedTime`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/MetadataTaskStorage.java
jtuglu1: Updated in `getActiveTasks` above as well.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -716,7 +711,7 @@ private void notifyStatus(final TaskEntry entry, final TaskStatus taskStatus, St
     // Mark this task as complete, so it isn't managed while being cleaned up.
     entry.isComplete = true;
     // Update the task status associated with this entry
-    entry.taskInfo = entry.taskInfo.withNewStatus(taskStatus);
+    entry.taskInfo = entry.taskInfo.withStatus(taskStatus);
```
kfaraz: Shouldn't this also go through `entry.updateStatus()`?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -1140,17 +1164,27 @@ private void removeTaskLock(Task task)
...
       this.lastUpdatedTime = DateTimes.nowUtc();
     }
+
+    /**
+     * Updates the {@link TaskStatus} for the task associated with this {@link TaskEntry} and sets the corresponding
+     * update time.
+     */
+    public void updateStatus(TaskStatus status, DateTime updateTime)
+    {
+      this.taskInfo.withStatus(status);
```
kfaraz: This just creates a new instance of `TaskInfo` but doesn't assign it to the `taskInfo` field.
So this method is not actually updating the status here.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
jtuglu1: Added `TaskEntry::getTask()` which references underlying `Task` stored inside `TaskInfo`. Since this is `final` inside the class, and we only create immutable `TaskInfo` objects using `withStatus` builder, I figured this should be fine.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -548,27 +549,28 @@ public boolean add(final Task task)
...
-        task.getId(),
+        taskInfo.getId(),
         prevEntry -> {
           if (prevEntry == null) {
             added.set(true);
-            return new TaskEntry(task);
+            return new TaskEntry(taskInfo);
           } else if (prevEntry.lastUpdatedTime.isBefore(updateTime)) {
-            prevEntry.lastUpdatedTime = updateTime;
+            // Ensure we keep the current status up-to-date
```
kfaraz: Nit: not really needed.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -835,26 +839,26 @@ void syncFromStorage()
 
     try {
       if (active) {
-        final Map<String, Task> newTasks =
-            CollectionUtils.toMap(taskStorage.getActiveTasks(), Task::getId, Function.identity());
-        final Map<String, Task> oldTasks =
-            CollectionUtils.mapValues(activeTasks, entry -> entry.task);
+        final Map<String, TaskInfo<Task, TaskStatus>> newTasks =
+            CollectionUtils.toMap(taskStorage.getActiveTaskInfos(), (taskInfo) -> taskInfo.getTask().getId(), Function.identity());
```
kfaraz: shorter alternative:
```suggestion
            CollectionUtils.toMap(taskStorage.getActiveTaskInfos(), TaskInfo::getId, Function.identity());
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -921,7 +925,7 @@ private Map<String, RowKey> getCurrentTaskDatasources()
   {
     return activeTasks.values().stream()
                       .filter(entry -> !entry.isComplete)
-                      .map(entry -> entry.task)
+                      .map(entry -> entry.getTask())
```
kfaraz:
```suggestion
                      .map(TaskEntry::getTask)
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -958,7 +962,7 @@ public Map<RowKey, Long> getWaitingTaskCount()
 
     return activeTasks.values().stream()
                       .filter(entry -> !entry.isComplete)
-                      .map(entry -> entry.task)
+                      .map(entry -> entry.getTask())
```
kfaraz:
```suggestion
                      .map(TaskEntry::getTask)
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -1020,24 +1024,44 @@ public Optional<Task> getActiveTask(String id)
...
+   * Gets the {@link TaskInfo} for the given {@code taskId} from {@link #activeTasks} if present,
+   * otherwise returns an empty optional.
+   */
+  public Optional<TaskInfo<Task, TaskStatus>> getActiveTaskInfo(String taskId)
+  {
+    final TaskEntry entry = activeTasks.get(taskId);
+    if (entry == null) {
+      return Optional.absent();
+    }
+    return Optional.of(entry.taskInfo);
```
kfaraz: Nit: shorter alternative
```suggestion
    return entry == null ? Optional.absent() : Optional.of(entry.taskInfo);
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -1140,17 +1164,35 @@ private void removeTaskLock(Task task)
...
     {
-      this.task = task;
+      this.taskInfo = taskInfo;
       this.lastUpdatedTime = DateTimes.nowUtc();
     }
+
+    /**
+     * Returns the task associated with this {@link TaskEntry}
+     */
+    public Task getTask()
```
kfaraz:
```suggestion
    Task getTask()
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -1140,17 +1164,35 @@ private void removeTaskLock(Task task)
...
+    {
+      return taskInfo.getTask();
+    }
+
+    /**
+     * Updates the {@link TaskStatus} for the task associated with this {@link TaskEntry} and sets the corresponding
+     * update time.
+     */
+    public void updateStatus(TaskStatus status, DateTime updateTime)
```
kfaraz:
```suggestion
    void updateStatus(TaskStatus status, DateTime updateTime)
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/MetadataTaskStorage.java
```diff
@@ -182,13 +192,22 @@ public List<Task> getActiveTasks()
   {
     // filter out taskInfo with a null 'task' which should only happen in practice if we are missing a jackson module
     // and don't know what to do with the payload, so we won't be able to make use of it anyway
-    return handler.getTaskInfos(Collections.singletonMap(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
+    return handler.getTaskInfos(Map.of(TaskLookupType.ACTIVE, ActiveTaskLookup.getInstance()), null)
                   .stream()
                   .filter(taskInfo -> taskInfo.getStatus().isRunnable() && taskInfo.getTask() != null)
```
kfaraz: Should we simplify this to the following?
```suggestion
    return handler.getActiveTaskInfos()
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -548,27 +549,28 @@ public boolean add(final Task task)
...
         }
     );
 
     if (added.get()) {
-      taskLockbox.add(task);
-    } else if (!entry.task.equals(task)) {
-      throw new ISE("Cannot add task[%s] as a different task for the same ID has already been added.", task.getId());
+      taskLockbox.add(taskInfo.getTask());
+    } else if (!entry.getTask().equals(taskInfo.getTask())) {
+      throw new ISE("Cannot add task[%s] as a different task for the same ID has already been added.", taskInfo.getTask().getId());
```
kfaraz: Can use `taskInfo.getId()` instead of `taskInfo.getTask().getId()`

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -708,6 +710,8 @@ private void notifyStatus(final TaskEntry entry, final TaskStatus taskStatus, St
 
     // Mark this task as complete, so it isn't managed while being cleaned up.
     entry.isComplete = true;
+    // Update the task status associated with this entry
```
kfaraz: Not really needed.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -1020,24 +1024,44 @@ public Optional<Task> getActiveTask(String id)
...
    */
   public Map<String, Task> getActiveTasksForDatasource(String datasource)
   {
     return activeTasks.values().stream().filter(
         entry -> !entry.isComplete
-                 && entry.task.getDataSource().equals(datasource)
+                 && entry.taskInfo.getDataSource().equals(datasource)
     ).map(
-        entry -> entry.task
+        entry -> entry.getTask()
```
kfaraz:
```suggestion
        TaskEntry::getTask
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/HeapMemoryTaskStorage.java
```diff
@@ -159,6 +160,18 @@ public List<Task> getActiveTasks()
...
 
+  @Override
+  public List<TaskInfo<Task, TaskStatus>> getActiveTaskInfos()
+  {
+    final ImmutableList.Builder<TaskInfo<Task, TaskStatus>> listBuilder = ImmutableList.builder();
+    for (final TaskStuff taskStuff : tasks.values()) {
+      if (taskStuff.getStatus().isRunnable()) {
+        listBuilder.add(TaskStuff.toTaskInfo(taskStuff));
+      }
+    }
```
samarthjain, line 171: Feel free to ignore the following suggestion. I suppose the following looks more readable:

```
@Override
public List<TaskInfo<Task, TaskStatus>> getActiveTaskInfos() {
  return tasks.values().stream()
      .filter(taskStuff -> taskStuff.getStatus().isRunnable())
      .map(TaskStuff::toTaskInfo)
      .collect(ImmutableList.toImmutableList());
}
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
jtuglu1: The updated timestamp will already be assigned after this function returns – I can still add it.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/HeapMemoryTaskStorage.java
jtuglu1: Sure, I just copied the format of the surrounding methods.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
jtuglu1: Yeah – in this case `taskInfo.getId()` will always be the task ID, I originally left it as accessing the task directly for safety.

kfaraz (approved):
LGTM 👍🏻

PR #18451, opened 2025-08-28
fixes for projection matching logic

clintropolis:
### Description
Fixes to projection matching logic, simplifying things by removing some match tracking which could cause bugs and was previously necessary to account for what was a flaw in the unnest `CursorBuildSpec` translation (also joins had this same flaw)

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/UnnestCursorFactory.java
+++ b/processing/src/main/java/org/apache/druid/segment/UnnestCursorFactory.java
@@ -22,8 +22,8 @@ package org.apache.druid.segment;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableMap;
-import org.apache.druid.java.util.common.Pair;
 import org.apache.druid.java.util.common.io.Closer;
+import org.apache.druid.query.Order;
 import org.apache.druid.query.OrderBy;
 import org.apache.druid.query.filter.BooleanFilter;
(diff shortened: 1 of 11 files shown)


processing/src/main/java/org/apache/druid/segment/UnnestCursorFactory.java
```diff
@@ -475,6 +453,73 @@ public static ColumnCapabilities computeOutputColumnCapabilities(
     }
   }
 
+  /**
+   * Converts a {@link CursorBuildSpec} to the base table {@link CursorBuildSpec}, ensuring that all required columns
+   * of are added to {@link CursorBuildSpec#getPhysicalColumns()} and the unnest column is added to {@link CursorBuildSpec#getVirtualColumns()}
```
capistrant: nit: `ensuring that all required columns of ? are added to ...` potentially missing a word where `?` is?

processing/src/main/java/org/apache/druid/segment/UnnestCursorFactory.java
```diff
@@ -475,6 +453,73 @@ public static ColumnCapabilities computeOutputColumnCapabilities(
...
+   */
+  @VisibleForTesting
+  public static CursorBuildSpec transformCursorBuildSpec(
```
capistrant: I don't see any usage outside of the `org.apache.druid.segment` package, should this drop the `public` visibility since it is marked as visiblefortesting?

PR #18455, opened 2025-08-29
Display time in local time

GabrielCWT:
Screenshot 2025-08-29 at 4 11 31 PM

To allow users to more easily interpret times.

### DIFF
--- a/web-console/package-lock.json
+++ b/web-console/package-lock.json
@@ -30,6 +30,7 @@
         "d3-shape": "^3.2.0",
         "d3-time-format": "^4.1.0",
         "date-fns": "^2.28.0",
+        "dayjs": "^1.11.15",
         "druid-query-toolkit": "^1.1.5",
         "echarts": "^5.5.1",
         "file-saver": "^2.0.5",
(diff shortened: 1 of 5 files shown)


FrankChen021:
@vogievetsky I think this change is reasonable. What do you think?

kfaraz:
Given that we have always had the time displayed in UTC, this should be a config in the web-console that can be changed in the settings (top right corner).

GabrielCWT:
> Given that we have always had the time displayed in UTC, this should be a config in the web-console that can be changed in the settings (top right corner).

Agreed. I have created a new tab for "Web Console Configurations", currently it will just be this local time config

vogievetsky:
Just saw this. Sounds like a cool idea. Will review in detail later today or tomorrow.

GabrielCWT:
Hi @vogievetsky could you take a look at this PR when you have the time? Thanks!

FrankChen021:
Hi @vogievetsky It's better that this can be merged into 35. 
If you're too busy to review this change, I will merge it first.

PR #18457, opened 2025-08-29
Optimize ObjectMapper reading in ChangeRequestHttpSyncer

GWphua:
#### Speed up `ChangeRequestHttpSyncer#onSuccess` by caching ObjectMapper's JavaType
- Cache JavaType from provided TypeReference, so ObjectMapper no longer needs to call `constructType` everytime it tries to read from the stream.  
- We can even consider making `ObjectMapper#readValue(InputStream, TypeReference<T>)` a forbidden API, given that we can use the more efficient `ObjectMapper#readValue(InputStream, JavaType)`.
- No behavior change; request count and APIs remain the same.

#### Release note

Internal deserialization optimization in `ChangeRequestHttpSyncer`

<hr>

 * `ChangeRequestHttpSyncer.java`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/coordination/ChangeRequestHttpSyncer.java
+++ b/server/src/main/java/org/apache/druid/server/coordination/ChangeRequestHttpSyncer.java
@@ -20,6 +20,7 @@
 package org.apache.druid.server.coordination;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.jaxrs.smile.SmileMediaTypes;
 import com.google.common.annotations.VisibleForTesting;
@@ -65,11 +66,11 @@ public class ChangeRequestHttpSyncer<T>
   private static final long MAX_RETRY_BACKOFF = TimeUnit.MINUTES.toMillis(2);
 
   private final ObjectMapper smileMapper;
+  private final JavaType changeRequestsSnapshotResponseType;
   private final HttpClient httpClient;
   private final ScheduledExecutorService executor;
   private final URL baseServerURL;
   private final String baseRequestPath;
-  private final TypeReference<ChangeRequestsSnapshot<T>> responseTypeReferences;
   private final long serverTimeoutMS;
   private final long serverHttpTimeout;
 
@@ -116,11 +117,11 @@ public class ChangeRequestHttpSyncer<T>
   )
   {
     this.smileMapper = smileMapper;
+    this.changeRequestsSnapshotResponseType = smileMapper.getTypeFactory().constructType(responseTypeReferences);
     this.httpClient = httpClient;
     this.executor = executor;
     this.baseServerURL = baseServerURL;
     this.baseRequestPath = baseRequestPath;
-    this.responseTypeReferences = responseTypeReferences;
     this.serverTimeoutMS = serverTimeoutMS;
     this.serverHttpTimeout = serverTimeoutMS + HTTP_TIMEOUT_EXTRA_MS;
     this.listener = listener;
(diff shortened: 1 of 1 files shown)


kfaraz:
@GWphua , do you have a benchmark on the perf improvement this change provides?
Wondering if this is something we should do across the board, particularly while reading things from the DB.

FrankChen021:
Based on the document of Jackson, such change is reasonable but if there's specific data that shows the improvement, it would be better. 

And one more thing is, the readValue is widely used in the repo, if we think that the performance matters, maybe we need to forbid current API to use the one that accepts JavaType

GWphua:
@kfaraz @FrankChen021  
I did a benchmark for 50,000 accesses of `readValue()`, and the benchmark favours using `JavaType` over `TypeReference` with the SmileMapper. Furthermore, after referencing [Jackson best practices](https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance), I have tried to benchmark for `ObjectReader`, and the performance does better. 

I think when working with SmileMapper, we can consider using `ObjectReader` in favour of `ObjectMapper`.

```
# JMH version: 1.21
# VM version: JDK 17.0.12, Java HotSpot(TM) 64-Bit Server VM, 17.0.12+8-LTS-286
# VM invoker: /Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home/bin/java
# VM options: <none>
# Warmup: 2 iterations, 10 s each
# Measurement: 3 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op

Benchmark                                             Mode  Cnt      Score     Error  Units
JsonLibrarySerializationBenchmark.typeReference_json  avgt   30  12284.689 ± 128.900  us/op
JsonLibrarySerializationBenchmark.javaType_json       avgt   30  11088.269 ± 260.141  us/op
JsonLibrarySerializationBenchmark.objectReader_json   avgt   30  10952.391 ± 119.274  us/op
```

GWphua:
A further update on the performances, these are the results from 50,000 accesses of readValue() on a small object (~10 characters of string), using a normal mapper (Results for SmileMapper above). While Jackson JR is the most performant, it might be complex to set up. 

Curiously, `JavaType` is more performant here. I tried repeating the benchmark ~2-3 times with similar results. We can look to using `JavaType` for the good ol' jsonMappers for now. They bring some performance boosts for the simplicity of change.

```
Benchmark                                                        Mode  Cnt      Score     Error  Units
JsonLibrarySerializationBenchmark.jacksonJr_json                 avgt   30  10779.843 ± 200.994  us/op
JsonLibrarySerializationBenchmark.javaType_json                  avgt   30  11279.839 ± 368.868  us/op
JsonLibrarySerializationBenchmark.javaType_json_afterburner      avgt   30  11080.339 ± 310.208  us/op
JsonLibrarySerializationBenchmark.objectReader_json              avgt   30  11519.170 ±  46.703  us/op
JsonLibrarySerializationBenchmark.objectReader_json_afterburner  avgt   30  11341.113 ± 183.754  us/op
JsonLibrarySerializationBenchmark.typeReference_json             avgt   30  12168.292 ±  35.448  us/op
```

kfaraz:
Thanks for sharing the benchmarks, @GWphua !

I am sorry but I fail to see a significant improvement, I must be missing something.
In the second dataset, is the perf improvement when replacing `TypeReference` with `JavaType` from 12 ms to 11 ms per operation?

We can merge off this PR if needed but I don't feel that it is really worth the effort right now to change all call sites that use `TypeReference`.

To really attack the perf bottlenecks (either querying, ingestion or management), we can look at flamegraphs of services and specific threads to see where we are spending most of our time.

GWphua:
Yea, you are right about the lack of performance improvement @kfaraz 

From my flame graph below, I thought the performance to be because of the JsonMapper spending a lot of time to break down the type. Though, i found the bulk of the bottleneck is caused by Interval deserialization, which is addressed in #18477

image

While not that impactful, it is still nice to see that we can get a performance enhancement via using JavaType / ObjectReader. We can merge this PR, especially since this path is called a lot especially for larger clusters.

PR #18460, opened 2025-08-30
Add builders for Kafka supervisor spec, ioConfig, tuningConfig

kfaraz:
### Changes

- Add `KafkaSupervisorSpecBuilder`
- Add `KafkaTuningConfigBuilder`
- Add `KafkaIOConfigBuilder`
- Use the above builders in tests

<hr>

This PR has:

- [ ] been self-reviewed.


extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorSpecBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+  public KafkaSupervisorSpecBuilder withId(String id)
+  {
+    this.id = id;
+    return this;
+  }
+
+  /**
+   * Builds a new {@link KafkaSupervisorSpec} with the specified parameters.
+   */
+  public KafkaSupervisorSpec build(String dataSource, String topic)
```
capistrant, line 64: Am I understanding correctly that if you want to use `topicPattern` in the IOConfig, it must have been explicitly set in the underlying IOConfig for this builder? And when doing taht, you then pass null at the call-site?

This feels a bit confusing and requiring insider knowledge. I think at minimum, we should have a `@param` field for topic in the javadoc calling out the case when you want to use topicPattern. But we should also consider the idea of removing `topic` from this method and requiring caller to have handled setting topic/topicPattern in ioConfig. Was it created this way because it made it the most easy to migrate since a vast majority were using topic and only few things use topicPattern?

kfaraz: Thanks for calling this out, @capistrant !
I feel the same, let me fix this up.

kfaraz: It happened to be like this since initially I wanted to force callers to provide a topic name but then later realized that `topicPattern` is an option too.

kfaraz:
Thanks for the review, @capistrant !
I have added a separate method to the builder : `buildWithTopicPattern`.
Please let me know if this helps resolve the ambiguity.

capistrant:
> Thanks for the review, @capistrant ! I have added a separate method to the builder : `buildWithTopicPattern`. Please let me know if this helps resolve the ambiguity.

This looks great, thanks for making the change

capistrant (approved):
+1 on green CI

PR #18463, opened 2025-09-01
Fix inactive task race in TaskLockbox

jtuglu1:
Currently, we check whether a task exists in the activeTasks set, then subsequently release the lock (allowing TaskQueue to call remove() on it). This can lead to undesirable outcomes where task locks are allocated to killed tasks, blocking other compaction, kill tasks.

Fixes #XXXX.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskLockbox.java
```diff
@@ -524,7 +523,9 @@ public List<SegmentAllocateResult> allocateSegments(
 
   /**
    * Marks the segment allocation as failed if the underlying task is not active.
+   * Should be called only while holding the {@code giant} lock.
    */
+  @GuardedBy("giant")
```
kfaraz, line 527: We should probably also annotate the field `activeTasks` with `@GuardedBy("giant")`.

kfaraz (approved):
Thanks for the PR, @jtuglu1 !

kfaraz:
@jtuglu1 , I see you have already mentioned in the PR that you intend to add concurrency tests.
A test setup which always fails before this change would be nice to have.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskLockbox.java
```diff
@@ -567,6 +569,10 @@ private TaskLockPosse createOrFindLockPosse(LockRequest request, Task task, bool
     giant.lock();
 
     try {
+      if (!activeTasks.contains(task.getId())) {
```
jtuglu1, line 570: Not currently necessary, but I put this in for good measure since there's a few call sites here (and in the future) which may pop-up. It's an O(1) (albeit a large constant with large # of tasks though) check already under lock so shouldn't be too expensive. Let me know if you want me to remove.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskLockbox.java
jtuglu1: `@GuardedBy` is function-based and not block-based. In a few places in `GlobalTaskLockbox` we rely on methods doing internal (block-based) locking of giant lock. The `@GuardedBy` cannot "look" inside the body of the function to ensure the lock is held, so it reports some of these codepaths as uncovered.

jtuglu1:
> @jtuglu1 , I see you have already mentioned in the PR that you intend to add concurrency tests. A test setup which always fails before this change would be nice to have.

Yes. I'd like to move the concurrency verification framework within `TaskQueueConcurrencyTest` into a separate package that all indexing-service tests can share. I'll add that in a separate PR.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskLockbox.java
kfaraz: No, I meant the `private final Set<String> activeTasks = new HashSet<>();` inside `TaskLockbox` class should be annotated with guarded by.
The only uncovered path would be `getActiveTasks()` which ~we should fix up~ you seem to have already fixed up.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskLockbox.java
kfaraz: No, I guess it makes sense. Method `tryLock` is already doing this. So maybe let's put it in a common method `ensureActiveTask()`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskLockbox.java
```diff
@@ -524,7 +523,9 @@ public List<SegmentAllocateResult> allocateSegments(
 
   /**
    * Marks the segment allocation as failed if the underlying task is not active.
+   * Should be called only while holding the {@code giant} lock.
```
kfaraz: Nit: I guess this is redundant since we have already annotated the method appropriately.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskLockbox.java
jtuglu1: Yes, that's what I was referring to. No, `getActiveTasks()` is not the only uncovered path (`remove()` is also, according to SureFire). Adding this annotation to that member causes other errors:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project druid-indexing-service: Compilation failure
[ERROR] druid/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskLockbox.java:[1246,8] error: [GuardedBy] This access should be guarded by 'this.giant', which is not currently held
[ERROR]     (see https://errorprone.info/bugpattern/GuardedBy)
```

It seems that it doesn't like:
```java
if (task != null) {
        activeTasks.remove(task.getId());
}
```
being called in the finally.

Doing something like:
```java
    giant.lock();
    try {
      try {
        if (!activeTasks.contains(task.getId())) {
          return;
        }
        log.info("Removing task[%s] from activeTasks", task.getId());
        cleanupUpgradeAndPendingSegments(task);
        unlockAll(task);
      }
      finally {
        if (task != null) {
          activeTasks.remove(task.getId());
        }
      }
    }
    finally {
      giant.unlock();
    }
```
works, however.

I can also disable the check on this specific function.

kfaraz: Ah, I see. Thanks for the clarification. For some reason, my IDE was not flagging this one.
We can skip adding the annotation on `activeTasks` for the time being then since having the two try-finally doesn't seem clean.

PR #18470, opened 2025-09-03
Concurrent linked queue for writes

GWphua:
#### Optimized data structure for heavy writes.

The ConcurrentLinkedQueue features a O(1) write time, as opposed to the O(N) write time for CopyOnWriteArrayList.
Not an impactful performance increase, given the `failedSegments` array should be as small as possible, but caught this while trying to optimize other stuff in `SegmentBootstrapper`.

We can also look to using `ConcurrentLinkedQueue` or `ConcurrentLinkedDeque` for other places in the codebase where there are heavy writes, but do not require array random access.

#### Release note

Optimize concurrent writing to failedSegments array when loading segment on startup.

<hr>

 * `SegmentBootstrapper`

<hr>

This PR has:

- [x] been self-reviewed.


server/src/main/java/org/apache/druid/server/coordination/SegmentBootstrapper.java
```diff
@@ -195,7 +195,7 @@ private void loadSegmentsOnStartup() throws IOException
       final int numSegments = segmentsOnStartup.size();
       final CountDownLatch latch = new CountDownLatch(numSegments);
       final AtomicInteger counter = new AtomicInteger(0);
-      final CopyOnWriteArrayList<DataSegment> failedSegments = new CopyOnWriteArrayList<>();
+      final ConcurrentLinkedQueue<DataSegment> failedSegments = new ConcurrentLinkedQueue<>();
```
kfaraz: Given that only ever add to this list, I wonder if it couldn't just be a `HashSet`.

kfaraz (approved):

server/src/main/java/org/apache/druid/server/coordination/SegmentBootstrapper.java
GWphua: > Given that only ever add to this list,

`failSegments` will also get converted to a ArrayList subsequently in here.

```
log.makeAlert("[%,d] errors seen while loading segments on startup", failedSegments.size())
             .addData("failedSegments", failedSegments)
             .emit();
```

Honestly, not an issue too. I'm ambivalent about both options. 

Would we prefer working with a HashSet here?

kfaraz: Yeah, I am okay either way too. Just wanted to understand if it is really required to be thread safe. 

By the way, why would it get converted to an array list in this line?

GWphua: > By the way, why would it get converted to an array list in this line?


Ah my bad, oversight on my part. 😂
It simply puts the `failedSegments` into a `LinkedHashMap`.

PR #18477, opened 2025-09-04
Optimize interval deserialization

GWphua:
### Description

This PR introduces a faster deserialization strategy when dealing with Intervals.

#### Motivation

Many Druid JSON payloads represent intervals as "YYYY-MM-DD'T'HH:mm:ss.SSSZ/YYYY-MM-DD'T'HH:mm:ss.SSSZ". The general parser behind Intervals.of(String) is flexible but performs extra work (branching, multiple format probes, allocations). This shows up as a hot spot in flame graphs, especially in streaming paths like JsonParserIterator.

image

#### Faster deserialization
Added a fast-path deserializer for `Interval` that works with the common interval string representations, and falls back to `Intervals.of(String)` for non-matching inputs.

We minimize work done per parse by:
- Avoid checking if String can be converted into a `ReadableInterval` / `ReadWriteableInterval`.
- Precompiled `DateTimeFormatter` replaces the flexible parser, with fixed chronology.
- Avoid building 2 `DateTime` objects per parse.

#### Benchmarks
Added a benchmark to test performance of old (legacy) and new (optimized) implementation on 2 cases:
1. The expected interval format
2. The interval format which causes a fallback to the old implementation. (Suffix: Fallback)

```
Benchmark                                                 (numValues)  Mode  Cnt   Score   Error  Units
JodaIntervalDeserialization.deserializeLegacy                   20000  avgt   30  21.466 ± 0.191  ms/op
JodaIntervalDeserialization.deserializeLegacyFallback           20000  avgt   30  13.953 ± 0.249  ms/op
JodaIntervalDeserialization.deserializeOptimized                20000  avgt   30  11.921 ± 0.082  ms/op
JodaIntervalDeserialization.deserializeOptimizedFallback        20000  avgt   30  30.500 ± 0.304  ms/op
```

Benchmark shows an estimated 44% speed-up when working with the interval of format "YYYY-MM-DD'T'HH:mm:ss.SSSZ/YYYY-MM-DD'T'HH:mm:ss.SSSZ".

However, for other formats, the fallback causes the performance to drop. 

#### Release note
Speed up deserialization time of Intervals by >40%

<hr>

 * `processing/src/main/java/org/apache/druid/java/util/common/Intervals.java`
 * `processing/src/main/java/org/apache/druid/jackson/JodaStuff.java`
 * `processing/src/test/java/org/apache/druid/java/util/common/IntervalsTest.java`
 * `benchmarks/src/test/java/org/apache/druid/benchmark/JodaIntervalDeserialization.java`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/benchmarks/src/test/java/org/apache/druid/benchmark/JodaIntervalDeserialization.java
+++ b/benchmarks/src/test/java/org/apache/druid/benchmark/JodaIntervalDeserialization.java
@@ -1,0 +1,152 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.benchmark;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import org.apache.druid.jackson.DefaultObjectMapper;
+import org.apache.druid.java.util.common.Intervals;
+import org.joda.time.Interval;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+@Fork(value = 1)
+@Warmup(iterations = 3)
+@Measurement(iterations = 5)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class JodaIntervalDeserialization
+{
+  @Param({"20000"})
+  public int numValues;
+  private ObjectMapper legacyMapper;
+  private ObjectMapper optimizedMapper;
+  private List<String> intervalJsonValues;
+  private List<String> fallbackIntervalJsonValues;
+
+  public static void main(String[] args) throws RunnerException
+  {
+    Options opt = new OptionsBuilder()
+        .include(JodaIntervalDeserialization.class.getSimpleName())
+        .forks(1)
+        .build();
+    new Runner(opt).run();
+  }
+
+  @Setup
+  public void setUp()
+  {
+    SimpleModule legacyModule = new SimpleModule()
+        .addDeserializer(
+            Interval.class,
+            new StdDeserializer<>(Interval.class)
+            {
+              @Override
+              public Interval deserialize(JsonParser jsonParser, DeserializationContext ctx) throws IOException
+              {
+                return Intervals.of(jsonParser.getText());
+              }
+            }
+        );
+
+    optimizedMapper = new DefaultObjectMapper();
+    legacyMapper = new DefaultObjectMapper().registerModule(legacyModule);
+
+    intervalJsonValues = new ArrayList<>(numValues);
+    fallbackIntervalJsonValues = new ArrayList<>(numValues);
+
+    // Use a small set of valid ISO UTC interval strings that hit the optimized fast path.
+    final String[] samples = new String[]{
+        "\"2022-09-16T00:00:00.000Z/2022-09-17T00:00:00.000Z\"",
+        "\"2021-01-01T12:34:56.789Z/2021-01-02T12:34:56.789Z\"",
+        "\"2010-06-30T23:59:59.000Z/2010-07-01T23:59:59.000Z\"",
+        "\"1999-12-31T00:00:00.123Z/2000-01-01T00:00:00.123Z\""
+    };
+
+    final String[] fallbackSamples = new String[]{
+        "\"2022-01-01T00:00:00.000Z/P1D\"",
+        "\"2022-01-01T12:00:00Z/PT6H\"",
+        "\"2022-01-01T00:00:00Z/P2DT3H4M5S\"",
+        "\"P2DT3H4M5S/2022-01-03T03:04:05Z\""
+    };
+
+    for (int i = 0; i < numValues; i++) {
+      intervalJsonValues.add(samples[i % samples.length]);
+      fallbackIntervalJsonValues.add(fallbackSamples[i % fallbackSamples.length]);
+    }
+  }
+
+  @Benchmark
+  public void deserializeOptimized(Blackhole blackhole) throws Exception
+  {
+    for (String json : intervalJsonValues) {
+      blackhole.consume(optimizedMapper.readValue(json, Interval.class));
+    }
+  }
+
+  @Benchmark
+  public void deserializeLegacy(Blackhole blackhole) throws Exception
+  {
+    for (String json : intervalJsonValues) {
+      blackhole.consume(legacyMapper.readValue(json, Interval.class));
+    }
+  }
+
+  @Benchmark
+  public void deserializeOptimizedFallback(Blackhole blackhole) throws Exception
+  {
+    for (String json : fallbackIntervalJsonValues) {
+      blackhole.consume(optimizedMapper.readValue(json, Interval.class));
+    }
+  }
+
+  @Benchmark
+  public void deserializeLegacyFallback(Blackhole blackhole) throws Exception
+  {
+    for (String json : fallbackIntervalJsonValues) {
+      blackhole.consume(legacyMapper.readValue(json, Interval.class));
+    }
+  }
+}
+
+
--- a/processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
+++ b/processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
@@ -25,6 +25,8 @@ import org.apache.druid.java.util.common.guava.Comparators;
 import org.joda.time.DateTime;
 import org.joda.time.Interval;
 import org.joda.time.chrono.ISOChronology;
+import org.joda.time.format.DateTimeFormatter;
+import org.joda.time.format.ISODateTimeFormat;
 
 import javax.annotation.Nullable;
 
@@ -32,6 +34,8 @@ public final class Intervals
 {
   public static final Interval ETERNITY = utc(JodaUtils.MIN_INSTANT, JodaUtils.MAX_INSTANT);
   public static final ImmutableList<Interval> ONLY_ETERNITY = ImmutableList.of(ETERNITY);
+  private static final DateTimeFormatter FAST_ISO_UTC_FORMATTER =
+      ISODateTimeFormat.dateTime().withChronology(ISOChronology.getInstanceUTC());
 
   public static Interval utc(long startInstant, long endInstant)
   {
@@ -51,6 +55,46 @@ public final class Intervals
   public static Interval of(String format, Object... formatArgs)
   {
     return of(StringUtils.format(format, formatArgs));
+  }
+
+  /**
+   * @return Null if cannot parse with optimized strategy, else return the Interval.
+   */
+  @Nullable
+  private static Interval tryOptimizedIntervalDeserialization(final String intervalText)
+  {
+    final int slashIndex = intervalText.indexOf('/');
+    if (slashIndex <= 0 || slashIndex >= intervalText.length() - 1) {
+      return null;
+    }
+    try {
+      final String startStr = intervalText.substring(0, slashIndex);
+      final long startMillis = FAST_ISO_UTC_FORMATTER.parseMillis(startStr);
+
+      final String endStr = intervalText.substring(slashIndex + 1);
+      final long endMillis = FAST_ISO_UTC_FORMATTER.parseMillis(endStr);
+      return Intervals.utc(startMillis, endMillis);
+    }
+    catch (IllegalArgumentException e) {
+      return null;
+    }
+  }
+
+  /**
+   * Parses a Joda-Time {@link Interval} from its string representation.
+   *
+   * Tries a fast path for ISO-8601 UTC, slash-delimited intervals (e.g.
+   * "2022-09-16T00:00:00.000Z/2022-09-17T00:00:00.000Z"); otherwise falls back to the
+   * general interval parser. The returned interval uses ISO chronology in UTC.
+   */
+  public static Interval deserialize(String string)
+  {
+    final Interval interval = tryOptimizedIntervalDeserialization(string);
+    if (interval == null) {
+      return Intervals.of(string);
+    } else {
+      return interval;
+    }
   }
 
   /**
--- a/processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
+++ b/processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
@@ -57,8 +57,46 @@ public final class Intervals
     return of(StringUtils.format(format, formatArgs));
   }
 
+  /**
+   * A performance-optimized method for parsing a Joda-Time {@link Interval} from a string.
+   * This method is significantly faster than the standard {@link Intervals#of(String)} for the following
+   * group of offsets:
+   * <ol>
+   *   <li>"2022-01-01T00:00:00.000Z/2022-01-02T00:00:00.000Z"</li>
+   *   <li>"2022-01-01T00:00:00.000+05:30/2022-01-01T01:00:00.000+05:30"</li>
+   *   <li>"2022-01-01T00:00:00.000+0530/2022-01-01T01:00:00.000+0530"</li>
+   * </ol>
+   * <p>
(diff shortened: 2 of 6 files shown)


processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
```diff
@@ -53,6 +57,46 @@ public static Interval of(String format, Object... formatArgs)
...
+  }
+
+  /**
+   * Parses a Joda-Time {@link Interval} from its string representation.
+   *
+   * Tries a fast path for ISO-8601 UTC, slash-delimited intervals (e.g.
+   * "2022-09-16T00:00:00.000Z/2022-09-17T00:00:00.000Z"); otherwise falls back to the
+   * general interval parser. The returned interval uses ISO chronology in UTC.
+   */
+  public static Interval deserialize(String string)
```
kfaraz:
```suggestion
  public static Interval fromString(String string)
```

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
```diff
@@ -53,6 +57,46 @@ public static Interval of(String format, Object... formatArgs)
     return of(StringUtils.format(format, formatArgs));
   }
 
+  /**
+   * @return Null if cannot parse with optimized strategy, else return the Interval.
```
kfaraz:
```suggestion
   * @return null if cannot parse with optimized strategy, else return the Interval.
```

kfaraz: For what kind of interval strings does this method return null?

Is there a simpler way to short circuit in those cases rather than trying the parse unnecessarily?\
Maybe a `canDeserializeOptimally()` method which checks it upfront? (Perhaps just check if there is `P` in there 😛 )

kfaraz: Otherwise, the new deserialize becomes much slower than the legacy deserialize for fallback cases (13 vs 30).

benchmarks/src/test/java/org/apache/druid/benchmark/JodaIntervalDeserialization.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+@Fork(value = 1)
+@Warmup(iterations = 3)
+@Measurement(iterations = 5)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class JodaIntervalDeserialization
```
kfaraz:
```suggestion
public class JodaIntervalDeserializationBenchmark
```

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
GWphua: Hi @kfaraz, thanks for the quick review. I have incorporated your suggestions. 

I have listed all cases i can think of under the test file.

```
// Zulu without millis
"2022-01-01T00:00:00Z/2022-01-02T00:00:00Z",
// Date-only
"2022-01-01/2022-01-02",
// start/period
"2022-01-01T12:00:00Z/PT6H",
// period/end
"P2DT3H4M5S/2022-01-01T00:00:00Z"
```

I will work more into short-circuit solutions. Will update this PR soon!

kfaraz: Thanks!

kfaraz: > I have listed all cases i can think of under the test file.

Please also list out these cases briefly in the javadoc of the new `Intervals.fromString()` method

processing/src/test/java/org/apache/druid/java/util/common/IntervalsTest.java
```diff
@@ -79,6 +79,44 @@ public void testFindOverlappingInterval()
...
+        // Date-only
+        "2022-01-01/2022-01-02",
+        // start/period
+        "2022-01-01T00:00:00.000Z/P1D",
+        "2022-01-01T12:00:00Z/PT6H",
+        "2022-01-01T00:00:00Z/P2DT3H4M5S",
+        // period/end
+        "P1D/2022-01-02T00:00:00.000Z",
+        "PT6H/2022-01-01T18:00:00Z",
+        "P2DT3H4M5S/2022-01-03T03:04:05Z"
```
kfaraz, line 110: I feel like these cases could also be deserialized in an optimal manner if we only know which category it falls into.
e.g. if we could quickly identify that the string has a `period/end` configuration, we could parse both parts separately.

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
GWphua: Hi @kfaraz an update on `canDeserializeOptimally()`, I added a check to see if the interval contains a `Period`, and conduct a fail-fast. The results look much better.

```
Benchmark                                                          (numValues)  Mode  Cnt   Score   Error  Units
JodaIntervalDeserializationBenchmark.deserializeLegacy                   20000  avgt    5  21.115 ± 0.302  ms/op
JodaIntervalDeserializationBenchmark.deserializeLegacyFallback           20000  avgt    5  13.831 ± 0.373  ms/op
JodaIntervalDeserializationBenchmark.deserializeOptimized                20000  avgt    5  11.909 ± 0.343  ms/op
JodaIntervalDeserializationBenchmark.deserializeOptimizedFallback        20000  avgt    5  14.014 ± 0.477  ms/op
```

processing/src/test/java/org/apache/druid/java/util/common/IntervalsTest.java
GWphua: Hi @kfaraz 

I attempted to make `tryOptimizedIntervalDeserialization` more inclusive (e.g. Work on Date-only, and Without Millis case), but this requires changing the `FAST_ISO_UTC_FORMATTER` DateTimeFormatter to a more flexible formatter. This causes performance to resemble the un-optimized versions.

These 2 formatters allow formats like `"2022-01-01/2022-01-02"` or `"2022-01-01T00:00:00Z/2022-01-02T00:00:00Z"`, but the performance is lacking...

```
private static final DateTimeFormatter FAST_ISO_UTC_FORMATTER =
      ISODateTimeFormat.dateOptionalTimeParser()
                       .withOffsetParsed()
                       .withChronology(ISOChronology.getInstanceUTC());

Benchmark                                                          (numValues)  Mode  Cnt   Score   Error  Units
JodaIntervalDeserializationBenchmark.deserializeLegacy                   20000  avgt    2  21.243          ms/op
JodaIntervalDeserializationBenchmark.deserializeLegacyFallback           20000  avgt    2  15.217          ms/op
JodaIntervalDeserializationBenchmark.deserializeOptimized                20000  avgt    2  20.605          ms/op
JodaIntervalDeserializationBenchmark.deserializeOptimizedFallback        20000  avgt    2  31.051          ms/op
```

```
private static final DateTimeFormatter FAST_ISO_UTC_FORMATTER =
      ISODateTimeFormat.dateTimeParser().withChronology(ISOChronology.getInstanceUTC());


Benchmark                                                          (numValues)  Mode  Cnt   Score   Error  Units
JodaIntervalDeserializationBenchmark.deserializeLegacy                   20000  avgt    3  21.502 ± 0.995  ms/op
JodaIntervalDeserializationBenchmark.deserializeLegacyFallback           20000  avgt    3  15.365 ± 1.227  ms/op
JodaIntervalDeserializationBenchmark.deserializeOptimized                20000  avgt    3  21.239 ± 1.543  ms/op
JodaIntervalDeserializationBenchmark.deserializeOptimizedFallback        20000  avgt    3  31.526 ± 3.083  ms/op
```

I think we can simply take advantage of knowing how our data will look like, and optimize the parse process for the most frequently-used data format.

GWphua: I think it is possible to check manually, but I feel this method feels abit too hacky, we might miss some DateTime cases:

For start/end string, we check
1. If is Date-only, append T00:00:000Z
2. If does not have millisecond part, append turn T00:00:00Z --> T00:00:00:000Z, and throw it into the formatter.

GWphua: > if we could quickly identify that the string has a period/end configuration, we could parse both parts separately.

Regarding this, I have this solution in mind:
1. Suppose start is a timestamp, end is a period string.
2. Parse the period string to a long. 
3. `endMillis` = period in long + `startMillis`

This method does not work all the time though. We may not get the correct number for leap years, or even months. We will need to calculate start/end period with awareness of these semantics, which is more complex. 

I feel we can stick with knowing that we are dealing with the most common data format (`"YYYY-MM-DD'T'HH:mm:ss.SSSZ/YYYY-MM-DD'T'HH:mm:ss.SSSZ"`) for this method, and use `Interval.of(String)` for other formats?

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
GWphua: I have updated the cases on the Javadocs of `Intervals.fromString()`, please take a look again. 😄

GWphua:
Hi @kfaraz 

Since the benchmark samples involving period show a similar result now, I have replaced the benchmark with new fallback samples, all of which uses non-period in start/end (e.g. without millis specified)

When calling deserialization with the optimized method, we should try our best to not use intervals of this format, else risking the below performance:

1. `"2022-01-01T00:00:00Z/2022-01-02T00:00:00Z"`
2. `"2022-01-01/2022-01-02"`

Benchmark results are as shown:
```
Benchmark                                                          (numValues)  Mode  Cnt   Score   Error  Units
JodaIntervalDeserializationBenchmark.deserializeLegacyFallback           20000  avgt       19.085          ms/op
JodaIntervalDeserializationBenchmark.deserializeOptimizedFallback        20000  avgt       33.561          ms/op
```

FrankChen021:
```
// Zulu without millis
"2022-01-01T00:00:00Z/2022-01-02T00:00:00Z",
// Date-only
"2022-01-01/2022-01-02",
// start/period
"2022-01-01T12:00:00Z/PT6H",
// period/end
"P2DT3H4M5S/2022-01-01T00:00:00Z"
```

do we have such interval string for a segment ? I think considering all cases should be too complicated.
the problem here is mainly related to the deserialization of interval of a segment because of extra large number of segment during loading.

My opinion is that to create a new class named as `SegmentInterval` which inherits from current `Interval` class and then create a dedicated deserialization for this one to handle the segment interval format.

This would simplify the change.

kfaraz:
> My opinion is that to create a new class named as SegmentInterval which inherits from current Interval class and then create a dedicated deserialization for this one to handle the segment interval format.

Ah, that would have been nice but unfortunately `joda.Interval`is a final class.
But we can still do this by passing in a `String` interval to `DataSegment` constructor
and then using `SegmentId.tryParse()` as it already uses the improved algorithm.

GWphua:
@kfaraz I'm having trouble trying to use segmentId's `tryParse()`, since it deals with data of this shape: `"datasource_2015-01-02T00:00:00.000Z_2015-01-03T00:00:00.000Z_ver_0_1"`, but our Interval representation is `"2015-01-02T00:00:00.000Z/2015-01-03T00:00:00.000Z"`. This means we will require some manipulation of the String that's going into DataSegment. lmk if im missing anything!

I can change DataSegment's `interval` input to `String` type, and use `Intervals.fromString()` on it to give the performance boost, since that's the goal of this PR -- to optimize the deserialization time for `DataSegment`. 

I have previously used `Intervals.fromString()` to be the default deserializer in `JodaStuff`. Now will revert that change, and use deserialization for only DataSegments. If we figure out that most of Druid handles interval string representation of the compatible type format, we can consider using the optimized deserialization strategy instead.

kfaraz:
> since it deals with data of this shape: "datasource_2015-01-02T00:00:00.000Z_2015-01-03T00:00:00.000Z_ver_0_1", but our Interval representation is "2015-01-02T00:00:00.000Z/2015-01-03T00:00:00.000Z".

Thanks for checking, @GWphua !
We would need to add some new constructor to `SegmentId` to make it work but we might as well
just use the new method you have already added in this PR.

> I can change DataSegment's interval input to String type, and use Intervals.fromString() on it to give the performance boost, since that's the goal of this PR -- to optimize the deserialization time for DataSegment.
I have previously used Intervals.fromString() to be the default deserializer in JodaStuff. Now will revert that change, and use deserialization for only DataSegments.

Yeah, this makes sense. Let's proceed with this.
Please make this JsonCreator constructor private and add a separate constructor so that all the call sites need not change.

In the javadoc of `Intervals.fromString()`, please indicate that this method is to be used for
segment intervals only (at least for the time being).

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
```diff
@@ -53,6 +57,74 @@ public static Interval of(String format, Object... formatArgs)
...
+   * consider using {@link Intervals#of(String)} instead:
+   * <ol>
+   *   <li>"2022-01-01T00:00:00Z/2022-01-02T00:00:00Z" (without millis)</li>
+   *   <li>"2022-01-01/2022-01-02" (Date only)</li>
+   *   <li>"2022-01-01T12:00:00.000Z/PT6H" (Periods in start / end)</li>
+   * </ol>
+   *
+   * Currently, this method is only used in {@link org.apache.druid.timeline.SegmentId}.
+   * Should you find additional places where the Interval format is guaranteed to be compatible with this method,
+   * feel free to open a PR.
```
kfaraz: Nit: We can leave out these lines.

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
```diff
@@ -25,13 +25,17 @@
...
+import org.joda.time.format.DateTimeFormatter;
+import org.joda.time.format.ISODateTimeFormat;
 
 import javax.annotation.Nullable;
 
 public final class Intervals
 {
   public static final Interval ETERNITY = utc(JodaUtils.MIN_INSTANT, JodaUtils.MAX_INSTANT);
   public static final ImmutableList<Interval> ONLY_ETERNITY = ImmutableList.of(ETERNITY);
+  private static final DateTimeFormatter FAST_ISO_UTC_FORMATTER =
```
kfaraz, line 37: I think you can use `DateTimes.ISO_DATE_TIME` instead of this since it uses the same formatter internally.

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
```diff
@@ -53,6 +57,74 @@ public static Interval of(String format, Object... formatArgs)
...
+    Interval interval = null;
+    if (canDeserializeIntervalOptimallyFromString(string)) {
+      interval = tryOptimizedIntervalDeserialization(string);
+    }
+
+    if (interval == null) {
+      return Intervals.of(string);
+    } else {
+      return interval;
+    }
```
kfaraz: Nit:
```suggestion
    return interval == null ? Intervals.of(string) : interval;
```

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
kfaraz: Maybe also check out the other two constant formatters in that class. See if they can help improve performance of the new method for other date formats. But this can be done in a follow up PR.

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
```diff
@@ -196,6 +197,37 @@ public DataSegment(
       @JsonProperty("size") long size,
       @JacksonInject PruneSpecsHolder pruneSpecsHolder
   )
+  {
+    this(
```
kfaraz, line 202: Please add a comment here mentioning that we take `interval` input as String so that we can deserialize it optimally.

kfaraz (commented):
Changes look good, left some minor nitpicks.

@GWphua , what does the flamegraph look like after the change in this PR?

GWphua:
> @GWphua , what does the flamegraph look like after the change in this PR?

Hi @kfaraz, unfortunately I cannot get a flamegraph this week due to internal code freezes, that's why I am presenting performances via benchmarks as of now. I will also be on a break for a while, so you can expect to hear back from me regarding non-trivial code changes after i return -- probably early October. 

Thanks!

kfaraz:
> Hi @kfaraz, unfortunately I cannot get a flamegraph this week due to internal code freezes, that's why I am presenting performances via benchmarks as of now. I will also be on a break for a while, so you can expect to hear back from me regarding non-trivial code changes after i return -- probably early October.

Sure, @GWphua , thanks for the update! 🙂

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
FrankChen021: Since this method is an optimization and used by the DataSegment only, my opinion is to move this method from the Interval class to the DataSegment.

FrankChen021 (approved):
LGTM with a minor suggestion.

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
GWphua: Hi @FrankChen021 

While this method is only used in `DataSegment` now, should another place find this deserialization useful, we may need to move the method back to `Intervals` to avoid DRY.

Furthermore, I feel that this we should place a method that deals with interval parsing in `Intervals`. What do you think?

processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
GWphua: Hi @kfaraz 

Got quite a bit of time to kill, so I did a benchmark for the formatters...

Seems `DateTimes.ISO_DATE_TIME` is a bit slower than the current optimization because `ISO_DATE_TIME` tries to create 2 unnecessary `DateTime` objects, before creating an `Interval` object. The current implementation creates an `Interval` directly using `long` via`Intervals#utc` and hence performs faster.

On the topic of using other data formats, seems like we can look for `DateTimes.ISO_DATE_OPTIONAL_TIME` to improve the deserialization, but the improvement is quite small.

```
Benchmark                                                                 (numValues)  Mode  Cnt   Score   Error  Units
JodaIntervalDeserializationBenchmark.deserializeLegacy                          20000  avgt   50  20.858 ± 0.035  ms/op
JodaIntervalDeserializationBenchmark.deserializeOptimized                       20000  avgt   50  11.940 ± 0.009  ms/op
JodaIntervalDeserializationBenchmark.deserializeDateTime                        20000  avgt   50  12.524 ± 0.025  ms/op
JodaIntervalDeserializationBenchmark.deserializeDateTimeOffset                  20000  avgt   50  23.505 ± 0.050  ms/op
JodaIntervalDeserializationBenchmark.deserializeOptionalDateTime                20000  avgt   50  20.663 ± 0.111  ms/op

JodaIntervalDeserializationBenchmark.deserializeLegacyFallback                  20000  avgt   50  19.035 ± 0.032  ms/op
JodaIntervalDeserializationBenchmark.deserializeOptimizedFallback               20000  avgt   50  33.782 ± 0.105  ms/op
JodaIntervalDeserializationBenchmark.deserializeDateTimeFallback                20000  avgt   50  36.642 ± 0.065  ms/op
JodaIntervalDeserializationBenchmark.deserializeDateTimeOffsetFallback          20000  avgt   50  22.067 ± 0.046  ms/op
JodaIntervalDeserializationBenchmark.deserializeOptionalDateTimeFallback        20000  avgt   50  18.861 ± 0.012  ms/op
```

kfaraz: Thanks a lot for trying these out, @GWphua ! 🙂

GWphua:
@kfaraz 

Here's the flame graph after the changes:
image

Now, the initialization of Interval is roughly equal to initialization of DataSegment.

PR #18478, opened 2025-09-04
Fix up metrics.md: OshiSysMonitor doesn't emit sys/storage/used metric

abhishekrb19:
Currently, OshiSysMonitor doesn't [emit](https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/java/util/metrics/OshiSysMonitor.java#L254) the `sys/storage/used` metric. Must probably be a copy-paste from the `SysMonitor` metrics.

Also, there's no such supervisor `ioConfig` called `windowPeriod`, so remove that reference from the `ingest/events/thrownAway` metric.

This PR has:

- [x] been self-reviewed.


abhishekrb19:
thanks for the quick review @kfaraz!

PR #17900, opened 2025-04-10
Add taskCountStart property to lag autoscaler

jtuglu-netflix:
Creates `taskCountStart` optional property which can specify the initial task count to start with. Allows users to be able to set a reasonable minimum without needing to worry about a spike in lag from a spec being resubmitted. If the scaler tends to track an average task count, this is where it can be specified in the config to avoid any excessive spikes during deployment.

### Description

Added `taskCountStart` optional property which can specify the initial task count for the supervisor to be submitted with. 

#### Release note

Added `taskCountStart` optional property to LagBasedAutoScaler which can specify the initial task count for the supervisor to be submitted with. 

<hr>

##### Key changed/added classes in this PR
* `extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorIOConfigTest.java`
* `indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java`
* `indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/AutoScalerConfig.java`
* `indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java`

<hr>

This PR has:

- [x] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [ ] been tested in a test Druid cluster.

PR #18480, opened 2025-09-04
Add stopTaskCountPercent field to LagBasedAutoscaler

jtuglu1:
### Description

Perf improvement that allows the maximum number of stopped tasks to scale proportional to the task autoscaler. This is useful in cases where scaling up/down a lot can mess up the (over/under) # of tasks being marked for publishing.

Also adds some long-overdue isolated test files for `LagBasedAutoscalerConfig` and `SeekableStreamSupervisorIOConfig`.

Example:
- You have `taskCountStart=10`, `stopTaskCount=1`, but auto-scale up to `taskCount=100`. In this case, you may want to instead stop at most 10 tasks.

#### Release note

Support proportional stopTaskCount with task count auto-scaler

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/docs/ingestion/supervisor.md
+++ b/docs/ingestion/supervisor.md
@@ -80,7 +80,7 @@ The following table outlines the configuration properties for `autoScalerConfig`
 |`taskCountStart`|Optional config to specify the number of ingestion tasks to start with. When you enable the autoscaler, Druid ignores the value of `taskCount` in `ioConfig` and, if specified, starts with the `taskCountStart` number of tasks. Otherwise, defaults to `taskCountMin`.|No|`taskCountMin`|
 |`minTriggerScaleActionFrequencyMillis`|The minimum time interval between two scale actions.| No|600000|
 |`autoScalerStrategy`|The algorithm of autoscaler. Druid only supports the `lagBased` strategy. See [Autoscaler strategy](#autoscaler-strategy) for more information.|No|`lagBased`|
-|`stopTaskCountPercent`|A variable version of `ioConfig.stopTaskCount`. Allows the maximum number of stoppable tasks in steady state to be proportional to the # of tasks running.|No||
+|`stopTaskCountRatio`|A variable version of `ioConfig.stopTaskCount` with a valid range of (0.0, 1.0]. Allows the maximum number of stoppable tasks in steady state to be proportional to the number of tasks currently running.|No||
 
 ##### Autoscaler strategy
 
--- a/docs/ingestion/supervisor.md
+++ b/docs/ingestion/supervisor.md
@@ -80,6 +80,7 @@ The following table outlines the configuration properties for `autoScalerConfig`
 |`taskCountStart`|Optional config to specify the number of ingestion tasks to start with. When you enable the autoscaler, Druid ignores the value of `taskCount` in `ioConfig` and, if specified, starts with the `taskCountStart` number of tasks. Otherwise, defaults to `taskCountMin`.|No|`taskCountMin`|
 |`minTriggerScaleActionFrequencyMillis`|The minimum time interval between two scale actions.| No|600000|
 |`autoScalerStrategy`|The algorithm of autoscaler. Druid only supports the `lagBased` strategy. See [Autoscaler strategy](#autoscaler-strategy) for more information.|No|`lagBased`|
+|`stopTaskCountPercent`|A variable version of `ioConfig.stopTaskCount`. Allows the maximum number of stoppable tasks in steady state to be proportional to the # of tasks running.|No||
 
 ##### Autoscaler strategy
 
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
@@ -22,6 +22,7 @@ package org.apache.druid.indexing.seekablestream.supervisor.autoscaler;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Preconditions;
 import org.apache.druid.indexing.overlord.supervisor.Supervisor;
 import org.apache.druid.indexing.overlord.supervisor.SupervisorSpec;
 import org.apache.druid.indexing.overlord.supervisor.autoscaler.AggregateFunction;
@@ -102,6 +103,8 @@ public class LagBasedAutoScalerConfig implements AutoScalerConfig
     this.scaleOutStep = scaleOutStep != null ? scaleOutStep : 2;
     this.minTriggerScaleActionFrequencyMillis = minTriggerScaleActionFrequencyMillis
         != null ? minTriggerScaleActionFrequencyMillis : 600000;
+
+    Preconditions.checkArgument(stopTaskCountPercent == null || (stopTaskCountPercent > 0.0 && stopTaskCountPercent <= 1.0), "0.0 < stopTaskCountPercent <= 1.0");
     this.stopTaskCountPercent = stopTaskCountPercent;
   }
 
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
@@ -31,7 +31,9 @@ import org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervi
 import org.apache.druid.java.util.emitter.service.ServiceEmitter;
 
 import javax.annotation.Nullable;
+import java.util.Objects;
 
+@JsonInclude(JsonInclude.Include.NON_NULL)
 public class LagBasedAutoScalerConfig implements AutoScalerConfig
 {
   private final long lagCollectionIntervalMillis;
@@ -44,13 +46,13 @@ public class LagBasedAutoScalerConfig implements AutoScalerConfig
   private final double triggerScaleInFractionThreshold;
   private int taskCountMax;
   private int taskCountMin;
-  @Nullable private Integer taskCountStart;
+  private Integer taskCountStart;
   private final int scaleInStep;
   private final int scaleOutStep;
   private final boolean enableTaskAutoScaler;
   private final long minTriggerScaleActionFrequencyMillis;
   private final AggregateFunction lagAggregate;
-  @Nullable private Double stopTaskCountPercent;
+  private final Double stopTaskCountRatio;
 
   @JsonCreator
   public LagBasedAutoScalerConfig(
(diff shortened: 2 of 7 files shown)


docs/ingestion/supervisor.md
```diff
@@ -80,6 +80,7 @@ The following table outlines the configuration properties for `autoScalerConfig`
 |`taskCountStart`|Optional config to specify the number of ingestion tasks to start with. When you enable the autoscaler, Druid ignores the value of `taskCount` in `ioConfig` and, if specified, starts with the `taskCountStart` number of tasks. Otherwise, defaults to `taskCountMin`.|No|`taskCountMin`|
 |`minTriggerScaleActionFrequencyMillis`|The minimum time interval between two scale actions.| No|600000|
 |`autoScalerStrategy`|The algorithm of autoscaler. Druid only supports the `lagBased` strategy. See [Autoscaler strategy](#autoscaler-strategy) for more information.|No|`lagBased`|
+|`stopTaskCountPercent`|A variable version of `ioConfig.stopTaskCount`. Allows the maximum number of stoppable tasks in steady state to be proportional to the # of tasks running.|No||
```
kfaraz:
```suggestion
|`stopTaskCountPercent`|A variable version of `ioConfig.stopTaskCount`. Allows the maximum number of stoppable tasks in steady state to be proportional to the number of tasks currently running.|No||
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
```diff
@@ -49,6 +50,7 @@ public class LagBasedAutoScalerConfig implements AutoScalerConfig
   private final boolean enableTaskAutoScaler;
   private final long minTriggerScaleActionFrequencyMillis;
   private final AggregateFunction lagAggregate;
+  @Nullable private Double stopTaskCountPercent;
```
kfaraz: Nit: I don't think we need to mark this as nullable too. It should be enough to mark just the getter as nullable.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
```diff
@@ -100,6 +103,9 @@ public LagBasedAutoScalerConfig(
     this.scaleOutStep = scaleOutStep != null ? scaleOutStep : 2;
     this.minTriggerScaleActionFrequencyMillis = minTriggerScaleActionFrequencyMillis
         != null ? minTriggerScaleActionFrequencyMillis : 600000;
+
+    Preconditions.checkArgument(stopTaskCountPercent == null || (stopTaskCountPercent > 0.0 && stopTaskCountPercent <= 1.0), "0.0 < stopTaskCountPercent <= 1.0");
```
kfaraz: If the value range is [0, 1], we should call the field `stopTaskCountRatio`.
Alternatively, if we want to keep a percentage in the range [0, 100], we should use an int field instead.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
```diff
@@ -212,6 +218,15 @@ public AggregateFunction getLagAggregate()
     return lagAggregate;
   }
 
+  @Override
+  @JsonProperty
+  @Nullable
+  @JsonInclude(JsonInclude.Include.NON_NULL)
```
kfaraz: This can be moved to the class level itself.

indexing-service/src/test/java/org/apache/druid/indexing/overlord/supervisor/autoscaler/LagBasedAutoScalerConfigTest.java
```diff
@@ -0,0 +1,294 @@
...
+ */
+
+package org.apache.druid.indexing.overlord.supervisor.autoscaler;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.druid.indexing.seekablestream.supervisor.autoscaler.LagBasedAutoScalerConfig;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class LagBasedAutoScalerConfigTest
```
kfaraz, line 27: Thanks for adding these tests!

indexing-service/src/test/java/org/apache/druid/indexing/overlord/supervisor/autoscaler/LagBasedAutoScalerConfigTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+  @Test
+  public void testSerdeWithDefaults() throws Exception
+  {
+    LagBasedAutoScalerConfig config = OBJECT_MAPPER.readValue(
+        OBJECT_MAPPER.writeValueAsString(
+            OBJECT_MAPPER.readValue(
+                "{}",
+                LagBasedAutoScalerConfig.class
+            )
```
kfaraz: Just this part should be enough to verify the default values.
We need not perform a repeated deserialize-serialize-deserialize.

kfaraz: We can have one separate test which verifies that the object obtained on serialize and then deserialize is the same as the original.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
```diff
@@ -2689,6 +2690,55 @@ public Duration getEmissionDuration()
     replayAll();
   }
 
+  @Test
+  public void testMaxAllowedStopsWithStopTaskCountPercent()
+  {
+    LagAggregator lagAggregator = createMock(LagAggregator.class);
+    AutoScalerConfig autoScalerConfig = new LagBasedAutoScalerConfig(
```
kfaraz, line 2697: Looking at the various test cases, I think it might make sense to add a `Builder` for the `LagBasedAutoScalerConfig` class. Not a blocker for this PR, though.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
jtuglu1: Value range is (0, 1.0]. I don't really like the idea of using an int field for this kind of thing. I'll switch the name to ratio.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
jtuglu1: Hmmm. Are we sure we want to do that? This would exclude ALL fields that are `null`. In the case where there's some malformed input we may not include it in the spec at all (not visible to the user). I'd prefer to reserve this for fields which make sense to be nullable, and nullable fields tend to bloat the config specs.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
jtuglu1: I'm following a pattern @gianm had [suggested](https://github.com/apache/druid/pull/17900#discussion_r2047872048) a while back.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
kfaraz: > I'd prefer to reserve this for fields which make sense to be nullable

I don't quite follow. A non-nullable field would never be null anyway.
The constructor would either assign a default value to it or throw an exception.
For anyone reading the json, it doesn't make a real difference if the field is absent or if it is assigned `"field": null`.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
kfaraz: Yes, the pattern suggested in that comment makes sense.

I just meant we should drop the `@Nullable` from `@Nullable private Double stopTaskCountPercent`
since none of the other fields have it and it doesn't serve any real purpose.
It is important to have `@Nullable` in the public APIs like the getter and the constructor arg.

kfaraz: Also, should this field be `final`?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
jtuglu1: > I don't quite follow. A non-nullable field would never be null anyway.

Of course. I think I was just trying to point out that we should have a standard for leaving null (but optional) fields in a config, versus keeping them out.

> For anyone reading the json, it doesn't make a real difference if the field is absent or if it is assigned "field": null.

I think it does. I believe what Gian was implying was that for non-required fields, having null values in the config looks weird/adds bloat.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
jtuglu1: > I just meant we should drop the @Nullable from @Nullable private Double stopTaskCountPercent

I personally like to mark these things as `@Nullable` or `Optional<T>` as it makes it clear when working the raw value that the value could be null (not just return value of the accessor method). In that case `startTaskCount` annotation should also be changed. We should have a clear standard on these things (like [this](https://google.github.io/styleguide/javaguide.html#s4.8.5-annotations)) so style doesn't differ based on who is reviewing.

kfaraz: Yeah, that's fair. It would be nice to have a common standard across the codebase.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScalerConfig.java
kfaraz: > Of course. I think I was just trying to point out that we should have a standard for leaving null (but optional) fields in a config, versus keeping them out.

Yeah, that would be nice. I think this could be a config on the object mapper itself.

> I think it does.

Oh, I must be missing something then. The only fields that would have ever have a null value are nullable (i.e. non required) fields. Required fields should always be validated in the constructor or assigned a default value if null. So the serialized JSON will always have a non-null value for required fields. A malformed input should fail during deserializing itself.

That said, it is really fine either way. It is only a style choice. 🙂 
And it seems that in the entirety of Druid code, this annotation is never used on the class level.
So we can just stick to that for the time being.

docs/ingestion/supervisor.md
```diff
@@ -80,6 +80,7 @@ The following table outlines the configuration properties for `autoScalerConfig`
 |`taskCountStart`|Optional config to specify the number of ingestion tasks to start with. When you enable the autoscaler, Druid ignores the value of `taskCount` in `ioConfig` and, if specified, starts with the `taskCountStart` number of tasks. Otherwise, defaults to `taskCountMin`.|No|`taskCountMin`|
 |`minTriggerScaleActionFrequencyMillis`|The minimum time interval between two scale actions.| No|600000|
 |`autoScalerStrategy`|The algorithm of autoscaler. Druid only supports the `lagBased` strategy. See [Autoscaler strategy](#autoscaler-strategy) for more information.|No|`lagBased`|
+|`stopTaskCountRatio`|A variable version of `ioConfig.stopTaskCount` with a valid range of (0.0, 1.0]. Allows the maximum number of stoppable tasks in steady state to be proportional to the number of tasks currently running.|No||
```
maytasm, line 83: nit: might worth mentioning that when stopTaskCountRatio is set, it takes precedence over stopTaskCount

PR #18486, opened 2025-09-05
fix projection cursor build spec interval matching to consider projection granularity

clintropolis:
### Description
Fixes projection matching to consider the interval specified on the `CursorBuildSpec` with the projection granularity used for the __time column. Prior to this patch, a projection match could produce incorrect results if the `CursorBuildSpec` specified a partially covered interval of the segment interval and that interval did not align with projection __time column granularity boundaries. This PR fixes this by ensuring that the interval defined on `CursorBuildSpec` align with the projection granularity if the query interval does not cover the data interval.


processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
```diff
@@ -220,6 +231,219 @@ void testSchemaMatchFilterIncludedInProjection()
...
+
+    projectionMatch = Projections.matchAggregateProjection(
+        spec.getSchema(),
+        cursorBuildSpecPartialInterval,
+        day,
+        new RowSignatureChecker(baseTable)
+    );
+    Assertions.assertNull(projectionMatch);
+
+    // interval larger than data interval can match, though this shouldn't really happen in practice
```
cecemei: not sure i follow, why would it not happen in practice? i thought the idea is that if we have daily segments, and query on partial+day1+day2+partial, we can still use projection on day1 and day2.

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -117,6 +121,11 @@ public static ProjectionMatch matchAggregateProjection(
     if (CollectionUtils.isNullOrEmpty(queryCursorBuildSpec.getPhysicalColumns())) {
       return null;
     }
+
+
+    if (isUnalignedInterval(projection, queryCursorBuildSpec, dataInterval)) {
```
cecemei, line 125: nit: additional line break.

also in line 382, it only checks for virtual column that only have __time as required input, but there can be multiple required inputs and __time being one of them, e.x. concat(column1, __time), in this case, granularity is not checked. maybe we should update the condition to be `requiredInputs.contains("__time")`

processing/src/test/java/org/apache/druid/segment/projections/ProjectionsTest.java
clintropolis: oh, i'll remove this comment because is confusing. i just meant that having a larger interval specifically like this where it is bigger than the segment interval, but not eternity will not likely happen much in practice. The broker wil issue queries broken down into which specific segments are involved with either the full interval of the segment specified or a partial if only partially covered. The only larger interval that is likely to happen in production code is eternity.

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
clintropolis: >also in line 382, it only checks for virtual column that only have __time as required input, but there can be multiple required inputs and __time being one of them, e.x. concat(column1, __time), in this case, granularity is not checked. maybe we should update the condition to be requiredInputs.contains("__time")

line 381 is checking that there is only 1 required input and that it is __time, since only time_floor can be considered as a granularity, if it requires __time but has only 1 input it requires none granularity. The else clause handles all other expressions, such as the concat example by just making sure all the input columns required by the expression are matched in the projection

cecemei: missed that it should be caught by `PhysicalColumnChecker`, so all good

cecemei (approved):

PR #18487, opened 2025-09-05
Add a new strategy to sanitize error messages based on the error message role

adarshsanjeev:
Add a new error message strategy to log the message and return an error id to the user, based on the role of the Druid exception.

To enable this strategy, set `druid.server.http.errorResponseTransform.strategy` to `persona`.

<h3>Patchnotes</h3>
- Add a new error message strategy to log the message and return an error id for non-user targeted messages. This can be configured by setting `druid.server.http.errorResponseTransform.strategy` to `persona`.

<hr>

This PR has:

- [ ] been self-reviewed.


processing/src/main/java/org/apache/druid/common/exception/ErrorResponseTransformStrategy.java
```diff
@@ -28,7 +28,8 @@
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "strategy", defaultImpl = NoErrorResponseTransformStrategy.class)
 @JsonSubTypes(value = {
     @JsonSubTypes.Type(name = "none", value = NoErrorResponseTransformStrategy.class),
-    @JsonSubTypes.Type(name = "allowedRegex", value = AllowedRegexErrorResponseTransformStrategy.class)
+    @JsonSubTypes.Type(name = "allowedRegex", value = AllowedRegexErrorResponseTransformStrategy.class),
+    @JsonSubTypes.Type(name = "roleBased", value = PersonaBasedErrorTransformStrategy.class),
```
abhishekagarwal87:
```suggestion
    @JsonSubTypes.Type(name = "persona", value = PersonaBasedErrorTransformStrategy.class),
```

processing/src/main/java/org/apache/druid/common/exception/PersonaBasedErrorTransformStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.apache.druid.error.DruidException;
+import org.apache.druid.java.util.common.StringUtils;
+import org.apache.druid.java.util.emitter.EmittingLogger;
+import org.apache.druid.java.util.emitter.service.AlertEvent;
+
+import java.util.UUID;
+import java.util.function.Function;
+
+public class PersonaBasedErrorTransformStrategy implements ErrorResponseTransformStrategy
```
abhishekagarwal87, line 34: please add some javadocs

processing/src/main/java/org/apache/druid/error/DruidException.java
```diff
@@ -129,7 +131,7 @@
  * introduced DruidException.
  */
 @NotThreadSafe
-public class DruidException extends RuntimeException
+public class DruidException extends RuntimeException implements SanitizableException
```
abhishekagarwal87: lets revert the changes here. I don't think the whole `SanitizableException` was well thought out and don't want DruidException to extend it. In an ideal world, we shouldn't have SanitizableException and just use DruidException

processing/src/main/java/org/apache/druid/common/exception/PersonaBasedErrorTransformStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+        return druidException;
+      } else {
+        return exception.sanitize(s -> {
+          final String errorId = UUID.randomUUID().toString();
+          LOG.makeAlert(druidException, StringUtils.format("Error ID: [%s]", errorId))
+             .addData(druidException.getContext())
+             .severity(AlertEvent.Severity.ANOMALY)
+             .emit();
+          return StringUtils.format(ERROR_WITH_ID_TEMPLATE, errorId);
+        });
```
abhishekagarwal87: instead of calling this, lets add a new function that generates a new error message. This method takes an exception and returns a string (not an exception). The returned string then is returned to the end user. The method should also take in an optional error id. Query endpoints can set this error id to the query id itself.

abhishekagarwal87: The returned method should be `Optional<String>` 
 if empty - a transformation occurred
 if not - no transformation 

caller can decide based on the return type, what to do.

abhishekagarwal87: though writing a string instead of a json for the end-user will start to break clients of query/sql API.

abhishekagarwal87: Instead of sanitizing the DruidException, we could just build a new one with user persona and the transformed error message and return that to the client. We should change the persona because the new error is safe to be sent to the end user. So the method will look like this 

```
returns Optional.empty - no transformation was applied
optional <exception> - transformation was applied
Optional<Exception> maybeTransform(Exception originalException, Optional<String> errorId) 
```
callers can check if transformation was applied and want to log the original exception too.

processing/src/main/java/org/apache/druid/error/DruidException.java
adarshsanjeev: I fully agree, but `SanitizableException` and `ErrorResponseTransformStrategy` are coupled. To work around this without having DruidException , I've had to create a new function in `ErrorResponseTransformStrategy`. Let me know if this is fine.

processing/src/main/java/org/apache/druid/common/exception/PersonaBasedErrorTransformStrategy.java
adarshsanjeev: I've attempted to make changes according to what you suggested, please take a look. I did not understand why we would need a `Optional<String> errorId` instead of always passing one.

cryptoe:
Could you please add the config change required in the PR description to enable this.

sql/src/main/java/org/apache/druid/sql/avatica/ErrorHandler.java
```diff
@@ -73,12 +74,18 @@ public <T extends Throwable> RuntimeException sanitize(T error)
     if (error instanceof SanitizableException) {
       return new RuntimeException(errorResponseTransformStrategy.transformIfNeeded((SanitizableException) error));
     }
+    if (error instanceof DruidException) {
+    }
```
abhishekagarwal87: whats this?

adarshsanjeev: This has been removed

PR #18489, opened 2025-09-05
Concurrent loading segment cache

GWphua:
#### Concurrent loading of cached segments during startup

Previously, we are using 1 thread to run `SegmentLocalCacheManager#getCachedSegments`. We may take quite a bit of time if the number of segments is large. 

We can do better by multithreading using `loadOnBootstrapExec`, and speed up segment retrieval. 

#### Refactoring
- Create segmentsToLoad array via extracted method: `retrieveSegmentMetadataFiles`.
- Extract logic to load segments into `addFilesToCachedSegments`

#### Log Changes
- Add stopwatch to report time taken to load cached segments.
- Removed log "Loading segment cache file [%d/%d][%s]." to remove log clutter, stopwatch will help to handle this better. (It was annoying to navigate ~100k lines of this when looking through logs)

#### Release note

Speed up load of cached segments during Historical startup.

<hr>

 * `SegmentLocalCacheManager`
 * `SegmentCacheManager` + Relevant implementations.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
+++ b/server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
@@ -232,8 +232,8 @@ public class SegmentLocalCacheManager implements SegmentCacheManager
     final CountDownLatch latch = new CountDownLatch(segmentsToLoad.length);
 
     // If there is no dedicated bootstrap executor, perform the loading sequentially on the current thread.
-    final boolean isLoadingSegmentCacheFromDiskWithoutBootstrapExecutor = loadOnBootstrapExec == null;
-    final ExecutorService executorService = isLoadingSegmentCacheFromDiskWithoutBootstrapExecutor
+    final boolean isLoadingSegmentsSequentially = loadOnBootstrapExec == null;
+    final ExecutorService executorService = isLoadingSegmentsSequentially
                                             ? MoreExecutors.newDirectExecutorService()
                                             : loadOnBootstrapExec;
 
@@ -269,7 +269,8 @@ public class SegmentLocalCacheManager implements SegmentCacheManager
     stopwatch.stop();
     log.info("Loaded [%d/%d] cached segments in [%d]ms.", cachedSegments.size(), segmentsToLoad.length, stopwatch.millisElapsed());
 
-    if (isLoadingSegmentCacheFromDiskWithoutBootstrapExecutor) {
+    if (isLoadingSegmentsSequentially) {
+      // Shutdown the direct executor service we created previously in this method.
       executorService.shutdown();
     }
 


server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -181,90 +185,120 @@ public boolean canHandleSegments()
...
+         .addData("numIgnored", ignoredfileCounter.get())
+         .emit();
+    }
+
+    return cachedSegments;
+  }
+
+  private File[] retrieveSegmentMetadataFiles() throws IOException
+  {
+    File baseDir = config.getInfoDir();
```
kfaraz: The original code used `getEffectiveInfoDir()`.

GWphua: Sure, I will make changes that makes it easy to review, and leave refactoring for a future PR :)

GWphua:
Hi @kfaraz, I have tried to changes, but the need to add a ExecutorService and use `exec.submit()` introduces another layer of indentation -- meaning the reviewing process will still be tedious.

I have created #18494 to help with reviewing this change, please take a look at that. I hope that after merging #18494, diff changes in this PR will be clearer. Thanks!

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -181,90 +185,120 @@ public boolean canHandleSegments()
...
-              final File[] newFiles = destination.listFiles();
-              // make sure old files exist and new files do not exist
-              if (oldFiles != null && oldFiles.length > 0 && newFiles != null && newFiles.length == 0) {
-                Files.move(legacyPath.toPath(), destination.toPath(), StandardCopyOption.ATOMIC_MOVE);
-              }
-              cleanupLegacyCacheLocation(location.getPath(), legacyPath);
-            }
+    AtomicInteger ignoredfileCounter = new AtomicInteger(0);
+    CountDownLatch latch = new CountDownLatch(segmentsToLoad.length);
+    ExecutorService exec = Objects.requireNonNullElseGet(loadOnBootstrapExec, MoreExecutors::newDirectExecutorService);
```
clintropolis: drive by comment since i haven't had a chance to do a full review yet:

given that the actual load of these segments is already happening concurrently on a pool (by the thing that calls `getCachedSegments`), I'm wondering if instead of putting deserializing these info files on a thread pool to instead combine reading the info file and mounting the segment into a single thread, instead of threads to load all the infos then more threads to do all the mounts.

That said, it is a bit more involved of a refactor, because `SegmentCacheBootstrapper` is what currently collects all the metadata today since it blends both the stuff on disk and the separate 'bootstrap' segments to then put onto the pool. Need to think a bit more about what that would look like.

GWphua: If I'm understanding you correctly, then the flow will be something like:

1. Put getBootstrapSegments() into array.
2. Retrieve `File[] segmentsToLoad` from `getEffectiveInfoDir()`
3. Parallelization, each thread does the following:
  - Load N cached segments
  - loadSegmentOnBootstrap for N cached segments
  - Announce the N cached segments.
5. Some of the threads that finished working can load the bootstrap segments. (This can happen anytime after (1))

This will indeed involve some refactoring... But I guess we can have the above steps? Let me know if i am missing anything.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -190,22 +193,47 @@ public List<DataSegment> getCachedSegments() throws IOException
...
-      log.info("Loading segment cache file [%d/%d][%s].", i + 1, segmentsToLoad.length, file);
-      try {
-        addFilesToCachedSegments(file, ignoredFilesCounter, cachedSegments);
-      }
-      catch (Exception e) {
-        log.makeAlert(e, "Failed to load segment from segment cache file.")
-           .addData("file", file)
-           .emit();
-      }
+    boolean createdNewExecutorServiceToLoadSegmentCache = loadOnBootstrapExec == null;
```
abhishekrb19: nit: something like `shouldShutdownExecutor`?

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -190,22 +193,47 @@ public List<DataSegment> getCachedSegments() throws IOException
...
-    final List<DataSegment> cachedSegments = new ArrayList<>();
     final File[] segmentsToLoad = retrieveSegmentMetadataFiles();
-
+    final ConcurrentLinkedQueue<DataSegment> cachedSegments = new ConcurrentLinkedQueue<>();
     AtomicInteger ignoredFilesCounter = new AtomicInteger(0);
+    CountDownLatch latch = new CountDownLatch(segmentsToLoad.length);
 
-    for (int i = 0; i < segmentsToLoad.length; i++) {
-      final File file = segmentsToLoad[i];
-      log.info("Loading segment cache file [%d/%d][%s].", i + 1, segmentsToLoad.length, file);
```
abhishekrb19: Curious, why was this info log removed? When an operator searches for a segment ID, it'd show the entire lifecycle of the segment - from loading on a data server to being dropped, so IMO this log line is useful to keep since it’s emitted only once and helps track progress as well. If it’s not of interest, we could always configure log4j to filter out these log lines instead.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -190,22 +193,47 @@ public List<DataSegment> getCachedSegments() throws IOException
...
+    try {
+      latch.await();
+    }
+    catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      log.makeAlert(e, "Interrupted when trying to retrieve cached segment metadata files");
+    }
+
+    stopwatch.stop();
+    log.info("Retrieved [%d,%d] cached segments in [%d]ms.", cachedSegments.size(), segmentsToLoad.length, stopwatch.millisElapsed());
```
abhishekrb19:
```suggestion
    log.info("Retrieved [%d]/[%d] cached segments in [%d]ms.", cachedSegments.size(), segmentsToLoad.length, stopwatch.millisElapsed());
```

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -190,22 +193,47 @@ public List<DataSegment> getCachedSegments() throws IOException
...
+        try {
+          loadToCachedSegmentsFromFile(cachedSegments, file, ignoredFilesCounter);
+        }
+        catch (Exception e) {
+          log.makeAlert(e, "Failed to load segment from segment cache file.")
+             .addData("file", file)
+             .emit();
+        }
+
+        latch.countDown();
```
abhishekrb19: put this in a `finally` block to ensure this is guaranteed to run.

FrankChen021: put in the finally block

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -190,22 +193,47 @@ public List<DataSegment> getCachedSegments() throws IOException
...
+      });
+    }
+
+    try {
+      latch.await();
+    }
+    catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      log.makeAlert(e, "Interrupted when trying to retrieve cached segment metadata files");
```
FrankChen021: I think there's no need to send an alert for this exception

GWphua:
Hi @kfaraz, hopefully the changes now are more reviewer-friendly. Can take a look again. Thanks!

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -225,22 +228,49 @@ public List<DataSegment> getCachedSegments() throws IOException
...
+        }
+      });
+    }
+
+    try {
+      latch.await();
+    }
+    catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      log.error(e, "Interrupted when trying to retrieve cached segment metadata files");
```
kfaraz: Since it was interrupted, I think we can skip logging the stack trace.
```suggestion
      log.noStackTrace().error(e, "Interrupted when trying to retrieve cached segment metadata files");
```

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -225,22 +228,49 @@ public List<DataSegment> getCachedSegments() throws IOException
...
+    }
+
+    stopwatch.stop();
+    log.info("Retrieved [%d,%d] cached segments in [%d]ms.", cachedSegments.size(), segmentsToLoad.length, stopwatch.millisElapsed());
```
kfaraz:
```suggestion
    log.info("Loaded [%d/%d] cached segments in [%d]ms.", cachedSegments.size(), segmentsToLoad.length, stopwatch.millisElapsed());
```

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -225,22 +228,49 @@ public List<DataSegment> getCachedSegments() throws IOException
       }
     }
 
-    final List<DataSegment> cachedSegments = new ArrayList<>();
+    final ConcurrentLinkedQueue<DataSegment> cachedSegments = new ConcurrentLinkedQueue<>();
     final File[] segmentsToLoad = retrieveSegmentMetadataFiles();
+    CountDownLatch latch = new CountDownLatch(segmentsToLoad.length);
+
+    boolean createdNewExecutorServiceToLoadSegmentCache = loadOnBootstrapExec == null;
+    ExecutorService executorService = createdNewExecutorServiceToLoadSegmentCache
```
kfaraz: Nit: Making these variables `final` would be nice.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -225,22 +228,49 @@ public List<DataSegment> getCachedSegments() throws IOException
...
-      try {
-        addFilesToCachedSegments(file, ignoredFilesCounter, cachedSegments);
-      }
-      catch (Exception e) {
-        log.makeAlert(e, "Failed to load segment from segment cache file.")
-           .addData("file", file)
-           .emit();
-      }
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    log.info("Retrieving [%d] cached segment metadata files to cache.", segmentsToLoad.length);
```
kfaraz:
```suggestion
    log.info("Loading [%d] segments from disk to cache.", segmentsToLoad.length);
```

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -225,22 +228,49 @@ public List<DataSegment> getCachedSegments() throws IOException
       }
     }
 
-    final List<DataSegment> cachedSegments = new ArrayList<>();
+    final ConcurrentLinkedQueue<DataSegment> cachedSegments = new ConcurrentLinkedQueue<>();
     final File[] segmentsToLoad = retrieveSegmentMetadataFiles();
+    CountDownLatch latch = new CountDownLatch(segmentsToLoad.length);
+
+    boolean createdNewExecutorServiceToLoadSegmentCache = loadOnBootstrapExec == null;
```
kfaraz:
```suggestion
    // If there is no dedicated bootstrap executor, perform the loading sequentially on the current thread
    boolean createdNewExecutorServiceToLoadSegmentCache = loadOnBootstrapExec == null;
```

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -249,10 +279,14 @@ public List<DataSegment> getCachedSegments() throws IOException
          .emit();
     }
 
-    return cachedSegments;
+    return new ArrayList<>(cachedSegments);
```
kfaraz: Return an immutable list:
```suggestion
    return List.copyOf(cachedSegments);
```

kfaraz (approved):
Thanks, @GWphua !
I have left minor suggestions, otherwise looks good.

GWphua:
Thanks @kfaraz for the review! Comments are resolved, feel free to take a look again. :)

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -225,22 +227,50 @@ public List<DataSegment> getCachedSegments() throws IOException
       }
     }
 
-    final List<DataSegment> cachedSegments = new ArrayList<>();
+    final ConcurrentLinkedQueue<DataSegment> cachedSegments = new ConcurrentLinkedQueue<>();
     final File[] segmentsToLoad = retrieveSegmentMetadataFiles();
+    final CountDownLatch latch = new CountDownLatch(segmentsToLoad.length);
+
+    // If there is no dedicated bootstrap executor, perform the loading sequentially on the current thread.
+    final boolean isLoadingSegmentCacheFromDiskWithoutBootstrapExecutor = loadOnBootstrapExec == null;
```
kfaraz: Nit: seems a little verbose 🙂 , especially since it is just to be used in this method.

You may call this variable `loadSegmentsSequentially` or `useBootstrapLoadExec` (with the inverse condition).

GWphua: Sounds good, I renamed it to isLoadingSegmentsSequentially. Thanks!

kfaraz: Thanks!

GWphua:
Hi @clintropolis are you still looking at this PR? 

If there are no further comments, I will be merging this PR soon (maybe by next Friday). Thanks!

GWphua:
Thanks all for the reviews!

I will merge this in first, if we need to do further refactoring, new PR's can be created.

PR #18493, opened 2025-09-07
Adjust formula for Netty worker threads.

gianm:
Generally, the new formula leads to much fewer threads at small processor counts, and similar numbers of threads at large processor counts. The purpose of this change is to avoid issues due to having too-large worker thread counts on small machines, such as OutOfMemoryError for too many threads or not enough direct memory.

Comparison table

| cpus | old | new |
|---|-----|-----|
| 1 | 40 | 2 |
| 2 | 40 | 4 |
| 3 | 40 | 5 |
| 4 | 40 | 7 |
| 5 | 40 | 8 |
| 6 | 40 | 10 |
| 7 | 40 | 11 |
| 8 | 40 | 13 |
| 16 | 49 | 25 |
| 24 | 57 | 37 |
| 32 | 66 | 49 |
| 48 | 83 | 73 |
| 64 | 100 | 97 |
| 96 | 134 | 145 |

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -702,7 +702,7 @@ All Druid components can communicate with each other over HTTP.
 |`druid.global.http.compressionCodec`|Compression codec to communicate with others. May be "gzip" or "identity".|`gzip`|
 |`druid.global.http.readTimeout`|The timeout for data reads.|`PT15M`|
 |`druid.global.http.unusedConnectionTimeout`|The timeout for idle connections in connection pool. The connection in the pool will be closed after this timeout and a new one will be established. This timeout should be less than `druid.global.http.readTimeout`. Set this timeout = ~90% of `druid.global.http.readTimeout`|`PT4M`|
-|`druid.global.http.numMaxThreads`|Maximum number of I/O worker threads|`max(10, ((number of cores * 17) / 16 + 2) + 30)`|
+|`druid.global.http.numMaxThreads`|Maximum number of I/O worker threads|`(number of cores) * 3 / 2 + 1`|
 |`druid.global.http.clientConnectTimeout`|The timeout (in milliseconds) for establishing client connections.|500|
 
 ### Common endpoints configuration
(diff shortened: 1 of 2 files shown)


kfaraz:
@gianm , it makes sense to reduce the number of netty worker threads since these
threads mostly sit idle most of the time.

But I wonder if the change is not too drastic, especially since this affects all services.
e.g. even for a machine with 16 cpus, we are halving the value from 49 to 25,
which can affect say an Overlord communicating with a large number of streaming tasks.

Do you think it would make sense to maybe roll this out gradually, like have a formula which reduces
the 49 to say 40-ish and decrease the values further in subsequent Druid releases?

gianm:
> @gianm , it makes sense to reduce the number of netty worker threads since these threads mostly sit idle most of the time.
> 
> But I wonder if the change is not too drastic, especially since this affects all services. e.g. even for a machine with 16 cpus, we are halving the value from 49 to 25, which can affect say an Overlord communicating with a large number of streaming tasks.
> 
> Do you think it would make sense to maybe roll this out gradually, like have a formula which reduces the 49 to say 40-ish and decrease the values further in subsequent Druid releases?

I believe the Netty worker threads are only supposed to be running callbacks, so they should need to scale with number of CPUs rather than number of outgoing connections. For that reason I am not expecting the smaller pools to be a bottleneck & I would rather jump straight to a new formula.

Maybe we can validate this with some scale testing prior to release?

kfaraz:
> I believe the Netty worker threads are only supposed to be running callbacks, so they should need to scale with number of CPUs rather than number of outgoing connections. For that reason I am not expecting the smaller pools to be a bottleneck & I would rather jump straight to a new formula.

Thanks for the clarification, @gianm .

> Maybe we can validate this with some scale testing prior to release?

Sure, I think this should be feasible.

FrankChen021:
image

documents should be updated to align this change.

uds5501:
Hey @gianm , we are testing this change in a cluster and will bombard the same with high load. We shall post the results once done.

gianm:
> documents should be updated to align this change.

Good point. Updated.

docs/configuration/index.md
```diff
@@ -702,7 +702,7 @@ All Druid components can communicate with each other over HTTP.
 |`druid.global.http.compressionCodec`|Compression codec to communicate with others. May be "gzip" or "identity".|`gzip`|
 |`druid.global.http.readTimeout`|The timeout for data reads.|`PT15M`|
 |`druid.global.http.unusedConnectionTimeout`|The timeout for idle connections in connection pool. The connection in the pool will be closed after this timeout and a new one will be established. This timeout should be less than `druid.global.http.readTimeout`. Set this timeout = ~90% of `druid.global.http.readTimeout`|`PT4M`|
-|`druid.global.http.numMaxThreads`|Maximum number of I/O worker threads|`max(10, ((number of cores * 17) / 16 + 2) + 30)`|
```
SatyaKuppam, line 705: Curious, how did we come up with these old formulas?

uds5501:
Hi @gianm , the local tests look good with these I/O worker threads setup (the supervisors for ingestion looks stable and are being queried just fine)

kfaraz:
Thanks for performing the verifications, @uds5501 !
Could you share some details of the kind of load you used in your testing?

docs/configuration/index.md
kfaraz: @gianm , could you please add a short comment in the code regarding the heuristics behind the formula?

PR #18495, opened 2025-09-08
Disable flaky ingestion test

uds5501:
There have been reports of flakiness for `test_runIndexTask_andKillData` where the broker doesn't reflect the state of the unused segments after the kill command.

A follow up PR will be raised after this that will :
1. Emit a metric once the data source is marked for dropoping
2. This embedded test waits for that event to show up before attempting to run this flaky query.


embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.java
```diff
@@ -149,6 +150,7 @@ protected int markSegmentsAsUnused(String dataSource)
   }
 
   @Test
+  @Disabled("Disabled due to flakiness after segment drops")
```
kfaraz: We should disable it only for the docker backward compatibility test where we are seeing the flakiness.

PR #18497, opened 2025-09-08
Emit event when a datasource has been dropped from the coordinator

uds5501:
Follow up PR for https://github.com/apache/druid/pull/18495

It introduces a new event `segment/schemaCache/datasource/dropped/count` that is emitted when the data source is dropped from the coordinator and the cache refresh in broker acknowledges it.

### DIFF
--- a/docs/operations/metrics.md
+++ b/docs/operations/metrics.md
@@ -441,6 +441,7 @@ These metrics are emitted by the Druid Coordinator in every run of the correspon
 |`serverview/sync/unstableTime`|Time in milliseconds for which the Coordinator has been failing to sync with a segment-loading server. Emitted only when [HTTP-based server view](../configuration/index.md#segment-management) is enabled.|`server`, `tier`|Not emitted for synced servers.|
 |`metadatacache/init/time`|Time taken to initialize the coordinator segment metadata cache.||Depends on the number of segments.|
 |`segment/schemaCache/refresh/count`|Number of segments for which schema was refreshed in coordinator segment schema cache.|`dataSource`||
+|`segment/schemaCache/datasource/dropped/count`|Number of datasources dropped from broker cache after segments were marked unused.|`dataSource`||
 |`segment/schemaCache/refresh/time`|Time taken to refresh segments in coordinator segment schema cache.|`dataSource`||
 |`segment/schemaCache/backfill/count`|Number of segments for which schema was back filled in the database.|`dataSource`||
 |`segment/schemaCache/realtime/count`|Number of realtime segments for which schema is cached.||Depends on the number of realtime segments in the cluster.|
--- a/docs/operations/metrics.md
+++ b/docs/operations/metrics.md
@@ -441,7 +441,7 @@ These metrics are emitted by the Druid Coordinator in every run of the correspon
 |`serverview/sync/unstableTime`|Time in milliseconds for which the Coordinator has been failing to sync with a segment-loading server. Emitted only when [HTTP-based server view](../configuration/index.md#segment-management) is enabled.|`server`, `tier`|Not emitted for synced servers.|
 |`metadatacache/init/time`|Time taken to initialize the coordinator segment metadata cache.||Depends on the number of segments.|
 |`segment/schemaCache/refresh/count`|Number of segments for which schema was refreshed in coordinator segment schema cache.|`dataSource`||
-|`segment/schemaCache/datasource/dropped/count`|Number of datasources dropped from broker cache after segments were marked unused.|`dataSource`||
+|`segment/schemaCache/dataSource/removed`|Emitted when a datasource is removed from the Broker cache due to segments being marked as unused.|`dataSource`||
 |`segment/schemaCache/refresh/time`|Time taken to refresh segments in coordinator segment schema cache.|`dataSource`||
 |`segment/schemaCache/backfill/count`|Number of segments for which schema was back filled in the database.|`dataSource`||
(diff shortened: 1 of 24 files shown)


sql/src/main/java/org/apache/druid/sql/calcite/schema/BrokerSegmentMetadataCache.java
```diff
@@ -243,13 +246,14 @@ public void refresh(final Set<SegmentId> segmentsToRefresh, final Set<String> da
...
     }
-
+    List<String> droppedDataSources = new ArrayList<>();
     // Rebuild the datasources.
     for (String dataSource : dataSourcesToRebuild) {
       final RowSignature rowSignature = buildDataSourceRowSignature(dataSource);
       if (rowSignature == null) {
         log.info("datasource [%s] no longer exists, all metadata removed.", dataSource);
         tables.remove(dataSource);
+        droppedDataSources.add(dataSource);
```
kfaraz: Rather than adding to the list, let's just emit the event here.

server/src/main/java/org/apache/druid/segment/metadata/Metric.java
```diff
@@ -49,6 +49,7 @@ public class Metric
   public static final String STARTUP_DURATION_MILLIS = "metadatacache/init/time";
   public static final String REFRESHED_SEGMENTS = PREFIX + "refresh/count";
   public static final String REFRESH_DURATION_MILLIS = PREFIX + "refresh/time";
+  public static final String DATASOURCE_DROPPED = PREFIX + "datasource/dropped/count";
```
kfaraz:
```suggestion
  public static final String DATASOURCE_REMOVED = PREFIX + "dataSource/removed";
```

docs/operations/metrics.md
```diff
@@ -441,6 +441,7 @@ These metrics are emitted by the Druid Coordinator in every run of the correspon
 |`serverview/sync/unstableTime`|Time in milliseconds for which the Coordinator has been failing to sync with a segment-loading server. Emitted only when [HTTP-based server view](../configuration/index.md#segment-management) is enabled.|`server`, `tier`|Not emitted for synced servers.|
 |`metadatacache/init/time`|Time taken to initialize the coordinator segment metadata cache.||Depends on the number of segments.|
 |`segment/schemaCache/refresh/count`|Number of segments for which schema was refreshed in coordinator segment schema cache.|`dataSource`||
+|`segment/schemaCache/datasource/dropped/count`|Number of datasources dropped from broker cache after segments were marked unused.|`dataSource`||
```
```suggestion
|`segment/schemaCache/dataSource/removed`|Emitted when a datasource is removed from the Broker cache due to segments being marked as unused.|`dataSource`||
```

kfaraz (commented):
Since we already have this PR, do we still need the one disabling the test?

kfaraz (approved):
@uds5501 , can we re-enable the flaky test as a part of this PR?

uds5501:
@kfaraz yeah let me do that.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -296,6 +296,10 @@ public void waitForAllSegmentsToBeAvailable(String dataSource, EmbeddedCoordinat
                       .hasDimension(DruidMetrics.DATASOURCE, dataSource),
         agg -> agg.hasSumAtLeast(numSegments)
     );
+    broker.latchableEmitter().waitForEvent(
```
kfaraz: Maybe we should we look for aggregate where the sum is equal to numSegments.
After this change, do we even need to wait to wait for the load success.
@uds5501 , could you try to push a change with the coordinator part commented out?
If not needed, we can remove that check completely.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -284,7 +284,7 @@ public void verifyNumVisibleSegmentsIs(int numExpectedSegments, String dataSourc
    * Waits for all used segments (including overshadowed) of the given datasource
    * to be loaded on historicals.
```
kfaraz:
```suggestion
   * Waits for all used segments (including overshadowed) of the given datasource
   * to be queryable by the given Broker.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.java
```diff
@@ -177,6 +177,13 @@ public void test_runIndexTask_andKillData()
                       .hasService("druid/coordinator"),
         agg -> agg.hasSumAtLeast(numSegments)
     );
+
+    eventCollector.latchableEmitter().waitForEvent(
```
kfaraz, line 181:
```suggestion
    // Wait for the Broker to remove this datasource from its schema cache
    eventCollector.latchableEmitter().waitForEvent(
```

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
uds5501: Tested locally, the tests seems to be working fine without the coordinator wait, removed in the latest comment.
Also updated the EmbeddedBroker as part of the numSegments change.

cryptoe:
One input source IT was running. The patch does not change anything there hence merged the PR.

PR #18498, opened 2025-09-08
SQL: Add rule for merging nested Aggregates.

gianm:
The rule is adapted from Calcite's AggregateMergeRule, with two changes:

1) Includes a workaround for https://issues.apache.org/jira/browse/CALCITE-7162

2) Includes the ability to merge two Aggregate with a Project between them, by pushing the Project below the new merged Aggregate.

The goal is to rewrite a query like:

```sql
SELECT
  hr,
  UPPER(t1.x) x,
  SUM(t1.cnt) cnt,
  MIN(t1.mn) mn,
  MAX(t1.mx) mx
FROM (
  SELECT
    floor(__time to hour) hr,
    dim2 x,
    COUNT(*) cnt,
    MIN(m1 * 5) mn,
    MAX(m1 + m2) mx
  FROM druid.foo
  WHERE dim2 IN ('abc', 'def', 'a', 'b', '')
  GROUP BY 1, 2
) t1
WHERE t1.x IN ('abc', 'foo', 'bar', 'a', '')
GROUP BY 1, 2
```

Into:

```sql
SELECT
  FLOOR(__time TO hour) hr,
  UPPER(dim2) x,
  COUNT(*) cnt,
  MIN(m1 * 5) mn,
  MAX(m1 + m2) mx
FROM druid.foo
WHERE dim2 IN ('abc', 'a', '')
GROUP BY 1, 2
```

### DIFF
--- a/sql/src/main/java/org/apache/druid/sql/calcite/rule/AggregateMergeRule.java
+++ b/sql/src/main/java/org/apache/druid/sql/calcite/rule/AggregateMergeRule.java
@@ -58,9 +58,9 @@ import java.util.Map;
  */
 public abstract class AggregateMergeRule extends RelOptRule
 {
-  protected AggregateMergeRule(final RelOptRuleOperand operand)
+  protected AggregateMergeRule(final RelOptRuleOperand operand, final String description)
   {
-    super(operand);
+    super(operand, description);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


clintropolis (approved):
:+1: these plans look nicer to me

PR #18500, opened 2025-09-09
Vectorized implementation of timestamp_ceil.

gianm:
By analogy to `timestamp_floor`.


processing/src/main/java/org/apache/druid/query/expression/TimestampExtractExprMacro.java
```diff
@@ -183,6 +188,38 @@ public ExpressionType getOutputType(InputBindingInspector inspector)
...
+    public <T> ExprVectorProcessor<T> asVectorProcessor(VectorInputBindingInspector inspector)
+    {
+      final ExprVectorProcessor<?> processor;
+
+      if (getOutputExpressionType(unit).is(ExprType.DOUBLE)) {
+        processor = new DoubleUnivariateLongFunctionVectorProcessor(
+            CastToTypeVectorProcessor.cast(args.get(0).asVectorProcessor(inspector), ExpressionType.LONG),
+            input -> {
+              final DateTime dateTime = new DateTime(input, chronology);
+              return getExprEval(dateTime, unit).asDouble();
```
clintropolis: kind of sad we have to make an `ExprEval` to throw away, one of the benefits of vector expr processing is much less garbage objects to deal with. maybe we should change this method to return a `Number` and make another helper method for things that need to wrap it in `ExprEval`?

processing/src/main/java/org/apache/druid/query/expression/TimestampExtractExprMacro.java
```diff
@@ -183,6 +188,38 @@ public ExpressionType getOutputType(InputBindingInspector inspector)
...
+    }
+
+    @Override
+    public <T> ExprVectorProcessor<T> asVectorProcessor(VectorInputBindingInspector inspector)
+    {
+      final ExprVectorProcessor<?> processor;
+
+      if (getOutputExpressionType(unit).is(ExprType.DOUBLE)) {
+        processor = new DoubleUnivariateLongFunctionVectorProcessor(
+            CastToTypeVectorProcessor.cast(args.get(0).asVectorProcessor(inspector), ExpressionType.LONG),
```
clintropolis: I missed cleaning this up, but the explicit `cast` call is no longer needed here, it is handled by the vector processor, https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/math/expr/vector/DoubleUnivariateLongFunctionVectorProcessor.java#L38

gianm: Cool, removed it.

processing/src/main/java/org/apache/druid/query/expression/TimestampExtractExprMacro.java
gianm: Reworked to split this into two methods that return `long` or `double` directly.

PR #17248, opened 2024-10-04
add druid.expressions.allowVectorizeFallback and default to false

clintropolis:
changes:
* adds `ExpressionProcessing.allowVectorizeFallback()` and `ExpressionProcessingConfig.allowVectorizeFallback()`, defaulting to false until few remaining bugs can be fixed (mostly complex types and some odd interactions with mixed types)
* add `cannotVectorizeUnlessFallback` functions to make it easy to toggle the default of this config, and easy to know what to delete when we remove it in the future

PR #17098, opened 2024-09-18
various fixes and improvements to vectorization fallback

clintropolis:
changes:
* add `ApplyFunction` support to vectorization fallback, allowing many of the remaining expressions to be vectorized
* add `CastToObjectVectorProcessor` so that vector engine can correctly cast any type
* add support for array and complex vector constants
* reduce number of cases which can block vectorization in expression planner to be unknown inputs (such as unknown multi-valuedness)
* fix array constructor expression, apply map expression to make actual evaluated type match the output type inference
* fix bug in array_contains where something like array_contains([null], 'hello') would return true if the array was a numeric array since the non-null string value would cast to a null numeric
* fix isNull/isNotNull to correctly handle any type of input argument

PR #16366, opened 2024-05-01
Fallback vectorization for FunctionExpr and BaseMacroFunctionExpr.

gianm:
This patch adds `FallbackVectorProcessor`, a processor that adapts non-vectorizable operations into vectorizable ones. It is used in `FunctionExpr` and `BaseMacroFunctionExpr`. As a result, all such expressions can now participate in vectorized queries.

In addition:

- Identifiers are updated to offer getObjectVector for ARRAY and COMPLEX in addition to STRING. ExprEvalObjectVector is updated to offer ARRAY and COMPLEX as well. Identifiers already did `return true` from `canVectorize`, so this enables them to live up to their claims.

- In SQL tests, cannotVectorize now fails tests if an exception is not thrown. This makes it easier to identify tests that can now vectorize.

- Fixes a null-matcher bug in StringObjectVectorValueMatcher that was uncovered by certain newly-vectorizable test cases.

Benchmarks follow for `SqlExpressionBenchmark` queries 26 and 27. These two queries are:

```
-- 26: group by string expr with non-expr agg
SELECT CONCAT(string2, '-', long2), SUM(double1) FROM foo GROUP BY 1 ORDER BY 2

-- 27: group by string expr with expr agg
SELECT CONCAT(string2, '-', long2), SUM(long1 * double4) FROM foo GROUP BY 1 ORDER BY 2
```

In these cases fallback vectorization is not as compelling as proper vectorization, but it's better than unvectorized execution. The relative benefit is greater for query 27, likely because fallback vectorization for `CONCAT` enables the `long1 * double4` to vectorize as well. In general I would expect the benefit to be greater for more complex queries, due to this effect.

```
Benchmark                        (query)  (rowsPerSegment)  (schema)  (vectorize)  Mode  Cnt    Score    Error  Units

SqlExpressionBenchmark.querySql       26            500000      auto        false  avgt    5  259.288 ±  4.863  ms/op
SqlExpressionBenchmark.querySql       26            500000      auto       native  avgt    5  194.405 ± 10.185  ms/op
SqlExpressionBenchmark.querySql       26            500000      auto     fallback  avgt    5  244.807 ± 6.851  ms/op

SqlExpressionBenchmark.querySql       27            500000      auto        false  avgt    5  289.065 ± 15.253  ms/op
SqlExpressionBenchmark.querySql       27            500000      auto       native  avgt    5  194.829 ±  7.732  ms/op
[...]

PR #18503, opened 2025-09-09
Additional expr type alignment.

gianm:
PR #16366 originally added fallback vectorization, a mechanism for making all expressions vectorizable. Later, #17098 fixed some issues that arose and #17248 disabled fallback vectorization in the out-of-box configuration.

This patch fixes various remaining issues with inconsistent type handling between the vectorized and nonvectorized expr implementations. It does not yet re-enable fallback vectorization out of the box, due to remaining inconsistencies with conditional exprs like "case_searched", "case_simple", and "if".

1) Aligns the behavior of missing columns and literal nulls so they are
   always treated as null longs. This was already the case for vectorized
   identifiers, but non-vectorized identifiers and literal nulls were still
   represented as strings.

2) Replaces all occurrences of "ExprEval.of(null)" with either an explicit
   type, or a call to "ExprEval.ofMissing()". ofMissing is a new function
   for situations where an eval represents a null value of unknown type.
   It is equivalent to "ExprEval.ofLong(null)", but is a separate function
   for clarity at the call site.

3) Update "cast" to return the target type even for null values.

4) Update "greatest", "least", and "array" so they eval to types that
   match what is reported by "getOutputType".

5) Update "scalb" to coerce input strings as numbers, to better allow
   for type evolution and missing columns.

6) Update "reverse" to coerce inputs to strings, to better allow for
   type evolution and missing columns.

### DIFF
--- a/processing/src/main/java/org/apache/druid/math/expr/vector/FallbackVectorProcessor.java
+++ b/processing/src/main/java/org/apache/druid/math/expr/vector/FallbackVectorProcessor.java
@@ -44,7 +44,7 @@ public abstract class FallbackVectorProcessor<T> implements ExprVectorProcessor<
   final Supplier<ExprEval<?>> fn;
   final List<AdaptedExpr> adaptedArgs;
 
-  private final ExpressionType outputType;
+  protected final ExpressionType outputType;
 
   private FallbackVectorProcessor(
       final Supplier<ExprEval<?>> fn,
@@ -200,10 +200,10 @@ public abstract class FallbackVectorProcessor<T> implements ExprVectorProcessor<
           adaptedArg.setRowNumber(i);
         }
 
-        outValues[i] = fn.get().value();
+        outValues[i] = fn.get().castTo(outputType).value();
       }
 
-      return new ExprEvalObjectVector(outValues, getOutputType());
+      return new ExprEvalObjectVector(outValues, outputType);
     }
 
     @Override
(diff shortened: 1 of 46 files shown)


processing/src/main/java/org/apache/druid/math/expr/vector/FallbackVectorProcessor.java
```diff
@@ -368,7 +368,7 @@ public ExprEval eval(ObjectBinding bindings)
         final boolean isNull = results.getNullVector() != null && results.getNullVector()[rowNum];
         return ExprEval.ofDouble(isNull ? null : results.getDoubleVector()[rowNum]);
       } else {
-        return ExprEval.ofType(type, results.getObjectVector()[rowNum]);
+        return ExprEval.bestEffortOf(results.getObjectVector()[rowNum]);
```
clintropolis, line 371: what was the problem with using `ofType` that is solved by best effort + cast?

gianm: I don't remember exactly which test failed, but the problem was that `ofType` doesn't coerce, it mostly trusts that the value you pass in is actually that type. But, there were some functions that lie about their output type (i.e. they return an expr that doesn't match the type from `getOutputType`), and to plug those in properly we need to coerce to the declared output type. I fixed a few in this patch but I'm not sure I got them all.

PR #18507, opened 2025-09-09
vectorize if function when then and else expressions have the same output type

clintropolis:
### Description
Vectorizes the `if` expression in cases where both then and else clause have the same `getOutputType`. This avoids solving a harder problem which is that the non-vectorized conditional expressions allow varying return types between the different result expressions and does not coerce them to a common type.

I think we should consider changing this to coerce to the least restrictive type to behave consistently with other functions, but not in this PR.

I haven't measured performance of this yet to compare non-vectorized and vectorized performance, but will try to do that soon.

### DIFF
--- a/processing/src/main/java/org/apache/druid/math/expr/Function.java
+++ b/processing/src/main/java/org/apache/druid/math/expr/Function.java
@@ -2144,6 +2144,8 @@ public interface Function extends NamedFunction
     @Override
     public boolean canVectorize(Expr.InputBindingInspector inspector, List<Expr> args)
     {
+      // vector engine requires consistent typing, but native if function does not coerce then and else expressions,
+      // so for now we can only vectorize if both args have the same output type to not have a behavior change
       final ExpressionType thenType = args.get(1).getOutputType(inspector);
       return Objects.equals(thenType, args.get(2).getOutputType(inspector));
     }
--- a/processing/src/main/java/org/apache/druid/math/expr/Function.java
+++ b/processing/src/main/java/org/apache/druid/math/expr/Function.java
@@ -2140,6 +2140,19 @@ public interface Function extends NamedFunction
     {
       return ExpressionTypeConversion.conditional(inspector, args.subList(1, 3));
     }
+
+    @Override
+    public boolean canVectorize(Expr.InputBindingInspector inspector, List<Expr> args)
+    {
+      final ExpressionType thenType = args.get(1).getOutputType(inspector);
+      return Objects.equals(thenType, args.get(2).getOutputType(inspector));
+    }
+
+    @Override
+    public <T> ExprVectorProcessor<T> asVectorProcessor(Expr.VectorInputBindingInspector inspector, List<Expr> args)
+    {
+      return VectorConditionalProcessors.ifFunction(inspector, args.get(0), args.get(1), args.get(2));
+    }
   }
 
   /**
--- a/processing/src/main/java/org/apache/druid/math/expr/vector/ExprEvalBindingVector.java
+++ b/processing/src/main/java/org/apache/druid/math/expr/vector/ExprEvalBindingVector.java
@@ -43,6 +43,7 @@ public class ExprEvalBindingVector<T> implements ExprEvalVector<T>
 
   @Nullable
   private boolean[] numericNulls;
+  private Object[] objects;
 
   public ExprEvalBindingVector(
       ExpressionType expressionType,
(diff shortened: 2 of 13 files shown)


processing/src/main/java/org/apache/druid/math/expr/Function.java
```diff
@@ -2140,6 +2140,19 @@ public ExpressionType getOutputType(Expr.InputBindingInspector inspector, List<E
     {
       return ExpressionTypeConversion.conditional(inspector, args.subList(1, 3));
     }
+
+    @Override
+    public boolean canVectorize(Expr.InputBindingInspector inspector, List<Expr> args)
+    {
+      final ExpressionType thenType = args.get(1).getOutputType(inspector);
+      return Objects.equals(thenType, args.get(2).getOutputType(inspector));
```
gianm, line 2150: Comment here about why the type check exists would be good.

processing/src/main/java/org/apache/druid/math/expr/vector/ExprEvalDoubleVector.java
```diff
@@ -69,4 +70,13 @@ public Object[] getObjectVector()
     }
     return objects;
   }
+
+  @Override
+  public boolean elementAsBoolean(int index)
+  {
+    if (nulls != null && nulls[index]) {
+      return Evals.asBoolean(0.0);
```
gianm: This is just `return false` right?

processing/src/main/java/org/apache/druid/math/expr/vector/ExprEvalBindingVector.java
```diff
@@ -133,9 +136,42 @@ public Object[] getObjectVector()
...
       }
-      return objects;
+    } else {
+      objects = bindings.getObjectVector(bindingName);
+    }
+    return objects;
+  }
+
+  @Override
+  public boolean elementAsBoolean(int index)
```
gianm, line 146: Would it make sense to vectorize this as well, so it's like `boolean[] getConditionVector()`?

processing/src/main/java/org/apache/druid/math/expr/vector/VectorConditionalProcessors.java
```diff
@@ -57,4 +57,43 @@ public static <T> ExprVectorProcessor<T> nvl(Expr.VectorInputBindingInspector in
...
   }
+
+  public static <T> ExprVectorProcessor<T> ifFunction(
+      Expr.VectorInputBindingInspector inspector,
+      Expr conditionExpr,
+      Expr thenExpr,
+      Expr elseExpr
+  )
+  {
+    // right now this function can only vectorize if then and else clause have same output type, if this changes then
```
gianm, line 68: is this comment cut off? it seems to end in the middle of a sentence

processing/src/test/java/org/apache/druid/math/expr/VectorExprResultConsistencyTest.java
```diff
@@ -348,6 +349,28 @@ public void testSymmetricalBivariateFunctions()
     testFunctions(types, templates, functions);
   }
 
+  @Test
+  public void testIfFunction()
+  {
```
gianm, line 354: Would be good to include some tests to validate that lazy evaluation works, such as `if(l1 % 2 == 0, l2 / (l1 % 2), -1)`. That would throw a division-by-zero exception if lazy evaluation wasn't working.

processing/src/main/java/org/apache/druid/math/expr/vector/ExprEvalBindingVector.java
clintropolis: i thought about this, but i wasn't sure it seemed better since we would have to loop over every value in the vector to make the `boolean[]`, and then again loop over all of the condition booleans array to then build the vector match, which seemed worse, but admittedly i didn't measure it.

processing/src/main/java/org/apache/druid/math/expr/vector/ExprEvalDoubleVector.java
clintropolis: it is, and i suppose it is unlikely we will ever change the behavior of the `Evals.asBoolean` calls, I was just being overly cautious to match logic in `ExprEval` exactly

processing/src/test/java/org/apache/druid/math/expr/VectorExprResultConsistencyTest.java
clintropolis: added test, though your comment has the then/else swapped because ^ is always divide by zero :bomb:

gianm: oops. TY

gianm (approved):

PR #18508, opened 2025-09-09
fixes for filtered projections that produce empty projection tables

clintropolis:
### Description
This PR fixes a bug that can occur when a filtered projection results in an empty dataset, where during a query the time boundary inspector assumes that there is at least 1 row in the dataset. A follow-up PR will optimize storage to not write out empty projections and instead only store metadata and produce an synthetic empty cursor at query time.

### DIFF
--- a/processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
+++ b/processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
@@ -1314,6 +1314,7 @@ public class CursorFactoryProjectionTest extends InitializedNullHandlingTest
   @Test
   public void testTimeseriesQueryGranularityFinerThanProjectionGranularity()
   {
+    // timeseries query only works on base table if base table is sorted by time
     Assume.assumeTrue(segmentSortedByTime);
     final TimeseriesQuery query = Druids.newTimeseriesQueryBuilder()
                                         .dataSource("test")
@@ -1536,6 +1537,7 @@ public class CursorFactoryProjectionTest extends InitializedNullHandlingTest
                     .build();
 
     final boolean isRealtime = projectionsCursorFactory instanceof IncrementalIndexCursorFactory;
+    // realtime projections don't have row count, so abfoo is chosen because of how projection sorting happens
     final ExpectedProjectionGroupBy queryMetrics = new ExpectedProjectionGroupBy(isRealtime ? "abfoo" : "a_hourly_c_sum_filter_a_to_a");
 
     final CursorBuildSpec buildSpec = GroupingEngine.makeCursorBuildSpec(query, queryMetrics);
@@ -1646,16 +1648,19 @@ public class CursorFactoryProjectionTest extends InitializedNullHandlingTest
         results
     );
 
+
     Assertions.assertEquals(TIMESTAMP, projectionsTimeBoundaryInspector.getMinTime());
     if (isRealtime || segmentSortedByTime) {
       Assertions.assertEquals(TIMESTAMP.plusHours(1).plusMinutes(1), projectionsTimeBoundaryInspector.getMaxTime());
     } else {
+      // min and max time are inexact for non time ordered segments
       Assertions.assertEquals(
           TIMESTAMP.plusHours(1).plusMinutes(1).plusMillis(1),
           projectionsTimeBoundaryInspector.getMaxTime()
       );
     }
 
+    // timeseries query only works on base table if base table is sorted by time
     Assume.assumeTrue(segmentSortedByTime);
     final Sequence<Result<TimeseriesResultValue>> resultRowsNoProjection = timeseriesEngine.process(
         query.withOverriddenContext(Map.of(QueryContexts.NO_PROJECTIONS, true)),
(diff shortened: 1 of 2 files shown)


processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -1484,6 +1523,155 @@ public void testProjectionSelectionTwoVirtual()
...
+        GroupByQuery.builder()
+                    .setDataSource("test")
+                    .setGranularity(Granularities.ALL)
+                    .setInterval(Intervals.ETERNITY)
+                    .setDimFilter(new EqualityFilter("a", ColumnType.STRING, "a", null))
+                    .addDimension("a")
+                    .build();
+
+    final boolean isRealtime = projectionsCursorFactory instanceof IncrementalIndexCursorFactory;
+    final ExpectedProjectionGroupBy queryMetrics = new ExpectedProjectionGroupBy(isRealtime ? "abfoo" : "a_hourly_c_sum_filter_a_to_a");
```
cecemei, line 1541: it's not obvious the projection matching returns different results for realtime/ data segments. digging into it, seems like it's about the sorting of projections (less # of rows first, then coarser gran first), and first match is returned. in realtime segments, # of rows is always 0 thus `abfoo` with ALL gran always returns first, while in data segments, `a_hourly_c_sum_filter_a_to_a` wins by having less total rows in projection (2 < 6). we should add some comments here in case ppl are wondering.

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -1484,6 +1523,155 @@ public void testProjectionSelectionTwoVirtual()
...
+    if (isRealtime || segmentSortedByTime) {
+      Assertions.assertEquals(TIMESTAMP.plusHours(1).plusMinutes(1), projectionsTimeBoundaryInspector.getMaxTime());
+    } else {
+      Assertions.assertEquals(
+          TIMESTAMP.plusHours(1).plusMinutes(1).plusMillis(1),
+          projectionsTimeBoundaryInspector.getMaxTime()
+      );
+    }
+
+    Assume.assumeTrue(segmentSortedByTime);
```
cecemei, line 1664: is this assumeTrue necessary?

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -1484,6 +1523,155 @@ public void testProjectionSelectionTwoVirtual()
...
+        new ExpectedProjectionTimeseries("a_hourly_c_sum_filter_a_to_empty");
+
+    final CursorBuildSpec buildSpec = TimeseriesQueryEngine.makeCursorBuildSpec(query, queryMetrics);
+
+    assertCursorProjection(buildSpec, queryMetrics, 0);
+
+    // realltime results are inconsistent between projection and base table since projection is totally empty, but base
+    // table is reduced with filter
+    final boolean isRealtime = projectionsCursorFactory instanceof IncrementalIndexCursorFactory;
+    final List<Object[]> expectedResults = Collections.singletonList(new Object[]{TIMESTAMP, null});
```
cecemei, line 1632: nit: could be more declarative with:
`final List<Object[]> expectedResults = isRealtime ?  List.of() : List.of(new Object[]{TIMESTAMP, null});`

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -1484,6 +1523,155 @@ public void testProjectionSelectionTwoVirtual()
...
+        query.getResultRowSignature(RowSignature.Finalization.YES),
+        isRealtime ? expectedRealtimeResults : expectedResults,
+        results
+    );
+
+    Assertions.assertEquals(TIMESTAMP, projectionsTimeBoundaryInspector.getMinTime());
+    if (isRealtime || segmentSortedByTime) {
+      Assertions.assertEquals(TIMESTAMP.plusHours(1).plusMinutes(1), projectionsTimeBoundaryInspector.getMaxTime());
+    } else {
+      Assertions.assertEquals(
```
cecemei, line 1657: maybe add a comment here explaining for data segments that's not sorted by time, the max time is inexact

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
clintropolis: i still always need `Collections.singletonList(new Object[]{TIMESTAMP, null})` because the base table doesn't return an empty result, so I left this like it is (also cannot use `List.of(new Object[]{...})` with only a single arg because it is ambiguous whether you want a `List<Object>` or `List<Object[]>`)

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
clintropolis: yea, added comment since timeseries query only works if the data is ordered by time

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
clintropolis: added comment

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
clintropolis: added comment

PR #18510, opened 2025-09-10
HttpRemoteTaskRunner: Fix hanging RUNNING task after overlord and worker restart

kgyrtkirk:
Phantom tasks could have appeared which were running indefinetly and may have caused the system to try to kill them repeatedly without much success

* if `HttpRemoteTaskRunner` is used
* launch a task (must be still running when the worker is stopped)
  *  the `Task`  object will be fully registered into the `HttpRemoteTaskRunnerWorkItem`
* cycle the overlord
  * the running task is picked up from the still alive worker; however it will be [registered](https://github.com/apache/druid/blob/45cf67d09e10aaf493556d2fe512f2034eead897/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java#L1541) w/o `Task` informations
* cycle the worker
  * during connection some [expectedAnnouncements](https://github.com/apache/druid/blob/45cf67d09e10aaf493556d2fe512f2034eead897/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java#L610-L625) are forged...however w/o the `Task` that's not possible
* the task remains in `RUNNING` state

note:
* killing a stuck task like the above will:
   * change its status to FAILED
   * sends  shutdown to the worker
      * which will be ignored
      * and thus no `Announcement` will be generated
   * it will remain stuck in the overlord
   * killed again in the next round

The proposed fix fills in the announcement w/o using the `Task` details; the only purpose of that announcement is to emit the the [fake Announcement](https://github.com/apache/druid/blob/45cf67d09e10aaf493556d2fe512f2034eead897/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/WorkerHolder.java#L354-L367) about that the task have disappeared.


indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -611,17 +612,18 @@ void addWorker(final Worker worker)
...
-                    )
-                );
-              }
+            HttpRemoteTaskRunnerWorkItem workItem = e.getValue();
+            if (workItem.isRunningOnWorker(worker)) {
+              expectedAnnouncements.add(
+                  TaskAnnouncement.create(
+                      workItem.getTaskId(),
+                      workItem.getTaskType(),
+                      null,
```
uds5501, line 626: What's the reason for setting taskResource and data source as null here?

kgyrtkirk: I think those questions can't really be answered without the `Task` object - 
these fake announcements will only be used to generate the task disappeared event....

the "damage" that when an overlord detects an unknown task running on a worker when it starts is already accepted ...see [here](https://github.com/apache/druid/blame/4fc154232e6789f8ce8020a7d286f55e61ce0f7d/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java#L1541)

uds5501: Ah fair, that makes sense, after this the task runner will be able to figure out that the task is in FAILED state and not keep it perpetually I assume? + It makes it more consistent regarding how we are dealing with task discoveries

uds5501 (approved):

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/OverlordIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.druid.testing.embedded.EmbeddedCoordinator;
+import org.apache.druid.testing.embedded.EmbeddedDruidCluster;
+import org.apache.druid.testing.embedded.EmbeddedIndexer;
+import org.apache.druid.testing.embedded.EmbeddedOverlord;
+import org.apache.druid.testing.embedded.junit5.EmbeddedClusterTestBase;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class OverlordIntegrationTest extends EmbeddedClusterTestBase
```
kfaraz: Please rename to `HttpTaskRunnerWorkerFailTest` or similar.

embedded-tests/src/test/resources/log4j2.xml
```diff
@@ -46,5 +46,8 @@
     <Logger level="info" name="org.apache.druid" additivity="false">
       <AppenderRef ref="RoutingAppender"/>
     </Logger>
+    <Logger level="error" name="org.apache.druid.metadata.SqlSegmentsMetadataManager" additivity="false">
```
kfaraz: Yeah, this log becomes very verbose in tests since we keep a very small metadata polling period. 😅 
Suggestion: Another way to avoid these verbose logs is to set the property `druid.manager.segments.useIncrementalCache=ifSynced` on the Coordinator.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/OverlordIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+    // give some time for the overlord to dispatch the task to the worker
+    Thread.sleep(100);
+    overlord.stop();
+    overlord.start();
+    // give some time for the overlord to load the task from the worker
+    Thread.sleep(100);
+    indexer.stop();
+    indexer.start();
+    // give time for the overlord to realize that the task is not there anymore
+    Thread.sleep(100);
```
kfaraz: Instead of this sleep, we can wait for the `task/run/time` metric to be emitted.

```suggestion
    // Wait for the Overlord to mark the task as FAILED
    overlord.latchableEmitter().waitForEvent(
       event -> event.hasMetricName("task/run/time")
                               .hasDimension(DruidMetrics.TASK_ID, taskId)
                               .hasDimension(DruidMetrics.TASK_STATUS, "FAILED")
    );
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/OverlordIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+    final String taskId = IdUtils.newTaskId("sim_test_noop", TestDataSource.WIKI, null);
+    cluster.callApi().onLeaderOverlord(
+        o -> o.runTask(taskId, new NoopTask(taskId, null, null, 8000L, 0L, null))
+    );
+    // give some time for the overlord to dispatch the task to the worker
+    Thread.sleep(100);
+    overlord.stop();
+    overlord.start();
+    // give some time for the overlord to load the task from the worker
+    Thread.sleep(100);
```
kfaraz: Instead of using these sleeps, we can override the `runTask()` method of the given `NoopTask` to count down a latch. The test can then wait on this latch to ensure that the task has started running on the Indexer.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/OverlordIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+    return EmbeddedDruidCluster.withEmbeddedDerbyAndZookeeper()
+        .useLatchableEmitter()
+        .addServer(new EmbeddedCoordinator())
+        .addServer(new EmbeddedBroker())
+        .addServer(overlord)
+        .addServer(indexer);
+  }
+
+  @Test
+  public void test_runTask_ofTypeNoop() throws Exception
```
kfaraz:
```suggestion
  public void test_overlord_marksTaskAsFailed_ifIndexerCrashes() throws Exception
```

kfaraz (commented):
Change looks good. Left some suggestions to simplify the test.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: Let's add this info as a comment.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -611,17 +612,18 @@ void addWorker(final Worker worker)
...
-                    TaskAnnouncement.create(
-                        e.getValue().getTask(),
-                        TaskStatus.running(e.getKey()),
-                        e.getValue().getLocation()
-                    )
-                );
-              }
+            HttpRemoteTaskRunnerWorkItem workItem = e.getValue();
+            if (workItem.isRunningOnWorker(worker)) {
+              expectedAnnouncements.add(
```
kfaraz, line 622:
```suggestion
              // This announcement is only used to notify when a task has disappeared on the worker
              // So it is okay to set the dataSource and taskResource to null as they will not be used
              expectedAnnouncements.add(
```

embedded-tests/src/test/resources/log4j2.xml
kgyrtkirk: I think the value of those `INFO` level log messages is practically zero if someone is running any tests....
setting a property on a single test will make it a little bit better for that one test....

removed these changes

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/OverlordIntegrationTest.java
kgyrtkirk: the task needs to be registered with jackson; instead I've enhanced the nooptask to emit an event when it starts running; also fixed the task to honor the "stop" so that it doesn't delay thing further if it was asked to stop....

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kgyrtkirk: there is already a comment in [HttpRemoteTaskRunnerWorkItem](https://github.com/apache/druid/blame/4fc154232e6789f8ce8020a7d286f55e61ce0f7d/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java#L1891-L1893)...
do you mean more?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/OverlordIntegrationTest.java
kfaraz: Great, that works too! 👍🏻

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IndexTaskTest.java
```diff
@@ -99,6 +99,9 @@ public void test_runIndexTask_forInlineDatasource()
     }
 
     cluster.callApi().waitForAllSegmentsToBeAvailable(dataSource, coordinator, broker);
+    broker.latchableEmitter().waitForMetricEvent(
+        event -> event.hasDimension(DruidMetrics.DATASOURCE, dataSource)
+    );
```
kfaraz: Is this needed? the call before this line (`waitForAllSegmentsToBeAvailable`) should already be taking care of segment availability.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NoopTask.java
```diff
@@ -29,33 +29,38 @@
...
   @JsonCreator
   public NoopTask(
       @JsonProperty("id") String id,
       @JsonProperty("groupId") String groupId,
       @JsonProperty("dataSource") String dataSource,
       @JsonProperty("runTime") long runTimeMillis,
       @JsonProperty("isReadyTime") long isReadyTime,
-      @JsonProperty("context") Map<String, Object> context
-  )
+      @JsonProperty("context") Map<String, Object> context)
```
kfaraz: Nit: not needed. older style was cleaner.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NoopTask.java
```diff
@@ -29,33 +29,38 @@
...
 public class NoopTask extends AbstractTask implements PendingSegmentAllocatingTask
 {
   public static final String TYPE = "noop";
+  public static final String NOOP_TASK_EVENT_STARTED = "noopTask-started";
   private static final int DEFAULT_RUN_TIME = 2500;
 
   @JsonIgnore
   private final long runTime;
 
+  @JsonIgnore
```
kfaraz: Nit: I think this annotation is not required. Any field/method not explicitly marked `@JsonProperty` is skipped from serialization by default.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NoopTask.java
```diff
@@ -97,13 +102,22 @@ public boolean isReady(TaskActionClient taskActionClient)
...
   @Override
   public TaskStatus runTask(TaskToolbox toolbox) throws Exception
   {
-    Thread.sleep(runTime);
-    return TaskStatus.success(getId());
+    toolbox.getEmitter().emit(ServiceMetricEvent.builder().setMetric(NOOP_TASK_EVENT_STARTED, 1));
+    long endTime = System.currentTimeMillis() + runTime;
+    while (endTime > System.currentTimeMillis() && !aborted.get()) {
+      Thread.sleep(100);
+    }
```
kfaraz: How about we use a `countDownLatch.await(runTime)` instead?
The latch will be counted down in `stopGracefully()`.
So we will wait on the latch until stop is requested or until the runTime elapses.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NoopTask.java
```diff
@@ -97,13 +102,22 @@ public boolean isReady(TaskActionClient taskActionClient)
...
   }
 
   @Override
   public TaskStatus runTask(TaskToolbox toolbox) throws Exception
   {
-    Thread.sleep(runTime);
-    return TaskStatus.success(getId());
+    toolbox.getEmitter().emit(ServiceMetricEvent.builder().setMetric(NOOP_TASK_EVENT_STARTED, 1));
+    long endTime = System.currentTimeMillis() + runTime;
+    while (endTime > System.currentTimeMillis() && !aborted.get()) {
```
kfaraz: Just curious as to why we needed to maintain the `aborted` field.
When we stop the `Indexer`, it should ideally interrupt all its tasks, thus triggering a `task/run/time` `FAILED` event.
Was that not happening?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: No, I just meant the one on line 620.
Resolving this.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1543,6 +1550,14 @@ public void taskAddedOrUpdated(final TaskAnnouncement announcement, final Worker
                   HttpRemoteTaskRunnerWorkItem.State.RUNNING
               );
               tasks.put(taskId, taskItem);
+              final ServiceMetricEvent.Builder metricBuilder = new ServiceMetricEvent.Builder();
+              metricBuilder.setDimension(DruidMetrics.TASK_ID, taskId);
+              emitter.emit(
+                  metricBuilder.setMetric(
+                      TASK_UNKNOWN_COUNT,
+                      (long) 1
+                  )
```
kfaraz: Nit: can be done in a single line.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/HttpRemoteTaskRunnerWorkerFailTest.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Test
+  public void test_overlord_marksTaskAsFailed_ifIndexerCrashes() throws Exception
+  {
+    final String taskId = IdUtils.newTaskId("sim_test_noop", TestDataSource.WIKI, null);
+    cluster.callApi().onLeaderOverlord(
+        o -> {
+          return o.runTask(taskId, new NoopTask(taskId, null, null, 8000L, 0L, null));
+        }
```
kfaraz: Nit: simpler lambda
```suggestion
        o -> o.runTask(taskId, new NoopTask(taskId, null, null, 8000L, 0L, null))
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NoopTask.java
```diff
@@ -29,33 +29,38 @@
...
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  */
 public class NoopTask extends AbstractTask implements PendingSegmentAllocatingTask
 {
   public static final String TYPE = "noop";
+  public static final String NOOP_TASK_EVENT_STARTED = "noopTask-started";
```
kfaraz: Nit:
```suggestion
  public static final String EVENT_STARTED = "task/noop/started";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NoopTask.java
```diff
@@ -97,13 +102,22 @@ public boolean isReady(TaskActionClient taskActionClient)
...
   {
+    aborted.set(true);
   }
 
   @Override
   public TaskStatus runTask(TaskToolbox toolbox) throws Exception
   {
-    Thread.sleep(runTime);
-    return TaskStatus.success(getId());
+    toolbox.getEmitter().emit(ServiceMetricEvent.builder().setMetric(NOOP_TASK_EVENT_STARTED, 1));
```
kfaraz: Shorthand:


```suggestion
    emitMetric(toolbox.getEmitter(), NOOP_TASK_EVENT_STARTED, 1);
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/HttpRemoteTaskRunnerWorkerFailTest.java
```diff
@@ -0,0 +1,93 @@
...
+    );
+    // wait for the overlord to dispatch the task and worker start it
+    indexer.latchableEmitter().waitForMetricEvent(
+        event -> event.hasMetricName(NoopTask.NOOP_TASK_EVENT_STARTED),
+        1000
```
kfaraz: Any specific reason to not use the default timeout (10s)?
Lower timeouts may cause flakiness in the tests.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/HttpRemoteTaskRunnerWorkerFailTest.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.druid.testing.embedded.EmbeddedOverlord;
+import org.apache.druid.testing.embedded.junit5.EmbeddedClusterTestBase;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class HttpRemoteTaskRunnerWorkerFailTest extends EmbeddedClusterTestBase
+{
+  private final EmbeddedOverlord overlord = new EmbeddedOverlord();
+  private final EmbeddedIndexer indexer = new EmbeddedIndexer().addProperty("druid.worker.capacity", "3");
```
kfaraz: Probably not needed for this test since we are going to run a single task.
```suggestion
  private final EmbeddedIndexer indexer = new EmbeddedIndexer();
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IndexTaskTest.java
kgyrtkirk: no

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NoopTask.java
kgyrtkirk: regardless what other classes do - this `runTask` method should return failure...as it should be failed as it was aborted

I don't think interrupts are used as it was running for the full time of the `runTime`; commenting the content of `stopGracefully` restores the original behaviour...which hangs for 8 second for the new test.

I wonder if the interrupt would be handled correctly - would there be still a need for the `stopGracefully` method?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/HttpRemoteTaskRunnerWorkerFailTest.java
kgyrtkirk: 10s is a lot....
removed the method and the abilitty to set the timeout

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IndexTaskTest.java
```diff
@@ -99,6 +100,9 @@ public void test_runIndexTask_forInlineDatasource()
     }
 
     cluster.callApi().waitForAllSegmentsToBeAvailable(dataSource, coordinator, broker);
+    broker.latchableEmitter().waitForEvent(
+        event -> event.hasDimension(DruidMetrics.DATASOURCE, dataSource)
+    );
```
kfaraz: Please remove this.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1543,6 +1550,9 @@ public void taskAddedOrUpdated(final TaskAnnouncement announcement, final Worker
                   HttpRemoteTaskRunnerWorkItem.State.RUNNING
               );
               tasks.put(taskId, taskItem);
+              final ServiceMetricEvent.Builder metricBuilder = new ServiceMetricEvent.Builder();
+              metricBuilder.setDimension(DruidMetrics.TASK_ID, taskId);
+              emitter.emit(metricBuilder.setMetric(TASK_UNKNOWN_COUNT, (long) 1));
```
kfaraz: Nit:
```suggestion
              emitter.emit(metricBuilder.setMetric(TASK_UNKNOWN_COUNT, 1L));
```

indexing-service/src/test/java/org/apache/druid/indexing/common/task/batch/parallel/TaskMonitorTest.java
```diff
@@ -296,10 +305,23 @@ public ListenableFuture<Void> runTask(String taskId, Object taskObject)
       if (task.throwUnknownTypeIdError) {
         throw new RuntimeException(new ISE("Could not resolve type id 'test_task_id'"));
       }
-      taskRunner.submit(() -> tasks.put(task.getId(), task.run(null).getStatusCode()));
+      TaskToolbox taskToolbox = makeToolbox();
+      taskRunner.submit(() -> tasks.put(task.getId(), task.run(taskToolbox).getStatusCode()));
```
kfaraz, line 309: Just curious why this change was needed?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: IIUC, this metric is emitted whenever the Overlord is notified of a running task for which it doesn't have a work item in the runner. If that is the case, maybe `task/unknown/count` is not the best metric name for it since it is not really an unknown task (the `TaskQueue` and `TaskStorage` know about it).

Also, the emission of this metric is susceptible to race conditions since after the Overlord restarts,
the `TaskQueue` may have called `taskRunner.run()` causing the `taskItem` to be non-null thus
skipping the emission of this metric.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/server/HttpRemoteTaskRunnerWorkerFailTest.java
```diff
@@ -0,0 +1,88 @@
...
+    );
+    // wait for the overlord to dispatch the task and worker start it
+    indexer.latchableEmitter().waitForEvent(
+        event -> event.hasMetricName(NoopTask.EVENT_STARTED)
+    );
+    overlord.stop();
+    overlord.start();
+    // give some time for the overlord to load the task from the worker
+    overlord.latchableEmitter().waitForEvent(
+        event -> event.hasMetricName(HttpRemoteTaskRunner.TASK_UNKNOWN_COUNT)
```
kfaraz: The emission of this metric is susceptible to race conditions, I think.
If the `TaskQueue` calls `taskRunner.run()` before the Overlord discovers the task,
the metric may never be emitted.

What happens if we don't wait for the Overlord to have re-discovered the task from the worker
and just proceed with killing it off?

kfaraz (commented):
Looks good. Just have some concerns regarding race conditions with the emission of the new metric.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IndexTaskTest.java
kgyrtkirk: oh... 69be8315d675376d7408224ee6c477fa705dade1 removed this and I had a conflict and this remained incorrectly

indexing-service/src/test/java/org/apache/druid/indexing/common/task/batch/parallel/TaskMonitorTest.java
kgyrtkirk: because test did not set up things correctly...emitting a metric in the task have hit the wall that its simply  not set...

I also had to undo the `stopGracefully` enhancements from the `NoopTask` as it seemed like it was destabilizing some other tests

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kgyrtkirk: renamed it to `task/discovered/count`

it seems like in case the new overlord becomes the leader it may launch that task again...
although I was not able to make that happen with debug points - as there is a latch which prevents this...

as [HttpRemoteTaskRunner](https://github.com/apache/druid/blob/9481a535be627d2e9cb1974e68eb12935e70e5b3/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java#L569) blocks untile workers sync state.
which blocks the `DruidOverlord#becomeLeader` thru `Lifecycle#start`

kfaraz: > as [HttpRemoteTaskRunner](https://github.com/apache/druid/blob/9481a535be627d2e9cb1974e68eb12935e70e5b3/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java#L569) blocks untile workers sync state.
which blocks the DruidOverlord#becomeLeader thru Lifecycle#start

I see. Thanks for the clarification!
So, IIUC, this metric will always be emitted for all running tasks whenever the Overlord is restarted?

kfaraz (approved):

kgyrtkirk:
Thank you @kfaraz  and @uds5501  for reviewing the changes!

PR #18515, opened 2025-09-12
Support unique SQL index names in metadata store

jtuglu1:
### Description

For metadata storage types with low (relatively) upper limits on identifiers(e.g. MySQL with 64 chars), creation of indices can sometimes fail. Because these failures are not critical to the operation of the cluster, they "silently" fail and are logged as warnings. Under large row counts, the lack of these indices can result in performance degradation. This patch 
allows for shorter index names while still maintaining uniqueness using a SHA-based hash of the index table name + columns of the index.

These changes also switch to using the `createIndex` helper tool for every index creation, to make future changes easier. Although this separates the table/index creation into 2 separate transactions, from what I can tell, this shouldn't have any correctness side-effects since the indices are always followed the createTable() statements which would just mean the index may not be created even if the table was created. I can revert/refactor this however..

When enabled, the new format is: `IDX_<SHA of table name + column list>`.

#### Version Forward-compatibility:
- If indices with the legacy name, we don't create a new index. 
#### Version Backwards-compatibility:
- In this case, we could end up creating 2 duplicate indices. I suppose having 2 ≥ n > 0 indices is better than having 0 (in the worst case where the old design wouldn't have an index because of length constraints).

#### Release note

Support short, unique index names in metadata store.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/metadata/MetadataStorageConnector.java
+++ b/processing/src/main/java/org/apache/druid/metadata/MetadataStorageConnector.java
@@ -95,4 +95,12 @@ public interface MetadataStorageConnector
    * SegmentSchema table is created only when CentralizedDatasourceSchema feature is enabled.
    */
   void createSegmentSchemasTable();
+
+  /**
+   * Returns the maximum identifier length for table indices in this MetadataStorageConnector implementation.
+   */
+  default int indexIdentifierLengthLimit()
+  {
+    return Integer.MAX_VALUE;
+  }
 }
--- a/processing/src/main/java/org/apache/druid/metadata/MetadataStorageConnector.java
+++ b/processing/src/main/java/org/apache/druid/metadata/MetadataStorageConnector.java
@@ -99,7 +99,7 @@ public interface MetadataStorageConnector
   /**
    * Returns the maximum identifier length for table indices in this MetadataStorageConnector implementation.
    */
-  default int indexIdentifierLengthLimit()
+  default int getMaxLengthOfIndexName()
   {
     return Integer.MAX_VALUE;
   }
--- a/processing/src/main/java/org/apache/druid/metadata/MetadataStorageConnector.java
+++ b/processing/src/main/java/org/apache/druid/metadata/MetadataStorageConnector.java
@@ -95,12 +95,4 @@ public interface MetadataStorageConnector
    * SegmentSchema table is created only when CentralizedDatasourceSchema feature is enabled.
    */
   void createSegmentSchemasTable();
-
-  /**
-   * Returns the maximum identifier length for table indices in this MetadataStorageConnector implementation.
-   */
-  default int getMaxLengthOfIndexName()
-  {
-    return Integer.MAX_VALUE;
-  }
 }
--- a/processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
+++ b/processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
@@ -21,6 +21,7 @@ package org.apache.druid.metadata;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.druid.common.config.Configs;
 import org.apache.druid.java.util.common.StringUtils;
 
 /**
@@ -31,7 +32,7 @@ public class MetadataStorageTablesConfig
 
   public static MetadataStorageTablesConfig fromBase(String base)
   {
-    return new MetadataStorageTablesConfig(base, null, null, null, null, null, null, null, null, null, null, null);
+    return new MetadataStorageTablesConfig(base, null, null, null, null, null, null, null, null, null, null, null, null);
   }
 
   private static final String DEFAULT_BASE = "druid";
@@ -72,6 +73,9 @@ public class MetadataStorageTablesConfig
   @JsonProperty("segmentSchemas")
   private final String segmentSchemasTable;
 
+  @JsonProperty("useShortIndexNames")
+  private final boolean useShortIndexNames;
+
   @JsonCreator
   public MetadataStorageTablesConfig(
       @JsonProperty("base") String base,
@@ -85,7 +89,8 @@ public class MetadataStorageTablesConfig
       @JsonProperty("audit") String auditTable,
       @JsonProperty("supervisors") String supervisorTable,
       @JsonProperty("upgradeSegments") String upgradeSegmentsTable,
-      @JsonProperty("segmentSchemas") String segmentSchemasTable
+      @JsonProperty("segmentSchemas") String segmentSchemasTable,
+      @JsonProperty("useShortIndexNames") Boolean useShortIndexNames
   )
   {
     this.base = (base == null) ? DEFAULT_BASE : base;
@@ -101,6 +106,7 @@ public class MetadataStorageTablesConfig
     this.auditTable = makeTableName(auditTable, "audit");
     this.supervisorTable = makeTableName(supervisorTable, "supervisors");
     this.segmentSchemasTable = makeTableName(segmentSchemasTable, "segmentSchemas");
+    this.useShortIndexNames = Configs.valueOrDefault(useShortIndexNames, false);
   }
 
   private String makeTableName(String explicitTableName, String defaultSuffix)
@@ -173,5 +179,10 @@ public class MetadataStorageTablesConfig
   public String getSegmentSchemasTable()
   {
     return segmentSchemasTable;
+  }
+
+  public boolean isUseShortIndexNames()
+  {
+    return useShortIndexNames;
   }
 }
--- a/processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
+++ b/processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
@@ -181,6 +181,9 @@ public class MetadataStorageTablesConfig
     return segmentSchemasTable;
   }
 
+  /**
+   * If enabled, this causes table indices to be created with short, unique SHA-based identifiers.
+   */
   public boolean isUseShortIndexNames()
   {
     return useShortIndexNames;
--- a/server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
+++ b/server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
@@ -62,7 +62,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -306,19 +305,16 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
             )
         )
     );
-    final Set<String> createdIndexSet = getIndexOnTable(tableName);
     createIndex(
         tableName,
         "IDX_%S_DATASOURCE_END",
-        List.of("dataSource", quoteColumn("end")),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("dataSource", quoteColumn("end"))
+    );
     createIndex(
         tableName,
         "IDX_%S_DATASOURCE_SEQUENCE",
-        List.of("dataSource", "sequence_name"),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("dataSource", "sequence_name")
+    );
     alterPendingSegmentsTable(tableName);
   }
 
@@ -384,13 +380,11 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
         )
     );
 
-    final Set<String> createdIndexSet = getIndexOnTable(tableName);
     createIndex(
         tableName,
         "IDX_%S_USED",
-        List.of("used"),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("used")
+    );
     createIndex(
         tableName,
         "IDX_%S_DATASOURCE_USED_END_START",
@@ -399,9 +393,8 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
             "used",
             quoteColumn("end"),
             "start"
-        ),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        )
+    );
   }
 
   private void createUpgradeSegmentsTable(final String tableName)
@@ -424,8 +417,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     createIndex(
         tableName,
         "IDX_%S_TASK",
-        List.of("task_id"),
-        getIndexOnTable(tableName)
+        List.of("task_id")
     );
   }
 
@@ -449,8 +441,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     createIndex(
         tableName,
         "IDX_%S_DATASOURCE",
-        List.of("dataSource"),
-        getIndexOnTable(tableName)
+        List.of("dataSource")
     );
   }
 
@@ -496,19 +487,16 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
             )
         )
     );
-    final Set<String> createdIndexSet = getIndexOnTable(tableName);
     createIndex(
         tableName,
         "IDX_%S_ACTIVE_CREATED_DATE",
-        List.of("active", "created_date"),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("active", "created_date")
+    );
     createIndex(
         tableName,
         "IDX_%S_DATASOURCE_ACTIVE",
-        List.of("datasource", "active"),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("datasource", "active")
+    );
   }
 
   private void alterEntryTableAddTypeAndGroupId(final String tableName)
@@ -562,8 +550,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     createIndex(
         tableName,
         "IDX_%S_DATASOURCE_TASK_ALLOCATOR_ID",
-        List.of("dataSource", "task_allocator_id"),
-        getIndexOnTable(tableName)
+        List.of("dataSource", "task_allocator_id")
     );
   }
 
@@ -586,8 +573,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     createIndex(
         tableName,
         "IDX_%S_TASK_ID",
-        List.of("task_id"),
-        getIndexOnTable(tableName)
+        List.of("task_id")
     );
   }
 
@@ -611,8 +597,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     createIndex(
         tableName,
         "IDX_%S_SPEC_ID ",
-        List.of("spec_id"),
-        getIndexOnTable(tableName)
+        List.of("spec_id")
     );
   }
 
@@ -665,8 +650,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     createIndex(
         tableName,
         "IDX_%S_DATASOURCE_UPGRADED_FROM_SEGMENT_ID",
-        List.of("dataSource", "upgraded_from_segment_id"),
-        getIndexOnTable(tableName)
+        List.of("dataSource", "upgraded_from_segment_id")
     );
   }
 
@@ -1015,25 +999,21 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
             )
         )
     );
-    final Set<String> createdIndexSet = getIndexOnTable(tableName);
     createIndex(
         tableName,
         "IDX_%s_KEY_TIME",
-        List.of("audit_key", "created_date"),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("audit_key", "created_date")
+    );
     createIndex(
         tableName,
         "IDX_%s_TYPE_TIME",
-        List.of("type", "created_date"),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("type", "created_date")
+    );
     createIndex(
         tableName,
         "IDX_%s_AUDIT_TIME",
-        List.of("created_date"),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("created_date")
+    );
   }
 
   @Override
@@ -1094,19 +1074,16 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
             )
         )
     );
-    final Set<String> createdIndexSet = getIndexOnTable(tableName);
     createIndex(
         tableName,
         "IDX_%s_FINGERPRINT",
-        List.of("fingerprint"),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("fingerprint")
+    );
     createIndex(
         tableName,
         "IDX_%s_USED",
-        List.of("used", "used_status_last_updated"),
-        createdIndexSet
-    ).ifPresent(createdIndexSet::add);
+        List.of("used", "used_status_last_updated")
+    );
   }
 
   @Override
@@ -1176,16 +1153,15 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
    * @param tableName             Name of the table to create index on
    * @param legacyIndexNameFormat Template to create index ID. If null, uses the format: IDX_{table name}_{columns}.
    * @param indexCols             List of un-escaped column names to be indexed on (case-sensitive).
-   * @param createdIndexSet       Set of pre-existing indices on the table (to avoid extra queries for multiple CREATEs).
    * @return Optional id of the index if successfully created.
    */
-  public Optional<String> createIndex(
+  public void createIndex(
       final String tableName,
       final String legacyIndexNameFormat,
-      final List<String> indexCols,
-      final Set<String> createdIndexSet
+      final List<String> indexCols
   )
   {
+    final Set<String> createdIndexSet = getIndexOnTable(tableName);
     final String shortIndexName = generateShortIndexName(tableName, indexCols);
     String legacyIndexName = legacyIndexNameFormat != null
                              ? StringUtils.toUpperCase(StringUtils.format(legacyIndexNameFormat, tableName))
@@ -1198,10 +1174,10 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     // Avoid creating duplicate indices if an index with either naming convention already exists
     if (createdIndexSet.contains(legacyIndexName)) {
       log.info("Legacy index[%s] on Table[%s] already exists", legacyIndexName, tableName);
-      return Optional.empty();
+      return;
     } else if (createdIndexSet.contains(shortIndexName)) {
       log.info("Short index[%s] on Table[%s] already exists", shortIndexName, tableName);
-      return Optional.empty();
+      return;
     }
 
     final String indexName = tablesConfigSupplier.get().isUseShortIndexNames() ? shortIndexName : legacyIndexName;
@@ -1219,13 +1195,10 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
             return null;
           }
       );
-      createdIndexSet.add(indexName);
-      return Optional.of(indexName);
     }
     catch (Exception e) {
       log.error(e, StringUtils.format("Exception while creating index on table [%s]", tableName));
     }
-    return Optional.empty();
   }
 
   /**
@@ -1295,7 +1268,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
   }
 
   /**
-   * Creates a unique length 24 index name of the format {@code IDX_{20 char SHA of table name + column list}}
+   * Creates a unique index name of length 24 with the format {@code IDX_{20 char SHA of table name + column list}}.
    *
    * @param tableName the table name
    * @param columns   the set of columns to create the index on (case-insensitive)
--- a/server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
+++ b/server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
@@ -392,7 +392,12 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     createIndex(
         tableName,
         "idx_%1$s_datasource_used_end_start",
-        ImmutableList.of("dataSource", "used", StringUtils.format("%send%S", getQuoteString(), getQuoteString()), "start"),
+        ImmutableList.of(
+            "dataSource",
+            "used",
+            StringUtils.format("%send%S", getQuoteString(), getQuoteString()),
+            "start"
+        ),
         createdIndexSet
     );
   }
@@ -1163,7 +1168,8 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
   }
 
   /**
-   * create index on the table with retry if not already exist, to be called after createTable
+   * Create index on the table with retry if not already exist, to be called after createTable
+   * Format of index name is either specified via legacy {@param indexTemplate} or {@code generateSHABasedIndexIdentifier}.
    *
    * @param tableName       Name of the table to create index on
    * @param indexTemplate   Template to create index ID (nullable for forwards compatibility with SHA-based indices)
@@ -1177,27 +1183,31 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
       final Set<String> createdIndexSet
   )
   {
-    final String shaIndexName = generateSHABasedIndexIdentifier(tableName, indexCols);
-    final String legacyIndexName = indexTemplate != null ? StringUtils.format(indexTemplate, tableName) : null;
+    final String shaIndexName = StringUtils.toUpperCase(generateSHABasedIndexIdentifier(tableName, indexCols));
+    final String legacyIndexName = indexTemplate != null ? StringUtils.toUpperCase(StringUtils.format(
+        indexTemplate,
+        tableName
+    )) : null;
+
     // Avoid creating duplicate indices if an index with either naming convention already exists
-    if ((legacyIndexName != null && createdIndexSet.contains(legacyIndexName)) || createdIndexSet.contains(shaIndexName)) {
+    if (legacyIndexName != null && createdIndexSet.contains(legacyIndexName)) {
+      log.info("Legacy index[%s] on Table[%s] already exists", legacyIndexName, tableName);
+      return;
+    } else if (createdIndexSet.contains(shaIndexName)) {
+      log.info("Index[%s] on Table[%s] already exists", shaIndexName, tableName);
       return;
     }
     try {
       retryWithHandle(
           (HandleCallback<Void>) handle -> {
-            if (!createdIndexSet.contains(StringUtils.toUpperCase(shaIndexName))) {
-              String indexSQL = StringUtils.format(
-                  "CREATE INDEX %1$s ON %2$s(%3$s)",
-                  shaIndexName,
-                  tableName,
-                  Joiner.on(",").join(indexCols)
-              );
-              log.info("Creating Index on Table [%s], sql: [%s] ", tableName, indexSQL);
-              handle.execute(indexSQL);
-            } else {
-              log.info("Index [%s] on Table [%s] already exists", shaIndexName, tableName);
-            }
+            String indexSQL = StringUtils.format(
+                "CREATE INDEX %1$s ON %2$s(%3$s)",
+                shaIndexName,
+                tableName,
+                Joiner.on(",").join(indexCols)
+            );
+            log.info("Creating Index on Table [%s], sql: [%s] ", tableName, indexSQL);
+            handle.execute(indexSQL);
             return null;
           }
       );
@@ -1268,15 +1278,23 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
   /**
    * Used to ensure full index coverage in the presence of large index names.
    * Returned indentifiers are of the format: `idx_{tableName}_{SHA of column list}`.
+   *
    * @param tableName the table name
-   * @param columns the set of columns to create the index on (case-insensitive)
+   * @param columns   the set of columns to create the index on (case-insensitive)
    * @return unique index identifier
    */
   @VisibleForTesting
   protected String generateSHABasedIndexIdentifier(String tableName, List<String> columns)
   {
     final String prefix = "idx_" + tableName + "_";
-    final String columnDigest = DigestUtils.sha1Hex(columns.stream().map(StringUtils::toLowerCase).collect(Collectors.joining("_")));
-    return prefix + columnDigest.substring(0, Math.min(columnDigest.length(), getMaxLengthOfIndexName() - prefix.length()));
+    final String columnDigest = DigestUtils.sha1Hex(
+        columns.stream()
+               .map(StringUtils::toLowerCase)
+               .collect(Collectors.joining("_"))
+    );
+    return prefix + columnDigest.substring(
+        0,
+        Math.min(columnDigest.length(), getMaxLengthOfIndexName() - prefix.length())
+    );
   }
 }
--- a/server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
+++ b/server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
@@ -397,7 +397,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     );
   }
 
-  protected void createUpgradeSegmentsTable(final String tableName)
+  private void createUpgradeSegmentsTable(final String tableName)
   {
     createTable(
         tableName,
@@ -988,7 +988,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     );
   }
 
-  protected void createAuditTable(final String tableName)
+  private void createAuditTable(final String tableName)
   {
     createTable(
         tableName,
--- a/server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
+++ b/server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
@@ -596,7 +596,7 @@ public abstract class SQLMetadataConnector implements MetadataStorageConnector
     );
     createIndex(
         tableName,
-        "IDX_%S_SPEC_ID ",
+        "IDX_%S_SPEC_ID",
         List.of("spec_id")
     );
   }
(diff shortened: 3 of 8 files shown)


server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
```diff
@@ -305,6 +307,43 @@ public void testIsTransientException()
...
+      public int indexIdentifierLengthLimit()
+      {
+        return 30;
+      }
+    };
+
+    if (truncate) {
+      Assert.assertEquals(
+          "idx_this_is__datasource_active",
+          testSQLMetadataConnector.getIndexIdentifier(identifierTemplate, longTableName)
```
jtuglu1: I'll switch the test to test `createIndex` behavior directly if we're ok with the `createIndex` changes. Otherwise, I'll leave this as-is for now.

processing/src/main/java/org/apache/druid/metadata/MetadataStorageConnector.java
```diff
@@ -95,4 +95,12 @@ default void exportTable(
    * SegmentSchema table is created only when CentralizedDatasourceSchema feature is enabled.
    */
   void createSegmentSchemasTable();
+
+  /**
+   * Returns the maximum identifier length for table indices in this MetadataStorageConnector implementation.
+   */
+  default int indexIdentifierLengthLimit()
```
kfaraz: Can we use the term `index name` instead of `index identifier` for simplicity in the java methods as well as the new config.

```suggestion
  default int getMaxLengthOfIndexName()
```

processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
```diff
@@ -72,6 +94,9 @@ public static MetadataStorageTablesConfig fromBase(String base)
   @JsonProperty("segmentSchemas")
   private final String segmentSchemasTable;
 
+  @JsonProperty("truncateIndexIdentifiers")
+  private final Boolean truncateIndexIdentifiers;
```
kfaraz: Please use a primitive `boolean` since this will always be non-null.

processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
```diff
@@ -174,4 +201,9 @@ public String getSegmentSchemasTable()
   {
     return segmentSchemasTable;
   }
+
+  public Boolean getTruncateIndexIdentifiers()
```
kfaraz:
```suggestion
  public boolean isTruncateIndexNames()
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -396,13 +409,16 @@ private void createUpgradeSegmentsTable(final String tableName)
...
-            StringUtils.format(
-                "CREATE INDEX idx_%1$s_task ON %1$s(task_id)",
-                tableName
             )
         )
     );
+    final Set<String> createdIndexSet = getIndexOnTable(tableName);
+    createIndex(
+        tableName,
+        "idx_%1$s_task",
```
kfaraz: It seems that this argument will always take the form `idx_%s_<columns>`.
Maybe we should remove this arg altogether, and pass the column names into `getIndexIdentifier` to generate the correct index name.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -419,10 +435,16 @@ public void createRulesTable(final String tableName)
                 + "  PRIMARY KEY (id)\n"
                 + ")",
                 tableName, getPayloadType(), getCollation()
-            ),
-            StringUtils.format("CREATE INDEX idx_%1$s_datasource ON %1$s(dataSource)", tableName)
+            )
         )
     );
+    final Set<String> createdIndexSet = getIndexOnTable(tableName);
```
kfaraz: Let's inline this call if the value is not used again. Same comment in other places too.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1206,4 +1268,16 @@ public static boolean isStatementException(Throwable e)
     return e instanceof StatementException ||
            (e instanceof CallbackFailedException && e.getCause() instanceof StatementException);
   }
+
+  @VisibleForTesting
+  String getIndexIdentifier(String identifierTemplate, String tableName)
```
kfaraz: Since we are adding this new method anyway, let's keep it protected so that metadata store implementation may override it if needed.

@jtuglu1 , as discussed offline, do you think it would make sense to include a fixed size hash of the column names rather than the column names themselves.
Although, that might pose some backward compat issues as we might end up recreating an index even though it already exists with the old name.

kfaraz (commented):
Left some thoughts.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: Yep, agree with the hash of the columns. Should protect against renaming + reordering within the index. 
I suppose for the backwards compatibility we can just check to see if index exists with the column names, and if not, create with the hash/create with the old way.

I'll note this isn't fully backwards compatible, but in this case I suppose having `0 < n ≤ 2` indices for a given (table, column set) is better than having 0.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: I considered this idea, but there's unfortunately a few cases where the column ordering doesn't actually match that which is placed in the index name. Similarly, some index names don't actually match the columns they are composed of (e.g. task_id vs _task).

I think it's probably best to move away from this unstandardized approach anyways, so might be worth just overriding these cases.

kfaraz: Yeah, I agree, let's just update these cases.

processing/src/main/java/org/apache/druid/metadata/MetadataStorageConnector.java
```diff
@@ -95,4 +95,12 @@ default void exportTable(
    * SegmentSchema table is created only when CentralizedDatasourceSchema feature is enabled.
    */
   void createSegmentSchemasTable();
+
+  /**
+   * Returns the maximum identifier length for table indices in this MetadataStorageConnector implementation.
```
kfaraz: Nit:
```suggestion
   * Returns the maximum allowed length of table index names in this MetadataStorageConnector implementation.
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,41 +1168,47 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
+      final String indexTemplate,
       final List<String> indexCols,
       final Set<String> createdIndexSet
   )
   {
+    final String shaIndexName = StringUtils.toUpperCase(generateSHABasedIndexIdentifier(tableName, indexCols));
+    final String legacyIndexName = indexTemplate != null ? StringUtils.toUpperCase(StringUtils.format(
+        indexTemplate,
+        tableName
+    )) : null;
```
kfaraz:
```suggestion
    final String legacyIndexName = indexTemplate != null
           ? StringUtils.toUpperCase(StringUtils.format(indexTemplate, tableName))
           : null;
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,41 +1168,47 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
    * @param tableName       Name of the table to create index on
-   * @param indexName       case-insensitive string index name, it helps to check the existing index on table
+   * @param indexTemplate   Template to create index ID (nullable for forwards compatibility with SHA-based indices)
    * @param indexCols       List of columns to be indexed on
    * @param createdIndexSet
    */
   public void createIndex(
       final String tableName,
-      final String indexName,
+      final String indexTemplate,
```
kfaraz:
```suggestion
      final String legacyIndexNameFormat,
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,41 +1168,47 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
-                    Joiner.on(",").join(indexCols)
-                );
-                log.info("Creating Index on Table [%s], sql: [%s] ", tableName, indexSQL);
-                handle.execute(indexSQL);
-              } else {
-                log.info("Index [%s] on Table [%s] already exists", indexName, tableName);
-              }
-              return null;
-            }
+          (HandleCallback<Void>) handle -> {
```
kfaraz, line 1188: Nit: not sure, but I think the casting can be avoided since there are no other methods which expect a similar lambda signature.
```suggestion
          handle -> {
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -572,10 +603,15 @@ public void createSupervisorsTable(final String tableName)
...
                 + ")",
                 tableName, getSerialType(), getPayloadType()
-            ),
-            StringUtils.format("CREATE INDEX idx_%1$s_spec_id ON %1$s(spec_id)", tableName)
+            )
         )
     );
+    createIndex(
+        tableName,
+        "idx_%1$s_spec_id ",
```
kfaraz: Nit: Since the index name has only one occurrence of the table name, can we just use `%s` instead of `%1$s` across all the invocations of the `createIndex` method?

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -371,15 +378,28 @@ public void createSegmentTable(final String tableName)
...
+        ImmutableList.of("used"),
+        createdIndexSet
+    );
+    createIndex(
+        tableName,
+        "idx_%1$s_datasource_used_end_start",
+        ImmutableList.of(
+            "dataSource",
+            "used",
+            StringUtils.format("%send%S", getQuoteString(), getQuoteString()),
```
kfaraz: can we delegate the quoting to the `createIndex` method itself? It could just quote all the methods by default.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -371,15 +378,28 @@ public void createSegmentTable(final String tableName)
...
+    createIndex(
+        tableName,
+        "idx_%1$s_used",
+        ImmutableList.of("used"),
+        createdIndexSet
+    );
+    createIndex(
+        tableName,
+        "idx_%1$s_datasource_used_end_start",
+        ImmutableList.of(
```
kfaraz: Nit: reads better
```suggestion
        List.of(
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1206,4 +1274,27 @@ public static boolean isStatementException(Throwable e)
...
+
+  /**
+   * Used to ensure full index coverage in the presence of large index names.
+   * Returned indentifiers are of the format: `idx_{tableName}_{SHA of column list}`.
+   *
+   * @param tableName the table name
+   * @param columns   the set of columns to create the index on (case-insensitive)
+   * @return unique index identifier
+   */
+  @VisibleForTesting
```
kfaraz: We can skip this annotation.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1206,4 +1274,27 @@ public static boolean isStatementException(Throwable e)
...
+  protected String generateSHABasedIndexIdentifier(String tableName, List<String> columns)
+  {
+    final String prefix = "idx_" + tableName + "_";
+    final String columnDigest = DigestUtils.sha1Hex(
+        columns.stream()
+               .map(StringUtils::toLowerCase)
+               .collect(Collectors.joining("_"))
+    );
+    return prefix + columnDigest.substring(
```
kfaraz: We should include the table name in generating the digest,
and rather than truncating the digest, we should truncate the table name itself.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,41 +1168,47 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
-              return null;
-            }
+          (HandleCallback<Void>) handle -> {
+            String indexSQL = StringUtils.format(
+                "CREATE INDEX %1$s ON %2$s(%3$s)",
+                shaIndexName,
+                tableName,
+                Joiner.on(",").join(indexCols)
+            );
+            log.info("Creating Index on Table [%s], sql: [%s] ", tableName, indexSQL);
```
kfaraz: Let's log the index name as it might have been truncated.
```suggestion
            log.info("Creating index[%s] on table[%s] using SQL[%s].", indexName, tableName, indexSQL);
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1206,4 +1274,27 @@ public static boolean isStatementException(Throwable e)
     return e instanceof StatementException ||
            (e instanceof CallbackFailedException && e.getCause() instanceof StatementException);
   }
+
+  /**
+   * Used to ensure full index coverage in the presence of large index names.
+   * Returned indentifiers are of the format: `idx_{tableName}_{SHA of column list}`.
```
kfaraz:
```suggestion
   * Creates a unique index name of the format {@code idx_<tableName>_<SHA of column list>}
   * while honoring the {@link #getMaxLengthOfIndexName}.
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1206,4 +1274,27 @@ public static boolean isStatementException(Throwable e)
...
+   *
+   * @param tableName the table name
+   * @param columns   the set of columns to create the index on (case-insensitive)
+   * @return unique index identifier
+   */
+  @VisibleForTesting
+  protected String generateSHABasedIndexIdentifier(String tableName, List<String> columns)
```
kfaraz:
```suggestion
  protected String generateUniqueIndexName(String tableName, List<String> columns)
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,41 +1168,47 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
    */
   public void createIndex(
       final String tableName,
-      final String indexName,
+      final String indexTemplate,
       final List<String> indexCols,
       final Set<String> createdIndexSet
   )
   {
+    final String shaIndexName = StringUtils.toUpperCase(generateSHABasedIndexIdentifier(tableName, indexCols));
```
kfaraz: Should using the new index name be an opt-in feature based on some config?

kfaraz (commented):
Left some final comments.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: I think it should be the default, but can certainly put behind a flag.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: > We should include the table name in generating the digest

Sure.

> rather than truncating the digest, we should truncate the table name itself.

Hmm, that doesn't sound great. I'd prefer to always have the table name. I'm optimizing for avoiding collisions with names within same cluster/table as well as tables between clusters. The SHA was supposed to introduce enough entropy where suffixes didn't overlap (I think the likelihood of this is extremely low as it is currently, given a sufficiently long digest (even 8 bytes should do), even lower than chance of colliding on table name if we were to truncate the table name). I'd suggest leaving for now.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: Sure. For the sake of review, I left these format strings the same to make it easier to view the diff between old/new methods.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: Yeah. I thought this as well, but seems like editor got confused once removed and gave a warning:
Screenshot 2025-09-19 at 11 59 12 AM

I left as-is to remove the warning.

kfaraz: Thanks for checking.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
kfaraz: Yeah, I was concerned about the uniqueness of the hash and thus wanted to always include the full length.
But you make a fair point and even a truncated hash should be good enough.

So, I would suggest _always_ using a fixed size substring of the hash (say 10 or 12 characters), similar to what we do for task IDs.
This ensures that the size of the hash is fixed across all the index names and they all
adhere to the same format.

The current logic might reduce the size of the hash string to zero (or even throw an exception)
if the table name is too long.
To handle that, if the `idx_<table-name>_<12-char-hash>` is still longer than the limit, we just raise an alert.(Failing service startup might be too drastic since we currently don't fail on this scenario.)

jtuglu1: > The current logic might reduce the size of the hash string to zero (or even throw an exception) if the table name is too long.

Sure. I don't think tracking the length per engine is even necessary at that point.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: > It could just quote all the methods by default.

This actually causes some trouble with how engines parse the quotes vs not when creating indices on the tables. With Derby, if the column isn't created in the table with the same quoting format the index creation fails(e.g. "used" vs used in the CT). I've left this as-is for now as it's only a handful of columns.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,41 +1168,48 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
+                "CREATE INDEX %1$s ON %2$s(%3$s)",
+                indexName,
+                tableName,
+                Joiner.on(",").join(indexCols)
+            );
+            log.info("Creating index[%s] on table[%s] using SQL[%s].", indexName, tableName, indexSQL);
+            handle.execute(indexSQL);
+            return null;
           }
       );
```
jtuglu1, line 1199: Considering it's use across multiple createIndex calls, for completeness, I could insert `indexName` into `createdIndexSet`, as I see this was never actually done before. Thoughts @kfaraz?

kfaraz: Yeah, that seems reasonable. 👍🏻

kfaraz: Or you could even return the created index name from the method.
Either way is fine.

processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
```diff
@@ -72,6 +73,9 @@ public static MetadataStorageTablesConfig fromBase(String base)
   @JsonProperty("segmentSchemas")
   private final String segmentSchemasTable;
 
+  @JsonProperty("useUniqueIndexNames")
+  private final Boolean useUniqueIndexNames;
```
kfaraz: Since this will always be non-null
```suggestion
  private final boolean useUniqueIndexNames;
```

processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
```diff
@@ -101,6 +106,7 @@ public MetadataStorageTablesConfig(
     this.auditTable = makeTableName(auditTable, "audit");
     this.supervisorTable = makeTableName(supervisorTable, "supervisors");
     this.segmentSchemasTable = makeTableName(segmentSchemasTable, "segmentSchemas");
+    this.useUniqueIndexNames = Configs.valueOrDefault(useUniqueIndexNames, Boolean.FALSE);
```
kfaraz:
```suggestion
    this.useUniqueIndexNames = Configs.valueOrDefault(useUniqueIndexNames, false);
```

processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
```diff
@@ -174,4 +180,9 @@ public String getSegmentSchemasTable()
   {
     return segmentSchemasTable;
   }
+
+  public boolean getUseUniqueIndexNames()
```
kfaraz:
```suggestion
  public boolean isUseUniqueIndexNames()
```

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
```diff
@@ -305,6 +307,170 @@ public void testIsTransientException()
...
+        "idx_%1$s_used",
+        "idx_%1$s_datasource_used_end_start",
+        "idx_%1$s_datasource_upgraded_from_segment_id"
+    );
+    final List<List<String>> indexColumns = List.of(
+        List.of("used"),
+        List.of(
+            "dataSource",
+            "used",
+            StringUtils.format("%send%S", connector.getQuoteString(), connector.getQuoteString()),
```
kfaraz: Nit: Why capitalize the second `%S`?

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1206,4 +1276,27 @@ public static boolean isStatementException(Throwable e)
...
+   * @param columns   the set of columns to create the index on (case-insensitive)
+   * @return unique index identifier
+   */
+  protected String generateUniqueIndexName(String tableName, List<String> columns)
+  {
+    final String prefix = "idx_" + tableName + "_";
+    final String tableAndColumnDigest = DigestUtils.sha1Hex(
+        StringUtils.format(
+            "%s_%s",
+            tableName,
```
kfaraz: I recall I had advised adding the table name in the generation of the digest.
But now since we are already prefixing the table name, we may omit this.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1206,4 +1276,27 @@ public static boolean isStatementException(Throwable e)
...
+            columns.stream()
+                   .map(StringUtils::toLowerCase)
+                   .collect(Collectors.joining("_")
```
kfaraz: This may cause duplicates since `_` can occur in a column name.
Better use a character that cannot occur in a column name like comma or something.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,43 +1168,51 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
   public void createIndex(
       final String tableName,
-      final String indexName,
+      final String legacyIndexNameFormat,
       final List<String> indexCols,
       final Set<String> createdIndexSet
   )
   {
+    final String shaIndexName = StringUtils.toUpperCase(generateUniqueIndexName(tableName, indexCols));
+    final String legacyIndexName = legacyIndexNameFormat != null
```
kfaraz: Since we use the `legacyIndexName` to ensure that the legacy index doesn't already exist,
I think the `legacyIndexNameFormat` should never be null.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,43 +1168,51 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
   /**
-   * create index on the table with retry if not already exist, to be called after createTable
+   * Create index on the table with retry if not already exist, to be called after createTable
+   * Format of index name is either specified via legacy {@param legacyIndexNameFormat} or {@code generateSHABasedIndexIdentifier}.
    *
    * @param tableName       Name of the table to create index on
-   * @param indexName       case-insensitive string index name, it helps to check the existing index on table
-   * @param indexCols       List of columns to be indexed on
+   * @param legacyIndexNameFormat   Template to create index ID (nullable for forwards compatibility with SHA-based indices)
+   * @param indexCols       List of un-escaped columns to be indexed on (case-sensitive).
```
kfaraz:
```suggestion
   * @param indexCols       List of un-escaped column names to be indexed on (case-sensitive).
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,43 +1168,51 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
 
   /**
-   * create index on the table with retry if not already exist, to be called after createTable
+   * Create index on the table with retry if not already exist, to be called after createTable
+   * Format of index name is either specified via legacy {@param legacyIndexNameFormat} or {@code generateSHABasedIndexIdentifier}.
    *
    * @param tableName       Name of the table to create index on
-   * @param indexName       case-insensitive string index name, it helps to check the existing index on table
-   * @param indexCols       List of columns to be indexed on
+   * @param legacyIndexNameFormat   Template to create index ID (nullable for forwards compatibility with SHA-based indices)
```
kfaraz: How does making this nullable help with forwards compat?

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1206,4 +1276,27 @@ public static boolean isStatementException(Throwable e)
...
+   * Creates a unique index name of the format {@code idx_<tableName>_<SHA of table name + column list>}
+   * with an SHA length of {@value MAX_INDEX_HASH_LENGTH}.
+   *
+   * @param tableName the table name
+   * @param columns   the set of columns to create the index on (case-insensitive)
+   * @return unique index identifier
+   */
+  protected String generateUniqueIndexName(String tableName, List<String> columns)
+  {
+    final String prefix = "idx_" + tableName + "_";
```
kfaraz: Nit: doesn't really affect functionality but probably helps with visualizing the final index names.

```suggestion
    final String prefix = "IDX_" + tableName + "_";
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1038,18 +1089,29 @@ public void createSegmentSchemaTable(final String tableName)
...
-            ),
-            StringUtils.format("CREATE INDEX idx_%1$s_fingerprint ON %1$s(fingerprint)", tableName),
-            StringUtils.format("CREATE INDEX idx_%1$s_used ON %1$s(used, used_status_last_updated)", tableName)
+            )
         )
     );
+    final Set<String> createdIndexSet = getIndexOnTable(tableName);
+    createIndex(
+        tableName,
+        "idx_%s_fingerprint",
```
kfaraz: Since we are touching this anyway, should we capitalize the index name formats?
I feel like it helps with visualizing the actual index name.

```suggestion
        "IDX_%s_FINGERPRINT",
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
kfaraz: I see. Thanks for checking.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -371,15 +378,28 @@ public void createSegmentTable(final String tableName)
...
+        List.of("used"),
+        createdIndexSet
+    );
+    createIndex(
+        tableName,
+        "idx_%s_datasource_used_end_start",
+        List.of(
+            "dataSource",
+            "used",
+            StringUtils.format("%send%S", getQuoteString(), getQuoteString()),
```
kfaraz: Why capitalize the second `%S`?

kfaraz:
```suggestion
            StringUtils.format("%1$send%1$s", getQuoteString()),
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
kfaraz: Side note (non blocking):
there can be another use for this nullability (assuming we don't need it for forward compat).

```java
String legacyIndexName = legacyIndexNameFormat == null
  ? upperCase(StringUtils.format("IDX_%s_%s", tableName, columns.join(`_`)))
  : upperCase(StringUtils.format("IDX_%s", tableName);
```

This will help simplify arguments to `createIndex` method. What do you think?

processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
kfaraz: Setting this flag to false would imply that index names may not be unique anymore which is not exactly right.

Maybe use some other name like `useShortIndexNames` or `useShaIndexNames` or something?

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
kfaraz: Suggestion (doesn't need to be in this PR):

To simplify this, we should probably add a method to `SQLMetadataConnector`,

```java
public String quote(String column)
{
   return StringUtils.format("%1$s%2$s%1$s", getQuoteString(), column);
}
```

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
```diff
@@ -305,6 +307,170 @@ public void testIsTransientException()
...
+            "start"
+        ),
+        List.of("dataSource", "upgraded_from_segment_id")
+    );
```
kfaraz, line 415: Since this is a test setup, we shouldn't have to go through the whole flow of generating index names using formatters etc.

Instead, we should verify the actual index names. Something like:

```java
connector.createSegmentsTable(...);

verifyIndicesPresentOnTable(
   "IDX_ABC",
   "IDX_DEF",
   "IDX_XYZ
);

```

This would make the tests much more readable and exposes the end result while hiding the impl details.

It is okay even if we have to change the test when the logic in the `SQLMetadataConnector.generateIndexName()` method changes
(in fact, it is desirable to have to update the test in cases where say the index name format changes
or SHA generation technique changes).

kfaraz: Please simplify the other test too.

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
```diff
@@ -305,6 +307,170 @@ public void testIsTransientException()
     );
   }
 
+  @ParameterizedTest
+  @ValueSource(booleans = {true, false})
```
kfaraz: These should be separate test cases.

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
```diff
@@ -305,6 +307,170 @@ public void testIsTransientException()
...
+                                                      .collect(Collectors.toList());
+
+    connector.createSegmentTable(segmentsTable);
+    connector.alterSegmentTable();
+
+    // drop legacy/unique indices and create the other type
+    connector.getDBI().withHandle(handle -> {
+      for (int i = 0; i < legacyIndexNames.size(); i++) {
+        handle.execute(StringUtils.toUpperCase(StringUtils.format(
+            "DROP INDEX %s",
```
kfaraz: We should avoid these formatters and instead just drop some of the indexes
(preferably not all so that we can verify the difference in behaviour)

e.g. `DROP INDEX IDX_ABC`.

kfaraz (commented):
The approach makes sense.
Left some more suggestions around the finer points of the impl and tests.

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
jtuglu1: Normally this will capitalize the provided string – it doesn't matter in this case as all supported engines' quotations don't contain any alphabetical chars, so I left it as-is to reduce diff.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: > I recall I had advised adding the table name in the generation of the digest. But now since we are already prefixing the table name, we may omit this.

We aren't prefixing table name. But, I still think it's fine to omit. SHA is the only thing being cut to 10 chars (down from 20 chars). Again, this is still best effort after a certain length of `idx_<table>`. In the worst case, for `datasource_upgraded_from_segment_id`, it's saving 25 characters of space.

If this is still too long for the engine, we throw an exception, log it, and silently continue. I feel like 10 chars is even generous (we could probably do 5 bytes of entropy and be fine).

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: > How does making this nullable help with forwards compat?

I think that's an outdated comment. This was before the config flag was re-added.

jtuglu1: > What do you think?
Sure

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: Previous code does this. This should normally capitalize the string but there's no capitalized equivalent for any engines' quote string. I had left to reduce diff area. Updated.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,46 +1150,56 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
-   * create index on the table with retry if not already exist, to be called after createTable
+   * Create index on the table {@code tableName} with retry if not already exist, to be called after {@link #createTable}.
+   * Format of index name is either specified via legacy {@code legacyIndexNameFormat} or {@link #generateShortIndexName}.
    *
-   * @param tableName       Name of the table to create index on
-   * @param indexName       case-insensitive string index name, it helps to check the existing index on table
-   * @param indexCols       List of columns to be indexed on
-   * @param createdIndexSet
+   * @param tableName             Name of the table to create index on
+   * @param legacyIndexNameFormat Template to create index ID. If null, uses the format: {@code IDX_{table name}_{columns}}.
```
maytasm: nit:
- Add to javadoc on the expected format if legacyIndexNameFormat is specified (since this method will only substitute the tableName and that should be the only placeholder)
- the name legacyIndexNameFormat might be misleading.  the current full index name is not legacy / replaced. It's just two alternative/options ... full vs short name , which we still support / keep both. So maybe just call it fullIndexNameFormat

maytasm: `@param fullIndexNameFormat Template to create index ID. If specified, the only placeholder should be for the tableName. If null, uses the format: {@code IDX_{table name}_{columns}}.`

maytasm (approved):

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: Will wait for @kfaraz to give final review and address comments in one pass.

processing/src/main/java/org/apache/druid/metadata/MetadataStorageTablesConfig.java
```diff
@@ -174,4 +180,9 @@ public String getSegmentSchemasTable()
   {
     return segmentSchemasTable;
   }
+
+  public boolean isUseShortIndexNames()
```
kfaraz, line 187: If you are making further commits, please add a short javadoc here mentioning that if enabled, this causes the tables to use SHA based index names etc.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,46 +1150,56 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
-      final String indexName,
-      final List<String> indexCols,
-      final Set<String> createdIndexSet
+      final String legacyIndexNameFormat,
+      final List<String> indexCols
   )
   {
+    final Set<String> createdIndexSet = getIndexOnTable(tableName);
+    final String shortIndexName = generateShortIndexName(tableName, indexCols);
+    String legacyIndexName = StringUtils.toUpperCase(legacyIndexNameFormat != null
```
kfaraz:
```suggestion
    final String legacyIndexName = StringUtils.toUpperCase(legacyIndexNameFormat != null
```

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1038,18 +1074,26 @@ public void createSegmentSchemaTable(final String tableName)
...
                 tableName, getSerialType(), getPayloadType()
-            ),
-            StringUtils.format("CREATE INDEX idx_%1$s_fingerprint ON %1$s(fingerprint)", tableName),
-            StringUtils.format("CREATE INDEX idx_%1$s_used ON %1$s(used, used_status_last_updated)", tableName)
+            )
         )
     );
+    createIndex(
+        tableName,
+        "IDX_%s_FINGERPRINT",
```
kfaraz, line 1082: Should we update this one and other cases to pass null here?

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
```diff
@@ -1106,46 +1150,56 @@ public ResultSet getIndexInfo(DatabaseMetaData databaseMetaData, String tableNam
...
+            );
+            log.info("Creating index[%s] on table[%s] using SQL[%s].", indexName, tableName, indexSQL);
+            handle.execute(indexSQL);
+            return null;
           }
       );
     }
     catch (Exception e) {
-      log.error(e, StringUtils.format("Exception while creating index on table [%s]", tableName));
+      log.error(e, StringUtils.format("Exception while creating index[%s] on table[%s]", indexName, tableName));
```
kfaraz:
```suggestion
      log.error(e, "Exception while creating index[%s] on table[%s]", indexName, tableName);
```

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
```diff
@@ -305,6 +303,140 @@ public void testIsTransientException()
     );
   }
 
+  @Test
+  public void test_useShortIndexNames_true_tableIndices_areNotAdded_ifExist()
+  {
+    tablesConfig = new MetadataStorageTablesConfig(
+        "druidTest_base_table_name_true",
```
kfaraz: Let's use a shorter base name here to make the verifications simpler:
```suggestion
        "druidTest",
```

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
```diff
@@ -305,6 +303,140 @@ public void testIsTransientException()
...
+      handle.execute("DROP INDEX IDX_B859F28F30657A58518416E445E0AA1A78A68177");
+      handle.execute(StringUtils.format("CREATE INDEX IDX_%1$s_USED ON %1$s(used)", segmentsTable));
+      return null;
+    });
+
+    connector.createSegmentTable(segmentsTable);
+    connector.alterSegmentTable();
+
+    final Set<String> expectedIndices = Sets.newHashSet(
+        StringUtils.format("IDX_%S_USED", segmentsTable),
```
kfaraz: Let's use the final index name where possible since we know what they are going to be.

```suggestion
        "IDX_DRUIDTEST_SEGMENTS_USED",
```

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
```diff
@@ -305,6 +303,140 @@ public void testIsTransientException()
...
+    connector.tearDown();
+  }
+
+  @Test
+  public void test_useShortIndexNames_false_tableIndices_areAdded_IfNotExist()
+  {
+    final String segmentsTable = tablesConfig.getSegmentsTable();
+
+    final Set<String> expectedIndices = Sets.newHashSet(
+        StringUtils.format("IDX_%S_USED", segmentsTable),
```
kfaraz: Please try to use the final index names here as well.

kfaraz (approved):
LGTM 🚀

server/src/test/java/org/apache/druid/metadata/SQLMetadataConnectorTest.java
jtuglu1: This one the connector choose random UUID for base table name. I'll override it but less clean IMO.

server/src/main/java/org/apache/druid/metadata/SQLMetadataConnector.java
jtuglu1: I'd like to keep as-is.

glasser:
> These changes also switch to using the createIndex helper tool for every index creation, to make future changes easier. Although this separates the table/index creation into 2 separate transactions, from what I can tell, this shouldn't have any correctness side-effects since the indices are always followed the createTable() statements which would just mean the index may not be created even if the table was created.

I think this is a correctness improvement — we have been using Druid since v0.12 and found out today that the pending segments indexes (introduced in v0.13) never got created in our DB!

PR #18520, opened 2025-09-12
MSQ: Add "maxThreads" context parameter.

gianm:
Allows users to reduce the number of threads per data server or task. Useful for benchmarking single-threaded performance, or for limiting the impact of a heavyweight query.

### DIFF
--- a/docs/multi-stage-query/reference.md
+++ b/docs/multi-stage-query/reference.md
@@ -418,6 +418,7 @@ The following table lists the context parameters for the MSQ task engine:
 | `storeCompactionState` | REPLACE<br /><br /> When set to true, a REPLACE query stores as part of each segment's metadata a `lastCompactionState` field that captures the various specs used to create the segment. Future compaction jobs skip segments whose `lastCompactionState` matches the desired compaction state. Works the same as [`storeCompactionState`](../ingestion/tasks.md#context-parameters) task context flag. | `false` |
 | `removeNullBytes` | SELECT, INSERT or REPLACE<br /><br /> The MSQ engine cannot process null bytes in strings and throws `InvalidNullByteFault` if it encounters them in the source data. If the parameter is set to true, The MSQ engine will remove the null bytes in string fields when reading the data. | `false` |
 | `maxFrameSize` | SELECT, INSERT or REPLACE<br /><br />Size of frames used for data transfer within the MSQ engine. You generally do not need to change this unless you have very large rows. | `1000000` (1 MB) |
+| `maxThreads` | SELECT, INSERT or REPLACE<br /><br />Maximum number of threads to use for processing. Acts as a cap on the number of threads that workers will use. If not specified, workers use their default thread count based on system configuration. | Not set (no limit) |
 
 ## Joins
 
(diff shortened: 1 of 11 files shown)


docs/multi-stage-query/reference.md
```diff
@@ -418,6 +418,7 @@ The following table lists the context parameters for the MSQ task engine:
...
 | `removeNullBytes` | SELECT, INSERT or REPLACE<br /><br /> The MSQ engine cannot process null bytes in strings and throws `InvalidNullByteFault` if it encounters them in the source data. If the parameter is set to true, The MSQ engine will remove the null bytes in string fields when reading the data. | `false` |
 | `maxFrameSize` | SELECT, INSERT or REPLACE<br /><br />Size of frames used for data transfer within the MSQ engine. You generally do not need to change this unless you have very large rows. | `1000000` (1 MB) |
+| `maxThreads` | SELECT, INSERT or REPLACE<br /><br />Maximum number of threads to use for processing. Acts as a cap on the number of threads that workers will use. If not specified, workers use their default thread count based on system configuration. | Not set (no limit) |
```
cryptoe:
```suggestion
| `maxThreads` | SELECT, INSERT or REPLACE<br /><br />Maximum number of threads to use for processing. Acts as a cap on the number of threads that workers will use. If not specified, workers use their default thread count based on system configuration. If the maxThreads given is greater than the default thread count, the default thread count is chosen.  | Not set (no limit) |
```

multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartWorkerContext.java
```diff
@@ -128,6 +129,11 @@ public class DartWorkerContext implements WorkerContext
     this.tempDir = tempDir;
     this.queryContext = Preconditions.checkNotNull(queryContext, "queryContext");
     this.emitter = emitter;
+    
+    // Compute thread count once in constructor
+    final int baseThreadCount = processingConfig.getNumThreads();
+    final Integer maxThreads = MultiStageQueryContext.getMaxThreads(queryContext);
```
cryptoe, line 135: We should probably have a -1 check since this property is user controlled.

cryptoe (approved):
Minor comments. LGTM otherwise.

docs/multi-stage-query/reference.md
gianm: Updated this to clarify the above and also reflect the zero / negative behavior.

```
| `maxThreads` | SELECT, INSERT or REPLACE<br /><br />Maximum number of threads to use for processing. This only has an effect if it is greater than zero and less than the default thread count based on system configuration. Otherwise, it is ignored, and workers use the default thread count. | Not set (use default thread count) |
```

multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartWorkerContext.java
gianm: I was just thinking about how to handle that. IMO, it's reasonable to treat an invalid number like 0 or -1 the same as `null`. I just changed it.

PR #18521, opened 2025-09-12
Specialized virtual columns for json_merge, json_object.

gianm:
Adds two virtual columns:

1) NestedMergeVirtualColumn, for `json_merge`.
2) NestedObjectVirtualColumn, for `json_object`.

Both of these functions return nested objects. Previously, it was not possible for a virtual column to act as a nested column, so this patch also adds machinery to make this possible. Specifically:

1) `ColumnCache` is updated to return holders for both physical and virtual columns. The subclass `BaseColumnHolder` is used when the column is guaranteed to be a physical column, to minimize the need for casting.

2) `NestedDataComplexColumn` (the abstract superclass of physical nested columns) is split up into interfaces, and `NestedFieldVirtualColumn` is adjusted to use those interfaces with `SelectableColumn#as` rather than looking for `NestedDataComplexColumn`.

3) `ColumnCache` is made into a `ColumnIndexSelector`, and `ColumnSelectorColumnIndexSelector` is removed.

4) `VirtualColumn` is updated to have a single method for creating each selector, accepting both `ColumnSelector` and `ColumnSelectorFactory` (instead of two methods, one accepting each). This is needed because when virtual nested columns are stacked on each other, some levels need one and some levels need the other. For example: `json_merge(nested, json_object(...)`.

This patch also includes an update to the expr implementation of `json_merge` to avoid needless round-trips to JSON strings. Previously, when processing an object, it would convert the object to a JSON string and then parse it again. The new tests validate that error-case behavior remains the same.


processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
```diff
@@ -274,6 +267,10 @@ public ColumnIndexSupplier getIndexSupplier(
       ColumnIndexSelector columnIndexSelector
   )
   {
+    if (isDirectAccess(columnIndexSelector)) {
+      return columnIndexSelector.getIndexSupplier(expression.parsed.get().getBindingIfIdentifier());
+    }
+
```
clintropolis, line 273: this isn't really necessary, `IdentifierExpr` implements `asColumnIndexSupplier` that does this

processing/src/main/java/org/apache/druid/segment/VirtualColumn.java
```diff
@@ -129,11 +192,40 @@ default boolean canVectorize(ColumnInspector inspector)
...
+   * @param factory        object for fetching underlying selectors.
+   * @param columnSelector object for fetching underlying columns, if available. Generally only available for
+   *                       regular segments.
+   * @param offset         offset to use with underlying columns. Available only if columnSelector is available.
    */
+  default SingleValueDimensionVectorSelector makeSingleValueVectorDimensionSelector(
+      DimensionSpec dimensionSpec,
+      VectorColumnSelectorFactory factory,
+      ColumnSelector columnSelector,
+      ReadableVectorOffset offset
```
clintropolis, line 210: should these be `@Nullable` to be consistent with the non-vectorized methods? same comment for other vectorized selector methods

processing/src/main/java/org/apache/druid/segment/VirtualColumns.java
```diff
@@ -297,137 +287,98 @@ public boolean canVectorize(ColumnInspector columnInspector)
...
+   * @param offset         offset to use with underlying columns
    *
    * @throws IllegalArgumentException if the virtual column does not exist (see {@link #exists(String)}
    */
-  @Nullable
   public SingleValueDimensionVectorSelector makeSingleValueDimensionVectorSelector(
       DimensionSpec dimensionSpec,
+      VectorColumnSelectorFactory factory,
       ColumnSelector columnSelector,
       ReadableVectorOffset offset
```
clintropolis, line 301: same comment about `@Nullable`

processing/src/main/java/org/apache/druid/segment/virtual/NestedObjectVirtualColumn.java
```diff
@@ -0,0 +1,497 @@
...
+      } else {
+        return null;
+      }
+    }
+  }
+
+  /**
+   * Represents an expression and a target type. Like a miniature {@link ExpressionVirtualColumn} with no output name.
+   */
+  public static class TypedExpression
```
clintropolis, line 445: feels like a bit of overlap with https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java#L435, though not exactly. I don't think need to worry about in this PR, just thinking about if there is some room for simplifying things later

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
gianm: It was just to align with all the other methods, which have a similar check. It looked odd to me that this one didn't.

processing/src/main/java/org/apache/druid/segment/virtual/NestedObjectVirtualColumn.java
gianm: Yeah, kinda, except this one never parses the expr and `ExpressionVirtualColumn.Expression` does parse it as part of `equals` and `hashCode`. I wanted to avoid parsing the expr in this class.

processing/src/main/java/org/apache/druid/segment/VirtualColumn.java
gianm: These were supposed to not be nullable, because the idea is for vectorized methods, the `ColumnSelector` is always available. I updated the javadoc to match.

processing/src/main/java/org/apache/druid/segment/VirtualColumns.java
gianm: These were supposed to not be nullable, because the idea is for vectorized methods, the `ColumnSelector` is always available. I updated the javadoc to match.

processing/src/main/java/org/apache/druid/segment/VirtualColumn.java
clintropolis: that is true now, will it always be true?

gianm:
Pushed a patch with some fixes for `nested-field` running on columns that contain nested objects but are not "nested columns". (Such as the output of an `expression` virtual column with type `COMPLEX<json>`.)

processing/src/main/java/org/apache/druid/segment/VirtualColumn.java
gianm: Well, who can say. I'm not aware of plans to make it not true. Right now there's two implementations of vector cursor factories— regular segments and columnar frames— and both have `ColumnSelector`.

There's also only one implementation of a specialized vectorized virtual column prior to this patch, `NestedFieldVirtualColumn`. That one doesn't have an implementation for the `VectorColumnSelectorFactory` based vector selector methods-- if they ever got called (i.e. if the `ColumnSelector` based one returned null) an exception would be thrown. So, the one example we have isn't really compatible with a world where the `ColumnSelector` is optional for vector cursors.

For these reasons I'd like to keep it required for now.

clintropolis (approved):
:+1: (assuming ci passes)

PR #18524, opened 2025-09-13
additional validation for projection schema

clintropolis:
### Description
Adds additional validation to projection schemas to ensure no duplicate grouping column or aggregator output names, and also checks that the __time column must be LONG typed.


processing/src/test/java/org/apache/druid/data/input/impl/AggregateProjectionSpecTest.java
```diff
@@ -112,9 +113,15 @@ void testComputeOrdering_granularity()
         ColumnType.LONG,
         TestExprMacroTable.INSTANCE
     );
+    ExpressionVirtualColumn ptEvery10MinLA = new ExpressionVirtualColumn(
+        "ptEvery10MinLA",
+        "timestamp_floor(__time, 'PT10M', null, 'America/Los_Angeles')",
+        ColumnType.LONG,
+        TestExprMacroTable.INSTANCE
+    );
     ExpressionVirtualColumn ptEvery10Min = new ExpressionVirtualColumn(
```
cecemei: since this is not longer pacific time zone, we can just rename it to every10Min

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
```diff
@@ -415,9 +415,73 @@ private static Set<String> computeAndValidateOutputFieldNames(
...
+                  "projection[%s] has granularity[%s] which must be finer than or equal to segment granularity[%s]",
+                  projection.getName(),
+                  projectionGranularity,
+                  segmentGranularity
+              );
+            }
+          }
+        }
+
+        final Map<String, Multiset<String>> fields = new TreeMap<>();
```
cecemei, line 454: nit: maybe `ImmutableMultimap.Builder` is more suitable?

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
```diff
@@ -415,9 +415,73 @@ private static Set<String> computeAndValidateOutputFieldNames(
...
+        int position = 0;
+        for (DimensionSchema grouping : projection.getGroupingColumns()) {
+          final String field = grouping.getName();
+          if (Strings.isNullOrEmpty(field)) {
+            throw DruidException
+                .forPersona(DruidException.Persona.USER)
+                .ofCategory(DruidException.Category.INVALID_INPUT)
+                .build("Encountered grouping column with null or empty name at position[%d]", position);
+          }
+          fields.computeIfAbsent(field, k -> TreeMultiset.create()).add("projection[" + projection.getName() + "] grouping column list");
```
cecemei, line 464: maybe worth use separate map for dimension and aggregator?

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
```diff
@@ -415,9 +415,73 @@ private static Set<String> computeAndValidateOutputFieldNames(
...
+      }
+    }
+  }
+
+  private static Set<String> getFieldsOrThrowIfErrors(Map<String, Multiset<String>> validatedFields)
+  {
     final List<String> errors = new ArrayList<>();
 
-    for (Map.Entry<String, Multiset<String>> fieldEntry : fields.entrySet()) {
+    for (Map.Entry<String, Multiset<String>> fieldEntry : validatedFields.entrySet()) {
```
cecemei, line 499: in line 495, entry.getCount() should always > 1 right?

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
```diff
@@ -415,9 +415,73 @@ private static Set<String> computeAndValidateOutputFieldNames(
...
+                .forPersona(DruidException.Persona.USER)
+                .ofCategory(DruidException.Category.INVALID_INPUT)
+                .build("Encountered aggregator with null or empty name at position[%d]", position);
+          }
+
+          fields.computeIfAbsent(field, k -> TreeMultiset.create()).add("projection[" + projection.getName() + "] aggregators list");
+          position++;
+        }
+
+        getFieldsOrThrowIfErrors(fields);
```
cecemei, line 480: the result is not used? maybe just `validateFieldsOrThrow`

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
```diff
@@ -415,9 +415,73 @@ private static Set<String> computeAndValidateOutputFieldNames(
       }
     }
 
+    return getFieldsOrThrowIfErrors(fields);
+  }
+
+  public static void validateProjections(
```
cecemei, line 426: maybe put a brief javadoc explain it validates:
1. segment granularity has to be at least as coarse as projection granularity.
2. no duplicated grouping column or aggregator names.

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
clintropolis: this is just following the same pattern as an existing method, `computeAndValidateOutputFieldNames` which this now shares some logic with, so i'm going to leave it like it is

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
clintropolis: this is to detect if there are grouping columns and aggs with the same name, see other comment about following the pattern of `computeAndValidateOutputFieldNames` both of which are making that the output schema is valid

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
clintropolis: it is used by `computeAndValidateOutputFieldNames`, which shares this method with projection validation

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
cecemei: actually `ArrayListMultimap.create` since it needs to handle duplicated values

processing/src/test/java/org/apache/druid/data/input/impl/AggregateProjectionSpecTest.java
clintropolis: this renamed `ptEvery10Min` to `ptEvery10MinLA`, and `ptEvery10Min` is now just utc, but both are used by different tests

server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
capistrant: +1 to a javadoc

processing/src/test/java/org/apache/druid/data/input/impl/AggregateProjectionSpecTest.java
cecemei: i mean, pt means pacific time,and this is utc time zone

clintropolis: oh, i guess it wasn't super obvious pt meant pacific time, since it is also `PT10M` i just assumed it was being oddly verbose about the 10 minutes part 😅

PR #18525, opened 2025-09-13
feat: add pre-ingestion filtering based on Kafka headers

### LINKED ISSUE #18476: Allow Header Based Filtering in Kafka Ingestion
### Description

* Druid's transformSpec filtering operates after row deserialization and data ingestion
* Add a feature for pre-ingestion kafka header-based filtering in Supervisor

### Motivation

Please provide the following for the desired feature or change:
- A detailed description of the intended use case, if applicable
- Rationale for why the desired feature/change would be beneficial

suraj-goel:
Fixes #18476

### Description

**Goal / summary**

Add pre-ingestion header-based filtering for Kafka ingestion. The supervisor / task `ioConfig` can now include an optional `headerBasedFilterConfig` which is evaluated *before* parsing: records whose header values do **not** match the configured filter are filtered out (dropped) early, offsets are still advanced so ingestion progress is preserved, and the standard ingestion metric `ingest/events/filtered` is emitted.

Key behavioral notes

- Only **`in`** (InDimFilter) filter specs are accepted — the config validates filter type and rejects unsupported filter types. The `in` filter’s `dimension` is interpreted as the header name and the `values` array as the inclusion set.  
- Header bytes are decoded using a configurable encoding (default `UTF-8`), with a configurable Caffeine-backed cache for decoded strings (default 10,000 entries) to avoid repeated decoding on hot values. Decode failures are logged and treated as `null`.  
- Filtering happens in the `KafkaRecordSupplier.poll(...)` loop; when a record is filtered the code creates a lightweight record marked as filtered (so offsets advance) instead of parsing it.  

#### Load testing
- Performed load testing on 4.3Million events per second.
- CPU improvements were seen when the feature was enabled.

#### Release note

Pre-ingestion header filtering for Kafka: you can now add `headerBasedFilterConfig` to Kafka IO config to filter records *before* parsing. Only **`in`** filters are supported (the `dimension` field maps to the header name; `values` are the allowed header values). Records matching the filter values are **dropped** before ingestion. Filtered events are counted on metric `ingest/events/filtered`.

---

#### Behavioral / design decisions & rationale

- **Pre-parse filtering**: avoids CPU/memory overhead for records that won’t be ingested.  
- **Only `in` filter supported**: simplifies evaluation semantics and ensures predictable behavior. The filter `dimension` maps directly to the header name.  
- **String decoding cache**: Caffeine cache prevents repeated expensive decoding of common header values.  
- **Offset advancement**: filtered records are represented as lightweight placeholders so consumer offsets continue to advance.  

---

 * `org.apache.druid.indexing.kafka.KafkaHeaderBasedFilterEvaluator` (new) — evaluator that decodes header bytes and evaluates the configured Druid `in` filter (uses Caffeine cache for decoded strings).  
 * `org.apache.druid.indexing.kafka.supervisor.KafkaHeaderBasedFilterConfig` (new) — config holder and validator (supports only `InDimFilter`, default encoding, default cache size).  
 * `org.apache.druid.indexing.kafka.KafkaRecordSupplier` (modified) — accepts header-based config, constructs evaluator, and skips/marks filtered records (while advancing offsets).  
 * `org.apache.druid.indexing.kafka.KafkaIndexTaskIOConfig` (modified) — stores `headerBasedFilterConfig` and exposes getter; propagated into task record supplier.  
 * `org.apache.druid.indexing.kafka.KafkaIndexTask` (modified) — passes header-based config into `KafkaRecordSupplier`.  
 * `org.apache.druid.indexing.kafka.KafkaIndexTaskModule` (modified) — registers Jackson subtype for the new config.  
 * `org.apache.druid.indexing.kafka.supervisor.KafkaSupervisorIOConfig` (modified) — accepts and exposes `headerBasedFilterConfig`.  
 * `org.apache.druid.indexing.kafka.supervisor.KafkaSupervisor` (modified) — wires the config through supervisor task creation.  
 * Metric / meters classes (modified) — emit / account for `ingest/events/filtered` (see emitter defaults and `DropwizardRowIngestionMeters` / other row-ingestion meters).  
 * `org.apache.druid.indexing.seekablestream.common.OrderedPartitionableRecord` (modified) — extended to represent filtered records so offsets advance even when records are dropped.  

---

### Checklist

This PR has:

- [x] been self-reviewed.
  - [x] using the concurrency checklist (n/a unless concurrency changes were added)
- [x] added Javadocs for most classes and all non-trivial methods.  
- [x] added unit tests or modified existing tests to cover new code paths.  


m-ghazanfar:
Hi @kfaraz / @abhishekagarwal87 !
Would one of you mind taking a look at this patch when you get a chance, please? We’d really appreciate your review on this.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaRecordSupplier.java
```diff
@@ -172,16 +198,35 @@ public Set<StreamPartition<KafkaTopicPartition>> getAssignment()
...
+        // Create filtered record for offset advancement with filtered=true flag
+        polledRecords.add(new OrderedPartitionableRecord<>(
+            record.topic(),
+            kafkaPartition,
+            record.offset(),
+            Collections.emptyList(), // Empty list for filtered records
+            record.timestamp(),
+            true // Mark as filtered
+        ));
+        continue;
```
cryptoe: This should be an if else block. 
```
boolean shouldInclude=true;
If (header!=null) {
shouldInclude=headerFilterEvaluator.shouldIncludeRecord(record)
}
polledRecords.add(new OrderedPartitionableRecord<>(
            record.topic(),
            kafkaPartition,
            record.offset(),
            Collections.emptyList(), // Empty list for filtered records
            record.timestamp(),
          shouldInclude
        ));
        
 ```
 
 Also we are switching between inclusion and filtering in the naming. Please stick to one convention. I guess filtering seems better.

File: extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaHeaderBasedInclusionConfig.java [hunk 2]
```diff
@@ -0,0 +1,137 @@
...
+import javax.annotation.Nullable;
+
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Kafka-specific implementation of header-based filtering.
+ * Allows filtering Kafka records based on message headers before deserialization.
+ */
+public class KafkaHeaderBasedInclusionConfig
```
cryptoe:
```suggestion
public class KafkaHeaderBasedFilterConfig
```

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaHeaderBasedFilterConfig.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private final DimFilter filter;
+  private final String encoding;
+  private final int stringDecodingCacheSize;
+
+  @JsonCreator
+  public KafkaHeaderBasedInclusionConfig(
+      @JsonProperty("filter") DimFilter filter,
+      @JsonProperty("encoding") @Nullable String encoding,
+      @JsonProperty("stringDecodingCacheSize") @Nullable Integer stringDecodingCacheSize
```
cryptoe, line 53: I understand you need to cache because of performance issues. 
The cache should have limit based on size in memory since that is more useful. 
I think its not a blocker for this PR but make this cache size automatically set lets say 1/20 of the jvm by default. Having such static overrides makes it hard for cluster operators when peon sizes changes makes of new instance types.

docs/ingestion/kafka-ingestion.md
```diff
@@ -125,6 +125,7 @@ For configuration properties shared across all streaming ingestion methods, refe
 |`consumerProperties`|String, Object|A map of properties to pass to the Kafka consumer. See [Consumer properties](#consumer-properties) for details.|Yes. At the minimum, you must set the `bootstrap.servers` property to establish the initial connection to the Kafka cluster.||
 |`pollTimeout`|Long|The length of time to wait for the Kafka consumer to poll records, in milliseconds.|No|100|
 |`useEarliestOffset`|Boolean|If a supervisor is managing a datasource for the first time, it obtains a set of starting offsets from Kafka. This flag determines whether the supervisor retrieves the earliest or latest offsets in Kafka. Under normal circumstances, subsequent tasks start from where the previous segments ended so this flag is only used on the first run.|No|`false`|
+|`headerBasedInclusionConfig`|Object|Configuration for including Kafka records based on their headers before ingestion. See [Header-based inclusion filtering](#header-based-inclusion-filtering) for more details.|No|null|
```
cryptoe: `headerFilter` ?

cryptoe (commented):
Left an initial review.

docs/ingestion/kafka-ingestion.md
suraj-goel: I think there's a confusion in naming as in the initial draft, I have used `headerBasedFilteringConfig`.
This was a bit confusing as the name suggests that we are filtering based on the provided filters in the config.

In reality, we are including the records based on the config and hence the name - `headerBasedInclusionConfig`

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaRecordSupplier.java
suraj-goel: We are also passing empty kafka record in case of shouldInclude is false.
imo, current implementation looks cleaner. Let me know if otherwise.

Regarding the inclusion and filtering in naming. Please refer to https://github.com/apache/druid/pull/18525#discussion_r2448146529

cryptoe: In that case could you please change it to a if else block. The current impl is a bit confusing. 
`continues` are a bit easy to miss in a refactor.

docs/ingestion/kafka-ingestion.md
cryptoe: https://druid.apache.org/docs/latest/ingestion/ingestion-spec#filter 
Druid ingestion filtering is generally inclusive.

suraj-goel: Sure, In that case I can revert rename commit - https://github.com/apache/druid/pull/18525/commits/fb0b1db503d4ea62a03936becc2ba72474130367

suraj-goel: Changed it back to `headerBasedFilteringConfig`

suraj-goel: Changed it to `headerBasedFilterConfig`

docs/ingestion/kafka-ingestion.md
```diff
@@ -264,6 +265,105 @@ The following example shows a supervisor spec with idle configuration enabled:
...
+      "inputFormat": {
+        "type": "json"
+      },
+      "consumerProperties": {
+        "bootstrap.servers": "localhost:9092"
+      },
+      "headerBasedFilterConfig": {
+        "filter": {
+          "type": "in",
+          "dimension": "environment",
```
FrankChen021, line 337: I think the name 'dimension' here is ambiguous. 'dimension' is a concept after a record is deserialized, but here we're defining a filter before deserialization, itconfuses users and they may ask what's the relationship with the dimension above defined in `dimensionSpec`.

Name such as `headerName` or `header` is better

docs/ingestion/kafka-ingestion.md
```diff
@@ -264,6 +265,105 @@ The following example shows a supervisor spec with idle configuration enabled:
...
 
+#### Header-based filtering
+
+Header-based filtering allows you to filter Kafka records based on their headers before ingestion, reducing the amount of data processed and improving ingestion performance. This is particularly useful when you want to ingest only a subset of records from a Kafka topic based on header values.
+
+The following table outlines the configuration options for `headerBasedFilterConfig`:
+
+|Property|Type|Description|Required|Default|
+|--------|----|-----------|--------|-------|
+|`filter`|Object|A Druid filter specification that defines which records to include based on header values. Only `in` filters are supported.|Yes||
```
FrankChen021:
```suggestion
|`filter`|Object|A Druid filter specification that defines which records to include if the evaluation result of this filter is true based on header values. Only `in` filter is supported.|Yes||
```

docs/ingestion/kafka-ingestion.md
```diff
@@ -264,6 +265,105 @@ The following example shows a supervisor spec with idle configuration enabled:
...
+      "type": "kafka",
+      "maxRowsPerSegment": 5000000
+    }
+  }
+}
+```
+</details>
+
+In this example:
+- Only records with `environment` header values of `production` or `staging` will be ingested
```
FrankChen021: based on the implementation below, the description here is not incomplete.
- records that do not have the environment headers are also included
- records that have environment header but the header value is null are included
- records that have environment header and value, but fail to decode value, are also included

But from the `filter` itself semantic, I think these 3 are contradiction. My understanding is that only records that have environment header and has either production or staging value will be included, meaning that above records should NOT be included

correct me if I'm wrong.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaHeaderBasedFilterEvaluator.java
```diff
@@ -0,0 +1,141 @@
...
+      return true;
+    }
+
+    String headerValue = getDecodedHeaderValue(header.value());
+    // Permissive behavior: failed to decode header value
+    if (headerValue == null) {
+      return true;
+    }
+
+    return filterValues.contains(headerValue);
```
FrankChen021: I think it's a little bit hard to add a new filter here. The implementation internally uses InDimFilter, but actually we don't use any methods or capability provided by current filter mechansim.

I think a better way is that:


```java
interface Filter {
   String getHeader();
    boolean shouldInclude(String headerValue)
}

class InFilter {
    public InFilter(String header, Set<String> values) {
    }

    public boolean shouldInclude(String headerValue) {
         return values.contains(headerValue);
    }
}
```

and in this `evaluateInclusion(Headers headers)` method, we just do sth like:
```java

...
Header header = headers.lastHeader(filter.getHeader());

String headerValue = getDecodedHeaderValue(header.value());

return this.filter.shouldInclude(headerValue);
```

By this way, when we want to add a new filter, we only add new implementation of Filter without changing any other parts of the code.

docs/ingestion/kafka-ingestion.md
FrankChen021: one more we need to state in the document is that if there's multiple values of a header, the last header is used(based on current implementation)

docs/ingestion/kafka-ingestion.md
suraj-goel: The `filter` in `headerBasedFilterConfig` is of type `DimFilter` (org.apache.druid.query.filter).
This is similar to what we have in transformSpec. I agree that dimension might be confusing but I wanted to avoid creation of another Filter Class and re-use the existing Filter.

Do you think we should create a new Filter class ?
imo, re-using it will be a good enough trade-off.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaHeaderBasedFilterEvaluator.java
suraj-goel: This is a good suggestion but I have added a similar comment here - https://github.com/apache/druid/pull/18525#discussion_r2461024627

docs/ingestion/kafka-ingestion.md
suraj-goel: Sure, I will update this in the docs.

suraj-goel: I have improved the documentation to clearly outlined the permissive behavior of header based filtering and role of `dimension` field w.r.t. header.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaHeaderBasedFilterEvaluator.java
suraj-goel: I have improved the extensibility and addition of new filters in the latest commit.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaHeaderBasedFilterConfig.java
```diff
@@ -0,0 +1,137 @@
...
+      InDimFilter.class
+  );
+
+  private final DimFilter filter;
+  private final String encoding;
+  private final int stringDecodingCacheSize;
+
+  @JsonCreator
+  public KafkaHeaderBasedFilterConfig(
+      @JsonProperty("filter") DimFilter filter,
```
FrankChen021, line 51: Since we have already defined our filter interface, here we can directly use our filter interface. No need to use DimFilter at all.

suraj-goel: We haven't defined our filter interface. Refer to [this](https://github.com/apache/druid/pull/18525#discussion_r2461024627) comment

suraj-goel:
I have deployed this change on Production clusters ingesting around 10Million events per second.
We are seeing improved CPU due to decrease in throwAway events and increase in filtered events.

cc: @cryptoe

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaHeaderBasedFilterEvaluator.java
```diff
@@ -0,0 +1,146 @@
...
+  {
+    // Permissive behavior: missing headers result in inclusion
+    if (headers == null) {
+      return true;
+    }
+
+    Header header = headers.lastHeader(headerName);
+    
+    // Permissive behavior: header is null or empty
+    if (header == null || header.value() == null) {
```
FrankChen021, line 115: [P1] Keep empty and malformed headers on the permissive path

The documented permissive contract is not met here. A zero-length value is not caught and decodes to an empty string; additionally, `new String(byte[], Charset)` replaces malformed input rather than throwing, so `getDecodedHeaderValue` does not return null for invalid bytes. Unless the empty or replacement string is explicitly allowed, these records are silently discarded. Check for zero-length values and decode with a `CharsetDecoder` configured with `CodingErrorAction.REPORT`, then include the record on failure.

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMeters.java
```diff
@@ -75,6 +76,9 @@ default long getProcessedBytes()
   long getThrownAway();
   void incrementThrownAway();
 
+  long getFiltered();
```
FrankChen021, line 79: [P1] Preserve the RowIngestionMeters extension contract

`RowIngestionMeters` is annotated `@ExtensionPoint`, but these new abstract methods make existing external implementations source-incompatible and cause `AbstractMethodError` when the filtered Kafka path calls `incrementFiltered`. Replacing the five-argument `RowIngestionMetersTotals` constructor can likewise cause `NoSuchMethodError` in previously compiled implementations. Provide default zero/no-op methods and retain a five-argument constructor overload delegating to `filtered = 0`.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 3 |
| P2 | 3 |
| P3 | 0 |
| Total | 6 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 3 |
| P2 | 3 |
| P3 | 0 |
| Total | 6 |

Reviewed 46 of 46 changed files. The original thread is handled in a separate summary-only review.

---

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaHeaderBasedFilterEvaluator.java
```diff
@@ -0,0 +1,146 @@
...
+   * Creates a new KafkaHeaderBasedFilterEvaluator with the given configuration.
+   *
+   * @param headerBasedFilterConfig the configuration containing filter, encoding, and cache settings
+   * @throws IllegalArgumentException if the filter type is not supported
+   */
+  public KafkaHeaderBasedFilterEvaluator(KafkaHeaderBasedFilterConfig headerBasedFilterConfig)
+  {
+    this.encoding = Charset.forName(headerBasedFilterConfig.getEncoding());
+    this.stringDecodingCache = Caffeine.newBuilder()
+        .maximumSize(headerBasedFilterConfig.getStringDecodingCacheSize())
```
FrankChen021, line 58: [P1] Bound the decoding cache by retained bytes

`maximumSize` bounds entry count, not retained memory. Each `ByteBuffer` key retains the entire Kafka header byte array, while the value retains its decoded string. With the default 10,000 entries, distinct large headers can retain gigabytes and exhaust a task heap; the documentation even recommends increasing the count for unique values. Use a byte-based `maximumWeight`/weigher and/or skip caching oversized values.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaRecordSupplier.java
```diff
@@ -93,18 +99,38 @@ public KafkaRecordSupplier(
       boolean multiTopic
   )
   {
-    this(getKafkaConsumer(sortingMapper, consumerProperties, configOverrides), multiTopic);
+    this(getKafkaConsumer(sortingMapper, consumerProperties, configOverrides), multiTopic, (KafkaHeaderBasedFilterConfig) null);
```
FrankChen021, line 102: [P2] Apply the header filter during Kafka sampling

`KafkaSamplerSpec#createRecordSupplier` still calls this four-argument overload, which hardcodes a null header filter. Sampling a supervisor spec containing `headerBasedFilterConfig` therefore displays records that actual ingestion tasks discard. Pass `kafkaSupervisorIOConfig.getheaderBasedFilterConfig()` from the sampler.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaHeaderBasedFilterConfig.java
```diff
@@ -0,0 +1,137 @@
...
+    validateSupportedFilter(this.filter);
+  }
+
+  /**
+   * Validates that the filter is one of the supported types.
+   * Only 'in' filters are supported for direct evaluation.
+   */
+  private void validateSupportedFilter(DimFilter dimFilter)
+  {
+    if (!SUPPORTED_FILTER_TYPES.contains(dimFilter.getClass())) {
```
FrankChen021, line 78: [P2] Reject unsupported InDimFilter extraction functions

This class-only validation accepts an `InDimFilter` with a non-null `extractionFn`, but `InDimFilterHandler` only performs a raw string-set lookup and silently ignores that function. Such a valid serialized Druid filter is evaluated with different semantics and may discard the wrong records. Either reject non-null extraction functions here or apply them in the handler.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaRecordSupplier.java
```diff
@@ -172,16 +198,34 @@ public Set<StreamPartition<KafkaTopicPartition>> getAssignment()
...
+      KafkaTopicPartition kafkaPartition = new KafkaTopicPartition(multiTopic, record.topic(), record.partition());
+
+      // Apply header filter if configured
+      if (headerFilterEvaluator != null && !headerFilterEvaluator.shouldIncludeRecord(record)) {
+        // Create filtered record for offset advancement with filtered=true flag
+        polledRecords.add(new OrderedPartitionableRecord<>(
+            record.topic(),
+            kafkaPartition,
+            record.offset(),
+            Collections.emptyList(), // Empty list for filtered records
```
FrankChen021, line 212: [P2] Keep filtered payload bytes in input-byte accounting

Replacing the payload with an empty list means `StreamChunkParser` can increment only the filtered-event counter; it never sees `record.value()` and never increments `processedBytes`. Consequently, `ingest/input/bytes` excludes all header-filtered payload bytes, contrary to its documented all-input-bytes contract. Carry the payload byte count on the marker, or retain the entity while bypassing parsing, and increment the byte meter.

FrankChen021 (commented):
Follow-up handled: the author’s clarification is correct. `HeaderFilterHandler` is a runtime evaluation adapter rather than a Jackson-polymorphic configuration contract, so keeping `DimFilter` in `KafkaHeaderBasedFilterConfig` is appropriate. Separate issues found during the broader review are reported independently.

Reviewed 46 of 46 changed files.

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/common/OrderedPartitionableRecord.java
```diff
@@ -50,15 +51,16 @@ public OrderedPartitionableRecord(
...
   }
 
   public OrderedPartitionableRecord(
       String stream,
       PartitionIdType partitionId,
       SequenceOffsetType sequenceNumber,
       List<RecordType> data,
-      Long timestamp
+      Long timestamp,
+      boolean filtered
```
FrankChen021, line 78: [P1] Preserve the public five-argument constructor

This changes the public `(String, PartitionIdType, SequenceOffsetType, List<RecordType>, Long)` constructor into a six-argument signature. Existing stream extensions compiled against this class will fail with `NoSuchMethodError`, and source extensions using the constructor will no longer compile. Keep the old overload and delegate it to the new one with `filtered = false`.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaSamplerSpec.java
```diff
@@ -74,7 +74,9 @@ protected KafkaRecordSupplier createRecordSupplier()
           objectMapper,
           kafkaSupervisorIOConfig.getConfigOverrides(),
           kafkaSupervisorIOConfig.isMultiTopic(),
-          null
+          null,
+          // Apply the same header-based filter during sampling so sampled records match what ingestion tasks keep.
+          kafkaSupervisorIOConfig.getheaderBasedFilterConfig()
```
FrankChen021, line 79: [P2] Skip filtered records in Kafka sampling

The filter is still ineffective for sampling. `SeekableStreamSamplerSpec` wraps this supplier in `RecordSupplierInputSource`, whose iterator consumes `recordIterator.next().getData()` without checking `OrderedPartitionableRecord.isFiltered()`. Because `KafkaRecordSupplier` intentionally retains the filtered payload in the marker, sample queries will parse and return records that ingestion tasks discard. Make the sampler input source skip filtered records while retaining payloads for task byte accounting, or use a sampler-specific record path.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaHeaderBasedFilterEvaluator.java
```diff
@@ -0,0 +1,182 @@
...
+      // Do not cache oversized values so the cache cannot retain unbounded heap for distinct large headers.
+      if (headerBytes.length > MAX_CACHEABLE_HEADER_BYTES) {
+        return decodeStrict(headerBytes);
+      }
+      ByteBuffer key = ByteBuffer.wrap(headerBytes);
+      return stringDecodingCache.get(key, k -> decodeStrict(headerBytes));
+    }
+    catch (Exception e) {
+      // Includes UncheckedIOException wrapping CharacterCodingException for malformed/unmappable input.
+      log.warn(e, "Failed to decode header bytes, treating as null");
```
FrankChen021: [P2] Rate-limit malformed-header warnings

`getDecodedHeaderValue` is on the record-processing path, but this catch logs the decoding exception at WARN, including a stack trace, for every malformed or unmappable header value. A high-rate topic with bad bytes or an encoding mismatch can therefore flood logs and spend substantial CPU and I/O on logging, degrading ingestion and obscuring useful signals. Rate-limit or deduplicate this warning, or lower the per-record path to debug while preserving the permissive include behavior.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 2 |
| P3 | 0 |
| Total | 3 |

Reviewed 47 of 47 changed files.

The findings are posted inline at the changed-file locations.



---

indexing-service/src/main/java/org/apache/druid/indexing/common/stats/TaskRealtimeMetricsMonitor.java
```diff
@@ -105,6 +105,9 @@ public boolean doMonitor(ServiceEmitter emitter)
 
     emitter.emit(builder.setMetric("ingest/events/processed", rowIngestionMetersTotals.getProcessed() - previousRowIngestionMetersTotals.getProcessed()));
 
+    final long filtered = rowIngestionMetersTotals.getFiltered() - previousRowIngestionMetersTotals.getFiltered();
+    emitter.emit(builder.setMetric("ingest/events/filtered", filtered));
```
FrankChen021, line 109: [P2] Update the default logging-emitter allowlist

`LoggingEmitter` drops every service metric not present in its bundled `processing/src/main/resources/loggingEmitterAllowedMetrics.json` when `shouldFilterMetrics` is enabled without a custom allowlist. This new `ingest/events/filtered` metric is not in that resource, so the documented metric silently disappears for that supported configuration even though the other emitter defaults were updated. Add it to the bundled allowlist and cover the default-filtered path.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorIOConfig.java
```diff
@@ -77,6 +78,7 @@ public KafkaSupervisorIOConfig(
       @JsonProperty("earlyMessageRejectionPeriod") Period earlyMessageRejectionPeriod,
       @JsonProperty("lateMessageRejectionStartDateTime") DateTime lateMessageRejectionStartDateTime,
       @JsonProperty("configOverrides") KafkaConfigOverrides configOverrides,
+      @JsonProperty("headerBasedFilterConfig") KafkaHeaderBasedFilterConfig headerBasedFilterConfig,
```
FrankChen021, line 81: [P2] Preserve the existing public constructor

Adding `headerBasedFilterConfig` to this public constructor removes the previous JVM descriptor, so callers compiled against the existing Kafka indexing-service API fail with `NoSuchMethodError` (and source callers stop compiling) even when filtering is unused. Keep the old overload delegating `null`, as this change already does for `KafkaIndexTaskIOConfig` and `OrderedPartitionableRecord`.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |



---

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaHeaderBasedFilterConfigIntegrationTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+    // Verify they're equivalent
+    Assertions.assertEquals(originalFilter.getFilter(), deserializedFilter.getFilter());
+    Assertions.assertEquals(originalFilter.getEncoding(), deserializedFilter.getEncoding());
+    Assertions.assertEquals(originalFilter.getStringDecodingCacheSize(), deserializedFilter.getStringDecodingCacheSize());
+
+    // Test that the deserialized filter works
+    evaluator = new KafkaHeaderBasedFilterEvaluator(deserializedFilter);
+    ConsumerRecord<byte[], byte[]> record = createRecord("events", 0, 100L, headers("environment", "production"));
+    Assertions.assertFalse(evaluator.shouldIncludeRecord(record), "Deserialized filter should work");
```
FrankChen021: [P1] Serialization test rejects an explicitly included value

The filter contains both "production" and "staging", and the evaluator includes a record when the header value is one of those values. This record therefore makes shouldIncludeRecord return true, so the newly added test fails deterministically despite the message saying the deserialized filter should work. Change this assertion to assertTrue, or use a non-matching header value if the test is intended to verify rejection.

FrankChen021 (commented):
## 🟡 Changes recommended

Updated-head static review rechecked the previously reported filtering, sampling, byte-accounting, compatibility, constructor, and logging-allowlist changes. One deterministic test failure remains and should be fixed before merging.

Reviewed 51 of 51 current changed files.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |



---

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskIOConfig.java
```diff
@@ -66,7 +68,8 @@ public KafkaIndexTaskIOConfig(
       @JsonProperty("configOverrides") @Nullable KafkaConfigOverrides configOverrides,
       @JsonProperty("multiTopic") @Nullable Boolean multiTopic,
       @JsonProperty("refreshRejectionPeriodsInMinutes") Long refreshRejectionPeriodsInMinutes,
-      @JsonProperty("boundedStreamConfig") @Nullable BoundedStreamConfig boundedStreamConfig
+      @JsonProperty("boundedStreamConfig") @Nullable BoundedStreamConfig boundedStreamConfig,
+      @JsonProperty("headerBasedFilterConfig") @Nullable KafkaHeaderBasedFilterConfig headerBasedFilterConfig
```
FrankChen021, line 72: [P1] Preserve the previous full constructor signature

Adding `headerBasedFilterConfig` changes the public 17-argument constructor's JVM descriptor. The overload added below only preserves the separate 12-argument convenience constructor, so extensions or other callers compiled against the previous full signature fail with `NoSuchMethodError` even when filtering is unused. Retain an overload with the old argument list that delegates with a null filter.

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
```diff
@@ -44,32 +45,66 @@ public RowIngestionMetersTotals(
       @JsonProperty("processedWithError") long processedWithError,
       @JsonProperty("thrownAway") long thrownAway,
       @JsonProperty("thrownAwayByReason") @Nullable Map<String, Long> thrownAwayByReason,
-      @JsonProperty("unparseable") long unparseable
+      @JsonProperty("unparseable") long unparseable,
+      @JsonProperty("filtered") long filtered
```
FrankChen021, line 49: [P1] Retain the old totals constructor overload

Adding `filtered` changes the public JSON-creator constructor from `(long, long, long, long, Map<String, Long>, long)` to a seven-argument signature, but no six-argument overload with both `thrownAway` and `thrownAwayByReason` remains. Already-compiled callers using the previous constructor will fail with `NoSuchMethodError` (and source callers will no longer compile); keep that overload and default `filtered` to zero.

FrankChen021 (commented):
## 🟡 Changes recommended

Updated-head static review rechecked the current head and the previously reported filtering, sampling, offset, byte-accounting, metric-allowlist, and serialization fixes. Two public constructor descriptors still change without preserving the prior overloads, so already-compiled callers can fail to link even when header filtering is unused.

Reviewed 51 of 51 current changed files.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 0 |
| P3 | 0 |
| Total | 2 |



---

This is an automated review by Codex GPT-5.6 Luna(Max)

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
FrankChen021: No such case. Ignore this. AI is too conservative

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskIOConfig.java
FrankChen021: No such case. Ignore this. AI is too conservative

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaHeaderBasedFilterConfig.java
```diff
@@ -0,0 +1,146 @@
...
+  @JsonCreator
+  public KafkaHeaderBasedFilterConfig(
+      @JsonProperty("filter") DimFilter filter,
+      @JsonProperty("encoding") @Nullable String encoding,
+      @JsonProperty("stringDecodingCacheSize") @Nullable Integer stringDecodingCacheSize
+  )
+  {
+    this.filter = Preconditions.checkNotNull(filter, "filter cannot be null");
+    this.encoding = encoding != null ? encoding : StandardCharsets.UTF_8.name();
+    this.stringDecodingCacheSize = stringDecodingCacheSize != null ? stringDecodingCacheSize : 10_000;
```
FrankChen021, line 58: `stringDecodingCacheSize` isn't validated to be non-negative here, unlike `encoding` and `filter` in the same constructor (which fail cleanly via `IllegalArgumentException`/`InvalidInput.exception`). A spec with e.g. `"stringDecodingCacheSize": -1` is accepted at submission time, but later blows up with an opaque `IllegalArgumentException("maximum size must not be negative")` from `Caffeine.newBuilder().maximumSize(...)` every time a record supplier is constructed (supervisor start, and every task launch) — effectively a crash loop instead of a clean validation error. Please add a non-negative check here.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaHeaderBasedFilterEvaluator.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }
+
+  /**
+   * Decodes bytes with the configured charset, reporting (rather than silently replacing with U+FFFD) malformed
+   * or unmappable input, so that invalid bytes surface as a failure instead of a bogus match value.
+   */
+  private String decodeStrict(byte[] headerBytes)
+  {
+    final CharsetDecoder decoder = encoding.newDecoder()
```
FrankChen021: `decodeStrict` allocates a new `CharsetDecoder` on every call (`encoding.newDecoder()...`). This runs on every cache miss and, for headers above `MAX_CACHEABLE_HEADER_BYTES`, on every single record. Since polling is single-threaded per `KafkaRecordSupplier`, this can be a reused instance field with `.reset()` called before each use instead of allocating fresh each time — worth doing given this PR's whole premise is reducing per-record CPU cost on a hot path handling millions of events/sec.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskIOConfig.java
```diff
@@ -188,6 +230,14 @@ public boolean isMultiTopic()
     return multiTopic;
   }
 
+
+  @JsonProperty
+  @Nullable
+  public KafkaHeaderBasedFilterConfig getheaderBasedFilterConfig()
```
FrankChen021: Getter is named `getheaderBasedFilterConfig()` (lowercase `h`) instead of the standard bean casing `getHeaderBasedFilterConfig()` used by every other getter in this class (`getConfigOverrides()`, etc.). JSON round-trips fine today only because the `@JsonCreator` constructor param and Jackson's inferred property name from this getter happen to both resolve to `headerBasedFilterConfig`, but the wrong casing will trip up any caller/tooling using standard naming. Same issue in `KafkaSupervisorIOConfig.getheaderBasedFilterConfig()` (line 239) — please fix both.

FrankChen021 (changes requested):
A few issues from review that I'd like addressed before merging:

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskIOConfig.java
FrankChen021, line 72: [P1] Retain the previous full IO-config constructor

**Finding:** Adding headerBasedFilterConfig to the public constructor that previously ended at boundedStreamConfig removes the old JVM descriptor. The overload below only preserves the separate convenience constructor; code or extensions compiled against the prior full constructor will fail with NoSuchMethodError at task/spec construction even when filtering is not configured.

**Suggestion:** Add a public overload with the old full parameter list that delegates to this constructor with a null header filter, and cover the descriptor in an API-compatibility test.

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
FrankChen021, line 49: [P1] Retain the previous full totals constructor

**Finding:** The @JsonCreator is also a public constructor; adding filtered changes the old (long, long, long, long, Map<String, Long>, long) descriptor, while the added overloads preserve only the five-argument constructors and the new six-argument no-map variant. Previously compiled callers using the full totals constructor can fail with NoSuchMethodError when building reports, independent of header filtering.

**Suggestion:** Retain a six-argument overload with thrownAwayByReason and no filtered argument that delegates with filtered=0 while leaving the Jackson creator on the new signature.

indexing-service/src/main/java/org/apache/druid/indexing/common/stats/TaskRealtimeMetricsMonitor.java
FrankChen021, line 109: [P2] Clear the reason dimension for filtered metrics

**Finding:** If this monitor interval emits any thrownAway-by-reason event, the shared ServiceMetricEvent.Builder still contains that last reason when the new ingest/events/filtered event is emitted. ServiceMetricEvent copies all user dimensions, so the filtered event is tagged with an unrelated reason (and the same stale dimension leaks into subsequent metrics), which corrupts logging or custom-emitter output and can split or mislabel metric series.

**Suggestion:** Build filtered and other non-reason metrics with a fresh builder or explicitly clear the reason dimension after the per-reason emissions, and add a test covering both counters in one monitor tick.

FrankChen021 (commented):
## 🟡 Changes recommended

Updated-head static review started with the incremental diff from 57f2c7a1b2f9435aa8b4a6c77e5844e35d5d2a9b to 553fd6f0666a3129e5134a58e375ad67d1a8a8b8, then rechecked the complete current diff and surrounding consumer, sampling, parsing, metrics, serialization, and compatibility paths. The incremental decoder reuse, getter casing, cache-size validation, and related tests are consistent with the existing lifecycle; the earlier filtering, sampling, byte-accounting, logging-allowlist, and serialization fixes remain in place. Two public constructor descriptors still lack the prior overloads, and the new filtered metric can inherit a stale reason dimension when thrown-away rows are emitted in the same monitor cycle.

Reviewed 51 of 51 changed files.

Static review only; no build or test commands were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 1 |
| P3 | 0 |
| Total | 3 |



---

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskIOConfig.java
suraj-goel: addressed

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
suraj-goel: addressed

indexing-service/src/main/java/org/apache/druid/indexing/common/stats/TaskRealtimeMetricsMonitor.java
suraj-goel: addressed

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskIOConfig.java
FrankChen021: I’m withdrawing this finding after rechecking Druid’s API contract. `KafkaIndexTaskIOConfig` and this constructor are not annotated `@PublicApi` or `@ExtensionPoint`; unannotated constructors are explicitly outside the compatibility guarantee. The descriptor change is real, but it does not require a compatibility overload.

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
FrankChen021: I’m withdrawing this finding after rechecking Druid’s API contract. `RowIngestionMetersTotals` and this constructor are not annotated `@PublicApi` or `@ExtensionPoint`; unannotated constructors are explicitly outside the compatibility guarantee. The descriptor change is real, but it does not require a compatibility overload.

indexing-service/src/main/java/org/apache/druid/indexing/common/stats/TaskRealtimeMetricsMonitor.java
FrankChen021: I rechecked the current head `553fd6f0666a3129e5134a58e375ad67d1a8a8b8` after the “addressed” reply. The head has not changed since the original review, and this path still reuses the same `ServiceMetricEvent.Builder`: it sets `DruidMetrics.REASON` for thrown-away-by-reason events, then emits `ingest/events/filtered` without clearing that dimension. I’m keeping this as the only active finding in the corrected summary; please push the corresponding code change or point me to a different current head for reassessment.

FrankChen021 (commented):
## 🟡 Changes recommended

Correction to the earlier review: the two constructor-compatibility findings are withdrawn. Although the old JVM descriptors are absent, these constructors belong to unannotated classes and are outside Druid’s `@PublicApi`/`@ExtensionPoint` compatibility contract. The current head is unchanged since the original review, and the metrics builder still carries `DruidMetrics.REASON` into `ingest/events/filtered`, so that remains the only active finding.

Reviewed 51 of 51 changed files; findings in 1 file.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

After addressing the finding or replying to the comment, you can request another review from me to trigger a new automated review.

FrankChen021 (commented):
## 🟢 Approval recommended

Static review of the updated head began with the incremental diff since the previous review and then covered the complete merge-base diff and surrounding code. The active prior finding—the `DruidMetrics.REASON` dimension leaking from per-reason `thrownAway` metrics into subsequent metrics—is fixed by clearing the shared builder after those emissions. Earlier constructor-compatibility findings were withdrawn and are not repeated. I found no remaining high-confidence correctness, data-loss, compatibility, security, lifecycle/concurrency, or missing-test issue that warrants a review finding.

Coverage: reviewed 53 of 53 current changed files.

Validation: `git diff --check` was run; it reported only two trailing-whitespace lines in newly added Javadoc. No build or tests were run; this was a static review as requested.



---

PR #18528, opened 2025-09-15
Bump derby to 10.14.3

cryptoe:
Bump derby to 10.14.3


cryptoe:
` 10.17.1.0` requires us to pin the version to `java 21` hence we cannot do that now. 
Closing this PR.

PR #18531, opened 2025-09-15
Add time window filter to Iceberg reader

a2l007:
### Description

Introduces a time window filter to the Druid Iceberg reader which can be used to define lookback and lookahead durations while filtering on a partition time column.
With the default configuration, this filter will only read data files from the last 24 hours. Combining this filter with the scheduled batch supervisor, an ingestion pipeline can be setup in Druid that reads the latest iceberg data every day. 

#### Release note
Adds the ability to filter iceberg data based on a configurable time window.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/docs/ingestion/input-sources.md
+++ b/docs/ingestion/input-sources.md
@@ -1119,7 +1119,7 @@ Sample:
 
 ### Iceberg filter object
 
-This input source provides the following filters: `and`, `equals`, `interval`, and `or`. You can use these filters to filter out data files from a snapshot, reducing the number of files Druid has to ingest.
+This input source provides the following filters: `and`, `equals`, `interval`, `timeWindow`, `range` and `or`. You can use these filters to filter out data files from a snapshot, reducing the number of files Druid has to ingest.
 It is strongly recommended to apply filtering only on Iceberg partition columns. When filtering on non-partition columns, Iceberg filters may return rows that do not fully match the expression. To address this, it may help to define an additional filter in the [`transformSpec`](./ingestion-spec.md#transformspec) to remove residual rows.
 
 `equals` Filter:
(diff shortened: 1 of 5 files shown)


maytasm:
How is this different than setting `intervals` in `granularitySpec`?

a2l007:
> How is this different than setting `intervals` in `granularitySpec`?

@maytasm Thanks for taking a look! This filter is at the input source level which means it controls the data file paths pulled from the iceberg table for ingestion.  Pruning the data files during the iceberg table scan phase helps reduce scan times and also reduces the number of files processed by the subsequent indexing phase.

maytasm:
> > How is this different than setting `intervals` in `granularitySpec`?
> 
> @maytasm Thanks for taking a look! This filter is at the input source level which means it controls the data file paths pulled from the iceberg table for ingestion. Pruning the data files during the iceberg table scan phase helps reduce scan times and also reduces the number of files processed by the subsequent indexing phase.

I see. Not directly related to this PR (time window filter) but Iceberg filter in general, what if the filterColumn is not a partition column? This would result in files that may contain rows not matching the filter right? We would then need to filter it on Druid side again using filter in transformSpec?

a2l007:
>  This would result in files that may contain rows not matching the filter right? We would then need to filter it on Druid side again using filter in transformSpec?

That is correct, filtering on non partition columns may pull residual rows which will have to be filtered on the Druid side.

maytasm:
> > This would result in files that may contain rows not matching the filter right? We would then need to filter it on Druid side again using filter in transformSpec?
> 
> That is correct, filtering on non partition columns may pull residual rows which will have to be filtered on the Druid side.

Do we rely on users to manually add those Iceberg filter as Druid filter in transformSpec? Out of scope of this PR, but I'm wondering if it's either possible to automatically convert Iceberg filter and add them to Druid filter in transformSpec for the users or warn/fail the job if the Iceberg filters result in residual files and there is no corresponding filters in transformSpec.

maytasm (approved):
LGTM. One quick comment....Do we need to make sure that the filterColumn in IcebergTimeWindowFilter is of TimestampType? Or at least mention in the docs that the column must be of TimestampType?

a2l007:
> Do we rely on users to manually add those Iceberg filter as Druid filter in transformSpec? Out of scope of this PR, but I'm wondering if it's either possible to automatically convert Iceberg filter and add them to Druid filter in transformSpec for the users or warn/fail the job if the Iceberg filters result in residual files and there is no corresponding filters in transformSpec.

That's a good idea, I think if we're able to push down the transformSpec to the inputSource, it should be possible.

PR #18535, opened 2025-09-16
allow queries with filtered aggs to match unfiltered projection aggs as long as the filter required columns are present

clintropolis:
### Description
Queries with filtered aggregators can use projections which have unfiltered versions of those same aggs as long as the projection also has the columns present to be able to satisfy evaluating the filter.


processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -296,6 +292,10 @@ public static ProjectionMatchBuilder matchAggregators(
     }
     boolean allMatch = true;
     for (AggregatorFactory queryAgg : queryCursorBuildSpec.getAggregators()) {
+      AggregatorFactory filterAgg = null;
```
cecemei, line 295: nit: maybe can move `FilteredAggregatorFactory filteredQueryAgg` up here and check null for that instead

clintropolis: it needs to go after the check for exact aggregator match because we would prefer an exact match in case the projection itself defined a filtered aggregator, so  this is a fallback to see if a query filtered aggregator can use a non-filtered agg

PR #18537, opened 2025-09-16
refactor vectorized nvl implementation into more general coalesce implementation

clintropolis:
### Description
This PR turns vectorized nvl into vectorized coalesce to support any number of arguments, filling in one of the remaining gaps of vectorized conditional expressions.


clintropolis:
After this PR, the last big thing to solve for conditionals that I'm aware of is how to handle mixed output types. The vector processing currently requires consistent typing, but the non-vectorized processing does not meaning that the selectors lazily cast to the claimed output type instead of eagerly make the evaluated result match the output type the expression claimed to produce.

I think it would be best to switch non-vectorized processing to use forcefully cast the evaluated result to match the type specified by `Expr.getOutputType`, but this could introduce a subtle behavior change for nested conditional expressions... so it might need to be behind a feature flag.

PR #18539, opened 2025-09-16
Remove ordering direction requirement on projection match

cecemei:
### Description
Remove ordering direction requirement on projection match. Also updated `CursorFactoryProjectionTest`:
- use `groupingEngine.mergeRunners` to enable result sorting, since segment sorting push down is not enabled in `groupingEngine.process`.
- update all ArrayResultSet to use List<Object[]> instead.
- prefer `Map.of` over `ImmutableMap.of`.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/CursorBuildSpec.java
+++ b/processing/src/main/java/org/apache/druid/segment/CursorBuildSpec.java
@@ -235,7 +235,7 @@ public class CursorBuildSpec
       return false;
     }
     for (int i = 0; i < preferredOrdering.size(); i++) {
-      if (!ordering.get(i).equals(preferredOrdering.get(i))) {
+      if (!ordering.get(i).getColumnName().equals(preferredOrdering.get(i).getColumnName())) {
         return false;
       }
     }
--- a/processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
+++ b/processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
@@ -20,9 +20,6 @@
 package org.apache.druid.segment;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import it.unimi.dsi.fastutil.Hash;
-import it.unimi.dsi.fastutil.objects.ObjectOpenCustomHashSet;
 import org.apache.druid.collections.CloseableDefaultBlockingPool;
 import org.apache.druid.collections.CloseableStupidPool;
 import org.apache.druid.collections.NonBlockingPool;
@@ -46,10 +43,14 @@ import org.apache.druid.java.util.common.guava.Sequence;
 import org.apache.druid.java.util.common.io.Closer;
 import org.apache.druid.math.expr.ExprMacroTable;
 import org.apache.druid.query.DefaultQueryMetrics;
+import org.apache.druid.query.DirectQueryProcessingPool;
 import org.apache.druid.query.DruidProcessingConfig;
 import org.apache.druid.query.Druids;
 import org.apache.druid.query.Query;
 import org.apache.druid.query.QueryContexts;
+import org.apache.druid.query.QueryPlus;
+import org.apache.druid.query.QueryResourceId;
+import org.apache.druid.query.QueryRunner;
 import org.apache.druid.query.Result;
 import org.apache.druid.query.aggregation.AggregatorFactory;
 import org.apache.druid.query.aggregation.CountAggregatorFactory;
@@ -108,7 +109,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.stream.Collectors;
 
 @RunWith(Parameterized.class)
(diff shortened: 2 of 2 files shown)


processing/src/main/java/org/apache/druid/segment/CursorBuildSpec.java
```diff
@@ -235,7 +235,7 @@ public boolean isCompatibleOrdering(List<OrderBy> ordering)
       return false;
     }
     for (int i = 0; i < preferredOrdering.size(); i++) {
-      if (!ordering.get(i).equals(preferredOrdering.get(i))) {
+      if (!ordering.get(i).getColumnName().equals(preferredOrdering.get(i).getColumnName())) {
```
clintropolis: I don't think we want to change this method since i think some callers do need to check that the directions are the same.

More importantly though, this needs to be more strict than just matching name, the check that projections needs is that the ordering is either same direction, or every column is exactly the opposite direction of the preferred ordering. So like `ORDER BY X ASC, Y DESC` can match `X ASC, Y DESC` or `X DESC, Y ASC`, but cannot match `X ASC Y ASC`.

This functionality is not really used much yet since today mainly the only thing callers care about is ordered by __time first, but we want stuff to be correct in the future as we expand

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -605,6 +590,10 @@ public void testProjectionSelectionTwoDims()
                     .setInterval(Intervals.ETERNITY)
                     .addDimension("a")
                     .addDimension("b")
+                    .addOrderByColumn("a", Direction.DESCENDING)
+                    .addOrderByColumn("b", Direction.DESCENDING)
+                    .setLimit(10)
+                    .setContext(Map.of(QueryContexts.QUERY_RESOURCE_ID, RESOURCE_ID))
```
clintropolis: if this is required for something, it should maybe be set by `testGroupBy` or something instead of needing each test to set it specifically

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
```diff
@@ -1898,75 +1909,40 @@ private void testGroupBy(
...
+    runner = (unused1, unused2) ->
+        groupingEngine.process(
+            query.withOverriddenContext(Map.of(QueryContexts.NO_PROJECTIONS, true)),
+            cursorFactory,
+            timeBoundaryInspector,
+            nonBlockingPool,
+            null
+        );
+    final List<ResultRow> resultsNoProjection = groupingEngine
+        .mergeRunners(DirectQueryProcessingPool.INSTANCE, List.of(runner))
```
clintropolis: it doesn't seem that important that these tests use `mergeRunners` since they are focused on testing cursors of the per segment processing, but i guess it isn't that harmful either...

cecemei: It needs mergeRunners because `GroupByMergingQueryRunner` used `RowBasedGrouperHelper.makeGrouperIterator`, which returns sorted result. `GroupByQueryEngine.GroupByEngineIterator` returns unsorted result.

clintropolis: why do we need a sorted result? (other than making it a bit easier to write tests since otherwise sometimes it is inconsistent between runs)

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
cecemei: this is moved to testGroupBy based on merging is needed or not

processing/src/main/java/org/apache/druid/segment/CursorBuildSpec.java
cecemei: Synced offline, this has been updated

processing/src/test/java/org/apache/druid/segment/CursorFactoryProjectionTest.java
cecemei: updated to make merging/sorting controllable.

processing/src/main/java/org/apache/druid/segment/CursorBuildSpec.java
```diff
@@ -234,8 +237,18 @@ public boolean isCompatibleOrdering(List<OrderBy> ordering)
     if (ordering.size() < preferredOrdering.size()) {
       return false;
     }
-    for (int i = 0; i < preferredOrdering.size(); i++) {
-      if (!ordering.get(i).equals(preferredOrdering.get(i))) {
+
+    boolean exactMatch = ordering.get(0).equals(preferredOrdering.get(0));
+    if (!exactMatch && !ordering.get(0).equals(preferredOrdering.get(0).reverse())) {
```
clintropolis: it might be nicer to do this without making additional `OrderBy` objects that `reverse` is going to be making since this is going to happen for every projection match of every segment being processed (to keep the garbage collector from having to do work); i think we could just check inline if the direction is same or reversed, and just keep counters and make sure either the same direction counter or reverse counter are the same as the total length (and bail early if neither count matches the processed count)

cecemei: updated to a `isExactReverse` method instead of creating additional objects

PR #18540, opened 2025-09-16
Make the HTTP client that lives beneath fabric8 KubernetesClient configurable

capistrant:
### Description

Allow an operator to select which HTTP client they want to use for their instance of fabric8 `KubernetesClient`

#### Http Client Selector

`druid.indexer.runner.k8sAndWorker.http.httpClientType`

Valid values: `vertx` (default), `okhttp`, `jdk`

> why no jetty?
* As of now no jetty http client support due to jetty dependency version clashing. It could be possible to look into this in the future though.

##### vertx

Extra configuration stays the same as it was in terms of key names and defaults, but the prefix changes to `druid.indexer.runner.k8sAndWorker.http.vertx`

##### okhttp

This is the old client that we used before #17913. Drawbacks are the default unbounded thread pool underpinning the client. This was a big reason behind making the change to vert.x in the first place (in addition to vert.x being the new default for fabric8 KubernetesClient in newer versions).

Extra configuration:

* `druid.indexer.runner.k8sAndWorker.http.okhttp.useCustomDispatcherExecutor`
    * Whether or not you want to override the default unbounded dispatcher.
    * default = `false` 
* `druid.indexer.runner.k8sAndWorker.http.okhttp.maxWorkerThreads`
    * The upper bound on the thread pool if you use the custom dispatcher
    * default `100`
* `druid.indexer.runner.k8sAndWorker.http.okhttp.coreWorkerThreads`
    * The lower bound on the thread pool size if you use the custom dispatcher 
    * default = `20`
* `druid.indexer.runner.k8sAndWorker.http.okhttp.workerThreadKeepAliveTime`
    * default = `60` (seconds)
    * how long idle threads will live for the thread pool if you use the custom dispatcher

##### jdk

This uses the native http client in Java. Not recommended if runtime is java11.

Extra configuration:

* `druid.indexer.runner.k8sAndWorker.http.jdk.maxWorkerThreads`
    * The upper bound on the thread pool
    * default `100`
* `druid.indexer.runner.k8sAndWorker.http.jdk.coreWorkerThreads`
    * The lower bound on the thread pool size
    * default = `20`
* `druid.indexer.runner.k8sAndWorker.http.jdk.workerThreadKeepAliveTime`
    * default = `60` (seconds)
    * how long idle threads will live for the thread pool.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
+++ b/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
@@ -54,7 +54,13 @@ import org.apache.druid.java.util.common.StringUtils;
 import org.apache.druid.java.util.common.lifecycle.Lifecycle;
 import org.apache.druid.java.util.common.logger.Logger;
 import org.apache.druid.k8s.overlord.common.DruidKubernetesClient;
-import org.apache.druid.k8s.overlord.common.DruidKubernetesHttpClientConfig;
+import org.apache.druid.k8s.overlord.common.httpclient.DruidKubernetesHttpClientFactory;
+import org.apache.druid.k8s.overlord.common.httpclient.jdk.DruidKubernetesJdkHttpClientConfig;
+import org.apache.druid.k8s.overlord.common.httpclient.jdk.DruidKubernetesJdkHttpClientFactory;
+import org.apache.druid.k8s.overlord.common.httpclient.okhttp.DruidKubernetesOkHttpHttpClientConfig;
+import org.apache.druid.k8s.overlord.common.httpclient.okhttp.DruidKubernetesOkHttpHttpClientFactory;
+import org.apache.druid.k8s.overlord.common.httpclient.vertx.DruidKubernetesVertxHttpClientConfig;
+import org.apache.druid.k8s.overlord.common.httpclient.vertx.DruidKubernetesVertxHttpClientFactory;
 import org.apache.druid.k8s.overlord.execution.KubernetesTaskExecutionConfigResource;
 import org.apache.druid.k8s.overlord.execution.KubernetesTaskRunnerDynamicConfig;
 import org.apache.druid.k8s.overlord.runnerstrategy.RunnerStrategy;
@@ -79,7 +85,10 @@ public class KubernetesOverlordModule implements DruidModule
                                                                + ".k8sAndWorker";
   private static final String RUNNERSTRATEGY_PROPERTIES_FORMAT_STRING = K8SANDWORKER_PROPERTIES_PREFIX
                                                                         + ".runnerStrategy.%s";
-  private static final String HTTPCLIENT_PROPERITES_PREFIX = K8SANDWORKER_PROPERTIES_PREFIX + ".http";
+  private static final String HTTPCLIENT_TYPE_PROPERTY = K8SANDWORKER_PROPERTIES_PREFIX + ".http.httpClientType";
+  private static final String VERTX_HTTPCLIENT_PROPERITES_PREFIX = K8SANDWORKER_PROPERTIES_PREFIX + ".http.vertx";
+  private static final String OKHTTP_HTTPCLIENT_PROPERITES_PREFIX = K8SANDWORKER_PROPERTIES_PREFIX + ".http.okhttp";
+  public static final String JDK_HTTPCLIENT_PROPERITES_PREFIX = K8SANDWORKER_PROPERTIES_PREFIX + ".http.jdk";
 
   @Override
   public void configure(Binder binder)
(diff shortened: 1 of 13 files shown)


extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
```diff
@@ -79,7 +85,10 @@ public class KubernetesOverlordModule implements DruidModule
                                                                + ".k8sAndWorker";
   private static final String RUNNERSTRATEGY_PROPERTIES_FORMAT_STRING = K8SANDWORKER_PROPERTIES_PREFIX
                                                                         + ".runnerStrategy.%s";
-  private static final String HTTPCLIENT_PROPERITES_PREFIX = K8SANDWORKER_PROPERTIES_PREFIX + ".http";
+  private static final String HTTPCLIENT_TYPE_PROPERTY = K8SANDWORKER_PROPERTIES_PREFIX + ".http.httpClientType";
```
gianm: You haven't added documentation for this parameter, which I think is good because I think it's a parameter that we may remove "soon". I suggest noting that in a comment here. Its purpose is to help cluster operators evaluate different http clients with fabric8, and we're adding it because we've noticed issues with two of the major ones:

- the old default (okhttp) can lead to excessive numbers of threads if you have a lot of tasks running
- the new default (vert.x) has been observed to generate spurious API call failures, which leads to seemingly-random task failures

Ideally, through some evaluation we will determine which one is best to use and how best to configure it, and at that point we could remove the property. It will help us slim down the distribution since it won't need to ship all 3 fabric8 client modules.

FrankChen021:
why is the jdk not recommend for java 11?  

Another point is that the type name `jdk` is ambiguous, people may think it's the http client shipped in the `sun.net.www.http` package.

capistrant:
> why is the jdk not recommend for java 11?
> 
> Another point is that the type name `jdk` is ambiguous, people may think it's the http client shipped in the `sun.net.www.http` package.

https://github.com/fabric8io/kubernetes-client/tree/main/httpclient-jdk just reading from their README they mentioned java16+ having fixes that some use cases need. Did not evaluate deeply since Druid is pending drop of java11 support due to jetty upgrade anyways.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
capistrant: I may or may not have been working on a commit to add docs after I realized I forgot 😆 but we can leave un-documented I guess. But maybe putting it in the extensions doc under an **experimental** label plus disclaimer that this is temporary while the community determines best path forward for a generic config that works well for all is good compromise. Then all users of the extension could help find the best path forward if they so choose.

FrankChen021:
> > why is the jdk not recommend for java 11?
> > Another point is that the type name `jdk` is ambiguous, people may think it's the http client shipped in the `sun.net.www.http` package.
> 
> https://github.com/fabric8io/kubernetes-client/tree/main/httpclient-jdk just reading from their README they mentioned java16+ having fixes that some use cases need. Did not evaluate deeply since Druid is pending drop of java11 support due to jetty upgrade anyways.

Thanks for the info.
From the link, I notice that the problem is related to web sockets, not the normal http request. So I think it's not our case here.
> It will not work with websocket requests containing queries with encoded characters until the fix  https://github.com/openjdk/jdk/commit/c07ce7eec71aefbd3cb624e03ca53f5148d01f19 which is available in Java 16 - https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8245245

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesOverlordModule.java
gianm: That'd also be OK, I just don't want to create a reason to worry about removing it in the future. If the doc page says that it may be removed in a future release then that works.

docs/development/extensions-core/k8s-jobs.md
```diff
@@ -855,3 +901,39 @@ To do this, set the following property.
...
+
+:::warning
+
+This configuration is experimental and subject to change. The Druid developer community intends on selecting a stable default HTTP client and configuration in the future, simplifying configuration and distribution packaging. This means that not all exposed HTTP clients and their configurations will be supported long term. If you opt in to using this experimental configuration, please provide feedback to the Druid developer community (Apache mailing list: `dev@druid.apache.org`) on your experience to help guide our long term decisions.
+
+:::
+
+|Property| Possible Values |Description| Default |required|
+|--------|-----------------|-----------|---------|--------|
+|`druid.indexer.runner.k8sAndWorker.http.httpClientType`|`String` (e.g., `okhttp`, `vertx`, `javaStandardHttp`)|Specifies the HTTP client library to be used by the worker task runner for communication with worker nodes.|`vertx`|No|
```
cryptoe, line 873: Should the default be okhttp. 
cc @gianm ?

gianm: It should be whatever we believe works best for the average user.

capistrant: I do not think we are ready to say vert.x has enough problems to remove it from being default. If we still want this option to pick a client in druid35 so the community can work to identify the best long term approach, I'm on board with that. But as of now, I think leaving the default as is is best for druid 35.

docs/development/extensions-core/k8s-jobs.md
```diff
@@ -316,6 +316,52 @@ Host: http://ROUTER_IP:ROUTER_PORT
...
 
+### Experimental HTTP Client Config
+
+:::warning
+
+This configuration is experimental and subject to change. The Druid developer community intends on selecting a stable default HTTP client and configuration in the future, simplifying configuration and distribution packaging. This means that not all exposed HTTP clients and their configurations will be supported long term. If you opt in to using this experimental configuration, please provide feedback to the Druid developer community (Apache mailing list: `dev@druid.apache.org`) on your experience to help guide our long term decisions.
+
+:::
+
+The extension uses [fabric8 KubernetesClient](https://github.com/fabric8io/kubernetes-client) to communicate with the Kubernetes API server. This client creates an
```
gianm: Since this is area where there are problems and our understanding of them is evolving, I think it would be better to move most of this info to a GitHub issue and link it here. Name the GitHub issue something like "Determine best Kubernetes API client"  and put the known issues and their mitigations there. That way, as our understanding evolves, old versions' documentation doesn't grow too stale. The only info we need here is a brief preamble, pointer to the GitHub issue, and the table with the descriptions of the configuration options.

capistrant: makes sense, have created https://github.com/apache/druid/issues/18629 for the tracking and discussion of the issue and removed most of the docs content outside of the configs

gianm:
TY @capistrant!

PR #18541, opened 2025-09-17
Add size check in DictionaryBuildingSingleValuedRowBasedKeySerdeHelper in putToKeyBuffer

cecemei:
### Description

Add size check in `DictionaryBuildingSingleValuedRowBasedKeySerdeHelper` in `putToKeyBuffer` . Without this change, dictionary size is unbounded and can cause historical node OOM. Additionally:
- add `NestedDataTypeStrategy` for json column
 
<hr>

This PR has:

- [x] been self-reviewed.


processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/RowBasedGrouperHelper.java
```diff
@@ -1619,6 +1627,11 @@ public boolean putToKeyBuffer(RowBasedKey key, int idx)
         final Object obj = key.getKey()[idx];
         int id = getReverseDictionary().getInt(obj);
         if (id == DimensionDictionary.ABSENT_VALUE_ID) {
+          int size = estimatedKeySize(key.getKey());
+          if (currentEstimatedSize + size > maxDictionarySize) {
+            return false;
+          }
+          currentEstimatedSize += size;
```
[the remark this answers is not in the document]
clintropolis: this comment is correct, it should be `estimateKeySize(obj)`, not the whole grouping key array

processing/src/main/java/org/apache/druid/segment/nested/StructuredData.java
```diff
@@ -133,16 +135,30 @@ private Number asNumber()
...
+    return sizeEstimate;
+  }
+
   @Override
   public int compareTo(StructuredData o)
   {
-    if (this.equals(o)) {
+    if (this == o) {
       return 0;
+    } else if (o == null) {
```
clintropolis: does this happen? Is fine since is harmless to be here, but i think most users of this would be dealing in `StructuredData` where a null value is still wrapped (so probably shouldn't appear here much in practice)

processing/src/main/java/org/apache/druid/segment/nested/NestedDataComplexTypeSerde.java
```diff
@@ -124,7 +123,7 @@ public Object fromByteBuffer(ByteBuffer buffer, int numBytes)
       @Override
       public byte[] toBytes(@Nullable Object val)
       {
-        return serializeToBytes(val);
+        return serializeToBytes(StructuredData.wrap(val));
```
clintropolis: what is this for? the `StructuredData` [serializer is just going to unwrap this when serializing]( https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/nested/StructuredDataJsonSerializer.java#L29)

processing/src/main/java/org/apache/druid/segment/column/TypeStrategies.java
```diff
@@ -697,4 +701,77 @@ public Class<?> getClazz()
       return Object[].class;
     }
   }
+
+  public static final class JsonTypeStrategy implements TypeStrategy
```
clintropolis: this should probably be named `NestedDataTypeStrategy` to be consistent with other internal naming

processing/src/main/java/org/apache/druid/segment/column/TypeStrategies.java
```diff
@@ -697,4 +701,77 @@ public Class<?> getClazz()
...
+  {
+    private final ObjectStrategy objectStrategy;
```
clintropolis: nit: it feels a bit off to me to have to make an `ObjectStrategy` to implement the `TypeStrategy`, where ideally I think I would expect both the nested column `ObjectStrategy` and `TypeStrategy` to be implemented such that they can just share some common methods instead of one relying on the other. `ObjectStrategyComplexTypeStrategy` was a sort of backwards compatible fallback to make it so that we could fill in `TypeStrategy` from pre-existing `ObjectStrategy`.

Not a big deal though

processing/src/main/java/org/apache/druid/segment/nested/StructuredData.java
cecemei: it's true, this would be caught by all the null checks and null comparators so it'd never happen, removed it

processing/src/main/java/org/apache/druid/segment/nested/NestedDataComplexTypeSerde.java
cecemei: true, the wrap is implemented in deserializer, removed here

processing/src/main/java/org/apache/druid/segment/column/TypeStrategies.java
cecemei: updated

processing/src/main/java/org/apache/druid/segment/column/TypeStrategies.java
cecemei: moved all functions to TypeStrategy, looks better for encapsulation!

processing/src/main/java/org/apache/druid/segment/column/TypeStrategies.java
```diff
@@ -697,4 +704,103 @@ public Class<?> getClazz()
       return Object[].class;
     }
   }
+
+  public static final class NestedDataTypeStrategy implements TypeStrategy<Object>
+  {
+    public static final NestedDataTypeStrategy INSTANCE = new NestedDataTypeStrategy();
```
clintropolis: nit, could put this at the top of `TypeStrategies` alongside the other static instances then would be `TypeStrategies.NESTED_DATA` or whatever

processing/src/main/java/org/apache/druid/segment/nested/StructuredData.java
```diff
@@ -133,16 +135,28 @@ private Number asNumber()
...
+    }
+    Preconditions.checkState(sizeEstimate >= 0, "sizeEstimate not initialized");
+    return sizeEstimate;
+  }
+
   @Override
   public int compareTo(StructuredData o)
   {
-    if (this.equals(o)) {
+    if (this == o) {
```
clintropolis: i feel like we should leave this, since `StructuredData` is also used as a wrapper at ingest time for all 'auto' types, primitive types can short-circuit out of this comparison before forcing stuff to be serialized, which seems like it would add a bit of overhead.

I guess you're trying to make all of equals/hashcode/compareTo like consistently weird the same way with the occasional hash collisions, instead of just isolating the weirdness where it falls to the hash in just the compareTo?

Fwiw, I was trying to keep the strangeness of using the hash instead of performing full byte comparison isolated in compareTo (since ideally we would be performing full comparison), is there a reason related to this PR of why these need to change?

cecemei: I thought this equal/compareTo is strange since it's not consistent. Equal doesn't handle nested array equal, if there's `Object[]` in a nested field, they don't equal, even though the compareTo returns equal since they're still the same data. But I've only seen places using compareTo so it might be fine. 

Anyway I reverted the compareTo/equals part of change, removed the original hashCode method since it doesn't seem to make sense to keep two hash method. My fix is also not a perfect solution, since: 1. it sort of require object to be serialized even for the equal check now, 2. it still have the issue of hash collision.

processing/src/main/java/org/apache/druid/segment/column/TypeStrategies.java
cecemei: updated

processing/src/test/java/org/apache/druid/segment/nested/StructuredDataTest.java
```diff
@@ -106,10 +105,4 @@ public void testCompareToWithDifferentJSONOrder()
     Assert.assertEquals(1, sd1.compareTo(sd2));
     Assert.assertEquals(0, sd0.compareTo(sd1));
   }
-
-  @Test
-  public void testEqualsAndHashcode()
-  {
-    EqualsVerifier.forClass(StructuredData.class).withIgnoredFields("hashInitialized", "hashValue", "hash").usingGetClass().verify();
-  }
```
clintropolis, line 114: why did we need to remove this?

clintropolis (approved):

processing/src/main/java/org/apache/druid/segment/nested/StructuredData.java
clintropolis: ah yea, it totally is strange, and i wasn't necessarily encouraging reverting it, just was curious if was directly related or not. I think it would be ok to consider making these all consistent with each other, but we can do that in a separate PR since it also doesn't really matter much either way since compareTo is the main thing that is used by stuff.

processing/src/test/java/org/apache/druid/segment/nested/StructuredDataTest.java
cecemei: something with `EqualsVerifier`, it doesn't allow the `hashCode` being delegated to an abstract method, i'm not quite sure, but looks like it doesnt like the hash longSupplier

PR #18505, opened 2025-09-09
Add user agent to headers

GabrielCWT:
Add user agent to headers for services.

This allows tracing to be easier as we can more easily tell which service is sending the request.

PR #18542, opened 2025-09-18
Add version to sys server

GabrielCWT:
This PR adds the Druid version information to the web console, so that each service displays the version it is running.

This is especially useful during rolling upgrades, since it allows operators to easily verify which services have been updated and to track version mismatches across the cluster.

Changes include:

- Adding a new `version` column to sys.server
- Displaying this new column in the "Services" tab.

Screenshot 2025-09-18 at 5 02 29 PM

### DIFF
--- a/docs/querying/sql-metadata-tables.md
+++ b/docs/querying/sql-metadata-tables.md
@@ -236,7 +236,7 @@ Servers table lists all discovered servers in the cluster.
 |max_size|BIGINT|Max size in bytes this server recommends to assign to segments see [druid.server.maxSize](../configuration/index.md#historical-general-configuration). Only valid for HISTORICAL type, for other types it's 0|
 |is_leader|BIGINT|1 if the server is currently the 'leader' (for services which have the concept of leadership), otherwise 0 if the server is not the leader, or null if the server type does not have the concept of leadership|
 |start_time|STRING|Timestamp in ISO8601 format when the server was announced in the cluster|
-|version|VARCHAR|Version of the server|
+|version|VARCHAR|Druid version running on the server|
 To retrieve information about all servers, use the query:
 
 ```sql
--- a/docs/querying/sql-metadata-tables.md
+++ b/docs/querying/sql-metadata-tables.md
@@ -236,6 +236,7 @@ Servers table lists all discovered servers in the cluster.
 |max_size|BIGINT|Max size in bytes this server recommends to assign to segments see [druid.server.maxSize](../configuration/index.md#historical-general-configuration). Only valid for HISTORICAL type, for other types it's 0|
 |is_leader|BIGINT|1 if the server is currently the 'leader' (for services which have the concept of leadership), otherwise 0 if the server is not the leader, or null if the server type does not have the concept of leadership|
 |start_time|STRING|Timestamp in ISO8601 format when the server was announced in the cluster|
+|version|VARCHAR|Version of the server|
 To retrieve information about all servers, use the query:
 
 ```sql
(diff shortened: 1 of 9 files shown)


vogievetsky:
Nice, this is super useful.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -697,7 +700,7 @@ private static Object[] buildRowForDiscoverableDataServer(
           currentSize,
           druidServerToUse.getMaxSize(),
           null,
-          toStringOrNull(discoveryDruidNode.getStartTime())
+          node.getVersion()
```
FrankChen021, line 704: the start time is deleted?

GabrielCWT: Fixed

docs/querying/sql-metadata-tables.md
```diff
@@ -236,6 +236,7 @@ Servers table lists all discovered servers in the cluster.
 |max_size|BIGINT|Max size in bytes this server recommends to assign to segments see [druid.server.maxSize](../configuration/index.md#historical-general-configuration). Only valid for HISTORICAL type, for other types it's 0|
 |is_leader|BIGINT|1 if the server is currently the 'leader' (for services which have the concept of leadership), otherwise 0 if the server is not the leader, or null if the server type does not have the concept of leadership|
 |start_time|STRING|Timestamp in ISO8601 format when the server was announced in the cluster|
+|version|VARCHAR|Version of the server|
```
```suggestion
|version|VARCHAR|Druid version running on the server|
```

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -381,6 +381,8 @@ public void setUp(@TempDir File tmpDir) throws Exception
 
   private final DateTime startTime = DateTimes.nowUtc();
 
+  private final String version = new DruidNode("s1", "localhost", false, 8081, null, true, false).getVersion();
```
kfaraz: I don't think `DruidNode` exposes any `getVersion` method.
Also, if we just need the version, why instantiate a `DruidNode`?

kfaraz (commented):
Thanks for the changes. Left some suggestions.

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
GabrielCWT: Updated to the code below. Just for your info, I have exposed `getVersion` for `DruidNode` in a previous PR https://github.com/apache/druid/pull/18505

```
private final String version = GuavaUtils.firstNonNull(
    SystemSchemaTest.class.getPackage().getImplementationVersion(),
    "unknown"
);
```

kfaraz: Thanks for the clarification! I must have been looking at older code.

Actually, the better thing to do would be to make the `unknown` a constant in `DruidNode` and use it here.

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -381,6 +382,10 @@ public void setUp(@TempDir File tmpDir) throws Exception
 
   private final DateTime startTime = DateTimes.nowUtc();
 
+  private final String version = GuavaUtils.firstNonNull(
+          SystemSchemaTest.class.getPackage().getImplementationVersion(),
+          DruidNode.UNKNOWN_VERSION);
```
kfaraz: Nit: style
```suggestion
          DruidNode.UNKNOWN_VERSION
      );
```

kfaraz:
There seem to be some flaky tests that were enabled in https://github.com/apache/druid/pull/18561
I tried to re-run the flow but they keep failing repeatedly.
Trying another re-run.

cc: @uds5501

FrankChen021 (approved):

PR #18543, opened 2025-09-18
Bouncer: Acquire child tickets before parent tickets.

gianm:
Follow up to #18520. The logic introduced in that patch for parent/child relationships was not ideal, since it would lead to parent tickets being acquired even if the child was at its max count. In production, this would cause a query with maxThreads = 1 to acquire more than 1 thread ticket from the global Bouncer.

### DIFF
--- a/examples/quickstart/tutorial/hadoop/docker/Dockerfile
+++ b/examples/quickstart/tutorial/hadoop/docker/Dockerfile
@@ -61,7 +61,8 @@ ENV JAVA_HOME=/usr/lib/jvm/zulu11
 ENV PATH=$PATH:$JAVA_HOME/bin
 
 # hadoop
-ARG APACHE_ARCHIVE_MIRROR_HOST=https://downloads.apache.org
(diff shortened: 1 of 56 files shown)


processing/src/main/java/org/apache/druid/frame/processor/Bouncer.java
```diff
@@ -89,24 +89,16 @@ public int getMaxCount()
...
-  }
-
-  /**
-   * Acquire a ticket from this Bouncer. Precondition: if there is a parentBouncer, only call this method when
-   * holding a parent ticket.
-   */
-  private ListenableFuture<Ticket> ticketInternal(@Nullable final Ticket parentTicket)
-  {
+    // Acquire our ticket first, then acquire a parent ticket. Only return our ticket once the parent ticket
+    // is also acquired.
```
abhishekrb19, line 93: IMO this could be called out in the method’s javadoc or at the class-level javadoc. I understood the behavior better after looking at the test code.

abhishekrb19 (approved):

PR #18544, opened 2025-09-18
druid.storage.type=s3 now honors druid.storage.zip to allow storing segments in s3 without zip compression

clintropolis:
### Description
This PR adds the capability to store segments in s3 without compressing with zip, similar to the 'local' deep storage option. This is mainly for experimentation at this point, but went ahead and documented it just in case anyone else wants to experiment :shrug:

When `druid.storage.zip` is false, the load spec stores the prefix to use instead of the exact location, and the pullers/killers/movers when for a path that ends with `/` and then do a list operation and iterate over and apply the operation to the results.

### DIFF
--- a/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentPusher.java
+++ b/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentPusher.java
@@ -24,6 +24,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import org.apache.druid.error.DruidException;
+import org.apache.druid.java.util.common.IOE;
 import org.apache.druid.java.util.common.StringUtils;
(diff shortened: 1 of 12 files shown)


extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentPusher.java
```diff
@@ -130,6 +141,56 @@ public DataSegment pushToPath(File indexFilesDir, DataSegment inSegment, String
...
+        try {
+          S3Utils.retryS3Operation(
+              () -> {
+                S3Utils.uploadFileIfPossible(s3Client, config.getDisableAcl(), config.getBucket(), s3Path + file.getName(), file);
+                return null;
+              }
+          );
+        }
+        catch (AmazonServiceException e) {
+          if (S3Utils.ERROR_ENTITY_TOO_LARGE.equals(S3Utils.getS3ErrorCode(e))) {
```
gianm: This whole block is duplicated, would be good to dedupe it.

gianm (approved):

PR #18546, opened 2025-09-18
Update to ZooKeeper 3.8.5.

gianm:
It was released today. ZooKeeper 3.8.4 has already been removed from the download mirrors, so our integration tests are failing. Might as well upgrade.

### DIFF
--- a/licenses.yaml
+++ b/licenses.yaml
@@ -2071,7 +2071,7 @@ name: Apache Zookeeper
 license_category: binary
 module: java-core
 license_name: Apache License version 2.0
-version: 3.8.4
+version: 3.8.5
 libraries:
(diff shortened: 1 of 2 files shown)


clintropolis (approved):

gianm:
Hmm, now some tests are failing because `zookeeper:3.8.5` isn't yet available on Docker.

I raised this other PR trying to fix the build a different way: https://github.com/apache/druid/pull/18548

gianm:
The Docker image we're using is https://hub.docker.com/_/zookeeper.

PR #18548 resolved the problem for now, so I'm putting this PR into draft mode. Once all the ZK 3.8.5 stuff is fully available we can revisit it.

PR #18547, opened 2025-09-19
Add labels to sys.server

GabrielCWT:
This MR adds support for labels being configured in the config file. It uses the key druid.labels and expects the value to be a JSON structure of key-value pairs.
The labels will be displayed in the web-console under the "Services" tab.

Example: `druid.labels={"broker-label":"broker-value","location": "Airtrunk"}`

Below is a screenshot of the implementation with different labels for different node types. Note that some columns have been hidden to capture the important details.

Screenshot 2025-09-18 at 5 42 01 PM

### DIFF
--- a/docs/querying/sql-metadata-tables.md
+++ b/docs/querying/sql-metadata-tables.md
@@ -236,6 +236,7 @@ Servers table lists all discovered servers in the cluster.
 |max_size|BIGINT|Max size in bytes this server recommends to assign to segments see [druid.server.maxSize](../configuration/index.md#historical-general-configuration). Only valid for HISTORICAL type, for other types it's 0|
 |is_leader|BIGINT|1 if the server is currently the 'leader' (for services which have the concept of leadership), otherwise 0 if the server is not the leader, or null if the server type does not have the concept of leadership|
 |start_time|STRING|Timestamp in ISO8601 format when the server was announced in the cluster|
+|labels|VARCHAR|Labels for the server see [druid.labels](../configuration/index.md)| 
 To retrieve information about all servers, use the query:
 
 ```sql
--- a/server/src/main/java/org/apache/druid/server/DruidNode.java
+++ b/server/src/main/java/org/apache/druid/server/DruidNode.java
@@ -32,11 +32,12 @@ import org.apache.druid.java.util.common.ISE;
 
 import javax.validation.constraints.Max;
 import javax.validation.constraints.NotNull;
-
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.UnknownHostException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 
 /**
@@ -89,6 +90,9 @@ public class DruidNode
       "unknown"
   );
 
+  @JsonProperty
+  private Map<String, String> labels = new HashMap<>();
+
   public DruidNode(
       String serviceName,
       String host,
@@ -99,7 +103,7 @@ public class DruidNode
       boolean enableTlsPort
   )
   {
-    this(serviceName, host, bindOnHost, plaintextPort, null, tlsPort, enablePlaintextPort, enableTlsPort);
+    this(serviceName, host, bindOnHost, plaintextPort, null, tlsPort, enablePlaintextPort, enableTlsPort, null);
   }
 
   /**
@@ -127,7 +131,8 @@ public class DruidNode
       @JacksonInject @Named("servicePort") @JsonProperty("port") Integer port,
       @JacksonInject @Named("tlsServicePort") @JsonProperty("tlsPort") Integer tlsPort,
       @JsonProperty("enablePlaintextPort") Boolean enablePlaintextPort,
-      @JsonProperty("enableTlsPort") boolean enableTlsPort
+      @JsonProperty("enableTlsPort") boolean enableTlsPort,
+      @JsonProperty("labels") Map<String, String> labels
   )
   {
     init(
@@ -137,7 +142,8 @@ public class DruidNode
         plaintextPort != null ? plaintextPort : port,
         tlsPort,
         enablePlaintextPort == null ? true : enablePlaintextPort.booleanValue(),
-        enableTlsPort
+        enableTlsPort,
+        labels
     );
   }
 
@@ -148,7 +154,8 @@ public class DruidNode
       Integer plainTextPort,
       Integer tlsPort,
       boolean enablePlaintextPort,
-      boolean enableTlsPort
+      boolean enableTlsPort,
+      Map<String, String> labels
   )
   {
     Preconditions.checkNotNull(serviceName);
@@ -208,6 +215,12 @@ public class DruidNode
     this.serviceName = serviceName;
     this.host = host;
     this.bindOnHost = bindOnHost;
+    this.labels = labels;
+  }
+
+  public Map<String, String> getLabels()
+  {
+    return labels;
   }
 
   public String getServiceName()
--- a/server/src/test/java/org/apache/druid/server/DruidNodeTest.java
+++ b/server/src/test/java/org/apache/druid/server/DruidNodeTest.java
@@ -285,7 +285,7 @@ public class DruidNodeTest
   public void testSerde1() throws Exception
   {
     DruidNode actual = mapper.readValue(
-        mapper.writeValueAsString(new DruidNode("service", "host", true, 1234, null, 5678, true, true)),
+        mapper.writeValueAsString(new DruidNode("service", "host", true, 1234, null, 5678, true, true, null)),
         DruidNode.class
     );
     Assert.assertEquals("service", actual.getServiceName());
@@ -301,7 +301,7 @@ public class DruidNodeTest
   public void testSerde2() throws Exception
   {
     DruidNode actual = mapper.readValue(
-        mapper.writeValueAsString(new DruidNode("service", "host", false, 1234, null, 5678, null, false)),
+        mapper.writeValueAsString(new DruidNode("service", "host", false, 1234, null, 5678, null, false, null)),
         DruidNode.class
     );
     Assert.assertEquals("service", actual.getServiceName());
@@ -317,7 +317,7 @@ public class DruidNodeTest
   public void testSerde3() throws Exception
   {
     DruidNode actual = mapper.readValue(
-        mapper.writeValueAsString(new DruidNode("service", "host", true, 1234, null, 5678, false, true)),
+        mapper.writeValueAsString(new DruidNode("service", "host", true, 1234, null, 5678, false, true, null)),
         DruidNode.class
     );
     Assert.assertEquals("service", actual.getServiceName());
@@ -344,7 +344,7 @@ public class DruidNodeTest
 
 
     DruidNode actual = mapper.readValue(json, DruidNode.class);
-    Assert.assertEquals(new DruidNode("service", "host", true, 1234, null, 5678, true, true), actual);
+    Assert.assertEquals(new DruidNode("service", "host", true, 1234, null, 5678, true, true, null), actual);
 
     Assert.assertEquals("https", actual.getServiceScheme());
     Assert.assertEquals("host:1234", actual.getHostAndPort());
@@ -365,7 +365,7 @@ public class DruidNodeTest
 
 
     DruidNode actual = mapper.readValue(json, DruidNode.class);
-    Assert.assertEquals(new DruidNode("service", "host", false, 1234, null, 5678, true, false), actual);
+    Assert.assertEquals(new DruidNode("service", "host", false, 1234, null, 5678, true, false, null), actual);
 
     Assert.assertEquals("http", actual.getServiceScheme());
     Assert.assertEquals("host:1234", actual.getHostAndPort());
@@ -385,7 +385,7 @@ public class DruidNodeTest
 
 
     DruidNode actual = mapper.readValue(json, DruidNode.class);
-    Assert.assertEquals(new DruidNode("service", "host", false, 1234, null, 5678, null, false), actual);
+    Assert.assertEquals(new DruidNode("service", "host", false, 1234, null, 5678, null, false, null), actual);
 
     Assert.assertEquals("http", actual.getServiceScheme());
     Assert.assertEquals("host:1234", actual.getHostAndPort());
@@ -405,7 +405,7 @@ public class DruidNodeTest
 
 
     DruidNode actual = mapper.readValue(json, DruidNode.class);
-    Assert.assertEquals(new DruidNode("service", "host", false, null, 1234, 5678, null, false), actual);
+    Assert.assertEquals(new DruidNode("service", "host", false, null, 1234, 5678, null, false, null), actual);
 
     Assert.assertEquals("http", actual.getServiceScheme());
     Assert.assertEquals("host:1234", actual.getHostAndPort());
--- a/sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
+++ b/sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
@@ -180,6 +180,7 @@ public class SystemSchema extends AbstractSchema
       .add("max_size", ColumnType.LONG)
       .add("is_leader", ColumnType.LONG)
       .add("start_time", ColumnType.STRING)
+      .add("labels", ColumnType.STRING)
       .build();
 
   static final RowSignature SERVER_SEGMENTS_SIGNATURE = RowSignature
@@ -244,7 +245,8 @@ public class SystemSchema extends AbstractSchema
             serverInventoryView,
             authorizerMapper,
             overlordClient,
-            coordinatorClient
+            coordinatorClient,
+            jsonMapper
         ),
         SERVER_SEGMENTS_TABLE,
         new ServerSegmentsTable(serverView, authorizerMapper),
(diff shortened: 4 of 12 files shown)


FrankChen021:
Since the label is defined as a Map internally, will these configuration work?
```
druid.label.key1 = val1
druid.label.key2 = val2
```

GabrielCWT:
> Since the label is defined as a Map internally, will these configuration work?
> 
> ```
> druid.label.key1 = val1
> druid.label.key2 = val2
> ```

You are right, I have updated the docs to reflect that both formats are acceptable

kfaraz:
@GabrielCWT , how do you intend to use these labels?
In particular, do they need to be displayed on the web-console and/or be a part of the `DruidNode` object?

GabrielCWT:
> @GabrielCWT , how do you intend to use these labels? In particular, do they need to be displayed on the web-console and/or be a part of the `DruidNode` object?

As of now, the use case is to display the IDC location of the service. Putting it in the web console would make it easier for us to see the locations of all services. 

This feature is similar to the [kubernetes label feature](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) which helps to add identifying attributes to each service.

I felt that since it is attached to each service, we should store it in the `DruidNode`. I am open to suggestions if you feel that this should be stored in a different place.

server/src/main/java/org/apache/druid/server/DruidNode.java
```diff
@@ -89,6 +90,9 @@ public class DruidNode
       "unknown"
   );
 
+  @JsonProperty
+  private Map<String, String> labels = new HashMap<>();
```
FrankChen021: I think the labels is always assigned in ctor, there's no need to assign a default value here.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -687,18 +704,24 @@ private static Object[] buildRowForDiscoverableDataServer(
...
+            node.getHost(),
+            (long) node.getPlaintextPort(),
+            (long) node.getTlsPort(),
+            StringUtils.toLowerCase(discoveryDruidNode.getNodeRole().toString()),
+            druidServerToUse.getTier(),
+            currentSize,
+            druidServerToUse.getMaxSize(),
+            null,
+            toStringOrNull(discoveryDruidNode.getStartTime()),
+            node.getLabels() == null ? null : jsonMapper.writeValueAsString(node.getLabels())
```
FrankChen021: I'm wondering if we can cache the string value of labels in the node to avoid serialization everytime the sys.servers table is accessed.

kfaraz:
Thanks for sharing the details, @GabrielCWT ! That sounds like a reasonable use case.

My only concern was with adding a new type of free-form info to `DruidNode` since it is primarily used for service discovery.
There can be issues with Zookeeper if a node ends up having very large label values,
but I suppose it is not very likely to happen and would be a configuration issue rather than a code issue anyway.

The alternative would be to just bind it to a completely new config object and then it would be served over the `/status/properties` API but then you wouldn't be able to display it on the web-console.

server/src/main/java/org/apache/druid/server/DruidNode.java
kfaraz: +1, the default should just be null.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
kfaraz: This method returns an `Object[]` which can contain a map, why do we need to serialize the labels using a `jsonMapper` at all?

GabrielCWT: > This method returns an Object[] which can contain a map, why do we need to serialize the labels using a jsonMapper at all?

When first implementing the feature, I saw that the nested objects returned are in string format, I have followed this convention. I do agree that it would be cleaner to return it in JSON format. If you are okay with it, I will update it to return JSON instead of it being a string. 

Side Note: Since we are returning it as a string, I suspect the Map's `toString` method is being called under the hood and therefore the return key-value is as follows `"labels": "{brokerTest=myValue, brokerTest2=myValue2}"`.

GabrielCWT:
> My only concern was with adding a new type of free-form info to `DruidNode` since it is primarily used for service discovery. There can be issues with Zookeeper if a node ends up having very large label values, but I suppose it is not very likely to happen and would be a configuration issue rather than a code issue anyway.

This was indeed one of my concerns as well but right now the use case will be a maximum of 2-3 values. As you mentioned, an absurdly large number of label values would be a configuration issue. 

> The alternative would be to just bind it to a completely new config object and then it would be served over the `/status/properties` API but then you wouldn't be able to display it on the web-console.

The main idea is for it to be visible on the web-console and therefore I think sticking to the current implementation is better.

FrankChen021:
> Thanks for sharing the details, @GabrielCWT ! That sounds like a reasonable use case.
> 
> My only concern was with adding a new type of free-form info to `DruidNode` since it is primarily used for service discovery. There can be issues with Zookeeper if a node ends up having very large label values, but I suppose it is not very likely to happen and would be a configuration issue rather than a code issue anyway.
> 
> The alternative would be to just bind it to a completely new config object and then it would be served over the `/status/properties` API but then you wouldn't be able to display it on the web-console.

I think it's better to show the label in the web console. Another use case is that we can inject some container meta data like container name to this property when druid is deployed in K8s, so that we know which service/host maps to which container from web-console

kfaraz:
> I think it's better to show the label in the web console. Another use case is that we can inject some container meta data like container name to this property when druid is deployed in K8s, so that we know which service/host maps to which container from web-console

Yeah, I agree. Labels can have several uses. I was only thinking out loud.
I prefer the current approach myself as long as we don't put unnecessarily large values in the labels.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
kfaraz: Yeah, we can stick to plain JSON since several other system tables are already returning full objects.
For example, `sys.segments` table:

https://github.com/apache/druid/blob/95de4c4b1ea134e32f322efb0e63f1112a64b5ab/sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java#L363-L367

docs/querying/sql-metadata-tables.md
```diff
@@ -236,6 +236,7 @@ Servers table lists all discovered servers in the cluster.
 |max_size|BIGINT|Max size in bytes this server recommends to assign to segments see [druid.server.maxSize](../configuration/index.md#historical-general-configuration). Only valid for HISTORICAL type, for other types it's 0|
 |is_leader|BIGINT|1 if the server is currently the 'leader' (for services which have the concept of leadership), otherwise 0 if the server is not the leader, or null if the server type does not have the concept of leadership|
 |start_time|STRING|Timestamp in ISO8601 format when the server was announced in the cluster|
+|labels|VARCHAR|Labels for the server see [druid.labels](../configuration/index.md)|
```
```suggestion
|labels|VARCHAR|Labels for the server configured using the property [`druid.labels`](../configuration/index.md)| 
```

server/src/main/java/org/apache/druid/server/DruidNode.java
```diff
@@ -89,6 +89,9 @@ public class DruidNode
       "unknown"
   );
 
+  @JsonProperty
+  private Map<String, String> labels;
```
kfaraz, line 96:
```suggestion
  private final Map<String, String> labels;
```

server/src/main/java/org/apache/druid/server/DruidNode.java
```diff
@@ -127,7 +130,8 @@ public DruidNode(
       @JacksonInject @Named("servicePort") @JsonProperty("port") Integer port,
       @JacksonInject @Named("tlsServicePort") @JsonProperty("tlsPort") Integer tlsPort,
       @JsonProperty("enablePlaintextPort") Boolean enablePlaintextPort,
-      @JsonProperty("enableTlsPort") boolean enableTlsPort
+      @JsonProperty("enableTlsPort") boolean enableTlsPort,
+      @JsonProperty("labels") Map<String, String> labels
```
kfaraz:
```suggestion
      @JsonProperty("labels") @Nullable Map<String, String> labels
```

server/src/main/java/org/apache/druid/server/DruidNode.java
```diff
@@ -208,6 +214,12 @@ private void init(
     this.serviceName = serviceName;
     this.host = host;
     this.bindOnHost = bindOnHost;
+    this.labels = labels;
+  }
+
+  public Map<String, String> getLabels()
```
kfaraz, line 224:
```suggestion
  @Nullable
  public Map<String, String> getLabels()
```

server/src/test/java/org/apache/druid/server/DruidNodeTest.java
kfaraz, line 1: Shouldn't we update a test to verify non-null values of `labels`?

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -180,6 +180,7 @@ public class SystemSchema extends AbstractSchema
       .add("max_size", ColumnType.LONG)
       .add("is_leader", ColumnType.LONG)
       .add("start_time", ColumnType.STRING)
+      .add("labels", ColumnType.NESTED_DATA)
```
kfaraz: I don't think any of the other columns use nested data as the type.
Even columns in `sys.segments` like `last_compaction_state`, `dimensions`, `metrics`, etc
use `STRING` as the type. I think we should just stick to that.

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -1077,10 +1092,11 @@ private Object[] createExpectedRow(
       @Nullable Long currSize,
       @Nullable Long maxSize,
       @Nullable Long isLeader,
-      String startTime
+      String startTime,
+      Map<String, String> labels
   )
   {
-    return new Object[]{
+    return new Object[] {
```
kfaraz: Nit: not needed

website/.spelling
```diff
@@ -2070,6 +2070,7 @@ druid.enableTlsPort
 druid.indexer.autoscale.workerVersion
 druid.service
 druid.storage.disableAcl
+druid.labels
```
kfaraz: This shouldn't be needed if we quote the property name as `druid.labels` in the docs.

kfaraz:
@vogievetsky, requested your review since this PR also touches multiple web-console files.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
GabrielCWT: Not sure if I misunderstood our previous discussion but the issue I was facing was that when it was a `STRING` type, the data being returned for the `Map` was in the form `"labels": "{brokerTest=myValue, brokerTest2=myValue2}"`, therefore I used `jsonMapper` to serialise the labels. 

We then agreed to return it in JSON format thus changing it to `NESTED_DATA` type. Just want to confirm that you want to revert it back to the original implementation

```
node.getLabels() == null ? null : jsonMapper.writeValueAsString(node.getLabels())
```

kfaraz: Oh, I see what's happening here. Sorry, I had missed this in the initial pass.
You _will_ need to use a `jsonMapper` and keep the column type as `STRING`
since that is what `SegmentsTable` is doing as well but lazilly.

The segments table uses a list of `SEGMENTS_JSON_FIELDS` to identify which fields to serialize as json.
Ideally, you would want to do something similar to avoid invoking `jsonMapper.writeValueAsString()` every time
we receive a query for `sys.servers` so that we serialize that field only when needed.

But for now, you may just stick to what you had originally that is:
`node.getLabels() == null ? null : jsonMapper.writeValueAsString(node.getLabels())`

The perf improvement can be done later.

Sorry for the confusion 😅 !

kfaraz: It seems that the optimization would also require making `ServersTable` implement `ProjectableFilterableTable` instead of `ScannableTable`. Not sure if that change would have any other impact.

server/src/main/java/org/apache/druid/server/DruidNode.java
GabrielCWT: Currently all variables are being assigned in the `init` function, this prevents the object's variables from being final. I will be following this convention and as such not be assigning `final` to the variable

kfaraz: Ah, I missed the `init()` call. `final` is always better for such bean fields.
But you are right, we cannot do it in this PR.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
FrankChen021: I don't think this optimization helps. because on the web-console, the `label` is always selected, this is the most scenario that `sys.server` table is queried.

Considering the number of servers is not very large(and even the number is huge, at the early phase of this feature, it may not be widely used), the serialization here might NOT be a problem. 

If we're going to optimize the serialization in future, my suggestion is to cache the serialized value in the `DruidNode`.

GabrielCWT: I will open up an issue for a possible refactor to implement `ProjectableFilterableTable`. It would be useful if the number of fields requiring serialising increases. For now I will stick with what was written originally.

> If we're going to optimize the serialization in future, my suggestion is to cache the serialized value in the DruidNode.

This was also an optimisation I was thinking of. What are your thoughts? @kfaraz

FrankChen021: for the serialization problem, I think the fundament problem here is that JSON is not supported by calcite. Not sure if there's a way to make it at the calcite side.

kfaraz: > If we're going to optimize the serialization in future, my suggestion is to cache the serialized value in the DruidNode.

Yeah, I suppose that is possible by passing a `@JacksonInject ObjectMapper` into the `DruidNode` constructor.
But it really seems unnecessary at this point.

@FrankChen021 , as you mention, serializing a small map is not a costly operation, we need not worry about that optimization right now.
I should think that there are many more costlier serialization steps involved in a single `SELECT * FROM sys.servers` query.
If the labels map becomes large, then we have other problems as mentioned before anyway.

kfaraz: > I don't think this optimization helps. because on the web-console, the label is always selected, this is the most scenario that sys.server table is queried.

> Considering the number of servers is not very large(and even the number is huge, at the early phase of this feature, it may not be widely used), the serialization here might NOT be a problem.

+1

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -626,52 +631,64 @@ public Enumerable<Object[]> scan(DataContext root)
...
-          (long) node.getPlaintextPort(),
-          (long) node.getTlsPort(),
-          StringUtils.toLowerCase(discoveryDruidNode.getNodeRole().toString()),
-          null,
-          UNKNOWN_SIZE,
-          UNKNOWN_SIZE,
-          isLeader ? 1L : 0L,
-          toStringOrNull(discoveryDruidNode.getStartTime())
-      };
+      try {
```
kfaraz: Rather than adding the try-catch in the 2 places, better to add a new method in `JacksonUtils` named `writeValueAsString()` which catches the `JsonProcessingException` and throws a `DruidException` instead. It will be useful for other places in the code too.

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -396,7 +396,7 @@ public void setUp(@TempDir File tmpDir) throws Exception
   );
 
   private final DiscoveryDruidNode overlord = new DiscoveryDruidNode(
-      new DruidNode("s2", "localhost", false, 8090, null, true, false),
+      new DruidNode("s2", "localhost", false, 8090, null, null, true, false, ImmutableMap.of("overlordKey", "overlordValue")),
```
kfaraz, line 405: Nit: Consider using `Map.of()` instead of `ImmutableMap.of()` for brevity.

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -1515,6 +1533,9 @@ private static void verifyTypes(final List<Object[]> rows, final RowSignature si
           case STRING:
             expectedClass = String.class;
             break;
+          case COMPLEX:
+            expectedClass = Map.class;
+            break;
```
kfaraz: Is this still needed?

kfaraz (approved):
Left some final non-blocking comments.

+1 after CI passes.

Best to get an approval from @vogievetsky too before merging this off.

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
GabrielCWT: Removed

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
GabrielCWT: I'll stick with `ImmutableMap.of()` to be consistent with other test cases.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
GabrielCWT: Here is my initial implementation, let me know if you have any comments about it.

```
  public static String writeValueAsString(ObjectMapper jsonMapper, Object value) throws DruidException
  {
    try {
      return jsonMapper.writeValueAsString(value);
    }
    catch (JsonProcessingException e) {
      throw InvalidInput.exception(e, "Failed to serialize object as JSON");
    }
  }
  ```

kfaraz: `throw InternalServerError.exception()` since we don't know exactly why the serialization failed.

FrankChen021:
Hi @vogievetsky 
I would like this one to be merged in 35. If you're too busy to review the changes, I will merge it first.

PR #18549, opened 2025-09-19
Enable fallback vectorization by default.

gianm:
This patch updates the default value of druid.expressions.allowVectorizeFallback to true.

In addition, there are the following fixes and improvements:

1) Remove safe_divide's explicit "return false" override for canVectorize,
   allowing it to fallback vectorize.

2) Fix getExponent's long-typed vectorized implementation. (It was casting
   to float, so the exponent was inconsistent with the non-vector version,
   which cast to double.)

3) Fix json_query_array's reported output type (was `COMPLEX<json>`, should
   be `ARRAY<COMPLEX<json>>`).

4) FallbackVectorProcessor and Div now have their errors wrapped using
   FunctionErrorReportingExprVectorProcessor, which handles errors
   similarly to how FunctionExpr does it. (Div is the only already-vectorized
   math function that we expect to throw errors).

5) ColumnarFrameCursorFactory now uses itself as the inspector for
   canVectorize, rather than its signature. The signature returns
   "unknown" for "hasMultipleValues", which leads to things being
   non-vectorizable that really should be vectorizable.

6) Add many more functions to VectorExprResultConsistencyTest, and update
   it to test negative numbers and zero. Because zero can lead to exceptions
   in division functions, the test code is also updated to verify equivalence
   of the thrown exceptions.

7) Call asSingleThreaded on Functions and ExprMacro exprs when they
   run in FallbackVectorProcessor.

### DIFF
--- a/processing/src/main/java/org/apache/druid/math/expr/vector/FunctionErrorReportingExprVectorProcessor.java
+++ b/processing/src/main/java/org/apache/druid/math/expr/vector/FunctionErrorReportingExprVectorProcessor.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.math.expr.vector;
+
+import org.apache.druid.error.DruidException;
+import org.apache.druid.math.expr.Expr;
+import org.apache.druid.math.expr.ExpressionProcessingException;
+import org.apache.druid.math.expr.ExpressionType;
+import org.apache.druid.math.expr.ExpressionValidationException;
+import org.apache.druid.segment.column.Types;
+
+/**
+ * Wrapper around {@link ExprVectorProcessor} that reports errors in the same style as {@code FunctionExpr}.
+ */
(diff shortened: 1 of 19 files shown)


processing/src/test/java/org/apache/druid/math/expr/VectorExprResultConsistencyTest.java
```diff
@@ -840,4 +961,62 @@ static NonnullPair<Expr.ObjectBinding[], Expr.VectorInputBinding> populateBindin
...
+    catch (ArithmeticException e) {
+      // After a few occasions, Java starts throwing ArithmeticException without a message for division by zero.
+      return Either.error(e.getClass().getName());
+    }
+    catch (Exception e) {
+      return Either.error(e.toString());
+    }
+
+    final Object[] vectorVals = vectorEval.getObjectVector();
+    // 'null' expressions can have an output type of null, but still evaluate in default mode, so skip type checks
```
clintropolis: nit: i know these aren't new comments (there is another one a bit later), but consider removing reference to default value mode, i think i missed this, its just talking about how outputType isn't reliable for null values so it was skipping comparison

processing/src/main/java/org/apache/druid/math/expr/vector/FunctionErrorReportingExprVectorProcessor.java
```diff
@@ -0,0 +1,79 @@
...
+                          .ofCategory(DruidException.Category.INVALID_INPUT)
+                          .build(e, e.getMessage());
+    }
+    catch (Types.InvalidCastException | Types.InvalidCastBooleanException e) {
+      throw DruidException.forPersona(DruidException.Persona.USER)
+                          .ofCategory(DruidException.Category.INVALID_INPUT)
+                          .build(e, "Function[%s] encountered exception: %s", functionName, e.getMessage());
+    }
+    catch (DruidException e) {
+      throw e;
```
clintropolis, line 61: i guess there are probably some cases where this path loses the function name? I'm not sure if it matters all that much, but this is part of why i've been a bit hesitant on migrating to using `DruidException` inside expr processing in favor of still using the expression specific exceptions and catching them in the callers to turn into `DruidException`. I feel like its kind of hard to know if a more general exception like DruidException should be caught and wrapped or rethrown.

All that said, this is fine i think too, since this is mostly just my opinion

clintropolis: ah looking around a bit more, is already a mix, so this is fine

PR #18553, opened 2025-09-19
SQL: Use specialized virtual columns for JSON_OBJECT, JSON_MERGE.

gianm:
Building on #18521, this patch makes virtual column specialization recursive and adds specializations for two additional JSON functions. The result is that chains of JSON_OBJECT, JSON_MERGE, and JSON_VALUE can preserve lazy evaluation, index usage, dictionary usage, etc.

There is a change to VirtualColumnCreator that can affect extensions that add SQL operators. To allow the creator to access rewritten arguments, a "DruidExpression self" parameter is added. The "String expression" is no longer needed so it is removed.


sql/src/main/java/org/apache/druid/sql/calcite/rel/DruidQuery.java
```diff
@@ -723,35 +724,15 @@ private static RowSignature computeOutputRowSignature(
...
-    // the various transforms and optimizations
-    Set<VirtualColumn> virtualColumns = new HashSet<>();
-
-
-    // rewrite any "specialized" virtual column expressions as top level virtual columns so that their native
-    // implementation can be used instead of being composed as part of some expression tree in an expresson virtual
-    // column
-    Set<String> specialized = new HashSet<>();
-    final boolean forceExpressionVirtualColumns =
-        plannerContext.getPlannerConfig().isForceExpressionVirtualColumns();
```
clintropolis, line 739: it seems like this flag is no longer honored, should we push the config into the virtual column registry?

gianm: I did do that, it's pushed in and if true then `VirtualColumnRegistry#specialize` exits without doing anything.

clintropolis: ah so it is 😅

PR #18567, opened 2025-09-24
Include IllegalArgumentException in vector processing error report

cecemei:
### Description

Include IllegalArgumentException in vector processing error report.

This PR has:

- [ ] been self-reviewed.


processing/src/main/java/org/apache/druid/math/expr/vector/FunctionErrorReportingExprVectorProcessor.java
```diff
@@ -57,7 +58,7 @@ public ExprEvalVector<T> evalVector(Expr.VectorInputBinding bindings)
                           .ofCategory(DruidException.Category.INVALID_INPUT)
                           .build(e, "Function[%s] encountered exception: %s", functionName, e.getMessage());
     }
-    catch (DruidException e) {
+    catch (DruidException | IAE e) {
```
kfaraz: I think it might be better to catch the IAE and convert it to a DruidException of type INVALID_INPUT instead, as it contains richer info for sending over the wire.

cecemei: Updated to convert to DruidException

processing/src/main/java/org/apache/druid/math/expr/vector/FunctionErrorReportingExprVectorProcessor.java
```diff
@@ -57,6 +58,11 @@ public ExprEvalVector<T> evalVector(Expr.VectorInputBinding bindings)
                           .ofCategory(DruidException.Category.INVALID_INPUT)
                           .build(e, "Function[%s] encountered exception: %s", functionName, e.getMessage());
     }
+    catch (IAE e) {
```
gianm, line 61: IMO, this is better dealt with inside each expression.

The idea with the code block here (and the equivalent one in `FunctionExpr`) is to pass through errors that the function clearly intended to be aimed at the end user (ExpressionValidationException, ExpressionProcessingException, DruidException, etc) and to map all others to an "unknown exception" message with the cause attached for debugging purposes.

So, if an `IAE` is making it out here with a user-facing error message, I think it's a sign that the function that generated the IAE should improve its error handling.

cecemei: Make sense. Will deal in expression.

PR #18568, opened 2025-09-24
Add support for caching ScanQuery results

jtuglu1:
### Description

Adds support for result-level caching of scan queries. Despite being a relatively infrequent query type, a small, sustained volume of these queries can put significant load on a cluster as these don't run on processing threads and tend to eat a lot of CPU + return big result sets.

#### Errors/Bugs
This PR also unearthed some bugs w.r.t segment-level cache handling on the Broker, as well as `bySegment` query cache handling in general. All cacheable query types which do not implement a proper mapping of result type `T` to `BySegmentResultValueClass<T>` (in other words, all but `topN` and `search`) will fail in the following scenarios:
- `bySegment:true` is passed in the query context.
- Broker has segment-level cache enabled. This is because [these](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/client/CachingClusteredClient.java#L313-L314) 2 lines effectively "force" the data node to return in bySegment result format, but caching a bySegment result breaks for all query types except `topN` + `search`, as they have their own custom `BySegmentResultValueClass` classes. This exception is thrown in the [first call](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/client/CachingClusteredClient.java#L773) to cache entry creation.

An example is as follows:
```
2025-09-28T21:34:48,651 ERROR [main] org.apache.druid.testsEx.indexer.AbstractITBatchIndexTest - Error while running test query at path /indexer/wikipedia_index_schemaless_queries.json
java.lang.RuntimeException: org.apache.druid.java.util.common.ISE: Error while querying[http://localhost:8082/druid/v2/sql] status[500 Server Error] content[{"error":"Unknown exception","errorClass":"java.lang.ClassCastException","host":"historical:8083","errorCode":"legacyQueryException","persona":"OPERATOR","category":"RUNTIME_FAILURE","errorMessage":"class org.apache.druid.query.Result cannot be cast to class org.apache.druid.query.scan.ScanResultValue (org.apache.druid.query.Result and org.apache.druid.query.scan.ScanResultValue are in unnamed module of loader 'app')","context":{"host":"historical:8083","errorClass":"java.lang.ClassCastException","legacyErrorCode":"Unknown exception"}}]
```

Attempts to coerce this new type of object `Result` into `ScanResultValue`, causing errors. Currently, we pin the type of the first template parameter `T` of [CacheStrategy](https://github.com/apache/druid/blob/-/processing/src/main/java/org/apache/druid/query/CacheStrategy.java#L38) to the type of the `Query<T>`, so we cannot support variable input types with `Object` (e.g. sometimes get `ScanResultValue` and sometimes `Result<BySegmentResultValue<ScanResultValue>>`). The conversion between these types is also a bit cumbersome as these return an iterable of `ScanResultValue`, but the segment cache expects a single `ScanResultValue`. Other query types that support `Result<T>` are `Iterable` and therefore are easier to coerce (see [BySegmentTopNResultValue](https://github.com/apache/druid/blob/-/processing/src/main/java/org/apache/druid/query/topn/BySegmentTopNResultValue.java#L33)). I think it would be doable to create a similar BySegment child for each of `ScanResultValue` and `SegmentAnalysis`, if it could be guaranteed that the `getResults()` was always a list of length 1.

##### Before
- Result-level/Segment-level caching did not work for scan queries
##### After
- Result-level caching will work iff:
  - The query is NOT `bySegment`.
- Segment-level caching will work iff:
  - The query is NOT `bySegment`.
  - The node is NOT a broker.

#### Alternatives considered:
I could change `CacheStrategy`'s templating layout and `ScanQuery` and `SegmentMetadataQuery` both get/set their cache values, but given these queries ALREADY return data keyed by segment, it didn't seem worth it. The caching would already be [disabled](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/client/CachingClusteredClient.java#L805) anyways for these `bySegment` queries. It's also my understanding that `bySegment` is primarily for debugging anyways.

#### Release note

Adds support for result-level caching of scan queries. 

This caching behavior is disabled by default for scan queries. To enable, override the `druid.*.cache.unCacheable` property.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/query/CacheStrategy.java
+++ b/processing/src/main/java/org/apache/druid/query/CacheStrategy.java
@@ -62,11 +62,11 @@ public interface CacheStrategy<T, CacheType, QueryType extends Query<T>>
    * @param query            the query to be cached
    * @param willMergeRunners indicates that {@link QueryRunnerFactory#mergeRunners(QueryProcessingPool, Iterable)} will be
    *                         called on the cached by-segment results
-   * @param bySegment        segment level or result level cache
+   * @param segmentLevel        segment level or result level cache
    *
    * @return true if the query is cacheable, otherwise false.
    */
-  default boolean isCacheable(QueryType query, boolean willMergeRunners, boolean bySegment)
+  default boolean isCacheable(QueryType query, boolean willMergeRunners, boolean segmentLevel)
   {
     return isCacheable(query, willMergeRunners);
   }
--- a/processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
+++ b/processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
@@ -52,6 +52,7 @@ import java.util.Optional;
 public class ScanQueryQueryToolChest extends QueryToolChest<ScanResultValue, ScanQuery>
 {
   private static final byte SCAN_QUERY = 0x13;
+  private static final byte CACHE_STRATEGY_VERSION = 0x1;
   private static final TypeReference<ScanResultValue> TYPE_REFERENCE = new TypeReference<>() {};
 
   private final GenericQueryMetricsFactory queryMetricsFactory;
@@ -219,6 +220,8 @@ public class ScanQueryQueryToolChest extends QueryToolChest<ScanResultValue, Sca
       public byte[] computeCacheKey(ScanQuery query)
       {
         CacheKeyBuilder builder = new CacheKeyBuilder(SCAN_QUERY)
+            .appendByte(CACHE_STRATEGY_VERSION)
+            .appendCacheable(query.getDataSource())
             .appendCacheable(query.getVirtualColumns())
             .appendString(query.getResultFormat().toString())
             .appendLong(query.getScanRowsOffset())
@@ -227,8 +230,9 @@ public class ScanQueryQueryToolChest extends QueryToolChest<ScanResultValue, Sca
             .appendStrings(query.getColumns() != null ? query.getColumns() : List.of())
             .appendString(query.getTimeOrder().toString());
 
-        if (query.getOrderBys() != null && !query.getOrderBys().isEmpty()) {
-          for (OrderBy orderBy : query.getOrderBys()) {
+        List<OrderBy> orderBys = query.getOrderBys();
+        if (orderBys != null) {
+          for (OrderBy orderBy : orderBys) {
             builder.appendString(orderBy.getColumnName())
                    .appendString(orderBy.getOrder().toString());
           }
--- a/processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
+++ b/processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
@@ -41,6 +41,7 @@ import org.apache.druid.query.QueryRunner;
 import org.apache.druid.query.QueryToolChest;
 import org.apache.druid.query.aggregation.MetricManipulationFn;
 import org.apache.druid.query.cache.CacheKeyBuilder;
+import org.apache.druid.segment.column.ColumnType;
 import org.apache.druid.segment.column.RowSignature;
 import org.apache.druid.utils.CloseableUtils;
 
@@ -221,21 +222,22 @@ public class ScanQueryQueryToolChest extends QueryToolChest<ScanResultValue, Sca
       {
         CacheKeyBuilder builder = new CacheKeyBuilder(SCAN_QUERY)
             .appendByte(CACHE_STRATEGY_VERSION)
-            .appendCacheable(query.getDataSource())
             .appendCacheable(query.getVirtualColumns())
             .appendString(query.getResultFormat().toString())
             .appendLong(query.getScanRowsOffset())
             .appendLong(query.getScanRowsLimit())
             .appendCacheable(query.getFilter())
             .appendStrings(query.getColumns() != null ? query.getColumns() : List.of())
-            .appendString(query.getTimeOrder().toString());
+            .appendCacheable(query.getTimeOrder());
 
         List<OrderBy> orderBys = query.getOrderBys();
         if (orderBys != null) {
-          for (OrderBy orderBy : orderBys) {
-            builder.appendString(orderBy.getColumnName())
-                   .appendString(orderBy.getOrder().toString());
-          }
+          builder.appendCacheables(orderBys);
+        }
+
+        List<ColumnType> columnTypes = query.getColumnTypes();
+        if (columnTypes != null) {
+          builder.appendCacheables(columnTypes);
         }
 
         return builder.build();
--- a/processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
+++ b/processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
@@ -20,6 +20,7 @@
 package org.apache.druid.query.scan;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Function;
 import com.google.common.base.Functions;
 import com.google.common.collect.Iterables;
@@ -30,22 +31,28 @@ import org.apache.druid.java.util.common.UOE;
 import org.apache.druid.java.util.common.guava.BaseSequence;
 import org.apache.druid.java.util.common.guava.Sequence;
 import org.apache.druid.java.util.common.guava.Sequences;
+import org.apache.druid.query.CacheStrategy;
 import org.apache.druid.query.FrameSignaturePair;
 import org.apache.druid.query.GenericQueryMetricsFactory;
+import org.apache.druid.query.OrderBy;
 import org.apache.druid.query.Query;
 import org.apache.druid.query.QueryMetrics;
 import org.apache.druid.query.QueryRunner;
 import org.apache.druid.query.QueryToolChest;
 import org.apache.druid.query.aggregation.MetricManipulationFn;
+import org.apache.druid.query.cache.CacheKeyBuilder;
+import org.apache.druid.segment.column.ColumnType;
 import org.apache.druid.segment.column.RowSignature;
 import org.apache.druid.utils.CloseableUtils;
 
+import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
 public class ScanQueryQueryToolChest extends QueryToolChest<ScanResultValue, ScanQuery>
 {
+  private static final byte SCAN_QUERY = 0x13;
   private static final TypeReference<ScanResultValue> TYPE_REFERENCE = new TypeReference<>() {};
 
   private final GenericQueryMetricsFactory queryMetricsFactory;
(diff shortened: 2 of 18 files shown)


processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
```diff
@@ -191,6 +198,80 @@ public Sequence<Object[]> resultsAsArrays(final ScanQuery query, final Sequence<
...
+        //   - A query is issued with bySegment:true
+        //   - Segment-level cache is enabled on the broker (in which case it sends bySegment queries to data nodes).
+        return !query.context().isBySegment() && (!segmentLevel || willMergeRunners);
+      }
+
+      @Override
+      public byte[] computeCacheKey(ScanQuery query)
+      {
+        CacheKeyBuilder builder = new CacheKeyBuilder(SCAN_QUERY)
+            .appendCacheable(query.getVirtualColumns())
```
samarthjain, line 225: I see `GroupByQueryToolChest` has a CACHE_STRATEGY_VERSION. I suppose the intent of having versioning is to discard old data from the cache if the cache key structure changes. It would make sense to add it here as well to ensure key evolution doesn't cause any unpleasant gotchas in the future.

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
```diff
@@ -191,6 +198,80 @@ public Sequence<Object[]> resultsAsArrays(final ScanQuery query, final Sequence<
...
+            .appendString(query.getResultFormat().toString())
```
samarthjain, line 226: Do you know if resultFormat is used after getting results out of the result set cache or before the results are written to the cache? If former, we probably can remove it from the cache key.

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
```diff
@@ -191,6 +198,80 @@ public Sequence<Object[]> resultsAsArrays(final ScanQuery query, final Sequence<
...
+            .appendInt(query.getBatchSize())
+            .appendLong(query.getScanRowsOffset())
+            .appendLong(query.getScanRowsLimit())
+            .appendCacheable(query.getFilter())
+            .appendStrings(query.getColumns() != null ? query.getColumns() : List.of())
+            .appendString(query.getTimeOrder().toString());
+
+        if (query.getOrderBys() != null && !query.getOrderBys().isEmpty()) {
```
samarthjain: nit: the empty check can be removed since the iterator in the loop will check that internally. 
```
List<OrderBy> orderBys = query.getOrderBys();
if (orderBys != null) {
    for (OrderBy orderBy : orderBys) {
        builder.appendString(orderBy.getColumnName())
               .appendString(orderBy.getOrder().toString());
    }
}
```

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
```diff
@@ -191,6 +198,80 @@ public Sequence<Object[]> resultsAsArrays(final ScanQuery query, final Sequence<
...
+          for (OrderBy orderBy : query.getOrderBys()) {
+            builder.appendString(orderBy.getColumnName())
+                   .appendString(orderBy.getOrder().toString());
+          }
+        }
+
+        if (query.getColumnTypes() != null && !query.getColumnTypes().isEmpty()) {
```
samarthjain: Same as above.

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
samarthjain, line 225: Interestingly, I don't see datasource being used in the cache key. Is that check performed somewhere else? It would be good to add a test like `testCacheKeyDifferentQueries` where you have everything else the same but the datasource names are different. If the test fails, I suppose we may have discovered a latent bug :)

processing/src/test/java/org/apache/druid/query/scan/ScanQueryQueryToolChestTest.java
```diff
@@ -441,4 +447,247 @@ private static Sequence<ScanResultValue> results3()
...
+        .build();
+
+    ScanQuery query2 = Druids.newScanQueryBuilder()
+        .dataSource("foo")
+        .intervals(new MultipleIntervalSegmentSpec(ImmutableList.of(Intervals.of("2025-01-01/2025-01-02"))))
+        .columns("dim1", "dim3")
+        .resultFormat(ScanQuery.ResultFormat.RESULT_FORMAT_LIST)
+        .build();
+
+    ScanQuery query3 = Druids.newScanQueryBuilder()
```
samarthjain, line 540: Just like you have tests for other fields, it might make sense to separate out the test for result format on its own. Also, we may or may need to have result format as the cache key in which case this test may not be needed.

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
jtuglu1: Good question, I had wondered this. It's the [latter](https://github.com/apache/druid/blob/-/processing/src/main/java/org/apache/druid/query/scan/ScanQueryEngine.java#L178).

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
jtuglu1: This was removed in prior commit, not relevant to the cache.

processing/src/test/java/org/apache/druid/query/scan/ScanQueryQueryToolChestTest.java
jtuglu1: I'll add another test, but it's needed in the cache key.

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
jtuglu1: Yeah, this I had also wondered. Bit of an odd design, but result level cache will actually [append](https://github.com/apache/druid/blob/-/processing/src/main/java/org/apache/druid/query/CacheStrategy.java#L88) datasource/interval already. That also happens [here](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/client/CachingClusteredClient.java#L870) and [here](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/client/CachingClusteredClient.java#L850) for both segment-level and result-level cache.

I personally would've just included these in the cache key so it's easily testable e2e but maybe there's some distinction here like sub-queries, etc.

processing/src/main/java/org/apache/druid/query/CacheStrategy.java
```diff
@@ -62,11 +62,11 @@ default boolean isCacheable(QueryType ignoredQuery, boolean ignoredWillMergeRunn
    * @param query            the query to be cached
    * @param willMergeRunners indicates that {@link QueryRunnerFactory#mergeRunners(QueryProcessingPool, Iterable)} will be
    *                         called on the cached by-segment results
-   * @param bySegment        segment level or result level cache
+   * @param segmentLevel        segment level or result level cache
```
maytasm, line 65: nit: whitespace/offset

jtuglu1:
> * Scan query with limit is not cached. This is because the etag is populated after createResultLevelCachePopulator is called, resulting in a null resultLevelCachePopulator.   In case you want to take a look at that in this PR. I'm also fine if we address that later too. Btw this might be a common use case with the LIMIT / OFFSET to support pagination.

I think this can be addressed in a follow-up PR. That would need to materialize in some form of "lazy" `ResultLevelCachePopulator` which would only be created once the etag was populated (e.g. after ClusteredClient is run).

> * I think scan query intentionally did not implement caching. This is because the result set of scan query is typically large and will quickly fill the cache up causing higher cache miss rate. I think we should still make it possible to cache scan query but it should be disable by default given the above reasoning.

I will add this to the default `druid...unCacheable=["scan"]` list. This way users need to opt-in to enable it.

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
```diff
@@ -191,6 +198,76 @@ public Sequence<Object[]> resultsAsArrays(final ScanQuery query, final Sequence<
...
+      {
+        // Currently, there is no bijective mapping from ScanResultValue to Result<BySegmentResultValueClass<ScanResultValue>>.
+        // This means queries will fail if:
+        //   - A query is issued with bySegment:true
+        //   - Segment-level cache is enabled on the broker (in which case it sends bySegment queries to data nodes).
+        return !query.context().isBySegment() && (!segmentLevel || willMergeRunners);
+      }
+
+      @Override
+      public byte[] computeCacheKey(ScanQuery query)
```
maytasm, line 221: Do we need the intervals and timeOrder?

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
```diff
@@ -191,6 +198,76 @@ public Sequence<Object[]> resultsAsArrays(final ScanQuery query, final Sequence<
...
+      }
+
+      @Override
+      public TypeReference<ScanResultValue> getCacheObjectClazz()
+      {
+        return TYPE_REFERENCE;
+      }
+
+      @Override
+      public Function<ScanResultValue, ScanResultValue> prepareForCache(boolean isResultLevelCache)
```
maytasm, line 260: Why are the cache for other queryType converting to Object?

jtuglu1: Because of perSegment query result type (it can be `BySegmentResultValue<T>` or the actual value `T`). See the explanation in the PR description.

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
jtuglu1: See my comment above. Samarth asked the same question.

maytasm: We don't need the intervals but will still need timeOrder and maybe columnTypes too

maytasm (approved):
Please address the comment on the computeCacheKey and add release note regarding the change in default for the unCacheable property

processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java
jtuglu1: [timeOrder](https://github.com/jtuglu1/druid/blob/afa7936c499bf385eb1e96256a8b35c3a2a5d62d/processing/src/main/java/org/apache/druid/query/scan/ScanQueryQueryToolChest.java#L231) is already there

jtuglu1: Adding back columnTypes

Issue #11652, opened 2021-09-01
QueryCountStatsMonitor doesn't track sql query metrics


a2l007:
QueryCountStatsMonitor provides the following query metrics
`query/count` 
`query/success/count`
`query/failed/count`
`query/interrupted/count`
 `query/timeout/count`
It looks like these metrics don't track the sql queries invoked against the `druid/v2/sql` endpoint.
This could be because the Broker is tied to `BrokerQueryResource` as the `QueryCountStatsProvider` implementation and even if `SqlResource` implements `QueryCountStatsProvider`, it wouldn't work. 
Unless the SqlResource was designed as such for some other reason, I'm looking at merging the `SqlResource` into `QueryResource` so that `druid/v2/sql` can take advantage of the query count stats monitoring within `QueryResource`.

@jihoonson Any thoughts on this?

maytasm:
@a2l007 Did you have a chance to work on this?

a2l007:
@maytasm Nope, but looks like there is already a PR out for this!

maytasm:
@a2l007 Yep. If you have a few minutes, would appreciate if you can review the PR https://github.com/apache/druid/pull/17452 Thanks!

PR #18569, opened 2025-09-24
Support verbose SQL query status metrics

jtuglu1:
Fixes https://github.com/apache/druid/issues/11652.

### Description

Ensures `query/success/count`, `query/failed/count`, `query/interrupted/count`, `query/timeout/count` are emitted for both query types (native + sql).

This pulls out the query count metrics logic from QueryResource and places it into BaseQueryCounterResource, which allows for SqlResource to share this codepath. An alternative approach is [here](https://github.com/jtuglu-netflix/druid/tree/add-query-metrics-for-sql), which avoids shared logic between the two resources. As stated in the issue, the ideal change would be to have SqlResource inherit, or simply pass on, all query handling to QueryResource, but that seemed a bit out of scope for this PR (but I can do this).

Also simplifies/unifies the query count metric tracking architecture between the multiple query resource classes.

#### Release note

Emit verbose query count metrics for SQL queries.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/test/java/org/apache/druid/server/QueryResourceTest.java
+++ b/server/src/test/java/org/apache/druid/server/QueryResourceTest.java
@@ -595,7 +595,8 @@ public class QueryResourceTest
             ResponseContextConfig.newConfig(true),
             DRUID_NODE
         ),
-        new ResourceIOReaderWriterFactory(jsonMapper, smileMapper)
+        new ResourceIOReaderWriterFactory(jsonMapper, smileMapper),
+        queryCountStatsProvider
     );
 
     expectPermissiveHappyPathAuth();


server/src/main/java/org/apache/druid/server/metrics/BaseQueryCountResource.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package org.apache.druid.server.metrics;
+
+import org.apache.druid.guice.LazySingleton;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/*
+  A thread-safe query count statistics tracker. Meant to be a shared between the various query resource handlers present
```
jtuglu1: will fix typo after review to minimize CI runs

server/src/main/java/org/apache/druid/server/QueryResource.java
```diff
@@ -97,7 +97,7 @@ public class QueryResource implements QueryCountStatsProvider
   private final AtomicLong failedQueryCount = new AtomicLong();
   private final AtomicLong interruptedQueryCount = new AtomicLong();
   private final AtomicLong timedOutQueryCount = new AtomicLong();
-  private final QueryResourceQueryMetricCounter counter = new QueryResourceQueryMetricCounter();
+  final QueryCountStatsProvider counter;
```
kfaraz:
```suggestion
  private final QueryCountStatsProvider counter;
```

server/src/test/java/org/apache/druid/server/metrics/QueryCountStatsMonitorTest.java
```diff
@@ -55,37 +56,57 @@ public void setUp()
 
     queryCountStatsProvider = new QueryCountStatsProvider()
```
kfaraz: Should we just use the new impl `BaseQueryCountResource` here?

server/src/test/java/org/apache/druid/server/metrics/QueryCountStatsMonitorTest.java
```diff
@@ -55,37 +56,57 @@ public void setUp()
...
       {
-        interruptedEmitCount += 3;
-        return interruptedEmitCount;
+        return interruptedEmitCount.get();
       }
 
       @Override
       public long getTimedOutQueryCount()
       {
-        timedOutEmitCount += 4;
```
kfaraz, line 87: Why was the increment +4? 😅

server/src/test/java/org/apache/druid/server/metrics/QueryCountStatsMonitorTest.java
```diff
@@ -106,15 +127,23 @@ public void testMonitor()
...
+
     emitter.flush();
     // Trigger metric emission
     monitor.doMonitor(emitter);
     Assert.assertEquals(5, emitter.getNumEmittedEvents());
     emitter.verifyValue("query/success/count", 1L);
     emitter.verifyValue("query/failed/count", 2L);
-    emitter.verifyValue("query/interrupted/count", 3L);
-    emitter.verifyValue("query/timeout/count", 4L);
-    emitter.verifyValue("query/count", 10L);
```
kfaraz, line 117: I assume these had to be changed due to the weird increments in the test `QueryCountStatsProvider` defined above?

server/src/test/java/org/apache/druid/server/metrics/QueryCountStatsMonitorTest.java
```diff
@@ -126,6 +155,14 @@ public void testMonitor_emitPendingRequests()
         new QueryCountStatsMonitor(queryCountStatsProvider, monitorsConfig, mergeBufferPool);
     final StubServiceEmitter emitter = new StubServiceEmitter("service", "host");
     monitor.doMonitor(emitter);
+
+    // Mock metrics emission
```
kfaraz:
```suggestion
    // Simulate metrics emission
```

server/src/test/java/org/apache/druid/server/QueryResourceTest.java
```diff
@@ -769,7 +777,7 @@ public void testTruncatedResponseContextShouldFail() throws IOException
         SIMPLE_TIMESERIES_QUERY.getBytes(StandardCharsets.UTF_8),
         queryResource
     );
-    Assert.assertEquals(1, queryResource.getInterruptedQueryCount());
+    Assert.assertEquals(1, queryResource.counter.getInterruptedQueryCount());
```
kfaraz: Let's not access the `counter` via the `QueryResource`
since it doesn't really belong to the resource and is a global service-level entity.
We should just hold a reference to the `counter` itself.

server/src/test/java/org/apache/druid/server/QueryResourceTest.java
```diff
@@ -1079,7 +1089,7 @@ public <T> QueryRunner<T> getQueryRunnerForSegments(Query<T> query, Iterable<Seg
     Assert.assertEquals("Query did not complete within configured timeout period. You can " +
                         "increase query timeout or tune the performance of query.", ex.getMessage());
     Assert.assertEquals(QueryException.QUERY_TIMEOUT_ERROR_CODE, ex.getErrorCode());
-    Assert.assertEquals(1, timeoutQueryResource.getTimedOutQueryCount());
+    Assert.assertEquals(1, timeoutQueryResource.counter.getTimedOutQueryCount());
```
kfaraz: Same comment here regarding accessing the `counter` via the resource class.

server/src/test/java/org/apache/druid/server/QueryResourceTest.java
```diff
@@ -1379,8 +1391,8 @@ public void testTooManyQuery() throws InterruptedException, ExecutionException
     for (Future<Boolean> theFuture : back2) {
       Assert.assertTrue(theFuture.get());
     }
-    Assert.assertEquals(2, queryResource.getSuccessfulQueryCount());
-    Assert.assertEquals(1, queryResource.getFailedQueryCount());
+    Assert.assertEquals(2, queryResource.counter.getSuccessfulQueryCount());
+    Assert.assertEquals(1, queryResource.counter.getFailedQueryCount());
```
kfaraz: Same.

server/src/test/java/org/apache/druid/server/QueryResultPusherTest.java
```diff
@@ -134,9 +133,33 @@ public ResultsWriter start()
     assertTrue("recordFailure(e) should have been invoked!", recordFailureInvoked.get());
   }
 
-  static class NoopQueryMetricCounter implements QueryMetricCounter
+  static class NoopQueryMetricCounter implements QueryCountStatsProvider
```
kfaraz: I suppose this might not be needed anymore either.
We may just use the new impl `BaseQueryCountResource`, unless something in the test asserts that some metric is zero (which we can update to reflect the true value).

services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
```diff
@@ -159,7 +160,8 @@ public AsyncQueryForwardingServlet(
       GenericQueryMetricsFactory queryMetricsFactory,
       AuthenticatorMapper authenticatorMapper,
       Properties properties,
-      final ServerConfig serverConfig
+      final ServerConfig serverConfig,
+      QueryCountStatsProvider counter
```
kfaraz: Nit: for homogeneity with the other fields
```suggestion
      final QueryCountStatsProvider counter
```

sql/src/main/java/org/apache/druid/sql/http/SqlHttpModule.java
```diff
@@ -35,6 +37,7 @@ public class SqlHttpModule implements Module
   public void configure(Binder binder)
   {
     binder.bind(SqlResource.class).in(LazySingleton.class);
+    binder.bind(QueryCountStatsProvider.class).to(BaseQueryCountResource.class).in(LazySingleton.class);
```
kfaraz: This is already bound on `CliBroker` which is the only service which uses `SqlHttpModule` (AFAICT).
Do we still need this additional binding?

sql/src/main/java/org/apache/druid/sql/http/SqlResource.java
```diff
@@ -80,14 +81,14 @@ public class SqlResource
...
   private static final Logger log = new Logger(SqlResource.class);
-  public static final SqlResourceQueryMetricCounter QUERY_METRIC_COUNTER = new SqlResourceQueryMetricCounter();
 
   private final AuthorizerMapper authorizerMapper;
   private final SqlResourceQueryResultPusherFactory resultPusherFactory;
   private final SqlLifecycleManager sqlLifecycleManager;
   private final SqlEngineRegistry sqlEngineRegistry;
   private final DefaultQueryConfig defaultQueryConfig;
   private final ServerConfig serverConfig;
+  final QueryCountStatsProvider counter;
```
kfaraz:
```suggestion
  private final QueryCountStatsProvider counter;
```

sql/src/test/java/org/apache/druid/sql/http/SqlResourceTest.java
```diff
@@ -1869,7 +1876,9 @@ public void testTooManyRequestsAfterTotalLaning() throws Exception
       }
     }
     Assert.assertEquals(2, success);
+    Assert.assertEquals(2, resource.counter.getSuccessfulQueryCount());
     Assert.assertEquals(1, limited);
+    Assert.assertEquals(1, resource.counter.getFailedQueryCount());
```
kfaraz: We shouldn't access `counter` via the resource class.

server/src/main/java/org/apache/druid/server/metrics/BaseQueryCountResource.java
```diff
@@ -0,0 +1,85 @@
...
+  on an instance.
+*/
+@LazySingleton
```
kfaraz: No need to annotate since we are already binding it appropriately where applicable.

sql/src/main/java/org/apache/druid/sql/http/SqlResource.java
```diff
@@ -80,14 +81,14 @@ public class SqlResource
   public static final String SQL_HEADER_VALUE = "yes";
 
   private static final Logger log = new Logger(SqlResource.class);
-  public static final SqlResourceQueryMetricCounter QUERY_METRIC_COUNTER = new SqlResourceQueryMetricCounter();
```
kfaraz, line 83: I suppose this was just a dummy instance that didn't even collect the stats?

server/src/main/java/org/apache/druid/server/metrics/BaseQueryCountResource.java
```diff
@@ -0,0 +1,85 @@
...
+public class BaseQueryCountResource implements QueryCountStatsProvider
```
kfaraz: Please rename this to something like `QueryCountStatsAccumulator` or `*Tracker` or `*ProviderImpl`.

(`Base*` implies that something extends this and `*Resource` is used for either REST resource classes or closeable resources).

server/src/main/java/org/apache/druid/server/metrics/QueryCountStatsProvider.java
```diff
@@ -40,4 +40,24 @@ public interface QueryCountStatsProvider
    * Returns the number of timed out queries during the emission period.
    */
   long getTimedOutQueryCount();
+
+  /**
+   * Increments successful query count.
+   */
+  void incrementSuccess();
```
kfaraz:
```suggestion
  void incrementSuccessful();
```

sql/src/main/java/org/apache/druid/sql/http/SqlResourceQueryResultPusherFactory.java
```diff
@@ -52,7 +53,7 @@ public SqlResourceQueryResultPusherFactory(
     this.selfNode = selfNode;
   }
 
-  public SqlResourceQueryResultPusher factorize(HttpServletRequest req, HttpStatement stmt, SqlQuery sqlQuery)
+  public SqlResourceQueryResultPusher factorize(final QueryCountStatsProvider counter, HttpServletRequest req, HttpStatement stmt, SqlQuery sqlQuery)
```
kfaraz: Please put args on separate lines

sql/src/main/java/org/apache/druid/sql/http/SqlHttpModule.java
kfaraz: We could however make a `Names.named("sql")` annotation binding here and use the named binding
in `SqlResource` which would allow separate tracking of sql and native query counts.

sql/src/main/java/org/apache/druid/sql/http/SqlResource.java
```diff
@@ -97,7 +98,8 @@ public SqlResource(
       final SqlEngineRegistry sqlEngineRegistry,
       final SqlResourceQueryResultPusherFactory resultPusherFactory,
       final DefaultQueryConfig defaultQueryConfig,
-      final ServerConfig serverConfig
+      final ServerConfig serverConfig,
+      final QueryCountStatsProvider counter
```
kfaraz, line 102: Inject the annotated impl here to keep SQL and native query stats separate.

server/src/test/java/org/apache/druid/server/metrics/QueryCountStatsMonitorTest.java
jtuglu1: No idea, I think to just add some variability so assertions were on some larger #s and so you don't need to do 4x calls.

sql/src/main/java/org/apache/druid/sql/http/SqlHttpModule.java
jtuglu1: I needed to add this binding to `SqlHttpModule` since quidem tests were failing.

sql/src/main/java/org/apache/druid/sql/http/SqlResource.java
jtuglu1: I don't want to keep them separate (ideally). Why do you want to keep them separate?

kfaraz: There could be some existing assumptions regarding the query counts in downstream applications. Also, it might be useful for future use cases to be able to distinguish between the two.

Say an application might want to determine number of failing Sql queries vs failing native queries.

sql/src/main/java/org/apache/druid/sql/http/SqlHttpModule.java
kfaraz: I see. That's probably because quidem test Guice setup doesn't use a CliBroker directly. 

In that case, let's move this binding to the appropriate place in the quidem test setup itself.

sql/src/main/java/org/apache/druid/sql/http/SqlResource.java
jtuglu1: That should be done with tags IMO and not differing metrics. What happened before this change is these 2 `*Resource` classes diverged to the point where there was gap in metrics between them. That success/failure can already be gauged from `sqlQuery/time`.

kfaraz: > That should be done with tags IMO and not differing metrics.

By tags, do you mean metric dimensions? That would be okay too [as suggested here](https://github.com/apache/druid/pull/18569#pullrequestreview-3265925315), as long as there is some way to distinguish between the two. But even different dimensions would still require you to collect the two stats separately via two different instances of the stats provider.

> What happened before this change is these 2 *Resource classes diverged to the point where there was gap in metrics between them.

Were SQL query count metrics ever emitted?

jtuglu1: > Were SQL query count metrics ever emitted?

At this granularity (timeout, failure, interrupt, etc.) no.

jtuglu1: I would prefer to just keep one shared instance. There hasn't been a use-case thus far where this was desired and adding more instances would make things more confusing IMO.

server/src/test/java/org/apache/druid/server/metrics/QueryCountStatsMonitorTest.java
jtuglu1: Sure, was trying to keep diff minimal.

sql/src/main/java/org/apache/druid/sql/http/SqlResource.java
kfaraz: Why would it be confusing? You just need to use an annotated binding.
Druid code uses it in several places and this is a perfectly valid use case.

We keep several instances of `ServiceClient`, `DruidLeaderSelector`, `ServiceLocator` and some classes with MSQ as well.
Check out the annotations `@IndexingService`, `@Coordinator`, `@Broker` which are used to do this.
There are also several places which use the `@Named` annotation.

jtuglu1: An alternative might be to keep a single class/instance, but introduce 1 layer of indirection to increment/fetch the metrics. This could support future combinations of (engine, queryFormat) where entries might look like (native, native) or (native, SQL) or (msq, SQL), etc. This way you don't get an explosion of unique metric names and/or annotated instances and that cardinality is instead pushed into the dimensions under a unified metric.

`QueryCountStatsMonitor` would know how to parse this map and create the relevant dimensions for the core set of metrics it emits.

I'd like to get to a point where people don't need to worry about a resource "picking up" a core metric because the code paths are separated. This might help with that.

jtuglu1: @gianm any thoughts here?

kfaraz: > This way you don't get an explosion of metrics and that cardinality is instead pushed into the dimensions under a unified metric.

Sure, as said earlier, same metric name with different dimension values works fine too.

kfaraz: > An alternative might be to keep a single class/instance, but introduce 1 layer of indirection to increment/fetch the metrics. This could support future combinations of (engine, queryFormat) where entries might look like (native, native) or (native, SQL) or (msq, SQL), etc. This way you don't get an explosion of unique metric names and/or annotated instances and that cardinality is instead pushed into the dimensions under a unified metric.

Yes, it would indeed be nice to accept some kind of `QueryType` in all the methods of `QueryCountStatsProvider`.

e.g

```java
void incrementSuccessful(QueryType type);

int getSuccessfulQueryCount(QueryType type);
AND/OR
Map<QueryType, Integer> getSuccessfulQueryCountsByType();

enum QueryType {NATIVE_JSON, NATIVE_SQL, MSQ, DART}
```

The query type will be emitted as a dimension.

But this will require more work as you would need to change the signature of `QueryCountStatsProvider` methods
and pass the query type down to the relevant classes like `QueryResultPusher`, etc.

If you feel like making that change, that would work for me too.

Multiple bindings achieves the same result for now (since we are only emitting SQL and native query metrics) but with less code change.

jtuglu1: 100% agree. Given that the behavior before this was `QueryResource` emitting query count metrics only (SQL was just doing no-op, because these classes didn't share any common class), do you think we can scope this PR to refactoring both to share a single metrics counter instance? Then, a follow-up can be adding the query execution type tag?

I think this change should address the core problem the GH issue mentions: that SQL counts were not being tracked in `query/*/count` metrics. Then, adding other ways of aggregating over these metrics with an extra dimension can be added in a follow-up task.

kfaraz: Agreed, we should address the core problem of publishing SQL metrics, but not at the cost of affecting existing metric values, with no way to tell whether a metric is originating from SQL or native.

The refactor can definitely be done in a follow up PR, but in that case, we should use the multiple bindings approach in this PR. The multiple bindings approach would have a very small footprint that can be easily removed later.

If we just a single instance with no dimensions in this PR,
I fear we might miss the Druid 35 code freeze (around Oct 7) and end up releasing a partial (possibly regressive) feature.

I would rather hold off on this PR until we have a complete solution, since we have never emitted SQL metrics anyway.

Let me know if you agree.

jtuglu1:
Closing as not needed at the moment

Issue #18573, opened 2025-09-25
NPE during result-level cache key construction


jtuglu1:
Please provide a detailed title (e.g. "Broker crashes when using TopN query with Bound filter" instead of just "Broker crashes").

### Affected Version

Druid Versions which allow null-handling support.

### Description

Example:
```json
{
  "dimensions": [
    {
      "type": "listFiltered",
      "delegate": "based_on",
      "values": [
        null,
        "A",
        "B"
      ]
    }
  ],
  "aggregations": [
    {
      "type": "doubleSum",
      "fieldName": "value",
      "name": "value"
    }
  ],
  "intervals": ["2025-01-01T00:00:00.000Z/2025-07-10T23:59:59.999Z"],
  "queryType": "groupBy",
  "granularity": "all",
  "dataSource": "datasource_A"
}
```

will fail with:
```
Error: RUNTIME_FAILURE (OPERATOR)

Cannot invoke "String.getBytes(String)" because "string" is null

java.lang.NullPointerException
```

PR #18575, opened 2025-09-25
Fix NPE in getCacheKey for ListFilteredDimensionSpec

jtuglu1:
Fixes https://github.com/apache/druid/issues/18573 and adds a test that reproduces the behavior.

### Description

Example:
```json
{
  "dimensions": [
    {
      "type": "listFiltered",
      "delegate": "based_on",
      "values": [
        null,
        "A",
        "B"
      ]
    }
  ],
  "aggregations": [
    {
      "type": "doubleSum",
      "fieldName": "value",
      "name": "value"
    }
  ],
  "intervals": ["2025-01-01T00:00:00.000Z/2025-07-10T23:59:59.999Z"],
  "queryType": "groupBy",
  "granularity": "all",
  "dataSource": "datasource_A"
}
```

will fail with:
```
Error: RUNTIME_FAILURE (OPERATOR)

Cannot invoke "String.getBytes(String)" because "string" is null

java.lang.NullPointerException
```

The line in question is [this](https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/query/dimension/ListFilteredDimensionSpec.java#L166).

#### Release note

Fix NPE in getCacheKey for ListFilteredDimensionSpec

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/query/dimension/ListFilteredDimensionSpec.java
+++ b/processing/src/main/java/org/apache/druid/query/dimension/ListFilteredDimensionSpec.java
@@ -22,14 +22,12 @@ package org.apache.druid.query.dimension;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicates;
-import org.apache.druid.java.util.common.StringUtils;
-import org.apache.druid.query.filter.DimFilterUtils;
+import org.apache.druid.query.cache.CacheKeyBuilder;
 import org.apache.druid.segment.DimensionSelector;
 import org.apache.druid.segment.IdLookup;
 import org.apache.druid.segment.IdMapping;
 
 import javax.annotation.Nullable;
-import java.nio.ByteBuffer;
 import java.util.Objects;
 import java.util.Set;
 
@@ -52,10 +50,10 @@ public class ListFilteredDimensionSpec extends BaseFilteredDimensionSpec
   {
     super(delegate);
 
-    Preconditions.checkArgument(values != null && values.size() > 0, "values list must be non-empty");
+    Preconditions.checkArgument(values != null && !values.isEmpty(), "values list must be non-empty");
     this.values = values;
 
-    this.isWhitelist = isWhitelist == null ? true : isWhitelist.booleanValue();
+    this.isWhitelist = isWhitelist == null || isWhitelist;
   }
 
   @JsonProperty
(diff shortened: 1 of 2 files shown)


processing/src/main/java/org/apache/druid/query/dimension/ListFilteredDimensionSpec.java
```diff
@@ -157,27 +155,11 @@ public static DimensionSelector filterDenyList(Set<String> values, DimensionSele
...
-                                          .put(DimFilterUtils.STRING_SEPARATOR);
-    for (byte[] bytes : valuesBytes) {
-      filterCacheKey.put(bytes)
-                    .put(DimFilterUtils.STRING_SEPARATOR);
-    }
-    return filterCacheKey.array();
+    CacheKeyBuilder builder = new CacheKeyBuilder(CACHE_TYPE_ID)
+        .appendBoolean(isWhitelist)
+        .appendCacheable(delegate)
+        .appendStringsIgnoringOrder(values);
```
jtuglu1, line 161: The order here shouldn't matter I believe since it's filtering in/out based on the items in the `values` set.

samarthjain: Thinking out loud here - consider a case where a distributed cache is being used for storing results cache that isn't restarted when a Druid cluster will be deployed with this change. Can it result in a cache miss now because the older key had values added in order and possibly had a different hash value vs key with this change?

jtuglu1: > Can it result in a cache miss now because the older key had values added in order and possibly had a different hash value vs key with this change?

Yes, this could happen and was considered. I think false negatives acceptable as long as false positives are not (at least extremely unlikely).

Generally this shouldn't be too big of an issue because:
- ≥ 1 old broker still serving from the old entry
- 1 new broker and ≥ 1 brokers reading from the new entry
- Eventually old entry gets TTL'd

jtuglu1: We should be moving away from manual key generation for all cacheable objects anyways (and switch to using builder) so this change will happen anyways.

samarthjain: Looks like the ListFilteredVirtualColumn [computes](https://github.com/apache/druid/blob/1b220b250426e6d1a6c8835040c8fcc0ec12ff8e/processing/src/main/java/org/apache/druid/segment/virtual/ListFilteredVirtualColumn.java#L132) key the same way.

+1 to the change.

samarthjain (approved):

PR #18576, opened 2025-09-26
Make KubernetesWorkItem.shutdown idempotent

kfaraz:
### Description

This patch tries to address the following condition due to a [bug in fabric8 client](https://github.com/fabric8io/kubernetes-client/issues/7163).

The callback executor in `TaskQueue` can get stuck in the following scenario.
- `TaskQueue` tries to shutdown a task via `notifyStatus()` and queues up a `TaskRunner.shutdown()` on the callback executor.
- `TaskQueue` then removes the entry for this task from its in-memory data-structures
- `TaskRunner.shutdown()` remains stuck due to the fabric8 bug
- `TaskRunner` still has the task in its memory
- When the `TaskQueue` periodically manages itself via `startPendingTasksOnRunner()`,
it realizes that the `TaskRunner` has this unknown task and tries to shut it down again.

### Fix

Subsequent calls to `KubernetesWorkItem.shutdown()` should not block and return immediately.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesWorkItem.java
+++ b/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesWorkItem.java
@@ -29,11 +29,14 @@ import org.apache.druid.indexing.overlord.TaskRunnerWorkItem;
 import org.apache.druid.java.util.common.ISE;
 
 import java.io.InputStream;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public class KubernetesWorkItem extends TaskRunnerWorkItem
 {
   private final Task task;
   private final KubernetesPeonLifecycle kubernetesPeonLifecycle;
+
+  private final AtomicBoolean isShuttingDown = new AtomicBoolean(false);
 
   public KubernetesWorkItem(Task task, ListenableFuture<TaskStatus> statusFuture, KubernetesPeonLifecycle kubernetesPeonLifecycle)
   {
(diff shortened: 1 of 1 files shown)


extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesWorkItem.java
```diff
@@ -29,23 +29,33 @@
...
+  /**
+   * Shuts down this work item. Subsequent calls to this method return immediately.
+   */
+  protected void shutdown()
   {
-    this.kubernetesPeonLifecycle.startWatchingLogs();
-    this.kubernetesPeonLifecycle.shutdown();
+    if (isShutdown.compareAndSet(false, true)) {
+      synchronized (this) {
+        this.kubernetesPeonLifecycle.startWatchingLogs();
```
capistrant, line 55: do we even need to be calling `startWatchingLogs()` here? Kind of off topic for the PR. But part of the reason shutdown is slow and can even hang, is because of the LogWatch being unhealthy. https://github.com/apache/druid/pull/18444 was implemented to put a time limit on log persist for saving off task logs. In that code path we are actually writing the logs out somewhere. I guess I don't see how the call here is useful at all if we are trying to just shut the task down?

capistrant (commented):
Seems like a logical solution to me. I think this might be a time to take a second and consider the log watch in the shutdown and if it is even necessary. what do you think?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesWorkItem.java
kfaraz: Yeah, I agree, the only apparent effect of calling `startWatchingLogs()` is to initialize `KubernetesPeonLifecycle.logWatch` field which would be initialized anyway via `join()` -> `finally` -> `saveLogs()` -> `doSaveLogs()`.

We can explore that in a separate PR.

capistrant:
https://github.com/apache/druid/pull/18579 opened this for proposing dropping the LogWatch init in shutdown

PR #18579, opened 2025-09-26
Remove LogWatch initialization during k8s task shutdown

capistrant:
### Description

Remove unnecessary LogWatch initialization from shutdown() in the K8s work item class. Under normal operations, the peon will already have initialized its LogWatch needed to stream logs. During shutdown it should either not need to create one because it exists, or skip creating it as we are just wanting to push through a shutdown at this point.

#18444 was created to try and mitigate known issues with LogWatch. This added a hacky timeout when trying to persist logs. Instead of doing that song and dance here, I propose we just drop the call to initialize it as the benefit of doing it is not clear to me. Removing it reduces risk of issues in shutdown leading to threads hung in fabric8 code like has been seen in live prod clusters.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesWorkItem.java
+++ b/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesWorkItem.java
@@ -44,7 +44,6 @@ public class KubernetesWorkItem extends TaskRunnerWorkItem
 
   protected synchronized void shutdown()
   {
-    this.kubernetesPeonLifecycle.startWatchingLogs();
     this.kubernetesPeonLifecycle.shutdown();
   }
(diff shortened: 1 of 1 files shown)


kfaraz:
Thanks for the PR, @capistrant !
I was in the middle of writing up a test to verify this 🙂 
Putting it up in #18580 .

capistrant:
> Thanks for the PR, @capistrant ! I was in the middle of writing up a test to verify this 🙂 Putting it up in #18580 .

Maybe we should consolidate the removal into your PR? Might make it easier for someone investigating the change at a later date to only end up at 1 PR with the change and the test for it

kfaraz:
Sure, @capistrant , I will update the test PR to include the fix.

PR #18587, opened 2025-09-30
Expand use of Druid side timeouts for fabric8 kubernetesclient timeouts

capistrant:
### Description

Expand on https://github.com/apache/druid/pull/18444 to cover more surface area. While putting saveLogs on a strict timer, helped prevent some issues with LogWatch interaction hanging, I have still seen issues in the path where `KubernetesWorkItem` calls shutdown which starts a LogWatch. Even the initialization of a LogWatch object can hang indefinitely.

#### Release note

new config for users of the `kubernetes-overlord-extensions` for K8s TaskRunner. `druid.indexer.runner.logWatchInitializationTimeout`. This puts a strict timeout on initializing a LogWatch object for a K8s task pod.

<hr>

 * `KubernetesPeonLifecycle`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/docs/development/extensions-core/k8s-jobs.md
+++ b/docs/development/extensions-core/k8s-jobs.md
@@ -793,7 +793,7 @@ Should you require the needed permissions for interacting across Kubernetes name
 | `druid.indexer.runner.capacity` | `Integer` | Number of concurrent jobs that can be sent to Kubernetes. | `2147483647` | No |
 | `druid.indexer.runner.cpuCoreInMicro` | `Integer` | Number of CPU micro core for the task. | `1000` | No |
 | `druid.indexer.runner.logSaveTimeout` | `Duration` | How long to wait for task logs to be saved before giving up. | `PT300S` | NO |
-| `druid.indexer.runner.logWatchInitializationTimeout` | `Duration` | How long to wait for Initializing a LogWatch on k8s peon pod before giving up. | `PT30S` | NO |
+| `druid.indexer.runner.logWatchInitializationTimeout` | `Duration` | How long to wait when initializing a log watch for a peon pod before giving up. | `PT30S` | NO |
 
 
 ### Metrics added
--- a/docs/development/extensions-core/k8s-jobs.md
+++ b/docs/development/extensions-core/k8s-jobs.md
@@ -792,8 +792,7 @@ Should you require the needed permissions for interacting across Kubernetes name
 | `druid.indexer.runner.graceTerminationPeriodSeconds` | `Long` | Number of seconds you want to wait after a sigterm for container lifecycle hooks to complete. Keep at a smaller value if you want tasks to hold locks for shorter periods. | `PT30S` (K8s default) | No |
 | `druid.indexer.runner.capacity` | `Integer` | Number of concurrent jobs that can be sent to Kubernetes. | `2147483647` | No |
 | `druid.indexer.runner.cpuCoreInMicro` | `Integer` | Number of CPU micro core for the task. | `1000` | No |
-| `druid.indexer.runner.logSaveTimeout` | `Duration` | How long to wait for task logs to be saved before giving up. | `PT300S` | NO |
-| `druid.indexer.runner.logWatchInitializationTimeout` | `Duration` | How long to wait when initializing a log watch for a peon pod before giving up. | `PT30S` | NO |
+| `druid.indexer.runner.podLogOperationTimeout` | `Duration` | Timeout for async operations that interact with `k8s` pod logs  | `PT300S` | NO |
 
 
 ### Metrics added
--- a/docs/development/extensions-core/k8s-jobs.md
+++ b/docs/development/extensions-core/k8s-jobs.md
@@ -792,6 +792,9 @@ Should you require the needed permissions for interacting across Kubernetes name
 | `druid.indexer.runner.graceTerminationPeriodSeconds` | `Long` | Number of seconds you want to wait after a sigterm for container lifecycle hooks to complete. Keep at a smaller value if you want tasks to hold locks for shorter periods. | `PT30S` (K8s default) | No |
 | `druid.indexer.runner.capacity` | `Integer` | Number of concurrent jobs that can be sent to Kubernetes. | `2147483647` | No |
 | `druid.indexer.runner.cpuCoreInMicro` | `Integer` | Number of CPU micro core for the task. | `1000` | No |
+| `druid.indexer.runner.logSaveTimeout` | `Duration` | How long to wait for task logs to be saved before giving up. | `PT300S` | NO |
+| `druid.indexer.runner.logWatchInitializationTimeout` | `Duration` | How long to wait for Initializing a LogWatch on k8s peon pod before giving up. | `PT30S` | NO |
+
 
 ### Metrics added
 
--- a/docs/development/extensions-core/k8s-jobs.md
+++ b/docs/development/extensions-core/k8s-jobs.md
@@ -792,7 +792,6 @@ Should you require the needed permissions for interacting across Kubernetes name
 | `druid.indexer.runner.graceTerminationPeriodSeconds` | `Long` | Number of seconds you want to wait after a sigterm for container lifecycle hooks to complete. Keep at a smaller value if you want tasks to hold locks for shorter periods. | `PT30S` (K8s default) | No |
 | `druid.indexer.runner.capacity` | `Integer` | Number of concurrent jobs that can be sent to Kubernetes. | `2147483647` | No |
 | `druid.indexer.runner.cpuCoreInMicro` | `Integer` | Number of CPU micro core for the task. | `1000` | No |
-| `druid.indexer.runner.logSaveTimeout` | `Duration` | How long to wait for task logs to be saved before giving up. | `PT300S` | NO |
 
 ### Metrics added
 
--- a/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
+++ b/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
@@ -34,7 +34,6 @@ import org.apache.druid.indexer.TaskLocation;
 import org.apache.druid.indexer.TaskStatus;
 import org.apache.druid.indexing.common.task.Task;
 import org.apache.druid.java.util.common.StringUtils;
-import org.apache.druid.java.util.common.concurrent.Execs;
 import org.apache.druid.java.util.emitter.EmittingLogger;
 import org.apache.druid.k8s.overlord.common.DruidK8sConstants;
 import org.apache.druid.k8s.overlord.common.JobResponse;
@@ -50,11 +49,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
@@ -97,7 +92,6 @@ public class KubernetesPeonLifecycle
   private final ObjectMapper mapper;
   private final TaskStateListener stateListener;
   private final SettableFuture<Boolean> taskStartedSuccessfullyFuture;
-  private final long logSaveTimeoutMs;
 
   @MonotonicNonNull
   private LogWatch logWatch;
@@ -110,8 +104,7 @@ public class KubernetesPeonLifecycle
       KubernetesPeonClient kubernetesClient,
       TaskLogs taskLogs,
       ObjectMapper mapper,
-      TaskStateListener stateListener,
-      long logSaveTimeoutMs
+      TaskStateListener stateListener
   )
   {
     this.task = task;
@@ -121,7 +114,6 @@ public class KubernetesPeonLifecycle
     this.mapper = mapper;
     this.stateListener = stateListener;
     this.taskStartedSuccessfullyFuture = SettableFuture.create();
-    this.logSaveTimeoutMs = logSaveTimeoutMs;
   }
 
   /**
@@ -355,33 +347,6 @@ public class KubernetesPeonLifecycle
   }
 
   protected void saveLogs()
-  {
-    ExecutorService executor = Executors.newSingleThreadExecutor(Execs.makeThreadFactory("k8s-tasklog-persist-%d"));
-    try {
-      Future<?> future = executor.submit(this::doSaveLogs);
-      future.get(logSaveTimeoutMs, TimeUnit.MILLISECONDS);
-    }
-    catch (TimeoutException e) {
-      log.warn("Persisting task logs timed out after %d ms for task [%s]. This does not have any impact on the"
-               + " work done by the task, but the logs may be innaccessible. If this continues to happen, check"
-               + " Kubernetes server logs for potential errors.", logSaveTimeoutMs, taskId.getOriginalTaskId());
-    }
-    catch (Exception e) {
-      log.error(e, "Persisting task logs failed for task [%s] This does not have any impact on the"
-                   + " work done by the task, but the logs may be innaccessible. If this continues to happen, check"
-                   + " Kubernetes server logs for potential errors.", taskId.getOriginalTaskId());
-    }
-    finally {
-      executor.shutdownNow();
-      // shutdownNow does not always allow finally blocks to run, so we make sure to close the logWatch here too if it
-      // wasn't closed in doSaveLogs
-      if (logWatch != null) {
-        logWatch.close();
-      }
-    }
-  }
-
-  private void doSaveLogs()
   {
     try {
       Path file = Files.createTempFile(taskId.getOriginalTaskId(), "log");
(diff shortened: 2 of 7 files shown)


extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycleTest.java
```diff
@@ -1050,74 +1082,6 @@ public void test_getTaskLocation_withStoppedTaskState_returnsUnknown()
     verifyAll();
   }
 
-  @Test
-  public void test_join_saveLogsTimeout_handledGracefully() throws IOException
```
kfaraz, line 1054: Wasn't this test verifying the save timeout change made in the previous PR?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -121,7 +124,11 @@ protected KubernetesPeonLifecycle(
     this.mapper = mapper;
     this.stateListener = stateListener;
     this.taskStartedSuccessfullyFuture = SettableFuture.create();
-    this.logSaveTimeoutMs = logSaveTimeoutMs;
+    this.logWatchInitializationTimeoutMs = logWatchInitializationTimeoutMs;
+    this.logWatchCopyLogTimeoutMs = logWatchCopyLogTimeoutMs;
+    this.logWatchExecutor = Executors.newSingleThreadExecutor(
```
kfaraz: I feel like the original approach of creating an `ExecutorService` on the fly when needed made more sense.
We don't really need this executor until we need to watch/copy over the logs, so there is no reason to have it take up unnecessary memory upfront.

docs/development/extensions-core/k8s-jobs.md
```diff
@@ -793,6 +793,8 @@ Should you require the needed permissions for interacting across Kubernetes name
 | `druid.indexer.runner.capacity` | `Integer` | Number of concurrent jobs that can be sent to Kubernetes. | `2147483647` | No |
 | `druid.indexer.runner.cpuCoreInMicro` | `Integer` | Number of CPU micro core for the task. | `1000` | No |
 | `druid.indexer.runner.logSaveTimeout` | `Duration` | How long to wait for task logs to be saved before giving up. | `PT300S` | NO |
+| `druid.indexer.runner.logWatchInitializationTimeout` | `Duration` | How long to wait when initializing a log watch for a peon pod before giving up. | `PT30S` | NO |
```
kfaraz: Let's avoid this new config and just use the original one for both the purposes.
A default timeout of 5 minutes is good enough for both initializing the `logWatch` field as well as
downloading the logs.

extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycleTest.java
capistrant: ya, but it only worked cuz the executor was also wrapping the log pusher which was mocked so I could artificially cause a timeout. The latest change tightens up what code is running in the executor so that approach didn't work anymore. I also think it was a mistake to have the code to push the logs in the executor in the first place. That is after the stream is copied off the pod so I don't think the logwatch is even in play at that point  so the timeout might prevent logs that would have successfully pushed from pushing

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
capistrant: ack. I don't like either approach 😨 but it makes sense to not eagerly consume memory if the log watch stuff is involved in a small chunk of the overall peon lifecycle

docs/development/extensions-core/k8s-jobs.md
capistrant: ok, I generalized the single config name a bit then

extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycleTest.java
capistrant: will look to see if there is other way to keep a similar test

docs/development/extensions-core/k8s-jobs.md
```diff
@@ -792,7 +792,8 @@ Should you require the needed permissions for interacting across Kubernetes name
 | `druid.indexer.runner.graceTerminationPeriodSeconds` | `Long` | Number of seconds you want to wait after a sigterm for container lifecycle hooks to complete. Keep at a smaller value if you want tasks to hold locks for shorter periods. | `PT30S` (K8s default) | No |
 | `druid.indexer.runner.capacity` | `Integer` | Number of concurrent jobs that can be sent to Kubernetes. | `2147483647` | No |
 | `druid.indexer.runner.cpuCoreInMicro` | `Integer` | Number of CPU micro core for the task. | `1000` | No |
-| `druid.indexer.runner.logSaveTimeout` | `Duration` | How long to wait for task logs to be saved before giving up. | `PT300S` | NO |
+| `druid.indexer.runner.podLogOperationTimeout` | `Duration` | Timeout for async operations that interact with `k8s` pod logs  | `PT300S` | NO |
```
kfaraz: `logOperationTimeout` seems more ambiguous than `logSaveTimeout`.
Do you prefer this name since the same timeout is used for multiple operations like copy and log watch initialization?

I think it is fine to continue calling the config `logSaveTimeout` since both are related to the saving of logs.
Although, please update the description to call out exactly which operations use this timeout and what is the result of that timeout (i.e. will logs for that task ever be accessible).

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerConfig.java
```diff
@@ -108,7 +108,7 @@ public class KubernetesTaskRunnerConfig
   @JsonProperty
   @NotNull
   // how long to wait for log saving operations to complete
-  private Period logSaveTimeout = new Period("PT300S");
+  private Period podLogOperationTimeout = new Period("PT300S");
```
kfaraz: If we continue with the same config name, we can just revert the changes made to this and some other files.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -343,52 +343,71 @@ protected void startWatchingLogs()
       log.debug("There is already a log watcher for %s", taskId.getOriginalTaskId());
       return;
     }
+    ExecutorService executor = Executors.newSingleThreadExecutor(Execs.makeThreadFactory("k8s-tasklog-init-watch-%d"));
```
kfaraz: The thread name should include the task id.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -343,52 +343,71 @@ protected void startWatchingLogs()
...
-    }
     catch (TimeoutException e) {
-      log.warn("Persisting task logs timed out after %d ms for task [%s]. This does not have any impact on the"
-               + " work done by the task, but the logs may be innaccessible. If this continues to happen, check"
-               + " Kubernetes server logs for potential errors.", logSaveTimeoutMs, taskId.getOriginalTaskId());
+      log.warn("Initializing log watcher timed out after %d ms for task [%s]. LogWatch not initialized.",
+               logWatchOperationTimeoutMs, taskId.getOriginalTaskId());
+    }
+    catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
kfaraz: Do we need to mark the thread as interrupted?
Is this status read by the `KubernetesTaskRunner`?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -343,52 +343,71 @@ protected void startWatchingLogs()
...
-      // shutdownNow does not always allow finally blocks to run, so we make sure to close the logWatch here too if it
-      // wasn't closed in doSaveLogs
-      if (logWatch != null) {
-        logWatch.close();
-      }
     }
   }
 
-  private void doSaveLogs()
+  protected void saveLogs()
```
kfaraz, line 373: I feel we might benefit from adding a common method which can be used both for logWatch init as well as saveLogs:

```java
private <T> T executeWithTimeout(Callable<T> runnable, long timeoutMillis, String operationName)
{
    // Create the executor
    // Start it
    // Handle the exceptions
    // Finally shutdown the executor
}
```

kfaraz (commented):
Thanks for addressing the feedback, @capistrant ! Left some final suggestions.

docs/development/extensions-core/k8s-jobs.md
capistrant: maybe I was over thinking it. I was concerned about doing the init under the same config as the save. But I suppose you are right that it is probably ok to just stick with it. I reverted and updated the doc with more info on what it is for.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
capistrant: hmmm. I guess this may just be force of habit to be defensive and do this whenever I catch an interrupted exception. Is there negatives to it even if the status isn't read?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
capistrant: good point, I took a shot at this

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerConfig.java
```diff
@@ -157,14 +157,14 @@ private KubernetesTaskRunnerConfig(
...
       Period k8sjobLaunchTimeout,
-      Period logSaveTimeout,
       List<String> peonMonitors,
       List<String> javaOptsArray,
       int cpuCoreInMicro,
       Map<String, String> labels,
       Map<String, String> annotations,
       Integer capacity,
-      Period taskJoinTimeout
+      Period taskJoinTimeout,
```
kfaraz: Please revert the changes to this file since they are not needed anymore.
We can stick to the original order of the constructor args.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -97,7 +99,7 @@ protected enum State
   private final ObjectMapper mapper;
   private final TaskStateListener stateListener;
   private final SettableFuture<Boolean> taskStartedSuccessfullyFuture;
-  private final long logSaveTimeoutMs;
+  private final long logSaveTimeout;
```
kfaraz: Nit: Please continue using `logSaveTimeoutMs` or `logSaveTimeoutMillis` as it removes any ambiguity regarding the time unit.

Alternatively, you may pass in a `Duration` object and use the name `logSaveTimeout`,
but I don't think that is needed here.

docs/development/extensions-core/k8s-jobs.md
```diff
@@ -792,7 +792,8 @@ Should you require the needed permissions for interacting across Kubernetes name
 | `druid.indexer.runner.graceTerminationPeriodSeconds` | `Long` | Number of seconds you want to wait after a sigterm for container lifecycle hooks to complete. Keep at a smaller value if you want tasks to hold locks for shorter periods. | `PT30S` (K8s default) | No |
 | `druid.indexer.runner.capacity` | `Integer` | Number of concurrent jobs that can be sent to Kubernetes. | `2147483647` | No |
 | `druid.indexer.runner.cpuCoreInMicro` | `Integer` | Number of CPU micro core for the task. | `1000` | No |
-| `druid.indexer.runner.logSaveTimeout` | `Duration` | How long to wait for task logs to be saved before giving up. | `PT300S` | NO |
+| `druid.indexer.runner.logSaveTimeout` | `Duration` | The peon executing the ingestion task makes a best effort to persist the pod logs from `k8s` to persistent task log storage. A timeout for this log persist is set to avoid `k8s` connection issues causing the peon to hang indefinitely if issues occur during the persisting of said logs. | `PT300S` | NO |
```
kfaraz:
```suggestion
| `druid.indexer.runner.logSaveTimeout` | `Duration` | The peon executing the ingestion task makes a best effort to persist the pod logs from `k8s` to persistent task log storage. The timeout ensures that `k8s` connection issues do not cause the pod to hang indefinitely thereby blocking Overlord operations. If the timeout occurs before the logs are saved, those logs will not be available in Druid. | `PT300S` | NO |
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -449,4 +449,35 @@ protected ListenableFuture<Boolean> getTaskStartedSuccessfullyFuture()
...
   }
+
+  /**
+   * Executes a callable with a timeout.
+   * <p>
+   * If the callable does not complete within the specified timeout or another exception occurs, the error will be
+   * logged and null will be returned.
+   * </p>
+   */
+  private <T> @Nullable T executeWithTimeout(Callable<T> runnable, long timeoutMillis, String operationName, String errorMessage)
```
kfaraz:
```suggestion
  private <T> @Nullable T executeWithTimeout(Callable<T> callable, long timeoutMillis, String operationName, String errorMessage)
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -449,4 +449,35 @@ protected ListenableFuture<Boolean> getTaskStartedSuccessfullyFuture()
...
+  {
+    ExecutorService executor = Executors.newSingleThreadExecutor(
+        Execs.makeThreadFactory("k8s-peon-lifecycle-util-" + taskId.getOriginalTaskId() + "-%d"));
+    try {
+      Future<T> future = executor.submit(runnable);
+      return future.get(timeoutMillis, TimeUnit.MILLISECONDS);
+    }
+    catch (TimeoutException e) {
+      log.warn("Operation [%s] timed out after %d ms for task [%s]. %s", operationName, timeoutMillis, taskId.getOriginalTaskId(), errorMessage);
```
kfaraz:
```suggestion
      log.warn("Operation[%s] for task[%s] timed out after [%d] ms with error[%s].", operationName, taskId.getOriginalTaskId(), timeoutMillis, errorMessage);
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -449,4 +449,35 @@ protected ListenableFuture<Boolean> getTaskStartedSuccessfullyFuture()
...
+    }
+    catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
kfaraz: @capistrant , regarding the previous discussion, I think we should avoid marking this thread as interrupted here since we are not reading/clearing this interrupted status anywhere which might have unintended side effects.

We can add it in the future if we need it.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -449,4 +449,35 @@ protected ListenableFuture<Boolean> getTaskStartedSuccessfullyFuture()
...
+      log.warn("Operation [%s] was interrupted for task [%s]. %s", operationName, taskId.getOriginalTaskId(), errorMessage);
```
kfaraz:
```suggestion
      log.warn("Operation[%s] for task[%s] was interrupted with error[%s].", operationName, taskId.getOriginalTaskId(), errorMessage);
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesPeonLifecycle.java
```diff
@@ -449,4 +449,35 @@ protected ListenableFuture<Boolean> getTaskStartedSuccessfullyFuture()
...
+    }
+    catch (Exception e) {
+      log.error(e, "Error during operation [%s] for task: %s. %s", operationName, taskId, errorMessage);
```
kfaraz:
```suggestion
      log.error(e, "Error during operation[%s] for task[%s]: %s", operationName, taskId, errorMessage);
```

PR #18589, opened 2025-09-30
derive-json

cecemei:
### Description

Derive json base on index.

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.


processing/src/main/java/org/apache/druid/segment/column/ColumnConfig.java
```diff
@@ -74,4 +82,9 @@ default double skipValueRangeIndexScale()
   {
     return DEFAULT_SKIP_VALUE_RANGE_INDEX_SCALE;
   }
+
+  default boolean deriveJsonColumnFromIndexes()
```
clintropolis: in my opinion i don't think we should add this since its just like a system level thing to control what is a segment level thing, though if we do leave it we should add javadoc to indicate that this is like a developer testing control knob, not something an operator would ever want to set (in case anyone stumbled on it in the code or a future developer wonders why it doesn't have docs or whatever)

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -156,7 +160,12 @@ public CompressedNestedDataComplexColumn(
     this.columnName = columnName;
     this.logicalType = logicalType;
     this.nullValues = nullValues;
-    this.fieldsSupplier = fieldsSupplier;
+    final TKeyDictionary fields = fieldsSupplier.get();
+    this.fieldPathMap = CollectionUtils.newLinkedHashMapWithExpectedSize(fields.size());
+    for (int i = 0; i < fields.size(); i++) {
+      String field = StringUtils.fromUtf8(fields.get(i));
+      fieldPathMap.put(parsePath(field), Pair.of(field, i));
+    }
```
clintropolis: i kind of think this should be lazy computed, really complicated json objects can have thousands of paths, and we really only need to know stuff about all of them if the query is doing something like json_keys/json_paths (or this new thing of reconstructing the object from the list of paths). If a query just has a json_value for a single field of the object, then it seems like it would be penalized by having to build this map for a bunch of paths it doesn't care about.

If changed to be lazy computed, the column selector methods for fields (and any other methods that take a list of path parts like `getFieldLogicalType`) should not use this map unless it was computed, since those are the things called when we want to make a selector for a specific path directly by things like NestedFieldVirtualColumn and other nested column specialized virtual columns.

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -329,28 +330,54 @@ public Object getRowValue(int rowNum)
...
+
+    ReadableOffset offset = new AtomicIntegerReadableOffset(new AtomicInteger(rowNum));
+    final List<StructuredDataBuilder.Element> elements =
+        fieldPathMap.keySet().stream()
+                    .map(path -> StructuredDataBuilder.Element.of(
+                        path,
+                        (Objects.requireNonNull(getColumnHolder(path)).getColumn()).makeColumnValueSelector(offset)
+                                                                                   .getObject()
+                    ))
+                    .collect(Collectors.toList());
```
clintropolis: making new offset and column value selector for every row seems pretty unchill... while I don't think much should be calling this method since we implement `makeColumnValueSelector`, could you look further into whether or not something will call this? If something does, it would be good to know to either consider changing it or knowing if it would be ok. or if nothing is expected to call this maybe we throw an exception

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -382,17 +418,32 @@ public void inspectRuntimeShape(RuntimeShapeInspector inspector)
...
     }
-    if (compressedRawColumn == null) {
+
+    if (compressedRawColumn == null && !columnConfig.deriveJsonColumnFromIndexes()) {
       compressedRawColumn = closer.register(compressedRawColumnSupplier.get());
     }
+    AtomicInteger rowNumber = new AtomicInteger(-1);
+    AtomicIntegerReadableOffset atomicOffset = new AtomicIntegerReadableOffset(rowNumber);
+    final List<Pair<List<NestedPathPart>, ColumnValueSelector>> fieldSelectors =
+        compressedRawColumn != null ? null :
```
clintropolis: maybe we should just make the `compressedRawColumnSupplier` nullable instead of checking the thing from the supplier is null?

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -434,8 +485,17 @@ private Object getForOffset(int offset)
...
+          final ByteBuffer valueBuffer = compressedRawColumn.get(offset);
+          return STRATEGY.fromByteBuffer(valueBuffer, valueBuffer.remaining());
+        } else {
+          rowNumber.set(offset);
+          List<StructuredDataBuilder.Element> elements = fieldSelectors
+              .stream()
+              .map(c -> StructuredDataBuilder.Element.of(c.lhs, c.rhs.getObject()))
+              .collect(Collectors.toList());
+          return new StructuredDataBuilder(elements).build();
+        }
```
clintropolis: same as non-vectorized, i think this would be cleaner to have separate `VectorObjectSelector` implementations returned depending on object encoding rather than checking inside every vector

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -361,8 +388,17 @@ public Object getObject()
...
+          final ByteBuffer valueBuffer = compressedRawColumn.get(offset.getOffset());
+          return STRATEGY.fromByteBuffer(valueBuffer, valueBuffer.remaining());
+        }
+        List<StructuredDataBuilder.Element> elements =
+            Objects.requireNonNull(fieldSelectors)
+                   .stream()
+                   .map(c -> StructuredDataBuilder.Element.of(c.lhs, c.rhs.getObject()))
+                   .collect(Collectors.toList());
+        return new StructuredDataBuilder(elements).build();
+
```
clintropolis: this might be a bit cleaner to just make separate `ObjectColumnSelector` implementations depending on object encoding (as we add more object encodings in the future).

processing/src/main/java/org/apache/druid/segment/data/CompressedVariableSizedBlobColumnSerializer.java
```diff
@@ -51,13 +54,15 @@ public class CompressedVariableSizedBlobColumnSerializer implements Serializer
...
       final SegmentWriteOutMedium segmentWriteOutMedium,
+      final ObjectStorageEncoding objectStorageEncoding,
       final CompressionStrategy compression
   )
   {
     this.filenameBase = filenameBase;
     this.offsetsFile = getCompressedOffsetsFileName(filenameBase);
     this.blobsFile = getCompressedBlobsFileName(filenameBase);
     this.segmentWriteOutMedium = segmentWriteOutMedium;
+    this.objectStorageEncoding = objectStorageEncoding;
```
clintropolis: this should not be passed into `CompressedVariableSizedBlobColumnSerializer`, since it is a json column specific thing. I think all of the changes to this and `CompressedVariableSizedBlobColumnSupplier` need to be reverted, and should be handled entirely within nested data column layer, which should not create this serializer (or store the internal 'raw' file that results from it) at all. The `objectStorageEncoding` can be passed into the column supplier from https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/serde/NestedCommonFormatColumnPartSerde.java#L343, which the read method can use to know whether or not to bother with the `CompressedVariableSizedBlobColumnSupplier`.

processing/src/main/java/org/apache/druid/segment/column/ColumnConfig.java
cecemei: reverted this

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
cecemei: true, add several utility method to get the field info instead

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
cecemei: updated to check null for `compressedRawColumnSupplier`

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
cecemei: updated to use a `valueProvider` so `compressedRawColumn` and `fieldSelectors` is now closer.

processing/src/main/java/org/apache/druid/segment/data/CompressedVariableSizedBlobColumnSerializer.java
cecemei: updated to use nullable `CompressedVariableSizedBlobColumnSupplier`

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -329,29 +323,69 @@ public Object getRowValue(int rowNum)
...
+    }
+
+    final List<StructuredDataBuilder.Element> elements = getAllParsedBaseFields()
+        .stream()
+        .map(pair -> {
+          NestedFieldDictionaryEncodedColumn column = (NestedFieldDictionaryEncodedColumn) getColumnHolder(
+              pair.lhs.fieldName,
+              pair.lhs.fieldIndex
+          ).getColumn();
+          return StructuredDataBuilder.Element.of(pair.rhs, column.lookupObject(rowNum));
```
clintropolis: this is not correct.i think, it looks like `lookupObject` implementation takes a dictionaryId, but you're passing a row number in here.

To get the dictionaryId of the `NestedFieldDictionaryEncodedColumn`, you need to get the value at `rowNum` of `NestedFieldDictionaryEncodedColumn.column`, so you want to be doing `column.lookupObject(column.getSingleRowValue(rowNum))` i think.

Did you ever figure out if this method gets called for real? It must not commonly at least or else this would have caused problems i think

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -382,17 +415,48 @@ public void inspectRuntimeShape(RuntimeShapeInspector inspector)
...
+                           pair.lhs.fieldIndex
+                       ).getColumn()).makeColumnValueSelector(readableAtomicOffset)
+                   ))
+                   .collect(Collectors.toList());
+      valueProvider = () -> {
+        List<StructuredDataBuilder.Element> elements = fieldSelectors
+            .stream()
+            .map(c -> StructuredDataBuilder.Element.of(c.lhs, c.rhs.getObject()))
+            .collect(Collectors.toList());
+        return new StructuredDataBuilder(elements).build();
```
clintropolis, line 457: doesn't have to be done in this PR, but my experience has been that java streams is less performant than normal loops in hot code-paths, might be worth measuring this in the future (I don't think we need to worry about this in this PR)

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -1206,4 +1245,102 @@ public int compare(Object o1, Object o2)
...
+    private final int elementNumber;
+
+    ArrayField(BaseField baseField, int elementNumber)
+    {
+      this.baseField = baseField;
+      this.elementNumber = elementNumber;
+    }
+  }
+
+  private static class AtomicIntegerReadableOffset implements ReadableOffset
```
clintropolis: this seems the same as https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/query/rowsandcols/concrete/ColumnHolderRACColumn.java#L165, maybe should move somewhere and share them?

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -1206,4 +1245,102 @@ public int compare(Object o1, Object o2)
...
+      }
+    }
+    return null;
+  }
+
+  interface Field
+  {
+  }
+
+  private static class BaseField implements Field
```
clintropolis: if i understand correctly, this stuff is used to indicate that a nested field column exists, and the other one a  nested array field column exists to try to extract a specific array element from, maybe clearer names about what they are like `NestedField` and `NestedArrayField` would be helpful, or at least some javadocs. I find the 'base' terminology a bit unintuitive, so would suggest dropping it, since what it really means is that a nested field column exists at some path

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -1206,4 +1245,102 @@ public int compare(Object o1, Object o2)
...
+    List<Pair<BaseField, List<NestedPathPart>>> allFields = new ArrayList<>(fields.size());
+    for (int i = 0; i < fields.size(); i++) {
+      String field = StringUtils.fromUtf8(fields.get(i));
+      allFields.add(Pair.of(new BaseField(field, i), parsePath(field)));
+    }
+    return allFields;
+  }
+
+  @Nullable
+  private Field getBaseOrArrayFieldFromPath(List<NestedPathPart> path)
```
clintropolis: this feels worth having javadocs on intended usage, since i think this returns something if there is a nested field column present for the path (or nested field array column to try to read to extract an element from)

clintropolis:
It would be nice to get some additional test coverage on this, could you consider adding tests in this mode to [NestedDataScanQueryTest.java ](https://github.com/apache/druid/blob/master/processing/src/test/java/org/apache/druid/query/scan/NestedDataScanQueryTest.java) or [CalciteNestedDataQueryTest.java](https://github.com/apache/druid/blob/master/sql/src/test/java/org/apache/druid/sql/calcite/CalciteNestedDataQueryTest.java). Ideally replicas of existing tests that are relying on the raw column, so we can compare the results between the 2 modes

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
cecemei: thanks for the catch! seems like it's not covered by test, added test coverage in `NestedDataColumnSupplierTest`

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
cecemei: updated to be a shared public class

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
cecemei: added javadoc for all the `NestedField` and `NestedArrayElement`

cecemei:
> It would be nice to get some additional test coverage on this, could you consider adding tests in this mode to [NestedDataScanQueryTest.java ](https://github.com/apache/druid/blob/master/processing/src/test/java/org/apache/druid/query/scan/NestedDataScanQueryTest.java) or [CalciteNestedDataQueryTest.java](https://github.com/apache/druid/blob/master/sql/src/test/java/org/apache/druid/sql/calcite/CalciteNestedDataQueryTest.java). Ideally replicas of existing tests that are relying on the raw column, so we can compare the results between the 2 modes

Added test coverage for both classes.

sql/src/test/java/org/apache/druid/sql/calcite/CalciteNestedDataQueryTest.java
```diff
@@ -150,15 +152,18 @@ public class CalciteNestedDataQueryTest extends BaseCalciteQueryTest
...
-                       .add(AutoTypeColumnSchema.of("string"))
-                       .add(AutoTypeColumnSchema.of("nest"))
-                       .add(AutoTypeColumnSchema.of("nester"))
-                       .add(AutoTypeColumnSchema.of("long"))
-                       .add(AutoTypeColumnSchema.of("string_sparse"))
+                       .add(new AutoTypeColumnSchema("string", null, NONE_OBJECT_STORAGE))
+                       .add(new AutoTypeColumnSchema("nest", null, NONE_OBJECT_STORAGE))
+                       .add(new AutoTypeColumnSchema("nester", null, NONE_OBJECT_STORAGE))
+                       .add(new AutoTypeColumnSchema("long", null, NONE_OBJECT_STORAGE))
+                       .add(new AutoTypeColumnSchema("string_sparse", null, NONE_OBJECT_STORAGE))
```
clintropolis: hmm, i don't think we should replace the existing tests with the new mode, could you add this as a separate datasource and just dupe the tests (or parameterize on table name or something) so we have coverage for both modes?

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -329,29 +324,69 @@ public Object getRowValue(int rowNum)
...
+                                        ).getColumn()).makeColumnValueSelector(offset)
+                                    ))
+                                    .collect(Collectors.toList());
+      valueProvider = () -> {
+        List<StructuredDataBuilder.Element> elements = fieldSelectors
+            .stream()
+            .map(c -> StructuredDataBuilder.Element.of(c.lhs, c.rhs.getObject()))
+            .collect(Collectors.toList());
+        return new StructuredDataBuilder(elements).build();
+      };
```
clintropolis, line 388: been thinking a bit on this, and I don't think we need to change anything in this PR since this is sufficient for testing and experimentation, but I think this could probably be done a lot more efficiently and just build the object on the fly while iterating over the set of fields only once per row produced and without extra transient object instantiations that need collected (the builder, elements, etc).

The field positions in fieldsSupplier are the same as the internal column names we get from getColumnHolder, so it would map pretty well to a plain for loop that just looks up the List<NestedPathPart> for the field position and use that to add the field value to the object. It would be more or less inlining the buildObject method of the builder i think, so like less pleasant to look at but also a fair bit less overhead I would think.

sql/src/test/java/org/apache/druid/sql/calcite/CalciteNestedDataQueryTest.java
cecemei: added as nested test, it's a bit tricky to use parameterized test since the component supplier thing has to be static.

cryptoe:
@cecemei 
Could you please update the description with the release notes. 
How to use this feature is missing in the release notes.

PR #18591, opened 2025-10-01
Implement dynamic capacity for kubernetes task runner

GabrielCWT:
### Description

The aim of the PR is to enable changing `capacity` for the `KubernetesTaskRunner`. This would be done through the existing POST API `/druid/indexer/v1/k8s/taskrunner/executionconfig`.

#### K8 configuration changes
In order to do this, I have added a new interface `KubernetesTaskRunnerConfig` and renamed the existing config to `KubernetesTaskRunnerStaticConfig`. The interface will be implemented by the existing static config and a new `KubernetesTaskRunnerEffectiveConfig` which will be a wrapper class to encapsulate both the dynamic and static configs. 

The effective config will fall back to the static config's capacity if the dynamic config has not been set. 

#### Changes to `/druid/indexer/v1/k8s/taskrunner/executionconfig` behaviour

The API will now take a new `capacity` field. On top of this, the fields will now be optional. If any field is `null` or not passed, we will use the existing dynamic config values. 

#### Release note

New `capacity` field for `/druid/indexer/v1/k8s/taskrunner/executionconfig` POST API. It will change the capacity for `KubernetesTaskRunner`.

### Challenges

In order to update the capacity for the task runner, I am calling a new function `syncCapacityWithDynamicConfig` before every task is run in order to update the thread pool to the newest config. 

The issue with this is that any changes by the user will not be immediately reflected on the web console's homepage under the "Tasks" widget. The "task slots" would only be updated after a new task has been run. 

I could not find a way to add a callback to the updating of dynamic configurations and felt that having a check every few seconds to see if the dynamic configuration had been updated was unnecessarily complex. I am open to suggestions if there are better ways to update the task runner.

Screenshot 2025-10-01 at 5 47 05 PM

### DIFF
--- a/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
+++ b/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
@@ -83,6 +83,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 /**
@@ -124,7 +125,7 @@ public class KubernetesTaskRunner implements TaskLogStreamer, TaskRunner
   // currently worker categories aren't supported, so it's hardcoded.
   protected static final String WORKER_CATEGORY = "_k8s_worker_category";
 
-  private int currentCapacity;
+  private final AtomicInteger currentCapacity;
 
   public KubernetesTaskRunner(
       TaskAdapter adapter,
@@ -144,8 +145,8 @@ public class KubernetesTaskRunner implements TaskLogStreamer, TaskRunner
     this.cleanupExecutor = Executors.newScheduledThreadPool(1);
     this.emitter = emitter;
 
-    this.currentCapacity = config.getCapacity();
-    this.tpe = new ThreadPoolExecutor(currentCapacity, currentCapacity, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(), Execs.makeThreadFactory("k8s-task-runner-%d", null));
+    this.currentCapacity = new AtomicInteger(config.getCapacity());
+    this.tpe = new ThreadPoolExecutor(currentCapacity.get(), currentCapacity.get(), 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(), Execs.makeThreadFactory("k8s-task-runner-%d", null));
     this.exec = MoreExecutors.listeningDecorator(this.tpe);
     configManager.addListener(KubernetesTaskRunnerDynamicConfig.CONFIG_KEY, StringUtils.format(OBSERVER_KEY, Thread.currentThread().getId()), this::syncCapacityWithDynamicConfig);
   }
@@ -194,19 +195,19 @@ public class KubernetesTaskRunner implements TaskLogStreamer, TaskRunner
   private void syncCapacityWithDynamicConfig(KubernetesTaskRunnerDynamicConfig config)
   {
     int newCapacity = config.getCapacity();
-    if (newCapacity == currentCapacity) {
+    if (newCapacity == currentCapacity.get()) {
       return;
     }
-    log.info("Adjusting k8s task runner capacity from [%d] to [%d]", currentCapacity, newCapacity);
+    log.info("Adjusting k8s task runner capacity from [%d] to [%d]", currentCapacity.get(), newCapacity);
     // maximum pool size must always be greater than or equal to the core pool size
-    if (newCapacity < currentCapacity) {
+    if (newCapacity < currentCapacity.get()) {
       tpe.setCorePoolSize(newCapacity);
       tpe.setMaximumPoolSize(newCapacity);
     } else {
       tpe.setMaximumPoolSize(newCapacity);
       tpe.setCorePoolSize(newCapacity);
     }
-    currentCapacity = newCapacity;
+    currentCapacity.set(newCapacity);
   }
 
   private TaskStatus runTask(Task task)
@@ -450,7 +451,7 @@ public class KubernetesTaskRunner implements TaskLogStreamer, TaskRunner
   @Override
   public Map<String, Long> getTotalTaskSlotCount()
   {
-    return ImmutableMap.of(WORKER_CATEGORY, (long) currentCapacity);
+    return ImmutableMap.of(WORKER_CATEGORY, (long) currentCapacity.get());
   }
 
   @Override
@@ -468,13 +469,13 @@ public class KubernetesTaskRunner implements TaskLogStreamer, TaskRunner
   @Override
   public Map<String, Long> getIdleTaskSlotCount()
   {
-    return ImmutableMap.of(WORKER_CATEGORY, (long) Math.max(0, currentCapacity - tasks.size()));
+    return ImmutableMap.of(WORKER_CATEGORY, (long) Math.max(0, currentCapacity.get() - tasks.size()));
   }
 
   @Override
   public Map<String, Long> getUsedTaskSlotCount()
   {
-    return ImmutableMap.of(WORKER_CATEGORY, (long) Math.min(currentCapacity, tasks.size()));
+    return ImmutableMap.of(WORKER_CATEGORY, (long) Math.min(currentCapacity.get(), tasks.size()));
   }
 
   @Override
@@ -565,7 +566,7 @@ public class KubernetesTaskRunner implements TaskLogStreamer, TaskRunner
   @Override
   public int getTotalCapacity()
   {
-    return currentCapacity;
+    return currentCapacity.get();
   }
 
   @Override


extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
```diff
@@ -109,17 +111,20 @@ public class KubernetesTaskRunner implements TaskLogStreamer, TaskRunner
   protected final TaskAdapter adapter;
 
   private final KubernetesPeonClient client;
-  private final KubernetesTaskRunnerConfig config;
+  private final KubernetesTaskRunnerEffectiveConfig config;
```
FrankChen021: I think we can still keep using the interface declaration

extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/overlord/taskadapter/K8sTaskAdapterTest.java
```diff
@@ -138,12 +138,12 @@ public PodSpec getSpec()
       }
     };
 
-    KubernetesTaskRunnerConfig config = KubernetesTaskRunnerConfig.builder()
-        .withNamespace("test")
-        .withOverlordNamespace("test_different")
-        .withAnnotations(ImmutableMap.of("annotation_key", "annotation_value"))
-        .withLabels(ImmutableMap.of("label_key", "label_value"))
-        .build();
+    KubernetesTaskRunnerStaticConfig config = KubernetesTaskRunnerStaticConfig.builder()
```
FrankChen021: I think we can still keep the builder method in the interface, so that there will not be any changes in the test cases

docs/development/extensions-core/k8s-jobs.md
```diff
@@ -48,9 +48,9 @@ Other configurations required are:
 Druid operators can dynamically tune certain features within this extension. You don't need to restart the Overlord
 service for these changes to take effect.
 
-Druid can dynamically tune [pod template selection](#pod-template-selection), which allows you to configure the pod 
-template based on the task to be run. To enable dynamic pod template selection, first configure the 
-[custom template pod adapter](#custom-template-pod-adapter).
+Druid can dynamically tune [pod template selection](#pod-template-selection) and [capacity](#properties).
```
GWphua: Would be clearer, IMO, to add a sentence to state 'capacity' refers to 'the maximum number of pods that can run at any time, or u can write it as `druid.indexer.runner.capacity`.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerStaticConfig.java
```diff
@@ -0,0 +1,562 @@
...
+import org.joda.time.Period;
+
+import javax.annotation.Nonnull;
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import java.util.List;
+import java.util.Map;
+
+public class KubernetesTaskRunnerStaticConfig implements KubernetesTaskRunnerConfig
```
GWphua, line 40: Would be great to have Javadocs to say how this config is used.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerEffectiveConfig.java
```diff
@@ -0,0 +1,185 @@
...
+import org.apache.druid.k8s.overlord.execution.KubernetesTaskRunnerDynamicConfig;
+import org.apache.druid.k8s.overlord.execution.PodTemplateSelectStrategy;
+import org.joda.time.Period;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Effective config object that combines static {@link KubernetesTaskRunnerConfig}
+ * with dynamic overrides from {@link KubernetesTaskRunnerDynamicConfig}.
```
GWphua:
```suggestion
 * Provides a flexible mechanism to configure Kubernetes task pods, 
 * by merging the static base settings from {@link KubernetesTaskRunnerConfig} 
 * with dynamic overrides from {@link KubernetesTaskRunnerDynamicConfig}.
 * <p>
 * Kubernetes will always use this effective config to run new tasks.
```

GWphua (approved):
LGTM, can remove the deprecated `defaultIfNull` calls.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerStaticConfig.java
GabrielCWT: Added the docs below.

```
/**
 * Static configuration for the Kubernetes-based task runner used by the Overlord.
 * <p>
 * This configuration is populated from runtime properties with the prefix
 * {@code druid.indexer.runner}.
 */
public class KubernetesTaskRunnerStaticConfig implements KubernetesTaskRunnerConfig
 ```

Comment (GWphua) [reply extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerStaticConfig.java:40]: Thanks for the change!

> Static configuration for the Kubernetes-based task runner used by the Overlord.

I think this is redundant -- already self-explanatory in the class name. Code readers who come to this class are probably more interested in how and when this class is used, instead of what this class is. 
e.g. Used as the starting config, and then will be gradually updated by dynamic configs if applicable.

Capturing this idea is probably more crucial.

**Comment by kfaraz** [comment]
@GabrielCWT , thanks for putting this together.

I might be mistaken but I am not entirely sure if the task capacity of a cluster should be a dynamic config.
For any cluster admin, this can have large cost implications.
It seems odd to be able to override the capacity of the cluster defined in the Overlord runtime property
simply by calling an API.

Changing the task capacity should not be a very frequent requirement for any cluster.
And when needed, it should be fairly reasonable to require an Overlord restart.

Could you please elaborate on why you feel the current setup is not adequate?

**Comment by GabrielCWT** [comment]
> @GabrielCWT , thanks for putting this together.
> 
> I might be mistaken but I am not entirely sure if the task capacity of a cluster should be a dynamic config. For any cluster admin, this can have large cost implications. It seems odd to be able to override the capacity of the cluster defined in the Overlord runtime property simply by calling an API.
> 
> Changing the task capacity should not be a very frequent requirement for any cluster. And when needed, it should be fairly reasonable to require an Overlord restart.
> 
> Could you please elaborate on why you feel the current setup is not adequate?

Restarting the Overlord can be time-consuming and potentially risky as we could face issues when trying to redeploy the Overlord instance. Updating the task capacity dynamically allows admins to adjust the cluster safely, reducing operational downtime and complexity. While changes to task capacity are infrequent, I feel that providing a safer runtime option would help to minimize disruptions.

**Comment by FrankChen021** [comment]
> @GabrielCWT , thanks for putting this together.
> 
> I might be mistaken but I am not entirely sure if the task capacity of a cluster should be a dynamic config. For any cluster admin, this can have large cost implications. It seems odd to be able to override the capacity of the cluster defined in the Overlord runtime property simply by calling an API.
> 
> Changing the task capacity should not be a very frequent requirement for any cluster. And when needed, it should be fairly reasonable to require an Overlord restart.
> 
> Could you please elaborate on why you feel the current setup is not adequate?

I think it's true that changing the task slot for middle manager requires the restart as we might need a redeployment of middelmanager to some servers with bigger resources. but for K8S-based task scheduling, the resources is allocated at K8S side which is out of druid, restarting of the overlord does not make any sense, we should have the ablity to reload the capacity dynamically. As @GabrielCWT has stated above, restarting overlord is a heavy and risky operation in production.

**Comment by kfaraz** [comment]
Thanks for the responses, @FrankChen021 , @GabrielCWT .

Even though with the K8s task runner, the task pods are not technically a part of the Druid cluster, the Overlord still has to manage those tasks. The KubernetesTaskRunner itself keeps separate threads for each running task to track their status (this PR also updates that thread count, if I am not mistaken).

So I should imagine that a major change in task capacity would also require some kind of scaling of the Overlord itself.

Also, why is the Overlord restart "risky" or even "slow"? Doesn't a version upgrade require an Overlord restart too?

**Comment by FrankChen021** [comment]
> too?

In a k8s deployment, after increasing the task capability, increasing the overlord resources may not be always needed. For example, we generally set the cpu limit to a higher value while keeping the cpu request relatively low at the initial deployment, when capacity is increased, no need to increase the cpu resource.  I mean risky because overlord needs to restore all tasks, previously we had some problems (maybe bug) that after switching leaders, overlord failed to elect a new leader. We try our best not to restart coordinator/overlord in production.

**Comment by kfaraz** [comment]
> I mean risky because overlord needs to restore all tasks, previously we had some problems (maybe bug) that after switching leaders, overlord failed to elect a new leader.

Yes, there might be some bugs around that. Also, the K8s task runner makes certain list pod calls, which are pretty heavy
and needs to be addressed. I think @capistrant is doing some work to improve that code flow.

> We try our best not to restart coordinator/overlord in production.

Oh, how frequently do you upgrade your cluster?
Is changing the task capacity going to be much more frequent than that.

I agree that K8s task runner is buggy and we should improve upon it.
But making the task capacity dynamic doesn't seem like the best solution.
It will open a whole another can of worms and make this piece only more complicated.

Instead, we should trying to fix up the actual problems in the task runner which make Overlord leader switch erroneous.

What are your thoughts, @FrankChen021 ?

**Comment by FrankChen021** [comment]
> > I mean risky because overlord needs to restore all tasks, previously we had some problems (maybe bug) that after switching leaders, overlord failed to elect a new leader.
> 
> Yes, there might be some bugs around that. Also, the K8s task runner makes certain list pod calls, which are pretty heavy and needs to be addressed. I think @capistrant is doing some work to improve that code flow.
> 
> > We try our best not to restart coordinator/overlord in production.
> 
> Oh, how frequently do you upgrade your cluster? Is changing the task capacity going to be much more frequent than that.
> 

We don't upgrade clusters very frequently, may be once a year or more than 1 year. We do adjust the capacity (upsize or downsize) regularly based on load/requirement.


> I agree that K8s task runner is buggy and we should improve upon it. But making the task capacity dynamic doesn't seem like the best solution. It will open a whole another can of worms and make this piece only more complicated.
> 
> Instead, we should trying to fix up the actual problems in the task runner which make Overlord leader switch erroneous.
> 
> What are your thoughts, @FrankChen021 ?

The main idea of dynamic configuration is not to circumvent problems at restarting phase, it's about reducing the operation complexity and saving time. even restarting overlord is smooth, I don't think changing such configuration requires a restart from users/operators' view. for static configurations, operators have to change configuration files, sync files to kubenetes, restarting components, it's a heavy work flow.

**Review by FrankChen021 — APPROVED** [review]

**Comment by kfaraz** [comment]
Thanks for the clarification, @FrankChen021 !

I am just a little apprehensive since the K8s task runner is already pretty buggy.
Also, it feels weird to have a config be specified by both static and dynamic means.
But I suppose the static config can be thought of as the default value.

I haven't gone through the whole PR yet. Will try to do a thorough review today.
@GabrielCWT , it would be nice if you could hold off on merging this PR until then.

**Comment by FrankChen021** [comment]
Hi @kfaraz are u reviewing this PR? I hope we can merge it into druid 35.

File: extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java [hunk 6]
```diff
@@ -179,6 +189,24 @@ protected KubernetesWorkItem joinAsync(Task task)
     }
   }
 
+  private void syncCapacityWithDynamicConfig()
```
cryptoe: This code is not maintainable. 
Can we do this in a seperate thread instead of calling sync everywhere. 
Is the currentCapacity  thread safe ?

GabrielCWT: It is possible to make `currentCapacity` to be thread safe. 

> Can we do this in a seperate thread instead of calling sync everywhere.

This was an alternative solution though I am not sure if we are thinking of the same implementation. My solution was to have a thread which periodically checks whether there has been any changes to the capacity and update `currentCapacity` accordingly. However, I wasn't sure if it was a waste of resources as there would need to be a trade off between responsiveness (how quick will the changes be visible) and resources (since we need to wake the thread up every X seconds).

kfaraz:
> Hi @kfaraz are u reviewing this PR? I hope we can merge it into druid 35.

Yes, @FrankChen021 , I will go through the changes either today or tomorrow.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
FrankChen021: using a separate thread to check is also a bad idea , it does not solve the real problem here but increase the complexity.

the real problem here is that the config manager does not provide a notification mechanism when it detects configuration changes.

If we look at the config manager implementation, it provides a method `swapIfNew` to check and set new values. This is a place where we can add notification.
I think we can add a new overridden `watch` method which  accepts a `Runnable` as callback.
This callback is kept in the internal `ConfigHolder`.

GabrielCWT: I've added observer support for ConfigHolders

FrankChen021:
@kfaraz any comments?

cecemei:
This pull request hasn’t been merged into the master branch yet, which means it won’t be included in the upcoming Druid 35 release. It will most likely be part of Druid 36 once it’s reviewed and merged.

kfaraz:
> @kfaraz any comments?

Hey @FrankChen021 , yeah, still planning on going through this PR. Will try to prioritize it for later this week.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerEffectiveConfig.java
```diff
@@ -0,0 +1,188 @@
...
+    return staticConfig.getAnnotations();
+  }
+
+  @Override
+  public Integer getCapacity()
+  {
+    if (dynamicConfigSupplier == null || dynamicConfigSupplier.get() == null || dynamicConfigSupplier.get().getCapacity() == null) {
+      return staticConfig.getCapacity();
+    }
+    return dynamicConfigSupplier.get().getCapacity();
```
cryptoe, line 177: Wont this be bit weird for the cluster admin. 
In the static configs, we supply lets say 10, but in the dynamic config we supply 20. How would the cluster admin reason about this ?
How are these configs reconciled ?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
```diff
@@ -179,6 +191,24 @@ protected KubernetesWorkItem joinAsync(Task task)
     }
   }
 
+  private void syncCapacityWithDynamicConfig(KubernetesTaskRunnerDynamicConfig config)
```
cryptoe, line 195: This looks much cleaner. Thanks @FrankChen021 for the suggestion and @GabrielCWT for the impl.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
cryptoe: Thank you the watcher based approach looks much more cleaner.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
```diff
@@ -132,10 +142,12 @@ public KubernetesTaskRunner(
...
     this.cleanupExecutor = Executors.newScheduledThreadPool(1);
-    this.exec = MoreExecutors.listeningDecorator(
-        Execs.multiThreaded(config.getCapacity(), "k8s-task-runner-%d")
-    );
     this.emitter = emitter;
+
+    this.currentCapacity = config.getCapacity();
+    this.tpe = new ThreadPoolExecutor(currentCapacity, currentCapacity, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(), Execs.makeThreadFactory("k8s-task-runner-%d", null));
+    this.exec = MoreExecutors.listeningDecorator(this.tpe);
+    configManager.addListener(KubernetesTaskRunnerDynamicConfig.CONFIG_KEY, StringUtils.format(OBSERVER_KEY, Thread.currentThread().getId()), this::syncCapacityWithDynamicConfig);
```
cryptoe, line 151: Would this run in the jetty thread?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
```diff
@@ -111,19 +117,23 @@ public class KubernetesTaskRunner implements TaskLogStreamer, TaskRunner
...
   private final KubernetesTaskRunnerConfig config;
   private final ListeningExecutorService exec;
+  private final ThreadPoolExecutor tpe;
   private final HttpClient httpClient;
   private final PeonLifecycleFactory peonLifecycleFactory;
   private final ServiceEmitter emitter;
   // currently worker categories aren't supported, so it's hardcoded.
   protected static final String WORKER_CATEGORY = "_k8s_worker_category";
 
+  private int currentCapacity;
```
cryptoe: This should be made thread safe. Use Atomic int.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerEffectiveConfig.java
GabrielCWT: You raise a good point. I don't think there is any way around this issue. I have updated the docs to state that the static config value will be overridden in the event that a dynamic config has been set. Hopefully this would add a little more clarity for users.

I am open to suggestions to make this even clearer.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
GabrielCWT: Yes it will run in the same thread as the one handling the API call. Would your concern be that if more listeners were to be added in the future, the updating of dynamic configs would be a long blocking call?

FrankChen021:
Hi @cryptoe Looks like all comments have been addressed. if no further comments, I will merge this

kfaraz:
Apologies! I have been bogged down with a bunch of other stuff, some to do with the D35 release as well.

@FrankChen021 , I would really appreciate it if you could hold off on merging this at least until Friday (tomorrow).
If I have not found some time to leave any feedback by then, and @cryptoe has approved, you can merge off this PR.

@cryptoe and I have discussed this briefly. For the most part, the capabilities added here make sense.
But I just want to take a quick pass through the code flow and structure since this is a critical piece of the task management system.

We could always merge off this PR and address my comments in a follow up PR but that would only make the changes more difficult to track.

I hope that makes sense.

FrankChen021:
@kfaraz can we merge this now? because I'm going to add a new kubernetes runner, and we hope this PR can be merged first otherwise there will be conflicts.

FrankChen021:
@kfaraz I merged this. if there's any concerns, let's resolve it in a new pr.

kfaraz:
Thanks for the update, @FrankChen021 !

My concern was mostly the code structure, the fact that we have to pass around an "effective" config and the merging of static and dynamic configs (even though most of the configs seem to be always static) without clearly communicating to the user the source of a specific config value, but we can always address it later if needed.

cryptoe:
+1 thanks @FrankChen021 for merging this. 
The new code structure seems much better.

PR #18592, opened 2025-10-01
Parallelize segment metadata queries at a data node-level

jtuglu1:
### Description

Currently, we use some legacy query runner to run segment metadata queries serially (segment-by-segment) on a data node. I'm not sure what the history is behind this decision; I assume it's due to resource contention concerns? Although there could be some merit to reducing query load on the data node, this could have other undesirable effects on the cluster by slowing down these queries used elsewhere (e.g. on brokers). Instead, I propose this noisy-neighbor concern can be prevented with the already-existing query prioritization (as well as the incoming per-segment timeout [feature](https://github.com/apache/druid/pull/18148)) to allow users to tune their queries' respective resource allocations rather than elide this parallelism from them.

This also has the added benefit of removing yet another one-off case (only one is now GroupByMergingQueryRunner) that doesn't use ChainedExecutionQueryRunner, helping to unify codepaths in the native engine. I feel like we should be "fixing" query resource contention issues with the help of default/user-supplied context parameters, not masking resource contention issues at the execution layer and/or putting this serial/parallel execution behavior behind a flag.

#### Release note

Parallelize segment metadata queries at a data node-level

<hr>

This PR has:

- [x] been self-reviewed.


maytasm (approved):

abhishekrb19 (approved):
Looks good to me, thanks!

After this change, it looks like the only remaining usage of `ConcatQueryRunner` is a test class `ServerManagerTest`, which can also be updated to use the `ChainedExecutionQueryRunner` and then we can remove the legacy runner altogether. I'm okay with doing that cleanup in a follow up too.

PR #18594, opened 2025-10-03
Interval optimization for subclasses of DataSegment

GWphua:
Follow-up to #18477 

### Description

Subclasses of `DataSegment` will still use the old deserialization method. Make JSON constructors private, and use the `Intervals.fromString()` method introduced in #18477.

<hr>

 * `DataSegmentWithLocation`
 * `LoadableDataSegment`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
+++ b/processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
@@ -78,39 +78,39 @@ public final class Intervals
    *
    * Currently, this method is only used in {@link org.apache.druid.timeline.SegmentId}.
    */
-  public static Interval fromString(String string)
+  public static Interval fromString(String serializedInterval)
   {
-    if (canDeserializeIntervalOptimallyFromString(string)) {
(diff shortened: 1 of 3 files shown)


processing/src/main/java/org/apache/druid/java/util/common/Intervals.java
```diff
@@ -80,12 +80,15 @@ public static Interval of(String format, Object... formatArgs)
    */
   public static Interval fromString(String string)
```
kfaraz: Nit:
```suggestion
  public static Interval fromString(String serializedInterval)
```

GWphua: @kfaraz  Changed, and propagated the change for related methods.

GWphua:
Failure unrelated to changes. Previous commit already cleared CI checks.

kfaraz:
@GWphua , could you please try pulling in the latest master? That would trigger another CI run.
Even though the failure is unrelated to your PR, I want to identify the source of flakiness.
Some of the flakiness had been recently addressed.

PR #18595, opened 2025-10-03
Fix Markdown table formatting in Delta input source docs

yurmix:
This PR fixes a documentation formatting issue.

A column separator was missing in the Markdown table, which caused it to render incorrectly. As a result, the content did not display as a table on the documentation site.


kfaraz (approved):
thanks for fixing this, @yurmix !

PR #18598, opened 2025-10-04
Implement metricTtl for prometheus-emitter

### LINKED ISSUE #14638: prometheus-emitter metrics ttl
We're using the prometheus-emitter for metrics on druid 26.0.0. This is all working OK, metrics are flowing and all is well with the world (almost).

Query metrics are showing up as expected, but the prometheus endpoint never appears to timeout the metrics. 

For example, we'll run a query, the metric for that query will show up (druid_query_wait_time as an example) and be scraped by prometheus. A minute/hour/day later, that same query metric is seemingly still being exposed by the prometheus emitter. Adding in some of the dimensions for query metrics (id being one), the cardinality is increased and it's causing the number of metrics to grow rapidly and causing stability issues. 

It's all pretty standard prometheus-emitter deployments using the exporter strategy, so I'm wondering whether I'm missing something (maybe a configuration option for the emitter to reset the metrics it's exposing after a pre-defined period).

The http emitter has a druid.emitter.http.flushMillis setting which (I appreciate is pushing rather than scraping) is the behaviour I think we're missing

aho135:
Fixes #14638.

### Description

This PR adds a new property to the prometheus-emitter called metricTtlMs. Prometheus is unique amongst emitters as it utilizes a pull model instead of a push model. This leads to an issue with stale metrics. For example, if the active Coordinator leader is emitting ingest/kafka/lag but then a leader election change occurs, the Coordinator would continue to emit that stale metric value until it is restarted.

An alternative option is to reset the metric values within DruidCoordinator. There we have access to the ServiceEmitter. We could update the Emitter interface to have a clear() method that is called in stopBeingLeader. For most emitters though, this method would do nothing since they use a push model. Instead of muddying the interface I thought it better to implement this in prometheus-emitter itself

#### Release note

Adds a new property to the prometheus-emitter called metricTtlMs which allows the operator to specify the ttl for a metric. If a metric value has not been updated within the ttl period then the value will be reset to its starting value

<hr>

 * `PrometheusEmitter`
 * `PrometheusEmitterConfig`
 * `DimensionsAndCollector`
 * `PrometheusEmitterConfigTest`
 * `PrometheusEmitterTest`

<hr>

This PR has:

- [X] been self-reviewed.

### DIFF
--- a/extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
+++ b/extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
@@ -20,6 +20,7 @@
 package org.apache.druid.emitter.prometheus;
 
 import io.prometheus.client.SimpleCollector;
+import java.util.concurrent.atomic.AtomicLong;
 
 public class DimensionsAndCollector
 {
@@ -27,6 +28,7 @@ public class DimensionsAndCollector
   private final SimpleCollector collector;
   private final double conversionFactor;
   private final double[] histogramBuckets;
+  private final AtomicLong lastUpdateTime;
 
   DimensionsAndCollector(String[] dimensions, SimpleCollector collector, double conversionFactor, double[] histogramBuckets)
   {
@@ -34,6 +36,7 @@ public class DimensionsAndCollector
     this.collector = collector;
     this.conversionFactor = conversionFactor;
     this.histogramBuckets = histogramBuckets;
+    this.lastUpdateTime = new AtomicLong(System.currentTimeMillis());
   }
 
   public String[] getDimensions()
@@ -54,5 +57,21 @@ public class DimensionsAndCollector
   public double[] getHistogramBuckets()
   {
     return histogramBuckets;
+  }
+
+  public void updateLastUpdateTime()
+  {
+    lastUpdateTime.set(System.currentTimeMillis());
+  }
+
+  public long getLastUpdateTime()
+  {
+    return lastUpdateTime.get();
+  }
+
+  public boolean isExpired(long ttlMillis)
+  {
+    long currentTime = System.currentTimeMillis();
+    return (currentTime - lastUpdateTime.get()) > ttlMillis;
   }
 }
--- a/extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
+++ b/extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
@@ -59,6 +59,7 @@ public class PrometheusEmitter implements Emitter
   private PushGateway pushGateway;
   private volatile String identifier;
   private ScheduledExecutorService exec;
+  private ScheduledExecutorService ttlExecutor;
 
   static PrometheusEmitter of(PrometheusEmitterConfig config)
   {
@@ -92,6 +93,18 @@ public class PrometheusEmitter implements Emitter
         }
       } else {
         log.error("HTTPServer is already started");
+      }
+      // Start TTL scheduler if TTL is configured
+      if (config.getMetricTtlMs() != null) {
+        ttlExecutor = ScheduledExecutors.fixed(1, "PrometheusTTLExecutor-%s");
+        // Check TTL every minute
+        ttlExecutor.scheduleAtFixedRate(
+            this::checkMetricTtl,
+            config.getMetricTtlMs(),
+            config.getMetricTtlMs(),
+            TimeUnit.MILLISECONDS
+        );
+        log.info("Started TTL scheduler with TTL of %d milliseconds", config.getMetricTtlMs());
       }
     } else if (strategy.equals(PrometheusEmitterConfig.Strategy.pushgateway)) {
       String address = config.getPushGatewayAddress();
@@ -140,6 +153,7 @@ public class PrometheusEmitter implements Emitter
 
     DimensionsAndCollector metric = metrics.getByName(name, service);
     if (metric != null) {
+      metric.updateLastUpdateTime();
       String[] labelValues = new String[metric.getDimensions().length];
       String[] labelNames = metric.getDimensions();
 
@@ -211,6 +225,9 @@ public class PrometheusEmitter implements Emitter
       if (server != null) {
         server.close();
       }
+      if (ttlExecutor != null) {
+        ttlExecutor.shutdownNow();
+      }
     } else {
       exec.shutdownNow();
       flush();
@@ -247,6 +264,11 @@ public class PrometheusEmitter implements Emitter
     return server;
   }
 
+  public Metrics getMetrics()
+  {
+    return metrics;
+  }
+
   public PushGateway getPushGateway()
   {
     return pushGateway;
@@ -255,5 +277,22 @@ public class PrometheusEmitter implements Emitter
   public void setPushGateway(PushGateway pushGateway)
   {
     this.pushGateway = pushGateway;
+  }
+
+  private void checkMetricTtl()
+  {
+    if (config.getMetricTtlMs() == null) {
+      return;
+    }
+
+    Map<String, DimensionsAndCollector> map = metrics.getRegisteredMetrics();
+    for (Map.Entry<String, DimensionsAndCollector> entry : map.entrySet()) {
+      if (entry.getValue().isExpired(config.getMetricTtlMs())) {
+        log.info("Metric [%s] has expired (last updated %d ms ago)",
+                 entry.getKey(),
+                 System.currentTimeMillis() - entry.getValue().getLastUpdateTime());
+        entry.getValue().getCollector().clear();
+      }
+    }
   }
 }
--- a/extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitterConfig.java
+++ b/extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitterConfig.java
@@ -77,6 +77,10 @@ public class PrometheusEmitterConfig
   @JsonProperty
   private final Duration waitForShutdownDelay;
 
+  @JsonProperty
+  @Nullable
+  private final Long metricTtlMs;
+
   @JsonCreator
   public PrometheusEmitterConfig(
       @JsonProperty("strategy") @Nullable Strategy strategy,
@@ -86,10 +90,11 @@ public class PrometheusEmitterConfig
       @JsonProperty("pushGatewayAddress") @Nullable String pushGatewayAddress,
       @JsonProperty("addHostAsLabel") boolean addHostAsLabel,
       @JsonProperty("addServiceAsLabel") boolean addServiceAsLabel,
-      @JsonProperty("flushPeriod") Integer flushPeriod,
+      @JsonProperty("flushPeriod") @Nullable Integer flushPeriod,
       @JsonProperty("extraLabels") @Nullable Map<String, String> extraLabels,
       @JsonProperty("deletePushGatewayMetricsOnShutdown") @Nullable Boolean deletePushGatewayMetricsOnShutdown,
-      @JsonProperty("waitForShutdownDelay") @Nullable Long waitForShutdownDelay
+      @JsonProperty("waitForShutdownDelay") @Nullable Long waitForShutdownDelay,
+      @JsonProperty("metricTtlMs") @Nullable Long metricTtlMs
   )
   {
     this.strategy = strategy != null ? strategy : Strategy.exporter;
@@ -128,6 +133,18 @@ public class PrometheusEmitterConfig
                               )
                           );
     }
+
+    if (metricTtlMs != null && metricTtlMs <= 0) {
+      throw DruidException.forPersona(DruidException.Persona.OPERATOR)
+                          .ofCategory(DruidException.Category.INVALID_INPUT)
+                          .build(
+                              StringUtils.format(
+                                  "Invalid value for metricTtlMs[%s] specified, metricTtlMs must be > 0.",
+                                  metricTtlMs
+                              )
+                          );
+    }
+    this.metricTtlMs = metricTtlMs;
 
     // Validate label names early to prevent Prometheus exceptions later.
     for (String key : this.extraLabels.keySet()) {
@@ -199,6 +216,12 @@ public class PrometheusEmitterConfig
   public Duration getWaitForShutdownDelay()
   {
     return waitForShutdownDelay;
+  }
+
+  @Nullable
+  public Long getMetricTtlMs()
+  {
+    return metricTtlMs;
   }
 
   public enum Strategy
--- a/extensions-contrib/prometheus-emitter/src/test/java/org/apache/druid/emitter/prometheus/PrometheusEmitterTest.java
+++ b/extensions-contrib/prometheus-emitter/src/test/java/org/apache/druid/emitter/prometheus/PrometheusEmitterTest.java
@@ -45,7 +45,7 @@ public class PrometheusEmitterTest
   public void testEmitterWithServiceLabel()
   {
     CollectorRegistry.defaultRegistry.clear();
-    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, null, false, null);
+    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, null, false, null, null);
     PrometheusEmitterModule prometheusEmitterModule = new PrometheusEmitterModule();
     Emitter emitter = prometheusEmitterModule.getEmitter(config);
     ServiceMetricEvent build = ServiceMetricEvent.builder()
@@ -66,7 +66,7 @@ public class PrometheusEmitterTest
   public void testEmitterWithServiceAndHostLabel()
   {
     CollectorRegistry.defaultRegistry.clear();
-    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, true, true, 60, null, false, null);
+    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, true, true, 60, null, false, null, null);
     PrometheusEmitterModule prometheusEmitterModule = new PrometheusEmitterModule();
     Emitter emitter = prometheusEmitterModule.getEmitter(config);
     ServiceMetricEvent build = ServiceMetricEvent.builder()
@@ -89,7 +89,7 @@ public class PrometheusEmitterTest
     CollectorRegistry.defaultRegistry.clear();
     Map<String, String> extraLabels = new HashMap<>();
     extraLabels.put("labelName", "labelValue");
-    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, false, 60, extraLabels, false, null);
+    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, false, 60, extraLabels, false, null, null);
     PrometheusEmitterModule prometheusEmitterModule = new PrometheusEmitterModule();
     Emitter emitter = prometheusEmitterModule.getEmitter(config);
     ServiceMetricEvent build = ServiceMetricEvent.builder()
@@ -111,7 +111,7 @@ public class PrometheusEmitterTest
     CollectorRegistry.defaultRegistry.clear();
     Map<String, String> extraLabels = new HashMap<>();
     extraLabels.put("labelName", "labelValue");
-    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, extraLabels, false, null);
+    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, extraLabels, false, null, null);
     PrometheusEmitterModule prometheusEmitterModule = new PrometheusEmitterModule();
     Emitter emitter = prometheusEmitterModule.getEmitter(config);
     ServiceMetricEvent build = ServiceMetricEvent.builder()
@@ -132,7 +132,7 @@ public class PrometheusEmitterTest
   {
     CollectorRegistry.defaultRegistry.clear();
     Map<String, String> extraLabels = new HashMap<>();
-    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, extraLabels, false, null);
+    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, extraLabels, false, null, null);
     PrometheusEmitterModule prometheusEmitterModule = new PrometheusEmitterModule();
     Emitter emitter = prometheusEmitterModule.getEmitter(config);
     ServiceMetricEvent build = ServiceMetricEvent.builder()
@@ -155,7 +155,7 @@ public class PrometheusEmitterTest
     Map<String, String> extraLabels = new HashMap<>();
     extraLabels.put("labelName1", "labelValue1");
     extraLabels.put("labelName2", "labelValue2");
-    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, extraLabels, false, null);
+    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, extraLabels, false, null, null);
     PrometheusEmitterModule prometheusEmitterModule = new PrometheusEmitterModule();
     Emitter emitter = prometheusEmitterModule.getEmitter(config);
     ServiceMetricEvent build = ServiceMetricEvent.builder()
@@ -178,7 +178,7 @@ public class PrometheusEmitterTest
     // ExtraLabels contains a label that collides with a service label
     Map<String, String> extraLabels = new HashMap<>();
     extraLabels.put("server", "collisionLabelValue");
-    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, extraLabels, false, null);
+    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, null, null, 0, null, false, true, 60, extraLabels, false, null, null);
     PrometheusEmitterModule prometheusEmitterModule = new PrometheusEmitterModule();
     Emitter emitter = prometheusEmitterModule.getEmitter(config);
     ServiceMetricEvent build = ServiceMetricEvent.builder()
@@ -205,7 +205,7 @@ public class PrometheusEmitterTest
   public void testEmitterMetric()
   {
     CollectorRegistry.defaultRegistry.clear();
-    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.pushgateway, "namespace", null, 0, "pushgateway", true, true, 60, null, false, null);
+    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.pushgateway, "namespace", null, 0, "pushgateway", true, true, 60, null, false, null, null);
     PrometheusEmitterModule prometheusEmitterModule = new PrometheusEmitterModule();
     Emitter emitter = prometheusEmitterModule.getEmitter(config);
     ServiceMetricEvent build = ServiceMetricEvent.builder()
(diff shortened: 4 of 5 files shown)


extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -140,6 +153,7 @@ private void emitMetric(ServiceMetricEvent metricEvent)
 
     DimensionsAndCollector metric = metrics.getByName(name, service);
     if (metric != null) {
+      metric.updateLastUpdateTime();
```
kfaraz: Update the time after the respective collector has been actually updated.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -20,20 +20,23 @@
...
 import io.prometheus.client.SimpleCollector;
+import java.util.concurrent.atomic.AtomicLong;
 
 public class DimensionsAndCollector
 {
   private final String[] dimensions;
   private final SimpleCollector collector;
   private final double conversionFactor;
   private final double[] histogramBuckets;
+  private final AtomicLong lastUpdateTime;
```
kfaraz: You can use a `org.apache.druid.java.util.common.Stopwatch` instead.
In `updateLastUpdateTime`, you can `restart` the stopwatch.
In `isExpired`, you can invoke `stopwatch.hasElapsed()`.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -93,6 +94,18 @@ public void start()
       } else {
         log.error("HTTPServer is already started");
       }
+      // Start TTL scheduler if TTL is configured
+      if (config.getMetricTtlMs() != null) {
+        ttlExecutor = ScheduledExecutors.fixed(1, "PrometheusTTLExecutor-%s");
+        // Check TTL every minute
+        ttlExecutor.scheduleAtFixedRate(
```
kfaraz: We should reuse the existing `exec`.
With `exporter` strategy, it will be used to clean up stale metrics.
With `pushgateway` strategy, it will be used to push metrics.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitterConfig.java
```diff
@@ -77,6 +77,10 @@ public class PrometheusEmitterConfig
   @JsonProperty
   private final Duration waitForShutdownDelay;
 
+  @JsonProperty
+  @Nullable
+  private final Long metricTtlMs;
```
kfaraz: We should use the existing `flushPeriod` config.
In `exporter` mode, it will represent the TTL.
In `pushgateway` mode, it will represent the push period.
We can update the docs to reflect this detail.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -55,4 +58,20 @@ public double[] getHistogramBuckets()
   {
     return histogramBuckets;
   }
+
+  public void updateLastUpdateTime()
```
kfaraz:
```suggestion
  public void resetLastUpdateTime()
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -256,4 +278,21 @@ public void setPushGateway(PushGateway pushGateway)
   {
     this.pushGateway = pushGateway;
   }
+
+  private void checkMetricTtl()
```
kfaraz:
```suggestion
  private void cleanUpStaleMetrics()
```

aho135:
> Thanks for the fix, @aho135 ! I have left some minor suggestions.
> 
> Could you share some screenshots where we can see stale metrics being reported?

Thanks for the review @kfaraz!

This is the ingest/kafka/partitionlag metric being emitted by 2 Coordinators. The active one is emitting the proper metric (0) but the previous Coordinator is emitting a stale metric that doesn't get reset until we manually restart it. The scenario we run into is that if a leader change occurs while there is lag on a topic then the old Coordinator continues to emit that stale lag metric. We have some alerting set up for lag, so the stale value ends up triggering false alarms.

Screenshot 2025-10-06 at 1 59 05 PM

docs/development/extensions-contrib/prometheus.md
```diff
@@ -44,7 +44,7 @@ All the configuration parameters for the Prometheus emitter are under `druid.emi
...
-| `druid.emitter.prometheus.flushPeriod`        | Emit metrics to Pushgateway every `flushPeriod` seconds. Required if `pushgateway` strategy is used.                                                                                                                                   | no        | 15                                   |
+| `druid.emitter.prometheus.flushPeriod`        | If strategy is `pushgateway`, emits metrics every `flushPeriod` seconds. Required if `pushgateway` strategy is used. If strategy is `exporter`, configures the metric ttl such that if the metric value is not updated within the `flushPeriod` then it will stop being emitted. Optional if `exporter` strategy is used.                                                                                                                                   | no        | 15 seconds for `pushgateway` strategy. None for `exporter` strategy                                  |
```
kfaraz:
```suggestion
| `druid.emitter.prometheus.flushPeriod`        | If strategy is `pushgateway`, emits metrics every `flushPeriod` seconds. If strategy is `exporter`, configures the metric ttl such that if the metric value is not updated within the `flushPeriod` then it will stop being emitted.                                                                                                                                   | Required if `pushgateway` strategy is used, optional otherwise.        | 15 seconds for `pushgateway` strategy. None for `exporter` strategy. |
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -55,4 +61,20 @@ public double[] getHistogramBuckets()
...
+  }
+
+  public long getTimeSinceLastUpdate()
+  {
+    return updateTimer.millisElapsed();
+  }
+
+  public boolean isExpired(long ttlSeconds)
+  {
+    return updateTimer.hasElapsed(new Duration(TimeUnit.SECONDS.toMillis(ttlSeconds)));
```
kfaraz:
```suggestion
    return updateTimer.hasElapsed(Duration.standardSeconds(ttlSeconds));
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -93,6 +93,18 @@ public void start()
       } else {
         log.error("HTTPServer is already started");
       }
+      // Start TTL scheduler if TTL is configured
+      if (config.getFlushPeriod() != null) {
+        exec = ScheduledExecutors.fixed(1, "PrometheusTTLExecutor-%s");
+        // Check TTL every minute
```
kfaraz: I don't think this comment is valid anymore, please remove it.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -93,6 +93,18 @@ public void start()
...
+        exec.scheduleAtFixedRate(
+            this::cleanUpStaleMetrics,
+            config.getFlushPeriod(),
+            config.getFlushPeriod(),
+            TimeUnit.SECONDS
+        );
+        log.info("Started TTL scheduler with TTL of %d seconds", config.getFlushPeriod());
```
kfaraz:
```suggestion
        log.info("Started TTL scheduler with TTL of [%d] seconds.", config.getFlushPeriod());
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -55,4 +61,20 @@ public double[] getHistogramBuckets()
...
     return histogramBuckets;
   }
+
+  public void resetLastUpdateTime()
+  {
+    updateTimer.reset();
+    updateTimer.start();
+  }
+
+  public long getTimeSinceLastUpdate()
```
kfaraz:
```suggestion
  public long getMillisSinceLastUpdate()
```

extensions-contrib/prometheus-emitter/src/test/java/org/apache/druid/emitter/prometheus/PrometheusEmitterTest.java
```diff
@@ -430,4 +431,78 @@ public void testEmitterWithoutDeleteOnShutdown() throws IOException
...
+                                                 .build(ImmutableMap.of("service", "historical", "host", "druid.test.cn"));
+    emitter.emit(event);
+
+    // Get the metrics and check that it's not expired initially
+    Map<String, DimensionsAndCollector> registeredMetrics = emitter.getMetrics().getRegisteredMetrics();
+    DimensionsAndCollector testMetric = registeredMetrics.get("segment/loadQueue/count");
+
+    Assert.assertNotNull("Test metric should be registered", testMetric);
+    Assert.assertFalse("Metric should not be expired initially",
+                       testMetric.isExpired(flushPeriod));
```
kfaraz: Style: cleaner to put the first arg in a new line. Please use the same style in other places too.
```suggestion
    Assert.assertFalse(
        "Metric should not be expired initially",
        testMetric.isExpired(flushPeriod)
    );
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -256,4 +279,21 @@ public void setPushGateway(PushGateway pushGateway)
...
+    if (config.getFlushPeriod() == null) {
+      return;
+    }
+
+    Map<String, DimensionsAndCollector> map = metrics.getRegisteredMetrics();
+    for (Map.Entry<String, DimensionsAndCollector> entry : map.entrySet()) {
+      if (entry.getValue().isExpired(config.getFlushPeriod())) {
+        log.debug("Metric [%s] has expired (last updated %d ms ago)",
+                 entry.getKey(),
+                 entry.getValue().getTimeSinceLastUpdate());
```
kfaraz: Style:
```suggestion
        log.debug(
            "Metric[%s] has expired as it was last updated [%d] ms ago.",
            entry.getKey(),
            entry.getValue().getTimeSinceLastUpdate()
        );
```

kfaraz (commented):
Looks good, left some final suggestions.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitterConfig.java
```diff
@@ -97,6 +97,9 @@ public PrometheusEmitterConfig(
     Preconditions.checkArgument(PATTERN.matcher(this.namespace).matches(), "Invalid namespace " + this.namespace);
     if (strategy == Strategy.exporter) {
       Preconditions.checkArgument(port != null, "For `exporter` strategy, port must be specified.");
+      if (Objects.nonNull(flushPeriod)) {
+        Preconditions.checkArgument(flushPeriod > 0, "flushPeriod must be greater than 0.");
+      }
```
abhishekrb19: - Would be good to include the invalid `flushPeriod` in the error message
- Looks like the other validation uses this to, could use `DruidException` for newer ones. Something like:

``` 
throw DruidException.forPersona(DruidException.Persona.OPERATOR)
                          .ofCategory(DruidException.Category.INVALID_INPUT)
                          .build(
                                  "Invalid value for 'druid.emitter.prometheus.flushPeriod'[%s] specified, flushPeriod must be > 0 if specified.",
                                  waitForShutdownDelay
                              )
                          );
```

docs/development/extensions-contrib/prometheus.md
abhishekrb19: Thanks for the docs! Minor suggestion:
```suggestion
| `druid.emitter.prometheus.flushPeriod`        | If strategy is `pushgateway`, metrics are emitted every `flushPeriod` seconds. Required if `pushgateway` strategy is used. If strategy is `exporter`, this configures the metric TTL such that if the metric value is not updated within the `flushPeriod` seconds, then it will stop being emitted. Optional if `exporter` strategy is used.                                                                                                                                   | no        | 15 seconds for `pushgateway` strategy. None for `exporter` strategy                                  |
```

Also, do you have guidance on what an operator should set this value to for TTL eviction (perhaps something as a function of emission period)?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
abhishekrb19: I think this comment is stale?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -256,4 +279,21 @@ public void setPushGateway(PushGateway pushGateway)
...
+
+  private void cleanUpStaleMetrics()
+  {
+    if (config.getFlushPeriod() == null) {
+      return;
+    }
+
+    Map<String, DimensionsAndCollector> map = metrics.getRegisteredMetrics();
+    for (Map.Entry<String, DimensionsAndCollector> entry : map.entrySet()) {
+      if (entry.getValue().isExpired(config.getFlushPeriod())) {
```
abhishekrb19: Extract `entry.getValue()` to a variable and reuse since it's used a few times in this loop :)

abhishekrb19: `DimensionsAndCollector` could accept the flush period directly, initialized from the Prometheus config that’s plumbed through `Metrics`. Then this code could become `if (entry.getValue().isExpired()) {}` where this fixed config isn't redundantly passed in

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -211,6 +226,9 @@ public void close()
       if (server != null) {
         server.close();
       }
+      if (exec != null) {
+        exec.shutdownNow();
+      }
```
abhishekrb19, line 220: We could reverse the ordering of operations in `close()` since it's generally good practice to close resources in the reverse order of how they were initialized in `start()`; move this before the code block where we do `server.close()`

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -256,4 +279,21 @@ public void setPushGateway(PushGateway pushGateway)
   {
     this.pushGateway = pushGateway;
   }
+
+  private void cleanUpStaleMetrics()
```
abhishekrb19, line 280: Could you please add a javadoc here? Or for the `exec` member?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -256,4 +279,21 @@ public void setPushGateway(PushGateway pushGateway)
...
+      return;
+    }
+
+    Map<String, DimensionsAndCollector> map = metrics.getRegisteredMetrics();
+    for (Map.Entry<String, DimensionsAndCollector> entry : map.entrySet()) {
+      if (entry.getValue().isExpired(config.getFlushPeriod())) {
+        log.debug("Metric [%s] has expired (last updated %d ms ago)",
+                 entry.getKey(),
+                 entry.getValue().getTimeSinceLastUpdate());
+        entry.getValue().getCollector().clear();
```
abhishekrb19: Is this `clear()` operation on `SimpleCollector` thread-safe? Doing a quick scan of the [docs](https://prometheus.io/docs/instrumenting/writing_clientlibs/#overall-structure) suggest that the client libraries must be thread-safe, but wanted to confirm that clearing can concurrently happen with the metric updates in `emitMetric()`


Side note: I think it would be better to move some of these methods inside `DimensionsAndCollector` to avoid abstraction leaks. But that refactor can be done later since it'd also require moving the metric updates from `emitMetric()` into this class.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -55,4 +61,20 @@ public double[] getHistogramBuckets()
   {
     return histogramBuckets;
   }
+
+  public void resetLastUpdateTime()
+  {
+    updateTimer.reset();
+    updateTimer.start();
```
abhishekrb19: nit: can be simplified to `updateTimer.restart();`

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
abhishekrb19:
```suggestion
        log.info("Started TTL scheduler with TTL of [%d] seconds", config.getFlushPeriod());
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
abhishekrb19: It looks like `clear()` removes all children for the metric. Do we instead want to use `remove()` to target a specific time series by specifying the corresponding set of label values?

Without that, metrics that track things like task ID or query ID, for example, `task/run/time` or `query/wait/time` mentioned in https://github.com/apache/druid/issues/14638 wouldn’t get evicted if they’re being updated frequently.

When we call `metric.resetLastUpdateTime()`, should we pass the appropriate label set and track it? Then in `cleanUpStaleMetrics()`, we could check the label sets and clear only the stale time series rather than clearing all of them or none at all. What do you think?

extensions-contrib/prometheus-emitter/src/test/java/org/apache/druid/emitter/prometheus/PrometheusEmitterTest.java
```diff
@@ -430,4 +431,78 @@ public void testEmitterWithoutDeleteOnShutdown() throws IOException
...
+  {
+    int flushPeriod = 3;
+    PrometheusEmitterConfig config = new PrometheusEmitterConfig(PrometheusEmitterConfig.Strategy.exporter, "test", null, 0, null, false, true, flushPeriod, null, false, null);
+    PrometheusEmitter emitter = new PrometheusEmitter(config);
+    emitter.start();
+
+    ServiceMetricEvent event = ServiceMetricEvent.builder()
+                                                 .setMetric("segment/loadQueue/count", 10)
+                                                 .build(ImmutableMap.of("service", "historical", "host", "druid.test.cn"));
+    emitter.emit(event);
```
abhishekrb19, line 484: Could you add the same metric with different label values so that it actrs as a different timeseries and then verify that one exists while the other one has expired?

abhishekrb19 (commented):
Thanks @aho135 for the feature, this is very useful! I've left some questions on the TTL-based tracking and some misc suggestions.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
kfaraz: Hmm, I suppose removing the children for specific labels would be a more complete solution.
But it could blow up the memory usage pretty fast since we would now be tracking the last updated time for every unique set of label values for every metric.

I think the issue this PR is trying to address is that if a certain metric stops being reported by a service
altogether (irrespective of label values), it should not remain in the collector anymore. e.g. in leadership changes.

Also, the staleness concern should really only apply to gauge metrics. So if timer (histogram) metrics like `task/run/time` and `query/wait/time` are still being updated by the current service (for any label), we should continue emitting them. Getting rid of the expired labels becomes more of a mem usage concern than anything else.

But with gauge metrics like say `segment/size` (which represents bytes of segment available on a historical for a given datasource), I agree that the story is different.
If all segments of a datasource are removed from a historical, we should definitely stop reporting `segment/size` for that datasource.

But I would suggest we add a comment and postpone the full solution for gauge metrics for a later PR.
Ideally, we would want to write up an embedded test (using Prometheus testcontainer) and verify the behaviour of gauge metrics in such cases and then proceed with a solution.

What are your thoughts, @abhishekrb19 , @aho135 ?

docs/development/extensions-contrib/prometheus.md
kfaraz: Yeah, good point. I guess it would have to account for the scraping frequency too.
So, a good value would be something like `3 * scrape period` or `3 * emission period`, whichever is larger.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
abhishekrb19: >Hmm, I suppose removing the children for specific labels would be a more complete solution.
But it could blow up the memory usage pretty fast since we would now be tracking the last updated time for every unique set of label values for every metric.


Yeah, good callout on the memory usage when tracking every unique combination of label sets per metric. Prometheus isn’t designed for very [high cardinality labels](https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels), and the Prometheus `SimpleCollector` in the java library already tracks the unique label sets (`children`) internally. On the extension side, we’d only be dealing with a subset of those `children`, since we’d remove the ephemeral “stale” entries anyway (if there's a way to leverage the collector’s internal state somehow, that would be even better). So I’m not too concerned about the additional memory this would require, though it might still be worth validating in a cluster with some labels with reasonable cardinality and load.

I’m also okay with the suggestion to start without labels for now and revisit this (or another solution) once we have a better understanding of the additional memory overhead it might introduce. The current implementation should remain extensible enough to support label tracking in the future if needed. We can keep [#14638](https://github.com/apache/druid/issues/14638)
 open and call out in the documentation that the TTL expiration only applies to certain metric types and not to individual timeseries'. 
@aho135 @kfaraz what do you think?

aho135: Yeah, I was also concerned about memory usage. Especially where folks are adding queryId as a dimension like in #14638. 

Sounds good to me to tackle the label problem in a separate PR. We can say that this PR specifically targets resetting the metric value in the leader election scenario

aho135: Yup, confirmed that the Map getting cleared is a ConcurrentHashMap

extensions-contrib/prometheus-emitter/src/test/java/org/apache/druid/emitter/prometheus/PrometheusEmitterTest.java
aho135: I guess we can defer this test since we're not considering labels when clearing in this PR

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -20,20 +20,26 @@
...
 {
   private final String[] dimensions;
   private final SimpleCollector collector;
   private final double conversionFactor;
   private final double[] histogramBuckets;
+  private final Stopwatch updateTimer;
+  private final Integer ttlSeconds;
 
-  DimensionsAndCollector(String[] dimensions, SimpleCollector collector, double conversionFactor, double[] histogramBuckets)
+  DimensionsAndCollector(String[] dimensions, SimpleCollector collector, double conversionFactor, double[] histogramBuckets, Integer ttlSeconds)
```
abhishekrb19:
```suggestion
  DimensionsAndCollector(String[] dimensions, SimpleCollector collector, double conversionFactor, double[] histogramBuckets, @Nullable Integer ttlSeconds)
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/Metrics.java
```diff
@@ -63,7 +63,7 @@ public DimensionsAndCollector getByName(String name, String service)
     }
   }
 
-  public Metrics(String namespace, String path, boolean isAddHostAsLabel, boolean isAddServiceAsLabel, Map<String, String> extraLabels)
+  public Metrics(String namespace, String path, boolean isAddHostAsLabel, boolean isAddServiceAsLabel, Map<String, String> extraLabels, Integer ttlSeconds)
```
abhishekrb19:
```suggestion
  public Metrics(String namespace, String path, boolean isAddHostAsLabel, boolean isAddServiceAsLabel, Map<String, String> extraLabels, @Nullable Integer ttlSeconds)
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -20,20 +20,26 @@
...
   {
     this.dimensions = dimensions;
     this.collector = collector;
     this.conversionFactor = conversionFactor;
     this.histogramBuckets = histogramBuckets;
+    this.updateTimer = Stopwatch.createStarted();
+    this.ttlSeconds = ttlSeconds;
```
abhishekrb19: nit: you could assign `Duration.standardSeconds(ttlSeconds)` in the constructor itself

website/.spelling
```diff
@@ -2456,3 +2456,6 @@ PropertyKeyName
 PropertyValueType
 tableSpec
 TableSpec
+
+- ../docs/development/extensions-contrib/prometheus.md
```
abhishekrb19, line 2460: I think this isn't needed?

```suggestion
```

docs/development/extensions-contrib/prometheus.md
```diff
@@ -44,7 +44,7 @@ All the configuration parameters for the Prometheus emitter are under `druid.emi
...
 | `druid.emitter.prometheus.pushGatewayAddress` | Pushgateway address. Required if using `pushgateway` strategy.                                                                                                                                                                         | no        | none                                 |
-| `druid.emitter.prometheus.flushPeriod`        | Emit metrics to Pushgateway every `flushPeriod` seconds. Required if `pushgateway` strategy is used.                                                                                                                                   | no        | 15                                   |
+| `druid.emitter.prometheus.flushPeriod`        | When using the `pushgateway` strategy metrics are emitted every `flushPeriod` seconds. <br/>When using the `exporter` strategy this configures the metric TTL such that if the metric value is not updated within `flushPeriod` seconds then it will stop being emitted. It is recommended to set this to at least 3 * `scrape_interval`. | Required if `pushgateway` strategy is used, optional otherwise. | 15 seconds for `pushgateway` strategy. <br/>None for `exporter` strategy. |
```
abhishekrb19: An additional note on the behavior for clarity:
```suggestion
| `druid.emitter.prometheus.flushPeriod`        | When using the `pushgateway` strategy metrics are emitted every `flushPeriod` seconds. <br/>When using the `exporter` strategy this configures the metric TTL such that if the metric value is not updated within `flushPeriod` seconds then it will stop being emitted. Note that unique label combinations per metric are currently not subject to TTL expiration. It is recommended to set this to at least 3 * `scrape_interval`. | Required if `pushgateway` strategy is used, optional otherwise. | 15 seconds for `pushgateway` strategy. <br/>None for `exporter` strategy. |
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -20,20 +20,26 @@
...
+import org.joda.time.Duration;
 
 public class DimensionsAndCollector
 {
   private final String[] dimensions;
   private final SimpleCollector collector;
   private final double conversionFactor;
   private final double[] histogramBuckets;
+  private final Stopwatch updateTimer;
+  private final Integer ttlSeconds;
```
kfaraz: Nit: This can be a `Duration` to avoid repeated created of a `Duration` object in `isExpired`.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -74,7 +74,8 @@ public PrometheusEmitter(PrometheusEmitterConfig config)
         config.getDimensionMapPath(),
         config.isAddHostAsLabel(),
         config.isAddServiceAsLabel(),
-        config.getExtraLabels()
+        config.getExtraLabels(),
+        config.getFlushPeriod()
```
kfaraz: Nit: we could pass the entire `config` object to the constructor rather than these 6 args.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitterConfig.java
```diff
@@ -97,10 +97,31 @@ public PrometheusEmitterConfig(
...
+          throw DruidException.forPersona(DruidException.Persona.OPERATOR)
+                              .ofCategory(DruidException.Category.INVALID_INPUT)
+                              .build(
+                                  StringUtils.format(
+                                      "Invalid value for flushPeriod[%s] specified, flushPeriod must be > 0.",
+                                      flushPeriod
+                                  )
+                              );
+        }
+      }
```
kfaraz: Nit: seems like this part can be commoned out.

aho135:
Thanks for the review @abhishekrb19 @kfaraz I went ahead and made those updates

website/.spelling
aho135: I was just copying the current format. Ex: https://github.com/apache/druid/blob/master/website/.spelling#L2455

I can remove this if it's not needed though

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -55,4 +65,23 @@ public double[] getHistogramBuckets()
...
+  public long getMillisSinceLastUpdate()
+  {
+    return updateTimer.millisElapsed();
+  }
+
+  public boolean isExpired()
+  {
+    if (ttlSeconds == null) {
+      log.error("Invalid usage of isExpired(), TTL has not been set");
+      return false;
```
abhishekrb19, line 83: This condition should never trigger in production. if it does, it’s a bug, so we could `throw DruidException.defensive()` to indicate invalid usage for developers. Given it’s configured correctly now, this should be fine and we can update this later.

PR #18599, opened 2025-10-06
[WIP] Create a new KubernetesPeonClient that uses fabric8 informers to reduce load on an underlying k8s API

capistrant:
### Description

**WORK IN PROGRESS**

#### Release note

<hr>

<hr>

This PR has:

- [ ] been self-reviewed.


kfaraz:
Thanks for putting this up, @capistrant !
Will try to do an initial review before the end of this week.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesClient.java
```diff
@@ -53,4 +106,154 @@ public KubernetesClient getClient()
...
+        String jobName = pod.getMetadata().getLabels().get("job-name");
+        if (jobName != null) {
+          return Collections.singletonList(jobName);
+        }
+      }
+      return Collections.emptyList();
+    };
+
+    Map<String, Function<Pod, List<String>>> customPodIndexers = new HashMap<>();
+    customPodIndexers.put("byJobName", jobNameIndexer);
```
capistrant: indexer name should be a constant. same as below

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunner.java
```diff
@@ -347,6 +347,7 @@ public void start()
     log.info("Starting K8sTaskRunner...");
     // Load tasks from previously running jobs and wait for their statuses to start running.
     final List<ListenableFuture<Boolean>> taskStatusActiveList = new ArrayList<>();
+    // Get all existing peon jobs.
```
capistrant: probably not needed.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesInformerExecutor.java
capistrant: javadoc

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/AbstractKubernetesPeonClient.java
```diff
@@ -0,0 +1,346 @@
...
+        log.info("K8s job does not exist: %s", taskId);
+      }
+      return result;
+    } else {
+      log.info("Not cleaning up job %s due to flag: debugJobs=true", taskId);
+      return true;
+    }
+  }
+
+  public Optional<LogWatch> getPeonLogWatcher(K8sTaskId taskId)
```
capistrant: javadoc

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/AbstractKubernetesPeonClient.java
```diff
@@ -0,0 +1,346 @@
...
+      }
+      return Optional.of(logStream);
+    }
+    catch (Exception e) {
+      log.error(e, "Error streaming logs from task: %s, pod: %s", taskId, podName);
+      return Optional.absent();
+    }
+  }
+
+  public int deleteCompletedPeonJobsOlderThan(long howFarBack, TimeUnit timeUnit)
```
capistrant: javadoc

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/AbstractKubernetesPeonClient.java
```diff
@@ -0,0 +1,346 @@
...
+   * fail immediately.
+   *
+   * @param client     the Kubernetes client to use for job creation
+   * @param job        the Kubernetes job to create
+   * @param quietTries number of initial retry attempts without logging warnings
+   * @param maxTries   maximum total number of retry attempts
+   * @throws DruidException if job creation fails after all retry attempts or encounters non-retryable errors
+   */
+  @VisibleForTesting
+  void createK8sJobWithRetries(KubernetesClient client, Job job, int quietTries, int maxTries)
```
capistrant: can this be private and call wrapper instead in tests?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,267 @@
...
+            log.info("Pod[%s] for job[%s] is running or complete", podName, jobName);
+            return pod;
+          } else {
+            log.debug("Pod[%s] for job[%s] exists but not ready yet", podName, jobName);
+          }
+        } else {
+          log.info("Pod for job[%s] not created yet", jobName);
+        }
+
+        // We wake up every informer resync period to avoid event notifier misses.
```
capistrant: might need an even more detailed comment here

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+  }
+
+  /**
+   * Check if the pod is in Running, Succeeded or Failed phase.
+   */
+  private boolean isPodRunningOrComplete(Pod pod)
+  {
+    // I could not find constants for Pod phases in fabric8, so hardcoding them here.
+    // They are documented here: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
```
capistrant: maybe put it in the javadoc in a link?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DirectKubernetesPeonClient.java
capistrant, line 1: This is a copy of all the read code from the legacy KubernetesPeonClient. There may be optimizations to be made for this mode, but I consider that out of scope for this PR.

extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/overlord/taskadapter/DruidPeonClientIntegrationTest.java
capistrant, line 1: I haven't read through the test cases closely, but this seems to be candidate for Docker test (or deletion if the docker tests cover it already). I'll leave for a follow on PR as it feels out of scope.

capistrant:
> Thanks for putting this up, @capistrant ! Will try to do an initial review before the end of this week.

thanks Kashif! sorry that I had a bunch of commits after you started reviewing. I ended up self reviewing last week and there were some places that clearly needed cleaning up

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.java
```diff
@@ -243,7 +242,7 @@ public void test_runIndexParallelTask_andCompactData()
         .dynamicPartitionWithMaxRows(5000)
         .withId(compactTaskId);
     cluster.callApi().onLeaderOverlord(o -> o.runTask(compactTaskId, compactionTask));
-    cluster.callApi().waitForTaskToSucceed(taskId, eventCollector.latchableEmitter());
+    cluster.callApi().waitForTaskToSucceed(compactTaskId, eventCollector.latchableEmitter());
```
kfaraz, line 242: Good catch!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.java
```diff
@@ -308,13 +307,11 @@ public void test_runKafkaSupervisor()
...
 
-    // Get the task statuses
-    List<TaskStatusPlus> taskStatuses = ImmutableList.copyOf(
-        (CloseableIterator<TaskStatusPlus>)
-            cluster.callApi().onLeaderOverlord(o -> o.taskStatuses(null, dataSource, 1))
-    );
-    Assertions.assertFalse(taskStatuses.isEmpty());
-    Assertions.assertEquals(TaskState.RUNNING, taskStatuses.get(0).getStatusCode());
+    // Confirm tasks are being created and running
+    // This more forgiving assertion avoids weird race conditions with super specific checks like "exactly 1 running task"
```
kfaraz: Yes, this is cleaner! I think we can remove this comment though.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.java
```diff
@@ -408,6 +405,17 @@ private void waitForSegmentsToBeQueryable(int numSegments)
     );
   }
 
+  /**
+   * Gets the count of tasks with the given status for the specified datasource.
+   */
+  private int getTaskCount(String status, String dataSource)
```
kfaraz: Consider moving this utility method to `EmbeddedClusterApis` itself so that it may be used by other tests too.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerFactory.java
```diff
@@ -70,22 +72,44 @@ public KubernetesTaskRunnerFactory(
...
+      if (useOverlordNamespace) {
+        peonClient = new CachingKubernetesPeonClient(
+            druidKubernetesClient,
+            kubernetesTaskRunnerConfig.getNamespace(),
+            kubernetesTaskRunnerConfig.getOverlordNamespace(),
+            kubernetesTaskRunnerConfig.isDebugJobs(),
+            emitter
+        );
+      } else {
+        peonClient = new CachingKubernetesPeonClient(
```
kfaraz: For brevity, let's just use a single constructor and pass the overlord namespace as null when `useOverlordNamespace` is `false`. The constructor can then default it to empty string or leave null, whichever is appropriate.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerConfig.java
```diff
@@ -80,6 +80,13 @@ public class KubernetesTaskRunnerConfig
   // disable using http / https proxy environment variables
   private boolean disableClientProxy;
 
+  @JsonProperty
+  // enable using kubernetes informer cache for peon client operations
+  private boolean enableKubernetesClientSharedInformers = false;
```
kfaraz: Maybe use a simpler name
```suggestion
  private boolean useK8sSharedInformers = false;
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerConfig.java
```diff
@@ -80,6 +80,13 @@ public class KubernetesTaskRunnerConfig
...
+
+  @JsonProperty
+  private Period kubernetesClientInformerResyncPeriod = new Period("PT5M");
```
kfaraz:
```suggestion
  private Period k8sSharedInformerResyncPeriod = new Period("PT5M");
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesClient.java
```diff
@@ -19,21 +19,64 @@
...
+import org.apache.druid.java.util.emitter.EmittingLogger;
 import org.apache.druid.k8s.overlord.common.httpclient.DruidKubernetesHttpClientFactory;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
 public class DruidKubernetesClient implements KubernetesClientApi
```
kfaraz, line 27: All the changes in this class (and the corresponding APIs in `KubernetesClientApi`) pertain only to the caching implementation.

I would suggest skipping the changes made to `KubernetesClientApi` and `DruidKubernetesClient` and move all the new logic into a separate class altogether and use it in `CachingKubernetesPeonClient` directly. The non-caching impl should not be able to invoke these methods at all (rather than initializing `podInformer` etc to null and throw exceptions if invoked).

The new class can be a wrapper on `DruidKubernetesClient` and may use `DruidKubernetesClient.getClient()`.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DirectKubernetesPeonClient.java
kfaraz: Yes, let's postpone the refactors/optimizations for follow up PRs.

It would also be helpful to get rid of `AbstractKubernetesClient` and retain `KubernetesPeonClient` same as before. The new caching-based impl may extend `KubernetesPeonClient` and override whatever is needed.
Let's make only the necessary minimal changes to `KubernetesPeonClient` in this patch.

This would help focus this PR only on the new stuff, avoid copy paste errors and greatly simplify reviews.
I was trying it out here https://github.com/kfaraz/druid/pull/356. You may cherrypick the last commits in that branch if you find them to be useful.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,284 @@
...
+        // This prevents the case where we miss a notification and wait forever.
+        log.debug("Timeout waiting for job change notification for job[%s]. If full job timeout has not been reached, the job completion wait will continue", taskId.getK8sJobName());
+      }
+      catch (InterruptedException e) {
+        throw DruidException.defensive(e, "Interrupted waiting for job change notification for job[%s]", taskId.getK8sJobName());
+      }
+      catch (Throwable e) {
+        log.warn("Exception[%s] waiting for job change notification for job[%s]. Error message[%s]", e.getClass().getName(), taskId.getK8sJobName(), e.getMessage());
+      }
+    } while ((System.currentTimeMillis() - startTime < timeoutMs) && (jobSeenInCache || System.currentTimeMillis() < jobMustBeSeenBy));
```
kfaraz: Consider using a Druid `Stopwatch` instead.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,284 @@
...
+      catch (TimeoutException e) {
+        // A timeout here is not a problem, it forces us to loop around and check the cache again.
+        // This prevents the case where we miss a notification and wait forever.
+        log.debug("Timeout waiting for job change notification for job[%s]. If full job timeout has not been reached, the job completion wait will continue", taskId.getK8sJobName());
+      }
+      catch (InterruptedException e) {
+        throw DruidException.defensive(e, "Interrupted waiting for job change notification for job[%s]", taskId.getK8sJobName());
+      }
+      catch (Throwable e) {
+        log.warn("Exception[%s] waiting for job change notification for job[%s]. Error message[%s]", e.getClass().getName(), taskId.getK8sJobName(), e.getMessage());
```
kfaraz:
```suggestion
        log.noStackTrace().warn(e, "Exception while waiting for change notification of job[%s]", taskId.getK8sJobName());
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,284 @@
...
+        if (currentResponse.getPhase() != PeonPhase.RUNNING) {
+          return currentResponse;
+        } else {
+          log.debug("K8s job[%s] is still running", taskId.getK8sJobName());
+        }
+      }
+      catch (TimeoutException e) {
+        // A timeout here is not a problem, it forces us to loop around and check the cache again.
+        // This prevents the case where we miss a notification and wait forever.
+        log.debug("Timeout waiting for job change notification for job[%s]. If full job timeout has not been reached, the job completion wait will continue", taskId.getK8sJobName());
```
kfaraz:
```suggestion
        log.debug("Timeout waiting for change notification of job[%s]. Waiting until full job timeout.", taskId.getK8sJobName());
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,284 @@
...
+          // Job was in cache before, but now it's gone - it was deleted and will never complete.
+          log.warn("K8s Job[%s] was not found. It can happen if the task was canceled", taskId.getK8sJobName());
+          return new JobResponse(null, PeonPhase.FAILED);
+        }
+
+        // We wake up every informer resync period to avoid event notifier misses.
+        Job job = jobFuture.get(clientApi.getInformerResyncPeriodMillis(), TimeUnit.MILLISECONDS);
+
+        // Immediately set up to watch for the next change in case we need to wait again
+        jobFuture = clientApi.getEventNotifier().waitForJobChange(taskId.getK8sJobName());
```
kfaraz: Can there be a race between these two lines causing us to miss out on notifications?
~How about we create the next future before invoking `.get()` on the previous one?~

I suppose the new future will just timeout in that case.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesClientApi.java
```diff
@@ -19,15 +19,30 @@
...
+  <T> T executeJobCacheRequest(KubernetesInformerExecutor<T, Job> executor);
+
   // use only when handling streams of data, example if you want to pass around an input stream from a pod,
   // then you would call this instead of executeRequest as you would want to keep the connection open until you
   // are done with the stream.  Callers responsibility to clean up when using this method
   KubernetesClient getClient();
+
+  SharedIndexInformer<Pod> getPodInformer();
+
+  SharedIndexInformer<Job> getJobInformer();
```
kfaraz: I guess these methods don't need to be exposed, as they are not being used anywhere.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesClientApi.java
```diff
@@ -19,15 +19,30 @@
...
+
+  long getInformerResyncPeriodMillis();
+
+  KubernetesResourceEventNotifier getEventNotifier();
```
kfaraz: Instead of returning the event notifier itself, we should expose the wait methods, `waitForJobChange()`, `waitForPodChange()`.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesClientApi.java
```diff
@@ -19,15 +19,30 @@
...
+import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
 
 // Wraps all kubernetes api calls, to ensure you open and close connections properly
 public interface KubernetesClientApi
 {
   <T> T executeRequest(KubernetesExecutor<T> executor) throws KubernetesResourceNotFoundException;
 
+  <T> T executePodCacheRequest(KubernetesInformerExecutor<T, Pod> executor);
+
+  <T> T executeJobCacheRequest(KubernetesInformerExecutor<T, Job> executor);
```
kfaraz: Nit: Should we simplify these names to `readPodCache` / `getFromPodCache` or similar?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesInformerExecutor.java
```diff
@@ -0,0 +1,28 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.k8s.overlord.common;
+
+import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
+
+@FunctionalInterface
+public interface KubernetesInformerExecutor<T, R>
```
kfaraz: Maybe rename to `SharedInformerCacheReadRequest` to be more specific?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/SharedInformerCacheReadRequestExecutor.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.apache.druid.k8s.overlord.common;
+
+import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
+
+@FunctionalInterface
+public interface KubernetesInformerExecutor<T, R>
+{
+  T executeRequest(SharedIndexInformer<R> informer);
```
kfaraz: Should this function be applied on the `Indexer` instead of the `SharedIndexInformer`?

extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerFactoryTest.java
```diff
@@ -57,8 +58,14 @@ public void setup()
         .withCapacity(1)
         .build();
     taskLogs = new NoopTaskLogs();
+
+    Config config = new ConfigBuilder().build();
+    config.setAdditionalProperty(
+        DruidKubernetesClient.ENABLE_INFORMERS_KEY,
+        kubernetesTaskRunnerConfig.isEnableKubernetesClientSharedInformers());
```
kfaraz: Nit: style
```suggestion
        kubernetesTaskRunnerConfig.isEnableKubernetesClientSharedInformers()
    );
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesClient.java
```diff
@@ -53,4 +115,176 @@ public KubernetesClient getClient()
...
+  private SharedIndexInformer<Job> setupJobInformer(String namespace)
+  {
+    SharedIndexInformer<Job> jobInformer =
+        kubernetesClient.batch()
+                        .v1()
+                        .jobs()
+                        .inNamespace(namespace)
+                        .withLabel(DruidK8sConstants.LABEL_KEY)
+                        .inform(
+                            new ResourceEventHandler<>()
```
kfaraz: The `ResourceEventHandler` impls seem to take up too many lines.

Let's create a concrete impl, say `InformerEventHandler<T>` which just accepts a single lambda `BiConsumer<T, EventType>` and invokes that lambda in all of `onAdd`, `onUpdate`, `onDelete` with the correct event type.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesResourceEventNotifier.java
```diff
@@ -0,0 +1,118 @@
...
+ * Allows tasks to wait for specific resource changes without polling, improving efficiency and responsiveness.
+ * Crtical component of {@link CachingKubernetesPeonClient} functionality.
+ * </p>
+ */
+public class KubernetesResourceEventNotifier
+{
+  private static final EmittingLogger log = new EmittingLogger(KubernetesResourceEventNotifier.class);
+
+  private final ConcurrentHashMap<String, List<CompletableFuture<Job>>> jobWatchers = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<String, List<CompletableFuture<Pod>>> podWatchers = new ConcurrentHashMap<>();
```
kfaraz: Why do we need to keep a list of futures? Can we keep a single future and give it out to anyone wanting to wait on it?
Once a notification is received, we atomically resolve and remove this future from the map.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesResourceEventNotifier.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param jobName The job-name label value that changed
+   */
+  public void notifyPodChange(String jobName, Pod pod)
+  {
+    List<CompletableFuture<Pod>> futures = podWatchers.get(jobName);
+    if (futures != null && !futures.isEmpty()) {
+      log.debug("Notifying %d watchers of pod change for job-name [%s]", futures.size(), jobName);
+      futures.forEach(f -> f.complete(pod));
+      futures.clear();
+    }
```
kfaraz: I think there are some race conditions possible here.
e.g. After we do `futures.forEach()`, a new future gets added to the list and it gets cleared without ever being resolved.

OR after we evaluate the `if (futures is not null or empty)`, a new future gets added and we never resolve it.
(I suppose such futures will eventually timeout in the `CachingKubernetesPeonClient`.)

For better thread safety, you could use `ConcurrentHashMap.compute()` and perform the entire operation atomically inside the `compute()`. We use a similar technique in `TaskQueue`.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,284 @@
...
+          JobResponse currentResponse = determineJobResponse(job);
+          if (currentResponse.getPhase() != PeonPhase.RUNNING) {
+            return currentResponse;
+          } else {
+            log.debug("K8s job[%s] found in cache and is still running", taskId.getK8sJobName());
+          }
+        } else if (jobSeenInCache) {
+          // Job was in cache before, but now it's gone - it was deleted and will never complete.
+          log.warn("K8s Job[%s] was not found. It can happen if the task was canceled", taskId.getK8sJobName());
+          return new JobResponse(null, PeonPhase.FAILED);
```
kfaraz, line 95: How long do successfully completed jobs remain in cache?
Can we tune it using some K8s config and should we expose that as a Druid config?

kfaraz, line 95: Note for follow up: We might consider adding an error message to the `JobResponse` to denote the reason for failure, e.g. `Job is not present in cache anymore` or something.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesResourceEventNotifier.java
```diff
@@ -0,0 +1,118 @@
...
+   *
+   * @param jobName The job-name label value that changed
+   */
+  public void notifyPodChange(String jobName, Pod pod)
+  {
+    List<CompletableFuture<Pod>> futures = podWatchers.get(jobName);
+    if (futures != null && !futures.isEmpty()) {
+      log.debug("Notifying %d watchers of pod change for job-name [%s]", futures.size(), jobName);
+      futures.forEach(f -> f.complete(pod));
+      futures.clear();
```
kfaraz: Should we completely remove the entry for this job name from the map? Otherwise, it will keep growing with empty entries for stale jobs.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesClient.java
```diff
@@ -19,21 +19,64 @@
...
+  public static final String JOB_NAME_INDEX = "byJobName";
+  public static final String OVERLORD_NAMESPACE_INDEX = "byOverlordNamespace";
+
+  public static final String ENABLE_INFORMERS_KEY = "druid.k8s.informers.enabled";
+  public static final String INFORMER_RESYNC_PERIOD_MS_KEY = "druid.k8s.informers.resyncPeriodMs";
+  private static final long DEFAULT_INFORMER_RESYNC_PERIOD_MS = 300000L; // 5 minutes
+
   private final KubernetesClient kubernetesClient;
+  private final SharedIndexInformer<Pod> podInformer;
+  private final SharedIndexInformer<Job> jobInformer;
```
kfaraz: Can we add a `stop()` method that invokes stop on the informers to clean up any resource usage?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        // We wake up every informer resync period to avoid event notifier misses.
+        Job job = jobFuture.get(clientApi.getInformerResyncPeriodMillis(), TimeUnit.MILLISECONDS);
+
+        // Immediately set up to watch for the next change in case we need to wait again
+        jobFuture = clientApi.getEventNotifier().waitForJobChange(taskId.getK8sJobName());
+        log.debug("Received job[%s] change notification", taskId.getK8sJobName());
+        jobSeenInCache = true;
```
kfaraz: Is this correct?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,284 @@
...
+
+        if (job == null) {
```
kfaraz: Should this check be done right after line 106 (`Job job = jobFuture.get()`)?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,284 @@
...
+
+        JobResponse currentResponse = determineJobResponse(job);
+        if (currentResponse.getPhase() != PeonPhase.RUNNING) {
+          return currentResponse;
+        } else {
+          log.debug("K8s job[%s] is still running", taskId.getK8sJobName());
+        }
+      }
+      catch (TimeoutException e) {
+        // A timeout here is not a problem, it forces us to loop around and check the cache again.
```
kfaraz: If the future times out, do we need to remove it from the event notifier map proactively?
I think it would get cleaned up only if we receive an event for that job/pod. Otherwise, it will always remain in memory.

I am not sure if it would be safe to just de-register using a `stopWatchingJobChange()` method, as it would essentially mark (all) the futures for that job as cancelled, thus causing exception in any other thread waiting for the future to get resolved.

I guess we could handle that case by catching a `CancellationException` and just looping back.

@capistrant , what do you think?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/SharedInformerCacheReadRequestExecutor.java
capistrant: hmmm interesting point. I don't think I knew up front that I'd only ever use the informer in a way that I always get an indexer up front and then work with it. It is reasonable to use an informer without querying with an index though.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesResourceEventNotifier.java
capistrant: heh... so I can debug log how many there are 🧌 I really don't remember, and can't find a reason for it. Thank you for calling it out!

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesResourceEventNotifier.java
capistrant: I think the changes to stop storing as a list should make this more safe. Let me know if I missed anything. Thanks for noticing this!

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesResourceEventNotifier.java
capistrant: yes, this was a big oversight

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
capistrant: Ya there is definitely a problem here. I am not positive on how to fix it yet. I am working on one angle that uses an atomic counter associated with each future so something that is not referenced anymore can be cleaned up proactively. But I'm not so sure that is going to work either. I'm going to keep thinking about this tonight and try to come up with a legitimate fix. The pattern just might be too brittle. Will also think about your idea of tolerating a cancelation exception and looping back around.

capistrant:
@kfaraz thank you for the detailed review. I pushed up changes that I made up until now. I am still trying to solve for the issue regarding leaving dangling futures in exceptional cases. Will hopefully have a proposal for fixing that in the next 24 hours. Thanks again, appreciate all your insight on how to improve this.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
kfaraz: Thanks, @capistrant!

The reference counting approach could be a viable option. `HeapMemoryDatasourceSegmentCache` does something similar.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/SharedInformerCacheReadRequestExecutor.java
kfaraz: Yeah, but it doesn't seem like we are using the informer itself in any of the invocations right now. The indexer is better to use since it refers to the actual cache, and doesn't allow the caller to perform any other operation on the informer.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
capistrant: upon further review, I think I'm trying to solve a problem that does not exist (at least not yet). That problem is, what to do if multiple "things" can be waiting on the same pod or job at the same time. But in reality if you look at the code in this extension, it does not work like that right now. The only thing ever setting up these waits is the owning task. and it would never create more than 1 wait within the task at the same time. I feel like removing this complexity makes the code a lot easier to reason about and will work until such a time that more than one thing need to register waits for the same job. What do you think?

kfaraz: Yeah, if we can be sure that only one thread is going to perform this code at a time (by putting this behind a taskId gate), we can get rid of this complexity.

capistrant: hmm, I guess I didn't use the task id as the key for managing the waits in my latest commits. I just use the job name still. I do think the job name is generated from the task id though https://github.com/apache/druid/blob/master/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesOverlordUtils.java#L49

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
capistrant: good question, I need to find this out and can report back

capistrant: claude says `druid.indexer.runner.taskCleanupDelay` is what manages when a task is cleaned up in the underlying task runner. And in our case of k8s here, that means the ttl on the underlying pod after it completes. So this config would drive how long until we purge the pods I guess. And that is when the delete event would fire. I think the case that this conditional is covering is more of a nasty case where someone force terminates the pod under the hood and it is removed from the k8s cluster

capistrant: but maybe folks using this mode would want to be alerted to the fact that they should probably not use the default of P2D cuz that would probably result in a lot of memory for successful tasks that have not been deleted from k8s yet.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesCachingClient.java
```diff
@@ -0,0 +1,280 @@
...
+    DELETE
+  }
+
+  /**
+   * Impl of ResourceEventHandler that simplifies event handling
+   * by accepting a single lambda BiConsumer for all event types (add, update, delete).
+   *
+   * @param <T> The Kubernetes resource type (e.g., Pod, Job)
+   */
+  public static class InformerEventHandler<T> implements ResourceEventHandler<T>
```
kfaraz: Nit: this class may be in a file of its own.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesCachingClient.java
```diff
@@ -0,0 +1,280 @@
...
+      eventConsumer.accept(newResource, InformerEventType.UPDATE);
+    }
+
+    @Override
+    public void onDelete(T resource, boolean deletedFinalStateUnknown)
+    {
+      eventConsumer.accept(resource, InformerEventType.DELETE);
+    }
+  }
+  private static final EmittingLogger log = new EmittingLogger(DruidKubernetesCachingClient.class);
```
kfaraz, line 37: Nit: please add a newline before this.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesCachingClient.java
```diff
@@ -0,0 +1,280 @@
...
+
+  public static final String JOB_NAME_INDEX = "byJobName";
+  public static final String OVERLORD_NAMESPACE_INDEX = "byOverlordNamespace";
+
+  private static final long DEFAULT_INFORMER_RESYNC_PERIOD_MS = 300000L; // 5 minutes
+
+  protected final SharedIndexInformer<Pod> podInformer;
+  protected final SharedIndexInformer<Job> jobInformer;
+  protected final KubernetesResourceEventNotifier eventNotifier;
```
kfaraz: Do these need to be `protected` or can they be `private` too?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesCachingClient.java
```diff
@@ -0,0 +1,280 @@
...
+  {
+    this.baseClient = baseClient;
+    this.informerResyncPeriodMillis = informerResyncPeriodMillis;
+    this.eventNotifier = new KubernetesResourceEventNotifier();
+
+    this.podInformer = setupPodInformer(namespace);
+    this.jobInformer = setupJobInformer(namespace);
+  }
+
+  public void stop()
```
kfaraz, line 65: Please add a short javadoc. Should this be called upon loss of leadership too or only service termination?

Should there also be an equivalent `start()` that is invoked on becoming leader?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesCachingClient.java
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  public KubernetesClient getClient()
+  {
+    return baseClient.getClient();
+  }
+
+  public <T> T readPodCache(SharedInformerCacheReadRequestExecutor<T, Pod> executor)
+  {
+    if (podInformer == null) {
```
kfaraz: `podInformer` and `jobInformer` will never be null.

extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/overlord/common/TestCachingKubernetesClient.java
```diff
@@ -0,0 +1,42 @@
...
+public class TestCachingKubernetesClient extends DruidKubernetesCachingClient
+{
+  private static final long TESTING_RESYNC_PERIOD_MS = 10L;
+
+  public TestCachingKubernetesClient(KubernetesClientApi clientApi, String namespace)
+  {
+    super(clientApi, namespace, TESTING_RESYNC_PERIOD_MS);
+  }
+
+  public void start()
```
kfaraz: Should this be in the super class itself? When do we call `podInformer.run()` for the super class?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/SharedInformerCacheReadRequestExecutor.java
capistrant: sure, makes sense. I will push a commit with this

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesCachingClient.java
```diff
@@ -0,0 +1,280 @@
...
+    eventNotifier.cancelAll();
+  }
+
+  public KubernetesClientApi getBaseClient()
+  {
+    return baseClient;
+  }
+
+  // Delegate write operations to base client
+  public <T> T executeRequest(KubernetesExecutor<T> executor) throws KubernetesResourceNotFoundException
```
kfaraz: Not used anymore.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/DruidKubernetesCachingClient.java
```diff
@@ -0,0 +1,280 @@
...
+  {
+    return baseClient.executeRequest(executor);
+  }
+
+  public KubernetesClient getClient()
+  {
+    return baseClient.getClient();
+  }
+
+  public <T> T readPodCache(SharedInformerCacheReadRequestExecutor<T, Pod> executor)
```
kfaraz: Please add 1-line javadocs for these methods.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/SharedInformerCacheReadRequestExecutor.java
```diff
@@ -0,0 +1,28 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.k8s.overlord.common;
+
+import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
+
+@FunctionalInterface
+public interface SharedInformerCacheReadRequestExecutor<T, R>
```
kfaraz:
```suggestion
public interface SharedInformerCacheReader<T, R>
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/SharedInformerCacheReadRequestExecutor.java
```diff
@@ -0,0 +1,28 @@
...
+{
+  T executeRequest(SharedIndexInformer<R> informer);
```
kfaraz:
```suggestion
  T readFromCache(Indexer<R> cacheIndexer);
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesResourceEventNotifier.java
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * Cancel and remove a job watcher. Safe to call even if the future has already completed.
+   *
+   * @param jobName The name of the job to stop watching
+   */
+  public void cancelJobWatcher(String jobName)
+  {
+    CompletableFuture<Job> future = jobWatchers.remove(jobName);
+    if (future != null && !future.isDone()) {
+      log.debug("Cancelling watcher for job [%s]", jobName);
```
kfaraz: Nit: Druid style logs
```suggestion
      log.debug("Cancelling watcher for job[%s]", jobName);
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,250 @@
...
+    super(cachingClient.getBaseClient(), namespace, overlordNamespace == null ? "" : overlordNamespace, debugJobs, emitter);
+    this.cachingClient = cachingClient;
+  }
+
+  @Override
+  public JobResponse waitForPeonJobCompletion(K8sTaskId taskId, long howLong, TimeUnit unit)
+  {
+    Duration timeout = Duration.millis(unit.toMillis(howLong));
+    Duration jobMustBeSeenWithin = Duration.millis(cachingClient.getInformerResyncPeriodMillis() * 2);
+    Stopwatch stopwatch = Stopwatch.createStarted();
```
kfaraz: Nit: Make these `final`?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,250 @@
...
+    boolean jobSeenInCache = false;
+
+    try {
+      CompletableFuture<Job> jobFuture = cachingClient.waitForJobChange(taskId.getK8sJobName());
+      while (stopwatch.hasNotElapsed(timeout) && (jobSeenInCache || stopwatch.hasNotElapsed(jobMustBeSeenWithin))) {
+        if (jobFuture.isDone()) {
+          jobFuture = cachingClient.waitForJobChange(taskId.getK8sJobName());
```
kfaraz, line 80:
```suggestion
          // Register a future to watch the next change to this job
          jobFuture = cachingClient.waitForJobChange(taskId.getK8sJobName());
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
```diff
@@ -0,0 +1,250 @@
...
+          } else {
+            log.noStackTrace().warn(cause, "Exception while waiting for change notification of job[%s]", taskId.getK8sJobName());
+          }
+        }
+        catch (TimeoutException e) {
+          // No job change event notified within the timeout time. If there is more time, it will loop back and check the cache again.
+          log.debug("Timeout waiting for change notification of job[%s].", taskId.getK8sJobName());
+        }
+        catch (InterruptedException e) {
+          throw DruidException.defensive(e, "Interrupted waiting for job change notification for job[%s]", taskId.getK8sJobName());
```
kfaraz, line 116: This would cause a bunch of DruidException to be thrown when an Overlord is terminated, but I am not sure what the alternative should be.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesPeonClient.java
```diff
@@ -156,46 +172,76 @@ public boolean deletePeonJob(K8sTaskId taskId)
...
+                                   .inContainer("main")
+                                   .watchLog();
       if (logWatch == null) {
         return Optional.absent();
       }
       return Optional.of(logWatch);
     }
     catch (Exception e) {
-      log.error(e, "Error watching logs from task: %s", taskId);
+      log.error(e, "Error watching logs from task: %s, pod: %s", taskId, podName);
```
kfaraz:
```suggestion
      log.error(e, "Error watching logs from task[%s], pod[%s].", taskId, podName);
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesPeonClient.java
```diff
@@ -156,46 +172,76 @@ public boolean deletePeonJob(K8sTaskId taskId)
...
       return Optional.absent();
     }
   }
 
+  /**
+   * Get an InputStream for the logs of the peon pod associated with the given taskId.
+   * <p>
+   * Any issues creating the InputStream will be logged and an absent Optional will be returned.
+   * </p>
+   *
```
kfaraz, line 214: Nit: I think this part can be omitted since the same info is already captured in `@return` tag too.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/KubernetesPeonClient.java
```diff
@@ -156,46 +172,76 @@ public boolean deletePeonJob(K8sTaskId taskId)
...
+   * @return an Optional containing the {@link InputStream} if the pod exists and logs could be streamed, or absent otherwise
```
kfaraz:
```suggestion
   * @return an Optional containing the {@link InputStream} for the logs of the pod, if it exists and logs could be streamed, or absent otherwise.
```

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerConfig.java
```diff
@@ -232,6 +238,18 @@ public Builder withLogSaveTimeout(Period logSaveTimeout)
       return this;
     }
 
+    public Builder withEnablePeonClientCache(boolean enableKubernetesClientCaching)
```
kfaraz: Please align the method name and arg name with the config i.e. `useK8sSharedInformers`.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/KubernetesTaskRunnerConfig.java
```diff
@@ -72,6 +72,10 @@ public interface KubernetesTaskRunnerConfig
 
   Integer getCapacity();
 
+  boolean isUseK8sSharedInformers();
+
+  Period getK8sSharedInformerResyncPeriod();
```
kfaraz, line 92: Maybe add 1-line javadocs for these.

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/CachingKubernetesPeonClient.java
kfaraz: Yes, I guess the jobName is a good enough proxy for the task ID itself.

extensions-core/kubernetes-overlord-extensions/src/test/java/org/apache/druid/k8s/overlord/common/TestCachingKubernetesClient.java
capistrant: This is actually not needed. It was leftover cruft from older versions of this class where I was using special test informers. We now use the real informers from the super class so this can be removed. when you intitialize the informers with the inform() method like we do in CachingKubernetesClient, it automatically starts the informers.

capistrant: but I guess this does mean that we are probably eagerly starting informers for both the standby and leader overlord. Is this okay or do we need becoming leader code to be what creates the informers? it would slow down failovers since the initial sync has to list all pods to build the initial cache

capistrant:
> Thanks for the update, @capistrant ! The changes look good.
> 
> The blocking comments are only these:
> 
> * [#18599 (comment)](https://github.com/apache/druid/pull/18599#discussion_r2599314643) (is `DruidKubernetestCachingClient.stop()` called upon loss of leadership? Should we also have a `start()` which is called on becoming leader?)
> * [#18599 (comment)](https://github.com/apache/druid/pull/18599#discussion_r2599334679) (when do we call `podInformer.run()` in `DruidKubernetesCachingClient`?
> 
> The other comments are all minor suggestions.

Thanks @kfaraz for another review round. I think that these questions are both inter-mingled actually.

How everything works now, the Informers start automatically when `DruidKubernetesCachingClient` is created. This is happening as a part of the overlord startup, regardless of leadership. The `start()` calls on the informers in the tests were benign and I removed them.

The above information leads to your other question on when to call stop. Right now we only call stop as a part of the JVM lifecycle and not on loss of leadership. This is because as currently constructed the informers are started and exist for the lifecycle of the jvm so they are active regardless of if an OL is leader. 

I think the key question is, do we want the informer/caching client to be tied to the JVM lifecycle as they are now or tied to leadership state. doing the latter would reduce resource required by the standby in exchange for lengthening failover time (I'm not exactly sure on statistics here regarding how long failover would be extended by. It would scale with the pod count in k8s, but I don't have any good estimate on what that looks like for a k8s cluster with 500 pods, 5000 pods, 50000 pods, etc)

kfaraz:
Thanks for the clarification, @capistrant !

I think it is better to have the cache pre-warmed on the follower Overlord, so that it is able to quickly take over in case of a leader failure. So let's keep the lifecycle of the cache tied to the Overlord service itself.

Issue #17481, opened 2024-11-14
Projections


clintropolis:
### Motivation

Today Druid rollup tables can be utilized to greatly improve performance by significantly reducing the overall number of rows to be scanned by pre-aggregating the data at ingest time. However, this does come at the cost of forcing operators to make difficult decisions about which columns to include and at what granularity to perform rollup.

If the detailed table is still needed, then the only current way to possibly handle this is by having totally separate datasources with different ingestion configurations, which creates additional complexity for users having to know which datasources contain which columns at what level of rollup. This can quite quickly get complicated if multiple views are needed, and is a large burden to place on users to 'get it right' at query time, not to mention the operational overhead, and any potential differences in how 'up to date' each individual view is.

### Proposed changes

This proposal intends to introduce the concept of projections to Druid, which for the initial implementation are materialized transformations to pre-aggregate data contained within Druid segments as part of the same datasource. By building projections for commonly used queries, especially those having columns with lower value cardinality, we can significantly reduce the number of rows to be processed by those queries, dramatically lowering both compute and io cost. This gives us the ability to take advantage of all of the features of rollup tables, but still retain a fine level of detail when required, all within the same table in exchange for some extra storage space.

#### dataSchema JSON spec
The intial implementation is focused on the low level mechanics of building projections for realtime segments and persisting them to immutable segments with query time support for both states. The implication here is that at first, projections only can be defined by classic JSON based ingestion (but they can still be used by queries using MSQ or the new Dart engine). Future development will allow projections to be created as part of MSQ based ingestion as well.

This approach was chosen to prove the viability of the implementation before we design a more user friendly (and modern) way to define and manage projections, which will be detailed in a follow-up proposal once many of the lower level pieces are in place. Ultimately, MSQ based ingestion goes through the same code-paths to actually generate segments, and so the work done so far will be re-usable for SQL inserts and replaces once we finish designing and implementing the higher level interfaces for defining projections for a table.

To define a projection, a new `projections` section to the `dataSchema` of JSON based ingestion specs has been added. Since the initial implementation is focusing on grouping and pre-aggregating, the spec looks kind of similar to what you might find in the JSON spec for a group by query.

| field | description | required |
| --- | --- | --- |
| `type` | must be `'aggregate'` | yes |
| `name` | The internal name of the projection. This value is not typically needed to be known by users, but can optionally be specified with the `useProjection` context flag | yes |
[...]

PR #18612, opened 2025-10-09
Enhance Compaction task to be able to write to a different/new datasource

maytasm:
Enhance Compaction task to be able to write to a different/new datasource

### Description

Compaction tasks, unlike reindexing (which is native batch + druid input source), are great as they allow user to only specific specs that is needed and can automatically infer unspecified specs from existing segments of the input datasource. For example, if a user doesn't want to change the metrics, they don't have to set the metricsSpec and the Compaction task will correctly create metricsSpec from existing segments. 

A limitation of Compaction tasks is that the datasource the task is writing to and reading from has to be the same datasource. There are multiple use cases where we may want to write to a different datasource. For example:
1. You have an existing datasource A with all your data. You want to overwrite the datasource with a new ingestionSpec modifying x dimensions and y metrics. Before you do, you want to create a backup in case anything goes wrong. You may want to use a Compaction task to read from datasource A and write to datasource A_temp (exact copy without any change in specs) to test out the new ingestionSpec.
2. You have an existing datasource A with all your data. You may want to do a periodic backup (clone of the table at that state in time). You may want to use a Compaction task to read from datasource A and write to datasource A_20250429 (exact copy without any change in specs). This can be used for audit purpose or as a backup in case an accidental and/or malicious drop/overwrite/modification happen to the datasource A.
3. You have an existing datasource A with all your data. You have some queries that is not performant on this datasource due to the low rollup ratio (i.e. this datasource has a requirement for fine granularity and/or large number of dimensions). You may want to use a Compaction task to read from datasource A and write to datasource A_optimizedForSomeGroupBy (exact copy with change in dimensionSpec and/or granularitySpec) to power this query. In this case, you wouldn't need to specific the metricsSpec as you are not changing the metricsSpec.

This PR adds a new Property to the Compaction task `inputSpec`. The new field is `dataSource`. This new field is not required. When not set, the Compaction task will read and write to the same datasource given by the top-most level `dataSource` field. When set, the Compaction task will read from inputSpec's `dataSource` and write to the top-most level `dataSource` field. I think it make sense for this new field to be in the Compaction task `inputSpec` as the `inputSpec` purpose is specifying the input of the task. 

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/.idea/inspectionProfiles/Druid.xml
+++ b/.idea/inspectionProfiles/Druid.xml
@@ -39,9 +39,13 @@
       <option name="SUGGEST_NULLABLE_ANNOTATIONS" value="false" />
       <option name="DONT_REPORT_TRUE_ASSERT_STATEMENTS" value="true" />
     </inspection_tool>
+    <inspection_tool class="ConstantValue" enabled="true" level="WARNING" enabled_by_default="true">
+      <option name="DONT_REPORT_TRUE_ASSERT_STATEMENTS" value="true" />
+    </inspection_tool>
     <inspection_tool class="Contract" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="CopyConstructorMissesField" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="CovariantEquals" enabled="true" level="ERROR" enabled_by_default="true" />
+    <inspection_tool class="DataFlowIssueMerged" />
     <inspection_tool class="DuplicateBooleanBranch" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="DuplicateCondition" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="ignoreSideEffectConditions" value="true" />
@@ -59,6 +63,7 @@
     <inspection_tool class="EqualsUsesNonFinalVariable" enabled="true" level="WARNING" enabled_by_default="true" />
     <inspection_tool class="EqualsWhichDoesntCheckParameterClass" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="EqualsWithItself" enabled="true" level="ERROR" enabled_by_default="true" />
+    <inspection_tool class="Eslint" enabled="true" level="WARNING" enabled_by_default="true" />
     <inspection_tool class="FieldAccessNotGuarded" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="FieldCanBeLocal" enabled="true" level="WARNING" enabled_by_default="true">
       <option name="EXCLUDE_ANNOS">
@@ -70,7 +75,6 @@
       </option>
       <option name="IGNORE_FIELDS_USED_IN_MULTIPLE_METHODS" value="true" />
     </inspection_tool>
-    <inspection_tool class="FieldMayBeFinal" enabled="true" level="WARNING" enabled_by_default="true" />
     <inspection_tool class="FinalStaticMethod" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="FlowJSError" enabled="false" level="Non-TeamCity Error" enabled_by_default="false" />
     <inspection_tool class="ForCanBeForeach" enabled="true" level="WARNING" enabled_by_default="true">
@@ -82,7 +86,6 @@
       <option name="ignoreUntypedCollections" value="false" />
     </inspection_tool>
     <inspection_tool class="ForLoopThatDoesntUseLoopVariable" enabled="true" level="ERROR" enabled_by_default="true" />
-    <inspection_tool class="Guava" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="HashCodeUsesNonFinalVariable" enabled="true" level="WARNING" enabled_by_default="true" />
     <inspection_tool class="ImplicitArrayToString" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="IndexOfReplaceableByContains" enabled="true" level="ERROR" enabled_by_default="true" />
@@ -111,6 +114,24 @@
     <inspection_tool class="LengthOneStringInIndexOf" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="ListIndexOfReplaceableByContains" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="MalformedFormatString" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="classNames">
+        <list>
+          <option value="org.apache.druid.java.util.common.StringUtils" />
+          <option value="org.apache.druid.java.util.common.logger.Logger" />
+        </list>
+      </option>
+      <option name="methodNames">
+        <list>
+          <option value="trace" />
+          <option value="debug" />
+          <option value="info" />
+          <option value="warn" />
+          <option value="error" />
+          <option value="wtf" />
+          <option value="format" />
+          <option value="nonStrictFormat" />
+        </list>
+      </option>
       <option name="additionalClasses" value="org.apache.druid.java.util.common.StringUtils,org.apache.druid.java.util.common.logger.Logger" />
       <option name="additionalMethods" value="trace,debug,info,warn,error,wtf,format,nonStrictFormat" />
     </inspection_tool>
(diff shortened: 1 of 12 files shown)


.idea/inspectionProfiles/Druid.xml
jtuglu1, line 1: Let's avoid writing this if possible

kfaraz:
@maytasm , how is this different from the existing re-index capability that reads from a `DruidInputSource` and writes into a new datasource using an `index_parallel` task? Will it not achieve the same result?

`compact` by definition means compaction of the data in the same datasource.

If we want to be able to leverage the exact same capabilities as a `CompactionTask`, I would advise we create a new task type which extends the `CompactionTask` rather than update the existing class.
Alternatively, we can try to fill in the gaps in re-indexing using `index_parallel`.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -517,6 +520,7 @@ public TaskStatus runTask(TaskToolbox toolbox) throws Exception
     emitMetric(toolbox.getEmitter(), "ingest/count", 1);
 
     final Map<Interval, DataSchema> intervalDataSchemas = createDataSchemasForIntervals(
```
jtuglu1, line 522: Given we are now ingesting into potentially a new datasource, have we checked whether all associated tasks for ingestion (e.g. lock acquisition, segment alloc) are now working?

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -242,7 +242,10 @@ public CompactionTask(
     }
     this.projections = projections;
     this.tuningConfig = tuningConfig != null ? getTuningConfig(tuningConfig) : null;
-    this.segmentProvider = new SegmentProvider(dataSource, this.ioConfig.getInputSpec());
+    this.segmentProvider = new SegmentProvider(
+        this.ioConfig.getInputSpec().getDataSource() == null ? dataSource : this.ioConfig.getInputSpec().getDataSource(),
```
jtuglu1, line 246:
```suggestion
       Configs.getOrDefault(this.ioConfig.getInputSpec().getDataSource(), dataSource)
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionInputSpec.java
```diff
@@ -51,4 +51,9 @@ public interface CompactionInputSpec
    * @param latestSegments       most recent published segments in the interval returned by {@link #findInterval}
    */
   boolean validateSegments(LockGranularity lockGranularityInUse, List<DataSegment> latestSegments);
+
+  /**
+   * Return the datasource to be used as input to the compaction task.
```
jtuglu1, line 56:
```suggestion
   * Returns the name of the datasource to ingest the compacted segments to.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionTaskTest.java
```diff
@@ -87,7 +87,7 @@ public class CompactionTaskTest extends CompactionTestBase
       () -> TaskBuilder
           .ofTypeCompact()
           .context("storeCompactionState", true)
-          .ioConfig(new CompactionIntervalSpec(Intervals.of("2013-08-31/2013-09-02"), null), false);
+          .ioConfig(new CompactionIntervalSpec(Intervals.of("2013-08-31/2013-09-02"), null, null), false);
```
jtuglu1, line 90: Let's add tests where this isn't null

indexing-service/src/main/java/org/apache/druid/indexing/common/task/SpecificSegmentsSpec.java
```diff
@@ -61,6 +70,14 @@ public List<String> getSegments()
     return segments;
   }
 
+  @Override
+  @Nullable
+  @JsonProperty
+  public String getDataSource()
```
jtuglu1, line 76: nit: let's add some javadoc to this

jtuglu1:
> @maytasm , how is this different from the existing re-index capability that reads from a DruidInputSource and writes into a new datasource using an index_parallel task? Will it not achieve the same result?

@kfaraz I think the intent here is to take advantage of the schema detection, spec-autofill etc. I agree that this probably makes more sense in an `index_parallel` task with a `dataSource` input type.

kfaraz:
> I think the intent here is to take advantage of the schema detection, spec-autofill etc.

Yeah, that's a fair point. Compaction task does provide those useful capabilities out-of-the-box.

> I agree that this probably makes more sense in an index_parallel task with a dataSource input type.

👍🏻

maytasm:
@kfaraz I just added more detail to the PR description. Compaction task doesn't require you to specific all the specs (metricsSpec, etc) and can discover from existing segments. Specifying all the specs of a datasource is a big pain as Druid doesn't have the concept of schema/catalog and the schema can evolve/differs between segments. For example, the metricsSpec requires you to specify the CombiningAggregator instead of the original aggregators. All of this is done for you in the Compaction task
The Compaction task is just a layer that create index_parallel task with a dataSource input type. We can have a new task type but I think that's making Druid harder to use/understand...just like adding new runtime properties and tuning configs. You can already use Compaction task for more than Compacting. There are case where users change the schema, drop dimensions, change granularity level, etc. You can even give it a finer segmentGranularity and it will be expanding the datasource 
Re-indexing is really just native batch + druid input source. I don't think this is really a gap in reindexing as the native batch supports various other input source. The schema detection, spec-autofill etc here only make sense for only one of those input source, the Druid input source. For compaction task, the input source is always Druid but for native batch it isn't. It would be kinda weird to have native batch support not providing dimensionSpec, metricSpec, etc when the input source is Druid and required it when input source is not. Seems very specific/particular...

clintropolis:
>You have an existing datasource A with all your data. You have some queries that is not performant on this datasource due to the low rollup ratio (i.e. this datasource has a requirement for fine granularity and/or large number of dimensions). You may want to use a Compaction task to read from datasource A and write to datasource A_optimizedForSomeGroupBy (exact copy with change in dimensionSpec and/or granularitySpec) to power this query. In this case, you wouldn't need to specific the metricsSpec as you are not changing the metricsSpec.

drive by comment (still digesting the main idea of this PR), you should check out [projections](https://github.com/apache/druid/issues/17481) for this use case, which allow building aggregated views inside of segments the same datasource to automatically do more performant queries for any query which can match the projection. Projections can be applied to existing datasources via compaction since https://github.com/apache/druid/pull/17803. Projections are still experimental, and existing released versions have some bugs which will be fixed in the upcoming 35 release, where we plan to document them as part of that release since they should be a lot more stable.

kfaraz:
Yes, @maytasm , I agree that we should be able to leverage the auto-discover capabilities of compaction task for re-indexing too. For that reason, it makes sense to extend that feature.

> We can have a new task type but I think that's making Druid harder to use/understand...just like adding new runtime properties and tuning configs.

Yeah, more runtime properties and tuning configs often make Druid harder to use and understand.
In fact, my concern is the same. Overloading the same feature to satisfy completely different use cases
makes things confusing and less maintainable. The question is of intent. There is no reason a user
trying to move data from one DS to another should have to launch a `compact` task.

Since the use case here is a new capability altogether, there is no harm in adding a new task type or a new input source type, whichever seems simpler to implement.

> You can already use Compaction task for more than Compacting. There are case where users change the schema, drop dimensions, change granularity level, etc. You can even give it a finer segmentGranularity and it will be expanding the datasource

Absolutely, this is one of the things we have been discussing, that a `compact` task should ideally never change the meaning of the data, only how it's laid out/partitioned (the change in this PR would only add to that discrepancy).
Since we have already added this capability, we don't want to get rid of it right now as users may already be using it.
In the future, the compaction templates in #18402 will have capability to validate that a template does not change the
meaning of data and does only "compaction".

>  The schema detection, spec-autofill etc here only make sense for only one of those input source, the Druid input source. For compaction task, the input source is always Druid but for native batch it isn't.

Oh, absolutely. To clarify, I meant that we should try to bring the auto-detection capabilities of `compact` task
into native batch + `druid` input source, not native batch in general. But I suppose that might be more involved to implement, and perhaps an overkill anyway. We might as well just extend the `compact` task, as that seems simpler.

Also, to add to the suggestions from @clintropolis , you could also consider using an MSQ INSERT/REPLACE statement.
I don't know for sure if they provide all the auto-discover niceties or not but probably worth a shot.

maytasm:
> drive by comment (still digesting the main idea of this PR), you should check out https://github.com/apache/druid/issues/17481 for this use case, which allow building aggregated views inside of segments the same datasource to automatically do more performant queries for any query which can match the projection. Projections can be applied to existing datasources via compaction since https://github.com/apache/druid/pull/17803. Projections are still experimental, and existing released versions have some bugs which will be fixed in the upcoming 35 release, where we plan to document them as part of that release since they should be a lot more stable.

Ah. I forgot about Projection. Yep, you are right. Projection can achieve the same result as this point too. Although I think Projection doesn't allow you to have a different queryGranularity though

maytasm:
> Also, to add to the suggestions from @clintropolis , you could also consider using an MSQ INSERT/REPLACE statement.
> I don't know for sure if they provide all the auto-discover niceties or not but probably worth a shot.

I think this actually work and can achieve all the use cases in the PR description. Thanks for the suggestion.
I think it should just be as simple as
```
INSERT INTO datasource_2 
SELECT * FROM datasource
PARTITIONED BY hour
```

PR #18613, opened 2025-10-09
Add CPU and memory information to web console

GabrielCWT:
This PR adds 2 new fields to the sys.servers table which is displayed on the web console. These will be stored on `DiscoveryDruidNode`

1. Available Processors
2. Total Memory (Displayed in **binary** bytes)

Both fields represent the available resource on the machine. If the service is hosted in a container, it will represent the resource available to the container instead. (Container Aware)

The reason for adding this feature is so that we are able to know the cost of running each of the servers. This is useful when deploying on Kubernetes as we have each server running on an isolated pod. 

Interestingly, it seems that from Java 17, the `OPERATING_SYSTEM_MX_BEAN` is container aware which helps to simplify the implementation. It might be possible to simplify some legacy code which can be found in the `cgroups` folder i.e. `Cpu.java` and `Memory.java`. I have not looked into detail but I feel that it is something worth looking into.

https://developers.redhat.com/articles/2022/04/19/java-17-whats-new-openjdks-container-awareness#recent_changes_in_openjdk_s_container_awareness_code


docs/querying/sql-metadata-tables.md
```diff
@@ -237,6 +237,9 @@ Servers table lists all discovered servers in the cluster.
 |is_leader|BIGINT|1 if the server is currently the 'leader' (for services which have the concept of leadership), otherwise 0 if the server is not the leader, or null if the server type does not have the concept of leadership|
 |start_time|STRING|Timestamp in ISO8601 format when the server was announced in the cluster|
 |version|VARCHAR|Druid version running on the server|
+|available_processors|BIGINT|Total number of processors available to the server|
```
FrankChen021:
```suggestion
|available_processors|BIGINT|Total number of CPU processors available to the server|
```

server/src/main/java/org/apache/druid/discovery/DiscoveryDruidNode.java
```diff
@@ -96,6 +120,8 @@ private static DiscoveryDruidNode fromJson(
       @JsonProperty("nodeType") NodeRole nodeRole,
       @JsonProperty("services") Map<String, StringObjectPairList> rawServices,
       @JsonProperty("startTime") DateTime startTime,
+      @JsonProperty("availableProcessors") Integer availableProcessors,
```
FrankChen021, line 124: if the `availableProcessors` is NULL, how is it processed? will it result in some problems at web-console side?

the null case may happens during upgrade, like coordinator/overlord are generally updated after brokers.

docs/querying/sql-metadata-tables.md
GabrielCWT: Updated

server/src/main/java/org/apache/druid/discovery/DiscoveryDruidNode.java
GabrielCWT: It should be handled in the constructor 
```java
    // Happens if service is running older version of Druid
    this.availableProcessors = availableProcessors != null ? availableProcessors : UNKNOWN_VALUE;
    this.totalMemory = totalMemory != null ? totalMemory : UNKNOWN_VALUE;
 ```

File: server/src/main/java/org/apache/druid/discovery/DiscoveryDruidNode.java [hunk 3]
```diff
@@ -65,20 +69,36 @@
...
+   */
+  @VisibleForTesting
+  public DiscoveryDruidNode(
+      DruidNode druidNode,
+      NodeRole nodeRole,
+      Map<String, DruidService> services,
+      DateTime startTime
+  )
+  {
+    this(druidNode, nodeRole, services, startTime, JvmUtils.getRuntimeInfo().getAvailableProcessors(), JvmUtils.getTotalMemory());
```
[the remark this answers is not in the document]
GabrielCWT: Fixed

server/src/main/java/org/apache/druid/discovery/DiscoveryDruidNode.java
```diff
@@ -65,20 +69,36 @@
...
+  }
+
   public DiscoveryDruidNode(
       DruidNode druidNode,
       NodeRole nodeRole,
       Map<String, DruidService> services
   )
   {
-    this(druidNode, nodeRole, services, DateTimes.nowUtc());
+    this(druidNode, nodeRole, services, DateTimes.nowUtc(), JvmUtils.getRuntimeInfo().getAvailableProcessors(), JvmUtils.getTotalMemory());
```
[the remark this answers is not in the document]
GabrielCWT: Fixed

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -381,6 +382,8 @@
   final List<DataSegment> realtimeSegments = ImmutableList.of(segment2, segment4, segment5);
 
   private final DateTime startTime = DateTimes.nowUtc();
+  private final long availableProcessors = JvmUtils.getRuntimeInfo().getAvailableProcessors();
```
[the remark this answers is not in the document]
GabrielCWT: Fixed

server/src/main/java/org/apache/druid/discovery/DiscoveryDruidNode.java
```diff
@@ -50,10 +52,13 @@
 public class DiscoveryDruidNode
 {
   private static final Logger LOG = new Logger(DiscoveryDruidNode.class);
+  private static final Integer UNKNOWN_VALUE = -1;
```
FrankChen021:
```suggestion
  private static final int UNKNOWN_VALUE = -1;
```

GabrielCWT: Updated

PR #18618, opened 2025-10-09
Web console: fix + check doc links

vogievetsky:
Made the expression input multiline, bumped axios to move past vulnerable version, and made a script to check for any broken links and then fixed all the broken links.

New transform UI with multiline input

image

### DIFF
--- a/web-console/script/check-doc-links.mjs
+++ b/web-console/script/check-doc-links.mjs
@@ -21,14 +21,13 @@
 /* eslint-disable no-undef */
 
 import fs from 'fs';
-import https from 'https';
 import path from 'path';
 import { fileURLToPath } from 'url';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
-const DOCS_BASE_URL = 'https://druid.apache.org/docs/latest';
 const SRC_DIR = path.join(__dirname, '../src');
+const DOCS_DIR = path.join(__dirname, '../../docs');
 
 // Regex to find ${getLink('DOCS')}/path patterns
 const DOC_LINK_REGEX = /\$\{getLink\(['"]DOCS['"]\)\}\/([^\s`'"}\]]+)/g;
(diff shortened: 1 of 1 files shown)


web-console/script/check-doc-links.mjs
```diff
@@ -0,0 +1,144 @@
...
+/* eslint-disable no-undef */
+
+import fs from 'fs';
+import https from 'https';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const DOCS_BASE_URL = 'https://druid.apache.org/docs/latest';
```
clintropolis: this i guess would have problems if the docs are not published yet? is it painful to like fall back to checking the docs markdown stuff for existence?

web-console/src/druid-models/transform-spec/transform-spec.tsx
```diff
@@ -50,12 +50,13 @@ export const TRANSFORM_FIELDS: Field<Transform>[] = [
...
     type: 'string',
-    placeholder: '"foo" + "bar"',
+    multiline: true,
+    placeholder: `ex: strlen(concat("foo", '-bar'))`,
     required: true,
     info: (
       <>
         A valid Druid{' '}
-        <ExternalLink href={`${getLink('DOCS')}/misc/math-expr`}>expression</ExternalLink>.
+        <ExternalLink href={`${getLink('DOCS')}/querying/math-expr`}>expression</ExternalLink>.
```
clintropolis, line 59: there are redirect entries for both of these things, are redirects not working? oh, is it because they only redirect for 'latest' and not specific version links?

web-console/script/check-doc-links.mjs
vogievetsky: That's a good idea to check the markdown

web-console/src/druid-models/transform-spec/transform-spec.tsx
vogievetsky: I think the redirects might be busted (try https://druid.apache.org/docs/latest/misc/math-expr)

clintropolis: hmm, that seems not great, i wonder when that broke?

clintropolis: ah, it does seem to work if i put a .html at the end of it, which is what the entry in the redirects is, so it is at least somewhat working

clintropolis: it also does work when specifying arbitrary versions (i picked 30 and it did a redirect for me)

cecemei:
Hi @vogievetsky, I'm including this PR in druid35 if that's ok since it fixes a severe CVE.
```
axios  1.0.0 - 1.11.0
Severity: high
Axios is vulnerable to DoS attack through lack of data size check - https://github.com/advisories/GHSA-4hjh-wcwx-xvwj
```

PR #18631, opened 2025-10-13
Emit status code tag in (sqlQuery|query)/time metrics

jtuglu1:
This adds a `code` dimension to the `query/time` and `sqlQuery/time` metrics which reports the result status code of the query. This is more verbose/helpful than success true/false and allows you to do quick aggregations filtering on code.

#### Release note

Add `code` dimension to {`query/time`, `sqlQuery/time`} metric containing the query result status code.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/QueryLifecycle.java
+++ b/server/src/main/java/org/apache/druid/server/QueryLifecycle.java
@@ -418,6 +418,7 @@ public class QueryLifecycle
       statsMap.put("query/time", TimeUnit.NANOSECONDS.toMillis(queryTimeNs));
       statsMap.put("query/bytes", bytesWritten);
       statsMap.put("success", success);
+      statsMap.put(DruidMetrics.CODE, DruidMetrics.computeStatusCode(e));
 
       if (authenticationResult != null) {
         statsMap.put("identity", authenticationResult.getIdentity());


jtuglu1:
I found some gaps with no query/time metric emission for various failure scenarios (403 auth, sql query unexpected arg, etc.) during this PR which I think can be followed up in a subsequent change. 

This is for two reasons:
1. IMO these changes are logically separate and shouldn't be coupled with each other.
1. Reduce the diff to make it easier to backport this commit in case it doesn't make it into v35.

kfaraz:
@jtuglu1 , I had a few suggestions:
- While the HTTP status code of a query does seem reasonably useful, it still provides very limited information.
It only tells us whether a query failed with say 400 or 500 etc.
It would be much more useful to include a short failure description (similar to what we do for `task/run/time` metric)
or the failure Druid error code (the code inside the `DruidException` could be used for this).
- The dimension name `code` is very ambiguous. Please use `errorCode` or `httpCode` instead. 
- I assume we are not emitting this dimension if the query was successful?

jtuglu1:
Good questions.

> It only tells us whether a query failed with say 400 or 500 

This is by design. If operators want to know why their query failed, they should consult the query logs. At high volume, emitting a repeated long string doesn't provide that much value in the presence of logs IMO and can be prohibitive in terms of emission tput and volume.

> The dimension name code is very ambiguous. Please use errorCode or httpCode instead.

Sure, will change in a follow-up.

> I assume we are not emitting this dimension if the query was successful?

We are emitting 200. That's the point of the status metric. It's a small piece of information for operators to easily aggregate on, etc. and not have to worry about odd edge cases like whether it's success=false, then go check `errorCode`, etc.

processing/src/test/java/org/apache/druid/query/DruidMetricsTest.java
```diff
@@ -0,0 +1,48 @@
...
+  {
+    Assert.assertEquals(200, DruidMetrics.computeStatusCode(null));
+  }
+
+  @Test
+  public void testComputeStatusCode_allDruidExceptionCategories()
+  {
+    for (DruidException.Category cat : DruidException.Category.values()) {
+      Assert.assertEquals(
+          cat.getExpectedStatus(), DruidMetrics.computeStatusCode(
```
samarthjain, line 39: Does this exercise `DEFENSIVE` code as well?

jtuglu1: It'll hit that category yes, all enum values (including `DEFENSIVE`) are covered.

jtuglu1:
Follow-up: https://github.com/apache/druid/pull/18633

PR #18634, opened 2025-10-15
Adding new task distribution strategy that assigns tasks to workers based on supervisor affinity

Pankaj260100:
Fixes #XXXX.

### Description
- Previously, each datasource was managed by a single supervisor (1:1 mapping).
- [This update](https://github.com/apache/druid/pull/18082) allows multiple supervisors to be assigned to a single datasource.
- Introduced a new task distribution strategy that assigns tasks to different tiers of workers based on their associated supervisor.
- Enables more flexible and efficient task allocation across the system.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
@@ -110,10 +110,7 @@ public class WorkerCategorySpec
     }
 
     /**
-     * Returns a map of datasource names to worker category names.
-     * Used to assign tasks to specific worker categories based on their datasource.
-     *
-     * @return map where key is datasource name and value is worker category name
+     * Returns a map from datasource name to the worker category name to be used for tasks of that datasource.
      */
     @JsonProperty
     public Map<String, String> getCategoryAffinity()
@@ -122,11 +119,8 @@ public class WorkerCategorySpec
     }
 
     /**
-     * Returns a map of supervisor IDs to worker category names.
-     * Used to assign tasks to specific worker categories based on their supervisor ID.
+     * Returns a map from supervisor ID to worker category name to be used for tasks of that supervisor.
      * This takes precedence over {@link #getCategoryAffinity()} when both are configured.
-     *
-     * @return map where key is supervisor ID and value is worker category name
      */
     @JsonProperty
     public Map<String, String> getSupervisorIdCategoryAffinity()
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerSelectUtils.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerSelectUtils.java
@@ -138,7 +138,7 @@ public class WorkerSelectUtils
           preferredCategory = categoryAffinity.get(task.getDataSource());
         }
         
-        // If there is no preferred category for the supervisorId or datasource, then using the defaultCategory.
+        // If there is no preferred category for the supervisorId or datasource, then use the defaultCategory.
         // However, the defaultCategory may be null too, so we need to do one more null check (see below).
         preferredCategory = preferredCategory == null ? defaultCategory : preferredCategory;
 
--- a/indexing-service/src/test/java/org/apache/druid/indexing/overlord/setup/EqualDistributionWithCategorySpecWorkerSelectStrategyTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/overlord/setup/EqualDistributionWithCategorySpecWorkerSelectStrategyTest.java
@@ -305,11 +305,11 @@ public class EqualDistributionWithCategorySpecWorkerSelectStrategyTest
         supervisorId,
         null,
         DataSchema.builder()
-            .withDataSource(datasource)
-            .withTimestamp(new TimestampSpec(null, null, null))
-            .withDimensions(new DimensionsSpec(Collections.emptyList()))
-            .withGranularity(new ArbitraryGranularitySpec(new AllGranularity(), Collections.emptyList()))
-            .build(),
+          .withDataSource(datasource)
+          .withTimestamp(new TimestampSpec(null, null, null))
+          .withDimensions(new DimensionsSpec(Collections.emptyList()))
+          .withGranularity(new ArbitraryGranularitySpec(new AllGranularity(), Collections.emptyList()))
+          .build(),
         Mockito.mock(SeekableStreamIndexTaskTuningConfig.class),
         Mockito.mock(SeekableStreamIndexTaskIOConfig.class),
         null,


indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/EqualDistributionWithSupervisorCategorySpecWorkerSelectStrategy.java
```diff
@@ -0,0 +1,108 @@
...
+import com.google.common.collect.ImmutableMap;
+import org.apache.druid.common.config.Configs;
+import org.apache.druid.indexing.common.task.Task;
+import org.apache.druid.indexing.overlord.ImmutableWorkerInfo;
+import org.apache.druid.indexing.overlord.config.WorkerTaskRunnerConfig;
+
+import javax.annotation.Nullable;
+import java.util.Objects;
+
+public class EqualDistributionWithSupervisorCategorySpecWorkerSelectStrategy implements WorkerSelectStrategy
```
kfaraz: This strategy seems to be mostly a duplicate of the existing `equalDistributionWithCategorySpec`.

Instead of this, add a new map in `CategoryConfig` class : `Map<String, String> supervisorIdCategoryAffinity`.
We will first check in this map, otherwise fallback to the datasource check.
This approach is better since there is no ambiguity regarding whether the map is keyed by datasource name or supervisor id.
Also, the `fillCapacityWithCategorySpec` strategy will be able to leverage the changes.

jtuglu1: +1 agree here with the above.

Pankaj260100: Make sense, updated the PR.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
```diff
@@ -88,15 +88,19 @@ public static class CategoryConfig
...
     private final Map<String, String> categoryAffinity;
+    // key: supervisorId, value: category
+    private final Map<String, String> supervisorIdCategoryAffinity;
 
     @JsonCreator
     public CategoryConfig(
         @JsonProperty("defaultCategory") String defaultCategory,
-        @JsonProperty("categoryAffinity") Map<String, String> categoryAffinity
+        @JsonProperty("categoryAffinity") Map<String, String> categoryAffinity,
+        @JsonProperty("supervisorIdCategoryAffinity") Map<String, String> supervisorIdCategoryAffinity
```
kfaraz:
```suggestion
        @JsonProperty("supervisorIdCategoryAffinity") @Nullable Map<String, String> supervisorIdCategoryAffinity
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
```diff
@@ -88,15 +88,19 @@ public static class CategoryConfig
     private final String defaultCategory;
     // key: datasource, value: category
     private final Map<String, String> categoryAffinity;
+    // key: supervisorId, value: category
```
kfaraz: It would be more useful to move this comment into a javadoc for the getter of this field.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
```diff
@@ -88,15 +88,19 @@ public static class CategoryConfig
...
     public CategoryConfig(
         @JsonProperty("defaultCategory") String defaultCategory,
-        @JsonProperty("categoryAffinity") Map<String, String> categoryAffinity
+        @JsonProperty("categoryAffinity") Map<String, String> categoryAffinity,
+        @JsonProperty("supervisorIdCategoryAffinity") Map<String, String> supervisorIdCategoryAffinity
     )
     {
       this.defaultCategory = defaultCategory;
       this.categoryAffinity = categoryAffinity == null ? Collections.emptyMap() : categoryAffinity;
+      this.supervisorIdCategoryAffinity = supervisorIdCategoryAffinity == null ? Collections.emptyMap() : supervisorIdCategoryAffinity;
```
kfaraz: shorthand:
```suggestion
      this.supervisorIdCategoryAffinity = Configs.valueOrDefault(supervisorIdCategoryAffinity, Map.of());
```

indexing-service/src/test/java/org/apache/druid/indexing/overlord/setup/EqualDistributionWithCategorySpecWorkerSelectStrategyTest.java
```diff
@@ -187,4 +313,65 @@ private ImmutableWorkerInfo selectWorker(WorkerCategorySpec workerCategorySpec)
 
     return worker;
   }
+
+  /**
+   * Test implementation of SeekableStreamIndexTask for testing supervisor ID affinity
+   */
+  private static class TestSeekableStreamIndexTask extends SeekableStreamIndexTask<String, String, ByteEntity>
```
kfaraz: `SeekableStreamSupervisorStateTest` also has a similar test class. Please see if it is possible to use the same one here. We can move it out into a file of its own.

indexing-service/src/test/java/org/apache/druid/indexing/overlord/setup/EqualDistributionWithCategorySpecWorkerSelectStrategyTest.java
```diff
@@ -174,6 +198,108 @@ public void testStrongTierSpec()
...
+  public void testSupervisorIdCategoryAffinityFallbackToDefault()
+  {
+    // Test that it falls back to default category when neither supervisor ID nor datasource affinity is found
+    final WorkerCategorySpec workerCategorySpec = new WorkerCategorySpec(
+        ImmutableMap.of(
+            "test_seekable_stream",
+            new WorkerCategorySpec.CategoryConfig(
+                "c2",  // default category
+                ImmutableMap.of("ds2", "c1"),  // datasource affinity (different datasource)
+                ImmutableMap.of("supervisor2", "c1")  // supervisor ID affinity (different supervisor)
```
kfaraz: the comments here are not really needed.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
kfaraz: Alternatively, we can also just leave the value as null and not assign any default value to it.

Pankaj260100: I updated it to the shorthand expression shared above. If we change it to null, then we will also have to update WorkerSelectUtils to check for null before the .get() operation.

kfaraz: It's okay to handle the null, I wanted to avoid the empty map being serialized out in the payload unnecessarily. But this is fine too.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
```diff
@@ -105,12 +109,31 @@ public String getDefaultCategory()
       return defaultCategory;
     }
 
+    /**
+     * Returns a map of datasource names to worker category names.
+     * Used to assign tasks to specific worker categories based on their datasource.
+     *
+     * @return map where key is datasource name and value is worker category name
```
kfaraz: Nit: Not needed since the first line already covers this info.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
```diff
@@ -105,12 +109,31 @@ public String getDefaultCategory()
       return defaultCategory;
     }
 
+    /**
+     * Returns a map of datasource names to worker category names.
```
kfaraz:
```suggestion
     * Returns a map from datasource name to the worker category name to be used for tasks of that datasource.
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
```diff
@@ -105,12 +109,31 @@ public String getDefaultCategory()
...
+     * @return map where key is datasource name and value is worker category name
+     */
     @JsonProperty
     public Map<String, String> getCategoryAffinity()
     {
       return categoryAffinity;
     }
 
+    /**
+     * Returns a map of supervisor IDs to worker category names.
```
kfaraz:
```suggestion
     * Returns a map from supervisor ID to worker category name to be used for tasks of that supervisor.
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerCategorySpec.java
```diff
@@ -105,12 +109,31 @@ public String getDefaultCategory()
...
+     * Used to assign tasks to specific worker categories based on their supervisor ID.
+     * This takes precedence over {@link #getCategoryAffinity()} when both are configured.
+     *
+     * @return map where key is supervisor ID and value is worker category name
```
kfaraz: Nit: can be omitted.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/setup/WorkerSelectUtils.java
```diff
@@ -120,10 +121,25 @@ public static ImmutableWorkerInfo selectWorker(
...
+            preferredCategory = supervisorIdCategoryAffinity.get(supervisorId);
+          }
+        }
+        
+        // If no supervisor-based category is found, fall back to datasource-based category affinity
+        if (preferredCategory == null) {
+          preferredCategory = categoryAffinity.get(task.getDataSource());
+        }
+        
+        // If there is no preferred category for the supervisorId or datasource, then using the defaultCategory.
```
kfaraz:
```suggestion
        // If there is no preferred category for the supervisorId or datasource, then use the defaultCategory.
```

indexing-service/src/test/java/org/apache/druid/indexing/overlord/setup/EqualDistributionWithCategorySpecWorkerSelectStrategyTest.java
```diff
@@ -187,4 +293,27 @@ private ImmutableWorkerInfo selectWorker(WorkerCategorySpec workerCategorySpec)
...
+  {
+    return new TestSeekableStreamIndexTask(
+        id,
+        supervisorId,
+        null,
+        DataSchema.builder()
+            .withDataSource(datasource)
+            .withTimestamp(new TimestampSpec(null, null, null))
+            .withDimensions(new DimensionsSpec(Collections.emptyList()))
+            .withGranularity(new ArbitraryGranularitySpec(new AllGranularity(), Collections.emptyList()))
```
kfaraz: Nit: formatting seems off.

kfaraz (approved):
Minor nitpicks, changes look good.

PR #18637, opened 2025-10-15
wire lazyLoadOnStart back up to segment loader config

clintropolis:
### Description
This was inadvertently removed in the very large refactor of #18176. Also fixed some possible null pointer exceptions that could occur in certain cases - the one in `ServerManager` can happen for sure, but less sure if the ones in `SegmentLocalCacheManager` can actually happen, but went ahead and added the checks anyway.

The added test doesn't really confirm that lazy column deserialization is actually happening since its quite internal to the `QueryableIndex`, but it at least ensures that the path is being run in tests.

### DIFF
--- a/server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
+++ b/server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
@@ -639,9 +639,11 @@ public class SegmentLocalCacheManager implements SegmentCacheManager
         if (cacheEntry.checkExists(location.getPath())) {
           if (location.isReserved(cacheEntry.id) || location.reserve(cacheEntry)) {
             final SegmentCacheEntry entry = location.getCacheEntry(cacheEntry.id);
-            entry.lazyLoadCallback = segmentLoadFailCallback;
-            entry.mount(location);
-            return entry;
+            if (entry != null) {
+              entry.lazyLoadCallback = segmentLoadFailCallback;
+              entry.mount(location);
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -831,7 +835,9 @@ public void mount(StorageLocation mountLocation) throws SegmentLoadingException
           }
           final SegmentizerFactory factory = getSegmentFactory(storageDir);
 
-          final Segment segment = factory.factorize(dataSegment, storageDir, false, lazyLoadCallback);
+          @SuppressWarnings("ObjectEquality")
+          final boolean lazy = config.isLazyLoadOnStart() && lazyLoadCallback != SegmentLazyLoadFailCallback.NOOP;
```
GWphua, line 839: Previously the code does not seem to have `lazyLoadCallback != SegmentLazyLoadFailCallback.NOOP` condition... Is it guaranteed that lazyLoadCallback is not NOOP when `isLazyLoadOnStart()` is configured?

clintropolis: yea, bootstrap is always called with a callback set https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/server/coordination/SegmentCacheBootstrapper.java#L213

GWphua (approved):
LGTM, thanks @clintropolis

PR #17762, opened 2025-02-27
add ability to customize json (and auto) column physical storage details on per column basis

clintropolis:
### Description

changes:
* `NestedDataColumnSchema` and `AutoTypeColumnSchema` now accept a `NestedCommonFormatColumnFormatSpec` as a new 'columnFormatSpec' property, which allows customizing how json and auto columns are compressed, how strings are encoded in dictionaries, etc, on a per column level. The nested column format spec has several properties: `objectKeysEncoding`, `objectStorageEncoding`, `objectStorageCompression`, `stringDictionaryEncoding`, `dictionaryEncodedColumnCompression`, `longColumnEncoding`, `longColumnCompression`, `doubleColumnCompression`
* JSON field names can now be stored with front-coding with `columnFormatSpec` property `objectKeysEncoding`
*  `IndexSpec` has a new `autoColumnFormatSpec` to set default `NestedCommonFormatColumnFormatSpec`
* `IndexSpec` fields are now all nullable and NOT populated by default. A new `getEffectiveSpec()` method has been added to combine an `IndexSpec` with the system default `IndexSpec` and finally the hard-coded defaults if no override values have been specified at any level
* Adds 'druid.indexing.formats.indexSpec' to define a system default `IndexSpec`. This system default will be blended with the job `IndexSpec` at segment building time, before `NestedCommonFormatColumn` blends in segment level config from the computed `IndexSpec` to populate any values not explicitly specified with segment level

PR #18638, opened 2025-10-15
fix bug with NestedCommonFormatColumnFormatSpec default value resolution

clintropolis:
Follow up to #17762, fixes a bug that prevented job and system level auto column format specs from overriding `objectFieldsDictionaryEncoding` and `objectStorageEncoding` even if the user specified no format spec on the column itself

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/nested/NestedCommonFormatColumnFormatSpec.java
+++ b/processing/src/main/java/org/apache/druid/segment/nested/NestedCommonFormatColumnFormatSpec.java
@@ -40,8 +40,7 @@ import java.util.Objects;
  */
 public class NestedCommonFormatColumnFormatSpec
 {
-  private static final NestedCommonFormatColumnFormatSpec OPERATOR_DEFAULT = builder().build();
-  private static final NestedCommonFormatColumnFormatSpec SYSTEM_DEFAULT =
+  private static final NestedCommonFormatColumnFormatSpec DEFAULT =
       builder().setObjectFieldsDictionaryEncoding(StringEncodingStrategy.UTF8_STRATEGY)
                .setObjectStorageEncoding(ObjectStorageEncoding.SMILE)
                .build();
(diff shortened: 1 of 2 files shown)


processing/src/main/java/org/apache/druid/segment/nested/NestedCommonFormatColumnFormatSpec.java
```diff
@@ -40,11 +40,11 @@
  */
 public class NestedCommonFormatColumnFormatSpec
 {
-  private static final NestedCommonFormatColumnFormatSpec DEFAULT =
-      NestedCommonFormatColumnFormatSpec.builder()
-                                        .setObjectFieldsDictionaryEncoding(StringEncodingStrategy.UTF8_STRATEGY)
-                                        .setObjectStorageEncoding(ObjectStorageEncoding.SMILE)
-                                        .build();
+  private static final NestedCommonFormatColumnFormatSpec OPERATOR_DEFAULT = builder().build();
```
cecemei: I feel like we should try to explain all these defaults and add some documentations on the overriding orders of the spec. The `OPERATOR_DEFAULT` seems like a convenient holder instance so that we can call `getEffectiveSpec`, it doesn't represent anything meaningful, maybe we should move this to builder class, and let builder accept null spec in its constructor. 

Maybe the system default can be moved to indexspec? I see there're some duplicates in defining the default for ObjectStorageEncoding and ObjectFieldsDictionaryEncoding, i guess its due to indexspec can have all-null values column spec as well, so maybe it might be nice to centralize them all in one place, like `getObjectFieldsDictionaryEncodingOrDefault`.

clintropolis: >I feel like we should try to explain all these defaults and add some documentations on the overriding orders of the spec. The OPERATOR_DEFAULT seems like a convenient holder instance so that we can call getEffectiveSpec, it doesn't represent anything meaningful, maybe we should move this to builder class, and let builder accept null spec in its constructor.

Dropped the instance method so that we don't need `OPERATOR_DEFAULT` anymore which makes stuff a bit more straightforward i think, so the static method just makes a builder from the column spec if not null, or just an empty builder if null, then we fill in the defaults from the indexspec

>Maybe the system default can be moved to indexspec? I see there're some duplicates in defining the default for ObjectStorageEncoding and ObjectFieldsDictionaryEncoding, i guess its due to indexspec can have all-null values column spec as well, so maybe it might be nice to centralize them all in one place, like getObjectFieldsDictionaryEncodingOrDefault.

The default in `IndexSpec` is null (since old stored indexSpec would not have this value set and we can handle everything consistently) so it doesn't make a lot of sense to me to define there since we would need to reference it from here. Also many of the defaults in `IndexSpec` are sourced from the property type, not defined in `IndexSpec` itself such as `CompressionStrategy.DEFAULT_COMPRESSION_STRATEGY`, `CompressionFactory.DEFAULT_LONG_ENCODING_STRATEGY`, and `StringEncodingStrategy.DEFAULT`, etc, so is currently consistent with other stuff.

cecemei (approved):

PR #18639, opened 2025-10-15
Fix query status metric reporting bug in router

jtuglu1:
Currently, all query errors that did not originate from the router are incorrectly being recorded as successful in metric emission at the router. For example, a query that timed out on a broker would be reported as a successful query by the router during metric emission.

This is because the current code checks an exception on the response object, rather than the actual status code of the response. This is incorrect because the router is operating as a proxy, and a "successful" result is considered one which was proxied + received a response, and will not throw an exception. We should be checking the status code instead as this is the true status determined by the downstream broker.

#### Release note

Fix inaccurate query status metric reporting bug in router.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/services/src/test/java/org/apache/druid/server/AsyncQueryForwardingServletTest.java
+++ b/services/src/test/java/org/apache/druid/server/AsyncQueryForwardingServletTest.java
@@ -467,6 +467,70 @@ public class AsyncQueryForwardingServletTest extends BaseJettyTest
   }
 
   @Test
+  public void testMetricsEmittedWithErrorStatusCodeButNoResultException()
+  {
+    final TimeseriesQuery query = Druids.newTimeseriesQueryBuilder()
+                                        .dataSource("foo")
+                                        .intervals("2000/P1D")
+                                        .granularity(Granularities.ALL)
+                                        .context(ImmutableMap.of("queryId", "test-query-504"))
+                                        .build();
+
+    final HttpServletRequest requestMock = Mockito.mock(HttpServletRequest.class);
+    Mockito.when(requestMock.getAttribute("org.apache.druid.proxy.avaticaQuery")).thenReturn(null);
+    Mockito.when(requestMock.getAttribute("org.apache.druid.proxy.query")).thenReturn(query);
+    Mockito.when(requestMock.getAttribute("org.apache.druid.proxy.sqlQuery")).thenReturn(null);
+    Mockito.when(requestMock.getRemoteAddr()).thenReturn("127.0.0.1");
+    Mockito.when(requestMock.getAttribute(AuthConfig.DRUID_AUTHENTICATION_RESULT))
+           .thenReturn(new AuthenticationResult("testUser", "basic", "basic", null));
+
+    final Request proxyRequestMock = Mockito.mock(Request.class);
+    final Response responseMock = Mockito.mock(Response.class);
+    Mockito.when(responseMock.getStatus()).thenReturn(504); // Gateway Timeout
+    Mockito.when(responseMock.getHeaders()).thenReturn(HttpFields.build());
+    Mockito.when(responseMock.getRequest()).thenReturn(proxyRequestMock);
+
+    final Result result = new Result(proxyRequestMock, responseMock)
+    {
+      @Override
+      public Throwable getFailure()
+      {
+        return null; // No exception thrown
+      }
+    };
+
+    final StubServiceEmitter stubServiceEmitter = new StubServiceEmitter("", "");
+    final AsyncQueryForwardingServlet servlet = new AsyncQueryForwardingServlet(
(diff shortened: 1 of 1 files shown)


services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
```diff
@@ -760,13 +760,17 @@ public void onComplete(Result result)
...
+      boolean success = statusCode == Status.OK.getStatusCode();
       if (success) {
         successfulQueryCount.incrementAndGet();
       } else {
         failedQueryCount.incrementAndGet();
       }
-      emitQueryTime(requestTimeNs, success, sqlQueryId, queryId, result.getFailure());
+
+      // As router is simply a proxy, we don't make an effort to construct the error code from the exception ourselves.
+      // We rely on broker to set this for us if the error occurs upstream.
```
abhishekrb19: I think this should be "downstream", since brokers are downstream of routers?

services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
```diff
@@ -760,13 +760,17 @@ public void onComplete(Result result)
         return;
       }
 
-      boolean success = result.isSucceeded();
```
abhishekrb19, line 763: Why was this logic for success changed from `result.isSucceeded()` to ` result.getResponse().getStatus()`? I think this would change the semantics of what counts as a successful query to strictly a query that's returning with a 200 OK code.

Also, is this related to the error code change?

processing/src/main/java/org/apache/druid/query/QueryMetrics.java
```diff
@@ -246,7 +245,7 @@ public interface QueryMetrics<QueryType extends Query<?>>
    * Translates the given query exception into the appropriate failure status code.
    * See {@link DruidMetrics#computeStatusCode}.
    */
-  void code(@Nullable Throwable error);
```
abhishekrb19, line 249: Javadoc needs an update since there's no exception passed in anymore

services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
```diff
@@ -760,13 +760,17 @@ public void onComplete(Result result)
...
+      final int statusCode = result.getResponse().getStatus();
```
abhishekrb19, line 763: On the router, I suppose this is fine especially as `result.getResponse().getStatus()` would effectively translate to the same error code as the underlying exception on the downstream query/data nodes, if it’s of type `DruidException`/`QueryException`?


Not a blocker for this PR, but I wonder if there are e2e embedded query tests, where we can validate that the metrics emitted from these different services are similar.

services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
jtuglu1: > Why was this logic for success changed from result.isSucceeded() to  result.getResponse().getStatus()? I think this would change the semantics of what counts as a successful query to strictly a query that's returning with a 200 OK code.

See PR description. `result.isSucceeded()` is the buggy logic. It checks `result.getFailure() == null` which doesn't encompass errors downstream of the router (failure will be null in those cases since it didn't fail in the router).

This assumption is already being made a few lines down [here](https://github.com/apache/druid/blob/master/services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java#L789-L790).

services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
jtuglu1: > On the router, I suppose this is fine especially as result.getResponse().getStatus() would effectively translate to the same error code as the underlying exception, if it’s of type DruidException/QueryException?

Yes. Router is stateless and doesn't handle any business logic (aside from potential auth errors, etc.). It shouldn't make an attempt to parse any of the exceptions in this case.

services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
abhishekrb19: Ah yes makes sense. Could you replace the `success && result.getResponse().getStatus() == Status.OK.getStatusCode()` you've referenced with simply `success` now (since the logic is effectively the same)?

FrankChen021:
Do we need or have a test case for this case? There're changes in test cases, but are mainly for the renaming

jtuglu1:
> Do we need or have a test case for this case? There're changes in test cases, but are mainly for the renaming

I added status code assertions in a previous commit, but added another unit test to catch this regression.

PR #18663, opened 2025-10-18
Clarify `compactTask/availableSlot/count`  metric description

abhishekrb19:
Clarify the `compactTask/availableSlot/count` metric description. 

The formula to compute available compact task slots in a run is: `availableCompactionTaskSlots = Math.max(0, compactionTaskCapacity - busyCompactionTaskSlots)`. Often times, it's confusing why `compactTask/availableSlot/count` is lower than expected. This happens because the value is capped by `maxNumConcurrentSubTasks`, and the current metric description can be slightly misleading.

This PR has:

- [x] been self-reviewed.


kfaraz:
> It turns out that the compact duty using the native engine just caps it using maxNumConcurrentSubTasks, regardless of the phase a current compact supervisor is in.

@abhishekrb19 , I am not sure how the `busyCompactionTaskSlots` being emitted in this PR will avoid this problem,
since the value we are emitting is effectively just `maxSlots - availableSlots`. Could you please clarify this?

To determine the actual count of tasks, we could count the sub-tasks for each currently running `compact` tasks,
but that can run into other issues. I feel it is better to launch fewer `compact` tasks (due to a smaller `availableSlot` count) than over-assign `compact` tasks causing them to potentially hog up slots reserved for ingestion.
So I feel the current approach of capping at `maxNumConcurrentSubTasks` while computing `availableSlot` count is reasonable.

What do you think?

Edit: Although, it might be useful to emit the "actual" busy slot count which would be computed using the sub-task counts as mentioned above. Is that what you mean to do here?

abhishekrb19:
@kfaraz thanks for taking a look!

>It turns out that the compact duty using the native engine just caps it using maxNumConcurrentSubTasks, regardless of the phase a current compact supervisor is in. 
@abhishekrb19 , I am not sure how the busyCompactionTaskSlots being emitted in this PR will avoid this problem,
since the value we are emitting is effectively just maxSlots - availableSlots.


Yeah, I was just noting my observation on how this metric is calculated. The docs for the `compactTask/availableSlot/count` metric tripped me up a bit: “This is the max number of task slots minus any currently running compaction tasks,” rather than mentioning that it’s an _estimated_ number of currently running compaction tasks.

While at it, I was thinking that the busy slot estimate could just be emitted directly given the [condition](https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java#L430-L438) - a consistent higher utilization of the busy slots metric (and a corresponding drop in available slots) would indicate the need to tune the compaction task slots, which is what we ended up doing for some clusters.

>Although, it might be useful to emit the "actual" busy slot count which would be computed using the sub-task counts as mentioned above. Is that what you mean to do here?

I wasn't necessarily thinking of emitting the "actual" busy slot count because it doesn't influence the auto-compaction algorithm currently; also I think this can be determined using `task/*/count`?

Let me know if that makes sense.

kfaraz:
> The docs for the compactTask/availableSlot/count metric tripped me up a bit: “This is the max number of task slots minus any currently running compaction tasks,” rather than mentioning that it’s an estimated number of currently running compaction tasks.

Fair point, @abhishekrb19 , it makes sense to update the docs.

> also I think this can be determined using task/*/count

Yes, that's true, we can get the running count for different task types/datasources using the `task/running/count` metric.

> a consistent higher utilization of the busy slots metric (and a corresponding drop in available slots) would indicate the need to tune the compaction task slots

I am not against emitting the busy slot count per se.
Just didn't seem to add any new info since we are already emitting the max slot count and the available slot count.
Please let me know if I am missing some use case.

abhishekrb19:
@kfaraz, I went ahead and just made the doc change to clarify and reverted the `busySlot` metric.

docs/operations/metrics.md
```diff
@@ -413,7 +413,7 @@ These metrics are emitted by the Druid Coordinator in every run of the correspon
 |`tier/total/capacity`|Total capacity in bytes available in each tier.|`tier`|Varies|
 |`compact/task/count`|Number of tasks issued in the auto compaction run.| |Varies|
 |`compactTask/maxSlot/count`|Maximum number of task slots available for auto compaction tasks in the auto compaction run.| |Varies|
-|`compactTask/availableSlot/count`|Number of available task slots that can be used for auto compaction tasks in the auto compaction run. This is the max number of task slots minus any currently running compaction tasks.| |Varies|
+|`compactTask/availableSlot/count`|Number of task slots available for auto compaction tasks during an auto compaction run. This value represents the total number of task slots allocated for compaction minus the estimated number of currently running compaction tasks. The estimated number of tasks is capped by `maxNumConcurrentSubTasks`, which serves as an upper bound to prevent over-allocation of compaction tasks.| |Varies|
```
kfaraz: Slight change of wording to avoid ambiguity. Let me know if it makes sense.

```suggestion
|`compactTask/availableSlot/count`|Number of currently vacant task slots out of the total slots allocated for auto-compaction tasks. This value is computed as the difference between the total number of task slots allocated for auto-compaction and the estimated number of task slots currently occupied by running compaction tasks. The number of sub-tasks of each compaction task is estimated to be `maxNumConcurrentSubTasks`.| |Varies|
```

docs/operations/metrics.md
```diff
@@ -413,7 +413,7 @@ These metrics are emitted by the Druid Coordinator in every run of the correspon
 |`tier/total/capacity`|Total capacity in bytes available in each tier.|`tier`|Varies|
 |`compact/task/count`|Number of tasks issued in the auto compaction run.| |Varies|
 |`compactTask/maxSlot/count`|Maximum number of task slots available for auto compaction tasks in the auto compaction run.| |Varies|
-|`compactTask/availableSlot/count`|Number of available task slots that can be used for auto compaction tasks in the auto compaction run. This is the max number of task slots minus any currently running compaction tasks.| |Varies|
+|`compactTask/availableSlot/count`|Number of currently vacant task slots out of the total slots allocated for auto-compaction tasks. This value is computed as the difference between the total number of task slots allocated for auto-compaction and the estimated number of task slots currently occupied by running compaction tasks. The number of sub-tasks of each compaction task is estimated to be `maxNumConcurrentSubTasks`.| |Varies|
```
abhishekrb19:
```suggestion
|`compactTask/availableSlot/count`|Number of currently vacant task slots out of the total slots allocated for auto compaction tasks. This value is computed as the difference between the total number of task slots allocated for auto compaction and the estimated number of task slots currently occupied by running compaction tasks. The number of sub-tasks of each compaction task is estimated to be `maxNumConcurrentSubTasks`.| |Varies|
```

docs/operations/metrics.md
abhishekrb19: Yes, it does, thanks. Just one small tweak, I left it as auto compaction instead of auto-compaction to keep it consistent with other metrics in this block.

PR #18677, opened 2025-10-21
maybe this will fix it

clintropolis:
CI is failing to startup to run unit tests, complaining about actions-timeline version not being allowed, switched to latest per https://github.com/apache/infrastructure-actions/blob/main/actions.yml


clintropolis:
looks like unit and integration tests are running at least

cecemei (approved):

PR #18683, opened 2025-10-22
virtual storage improvements

clintropolis:
### Description
This PR makes some behavioral changes to how historical servers manage their cache files when operating in the 'virtual storage' mode added in #18176, to provide a better experience and more flexibility to callers.

The main change is that in virtual storage mode, it is now impossible for a segment to be 'missing' in the `SegmentManager` if a caller has a `DataSegment`. By 'missing' segments, I am referring to the specific Druid response a historical server provides in the query response context to cover the case of a segment drop command occurring before the query engine has a chance to obtain a reference to the files (or even before ever receiving the request), so that the broker can retry these requests to the new location the segment has been moved to (if retries are enabled the the segment is available elsewhere).

Now, in virtual storage mode, the historical will simply download the segment on demand and process it for any `DataSegment` object. So for example, when using `ServerManager` to get segments, this means that after resolving the set of `DataSegments` participating in a query from the timeline, the query engine can always obtain a `SegmentReference` so long as there is adequate disk space to download (and it is present in deep storage of course).

To complement this behavior, the segment cache itself no longer actively unmounts weakly held segment cache entries, making the coordinator drop command a near no-op. Instead, files will remain on disk until they are evicted because some other segment needs to reclaim the space so that it can be loaded. This does mean that the disk will 'fill up' over time, but is worth the benefits. To accommodate this and ensure that segment files are tracked across process boundaries (in the event of a failure or whatever), segment info files are now deleted via an 'unmount' hook on the segment files cache entry, meaning that they will not be removed until the segment files are also removed.

### DIFF
--- a/server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
+++ b/server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
@@ -977,14 +977,15 @@ public class SegmentLocalCacheManager implements SegmentCacheManager
             storageDir = null;
             location = null;
           }
+
+          final Runnable onUnmountRunnable = onUnmount.get();
+          if (onUnmountRunnable != null) {
+            onUnmountRunnable.run();
+          }
         }
       }
       finally {
         lock.unlock();
-      }
-      final Runnable onUnmountRunnable = onUnmount.get();
-      if (onUnmountRunnable != null) {
-        onUnmountRunnable.run();
       }
     }
 


server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -362,6 +383,13 @@ public AcquireSegmentAction acquireSegment(final DataSegment dataSegment) throws
           );
           try {
             if (hold != null) {
+              // write the segment info file if it doesn't exist. this can happen if we are loading after a drop
+              final File segmentInfoCacheFile = new File(getEffectiveInfoDir(), dataSegment.getId().toString());
+              if (!segmentInfoCacheFile.exists()) {
+                jsonMapper.writeValue(segmentInfoCacheFile, dataSegment);
```
gianm: Use `FileUtils.writeAtomically` to avoid leaving a partially-written file after a crash.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -421,7 +449,23 @@ private AcquireSegmentAction acquireExistingSegment(SegmentCacheEntryIdentifier
...
+      // virtual storage doesn't do anything with loading immediately, but check to see if the segment is already cached
+      // and if so, clear out the onUnmount action
+      final ReferenceCountingLock lock = lock(dataSegment);
+      synchronized (lock) {
+        try {
+          final SegmentCacheEntryIdentifier cacheEntryIdentifier = new SegmentCacheEntryIdentifier(dataSegment.getId());
+          for (StorageLocation location : locations) {
+            final SegmentCacheEntry cacheEntry = location.getCacheEntry(cacheEntryIdentifier);
+            if (cacheEntry != null) {
+              cacheEntry.onUnmount.set(null);
```
gianm: Can this race with the call in `removeInfoFile`? That one doesn't seem to acquire a lock. Same comment for the other various calls to `set(null)`.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -290,6 +291,27 @@ public void storeInfoFile(final DataSegment segment) throws IOException
...
+    final Runnable delete = () -> deleteSegmentInfoFile(segment);
+    final SegmentCacheEntryIdentifier entryId = new SegmentCacheEntryIdentifier(segment.getId());
+    boolean isCached = false;
+    // defer deleting until the unmount operation of the cache entry, if possible, so that if the process stops before
+    // the segment files are deleted, they can be properly managed on startup (since the info entry still exists)
+    for (StorageLocation location : locations) {
+      final SegmentCacheEntry cacheEntry = location.getCacheEntry(entryId);
+      if (cacheEntry != null) {
+        isCached = true;
+        cacheEntry.onUnmount.set(delete);
```
gianm: Can this race with an actual unmounting? A lock doesn't seem to be acquired at this point in time.

clintropolis: yea, this one isn't under a segment lock so could, fixed to just make methods on cache entry to synchronize with mount/unmount

PR #18684, opened 2025-10-22
Validate that dynamic partition spec values are positive

GabrielCWT:
Ensures that maxRowsPerSegment and maxTotalRows are either null or a positive value.

### DIFF
--- a/docs/ingestion/native-batch.md
+++ b/docs/ingestion/native-batch.md
@@ -333,8 +333,8 @@ The `partitionsSpec` types have different characteristics.
 |Property|Description|Default|Required|
 |--------|-----------|-------|---------|
 |`type`|Set the value to `dynamic`.|none|yes|
-|`maxRowsPerSegment`|Used in sharding. Determines how many rows are in each segment.|5000000|no|
-|`maxTotalRows`|Total number of rows across all segments waiting for being pushed. Used in determining when intermediate segment push should occur.|20000000|no|
+|`maxRowsPerSegment`|Used in sharding. Determines how many rows are in each segment. Value must be greater than 0.|5000000|no|
+|`maxTotalRows`|Total number of rows across all segments waiting for being pushed. Used in determining when intermediate segment push should occur. Value must be greater than 0.|20000000|no|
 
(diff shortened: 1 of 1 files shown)


processing/src/main/java/org/apache/druid/indexer/partitions/DynamicPartitionsSpec.java
```diff
@@ -50,10 +50,13 @@ public DynamicPartitionsSpec(
       @JsonProperty("maxTotalRows") @Nullable Long maxTotalRows
   )
   {
-    this.maxRowsPerSegment = PartitionsSpec.isEffectivelyNull(maxRowsPerSegment)
+
+    this.maxRowsPerSegment = PartitionsSpec.isEffectivelyNull(maxRowsPerSegment) || maxRowsPerSegment <= 0
```
kfaraz: If the given value is negative, we should throw an `InvalidInput` exception rather than defaulting to some other value.

server/src/test/java/org/apache/druid/server/compaction/CompactionStatusTest.java
```diff
@@ -213,7 +213,7 @@ public void testStatusWhenLastCompactionStateIsEmpty()
   @Test
   public void testStatusOnPartitionsSpecMismatch()
```
kfaraz, line 214: We can probably add a test for negative/zero values.

GabrielCWT: I was thinking of adding tests for this as well. Should we create a new class `DynamicPartitionsSpecTest` instead? I think testing the constructor behavior in `CompactionStatusTest` is not a good idea

GabrielCWT: Just a follow up, I've created `DynamicPartitionsSpecTest`

FrankChen021 (approved):
LGTM, just one thing to confirm, does the document clearly state tthat hese properties must be greater than zero?

PR #18686, opened 2025-10-22
Implement Interval toString optimisation

GabrielCWT:
This PR implements an optimisation for the Interval.toString method.

Based on the flame graph before optimisation, we can see that the toString method takes up a significant amount of the running time (~16%). After the optimisation, the optimised format method only accounts for ~4% of the running time.

We use the algorithm found [here](https://howardhinnant.github.io/date_algorithms.html#civil_from_days) in order to optimise the toString method. 

The implementation differs slightly as we know Druid's intervals. As such the following tweaks have been made:

- We fix era to be 5 since we know the dates will be within [2000-03-01T00:00:00.000Z, 2400-02-29T23:59:59.999Z]
- We omit any negative dates checks.

Any dates which do not fall within the era or doesn't have UTC timezone will use `Interval.toString()` as the fallback.

```
Benchmark                                                              Mode  Cnt  Score   Error  Units
DefaultQueryMetricsBenchmark.baselineMetrics                           avgt   40  1.207 ± 0.022  us/op
DefaultQueryMetricsBenchmark.withCustomIntervalToStringDimension       avgt   40  0.413 ± 0.010  us/op
```

Note: I do agree that the algorithm is difficult to understand and there are magic numbers still existing. I did my best to refactor most of the numbers to variables if they were constants and I've also tried to maintain the same structure as the algorithm provided in order to make it easier to understand. 

## Flame Graph before Optimisation
image

## Flame Graph after Optimisation
image

### DIFF
--- a/processing/src/test/java/org/apache/druid/java/util/common/FastIntervalStringFormatterTest.java
+++ b/processing/src/test/java/org/apache/druid/java/util/common/FastIntervalStringFormatterTest.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.java.util.common;
+
+import org.joda.time.Interval;
+import org.joda.time.chrono.ISOChronology;
+import org.junit.Assert;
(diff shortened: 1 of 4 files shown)


processing/src/test/java/org/apache/druid/java/util/common/FastIntervalStringFormatterTest.java
```diff
@@ -0,0 +1,105 @@
...
+    Mockito.when(intervalMock.toString()).thenReturn(dummyString);
+
+    String out = FastIntervalStringFormatter.format(intervalMock);
+
+    Assert.assertEquals(dummyString, out);
+
+    final Interval[] fallbackIntervals = new Interval[]{
+        // Boundary values
+        Intervals.of("2000-02-28T00:00:00.000Z/2000-02-29T00:00:00.000Z"),
+        Intervals.of("2000-02-29T00:00:00.000Z/2000-03-01T00:00:00.000Z"),
```
FrankChen021, line 87: a good case that verifies the leap year

kfaraz:
Thanks for the perf improvement, @GabrielCWT .

While an improvement, the impact seems to be only in the micro-second range.
Will the effect be more pronounced if the query interval spec has a large number of intervals?

GabrielCWT:
> While an improvement, the impact seems to be only in the micro-second range. Will the effect be more pronounced if the query interval spec has a large number of intervals?

@kfaraz, the improvement should be more pronounced when there is a high QPS. The `Interval.toString()` is called when emitting metrics during every query. Obviously with low QPS, this optimisation will be barely noticeable.  You are also right that we would see an improvement if there was a large number of intervals.

PR #18687, opened 2025-10-22
Historical Startup -- Configurable loading strategy

### LINKED ISSUE #18446: Historical Segment Cache Loading Strategy on Start-up
### Description

- Propose a configurable startup strategy that eagerly loads only recent (“hot”) segments, while leaving older (“cold”) segments to load lazily on first access.
- Propose to deprecate druid.segmentCache.lazyLoadOnStart in favour for configs that gives more flexibility to historical's segment cache loading during startup.

### Motivation

- Non-lazy segment loading takes long if Historical segment count is high (observed ~22 minutes per Historical; ~39 hours cluster-wide).
- Lazy-loading improves startup time but initial queries over hot data can be slow.
- Many clusters primarily query the last N days/weeks; we can make that slice eager at startup to maintain query performance.

### Proposal
Deprecate druid.segmentCache.lazyLoadOnStart in favor of a single strategy-driven config:

New: historicalStartupSegmentCacheStrategy with options: 
1. loadLazily (all segments lazy)
2. loadAllEagerly (all segments eager)
3. loadEagerlyForPeriod (recent window eager, older lazy)

When loadEagerlyForPeriod is selected, require:
historicalStartupSegmentCacheStrategy.loadPeriod (ISO-8601 period, e.g., P7D, P30D)

#### Backward compatibility and migration
Keep reading druid.segmentCache.lazyLoadOnStart for at least a few more releases with a deprecation warning.
We can map true -> loadLazily, false -> loadAllEagerly. 
Using the new `historicalStartupSegmentCacheStrategy` overwrites the `lazyLoadOnStart` setting, [Optional: and a warning is logged if both settings are configured].

The pros of relying on the new config allows us to implement more load strategies that we want.

Config names are open for discussion, do drop some suggestions!

GWphua:
Fixes #18446 

### Description

This PR belongs to a set of PR's that hope to optimize the start-up time of Historical. I came across this problem when I am running 100+ Historical servers, each needing to process a large number of segments during start-up (~85k). When conducting any updates to the Historical, each segment will take 15.3ms to load, and the start-up time for one historical will easily take >20mins. (Meaning 1.5 days to complete update for all Historical servers!)

I looked to using `lazyLoadOnStart` to speed up startup time. Lazy loading processes each segment metadata in 3.23ms, and this shortens the start-up time to ~4min. However, using this strategy will cause some hiccups to query latency when we are trying to conduct an upgrade. I plan to solve this by selectively choosing which segments to load during Historical startup. 

image

#### Finding a middle ground

By studying the usage pattern of my clusters, I noticed that each Historical stores 1 month worth of data, but the heavy querying is only restricted to 7 days, while occasional queries are issued for the time period out of the last 7 days. Hence, I changed the logic of segment loading during Historical startup to provide a configurable time period to load segments eagerly (and the rest lazily). The use of the time period is dependent on the querying habits of the cluster.

Here's a benchmark of the improvements. I also included a test for #18489, which helps me to shave 10s off the start-up time. (Total of 78.5% improvement in loading time)

|Description | Loading Time|
|-- | --|
| Original | 1,494.631 s |
| EagerLoadingForPeriod | 328.819 s |
| EagerLoadingForPeriod + ConcurrentSegmentFileLoad | 318.262 s |

#### Release note

Segment loading during Historical service startup is now configurable with `druid.segmentCache.startupLoadStrategy`. This new setting allows users to choose between the existing eager loading (loadAllEagerly), a new lazy loading (loadAllLazily) option for faster startups, and a hybrid strategy (loadEagerlyBeforePeriod) that ensures low query latency for the most recent data while deferring the loading cost of older data.

Deprecated isLazyLoadOnStart.

<hr>

 * `docs/configuration/index.md`
 * `SegmentStatsMonitor`
 * `SegmentLoaderConfig`
 * `SegmentLocalCacheManager`
 * `startup/HistoricalStartupCacheLoadStrategy`
 * `startup/HistoricalStartupCacheLoadStrategyFactory`
 * `startup/LoadAllEagerlyStrategy`
 * `startup/LoadAllLazilyStrategy`
 * `startup/LoadEagerlyBeforePeriod`

<hr>

This PR has:

- [x] been self-reviewed.


server/src/main/java/org/apache/druid/server/metrics/SegmentStatsMonitor.java
```diff
@@ -63,7 +64,7 @@ public SegmentStatsMonitor(
       SegmentLoaderConfig segmentLoaderConfig
   )
   {
-    if (segmentLoaderConfig.isLazyLoadOnStart()) {
+    if (segmentLoaderConfig.getStartupCacheLoadStrategy().equals(LoadAllLazilyStrategy.STRATEGY_NAME)) {
```
FrankChen021: flip the equals

docs/configuration/index.md
```diff
@@ -1585,7 +1585,9 @@ These Historical configurations can be defined in the `historical/runtime.proper
...
+|`druid.segmentCache.lazyLoadOnStart`|_DEPRECATED_ Use `druid.segmentCache.startupLoadStrategy` instead. Whether or not to load segment columns metadata lazily during historical startup. When set to true, Historical startup time will be dramatically improved by deferring segment loading until the first time that segment takes part in a query, which will incur this cost instead.|false|
+|`druid.segmentCache.startupLoadStrategy`|Selects the segment column metadata loading strategy during historical startup. Possible values are `loadAllEagerly`, `loadAllLazily`, and `loadEagerlyBeforePeriod`. More details on each strategy below.|`loadAllEagerly`|
+|`druid.segmentCache.startupLoadPeriod`| Used only when startup load strategy `loadEagerlyBeforePeriod` is configured. Suppose timestamp `t` is when the Historical started up. Any segment metadata with interval that does not overlap with the interval of `[t - startupLoadPeriod, t]` will be lazily loaded.|`P7D`|
```
FrankChen021:
```suggestion
|`druid.segmentCache.startupLoadPeriod`| Used only when startup load strategy `loadEagerlyBeforePeriod` is configured. Suppose timestamp `t` is when the Historical started up. Any segment metadata with interval that falls in the interval of `[t - startupLoadPeriod, t]` will be loaded.|`P7D`|
```

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -853,7 +858,8 @@ public void mount(StorageLocation mountLocation) throws SegmentLoadingException
           final SegmentizerFactory factory = getSegmentFactory(storageDir);
 
           @SuppressWarnings("ObjectEquality")
-          final boolean lazy = config.isLazyLoadOnStart() && lazyLoadCallback != SegmentLazyLoadFailCallback.NOOP;
+          final boolean lazy = loadStrategy.shouldLoadLazily(dataSegment)
+                               && lazyLoadCallback != SegmentLazyLoadFailCallback.NOOP;
```
FrankChen021: put the shouldLoadLazily condition in the last

FrankChen021 (approved):
LGTM with minor suggestions

docs/configuration/index.md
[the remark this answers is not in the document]
abhishekrb19: Also, how feasible/extensible is it to accept a map of datasource to load period, to allow configurable periods per datasource? (similar to the load [rules](https://druid.apache.org/docs/latest/operations/rule-configuration#period-load-rule) config where each datasource can have different load retention rules)

abhishekrb19: Why do we default to `P7D`? IMO this should be a required property configured by operators and fail startup if unspecified (since there's no reasonable default that will apply to most users).

docs/configuration/index.md
```diff
@@ -1602,6 +1604,14 @@ In `druid.segmentCache.locationSelector.strategy`, one of `leastBytesUsed`, `rou
...
 
+In `druid.segmentCache.startupLoadStrategy`, one of `loadAllEagerly`, `loadAllLazily`, or `loadEagerlyBeforePeriod` could be specified to represent the strategy to load segments when starting the Historical service.
+
+|Strategy|Description|
+|--------|-----------|
+|`loadAllEagerly`|The default startup strategy. The Historical service will load all segment column metadata immediately during the initial startup process.|
+|`loadAllLazily`|To significantly improve historical system startup time, segments are not loaded during the initial startup sequence. Instead, the loading cost is deferred, and will be incurred the first time a segment is referenced by a query.|
+|`loadEagerlyBeforePeriod`|Provides a balance between fast startup and query performance. The Historical service will eagerly load column metadata only for segments that fall within the most recent period defined by `druid.segmentCache.startupLoadPeriod`. Segments outside this recent period will be loaded on-demand when first queried.|
```
abhishekrb19, line 1613: How feasible/extensible is it to accept a map of datasource to load period, to allow configurable periods per datasource? (similar to the `loadByPeriod` - load [rules](https://druid.apache.org/docs/latest/operations/rule-configuration#period-load-rule) config where each datasource can have different load retention rules)

I think having that option would allow a lot more flexibility to operators as the query workloads can be vastly different.

server/src/main/java/org/apache/druid/segment/loading/SegmentLoaderConfig.java
```diff
@@ -84,11 +97,31 @@ public List<StorageLocationConfig> getLocations()
...
+  @Deprecated
   public boolean isLazyLoadOnStart()
   {
     return lazyLoadOnStart;
   }
 
+  public String getStartupCacheLoadStrategy()
+  {
+    return startupLoadStrategy == null
+           ? isLazyLoadOnStart()
```
abhishekrb19: Do we have validations that we don't have incompatible configurations enabled accidentally? for example, when `lazyLoadOnStart = true` and `startupLoadStrategy` are both  set

server/src/main/java/org/apache/druid/server/coordination/startup/HistoricalStartupCacheLoadStrategy.java
```diff
@@ -0,0 +1,27 @@
...
+ * under the License.
+ */
+
+package org.apache.druid.server.coordination.startup;
+
+import org.apache.druid.timeline.DataSegment;
+
+public interface HistoricalStartupCacheLoadStrategy
+{
+  boolean shouldLoadLazily(DataSegment segment);
```
abhishekrb19, line 45: Please add javadocs on this interface and the method

server/src/main/java/org/apache/druid/server/coordination/startup/HistoricalStartupCacheLoadStrategyFactory.java
```diff
@@ -0,0 +1,43 @@
...
+      case LoadAllLazilyStrategy.STRATEGY_NAME:
+        return new LoadAllLazilyStrategy();
+      case LoadAllEagerlyStrategy.STRATEGY_NAME:
+        return new LoadAllEagerlyStrategy();
+      case LoadEagerlyBeforePeriod.STRATEGY_NAME:
+        return new LoadEagerlyBeforePeriod(config.getStartupLoadPeriod());
+      default:
+        throw DruidException.forPersona(DruidException.Persona.OPERATOR)
+                            .ofCategory(DruidException.Category.UNSUPPORTED)
+                            .build("Unknown configured Historical Startup Loading Strategy[%s]", strategyName);
```
abhishekrb19: This could just be a simpler helper method in `SegmentLocalCacheManager` where this is being used. I think calling it `factorize()` is not needed.

server/src/main/java/org/apache/druid/server/coordination/startup/LoadEagerlyBeforePeriod.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.druid.java.util.common.DateTimes;
+import org.apache.druid.java.util.common.logger.Logger;
+import org.apache.druid.timeline.DataSegment;
+import org.joda.time.DateTime;
+import org.joda.time.Interval;
+import org.joda.time.Period;
+
+public class LoadEagerlyBeforePeriod implements HistoricalStartupCacheLoadStrategy
```
abhishekrb19, line 41: Please add a brief javadoc for these implementations

server/src/main/java/org/apache/druid/server/coordination/startup/LoadEagerlyBeforePeriod.java
```diff
@@ -0,0 +1,56 @@
...
+{
+  private static final Logger log = new Logger(LoadEagerlyBeforePeriod.class);
+  public static final String STRATEGY_NAME = "loadEagerlyBeforePeriod";
+
+  private final Interval eagerLoadingWindow;
```
abhishekrb19: nit: `eagerLoadingInterval`

server/src/main/java/org/apache/druid/segment/loading/SegmentLoaderConfig.java
```diff
@@ -84,11 +97,31 @@ public List<StorageLocationConfig> getLocations()
...
+   * Removal of this method in the future will requires a change in {@link #getStartupCacheLoadStrategy()}
+   * to default to {@link LoadAllEagerlyStrategy#STRATEGY_NAME} when {@link #startupLoadStrategy} is null.
+   */
+  @Deprecated
   public boolean isLazyLoadOnStart()
   {
     return lazyLoadOnStart;
   }
 
+  public String getStartupCacheLoadStrategy()
```
abhishekrb19: This API contract seems weird...it's returning the strategy name rather than the concrete strategy itself. What do you think about inlining the logic from `factorize()` and returning `HistoricalStartupCacheLoadStrategy`?

server/src/main/java/org/apache/druid/server/coordination/startup/HistoricalStartupCacheLoadStrategyFactory.java
abhishekrb19: please see my related comment on simplifying this with `getStartupCacheLoadStrategy`

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
abhishekrb19: Just to confirm that this would only affect bootstrapping flow and not other features like virtual fabric / on-demand queries that @clintropolis was working on? 
I'm asking because this config usage is deep in the segment load flow that can  inadvertently affect other features since it's used by 4-5 related callers (I can help double check on this later)


Alternatively, it'd be cleaner and safer to move the lod logic to `SegmentCacheBootstrapper` where we know certainly that it'd be secluded to the segment bootstrapping logic on data servers. For example, it can be somewhere in `SegmentCacheBootstrapper#loadSegmentsOnStartup()` where the segment loader config is present.

server/src/main/java/org/apache/druid/server/coordination/startup/HistoricalStartupCacheLoadStrategyFactory.java
clintropolis: couldn't these just be json annotated types or something and just let the json config stuff handle materializing the right one like on the segment loader config instead of having this factorize method?

like the interface would be decorated with 
```
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
@JsonSubTypes(value = {
    @JsonSubTypes.Type(name = LoadAllEagerlyStrategy.STRATEGY_NAME, value = LoadAllEagerlyStrategy.class),
    @JsonSubTypes.Type(name = LoadAllLazilyStrategy.STRATEGY_NAME, value = LoadAllLazilyStrategy.class),
    @JsonSubTypes.Type(name = LoadEagerlyBeforePeriod.STRATEGY_NAME, value = LoadEagerlyBeforePeriod.class)
})
public interface HistoricalStartupCacheLoadStrategy
{
...
}
```

and then make all of the implementations annotated with `@JsonCreator` and `@JsonProperty`, so then this is just a property on `SegmentLoaderConfig`, and the configs would be shifted to like

```
|`druid.segmentCache.startupLoadStrategy.type`|Selects the segment column metadata loading strategy during historical startup. Possible values are `loadAllEagerly`, `loadAllLazily`, and `loadEagerlyBeforePeriod`. More details on each strategy below.|`loadAllEagerly`|
|`druid.segmentCache.startupLoadStrategy.period`| Used only when startup load strategy `loadEagerlyBeforePeriod` is configured. Suppose timestamp `t` is when the Historical started up. Any segment metadata with interval that does not overlap with the interval of `[t - startupLoadPeriod, t]` will be lazily loaded.|`P7D`|
```

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
clintropolis: i think this should not really interfere, lazyLoadCallback is only set to a value other than NOOP during bootstrapping, and in virtual storage mode bootstrapping only mounts already cached segments (though maybe it should download bootstrapping segments eagerly... haven't thought about it all that much yet tbh), and should probably follow whatever logic is set here.

I don't think it makes sense to move this logic out of here

server/src/main/java/org/apache/druid/segment/loading/SegmentLoaderConfig.java
clintropolis: or https://github.com/apache/druid/pull/18687#discussion_r2458521243 instead of inlining and let json deserialization do the work

server/src/main/java/org/apache/druid/server/coordination/startup/HistoricalStartupCacheLoadStrategyFactory.java
abhishekrb19: Ah yes, we can let jackson do its thing here

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
abhishekrb19: okay thanks for confirming that it wouldn't interfere with other flows

server/src/main/java/org/apache/druid/segment/loading/SegmentLoaderConfig.java
GWphua: The current implementation ensures that a declared `startupLoadStrategy` overwrites whatever `isLazyLoadOnStart` is set to, else `isLazyLoadOnStart` is used for forward compatibility purposes. Personally, there is no need to provide validation with this overwrite functionality -- though it is relatively simple to implement one if you feel it's needed.

docs/configuration/index.md
GWphua: `P7D` is an arbitrary number I have set up that fits my use-cases. I will add appropriate validations and warn remind to configure this value if unset.

docs/configuration/index.md
GWphua: This is workable -- 

I can change `startupLoadStrategy.period` to `startupLoadStrategy.datasourceToPeriodMapping`, which receives something like a JSON 

e.g. 
`{"DS1": "P7D", "DS2": "P2D", ".": "P7D"}`

Where `.` refers to the default configuration (since datasources cannot start with `.`)

server/src/main/java/org/apache/druid/server/coordination/startup/HistoricalStartupCacheLoadStrategyFactory.java
GWphua: @clintropolis @abhishekrb19 

This is helpful, thanks for the suggestion!

docs/configuration/index.md
GWphua: Updated default setting.

kfaraz:
I guess it is natural to allow more start up strategies since we already support - loadAllLazily and loadAllEagerly.
But the approach here feels a little complicated. We are also trying to re-invent the things which load rules already do.
It would perhaps make more sense to use "Virtual Storage Fabric" + load rule combo since they are much more powerful and will only evolve over time.

IIUC, the use case here is that we frequently query the data of the last `x` days and also query the data of the last `x + y` days but less frequently so. Rather than doing the eager vs lazy load of the segments in the `y` days only on startup, can we _always_ load these segments lazily?

So, maybe the solution could be a mix of the recently added "Virtual Storage Fabric" + some load rules.

Load rules would perhaps look something like:

Rule 1: `loadByPeriod` last 7 days (i.e. load eagaerly on startup and otherwise)
Rule 2: `loadLazilyByPeriod` last 30 days (i.e. these segments are present on virtual storage fabric and will be loaded when the first query is received for them. Coordinator remains agnostic of this fact. The virtual storage can also be reclaimed to load other virtual segments.).
Rule 3: `dropForever` any data beyond the last 30 days.

@FrankChen021 , @GWphua , would this satisfy your use case?
@clintropolis , @abhishekrb19 , what are your thoughts?

FrankChen021:
> I guess it is natural to allow more start up strategies since we already support - loadAllLazily and loadAllEagerly. But the approach here feels a little complicated. We are also trying to re-invent the things which load rules already do. It would perhaps make more sense to use "Virtual Storage Fabric" + load rule combo since they are much more powerful and will only evolve over time.
> 
> IIUC, the use case here is that we frequently query the data of the last `x` days and also query the data of the last `x + y` days but less frequently so. Rather than doing the eager vs lazy load of the segments in the `y` days only on startup, can we _always_ load these segments lazily?
> 
> So, maybe the solution could be a mix of the recently added "Virtual Storage Fabric" + some load rules.
> 
> Load rules would perhaps look something like:
> 
> Rule 1: `loadByPeriod` last 7 days (i.e. load eagaerly on startup and otherwise) Rule 2: `loadLazilyByPeriod` last 30 days (i.e. these segments are present on virtual storage fabric and will be loaded when the first query is received for them. Coordinator remains agnostic of this fact. The virtual storage can also be reclaimed to load other virtual segments.). Rule 3: `dropForever` any data beyond the last 30 days.
> 
> @FrankChen021 , @GWphua , would this satisfy your use case? @clintropolis , @abhishekrb19 , what are your thoughts?

I think virtual storage + loading rule is different from what this PR is doing.
under the virtual storage mode, the segment is not loaded from deep storage until first query comes. 
Here, the segment is not loaded into page `CACHE` during start up of historical, after a segment has been loaded by a historical node. the implementation here is not re-invent, and not complicated.

If the goal of virtual storage will be the dominant mode in future, introducing `loadLazilyByPeriod` to loading rules makes sense.

Virtual storage is just merged and is still under experiemental, I don't know when it will be production ready, and what's the roadmap for it. Making a small change to existing segment cache loading is still worthy.

docs/configuration/index.md
FrankChen021: My opinion is that let's keep the change in this PR small enough. for datasource level configuration, if there's really need for this feature, we can implement it by defining a datasource level configuration

GWphua: > How feasible/extensible is it to accept a map of datasource to load period, to allow configurable periods per datasource? (similar to the `loadByPeriod` - load [rules](https://druid.apache.org/docs/latest/operations/rule-configuration#period-load-rule) config where each datasource can have different load retention rules)
> 
> I think having that option would allow a lot more flexibility to operators as the query workloads can be vastly different.

I feel we can leave this for another PR, since it is out of scope of this intended PR. WDYT? @abhishekrb19

kfaraz:
Thanks for the clarification, @FrankChen021 !
I agree that the solution here is different from what vsf + load rule would do.
But I was wondering if it was close enough to satisfy your use case. That is, why not delay downloading the segment from the deep storage too?

> the implementation here is not re-invent, and not complicated.
> Making a small change to existing segment cache loading is still worthy.

True, as mentioned, I am not opposed to the idea of new startup cache load strategies. It only seems natural.
But the additional configs mean more maintenance work for cluster admins. They would need to think about which segments they want to be loaded on the cache eagerly and which are okay to be left for later. It is a question (kind of) similar to load rules i.e. which segments to keep on historicals and which are okay to be left just on the deep store. The page cache is just the first level of caching, the second being the disk of the historical itself.

Since load rules already work well with the concept of period-based loading, I hoped it would be more useful for the future to just extend that concept to cover such use cases as well.

But if that doesn't cover your use case, I can understand.

> If the goal of virtual storage will be the dominant mode in future, introducing loadLazilyByPeriod to loading rules makes sense.
> Virtual storage is just merged and is still under experiemental, I don't know when it will be production ready, and what's the roadmap for it.

I think @clintropolis would have some insights there but I imagine it should see good adoption in the near future. I was hoping you would be one of the early birds! 😉

PR #17916, opened 2025-04-15
Add audit logging to BasicAuthorizerResource update methods

lujiefsi:
# Pull Request: Add audit logging to BasicAuthorizerResource update methods

## Related Issue
Fixes #17914

## Description
This PR adds missing audit logging to three update methods in the BasicAuthorizerResource class. Without these audit logs, changes to user authorizations and group mappings were not being properly tracked, creating a security monitoring gap.

The following methods now include audit logging:
- `authorizerUserUpdateListener`
- `authorizerGroupMappingUpdateListener`
- `authorizerUpdateListener` (deprecated)

Each method now calls `performAuditIfSuccess()` after processing but before returning the response, ensuring that successful update operations are properly recorded in the audit log. The audit messages include the authorizer name and payload size to provide context about the operations being performed.

## Changes Made
Added audit logging calls to three methods in `BasicAuthorizerResource.java`:

```java
performAuditIfSuccess(
    authorizerName,
    req,
    response,
    "Update user authorizations for authorizer[%s]",
    authorizerName);
```

Similar patterns were added to the other two methods, following the existing audit logging pattern used elsewhere in the class.

## Testing Done
- Verified that audit logs are correctly generated when these methods are called
- Confirmed that the log messages contain the appropriate context information
- Ensured that the existing functionality works correctly with the added audit logging

## Additional Notes
This change improves security monitoring and compliance by ensuring all authorization-related changes are properly tracked in audit logs, closing a security gap identified in issue #17914.

PR #18688, opened 2025-10-23
Fix bug in basic authorizer auditing and add embedded test

kfaraz:
### Bug

#17916 introduced a bug where it creates audit log entries when a change made to authorizer roles and permissions are synced from the Coordinator to other services.
Audit logs should not be created in the `/listen` APIs and should be created only by the Coordinator.

### Fix

- Revert the changes made in #17916 
- Add embedded test to verify created audit entries

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/extensions-core/druid-basic-security/src/main/java/org/apache/druid/security/basic/authorization/endpoint/BasicAuthorizerResource.java
+++ b/extensions-core/druid-basic-security/src/main/java/org/apache/druid/security/basic/authorization/endpoint/BasicAuthorizerResource.java
@@ -24,7 +24,6 @@ import com.google.inject.Inject;
 import com.sun.jersey.spi.container.ResourceFilters;
 import org.apache.druid.audit.AuditEntry;
 import org.apache.druid.audit.AuditManager;
-import org.apache.druid.common.config.Configs;
 import org.apache.druid.guice.LazySingleton;
 import org.apache.druid.java.util.common.StringUtils;
 import org.apache.druid.security.basic.BasicSecurityResourceFilter;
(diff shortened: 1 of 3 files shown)


kfaraz:
FYI @lujiefsi

extensions-core/druid-basic-security/src/main/java/org/apache/druid/security/basic/authorization/endpoint/BasicAuthorizerResource.java
```diff
@@ -610,6 +612,24 @@ public Response getCachedSerializedGroupMap(
     return resourceHandler.getCachedGroupMappingMaps(authorizerName);
   }
 
+  @GET
+  @Path("/db/{authorizerName}/history")
+  @Produces(MediaType.APPLICATION_JSON)
+  @ResourceFilters(BasicSecurityResourceFilter.class)
+  public Response getAuthorizerUpdateHistory(
```
cryptoe: It looks like the only caller for this resource is the embedded test. 
Do we really need this API ?

cryptoe: If you really feel strongly about having this API, I guess a revert PR followed by a new patch for adding an API makes more sense no?

kfaraz: Sure, we can leave out the API. It just seemed weird to have audit logs and no way to see them for the end user. But we can add it later when needed.

For now, I have updated the embedded test so that it fetches the audit entries from the DB directly.

cryptoe: Please do update the description whenever you have the chance. 
Rest all lgtm!!. 
Thanks @kfaraz for figuring this one out.

PR #18689, opened 2025-10-23
Prometheus per label ttl

### LINKED ISSUE #14638: prometheus-emitter metrics ttl

aho135:
Fixes #14638

This is an extension of https://github.com/apache/druid/pull/18598 to enhance the TTL to track metrics at the label level instead of just the metric level. 

### Description

This PR enhances the current metric TTL tracking in the prometheus-emitter so that all unique label combinations are tracked instead of just the metric name. This is done using a ConcurrentHashMap which maintains a mapping of label values to a Stopwatch. The map needs to handle concurrent updates because there is a possibility for the TTL scheduler to remove a key from a map at the same that the key is updated during metric emission. These are handled through the two atomic operations removeIfExpired and resetLastUpdateTime

#### Release note
Enhances the prometheus-emitter metric TTL tracking to consider all label value combinations instead of just the metric name. 

<hr>

 * `PrometheusEmitter`
 * `PrometheusEmitterTest`
 * `DimensionsAndCollector`

<hr>

This PR has:

- [x] been self-reviewed.


extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -286,13 +290,16 @@ private void cleanUpStaleMetrics()
...
       DimensionsAndCollector metric = entry.getValue();
-      if (metric.isExpired()) {
-        log.debug(
-            "Metric [%s] has expired (last updated [%d] ms ago)",
-            entry.getKey(),
-            metric.getMillisSinceLastUpdate()
-        );
-        metric.getCollector().clear();
+      Set<List<String>> labelValuesCopy = new HashSet<>(metric.getLabelValuesToStopwatch().keySet());
+      for (List<String> labelValues : labelValuesCopy) {
```
abhishekrb19: Is a copy required?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -286,13 +290,16 @@ private void cleanUpStaleMetrics()
...
+        if (metric.removeIfExpired(labelValues)) {
+          log.debug(
+              "Metric [%s] with labels [%s] has expired",
+              entry.getKey(),
+              labelValues
+          );
+          metric.getCollector().remove(labelValues.toArray(new String[0]));
```
abhishekrb19: Should the corresponding entry from `labelValuesToStopwatch` also be removed?  Otherwise, `labelValuesToStopwatch` can grow unbounded right?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -66,22 +70,41 @@ public double[] getHistogramBuckets()
...
-    return updateTimer.millisElapsed();
+    return labelValuesToStopwatch;
   }
 
-  public boolean isExpired()
+  /**
+   * For the given labelValues, checks if the metric value has been updated within the configured flushPeriod.
+   * Returns true and removes the entry from the map if it has expired, otherwise returns false.
+   */
+  public boolean removeIfExpired(List<String> labelValues)
```
abhishekrb19: nit: what do you think of something like `areLabelValuesExpired` or `isMetricExpired`?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -33,7 +37,7 @@ public class DimensionsAndCollector
   private final SimpleCollector collector;
   private final double conversionFactor;
   private final double[] histogramBuckets;
-  private final Stopwatch updateTimer;
+  private final ConcurrentHashMap<List<String>, Stopwatch> labelValuesToStopwatch;
```
abhishekrb19: Did you see any memory pressure with this change? Previously [there](https://github.com/apache/druid/pull/18598#discussion_r2409716522) were some concerns on the memory usage with this approach, so it would be nice if you could share any observations

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
abhishekrb19: nit: to make this more boolean-y :) what do you think of something like `shouldRemoveIfExpired`?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
abhishekrb19, line 40: Did you observe any memory pressure with this change? Previously [there](https://github.com/apache/druid/pull/18598#discussion_r2409716522) were some concerns on the memory usage with this approach, so it would be nice if you could share any observations

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
abhishekrb19: nit: to make this more boolean-y :) what do you think of calling this something like `shouldRemoveIfExpired` or `haveLabelsExpired`?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -286,13 +290,16 @@ private void cleanUpStaleMetrics()
...
     for (Map.Entry<String, DimensionsAndCollector> entry : map.entrySet()) {
       DimensionsAndCollector metric = entry.getValue();
-      if (metric.isExpired()) {
-        log.debug(
-            "Metric [%s] has expired (last updated [%d] ms ago)",
-            entry.getKey(),
-            metric.getMillisSinceLastUpdate()
-        );
-        metric.getCollector().clear();
+      Set<List<String>> labelValuesCopy = new HashSet<>(metric.getLabelValuesToStopwatch().keySet());
```
abhishekrb19: Is a copy of the map required here if the underlying map is already concurrent safe from using `ConcurrentMap`? We're also doing `computeIfPresent()` on the map entries whihh should be safe

extensions-contrib/prometheus-emitter/src/test/java/org/apache/druid/emitter/prometheus/PrometheusEmitterTest.java
```diff
@@ -507,14 +510,81 @@ public void testMetricTtlUpdate()
...
+    catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+    }
     Assert.assertTrue(
-        "Update time should have been refreshed",
-        timeSinceLastUpdate < waitTime
+        "Metric should be expired",
+        testMetric.removeIfExpired(Arrays.asList("historical", "druid.test.cn", "historical1"))
+    );
+    Assert.assertFalse(
```
abhishekrb19: Thanks for these tests!

From a functional PoV, it would be better to verify the end state of the collector, that the metrics are indeed cleared, rather than calling the internal function `removeIfExpired()`; as it is, I think we don't have coverage for `cleanUpStaleMetrics()`.

What do you think of something like:

- `emitter.emit(metric1 - labelset1)`
- `emitter.emit(metric1 - labelset2)`
- `emitter.emit(metric1 - labelset1)`

verify that `metric1 - labelset1` is not cleared from the collector, whereas `metric1 - labelset2` is cleared.


I think we could leverage the `SimpleCollector`'s samples for verification. We could also pass `BlockingExecutorService` from the test to the prometheus emitter to control the executions (a bunch of tests do this to have better control of async executors). Please let me know what you think.

abhishekrb19 (commented):
Thanks @aho135! Did a high level review and left some thoughts

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
aho135: I haven't done any in depth memory profiling for this. I think it largely depends on how many unique label combinations there are. For something like queryId I'd imagine the memory usage would be higher under heavy query load. But for dimensions like taskId the memory usage should be much lower. I think it roughly gets offset anyway because we are clearing the label from the underlying SimpleCollector map now

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
aho135: Yeah good point. Felt a little dangerous to be iterating over while removing keys, but I think the consistency guarantees of the ConcurrentMap make this safe

extensions-contrib/prometheus-emitter/src/test/java/org/apache/druid/emitter/prometheus/PrometheusEmitterTest.java
aho135: Yeah agreed. Ideally I'd wanted to check SimpleCollector.children and ensure that the labels were removed by  `cleanUpStaleMetrics()`. Unfortunately the SimpleCollector doesn't expose any function to access the children Map. I did verify via the debugger that the entry is removed from the child map after the TTL though

The tests do actually run `cleanUpStaleMetrics()` because the executor is created in `emitter.start()` and manual sleep times allow the scheduled executor to run

aho135: This test case does roughly model the behavior you suggested:

Except I update metric1 - labelset2 instead

emitter.emit(metric1 - labelset1)
emitter.emit(metric1 - labelset2)
emitter.emit(metric1 - labelset2)

docs/development/extensions-contrib/prometheus.md
```diff
@@ -44,7 +44,7 @@ All the configuration parameters for the Prometheus emitter are under `druid.emi
...
-| `druid.emitter.prometheus.flushPeriod`        | When using the `pushgateway` strategy metrics are emitted every `flushPeriod` seconds. <br/>When using the `exporter` strategy this configures the metric TTL such that if the metric value is not updated within `flushPeriod` seconds then it will stop being emitted. Note that unique label combinations per metric are currently not subject to TTL expiration. It is recommended to set this to at least 3 * `scrape_interval`. | Required if `pushgateway` strategy is used, optional otherwise. | 15 seconds for `pushgateway` strategy. <br/>None for `exporter` strategy. |
+| `druid.emitter.prometheus.flushPeriod`        | When using the `pushgateway` strategy metrics are emitted every `flushPeriod` seconds. <br/>When using the `exporter` strategy this configures the metric TTL such that if the metric value is not updated within `flushPeriod` seconds then it will stop being emitted. It is recommended to set this to at least 3 * `scrape_interval`. | Required if `pushgateway` strategy is used, optional otherwise. | 15 seconds for `pushgateway` strategy. <br/>None for `exporter` strategy. |
```
abhishekrb19: Should we perhaps amend this to say that this reset also applies to unique label combinations?

extensions-contrib/prometheus-emitter/src/test/java/org/apache/druid/emitter/prometheus/PrometheusEmitterTest.java
abhishekrb19: Gotcha, the SimpleCollector [exposes](https://www.javadoc.io/doc/io.prometheus/simpleclient/0.0.5/io/prometheus/client/Collector.html#collect()) a `collect()` API to get the metrics collected/samples. Perhaps something like this should work?
```
List<Collector.MetricFamilySamples.Sample> samples = testMetric.getCollector().collect().get(0).samples;
// assert the labels & the sample size is 1 during a valid period
```
After the TTL period, the sample size would be 0? I think that would give us sufficient functionality validation on `cleanUpStaleMetrics()`

docs/development/extensions-contrib/prometheus.md
abhishekrb19: Should we perhaps amend this to say that this reset also applies to unique label combinations?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -66,22 +70,42 @@ public double[] getHistogramBuckets()
...
-  public long getMillisSinceLastUpdate()
+  public ConcurrentMap<List<String>, Stopwatch> getLabelValuesToStopwatch()
   {
-    return updateTimer.millisElapsed();
+    return labelValuesToStopwatch;
   }
 
-  public boolean isExpired()
+  /**
+   * For the given labelValues, checks if the metric value has been updated within the configured flushPeriod.
```
abhishekrb19:
```suggestion
   * For the given labelValues, checks if the metric value has been updated within the configured {@link #ttlSeconds}.
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -66,22 +70,42 @@ public double[] getHistogramBuckets()
...
+   * Returns true and removes the entry from the map if it has expired or if the entry doesn't exist, otherwise
+   * returns false.
+   */
+  public boolean shouldRemoveIfExpired(List<String> labelValues)
   {
     if (ttlSeconds == null) {
-      log.error("Invalid usage of isExpired(), TTL has not been set");
-      return false;
+      throw DruidException.defensive("Invalid usage of removeIfExpired, flushPeriod has not been set");
```
abhishekrb19:
```suggestion
      throw DruidException.defensive("Invalid usage of shouldRemoveIfExpired, ttlSeconds has not been set");
```

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -87,7 +95,7 @@ public void start()
         log.error("HTTPServer is already started");
       }
       // Start TTL scheduler if TTL is configured
-      if (config.getFlushPeriod() != null) {
+      if (config.getFlushPeriod() != null && exec == null) {
         exec = ScheduledExecutors.fixed(1, "PrometheusTTLExecutor-%s");
```
abhishekrb19: Not a blocker since this is existing code... we could perhaps move the initialization of `exec` in the constructor and make it a `final` variable. `start()` would then just do the execution start

extensions-contrib/prometheus-emitter/src/test/java/org/apache/druid/emitter/prometheus/PrometheusEmitterTest.java
```diff
@@ -493,29 +494,94 @@ public void testMetricTtlUpdate()
...
+    exec.submit(emitter::cleanUpStaleMetrics).get();
+    Assert.assertEquals(2, testMetric.getCollector().collect().get(0).samples.size());
 
-    long timeSinceLastUpdate = testMetric.getMillisSinceLastUpdate();
-    Assert.assertTrue(
-        "Update time should have been refreshed",
-        timeSinceLastUpdate < waitTime
+    // Wait for a little, but not long enough for the metric to expire
+    long waitTime = TimeUnit.SECONDS.toMillis(flushPeriod) / 5;
+    sleep(waitTime);
```
abhishekrb19: I think we can directly call `Thread.sleep()` since the unit tests already throw `InterruptedException`

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -277,7 +285,7 @@ public void setPushGateway(PushGateway pushGateway)
    * This method is called periodically by the TTL scheduler when using the 'exporter' strategy with
    * a configured flushPeriod.
    */
-  private void cleanUpStaleMetrics()
+  protected void cleanUpStaleMetrics()
```
abhishekrb19, line 290: Could we make this `public` scope with a `@VisibleForTesting` annotation? Alternatively, we can pass the `WrappingScheduledExecutorService` from tests and keep this private

aho135: Thanks for the review @abhishekrb19! 
I added the `@VisibleForTesting` annotation, but I think we can just keep the scope `protected` since it's only used in the test class

abhishekrb19:
@aho135 when you get a chance, could you please pull the latest master to this branch? That should fix the IT failures due to openjdk version.

abhishekrb19:
The failing embedded test is unrelated to this PR. It's been flakey on a few other PRs, a retry didn't help:
```
Error:  Tests run: 5, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 246.174 s <<< FAILURE! - in org.apache.druid.testing.embedded.docker.IngestionDockerTest
Error:  test_runIndexTask_andKillData  Time elapsed: 128.148 s  <<< ERROR!
org.apache.druid.java.util.common.ISE: Timed out waiting for event
```

cc: @kfaraz

kfaraz:
Thanks, @abhishekrb19 ! I think this test flakes out every now and then.
Will try to debug this soon.

PR #18692, opened 2025-10-24
Add system properties table

GabrielCWT:
This PR adds a new properties table which will call the `/status/properties` endpoint of each of the nodes.

The feature allows for an aggregated view of the properties of all services.


sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,163 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
+
+public final class SystemPropertiesTable extends AbstractTable implements ScannableTable
```
kfaraz: Please add a short javadoc here.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,163 @@
...
+{
+  public static final String PROPERTIES_TABLE = "properties";
```
kfaraz:
```suggestion
  public static final String PROPERTIES_TABLE = "server_properties";
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,163 @@
...
+
+  static final RowSignature PROPERTIES_SIGNATURE = RowSignature
+      .builder()
+      .add("service", ColumnType.STRING)
+      .add("host", ColumnType.STRING)
```
kfaraz: To avoid confusion
```suggestion
      .add("service_name", ColumnType.STRING)
      .add("server", ColumnType.STRING)
```

the `server` field should take the value `node.getHostAndPortToUse()` as several Druid services may be running on the same host.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,163 @@
...
+
+    final FluentIterable<Object[]> results = FluentIterable
+        .from(() -> druidServers)
+        .transformAndConcat((DiscoveryDruidNode discoveryDruidNode) -> {
+          final DruidNode druidNode = discoveryDruidNode.getDruidNode();
+          final Map<String, String> propertiesMap = getProperties(druidNode);
+          return propertiesMap.entrySet().stream()
+                              .map(entry -> new Object[]{
+                                  druidNode.getServiceName(),
+                                  druidNode.getHost(),
```
kfaraz: to align with `sys.servers` table and to avoid ambiguity in case multiple Druid services are running on the same host.
```suggestion
                                  druidNode.getHostAndPortToUse(),
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,163 @@
...
+    return Linq4j.asEnumerable(results);
+  }
+
+  private Map<String, String> getProperties(DruidNode druidNode)
+  {
+    try {
+      final String url = druidNode.getUriToUse().resolve("/status/properties").toString();
+      final Request request = new Request(HttpMethod.GET, new URL(url));
+      final StringFullResponseHolder response;
+      try {
```
kfaraz: Please don't nest try-catch.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,163 @@
...
+            "Failed to get properties from node at [%s]. Error code [%d], description [%s].",
+            url,
+            response.getStatus().getCode(),
+            response.getStatus().getReasonPhrase()
+        );
+      }
+      return jsonMapper.readValue(
+          response.getContent(), new TypeReference<Map<String, String>>()
+          {
+          }
```
kfaraz:
```suggestion
          response.getContent(),
          new TypeReference<>(){}
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -234,29 +236,29 @@ public SystemSchema(
...
-            jsonMapper
-        ),
-        SERVER_SEGMENTS_TABLE,
-        new ServerSegmentsTable(serverView, authorizerMapper),
-        TASKS_TABLE,
-        new TasksTable(overlordClient, authorizerMapper),
-        SUPERVISOR_TABLE,
-        new SupervisorsTable(overlordClient, authorizerMapper)
-    );
+    this.tableMap = ImmutableMap.<String, Table>builder()
```
kfaraz: Please use the original syntax `ImmutableMap.of()` and reduce the diff size here.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -1121,6 +1116,14 @@ private static void checkStateReadAccessForServers(
     }
   }
 
+  public static Iterator<DiscoveryDruidNode> getDruidServers(DruidNodeDiscoveryProvider druidNodeDiscoveryProvider)
```
kfaraz, line 1125: Please add a short javadoc to this method.

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -1448,6 +1463,129 @@ public void testSupervisorTableAuth()
...
+    final DruidNodeDiscovery coordinatorNodeDiscovery = EasyMock.createMock(DruidNodeDiscovery.class);
+    final DruidNodeDiscovery middleManagerNodeDiscovery = EasyMock.createMock(DruidNodeDiscovery.class);
+    final DruidNodeDiscovery brokerNodeDiscovery = EasyMock.createMock(DruidNodeDiscovery.class);
+    final DruidNodeDiscovery routerNodeDiscovery = EasyMock.createMock(DruidNodeDiscovery.class);
+    final DruidNodeDiscovery historicalNodeDiscovery = EasyMock.createMock(DruidNodeDiscovery.class);
+    final DruidNodeDiscovery overlordNodeDiscovery = EasyMock.createMock(DruidNodeDiscovery.class);
+    final DruidNodeDiscovery peonNodeDiscovery = EasyMock.createMock(DruidNodeDiscovery.class);
+    final DruidNodeDiscovery indexerNodeDiscovery = EasyMock.createMock(DruidNodeDiscovery.class);
+
+    EasyMock.expect(druidNodeDiscoveryProvider.getForNodeRole(NodeRole.COORDINATOR)).andReturn(coordinatorNodeDiscovery).once();
```
kfaraz: Please group the definition and expectation of each mock together.
Also, since these snippets will follow the same format, please see if they can't be put in a common private method.

Something liek:

```
final DruidNodeDiscovery coordinatorNodeDiscovery = mockNodeDiscovery(
    // list of nodes to return
    new DruidNode(...),
    new DruidNode(...)
);

final DruidNodeDiscovery overlordNodeDiscovery = mockNodeDiscovery(
       new DruidNode(...),
    new DruidNode(...)
);
```

```
private DruidNodeDiscovery mockNodeDiscovery(DruidNode... nodes)
{
   // Create the mock
   // Set expectations
   // Replay 
}
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,163 @@
...
+    );
+    SystemSchema.checkStateReadAccessForServers(authenticationResult, authorizerMapper);
+    final Iterator<DiscoveryDruidNode> druidServers = SystemSchema.getDruidServers(druidNodeDiscoveryProvider);
+
+    final FluentIterable<Object[]> results = FluentIterable
+        .from(() -> druidServers)
+        .transformAndConcat((DiscoveryDruidNode discoveryDruidNode) -> {
+          final DruidNode druidNode = discoveryDruidNode.getDruidNode();
+          final Map<String, String> propertiesMap = getProperties(druidNode);
+          return propertiesMap.entrySet().stream()
```
kfaraz: Possible improvement:
I think we could club together the entries if the same hostAndPort occurs multiple times with different node roles, e.g. this can happen when running a combined Coordinator/Overlord. Then instead of `server_type`, we could have a multi valued column `node_roles`. Otherwise, we would duplicating a bunch of properties unnecessarily.

A simple way to do that would be to have `results` be a `Map<String, Object[]>` with key as `hostAndPort` (i.e. same value as the column `server`) and add or update entries into it as applicable. Then just return `Linq4.asEnumerable(map.values())`.

GabrielCWT: Thanks for the review @kfaraz! This is a good idea, I will implement this feature.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
GabrielCWT: ImmutableMap.of() has only been overloaded to up to 10 entries. We now have 12... so I refactored to use the builder instead

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
GabrielCWT: Hi @kfaraz, I realised that if we were to combine only the roles, we would be losing out on the service_name. I have combined them into 1 column with the following format: 

`[druid/coordinator,coordinator],[druid/coordinator,overlord]`

Not too happy with the naming of the column `service_name_and_role` but it was the best I could come up with. Open to suggestions!

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
kfaraz: Yes, it can take up to 10 key value pairs.
We are currently only at 6 pairs.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
kfaraz: Can't we keep the `service_names` column separate from `node_roles`?

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
GabrielCWT: Sorry I meant 5 entries, 10 arguments (5 of each)

`of(K k1, V v1, K k2, V v2, K k3, V v3, K k4, V v4, K k5, V v5)`

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
GabrielCWT: I thought it would be clearer if we grouped the corresponding service name and role together

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
kfaraz: Yes. But newer guava versions support upto 10 kv pairs.
https://guava.dev/releases/32.1.2-jre/api/docs/com/google/common/collect/ImmutableMap.html#of(K,V,K,V,K,V,K,V,K,V,K,V,K,V,K,V,K,V,K,V)

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
GabrielCWT: Ignore what I mentioned above... I realised that if they are hosted on the same port, they cannot have different service_names

GabrielCWT:
Hi @kfaraz, when implementing the embedded test, I encountered a strange bug which I could not pinpoint the issue. I'm hoping you have more insight on this issue.

When sending the `/status/properties` request to the coordinator's `localhost:8081`, it would return the broker's properties.  I confirmed that the coordinator was initialising it's own StatusResource and Properties object. 

It seems like the coordinator is hitting the broker's endpoint even though I am using `localhost:8081`. 

When I added a router server, the coordinator would return the router's properties instead.

```
  protected EmbeddedDruidCluster createCluster()
  {
    return EmbeddedDruidCluster
        .withZookeeper()
        .addServer(new EmbeddedCoordinator())
        .addServer(overlord)
        .addServer(broker)
        .addServer(new EmbeddedRouter()) // added here
        .addCommonProperty("commonProperty", "commonValue");
  }
  ```

Note: Currently I have omitted verifying the coordinator's properties in the embedded test for the reason above.

**Comment by kfaraz** [comment]
Thanks for the update, @GabrielCWT . I will take a look at the test.

File: embedded-tests/src/test/java/org/apache/druid/testing/embedded/schema/SystemPropertiesTableTest.java [hunk 11]
```diff
@@ -0,0 +1,122 @@
...
+  private final EmbeddedBroker broker = new EmbeddedBroker()
+      .addProperty("test.onlyBroker", "brokerValue")
+      .addProperty("mytestbrokerproperty", "mytestbrokervalue")
+      .addProperty("druid.host", HOST)
+      .addProperty("druid.plaintextPort", BROKER_PORT)
+      .addProperty("druid.service", BROKER_SERVICE);
+
+  private final EmbeddedOverlord overlord = new EmbeddedOverlord()
+      .addProperty("druid.service", OVERLORD_SERVICE)
+      .addProperty("druid.host", HOST)
```
kfaraz: I don't think there is much point in trying to override the host since it always needs to be `localhost` for the test to work.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/schema/SystemServerPropertiesTableTest.java
```diff
@@ -0,0 +1,122 @@
...
+  private static final String BROKER_PORT = "8082";
+  private static final String OVERLORD_PORT = "8090";
+  private static final String BROKER_SERVICE = "test/broker";
+  private static final String OVERLORD_SERVICE = "test/overlord";
+
+  private final EmbeddedBroker broker = new EmbeddedBroker()
+      .addProperty("test.onlyBroker", "brokerValue")
+      .addProperty("mytestbrokerproperty", "mytestbrokervalue")
+      .addProperty("druid.host", HOST)
+      .addProperty("druid.plaintextPort", BROKER_PORT)
```
kfaraz, line 50: Maybe override this to a value which is not already the default port for Broker (8082).

embedded-tests/src/test/java/org/apache/druid/testing/embedded/schema/SystemPropertiesTableTest.java
```diff
@@ -0,0 +1,122 @@
...
+        new TypeReference<>(){}
+    );
+    verifyPropertiesForServer(overlordProps, OVERLORD_SERVICE, getHostAndPort(HOST, OVERLORD_PORT), NodeRole.OVERLORD_JSON_NAME);
+
+    final Map<String, String> brokerProps = cluster.callApi().serviceClient().onAnyBroker(
+        mapper -> new RequestBuilder(HttpMethod.GET, "/status/properties"),
+        new TypeReference<>(){}
+    );
+    verifyPropertiesForServer(brokerProps, BROKER_SERVICE, getHostAndPort(HOST, BROKER_PORT), NodeRole.BROKER_JSON_NAME);
+    final String test = cluster.runSql("SELECT * FROM sys.server_properties");
```
kfaraz: Did you mean to verify this result?

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -234,28 +236,31 @@ public SystemSchema(
...
-            overlordClient,
-            coordinatorClient,
-            jsonMapper
-        ),
-        SERVER_SEGMENTS_TABLE,
-        new ServerSegmentsTable(serverView, authorizerMapper),
-        TASKS_TABLE,
-        new TasksTable(overlordClient, authorizerMapper),
-        SUPERVISOR_TABLE,
-        new SupervisorsTable(overlordClient, authorizerMapper)
```
kfaraz, line 258: Nit: Please revert the formatting change here.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -234,28 +236,31 @@ public SystemSchema(
       final CoordinatorClient coordinatorClient,
       final OverlordClient overlordClient,
       final DruidNodeDiscoveryProvider druidNodeDiscoveryProvider,
-      final ObjectMapper jsonMapper
+      final ObjectMapper jsonMapper,
+      final @EscalatedClient HttpClient httpClient
```
kfaraz: Nit: to adhere to the Druid style used elsewhere
```suggestion
      @EscalatedClient final HttpClient httpClient
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,174 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * This table contains row per property. It contains all the properties of all druid servers.
```
kfaraz:
```suggestion
 * System schema table {@code sys.server_properties} that contains the properties of all Druid servers.
 * Each row contains the value of a single property. If a server has multiple node roles, all the rows for
 * that server would have multiple values in the column {@code node_roles} rather than duplicating all the
 * rows.
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,174 @@
...
+ */
+public class SystemPropertiesTable extends AbstractTable implements ScannableTable
+{
+  public static final String PROPERTIES_TABLE = "server_properties";
+
+  static final RowSignature PROPERTIES_SIGNATURE = RowSignature
```
kfaraz:
```suggestion
  static final RowSignature ROW_SIGNATURE = RowSignature
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * This table contains row per property. It contains all the properties of all druid servers.
+ */
+public class SystemPropertiesTable extends AbstractTable implements ScannableTable
+{
+  public static final String PROPERTIES_TABLE = "server_properties";
```
kfaraz:
```suggestion
  public static final String TABLE_NAME = "server_properties";
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,174 @@
...
+            url,
+            response.getStatus().getCode(),
+            response.getStatus().getReasonPhrase()
+        );
+      }
+      return jsonMapper.readValue(
+          response.getContent(),
+          new TypeReference<Map<String, String>>()
+          {
+          }
```
kfaraz:
```suggestion
          new TypeReference<>(){}
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,174 @@
...
+                                                                                                                    druidNode.getServiceName(),
+                                                                                                                    druidNode.getHostAndPortToUse(),
+                                                                                                                    entry.getKey(),
+                                                                                                                    entry.getValue()
+                                                                                                              })
+            )
+        );
+      }
+    });
+    return Linq4j.asEnumerable(serverToPropertiesMap.values().stream().flatMap(pair -> pair.rhs.map(entry -> new Object[]{entry[0], entry[1], pair.lhs.toString(), entry[2], entry[3]})).collect(Collectors.toList()));
```
kfaraz: Please check if this can be formatted better by breaking up the argument of `asEnumerable` into multiple lines.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,174 @@
...
+  public Enumerable<Object[]> scan(DataContext root)
+  {
+    final AuthenticationResult authenticationResult = (AuthenticationResult) Preconditions.checkNotNull(
+        root.get(PlannerContext.DATA_CTX_AUTHENTICATION_RESULT),
+        "authenticationResult in dataContext"
+    );
+    SystemSchema.checkStateReadAccessForServers(authenticationResult, authorizerMapper);
+    final Iterator<DiscoveryDruidNode> druidServers = SystemSchema.getDruidServers(druidNodeDiscoveryProvider);
+
+    final Map<String, Pair<StringBuilder, Stream<Object[]>>> serverToPropertiesMap = new HashMap<>();
```
kfaraz: Rather than a Pair, we can create a private class with some appropriate methods.
Also, let's avoid the `StringBuilder` and just use a list. We can serialize it to a String when converting to the `Linq4.asEnumerable()`.

```java
private class ServerProperties
{
   final String serviceName;
   final String hostAndPortToUse;
   final List<String> nodesRoles = new ArrayList<>();
   final Map<String, String> properties;
}

File: sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java [hunk 22]
```diff
@@ -0,0 +1,174 @@
...
+    try {
+      final Request request = new Request(HttpMethod.GET, new URL(url));
+      final StringFullResponseHolder response;
+      response = httpClient
+          .go(request, new StringFullResponseHandler(StandardCharsets.UTF_8))
+          .get();
+
+      if (response.getStatus().getCode() != HttpServletResponse.SC_OK) {
+        throw new RE(
+            "Failed to get properties from node at [%s]. Error code [%d], description [%s].",
```
kfaraz:
```suggestion
            "Failed to get properties from node[%s]. Error code[%d], description[%s].",
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,174 @@
...
+          response.getContent(),
+          new TypeReference<Map<String, String>>()
+          {
+          }
+      );
+    }
+    catch (IOException | InterruptedException e) {
+      throw new RuntimeException(e);
+    }
+    catch (ExecutionException e) {
```
kfaraz: Let's club these to just catch `Exception` and throw a `DruidException` since this will be sent back to the user.

kfaraz (commented):
Left some more comments.

@GabrielCWT , to fix the test, please
change this line https://github.com/apache/druid/blob/4de1909d79bddfa751b32529989cab637ae62591/services/src/main/java/org/apache/druid/cli/CoordinatorJettyServerInitializer.java#L112-L113

to be the following instead:

```java
final FilterHolder guiceFilterHolder = JettyServerInitUtils.getGuiceFilterHolder(injector);
root.addFilter(guiceFilterHolder, "/status/*", null);
```

I must have missed this one while fixing up the rest for the embedded test framework.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    public void addNodeRole(String nodeRole)
+    {
+      nodeRoles.add(nodeRole);
+    }
+
+    public Stream<Object[]> toRows()
+    {
+      return properties.entrySet().stream().map(entry -> new Object[]{serviceName, server, nodeRoles.toString(), entry.getKey(), entry.getValue()});
```
FrankChen021: move the `nodeRoles.toString()` out of the stream

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * System schema table {@code sys.server_properties} that contains the properties of all Druid servers.
+ * Each row contains the value of a single property. If a server has multiple node roles, all the rows for
+ * that server would have multiple values in the column {@code node_roles} rather than duplicating all the
+ * rows.
+ */
+public class SystemPropertiesTable extends AbstractTable implements ScannableTable
```
abhishekrb19: Could we call this `ServerPropertiesTable` to be clearer?  (similar to `ServerSegmentsTable` for `server_segments` table)

abhishekrb19:
This is cool! The hidden or secret properties on each server wouldn’t be listed in this systems table as it goes through the `/status/properties`, right? I think it would be good to add a test for that. Could you also please update the docs for this table here https://druid.apache.org/docs/latest/querying/sql-metadata-tables/#system-schema?

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,191 @@
...
+      return jsonMapper.readValue(
+          response.getContent(),
+          new TypeReference<Map<String, String>>(){}
+      );
+    }
+    catch (Exception e) {
+      throw DruidException.forPersona(DruidException.Persona.USER)
+                          .ofCategory(DruidException.Category.UNCATEGORIZED)
+                          .build(e, "HTTP request to[%s] failed", url);
+    }
```
kfaraz: Maybe use `InternalServerError.exception()` instead.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,191 @@
...
+        throw new RE(
+            "Failed to get properties from node[%s]. Error code[%d], description[%s].",
+            url,
+            response.getStatus().getCode(),
+            response.getStatus().getReasonPhrase()
+        );
+      }
+      return jsonMapper.readValue(
+          response.getContent(),
+          new TypeReference<Map<String, String>>(){}
```
kfaraz: I think this can be omitted.
```suggestion
          new TypeReference<>(){}
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,191 @@
...
+
+    final Map<String, ServerProperties> serverToPropertiesMap = new HashMap<>();
+    druidServers.forEachRemaining(discoveryDruidNode -> {
+      final DruidNode druidNode = discoveryDruidNode.getDruidNode();
+      final Map<String, String> propertiesMap = getProperties(druidNode);
+      if (serverToPropertiesMap.containsKey(druidNode.getHostAndPortToUse())) {
+        ServerProperties serverProperties = serverToPropertiesMap.get(druidNode.getHostAndPortToUse());
+        serverProperties.addNodeRole(discoveryDruidNode.getNodeRole().getJsonName());
+      } else {
+        serverToPropertiesMap.put(druidNode.getHostAndPortToUse(), new ServerProperties(druidNode.getServiceName(), druidNode.getHostAndPortToUse(), new ArrayList<>(Arrays.asList(discoveryDruidNode.getNodeRole().getJsonName())), propertiesMap));
```
kfaraz: formatting:
```suggestion
        serverToPropertiesMap.put(
            druidNode.getHostAndPortToUse(),
            new ServerProperties(
                druidNode.getServiceName(),
                druidNode.getHostAndPortToUse(),
                new ArrayList<>(Arrays.asList(discoveryDruidNode.getNodeRole().getJsonName())),
                propertiesMap
            )
        );
```

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,191 @@
...
+      this.nodeRoles = nodeRoles;
+      this.properties = properties;
+    }
+
+    public void addNodeRole(String nodeRole)
+    {
+      nodeRoles.add(nodeRole);
+    }
+
+    public Stream<Object[]> toRows()
```
kfaraz: this method should return a `List<Object[]>` rows since we are going to always convert it to a list anyway.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
```diff
@@ -0,0 +1,191 @@
...
+ * rows.
+ */
+public class SystemPropertiesTable extends AbstractTable implements ScannableTable
+{
+  public static final String TABLE_NAME = "server_properties";
+
+  static final RowSignature ROW_SIGNATURE = RowSignature
+      .builder()
+      .add("service_name", ColumnType.STRING)
+      .add("server", ColumnType.STRING)
```
kfaraz: Nit: Please make `server` the first column since it is unique and serves as the "id" in this table.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/schema/SystemPropertiesTableTest.java
```diff
@@ -0,0 +1,125 @@
...
+    return EmbeddedDruidCluster
+        .withZookeeper()
+        .addServer(coordinator)
+        .addServer(overlord)
+        .addServer(broker)
+        .addCommonProperty("commonProperty", "commonValue");
+  }
+
+  @Test
+  public void test_serverPropertiesTable()
```
kfaraz: - Please add a separate test method for each service.
- Please add another test method which verifies the values of specific properties using some WHERE clauses.

e.g.

```sql
1. SELECT value FROM sys.server_properties WHERE server = xyz AND property = abc;
2. SELECT value FROM sys.server_properties WHERE service_name = xyz AND property = abc;
3. SELECT * FROM sys.server_properties WHERE service_name = xyz AND property = abc;
```

Example shorthand snippet:

```java
Assertions.assertEquals(
    "expectedValue",
    cluster.runSql("SELECT value FROM sys.server_properties WHERE server = xyz AND property = abc")
);
Assertions.assertEquals(
    "localhost:1234,test/overlord,overlord,abc,propertyValue",
    cluster.runSql("SELECT * FROM sys.server_properties WHERE server = xyz AND property = abc")
);
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/schema/SystemPropertiesTableTest.java
```diff
@@ -0,0 +1,125 @@
...
+    String[] expectedRows = properties.entrySet().stream().map(entry -> String.join(
+        ",",
+        escapeCsvField(serivceName),
+        escapeCsvField(hostAndPort),
+        escapeCsvField(ImmutableList.of(nodeRole).toString()),
+        escapeCsvField(entry.getKey()),
+        escapeCsvField(entry.getValue())
+    )).toArray(String[]::new);
+    Arrays.sort(expectedRows, String::compareTo);
+    String[] actualRows = Arrays.stream(cluster.runSql("SELECT * FROM sys.server_properties WHERE server='%s'", hostAndPort).split("\n")).map(entry -> StringUtils.replace(entry, "...", "\"\"")).toArray(String[]::new);
```
kfaraz: Nit: break up the line for readability

Also, is the removal of ellipsis (`...`) required or was it copied over from some other test?

```suggestion
    final String result = cluster.runSql("SELECT * FROM sys.server_properties WHERE server='%s'", hostAndPort);
    String[] actualRows = result.split("\n");
```

kfaraz (commented):
Left some minor suggestions. Changes look good.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemPropertiesTable.java
GabrielCWT: Will rename it to `SystemServerPropertiesTable` instead to maintain the idea that it is a system table

docs/querying/sql-metadata-tables.md
```diff
@@ -315,3 +316,21 @@ For example, to retrieve supervisor tasks information filtered by health status,
 ```sql
 SELECT * FROM sys.supervisors WHERE healthy=0;
 ```
+
+### SERVER_PROPERTIES table
+
+The `server_properties` table exposes configuration properties for each Druid server. Each row represents a single property key-value pair associated with a specific server.
```
```suggestion
The `server_properties` table exposes the runtime properties configured on each Druid server. Each row represents a single property key-value pair associated with a specific server.
```

docs/querying/sql-metadata-tables.md
```diff
@@ -315,3 +316,21 @@ For example, to retrieve supervisor tasks information filtered by health status,
...
+
+|Column|Type|Notes|
+|------|-----|-----|
+|server|VARCHAR|Host and port of the server, in the form host:port|
```
```suggestion
|server|VARCHAR|Host and port of the server, in the form `host:port`|
```

GabrielCWT:
> We should also add the test suggested by @abhishekrb19 , but that is not a blocker for this PR and can be done in a follow up. 

@kfaraz I've added a minimal test to check if hidden properties are exposed. 

Also do we still need tests in `SystemSchemaTest` since we implemented the tests in `SystemServerPropertiesTableTest`?

kfaraz:
> @kfaraz I've added a minimal test to check if hidden properties are exposed.

Thanks, @GabrielCWT !

> Also do we still need tests in SystemSchemaTest since we implemented the tests in SystemServerPropertiesTableTest?

Yes, since you have already added those UTs, let's keep them.
Embedded tests are primarily for integration testing.
Method/class level functional testing should still be done in UTs.

docs/querying/sql-metadata-tables.md
```diff
@@ -315,3 +316,21 @@ For example, to retrieve supervisor tasks information filtered by health status,
...
+|server|VARCHAR|Host and port of the server, in the form `host:port`|
+|service_name|VARCHAR|Service name of the server, as defined by `druid.service`|
+|node_roles|VARCHAR|Comma-separated list of roles that the server performs. For example, `[coordinator,overlord]` if the server functions as both a Coordinator and an Overlord.|
+|property|VARCHAR|Name of the property|
+|value|VARCHAR|Value of the property|
+
+For example, to retrieve properties for a specific server, use the query
+
+```sql
+SELECT * FROM sys.server_properties WHERE server='123.456.789.1:8081'
```
FrankChen021:
```suggestion
SELECT * FROM sys.server_properties WHERE server='192.168.1.1:8081'
```

kfaraz:
@FrankChen021 , @GabrielCWT , are any further changes pending in this PR?

GabrielCWT:
No other changes @kfaraz

PR #18696, opened 2025-10-24
atomic move segment files to staging location prior to delete

clintropolis:
This PR changes how segment files are deleted from the cache by introducing an atomic move to a staging directory before performing the delete. This ensures that there will not be partial segment files hanging out in the event of a failure during a delete operation, and the staging directory is cleaned up on process start.

The staging directory is in the root of the storage location, so if the segment cache path for a segment is something like 
```
.../segment_cache_folder/test_2019-01-01T00:00:00.000Z_2019-01-02T00:00:00.000Z_2025-10-24T23:44:38.503Z_5
```
it would be moved to 
```
.../segment_cache_folder/__drop/test_2019-01-01T00:00:00.000Z_2019-01-02T00:00:00.000Z_2025-10-24T23:44:38.503Z_5
```
prior to deleting.

### DIFF
--- a/server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
+++ b/server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
@@ -205,7 +205,7 @@ public class SegmentLocalCacheManager implements SegmentCacheManager
               FileUtils.deleteDirectory(droppedFile);
             }
             catch (Exception e) {
-              log.error(e, "Unable to remove directory[%s]", droppedFile);
+              log.warn(e, "Unable to remove dropped segment directory[%s]", droppedFile);
             }
           }
         }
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -747,21 +768,30 @@ private SegmentCacheEntry assignLocationAndMount(
...
+    final File parent = path.getParentFile();
+    final File tempLocation = new File(parent, DROP_PATH);
     try {
-      FileUtils.deleteDirectory(path);
+      if (!tempLocation.exists()) {
+        FileUtils.mkdirp(tempLocation);
+      }
+      final File tempPath = new File(tempLocation, path.getName());
+      log.debug("moving[%s] to temp location[%s]", path, tempLocation);
+      Files.move(path.toPath(), tempPath.toPath(), StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
```
gianm: REPLACE_EXISTING doesn't work for directories, but it might be a no-op. At any rate you might as well remove it.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -771,7 +801,13 @@ private static void cleanupLegacyCacheLocation(final File baseFile, final File c
       return;
     }
 
-    deleteCacheEntryDirectory(cacheFile);
+    try {
+      log.info("Deleting directory[%s]", cacheFile);
+      FileUtils.deleteDirectory(cacheFile);
+    }
+    catch (Exception e) {
+      log.error(e, "Unable to remove directory[%s]", cacheFile);
```
gianm: should probably be `warn`, doesn't seem serious enough for `error`

cryptoe:
@clintropolis Is this a Druid 35 blocker ?

PR #18699, opened 2025-10-28
Fix extra router statusCode metric edge cases

jtuglu1:
### Description

`statusCode` metrics will misreport if connection exceptions occur while router/broker are communicating (e.g. there's no HTTP status code yet – Jetty sets defaults this to `0`) or when a client manually closes the connection while a downstream query is still in-flight. We cannot perfectly map every error in this case (nor do I think we should), but it's better than misreporting incorrect codes. Since we need to report a metric event anyways, maintaining a fallback of 500/200 if we cannot determine the error code is consistent with how things are handled already on the `.*QueryResource` classes and router should do the same.

#### Release note

Fix connection error statusCode metrics reporting on router

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
+++ b/services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
@@ -107,7 +107,8 @@ public class AsyncQueryForwardingServlet extends AsyncProxyServlet implements Qu
   private static final String PROPERTY_SQL_ENABLE_DEFAULT = "false";
 
   private static final long CANCELLATION_TIMEOUT_MILLIS = TimeUnit.SECONDS.toMillis(5);
-  private static final int UNASSIGNED_DEFAULT_STATUS_CODE = 0; // Jetty-specific default (un-assigned) status code
+  // Jetty-specific default (un-assigned) status code
+  private static final int UNASSIGNED_DEFAULT_STATUS_CODE = 0;
 
   private final AtomicLong successfulQueryCount = new AtomicLong();
   private final AtomicLong failedQueryCount = new AtomicLong();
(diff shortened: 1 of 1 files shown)


services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
```diff
@@ -107,6 +107,7 @@ public class AsyncQueryForwardingServlet extends AsyncProxyServlet implements Qu
   private static final String PROPERTY_SQL_ENABLE_DEFAULT = "false";
 
   private static final long CANCELLATION_TIMEOUT_MILLIS = TimeUnit.SECONDS.toMillis(5);
+  private static final int UNASSIGNED_DEFAULT_STATUS_CODE = 0; // Jetty-specific default (un-assigned) status code
```
kfaraz:
```suggestion
  // Jetty-specific default (un-assigned) status code
  private static final int UNASSIGNED_DEFAULT_STATUS_CODE = 0;
```

services/src/main/java/org/apache/druid/server/AsyncQueryForwardingServlet.java
```diff
@@ -857,8 +858,9 @@ public void onFailure(Response response, Throwable failure)
       failedQueryCount.incrementAndGet();
 
       // As router is simply a proxy, we don't make an effort to construct the error code from the exception ourselves.
-      // We rely on broker to set this for us if the error occurs downstream.
-      emitQueryTime(requestTimeNs, false, sqlQueryId, queryId, response.getStatus());
+      // We rely on broker to set this for us if the error occurs downstream. Otherwise, if there's a router/client error, we log this as an ISE.
```
kfaraz: ISE might be confused with the Druid `ISE` (flavor of IllegalStateException).

```suggestion
      // We rely on broker to set this for us if the error occurs downstream.
      // Otherwise, if there's a router/client error, we log this as an unknown error.
```

PR #18703, opened 2025-10-30
consolidate duplicate implementations of ServerManagerForQueryErrorTest and CliHistoricalForQueryErrorTest

clintropolis:
Moved the implementation of `integration-tests/src/main/java/org/apache/druid/server/coordination/ServerManagerForQueryErrorTest.java` to integration-tests-ex/tools/src/main/java/org/apache/druid/testing/tools/ServerManagerForQueryErrorTest.java since the former was the one apparently being used by still running tests, but the latter feels more portable in the tools module


kfaraz (approved):
Thanks for the cleanup, @clintropolis !

PR #18704, opened 2025-10-30
fix flaky SegmentLocalCacheManagerConcurrencyTest

clintropolis:
flaky caused by test not being updated in #18696 (the extra __drop directory can appear in the location)


server/src/test/java/org/apache/druid/segment/loading/SegmentLocalCacheManagerConcurrencyTest.java
```diff
@@ -428,8 +428,8 @@ public void testAcquireSegmentOnDemandRandomSegmentWithTimeoutBeforeAcquire() th
     Assertions.assertEquals(0, location2.getActiveWeakHolds());
     Assertions.assertTrue(4 >= location.getWeakEntryCount());
     Assertions.assertTrue(4 >= location2.getWeakEntryCount());
-    Assertions.assertTrue(4 >= location.getPath().listFiles().length);
-    Assertions.assertTrue(4 >= location2.getPath().listFiles().length);
+    Assertions.assertTrue(5 >= location.getPath().listFiles().length);
```
kfaraz, line 432: Please add a one-line comment on why we expect the count to be 5.

kfaraz (approved):

PR #18708, opened 2025-10-31
Remove various version overrides from module poms.

gianm:
Many poms explicitly specified versions for dependencies and plugins that have centrally-managed versions in the master pom's dependencyManagement and pluginManagement section.

For the most part, the overrides were simply restating the centrally-managed version, so they weren't causing an actual divergence. In these cases, removing the overrides simply improves maintainability by promoting the usages of dependencyManagement and pluginManagement.

The most substantive change is that hamcrest is upgraded from 1.3 to 2.2. Version 2.2 was already being used by kubernetes-overlord-extensions.

### DIFF
--- a/extensions-contrib/moving-average-query/pom.xml
+++ b/extensions-contrib/moving-average-query/pom.xml
@@ -115,5 +115,10 @@
       <artifactId>hamcrest-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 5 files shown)


jtuglu1:
I wonder if there's a way to do this with Druid versions too? Instead of stating `36.0.0-SNAPSHOT` everywhere, just define it once in the root pom?

kfaraz (approved):
Thanks for the cleanup, @gianm !
Should we also consider adding a checkstyle rule that ensures that future PRs do not add explicit versions?

kfaraz:
> I wonder if there's a way to do this with Druid versions too? Instead of stating 36.0.0-SNAPSHOT everywhere, just define it once in the root pom?

yeah, that would be nice too.

gianm:
> I wonder if there's a way to do this with Druid versions too? Instead of stating `36.0.0-SNAPSHOT` everywhere, just define it once in the root pom?

I'm not sure if there's a way to do that while still preserving maven's ability to do `release:prepare` and `release:perform`. It has some automation for updating all of these versions in multi-module projects.

> Should we also consider adding a checkstyle rule that ensures that future PRs do not add explicit versions?

I actually wanted to do this but wasn't sure how, so I skipped it. Do you happen to know how to implement a check like this for `pom.xml`?

PR #18709, opened 2025-10-31
Add dataSource and taskId dimensions to GroupByStatsMonitor for peons

abhishekrb19:
The `GroupByStatsMonitor` helps highlight potential performance bottlenecks on the query and data nodes. For peons, it’s useful to include the `dataSource` and `taskId` dimensions to get an aggregated view of merge buffer acquisition times, the number of merge buffer queries per datasource, etc. This is useful for diagnosing identifying bottlenecks on the peons when there's a flurry of realtime group By queries to different tables, so the default tunings are not sufficient.

The existing utility `MontiorsConfig.mapOfDatasourceAndTaskID()` adds `id` as the dimension for task id. This might be okay for jvm and other sys metrics on the peons, but it's certainly confusing for the group by stats. So I've added a new method `mapOfDatasourceAndTaskIdDimension()` to add the `taskId` dimension and retained the old one.

Also, added missing `@Nullable` annotations to `DataSourceTaskIdHolder` and its callers as those are optional.

### Release note
Monitors on peons that previously emitted the `id` dimension from `JettyMonitor`, `OshiSysMonitor`, `JvmMonitor`, `JvmCpuMonitor`, `JvmThreadsMonitor` and `SysMonitor` to represent the task ID are deprecated and will be removed in a future release. Use the `taskId` dimension instead.

Also, the `GroupByStatsMonitor` includes `dataSource` and `taskId` dimensions for metrics emitted on the peons.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/integration-tests/docker/Dockerfile
+++ b/integration-tests/docker/Dockerfile
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG JDK_VERSION=17-slim-bullseye
-FROM openjdk:$JDK_VERSION as druidbase
+ARG JDK_VERSION=21-jre-jammy
+FROM eclipse-temurin:$JDK_VERSION AS druidbase
 
 # Bundle everything into one script so cleanup can reduce image size.
 # Otherwise docker's layered images mean that things are not actually deleted.
--- a/integration-tests/docker/base-setup.sh
+++ b/integration-tests/docker/base-setup.sh
@@ -27,8 +27,8 @@ apt-get update
 # wget
 apt-get install -y wget
 
-# MySQL (Metadata store), modern Debian-based systems install mariadb-server
-apt-get install -y default-mysql-server
+# MariaDB metadata store
+apt-get install -y mariadb-server
 
 # Supervisor
 apt-get install -y supervisor
--- a/integration-tests/script/docker_build_containers.sh
+++ b/integration-tests/script/docker_build_containers.sh
@@ -34,7 +34,7 @@ else
   11 | 17 | 21)
     echo "Build druid-cluster with Java $DRUID_INTEGRATION_TEST_JVM_RUNTIME"
     docker build -t druid/cluster \
-      --build-arg JDK_VERSION=$DRUID_INTEGRATION_TEST_JVM_RUNTIME-slim-bullseye \
+      --build-arg JDK_VERSION=$DRUID_INTEGRATION_TEST_JVM_RUNTIME-jre-jammy \
       --build-arg ZK_VERSION \
       --build-arg KAFKA_VERSION \
       --build-arg CONFLUENT_VERSION \


server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerModule.java
```diff
@@ -526,7 +527,7 @@ public static class JettyMonitor extends AbstractMonitor
   {
     private final Map<String, String[]> dimensions;
 
-    public JettyMonitor(String dataSource, String taskId)
+    public JettyMonitor(@Nullable String dataSource, @Nullable String taskId)
```
kfaraz: The `JettyMonitor` is used only for peons. Why should this be a nullable field?

server/src/main/java/org/apache/druid/server/metrics/DataSourceTaskIdHolder.java
```diff
@@ -46,11 +48,13 @@ public class DataSourceTaskIdHolder
   @Inject(optional = true)
   BroadcastDatasourceLoadingSpec broadcastDatasourceLoadingSpec = BroadcastDatasourceLoadingSpec.ALL;
 
+  @Nullable
```
kfaraz, line 58: When can this be null?

server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerModule.java
abhishekrb19: Isn’t the `JettyMonitor` applicable to all servers, not just peons? It’s bound through the `JettyServerModule` [here](https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/initialization/CoreInjectorBuilder.java#L99).

server/src/main/java/org/apache/druid/server/metrics/DataSourceTaskIdHolder.java
abhishekrb19: For peons, the `CliPeon` would [provide](https://github.com/apache/druid/blob/master/services/src/main/java/org/apache/druid/cli/CliPeon.java#L334) the values of `taskId` and `dataSource`.

For all other servers, this would be null as these fields are initialized to null::
```java
  @Named(DATA_SOURCE_BINDING)
  @Inject(optional = true)
  String dataSource = null;
  @Named(TASK_ID_BINDING)
  @Inject(optional = true)
  String taskId = null;
```

server/src/main/java/org/apache/druid/server/initialization/jetty/JettyServerModule.java
kfaraz: Ah, of course! Thanks for the clarification.
I don't know why I assumed this class was being used only by the peons. I have myself depended on the `jetty/*` metrics several times. 😅 .
I think it must have been the presence of the `DatasourceTaskIdHolder` in the call site which put me in the peon mindset. 😛 

---

Since other services are going to use it too, ~let's add a zero-arg constructor to this class too.~, please accept the dimensions map in the constructor instead of an explicit datasource and taskID, since these two fields are very specific to peons.

server/src/main/java/org/apache/druid/server/metrics/DataSourceTaskIdHolder.java
kfaraz: Please add a short javadoc on these getters mentioning that these are non-null only for CliPeon servers.

In a follow up PR, we should also fix up the way that `DatasourceTaskIdHolder` is wired and bind that class itself only for peons. There is no meaning in having this instance bound but contain null values for datasource and task ID.

server/src/main/java/org/apache/druid/server/metrics/GroupByStatsMonitor.java
```diff
@@ -41,21 +43,28 @@ public class GroupByStatsMonitor extends AbstractMonitor
...
   private final GroupByStatsProvider groupByStatsProvider;
   private final BlockingPool<ByteBuffer> mergeBufferPool;
+  private final Map<String, String[]> dimensions;
 
   @Inject
   public GroupByStatsMonitor(
       GroupByStatsProvider groupByStatsProvider,
-      @Merging BlockingPool<ByteBuffer> mergeBufferPool
+      @Merging BlockingPool<ByteBuffer> mergeBufferPool,
+      DataSourceTaskIdHolder dataSourceTaskIdHolder
```
kfaraz, line 52: Please pass a dimensions map here instead of this holder since the holder class has meaning only for peons.

server/src/main/java/org/apache/druid/server/metrics/GroupByStatsMonitor.java
```diff
@@ -41,21 +43,28 @@ public class GroupByStatsMonitor extends AbstractMonitor
 {
   private final GroupByStatsProvider groupByStatsProvider;
   private final BlockingPool<ByteBuffer> mergeBufferPool;
+  private final Map<String, String[]> dimensions;
```
kfaraz, line 46: Maybe rename to `serviceDimensions`?

In fact, I wonder if it wouldn't be cleaner to just bind the datasource and taskID as `@ExtraServiceDimensions` so that they are used in `EmitterModule` and are emitted for all metrics emitted by a peon.

server/src/main/java/org/apache/druid/server/metrics/DataSourceTaskIdHolder.java
abhishekrb19: Added javadoc!

>In a follow up PR, we should also fix up the way that DatasourceTaskIdHolder is wired and bind that class itself only for peons. There is no meaning in having this instance bound but contain null values for datasource and task ID.


Yeah, there's some [commentary](https://github.com/apache/druid/blame/master/server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java#L115) on why it's setup the way it is. I think we'll need to revisit this Guice bindings for this.

server/src/main/java/org/apache/druid/server/metrics/GroupByStatsMonitor.java
abhishekrb19: I can rename, but I left this field as `dimensions` to be consistent with the other monitors.

>I wonder if it wouldn't be cleaner to just bind the datasource and taskID as @ExtraServiceDimensions so that they are used in EmitterModule and are emitted for all metrics emitted by a peon.


This is a good idea! I will try to take a stab at it separately and put up a PR as it affects all monitors - we'll need to [rework](https://github.com/apache/druid/pull/18709/files#r2488395049) the Guice bindings and likely need to reconcile how the extra service dimensions will be bound to monitors in a backwards-compatible manner too (`id` vs `taskId` is something to consider).

server/src/main/java/org/apache/druid/server/metrics/GroupByStatsMonitor.java
abhishekrb19: Since this monitor is injected, I don’t think it can accept a `Map<String, String[]>` in the constructor unless it’s bound with a specific name or qualifier?

server/src/main/java/org/apache/druid/server/metrics/DataSourceTaskIdHolder.java
kfaraz: Yeah, given that we have embedded tests now, it should be easy to fix as we can iterate much faster.

server/src/main/java/org/apache/druid/server/metrics/GroupByStatsMonitor.java
kfaraz: Yeah, we can leave that for now. When we fix up the binding of `DatasourceTaskIdHolder`, we can figure that out.

server/src/main/java/org/apache/druid/server/metrics/GroupByStatsMonitor.java
kfaraz: > I can rename, but I left this field as dimensions to be consistent with the other monitors.

I see, thanks for the clarification.

docs/operations/metrics.md
```diff
@@ -131,13 +131,13 @@ Most metric values reset each emission period, as specified in `druid.monitoring
...
-|`mergeBuffer/acquisitionTimeNs`|Total time in nanoseconds to acquire merge buffer for groupBy queries.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
-|`groupBy/spilledQueries`|Number of groupBy queries that have spilled onto the disk.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
-|`groupBy/spilledBytes`|Number of bytes spilled on the disk by the groupBy queries.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
-|`groupBy/mergeDictionarySize`|Size of on-heap merge dictionary in bytes.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
+|`mergeBuffer/pendingRequests`|Number of requests waiting to acquire a batch of buffers from the merge buffer pool.|`dataSource`, `taskId`. This metric is only available if the `GroupByStatsMonitor` module is included.|Should be ideally 0, though a higher number isn't representative of a problem.|
```
```suggestion
|`mergeBuffer/pendingRequests`|Number of requests waiting to acquire a batch of buffers from the merge buffer pool. This metric is only available if the `GroupByStatsMonitor` module is included.| `dataSource`, `taskId`|Should be ideally 0, though a higher number isn't representative of a problem.|
```

kfaraz: same change might be needed in other lines.

server/src/main/java/org/apache/druid/server/metrics/MonitorsConfig.java
```diff
@@ -79,14 +80,44 @@ public String toString()
            '}';
   }
 
-  public static Map<String, String[]> mapOfDatasourceAndTaskID(final String datasource, final String taskId)
```
kfaraz, line 82: Instead of adding a new method, let's do the following instead:
- use the old method itself to emit the task ID under both the `id` dimension and `taskId` dimension
- call out the `id` dimension as deprecated in the docs and release notes and remove it in D37
- update the docs for the existing metrics to indicate that they emit `taskId` dimension too.

docs/operations/metrics.md
abhishekrb19: Updated

server/src/main/java/org/apache/druid/server/metrics/MonitorsConfig.java
abhishekrb19: Done. 

>call out the id dimension as deprecated in the docs and release notes and remove it in D37

It looks like the [docs](https://druid.apache.org/docs/latest/operations/metrics/#jetty) for the six monitors that use this method don’t specify their dimensions clearly. I’ve added a general info note under the Real-time metrics seciton. Please let me know if I missed anything.

docs/operations/metrics.md
```diff
@@ -119,6 +119,11 @@ Most metric values reset each emission period, as specified in `druid.monitoring
 
 ### Real-time
 
+:::info
+Monitors on peons that previously emitted the `id` dimension from `JettyMonitor`, `OshiSysMonitor`, `JvmMonitor`, `JvmCpuMonitor`, `JvmThreadsMonitor` and `SysMonitor` 
+to represent the task ID are deprecated and will be removed in a future release. Use the `taskId` dimension instead.
```
kfaraz, line 124:
```suggestion
The dimension `id` emitted by peons to represent the task ID (by monitors such as `JettyMonitor`, `OshiSysMonitor`, `JvmMonitor`, `JvmCpuMonitor`, `JvmThreadsMonitor` and `SysMonitor`) 
is now deprecated and will be removed in a future release. Use the `taskId` dimension instead.
```

kfaraz (approved):

abhishekrb19:
Thanks for the review, @kfaraz! I will incorporate the doc suggestion in some other change.

PR #18715, opened 2025-11-04
Do not clear `pendingCompletionTaskGroups` in `clearAllocationInfo`

amaechler:
### Description

Fixes a bug where the SeekableStream supervisor autoscaler creates duplicate history entries every `minTriggerScaleActionFrequencyMillis` (default 10min) during scale-down operations, causing database pollution and preventing scale-down from completing.

_Lots of help from Claude._

#### Problem

When the autoscaler scales down tasks, `clearAllocationInfo()` prematurely clears `pendingCompletionTaskGroups`, causing the supervisor to "forget" about tasks transitioning from READING to PUBLISHING state. On the next supervisor cycle, these tasks are rediscovered and re-added to `activelyReadingTaskGroups`, triggering another scale-down attempt and creating a duplicate history entry. This repeats every `minTriggerScaleActionFrequencyMillis` (default: 10 minutes). I saw hundreds of duplicate history entries, with entries created at exact 10-minute intervals.

The root cause is that the autoscaler has a built-in safeguard (line 480-496) to skip scale actions when `pendingCompletionTaskGroups` is non-empty, but this check is ineffective because `clearAllocationInfo()` clears the map immediately after tasks were moved there.

#### Solution

Preserve `pendingCompletionTaskGroups` in `clearAllocationInfo()`. This allows the autoscaler's existing skip logic to function correctly, preventing duplicate scale attempts until tasks naturally complete (removed by `checkPendingCompletionTasks()` every supervisor cycle).

#### Release note

Fixed a bug in the SeekableStream supervisor autoscaler where scale-down operations would create duplicate supervisor history entries. The autoscaler now correctly waits for tasks to complete before attempting subsequent scale operations.

<hr>

* `SeekableStreamSupervisor` - Modified `clearAllocationInfo()` to preserve `pendingCompletionTaskGroups`

<hr>

This PR has:

* [x] been self-reviewed.
* [ ] added documentation for new or modified features or behaviors.
* [x] a release note entry in the PR description.
* [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
* [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
* [x] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
* [x] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
* [ ] added integration tests.
* [ ] been tested in a test Druid cluster.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
@@ -621,9 +621,6 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
     activelyReadingTaskGroups.clear();
     partitionGroups.clear();
     partitionOffsets.clear();
-
-    // Note: We intentionally do NOT clear pendingCompletionTaskGroups here.
-
     partitionIds.clear();
   }
 


extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
```diff
@@ -5047,6 +5050,216 @@ public void test_doesTaskMatchSupervisor()
...
+   * to pendingCompletionTaskGroups, then calls clearAllocationInfo(). If clearAllocationInfo()
+   * clears pendingCompletionTaskGroups, the supervisor "forgets" about publishing tasks.
+   * During the next discoverTasks(), these tasks get rediscovered and re-added to
+   * activelyReadingTaskGroups, causing the autoscaler to repeatedly attempt scale-down
+   * and create duplicate history entries.
+   * <p>
+   * Fix: clearAllocationInfo() must preserve pendingCompletionTaskGroups so the autoscaler's
+   * built-in check (DynamicAllocationTasksNotice.handle()) can skip scale actions while
+   * tasks are completing.
+   */
```
kfaraz: This can be omitted as it is an implementation detail.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
```diff
@@ -5047,6 +5050,216 @@ public void test_doesTaskMatchSupervisor()
...
+  @Test
+  public void test_clearAllocationInfo_preservesPendingCompletionTaskGroups() throws Exception
```
kfaraz: Rather than verifying the internal data structures of the supervisor, the test should try to verify the behaviour of the auto-scaler.


```suggestion
  public void test_autoScaler_doesNotRepeatScaleDownActions_ifTasksAreStillPublishing() throws Exception
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -614,7 +614,15 @@ private void clearAllocationInfo()
...
 
-    pendingCompletionTaskGroups.clear();
+    // Note: We intentionally do NOT clear pendingCompletionTaskGroups here.
+    // When the autoscaler calls this method after gracefulShutdownInternal(), tasks have been moved to
+    // pendingCompletionTaskGroups and are transitioning from READING -> PUBLISHING state.
+    // If we clear pendingCompletionTaskGroups, the supervisor "forgets" about these publishing tasks.
+    // Then, during the next discoverTasks(), if tasks haven't transitioned yet, they get re-added to
+    // activelyReadingTaskGroups, causing the autoscaler to repeatedly attempt scale-down and create
+    // duplicate supervisor history entries. By preserving pendingCompletionTaskGroups, the autoscaler's
+    // check at DynamicAllocationTasksNotice.handle() will correctly skip scale actions until tasks complete.
```
kfaraz: Tried to make it a little concise. Also please move it to the javadoc of this method.


```suggestion
    /**
     * Does not clear {@link #pendingCompletionTaskGroups} so that the supervisor remembers that these
     * tasks are publishing and auto-scaler does not repeatedly attempt a scale down until these tasks
     * complete. If this is cleared, the next {@link #discoverTasks()} might add these tasks to
     * {@link #activelyReadingTaskGroups}.
     */
```

kfaraz (commented):
Thanks for the fix, @amaechler !

I have left a couple of minor suggestions.
We might want to add an embedded test for this race condition. But that need not block this bugfix.

amaechler:
Thanks @kfaraz for taking the time to review! I updated the wording a bit based on your suggestions. I'm not sure about how I could rewrite the test to be more high-level, so I kept the actual test for now.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -608,13 +608,22 @@ private void changeTaskCountInIOConfig(int desiredActiveTaskCount)
...
+   * {@link #activelyReadingTaskGroups}.
+   */
   private void clearAllocationInfo()
   {
     activelyReadingTaskGroups.clear();
     partitionGroups.clear();
     partitionOffsets.clear();
 
-    pendingCompletionTaskGroups.clear();
+    // Note: We intentionally do NOT clear pendingCompletionTaskGroups here.
```
kfaraz: Since the original line of code has already been removed, this comment seems out of place here. It has already been called out in the javadoc anyway.

kfaraz:
> I'm not sure about how I could rewrite the test to be more high-level, so I kept the actual test for now.

Fair enough, we can address that and add an embedded test in follow up PRs.

PR #18717, opened 2025-11-05
Run JDK 21 tests on PRs

abhishekrb19:
The JDK 21 tests don't currently run by default on PRs. One of the `CompressionUtilsTest` cases has been failing in [master](https://github.com/apache/druid/commits/master/) and the [35.0.0](https://github.com/apache/druid/commits/35.0.0) release branches for ~3 weeks, likely due to the bug fixed in latest JDK version 21.0.9. This test has been disabled for now; we can perhaps remove the test as well as it doesn't seem to test the internals of `CompressionUtils`.

With [#17876](https://github.com/apache/druid/pull/17876), PRs require a label to trigger all JDK tests on PRs (this runs by default on merge commits). I believe this was introduced to reduce CI runtime when we previously tested across three JDK versions. Since JDK 11 has been removed and Druid now officially supports JDK 21, this PR re-enables tests on JDK 21 by default on PRs, removing the requirement to explicitly add the label.

See related slack thread - https://apachedruidworkspace.slack.com/archives/C030CMF6B70/p1762309385094569

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/processing/src/test/java/org/apache/druid/java/util/common/CompressionUtilsTest.java
+++ b/processing/src/test/java/org/apache/druid/java/util/common/CompressionUtilsTest.java
@@ -33,6 +33,7 @@ import org.apache.commons.compress.compressors.zstandard.ZstdCompressorOutputStr
 import org.apache.druid.utils.CompressionUtils;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -551,8 +552,9 @@ public class CompressionUtilsTest
   }
 
   // If this ever passes, er... fails to fail... then the bug is fixed
-  @Test(expected = AssertionError.class)
   // http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7036144
+  @Test(expected = AssertionError.class)
+  @Ignore("This test fails in JDK 21, looks like the bug was fixed in 21.0.9 at least.")
   public void testGunzipBug() throws IOException
   {
     final ByteArrayOutputStream tripleGzByteStream = new ByteArrayOutputStream(GZ_BYTES.length * 3);


.github/workflows/ci.yml
```diff
@@ -71,21 +72,6 @@ jobs:
       artifacts_to_download: "unit-test-reports-*"
       key: "coverage-jacoco"
 
-  run-unit-tests-all-jdk:
```
gianm, line 74: In the interests of making CI run faster, IMO it would be good to keep this, but flip 17 and 21. So, every PR runs java 21 tests and some run 17 tests. That way we're always doing our regular testing with the latest supported JDK. There is a tradeoff here in terms of catching stuff earlier, but I tend to believe that CI running quickly is worth the tradeoff.

abhishekrb19: The unit tests seem to run in parallel for the matrix entries - one with [both 17 and 21 ](https://github.com/apache/druid/actions/runs/19174393128) in this PR took ~1h 13m and another PR with only [17](https://github.com/apache/druid/actions/runs/19166672258) took ~1h 12m (here's another one with [only](https://github.com/apache/druid/actions/runs/19010125601) 17 that took ~1h 5m).  Unless we’re hitting a concurrency limit on GHA jobs, this shouldn’t be a problem purely for CI runtime?

gianm: Ah, got it. In that case running both seems fine.

kgyrtkirk:
the purpose of it was to run the tests once on the primary jdk; and optional only then proceed to others (possibly reducing resource use if there are failed tests; note: around when this was made testing was run on 3 jdks)

alternate jdk runs were deliberatlely skipped on PRs which were not requesting for it via the label - to reduce resource usage as in most cases these wont add further coverage to a PR.

the current approach parallelizes the 2 jdk results and merges them in the test-report - which may make it less obvious if a test fails only with 1 jdk

ideally the alternate jdk runs should run in parallel with the integration test like things...but never got to that

I would like to highlight that in case of PR testing its important to provide results for possible multiple PRs - optimizing to use up resources to service a single PR will not be the best use of resources as it could force the system to queue up processing earlier.

Also note that usually the CI runs multiple times on single PR - in most cases its practiaclly has no value to run also on alternate jdks as most changes doesn't affect it...but after the merge all other jdks checks were run on the branch commit - so it was still checked.

processing/src/test/java/org/apache/druid/java/util/common/CompressionUtilsTest.java
```diff
@@ -551,8 +552,9 @@ public void testZeroRemainingInputStream() throws IOException
   }
 
   // If this ever passes, er... fails to fail... then the bug is fixed
-  @Test(expected = AssertionError.class)
   // http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7036144
+  @Test(expected = AssertionError.class)
+  @Ignore("This test fails in JDK 21, looks like the bug was fixed in 21.0.9 at least.")
```
kgyrtkirk, line 557: couldn't you skip this testcase in a jdk version aware way?

PR #18718, opened 2025-11-05
prometheus-emitter: disable label tracking when the TTL isn't set

aho135:
A small follow up to https://github.com/apache/druid/pull/18689 to not track labels if the TTL isn't set

<hr>

 * `DimensionsAndCollector`
 * `PrometheusEmitterTest`

<hr>

This PR has:

- [X] been self-reviewed.

### DIFF
--- a/extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
+++ b/extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
@@ -72,12 +72,11 @@ public class DimensionsAndCollector
 
   /**
    * For each unique set of labelValues, keeps track of the amount of time that has elapsed since its metric
-   * value has been updated.
+   * value has been updated. Label tracking is only required if a metric TTL has been configured
    */
   public void resetLastUpdateTime(List<String> labelValues)
   {
     if (ttlSeconds == null) {
-      // Label tracking is only required if a metric TTL has been configured
       return;
     }
     labelValuesToStopwatch.compute(labelValues, (k, v) -> {


extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/DimensionsAndCollector.java
```diff
@@ -76,6 +76,10 @@ public double[] getHistogramBuckets()
    */
   public void resetLastUpdateTime(List<String> labelValues)
   {
+    if (ttlSeconds == null) {
+      // Label tracking is only required if a metric TTL has been configured
```
abhishekrb19: Could move this comment to the method's javadoc

abhishekrb19 (approved):

abhishekrb19:
The failing embedded `BasicAuthMSQTest` is unrelated to this PR

kfaraz:
@abhishekrb19 , even though the primary failure was unrelated, we should probably try re-runs in this case as there might be some ITs/docker tests that use the prometheus emitter (I don't know if there any though) and may be affected by the changes in the PR.

On that note, I think we should try adding an embedded test for the Prometheus emitter.

kfaraz:
The failure seems to have been due to a race condition in the auth embedded test.

abhishekrb19:
>@abhishekrb19 , even though the primary failure was unrelated, we should probably try re-runs in this case as there might be some ITs/docker tests that use the prometheus emitter (I don't know if there any though) and may be affected by the changes in the PR.

Ah yes, sorry, I should've clarified, the CI was retried and it failed twice in a row before merging (and I recall seeing the same failure on another PR too).

kfaraz:
> Ah yes, sorry, I should've clarified, the CI was retried and it failed twice in a row before merging (and I recall seeing the same failure on another PR too).

Thanks for the clarification! Yes, embedded tests have been acting out a little recently as we migrate more and more to them. I think they may have stumbled upon some race conditions in the Overlord code. I plan to spend some time investigating it next week.

PR #18722, opened 2025-11-06
Configurable bitmap index type for numeric fields in nested data columns

cecemei:
### Description

- Main change: users can now choose between full dictionary-based indexing and nulls-only indexing for long/double fields in nested column. This is done via a new `BitmapIndexType` abstraction with two implementations: DictionaryEncodedValueIndex (full indexing) and NullValueIndex (nulls-only indexing). It can be configured via `LongFieldBitmapIndexEncoding` and `DoubleFieldBitmapIndexEncoding` in `NestedCommonFormatColumnFormatSpec`. 
- Refactored `NestedDataScanQueryTest` to use parameterized testing for better coverage and maintainability, added a `ResourceFileSegmentBuilder` class to build segment for cleaner test code. 
- Minor update on `GlobalDictionaryEncodedFieldColumnWriter` to make  `getSerializedColumnSize` and `writeColumnTo` size match in the same class for consistency.

 * `BitmapIndexType`
 * `BitmapIndexType.DictionaryEncodedValueIndex`
 * `BitmapIndexType.NullValueIndex`
 * `NestedCommonFormatColumnFormatSpec`
 * `NestedDataScanQueryTest`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
@@ -20,7 +20,6 @@
 package org.apache.druid.testing.embedded.indexing;
 
 import org.apache.druid.indexing.common.task.TaskBuilder;
-import org.apache.druid.java.util.common.StringUtils;
 import org.apache.druid.segment.IndexSpec;
 import org.apache.druid.segment.nested.NestedCommonFormatColumnFormatSpec;
 import org.apache.druid.segment.nested.ObjectStorageEncoding;
@@ -96,12 +95,23 @@ public class NestedDataFormatsTest extends EmbeddedClusterTestBase
     cluster.callApi().runTask(indexTask.withId(taskId), overlord);
     cluster.callApi().waitForAllSegmentsToBeAvailable(dataSource, coordinator, broker);
 
-    // Test ingesting with skipping raw json smile format works, same row count, with ~20% storage saving
-    final String metadataSql = "select sum(num_rows), sum(size) from sys.segments where datasource = '%s'";
-    final String defaultFormatResult = cluster.runSql(metadataSql, datasourceWithDefaultFormat);
-    final String noneObjectStorageFormatResult = cluster.runSql(metadataSql, dataSource);
-    Assertions.assertEquals(StringUtils.format("%d,%d", 465_346, 53_000_804), defaultFormatResult);
-    Assertions.assertEquals(StringUtils.format("%d,%d", 465_346, 41_938_750), noneObjectStorageFormatResult);
+    // Test ingesting with skipping raw json smile format works, with the same row count
+    final String rowsSql = "select sum(num_rows) from sys.segments where datasource = '%s'";
+    final String defaultFormatRows = cluster.runSql(rowsSql, datasourceWithDefaultFormat);
+    final String noneObjectStorageFormatRows = cluster.runSql(rowsSql, dataSource);
+    Assertions.assertEquals(String.valueOf(465_346), defaultFormatRows);
+    Assertions.assertEquals(String.valueOf(465_346), noneObjectStorageFormatRows);
+
+    // Test ingesting with skipping raw json smile format works, with ~20% storage saving
+    final String metadataSql = "select sum(size) from sys.segments where datasource = '%s'";
+    final String defaultFormatSize = cluster.runSql(metadataSql, datasourceWithDefaultFormat);
+    final String noneObjectStorageFormatSize = cluster.runSql(metadataSql, dataSource);
+    System.out.println(noneObjectStorageFormatSize);
+    System.out.println(defaultFormatSize);
+    Assertions.assertTrue(
+        Integer.parseInt(noneObjectStorageFormatSize) <= Integer.parseInt(defaultFormatSize) * 0.8,
+        "Expected at least 20% space savings"
+    );
 
     // Test querying on a nested field works
     final String groupByQuery =


processing/src/main/java/org/apache/druid/segment/column/BitmapIndexEncodingStrategy.java
```diff
@@ -0,0 +1,165 @@
...
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.nio.channels.WritableByteChannel;
+import java.util.Objects;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonSubTypes({
+    @JsonSubTypes.Type(value = BitmapIndexEncodingStrategy.DictionaryId.class, name = "dictionaryId"),
+    @JsonSubTypes.Type(value = BitmapIndexEncodingStrategy.NullsOnly.class, name = "nullsOnly")
```
clintropolis: i think lets call these `dictionaryEncodedValuesIndex` and `nullValueIndex` to be consistent with the internal class names that use them on the other side (`DictionaryEncodedValueIndex` and `NullValueIndex`)

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnV5.java
```diff
@@ -61,7 +60,7 @@ public NestedDataColumnV5(
       Supplier<FixedIndexed<Double>> doubleDictionarySupplier,
       Supplier<FrontCodedIntArrayIndexed> arrayDictionarySupplier,
       SegmentFileMapper fileMapper,
-      BitmapSerdeFactory bitmapSerdeFactory,
+      NestedCommonFormatColumnFormatSpec bitmapSerdeFactory,
```
clintropolis: nit: variable name should be `columnFormatSpec`

processing/src/main/java/org/apache/druid/segment/nested/NestedCommonFormatColumnFormatSpec.java
```diff
@@ -172,14 +184,18 @@ public static NestedCommonFormatColumnFormatSpec getEffectiveFormatSpec(
   private final CompressionStrategy doubleColumnCompression;
   @Nullable
   private final BitmapSerdeFactory bitmapEncoding;
+  @Nullable
+  private final BitmapIndexEncodingStrategy numericFieldsBitmapIndexEncoding;
 
   @JsonCreator
   public NestedCommonFormatColumnFormatSpec(
       @JsonProperty("objectFieldsDictionaryEncoding") @Nullable StringEncodingStrategy objectFieldsDictionaryEncoding,
+      @JsonProperty("numericFieldsBitmapIndexEncoding") @Nullable BitmapIndexEncodingStrategy numericFieldsBitmapIndexEncoding,
```
clintropolis: this should be separately controllable for long and double fields, also I don't think 'encoding' is quite the correct thing to call this as the type of bitmap encoding itself is controlled by `IndexSpec`.

How about something like `longFieldIndexType` and `doubleFieldIndexType` since these are controlling the type of indexes we build for the field.

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexEncodingStrategy.java
```diff
@@ -0,0 +1,165 @@
...
+   * Assigned in {@link #close(BitmapFactory, GenericIndexedWriter)}
+   */
+  @Nullable
+  GenericIndexedWriter<ImmutableBitmap> writer;
+
+  public abstract void init(BitmapFactory bitmapFactory, int dictionarySize);
+
+  public abstract void add(int row, int sortedId, @Nullable Object o);
+
+  public void close(BitmapFactory bitmapFactory, GenericIndexedWriter<ImmutableBitmap> writer) throws IOException
```
clintropolis: nit: this method name is kind of confusing, maybe calling it `closeForWrite` or `writeToWriter` or something would be clearer, since this writing to the writer that is passed in, and `getSerializedSize` and `writeTo` will be called after this method (which feels wrong after calling close on something)

That said, I have a bunch of a changes I want to make to this abstraction, so is fine to leave as is too since I kind of want to change these interfaces a lot in a follow-up

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexType.java
```diff
@@ -0,0 +1,165 @@
...
+  public long getSerializedSize()
+  {
+    return writer.getSerializedSize();
+  }
+
+  @Override
+  public void writeTo(WritableByteChannel channel, SegmentFileBuilder fileBuilder) throws IOException
+  {
+    writer.writeTo(channel, fileBuilder);
+  }
```
clintropolis: these methods should maybe defensively check that writer has been set (the method currently called `close` has been called)

processing/src/main/java/org/apache/druid/segment/nested/GlobalDictionaryEncodedFieldColumnWriter.java
```diff
@@ -272,21 +272,21 @@ public long getSerializedSize() throws IOException
...
+               bitmapIndexEncoding.getSerializedSize() +
+               arraySize;
       }
 
       @Override
       public void writeTo(WritableByteChannel channel, SegmentFileBuilder fileBuilder) throws IOException
       {
         Channels.writeFully(channel, ByteBuffer.wrap(new byte[]{version.asByte()}));
-        channel.write(ByteBuffer.wrap(Ints.toByteArray(flags)));
+        Channels.writeFully(channel, ByteBuffer.wrap(Ints.toByteArray(flags)));
```
clintropolis, line 281: was this causing a problem?

processing/src/test/java/org/apache/druid/query/NestedDataTestUtils.java
```diff
@@ -307,76 +160,144 @@ public static List<Segment> createSegmentsWithConcatenatedJsonInput(
...
   }
 
-  public static List<Segment> createSegmentsForJsonInput(
-      TemporaryFolder tempFolder,
-      Closer closer,
-      String inputFile,
-      IndexSpec indexSpec
-  )
-      throws Exception
+  public static class SegmentBuilder
```
clintropolis: I think we should call this `ResourceFileSegmentBuilder` so it is more obvious that this is a wrapper around `IndexBuilder` to streamline using it with resource files.

processing/src/main/java/org/apache/druid/segment/nested/GlobalDictionaryEncodedFieldColumnWriter.java
```diff
@@ -187,7 +189,11 @@ long getSerializedColumnSize() throws IOException
    * Defines how to write the column, including the dictionary id column, along with any additional columns
    * such as the long or double value column as type appropriate.
    */
-  abstract void writeColumnTo(WritableByteChannel channel, SegmentFileBuilder fileBuilder) throws IOException;
+  void writeColumnTo(WritableByteChannel channel, SegmentFileBuilder fileBuilder) throws IOException
+  {
+    writeLongAndDoubleColumnLength(channel, 0, 0);
+    encodedValueSerializer.writeTo(channel, fileBuilder);
+  }
```
clintropolis, line 196: why remove the abstraction here? iirc I made it abstract on purpose to ensure that every field implements this method since it is pretty important to be correct

processing/src/main/java/org/apache/druid/segment/nested/NestedCommonFormatColumnFormatSpec.java
```diff
@@ -85,6 +90,13 @@ public static NestedCommonFormatColumnFormatSpec getEffectiveFormatSpec(
       defaultSpec = DEFAULT;
     }
 
+    if (builder.numericFieldsBitmapIndexEncoding == null) {
+      builder.setNumericFieldsBitmapIndexEncoding(GuavaUtils.firstNonNull(
+          defaultSpec.getNumericFieldsBitmapIndexEncoding(),
+          DEFAULT.getNumericFieldsBitmapIndexEncoding()
+      ));
```
clintropolis: nit: this would be more consistent with everything else in this method if it just used regular if/else... or I guess could change everything else, but i find the if/else a bit more readable/clearer for something with a lot of options resolving defaults like this..

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexEncodingStrategy.java
```diff
@@ -0,0 +1,165 @@
...
+    @Override
+    public int hashCode()
+    {
+      return Objects.hashCode(getClass());
+    }
+
+    @Override
+    public String toString()
+    {
+      return "DictionaryId";
```
clintropolis: nit: maybe use the same json type name string?

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexEncodingStrategy.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+
+  public static class NullsOnly extends BitmapIndexEncodingStrategy
```
clintropolis: its kind of weird that the null index needs a `GenericIndexedWriter` to write only a single bitmap... I think these should just write the bitmap like other implementations of stuff that gets wired up to `NullValueIndex`, which just use a `ByteBufferWriter` to write the blob (see `LongColumnSerializerV2`, `DoubleColumnSerializerV2`, `NestedDataColumnSerializer`, etc)

I see why it is like this, so you could have some shared code, but I think as we add other types of indexes there will be less and less shared code and having an abstract base type isn't really the correct abstraction.

processing/src/main/java/org/apache/druid/segment/nested/GlobalDictionaryEncodedFieldColumnWriter.java
cecemei: i was jumping through classes to compare `getSerializedColumnSize` and `writeColumnTo` and thought it'd be nice if inside the abstraction class these could all match. so like either overwrite all or overwrite none.

processing/src/main/java/org/apache/druid/segment/nested/GlobalDictionaryEncodedFieldColumnWriter.java
cecemei: no just for consistency with line above

processing/src/main/java/org/apache/druid/segment/nested/NestedCommonFormatColumnFormatSpec.java
cecemei: i was thinking more granular controls by field names, with numeric as default. kinda feel like nothing in particular that double and long might be different? adding `longFieldIndexType` and `doubleFieldIndexType` is not too much work either, can do that if you have strong opinions on this.

rn i renamed to `numericFieldsBitmapIndexType` and `BitmapIndexType`.

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexEncodingStrategy.java
cecemei: renamed to finalize

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexEncodingStrategy.java
cecemei: are you thinking that the `BitmapIndexType` should handle deserialization? i thought about that but was not sure where the `meta` for other bitmap index type would be stored in, so i didint explore further.

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexEncodingStrategy.java
cecemei: actually renamed to finalizeWriter because spotbug got confused with 
>Error:  Medium: BitmapIndexType.java:[line 73] is set to null inside finalize method in org.apache.druid.segment.column.BitmapIndexType [org.apache.druid.segment.column.BitmapIndexType] At BitmapIndexType.java:[line 73] FI_FINALIZER_NULLS_FIELDS

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexType.java
```diff
@@ -0,0 +1,171 @@
...
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonSubTypes({
+    @JsonSubTypes.Type(value = BitmapIndexType.DictionaryEncodedValueIndex.class, name = BitmapIndexType.TYPE_DICTIONARY),
+    @JsonSubTypes.Type(value = BitmapIndexType.NullValueIndex.class, name = BitmapIndexType.TYPE_NULL)
+})
+public abstract class BitmapIndexType implements Serializer
+{
+  protected static final String TYPE_DICTIONARY = "DictionaryEncodedValueIndex";
+  protected static final String TYPE_NULL = "NullValueIndex";
```
clintropolis: these should probably be `dictionaryEncodedValueIndex` and `nullValueIndex` to be consistent with other configuration type casing

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -1150,52 +1143,69 @@ private BaseColumnHolder readNestedFieldColumn(String field, int fieldIndex)
...
+        if (formatSpec.getNumericFieldsBitmapIndexType() instanceof BitmapIndexType.NullValueIndex) {
+          if (rBitmaps.size() != 1) {
+            throw DruidException.forPersona(DruidException.Persona.USER)
+                                .ofCategory(DruidException.Category.INVALID_INPUT)
+                                .build(StringUtils.format(
+                                    "expecting 1 bitmap for %s type index, got [%d]",
+                                    indexType.toString(),
+                                    rBitmaps.size()
+                                ));
+          }
```
clintropolis: this case should set a `NullValueIndexSupplier`

processing/src/test/java/org/apache/druid/guice/BuiltInTypesModuleTest.java
```diff
@@ -106,6 +107,7 @@ public void testOverride()
     props.setProperty("druid.indexing.formats.indexSpec.autoColumnFormatSpec.stringDictionaryEncoding.type", StringEncodingStrategy.FRONT_CODED);
     props.setProperty("druid.indexing.formats.indexSpec.autoColumnFormatSpec.stringDictionaryEncoding.bucketSize", "16");
     props.setProperty("druid.indexing.formats.indexSpec.autoColumnFormatSpec.stringDictionaryEncoding.formatVersion", "1");
+    props.setProperty("druid.indexing.formats.indexSpec.autoColumnFormatSpec.numericFieldsBitmapIndexEncoding.type", "nullsOnly");
```
clintropolis: oops, i think this needs updated

processing/src/main/java/org/apache/druid/segment/nested/GlobalDictionaryEncodedFieldColumnWriter.java
clintropolis: yea, fair, consistent is good, i don't feel super strongly

processing/src/main/java/org/apache/druid/segment/nested/NestedCommonFormatColumnFormatSpec.java
clintropolis: i was wanting long and double to be separate so we could support use cases where json has fields with long values that are used as dimensions but doubles as measures, before we implement per field customization. I think I was imagining per field customization would allow partial declaration, so that it could fall back to the per type default if an explicit configuration was not specified for a given field, so it would still be nice to be able to control them separately

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexEncodingStrategy.java
clintropolis: >are you thinking that the BitmapIndexType should handle deserialization?

yes, but not in this PR, i have some changes in mind to do as a follow-up to overhaul these interfaces a bit.

Since the indexes are stored at the end of the buffer, it wouldn't be that much trouble to switch to using `ByteBufferWriter` to just write the single bitmap and be consistent with other numeric columns and save some space compared to the overhead of `GenericIndexed` for fields that only have null value indexes. For now i think could just swap the index handling logic in `CompressedNestedDataComplexColumn.readNestedFieldColumn` to handle the index type at the part where we currently do `GenericIndexed.read`, and use `bitmapSerdeFactory.getObjectStrategy().fromByteBufferWithSize` if only has null index and to use the `NullValueIndexSupplier` as the index supplier.

processing/src/main/java/org/apache/druid/segment/nested/NestedCommonFormatColumnFormatSpec.java
cecemei: added the `longFieldBitmapIndex` and `doubleFieldBitmapIndex` config, looks like 134 bytes (for both config) per field increase in the overhead.

processing/src/main/java/org/apache/druid/segment/column/BitmapIndexEncodingStrategy.java
cecemei: updated to use `ByteBufferWriter` for null index, also realized that since the writer has state we cant really use `BitmapIndexType` directly so added a `getWriter` method

processing/src/test/java/org/apache/druid/guice/BuiltInTypesModuleTest.java
cecemei: my mistake, all have been updated

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
cecemei: this is done

PR #18729, opened 2025-11-08
HttpRemoteTaskRunnerV2

jtuglu1:
### Description

Still a draft.

I've seen on the giant lock in `HttpRemoteTaskRunner` cause severe performance degradation under heavy load(200-500ms per acquisition with 1000s of activeTasks can slow down the startPendingTasks loop in TaskQueue). This leads to scheduling delays, which leads to more lag, which auto-scales more tasks, ..., etc. The runner also has a few (un)documented races abundant in the code. This overhead also slows down query tasks under load (e.g. MSQE and others) which utilize the scheduler for execution.

I'm attempting a rewrite of this class to optimize for throughput and safety.

Apart from the performance improvements/bug fixes, this will also include some new features:
- Simpler code. The old task runner had old, legacy ZK references dangling around as well as a pretty complicated scheduling loop.
- Task priority-based scheduling onto workers (e.g. prioritize realtime tasks for scheduling), follows a simple priority-based fair-queueing policy. 

I would ultimately like to make this the default `HttpRemoteTaskRunner` and have it run in all tests/production clusters, etc. as I think that would help catch more bugs/issues. 

#### Performance Testing

Test results thus far have shown ~100-300ms speed up per task runner operation (`add()`, etc.). Over 1000s of tasks, this amounts to minutes of delay saved.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerV2.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerV2.java
@@ -1,0 +1,1971 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


jtuglu1:
@kfaraz any thoughts here?

jtuglu1:
@gianm @clintropolis could I get your thoughts here?

jtuglu1:
Bumping this @kfaraz 🙃.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerV2.java
```diff
@@ -0,0 +1,1971 @@
...
+
+                entry.setState(HttpRemoteTaskRunnerWorkItem.State.PENDING_WORKER_ASSIGN);
+                return entry;
+              }
+          );
+        }
+
+        final String workerHost = workerToAssign.getWorker().getHost();
+        try {
+          if (!runTaskOnWorker(taskId, workerHost)) {
```
jtuglu1, line 1222: A quick note: @kfaraz I was tempted to place this inside the above .compute() to avoid the race, but given it could be holding the lock for a while (and concurrent hashmap locks a range of keys, not necessarily just the key), I opted to unlock (similar to what the existing runner does). Additionally, if something like a removeWorker callback occurs, I'd prefer to update the state right away, rather than wait for an assignment timeout, etc.

PR #18731, opened 2025-11-10
[Groupby Query Metrics] Add merge buffer tracking

### LINKED ISSUE #17902: Add metric for merge buffers usage in bytes
### Description

We use Druid's Groupby queries heavily. These Groupby queries use the merge buffers to merge results from different historicals, real-time peons/indexers. We currently have the metric for the number of merge buffers, but we don't know whether we're optimizing the size of these merge buffers. Currently, we're setting it to 512 Mb, but we suspect that we can reduce it. We can't take such an action without having concrete metrics.
It would be nice if we have a metric indicating the bytes used for merge buffers.

### Motivation

The above description already has the motivation behind this feature.

GWphua:
Fixes #17902

Huge thanks to @gianm for the implementation tip in the issue!

#### Tracking merge buffer usage
- Usage of a direct byte buffer is done under `AbstractBufferHashGrouper` and its implementations.
1. Each direct byte buffer uses a `ByteBufferHashTable` along with an offset tracker.
2. Usage is calculated by tracking the maximum capacity of the byte buffer in `ByteBufferHashTable`, and maximum offset size calculated throughout the query's lifecycle. 

~~Incorporated a helpful suggestion by @aho135 : since the size of the hash tables are ever-changing, it makes sense to conduct calculations by taking the maximum values across queries -- so operators can have a better understanding of how the size of merge buffers can be configured.~~ 
_Edit: max metrics provided in #18934_

Here's an example of the current SUM implementations, vs the MAX implementation The latter helps to tell us that we should probably configure merge buffer sizes to 2G for this case:
image

#### Release note
`GroupByStatsMonitor` now provides metrics "mergeBuffer/bytesUsed", and **max metrics** for merge buffer acquisition time, bytes used, spilled bytes, and merge dictionary size.

<hr>

 * `GroupByStatsProvider`
 * Groupers + underlying ByteBuffer table/lists.

<hr>

This PR has:

- [x] been self-reviewed.

### Possible further enhancements
While building this PR, I have come across some problems which we can further enhance in the future: 

#### Nested Group-bys
The current metric is great, but will not report accurately for nested group-by's.  (Do correct me on this if I'm mistaken though!)

As far as I know, nested groupby limits the merge buffers usage count to 2, meaning the merge buffer will be re-used. IIUC, every ConcurrentGrouper (if concurrency is enabled) / SpillingGrouper (if concurrency disabled) is created and closed multiple times, and hence a per-query metric will likely over-report the merge buffer usage.

#### Simplify Memory Management
Right now we need to configure the following for each queryable service:
1. size of merge buffer
2. number of merge buffer
3. direct memory = (numProcessingThreads + numMergeBuffer + 1) * mergeBufferSizeBytes

It will be great if we can simplify the calculations down to simply configuring direct memory, and we can manage a memory pool instead. This allows for more flexibility (unused memory allocated for merge buffers may be used by processing threads instead).

### DIFF
--- a/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
+++ b/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
@@ -395,6 +395,10 @@ public class ByteBufferHashTable
     return growthCount;
   }
 
+  /**
+   * To maintain an accurate tracking of the maximum bytes used per query, this function is to be called immediately
+   * whenever either of {@link #size} or {@link #bucketSizeWithHash} is changed.
+   */
   protected void updateMaxMergeBufferUsedBytes()
   {
     maxMergeBufferUsedBytes = Math.max(maxMergeBufferUsedBytes, (long) size * bucketSizeWithHash);
--- a/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/LimitedBufferHashGrouper.java
+++ b/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/LimitedBufferHashGrouper.java
@@ -521,8 +521,8 @@ public class LimitedBufferHashGrouper<KeyType> extends AbstractBufferHashGrouper
     public void reset()
     {
       size = 0;
-      growthCount = 0;
       updateMaxMergeBufferUsedBytes();
+      growthCount = 0;
       // clear the used bits of the first buffer
       for (int i = 0; i < maxBuckets; i++) {
         subHashTableBuffers[0].put(i * bucketSizeWithHash, (byte) 0);


processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
```diff
@@ -120,6 +143,7 @@ public void addQueryStats(PerQueryStats perQueryStats)
       if (perQueryStats.getMergeBufferAcquisitionTimeNs() > 0) {
         mergeBufferQueries++;
         mergeBufferAcquisitionTimeNs += perQueryStats.getMergeBufferAcquisitionTimeNs();
+        mergeBufferTotalUsage += perQueryStats.getMergeBufferTotalUsage();
```
aho135: @GWphua Instead of summing here, what do you think about taking the max? Then the metric emitted would be the max merge buffer usage of a single query in that emission period. This would be a good signal for operators on whether they need to tweak the mergeBuffer size.

GWphua: I was thinking about this. 

There are other metrics where taking MAX will also make sense -- 
spilledBytes --> How much storage would be good to configure?
dictionarySize --> How large can the merge dictionary size get?

I am considering adding another metric (maxSpilledBytes, maxDictionarySize, maxSpilledBytes). What do you think?

aho135: Yeah agreed. I do think it makes sense to have it for those 3 metrics

Even for `mergeBuffer/acquisitionTimeNs` I think there's value in having the max, as it gives operators a signal on whether to increase numMergeBuffers

aho135:
Thanks for adding those max metrics @GWphua!

What do you think about adding `sqlQueryId` as a dimension only for the MAX metrics? I think this would be useful for understanding how much the query execution time was affected by the mergeBufferAcquisition. Can also do this in a follow-up PR if you think it's useful.

GWphua:
Hi @aho135 

Thanks for the review! I also find that it will be very helpful to emit metrics for each query, so we know which query will take up alot of resources. In our version of Druid, we simply appended each of the `PerQueryStat` to the `statsMap` in `QueryLifecycle#emitLogsAndMetrics`, but I feel its quite a hacky way of doing it. `sqlQueryId` as a dimension in the GroupByStatsMonitor will definitely help.

Alternatively, we can look into migrating the groupBy query metrics in `GroupByStatsMonitor` to `GroupByQueryMetrics`, which should emit metrics for each GroupBy query. In that way, this can make the `MAX` and `SUM` metrics redundant as we can now emit metrics for each query.

We can do more of this in a seperate PR.

aho135:
> Hi @aho135
> 
> Thanks for the review! I also find that it will be very helpful to emit metrics for each query, so we know which query will take up alot of resources. In our version of Druid, we simply appended each of the `PerQueryStat` to the `statsMap` in `QueryLifecycle#emitLogsAndMetrics`, but I feel its quite a hacky way of doing it. `sqlQueryId` as a dimension in the GroupByStatsMonitor will definitely help.
> 
> Alternatively, we can look into migrating the groupBy query metrics in `GroupByStatsMonitor` to `GroupByQueryMetrics`, which should emit metrics for each GroupBy query. In that way, this can make the `MAX` and `SUM` metrics redundant as we can now emit metrics for each query.
> 
> We can do more of this in a seperate PR.

Sounds good @GWphua I was thinking on very similar lines to emit these from `GroupByQueryMetrics`

I have a first draft on this: https://github.com/aho135/druid/commit/9f8209149221be0ff6714dc4faa1c10e6e92274e
Lmk if you have any thoughts on this. Thanks!

GWphua:
Hi @aho135, since the scope of adding `GroupByQueryMetrics` is out of this PR, I have created #18781 to allow us to further  discuss it there.

> I have a first draft on this: https://github.com/aho135/druid/commit/9f8209149221be0ff6714dc4faa1c10e6e92274e
> Lmk if you have any thoughts on this. Thanks!

I have a draft for `GroupByQueryMetrics` before creating this PR,  and my draft is a direct extension of your implementation shared. I think I will try and create a PR with that draft soon. I was actually hoping to get this PR merged, before sharing the draft, because that draft is done as a follow-up to this PR.

docs/operations/metrics.md
```diff
@@ -89,9 +89,14 @@ Most metric values reset each emission period, as specified in `druid.monitoring
 |`mergeBuffer/used`|Number of merge buffers used from the merge buffer pool.|This metric is only available if the `GroupByStatsMonitor` module is included.|Depends on the number of groupBy queries needing merge buffers.|
 |`mergeBuffer/queries`|Number of groupBy queries that acquired a batch of buffers from the merge buffer pool.|This metric is only available if the `GroupByStatsMonitor` module is included.|Depends on the number of groupBy queries needing merge buffers.|
 |`mergeBuffer/acquisitionTimeNs`|Total time in nanoseconds to acquire merge buffer for groupBy queries.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
+|`mergeBuffer/bytesUsed`|Number of bytes used by merge buffers to process groupBy queries.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
+|`mergeBuffer/maxAcquisitionTimeNs`|Maximum time in nanoseconds to acquire merge buffer for any single groupBy query within the emission period.|This metric is only available if the `GroupByStatsMonitor` module is included.|Varies|
```
abhishekrb19: nit: move this immediately after ``mergeBuffer/acquisitionTimeNs`

GWphua:
Hi @gianm, would appreciate it if I receive a review/feedback on this PR. Thanks!

server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
```diff
@@ -59,14 +59,19 @@ public synchronized AggregateStats getStatsSince()
...
             2L,
             200L,
+            200L,
+            300L,
             300L
         );
       }
     };
 
-    mergeBufferPool = new DefaultBlockingPool(() -> ByteBuffer.allocate(1024), 5);
```
abhishekrb19: Could you please leave out all the formatting changes here and rest of the places for simplicity?

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
```diff
@@ -110,50 +159,63 @@ public long getSpilledBytes()
...
     {
-      AggregateStats aggregateStats =
-          new AggregateStats(
-              mergeBufferQueries,
-              mergeBufferAcquisitionTimeNs,
-              spilledQueries,
-              spilledBytes,
-              mergeDictionarySize
-          );
-
```
abhishekrb19, line 176: Any reason to change this contract? I see the caller now calls `AggregateStats()` explicitly before `reset()` to save the state, I think the old approach was better.

processing/src/test/java/org/apache/druid/query/groupby/GroupByStatsProviderTest.java
```diff
@@ -63,6 +66,7 @@ public void testMetricCollection()
     aggregateStats = statsProvider.getStatsSince();
     Assert.assertEquals(2, aggregateStats.getMergeBufferQueries());
     Assert.assertEquals(1800L, aggregateStats.getMergeBufferAcquisitionTimeNs());
+    Assert.assertEquals(150L, aggregateStats.getMergeBufferTotalUsage());
```
abhishekrb19: Should we also add verify the new maxMergeBufferAcquisitionTimeNs, maxMergeBufferUsage, etc here?

abhishekrb19: Also, I think it's worth adding a test to validate these max metrics explicitly. For example, three resources r1, r2 and r3 with different spilled bytes, acquisition times, etc.

processing/src/test/java/org/apache/druid/query/groupby/GroupByStatsProviderTest.java
```diff
@@ -45,6 +46,7 @@ public void testMetricCollection()
 
     stats2.mergeBufferAcquisitionTime(500);
     stats2.mergeBufferAcquisitionTime(600);
+    stats1.mergeBufferTotalUsage(100);
```
abhishekrb19: nit: did you intend to mix stats1 operation with stats2? If not, please move this above these stats2 block

docs/operations/metrics.md
abhishekrb19, line 92: nit: move this after `mergeBuffer/acquisitionTimeNs` similar to the other corresponding max metrics

docs/operations/metrics.md
```diff
@@ -113,9 +118,14 @@ Most metric values reset each emission period, as specified in `druid.monitoring
...
```
abhishekrb19, line 121: ditto here and below

server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
```diff
@@ -128,11 +134,17 @@ public void testMonitorWithServiceDimensions()
...
     verifyMetricValue(emitter, "mergeBuffer/queries", dimFilters, 1L);
     verifyMetricValue(emitter, "mergeBuffer/acquisitionTimeNs", dimFilters, 100L);
+    verifyMetricValue(emitter, "mergeBuffer/bytesUsed", dimFilters, 200L);
+    verifyMetricValue(emitter, "mergeBuffer/maxAcquisitionTimeNs", dimFilters, 100L);
+    verifyMetricValue(emitter, "mergeBuffer/maxBytesUsed", dimFilters, 200L);
     verifyMetricValue(emitter, "groupBy/spilledQueries", dimFilters, 2L);
     verifyMetricValue(emitter, "groupBy/spilledBytes", dimFilters, 200L);
+    verifyMetricValue(emitter, "groupBy/maxSpilledBytes", dimFilters, 200L);
     verifyMetricValue(emitter, "groupBy/mergeDictionarySize", dimFilters, 300L);
+    verifyMetricValue(emitter, "groupBy/maxMergeDictionarySize", dimFilters, 300L);
```
abhishekrb19, line 144: I think it will be nice to have a separate test that mimics multiple resource IDs - something that doesn't use the static `AggregateStats getStatsSince()`  defined in the test setup.  The existing tests can continue to the setup for convenience.

A new test could leverage the provider's methods on different resource IDs, similar to what the `GroupByMergingQueryRunner` would actually do:
- `groupByStatsProvider.getPerQueryStatsContainer()` 
-  `groupByStatsProvider.closeQuery()`


Please let me know what you think.

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
abhishekrb19: Actually on second thought, I think the new contract `void reset()` is better. It aligns with the `reset()` signature defined in the grouper and generally other places too

processing/src/test/java/org/apache/druid/query/groupby/GroupByStatsProviderTest.java
GWphua: Thanks for the catch, I meant to add it in stats2. Fixed!

processing/src/test/java/org/apache/druid/query/groupby/GroupByStatsProviderTest.java
GWphua: Added

server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
GWphua: I do think that it's better to test the `GroupByStatsProvider` methods in the  `GroupByStatsProviderTest`, but I do not have a strong preference for this.

Regardless, I have added a testcase under the monitor test.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ConcurrentGrouper.java
```diff
@@ -332,7 +332,7 @@ public void reset()
       throw new ISE("Grouper is closed");
     }
 
-    groupers.forEach(Grouper::reset);
```
abhishekrb19: Is this change required? Given that this operates on `SpillingGrouper`, I suppose this is technically correct. But calling `Grouper::reset` should already ensure that the specific reset/close methods are called?

server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
```diff
@@ -185,6 +194,55 @@ public void testMonitoringMergeBuffer_pendingRequests()
     }
   }
 
+  @Test
+  public void testMonitoringWithoutMockingGroupByStatsProvider()
```
abhishekrb19: Perhaps something like this (the existing `groupByStatsProvider` is technically still a concrete type):
```suggestion
  public void testMonitoringWithMultipleResources()
```

GWphua:
@abhishekrb19, thanks for the review! 

I have made changes according to your suggestions.

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
GWphua: My reason for the change is more from a new reader's perspective.

To understand `getStatsSince()`, I had to understand that `reset()` not only resets the stats monitored, it also returns a copy of all stats being monitored so far.

Hopefully, the current implementation will make it easier to understand the intent of `getStatsSince` without jumping to `reset()`.

server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
abhishekrb19: Perhaps something like this to avoid confusion - the existing `groupByStatsProvider` in the test setup is still a concrete type (not a mock), even though it overrides a method

```suggestion
  public void testMonitoringWithMultipleResources()
```

server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
abhishekrb19: It looks much cleaner, thanks for adding it!

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ConcurrentGrouper.java
abhishekrb19, line 335: Is this change required? Given that the `ConcurrentGrouper` operates on `SpillingGrouper` instance, I suppose this is technically correct. But calling `Grouper::reset` as it was earlier should already ensure that the specific reset/close methods from `SpillingGrouper` are invoked? If so, could we revert this change here and below?

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/BufferHashGrouper.java
```diff
@@ -199,18 +209,15 @@ public int size()
...
-          (lhs, rhs) -> {
-            final ByteBuffer tableBuffer = hashTable.getTableBuffer();
-            return comparator.compare(
-                tableBuffer,
-                tableBuffer,
-                lhs + HASH_SIZE,
-                rhs + HASH_SIZE
-            );
-          }
-      );
```
abhishekrb19, line 213: Functionally there's no difference between the two sorting methods, right? If this is a stylistic preference, please revert this from the PR and we can do these changes separately if desired.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/SpillingGrouper.java
```diff
@@ -68,7 +68,7 @@ public class SpillingGrouper<KeyType> implements Grouper<KeyType>
       "Not enough disk space to execute this query. Try raising druid.query.groupBy.maxOnDiskStorage."
   );
 
-  private final Grouper<KeyType> grouper;
+  private final AbstractBufferHashGrouper<KeyType> grouper;
```
abhishekrb19, line 71: Same comment as above. I think it would be better to define these methods in the `Grouper` interface and leave this change as it was earlier. Mixing `Grouper` with impls otherwise seems confusing

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/AbstractBufferHashGrouper.java
```diff
@@ -173,6 +173,15 @@ public void close()
...
   }
 
+  /**
+   * This method is implemented to return the highest memory value claimed by the Grouper. This is only
+   * used for monitoring the size of the merge buffers used.
+   */
+  public long getMergeBufferUsage()
+  {
+    return hashTable.getMaxTableBufferUsage();
+  }
```
abhishekrb19: please move this to the `Grouper` interface and have a default implementation for it. Override it in all the concrete classes as needed:
```
  default long getMergeBufferUsage()
  {
    return 0l;
  }
```

abhishekrb19: Also, `getMergeBufferUsage()` seems a bit vague. What do you think about `getMaxMergeBufferUsedBytes()`? Once decided on a name, please use the name across the board for consistency

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferIntList.java
```diff
@@ -71,4 +72,9 @@ public void reset()
   {
     numElements = 0;
```
abhishekrb19, line 76: Shouldn't `reset()` also set `maxMergeBufferUsageBytes` to 0?

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
```diff
@@ -100,6 +134,21 @@ public long getMergeBufferAcquisitionTimeNs()
       return mergeBufferAcquisitionTimeNs;
     }
 
+    public long getMergeBufferTotalUsage()
+    {
+      return mergeBufferTotalUsage;
+    }
```
abhishekrb19: On the naming, consider:
- `getMergeBufferTotalUsage() -> `getMergeBufferTotalUsedBytes()`
- `getMaxMergeBufferUsage()` -> `getMaxMergeBufferUsedBytes()

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
```diff
@@ -100,6 +134,21 @@ public long getMergeBufferAcquisitionTimeNs()
...
+
+    public long getMaxMergeBufferAcquisitionTimeNs()
+    {
+      return maxMergeBufferAcquisitionTimeNs;
+    }
```
abhishekrb19: nit: consider moving this after `getMergeBufferAcquisitionTimeNs()` so the maxes are grouped together with their total counterparts

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferIntList.java
```diff
@@ -55,6 +55,7 @@ public void add(int val)
     }
     buffer.putInt(numElements * Integer.BYTES, val);
     numElements++;
+    maxMergeBufferUsageBytes = Math.max(maxMergeBufferUsageBytes, numElements * Integer.BYTES);
```
abhishekrb19: Actually I think this variable and state tracking isn't needed in `add()` since we're tracking `numElements` already. We can just do it inline `numElements * Integer.BYTES` from `getMaxMergeBufferUsageBytes()`

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
```diff
@@ -381,6 +387,16 @@ public int getGrowthCount()
     return growthCount;
   }
 
+  protected void updateMaxTableBufferUsage()
+  {
+    maxTableBufferUsage = Math.max(maxTableBufferUsage, tableBuffer.capacity());
```
abhishekrb19: Unless I'm missing something, the issue https://github.com/apache/druid/issues/17902 was created to actually get some visibility into actual merge buffer usage, but this metric would just tell us how much was actually configured instead? 

`tableBuffer.capacity()` would just indicate the capacity of the buffers, so more or less what was configured via `druid.processing.buffer.sizeBytes`.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/BufferHashGrouper.java
```diff
@@ -154,6 +152,18 @@ public void reset()
...
+  @Override
+  public long getMergeBufferUsage()
+  {
+    if (!initialized) {
+      return 0L;
+    }
+
+    long hashTableUsage = hashTable.getMaxTableBufferUsage();
+    long offSetListUsage = offsetList.getMaxMergeBufferUsageBytes();
+    return hashTableUsage + offSetListUsage;
```
abhishekrb19: Same comment, I think this would just more or less tell us configured size rather than actual buffer usage. (more or less because of offset list tracking)

abhishekrb19 (commented):
Thanks @GWphua! I tried to take a closer look at the grouper changes - please see my latest comments. To track actual merge buffer usage the way https://github.com/apache/druid/issues/17902
 proposes, we may need some additional thought: using `buffer.capacity()` in the various places is going to reflect the configured buffer size, not the actual bytes used.


My suggestion would be to split all the new max metrics into a separate PR: `mergeBuffer/maxAcquisitionTimeNs`, `groupBy/maxSpilledBytes`, `groupBy/maxMergeDictionarySize`. Those look more straightforward and would still be useful for operators to track.

Then we can keep this PR and the linked issue open to track `mergeBuffer/bytesUsed` and `mergeBuffer/maxBytesUsed`. We'll also want to think about and add some test coverage for the grouper changes once the approach is finalized.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/AbstractBufferHashGrouper.java
abhishekrb19: Some of these comments related to merge buffer usage may be stale. Please see https://github.com/apache/druid/pull/18731#discussion_r2687839923

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ConcurrentGrouper.java
GWphua: You are correct, only SpillingGrouper's method is called here. 

I changed this to make my life easier during development:
1. IntelliJ can jump to the SpillingGrouper method, instead of going to the Grouper interface.
2. Future readers will be able to tell that the groupers object held by `ConcurrentGrouper` will be `SpillingGrouper`s.

If the purpose is to keep the changes limited to this PR only connected to the Groupby metrics, I will be opening a new PR. Let me know if you would prefer that.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/BufferHashGrouper.java
GWphua: Noted on this, changed!

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
GWphua: Done

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/SpillingGrouper.java
GWphua: For this, I will need to maintain the AbstractBufferHashGrouper. The reason will also address why `getMergeBufferUsedBytes` is not placed in the `Groupers` interface.

This is because the groupby metrics are only collated when `Groupers#close` is called. This means that we will not be interfacing with `Groupers#getMergeBufferUsedBytes` at all, but letting `Groupers#close` retrieve `getMergeBufferUsedBytes`. This is why `SpillingGrouper#getMergeBufferUsedBytes` is private, following the example set by `SpillingGrouper#getDictionarySizeEstimate`. 

The only time we need to interface with `getMergeBufferUsedBytes` is in `AbstractBufferHashGrouper`, which are the underlying groupers used by the `SpillingGrouper`.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/AbstractBufferHashGrouper.java
GWphua: the new name `getMaxMergeBufferUsedBytes()` sounds great, used it.

For the Groupers interface, please see my above comments about `AbstractBufferHashGrouper` in `SpillingGrouper`

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferIntList.java
GWphua: I was thinking to get the maximum usage, because we request a `getMergeBufferUsage` when `SpillingGrouper#close` is called. I do not really know whether we are guaranteed to see the maximum usage when the grouper is closing this class...

I am worried about the case where maybe we configure 1GiB to the merge buffer, and the usage in the middle goes to, say 900MiB, but when the Grouper is closed, the usage shows ~300MiB. The user is then encouraged to lower the merge buffer allocation to 500MiB, which will be problematic.

This comment would also hopefully address your query about why `reset` does not change `maxMergeBufferUsageBytes`.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferIntList.java
GWphua: The `ByteBufferIntList` in BufferHashGrouper should by all means be `final`, but we cannot make it so because it is only initialized when `BufferHashGrouper#init` is called. 

Here I am intending to record the maximum usage for the entire `BufferHashGrouper`'s lifecycle. So, reset will not set the max usage to 0.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
GWphua: You are right about the current metric reporting the allocation. 

I relooked the implementation, and found that maybe we can better estimate the ByteBufferHashTable. Since the ByteBufferHashTable is an open-addressing hash table, we can use the number of elements in table (size) * space taken up per bucket (bucketSizeWithHash) to estimate the usage in bytes.

GWphua:
Hi @abhishekrb19, I have revisited your comments, and have made the relevant changes/replies. Please take a look at the new approach to calculating the usage. Thanks!

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/LimitedBufferHashGrouper.java
```diff
@@ -571,7 +585,19 @@ public void adjustTableWhenFull()
...
       growthCount++;
     }
+
+    @Override
+    protected void updateMaxTableBufferUsedBytes()
+    {
+      long currentBufferUsedBytes = 0;
+      for (ByteBuffer buffer : subHashTableBuffers) {
+        currentBufferUsedBytes += buffer.capacity();
+      }
```
abhishekrb19: I think this effectively would just be `tableArenaSize` which would reflect the allocated configured size rather than actual used size?

abhishekrb19: Please update `LimitedBufferHashGrouperTest`, `BufferHashGrouperTest`  and related grouper tests to validate the correctness of these implementations.

I just pulled in the latest patch locally and ran some group by queries and noticed that the `bytesUsed` and `maxBytesUsed` were more or less what was configured `druid.processing.buffer.sizeBytes` 🤔 


I’ll try to dig into this more, but in the meantime, I’d still recommend splitting the PR into two parts: 1. max metrics 2. the bytesUsed and maxBytesUsed metrics
2 seems a bit more involved.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
abhishekrb19: Thanks for the update. Please see https://github.com/apache/druid/pull/18731/changes#r2706872449

GWphua:
> My suggestion would be to split all the new max metrics into a separate PR: `mergeBuffer/maxAcquisitionTimeNs`, `groupBy/maxSpilledBytes`, `groupBy/maxMergeDictionarySize`. Those look more straightforward and would still be useful for operators to track.
> 
> Then we can keep this PR and the linked issue open to track `mergeBuffer/bytesUsed` and `mergeBuffer/maxBytesUsed`.

As suggested, moved 3 max metrics to #18934, leaving `maxBytesUsed` as the only max metric introduced in this PR.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/LimitedBufferHashGrouper.java
GWphua: Hello, I have added the tests for the groupers.

I did not get the same results as you, maybe because I used queries for a smaller dataset. 

How I did in my tests is to query with spill to disk enabled:
1. Set druid.processing.buffer.sizeBytes = 1GB
2. Query on a dataset. (Let's say the results for this is 100MB)
3. Set druid.processing.buffer.sizeBytes to a much smaller value ~5MB
4. Query on the same dataset, and watch the usage metrics cap at 5MB, with spillage to disk ~95MB.

Here's an example of how my max metrics look like:
image

I do have to admit, some of the values are kinda "blocky", like it will report ~28MB repeatedly for, say 3 consecutive metrics, then report some other value. Maybe this is because similar queries are being sent during a short period of time, and perhaps the allocated space is the same for these similar queries. Hopefully, this will be fixed by your catch -- reporting the usage instead of the capacity. 😄

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
GWphua: Made the AlternatingByteBufferHashTable inherit the max metrics reporting from the superclass. Should now accurately report the usage :)

GWphua:
@abhishekrb19 

The conflicts have been resolved. You can look again when you're free. Thanks!

abhishekrb19:
> @abhishekrb19
> 
> The conflicts have been resolved. You can look again when you're free. Thanks!

Ack, thanks @GWphua! I will take a look at the latest changes later this week/early next week

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/AbstractBufferHashGrouper.java
```diff
@@ -173,6 +173,18 @@ public void close()
...
   }
 
+  /**
+   * Retrieves the size of the merge buffers used for this groupby query. This value is retrieved when
+   * {@link SpillingGrouper#close()} is called.
+   * <p></p>
+   * This method is implemented to return the highest memory value used, this is helpful especially in
+   * reporting the highest number of bytes used throughout the entire query lifecycle.
+   */
+  public long getMergeBufferUsedBytes()
```
abhishekrb19: Should this method and all of its implementation overrides be replaced by `getMaxMergeBufferUsedBytes()` ? To complement `updateMaxMergeBufferUsedBytes()` and since the tracking is essentially max.

processing/src/test/java/org/apache/druid/query/groupby/epinephelinae/BufferHashGrouperTest.java
```diff
@@ -156,6 +156,61 @@ public void testNoGrowing()
     }
   }
 
+  @Test
+  public void testMaxMergeBufferUsedBytesWorksNormally()
```
abhishekrb19:
```suggestion
  public void testMaxMergeBufferUsedBytes()
```

processing/src/test/java/org/apache/druid/query/groupby/epinephelinae/BufferHashGrouperTest.java
```diff
@@ -156,6 +156,61 @@ public void testNoGrowing()
...
+
+    grouper.aggregate(new IntKey(2));
+    grouper.aggregate(new IntKey(3));
+
+    Assert.assertEquals(3L * expectedBucketSize, grouper.getMergeBufferUsedBytes());
+
+    grouper.aggregate(new IntKey(4));
+    grouper.aggregate(new IntKey(5));
+
+    Assert.assertEquals(5L * expectedBucketSize, grouper.getMergeBufferUsedBytes());
```
abhishekrb19: It'll be good to include and verify duplicate keys in `aggregate()` doesn't cause the buffer to grow in size:
```
    grouper.aggregate(new IntKey(5));
    grouper.aggregate(new IntKey(1));
    Assert.assertEquals(5L * expectedBucketSize, grouper.getMergeBufferUsedBytes());
```

processing/src/test/java/org/apache/druid/query/groupby/epinephelinae/BufferHashGrouperTest.java
```diff
@@ -156,6 +156,61 @@ public void testNoGrowing()
...
+
+    grouper.reset();
+    Assert.assertEquals(0, grouper.getSize());
+    Assert.assertEquals(5L * expectedBucketSize, grouper.getMergeBufferUsedBytes());
+
```
abhishekrb19: Was curious about the behavior here. Any idea why after a `reset()`, just adding these isn't working as expected? The 6 entries below do work (just anything under 5 aggregates() fail):
```
grouper.aggregate(new IntKey(1));
grouper.aggregate(new IntKey(2));
Assert.assertEquals(2L * expectedBucketSize, grouper.getMergeBufferUsedBytes());
```

For anything less than what was aggregate() from before reset(), it seems to keep track of the max bytes from before, so it's essentially overreporting.

```
java.lang.AssertionError: 
Expected :58
Actual   :145
```
(145 = 5 * expectedBucketSize, it keeps memory of stuff from before `reset()` so it's saving the value from before)

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/LimitedBufferHashGrouper.java
```diff
@@ -458,6 +458,18 @@ public boolean validateBufferCapacity(int bufferCapacity)
     }
   }
 
+  @Override
+  public long getMergeBufferUsedBytes()
```
abhishekrb19: `getMaxMergeBufferUsedBytes()`?

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
```diff
@@ -245,6 +258,7 @@ protected void initializeNewBucketKey(
     tableBuffer.putInt(Groupers.getUsedFlag(keyHash));
     tableBuffer.put(keyBuffer);
     size++;
+    updateMaxMergeBufferUsedBytes();
```
abhishekrb19, line 261: Should there be a similar call to `updateMaxMergeBufferUsedBytes()` from `adjustTableWhenFull()` as well? 

The `LimitedBufferHashGrouper` does that too

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ConcurrentGrouper.java
abhishekrb19: Got it. It’s generally preferable not to mix refactoring and functional changes together, but this one is relatively straightforward, so it’s okay with me.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/LimitedBufferHashGrouper.java
abhishekrb19: Yeah, I was playing around with it and I noticed more accurate reporting with the latest change than with the previous iterations. 

>I do have to admit, some of the values are kinda "blocky", like it will report ~28MB repeatedly for, say 3 consecutive metrics, then report some other value.

I wonder if the "stickiness" in the reporting is coming from `reset()` that was observed in the test: https://github.com/apache/druid/pull/18731#discussion_r2857228278

processing/src/test/java/org/apache/druid/query/groupby/epinephelinae/BufferHashGrouperTest.java
GWphua: This is because `grouper.getMergeBufferUsedBytes` tracks the max bytes used for a single query.

During a `reset()`, the history of the max bytes used is not reset. So after `reset()`, the value of `grouper.getMergeBufferUsedBytes` is unchanged (stays at 5). The test below it shows how much bytes are used by the grouper if we try to add a bunch of new items. 

The new value of `grouper.getMergeBufferUsedBytes` will be `Math.max(used bytes before reset, used bytes now)`

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
GWphua: Since `updateMaxMergeBufferUsedBytes` is calculated by `size * bucketSizeWithHash`, I added this function wherever `size` or `bucketSizeWithHash` is changed. 

ByteBufferHashTable does not change both values in `adjustTableWhenFull`, so I did not add that. Granted, from a code reader's perspective, the places where the max merge buffer usage is updated seems quite random...

I have added a Javadoc to the related function.

processing/src/test/java/org/apache/druid/query/groupby/epinephelinae/BufferHashGrouperTest.java
GWphua: Added test and it works. :)

PR #16140, opened 2024-03-15
MetricsModule: inject DataSourceTaskIdHolder early

arunramani:
### Description
This fixes an issue where MSQ ingestion failed if the first monitor in the list is `ServiceStatusMonitor`. Before this change, setting the `ServiceStatusMonitor` as the first monitor will fail with errors in `CliPeon` like below

```
1) Error in custom provider, java.lang.RuntimeException: com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `org.apache.druid.segment.virtual.ExpressionVirtualColumn`, problem: Unable to provision, see the following errors:

1) Problem parsing object at prefix[druid.lookup]: Cannot construct instance of `org.apache.druid.query.lookup.LookupListeningAnnouncerConfig`, problem: Unable to provision, see the following errors:

1) Error in custom provider, java.lang.IllegalStateException: This is a proxy used to support circular references. The object we're proxying is not constructed yet. Please wait until after injection has completed to use this object.
  at org.apache.druid.cli.CliPeon$1.getDataSourceFromTask(CliPeon.java:344) (via modules: com.google.inject.util.Modules$OverrideModule -> com.google.inject.util.Modules$OverrideModule -> org.apache.druid.cli.CliPeon$1)
  at org.apache.druid.cli.CliPeon$1.getDataSourceFromTask(CliPeon.java:344) (via modules: com.google.inject.util.Modules$OverrideModule -> com.google.inject.util.Modules$OverrideModule -> org.apache.druid.cli.CliPeon$1)
  while locating java.lang.String annotated with @com.google.inject.name.Named(value="druidDataSource")
    for field at org.apache.druid.server.metrics.DataSourceTaskIdHolder.dataSource(DataSourceTaskIdHolder.java:29)
  at org.apache.druid.server.metrics.MetricsModule.configure(MetricsModule.java:92) (via modules: com.google.inject.util.Modules$OverrideModule -> com.google.inject.util.Modules$OverrideModule -> org.apache.druid.server.metrics.MetricsModule)
  while locating org.apache.druid.server.metrics.DataSourceTaskIdHolder
Caused by: java.lang.IllegalStateException: This is a proxy used to support circular references. The object we're proxying is not constructed yet. Please wait until after injection has completed to use this object.
	at com.google.common.base.Preconditions.checkState(Preconditions.java:512)
	at com.google.inject.internal.DelegatingInvocationHandler.invoke(DelegatingInvocationHandler.java:36)
[...]

PR #18732, opened 2025-11-10
Refactor: Rename DataSourceTaskIdHolder and extract load spec holder to its own class

abhishekrb19:
Refactor:
- Remove named injections of `DataSourceTaskIdHolder` in favor of:
   - `TaskHolder` interface: implemented by `PeonTaskHolder` for peon processes and by `NoopTaskHolder` for all other servers, where the methods return null. The alternative was to throw `UnsupportedException`s in `NoopTaskHolder` and have callers verify type, but that made the usages a bit [difficult](https://github.com/apache/druid/pull/18732/commits/91acfd34a32d2c3e26fcc09d8d37fafaf0a3802d) and the callers would still have to effectively assign null anyway. 
   - `LoadSpecHolder` interface: implemented by `PeonLoadSpecHolder` for peon processes and `DefaultLoadSpecHolder` for all other servers. 
- A module `DefaultServerHolderModule` that binds the above modules as part of the core injector module for all servers except CliPeon. For CliPeon, the binding to `PeonTaskHolder` and `PeonLoadSpecHolder` is explicitly done in CliPeon itself.

Some implementation specific choices have been documented as javadocs in the code.  In the future, we can extend `TaskHolder` to also include `groupId`, `taskType`, etc., from the Task and add those dimensions to all the metric modules as applicable.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/collections/DummyBlockingPool.java
+++ b/processing/src/main/java/org/apache/druid/collections/DummyBlockingPool.java
@@ -22,16 +22,16 @@ package org.apache.druid.collections;
 import java.util.List;
 
 /**
- * BlockingPool with 0 maxSize, all take*() methods immediately throw {@link UnsupportedOperationException}.
+ * BlockingPool with 0 maxSize, all {@code takeBatch} methods immediately throw {@link UnsupportedOperationException}.
  */
-public final class DummyBlockingPool<T> implements BlockingPool<T>
+public final class DummyBlockingPool implements BlockingPool<Object>
 {
   private static final DummyBlockingPool INSTANCE = new DummyBlockingPool();
 
   @SuppressWarnings("unchecked")
   public static <T> BlockingPool<T> instance()
   {
-    return INSTANCE;
+    return (BlockingPool<T>) INSTANCE;
   }
 
   private DummyBlockingPool()
@@ -45,13 +45,13 @@ public final class DummyBlockingPool<T> implements BlockingPool<T>
   }
 
   @Override
-  public List<ReferenceCountingResourceHolder<T>> takeBatch(int elementNum, long timeoutMs)
+  public List<ReferenceCountingResourceHolder<Object>> takeBatch(int elementNum, long timeoutMs)
   {
     throw new UnsupportedOperationException();
   }
 
   @Override
-  public List<ReferenceCountingResourceHolder<T>> takeBatch(int elementNum)
+  public List<ReferenceCountingResourceHolder<Object>> takeBatch(int elementNum)
   {
     throw new UnsupportedOperationException();
   }
--- a/processing/src/test/java/org/apache/druid/collections/DummyBlockingPoolTest.java
+++ b/processing/src/test/java/org/apache/druid/collections/DummyBlockingPoolTest.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.collections;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class DummyBlockingPoolTest
+{
+  @Test
+  public void testSingletonInstance()
+  {
+    final BlockingPool<Object> p1 = DummyBlockingPool.instance();
+    final BlockingPool<Integer> p2 = DummyBlockingPool.instance();
+    Assertions.assertSame(p1, p2);
+  }
+
+  @Test
+  public void testMaxSizeAndMetricsAreZero()
+  {
+    final BlockingPool<Integer> pool = DummyBlockingPool.instance();
+    Assertions.assertEquals(0, pool.maxSize());
+    Assertions.assertEquals(0, pool.getPendingRequests());
+    Assertions.assertEquals(0, pool.getUsedResourcesCount());
+  }
+
+  @Test
+  public void testTakeBatchThrows()
+  {
+    final BlockingPool<Integer> pool = DummyBlockingPool.instance();
+    Assertions.assertThrows(UnsupportedOperationException.class, () -> pool.takeBatch(1));
+    Assertions.assertThrows(UnsupportedOperationException.class, () -> pool.takeBatch(1, 10L));
+  }
+}
--- a/processing/src/test/java/org/apache/druid/collections/DummyNonBlockingPoolTest.java
+++ b/processing/src/test/java/org/apache/druid/collections/DummyNonBlockingPoolTest.java
@@ -19,26 +19,25 @@
 
 package org.apache.druid.collections;
 
-/**
- * NonBlockingPool that cannot allocate objects, {@link #take()} throws {@link UnsupportedOperationException}.
- */
-public final class DummyNonBlockingPool implements NonBlockingPool<Object>
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class DummyNonBlockingPoolTest
 {
-  private static final DummyNonBlockingPool INSTANCE = new DummyNonBlockingPool();
-
-  @SuppressWarnings("unchecked")
-  public static <T> NonBlockingPool<T> instance()
+  @Test
+  public void testSingletonInstance()
   {
-    return (NonBlockingPool<T>) INSTANCE;
+    final NonBlockingPool<Object> p1 = DummyNonBlockingPool.instance();
+    final NonBlockingPool<Integer> p2 = DummyNonBlockingPool.instance();
+    Assertions.assertSame(p1, p2);
   }
 
-  private DummyNonBlockingPool()
+  @Test
+  public void testTakeThrows()
   {
-  }
-
-  @Override
-  public ResourceHolder<Object> take()
-  {
-    throw new UnsupportedOperationException();
+    final NonBlockingPool<Integer> pool = DummyNonBlockingPool.instance();
+    Assertions.assertThrows(UnsupportedOperationException.class, pool::take);
   }
 }
+
+
--- a/processing/src/test/java/org/apache/druid/segment/TestSegmentUtils.java
+++ b/processing/src/test/java/org/apache/druid/segment/TestSegmentUtils.java
@@ -126,6 +126,12 @@ public class TestSegmentUtils
     {
       return Cursors.ascendingTimeOrder();
     }
+
+    @Override
+    public int getNumRows()
+    {
+      return 1234;
+    }
   };
 
   public static class SegmentForTesting extends QueryableIndexSegment
@@ -178,6 +184,8 @@ public class TestSegmentUtils
         return (T) INDEX;
       } else if (clazz.equals(CursorFactory.class)) {
         return (T) new QueryableIndexCursorFactory(INDEX);
+      } else if (clazz.equals(PhysicalSegmentInspector.class)) {
+        return (T) new QueryableIndexPhysicalSegmentInspector(INDEX);
       }
       return null;
     }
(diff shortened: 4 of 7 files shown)


kfaraz (commented):
Thanks for putting this together, @abhishekrb19 !

I had a slightly different approach in mind.

In the original class, `DatasourceTaskIdHolder` all the fields are derived from `Task` itself, so we should probably just:
- Rename it to `PeonTaskHolder` to denote that this class is meaningful only for a peon and that it contains the task that the peon is running
- Just keep one field inside it `Task task` and add method `getTask()`
- Retain the original utility methods `getDatasource()`, `getTaskId()`, `getLookupLoadinSpec()` etc.
- Create the `PeonTaskHolder` in a `@Provides` method in `CliPeon`.
- In non peon servers, bind the `PeonTaskHolder` to null.

Let me know if this would make sense.

abhishekrb19:
Thanks for the review, @kfaraz!

I think the approach of keeping one renamed class would work well and makes sense. Just one caveat: the `Task` class is in the `indexing-service` module, so we can’t use it directly in the `server` module. However, we could create a constructor that accepts all the task properties of interest and have it `@Provides` from `CliPeon` like you suggested, along with a default constructor for all other servers. This way, we can remove all the individual `@Named` annotation bindings from `CliPeon` to `DatasourceTaskIdHolder`. Does that work?

kfaraz:
Ah, I see, thanks for clarifying that, @abhishekrb19 ! All the current usages of `DatasourceTaskIdHolder` are indeed in the `druid-server` module.

In that case, I think we should add an interface `TaskHolder` to `druid-server` and use it wherever needed.
Bind the `TaskHolder` to a `PeonTaskHolder` in `CliPeon` and bind it elsewhere to a `NoopTaskHolder` which throws `UnsupportedOperationException` upon calling any method. I feel this clarifies the intent better and makes it more extensible for the future.

Sound good?

abhishekrb19:
@kfaraz sorry I ran into some hiccups with the bindings last week and didn't get a chance to take a closer look, but I'm hoping to revive this change again soon.

Btw I noticed this deprecated comment on [EmbeddedMiddleManager](https://github.com/apache/druid/blob/master/services/src/test/java/org/apache/druid/testing/embedded/EmbeddedMiddleManager.java#L36), so most (or all?) the tests seem to use the indexer for good reasons. Given that there are some subtle differences between indexers and MMs, do you have any suggestions for embedded tests that use MMs? 

At the very least, I was thinking of updating some of the embedded tests to use `EmbeddedMiddleManager` instead of `EmbeddedIndexer` locally. I’m also planning to add some unit tests.

kfaraz:
> @kfaraz sorry I ran into some hiccups with the bindings last week and didn't get a chance to take a closer look, but I'm hoping to revive this change again soon.

No worries, @abhishekrb19 ! 🙂 

> Given that there are some subtle differences between indexers and MMs, do you have any suggestions for embedded tests that use MMs?

If the purpose is to test out peons in an embedded test setup, you could do a couple of things:
- Use `EmbeddedMiddleManager` and annotate it `@SuppressForbidden(reason = "EmbeddedMiddleManager#init")`
- Use the k3s environment with the help of `K3sClusterResource`, e.g. `KubernetesClusterDockerTest` (These tests would run in the `docker-tests` GHA workflow. It is also possible to run them locally after setting the `druid.testing.docker.image` property.)

For local testing, please feel free to update any test to use an MM instead of Indexer. Most of them should work fine for both Indexer and MM.
But while committing, I would advise not to change any of the existing tests to use MMs (since Indexers are faster and easier to debug) and add new MM-flavored variants of any test where necessary.

Please let me know if this meets yours needs.

abhishekrb19:
Sounds good, thanks for the pointer @kfaraz! I will give that a try

server/src/main/java/org/apache/druid/query/lookup/LookupReferencesManager.java
```diff
@@ -380,7 +379,7 @@ private void takeSnapshot(Map<String, LookupExtractorFactoryContainer> lookupMap
   }
 
   /**
-   * Load a set of lookups based on the injected value in {@link DataSourceTaskIdHolder#getLookupLoadingSpec()}.
+   * Load a set of lookups based on the injected value in {@link TaskPropertiesHolder#getLookupLoadingSpec()}.
```
kfaraz:
```suggestion
   * Load a set of lookups based on the injected value in {@link LoadSpecHolder#getLookupLoadingSpec()}.
```

server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java
```diff
@@ -110,9 +108,10 @@ public MonitorScheduler getMonitorScheduler(
   {
     List<Monitor> monitors = new ArrayList<>();
     // HACK: when ServiceStatusMonitor is the first to be loaded, it introduces a circular dependency between
-    // CliPeon.runTask and CliPeon.getDataSourceFromTask/CliPeon.getTaskIDFromTask. The reason for this is unclear
-    // but by injecting DataSourceTaskIdHolder early this cycle is avoided.
-    injector.getInstance(DataSourceTaskIdHolder.class);
+    // CliPeon.runTask and TaskHolder.getDataSource()/TaskHolder.getTaskId(). The reason for this is unclear
+    // but by injecting TaskPropertiesHolder early this cycle is avoided.
```
kfaraz: Is this true even after the changes in this PR?

server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java
```diff
@@ -150,50 +149,46 @@ public MonitorScheduler getMonitorScheduler(
...
   @Provides
   @ManageLifecycle
-  public JvmThreadsMonitor getJvmThreadsMonitor(DataSourceTaskIdHolder dataSourceTaskIdHolder)
+  public JvmThreadsMonitor getJvmThreadsMonitor(TaskHolder taskHolder)
   {
-    Map<String, String[]> dimensions = MonitorsConfig.mapOfDatasourceAndTaskID(
-        dataSourceTaskIdHolder.getDataSource(),
-        dataSourceTaskIdHolder.getTaskId()
+    Map<String, String[]> dimensions = MonitorsConfig.mapOfTaskHolderDimensions(
+        taskHolder
```
kfaraz: Nit: Can we put this (and other similar invocations of this new method) in a single line?

server/src/main/java/org/apache/druid/server/metrics/MonitorsConfig.java
```diff
@@ -81,19 +78,19 @@ public String toString()
...
+   * and {@link TaskHolder#getTaskId()} are non-null.
+   * <p>The task ID ({@link TaskHolder#getTaskId()}) is added to both {@link DruidMetrics#TASK_ID}
+   * and {@link DruidMetrics#ID} dimensions to the map for backward compatibility. {@link DruidMetrics#ID} is
+   * deprecated because it's ambiguous and will be removed in a future release.</p>
    */
-  public static Map<String, String[]> mapOfDatasourceAndTaskID(
-      @Nullable final String datasource,
-      @Nullable final String taskId
-  )
+  public static Map<String, String[]> mapOfTaskHolderDimensions(final TaskHolder taskHolder)
```
kfaraz, line 87: I think it would now make sense to move this method into `TaskHolder` itself. The map would be created just once when the `TaskHolder` instance is created.

Then we can just call `taskHolder.getMetricDimensions()` and get the same immutable wherever needed.

services/src/main/java/org/apache/druid/cli/PeonTaskHolder.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * <p>This holder retrieves task information lazily via a {@link Provider} to avoid
+ * circular dependencies during Guice initialization.</p>
+ */
+public class PeonTaskHolder implements TaskHolder
+{
+  private Provider<Task> taskProvider;
+
+  @Inject
+  public PeonTaskHolder(
```
kfaraz: Suggestion: Do not use `@Inject` with this constructor and just pass a `Task` into it instead of a `Provider`.
Use an `@Provides` method in `CliPeon` to bind it correctly.

```java
@Provides
public TaskHolder getTaskHolder(Task task)
{
   return new PeonTaskHolder(task);
}
```

services/src/main/java/org/apache/druid/cli/PeonLoadSpecHolder.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * <p>This holder retrieves load specifications lazily via a {@link Provider} to avoid
+ * circular dependencies during Guice initialization.</p>
+ */
+public class PeonLoadSpecHolder implements LoadSpecHolder
+{
+  private Provider<Task> taskProvider;
+
+  @Inject
+  public PeonLoadSpecHolder(
```
kfaraz: Similar to `PeonTaskHolder`, this constructor should not need `@Inject`. Use a `@Provides` method in `CliPeon` instead.

server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
```diff
@@ -106,24 +98,22 @@ public void testMonitorWithServiceDimensions()
...
-        })
-    );
-    final DataSourceTaskIdHolder dsTaskIdHolder = new DataSourceTaskIdHolder();
-    injector.injectMembers(dsTaskIdHolder);
-    final GroupByStatsMonitor monitor =
-        new GroupByStatsMonitor(groupByStatsProvider, mergeBufferPool, dsTaskIdHolder);
+    final GroupByStatsMonitor monitor = new GroupByStatsMonitor(
+        groupByStatsProvider,
+        mergeBufferPool,
+        new TaskHolder() {
```
kfaraz: Maybe create a `TestTaskHolder` so that we don't have to create this inline class in multiple places.

services/src/main/java/org/apache/druid/cli/PeonTaskHolder.java
abhishekrb19: There are cyclic dependencies involving `Task` when it contains a `TransformSpec`, an `AggregatorFactory` or a monitor, which will pull in `LookupModule`, `MetricsModule`, etc, which depend on these holders. I’ve expanded the Javadoc to explain this in more detail. To break the cycle, this binding needs to use a [Provider](https://stackoverflow.com/questions/37276431/why-using-guice-provider-instead-of-normal-injection)
 instead of direct injection. Switching to `@Provides` pattern will break UTs and some ITs with a cyclic dependency Guice instantiation error:
```
End of classname legend:
========================

	at org.apache.druid.cli.CliPeon.run(CliPeon.java:413)
	at org.apache.druid.cli.Main.main(Main.java:111)
Caused by: java.lang.RuntimeException: com.google.inject.CreationException: Unable to create injector, see the following errors:

1) [Guice/ErrorInCustomProvider]: RuntimeException: ValueInstantiationException: Cannot construct instance of `ExpressionTransform`, problem: Unable to provision, see the following errors:

1) Problem parsing object at prefix[druid.lookup]: Cannot construct instance of `LookupListeningAnnouncerConfig`, problem: Unable to provision, see the following errors:

1) [Guice/ErrorInCustomProvider]: IllegalStateException: This is a proxy used to support circular references.
```

services/src/main/java/org/apache/druid/cli/PeonLoadSpecHolder.java
abhishekrb19: This is intentional - please see https://github.com/apache/druid/pull/18732#discussion_r2573303155

server/src/main/java/org/apache/druid/server/metrics/MetricsModule.java
abhishekrb19: Ah yes, good point, they're no longer required. I've removed the changes originally included in https://github.com/apache/druid/pull/16140 and added some tests that would otherwise fail if we remove this "hack".

services/src/main/java/org/apache/druid/cli/PeonTaskHolder.java
kfaraz: I see, thanks for the clarification!

services/src/test/java/org/apache/druid/cli/CliPeonTest.java
```diff
@@ -198,77 +208,96 @@ public void testCliPeonLocalTmpStorage() throws IOException
...
+        .build();
+
+    final Injector peonInjector = makePeonInjector(new Properties(), compactionTask);
+
+    verifyLoadSpecHolder(peonInjector.getInstance(LoadSpecHolder.class), compactionTask);
+    verifyTaskHolder(peonInjector.getInstance(TaskHolder.class), compactionTask);
+  }
+
+  @Test
+  public void testTaskWithMonitorsAndMetricsSpecDoNotCauseCyclicDependency() throws IOException
```
kfaraz, line 253: thanks for adding these tests!

kfaraz (approved):
Only pending comment is moving the `mapOfDimensions` method to `TaskHolder`, but that is not a blocker for this PR.

abhishekrb19:
Thanks, @kfaraz!

PR #18020, opened 2025-05-20
Fix race in task queue between management and sync loops

jtuglu-netflix:
Fixes a race in the task queue scheduling logic. The race is as follows:

1. T1: Creates a new task T, adds it to `activeTasks` ([here](http://github.com/apache/druid/blob/-/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java#L534)).
2. T2: Starts manageQueuedTasks() [here](https://github.com/apache/druid/blob/-/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java#L390).
3. T3: Starts syncFromStorage() [here](https://github.com/apache/druid/blob/-/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java#L834).
4. T3: Removes the task T since currently we always remove the tasks from `activeTasks`.
5. T2: Sees this task has been removed and proceeds to shut it down [here](https://github.com/apache/druid/blob/-/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java#L399).
6. T3: Adds back the task T to the `activeTasks` [here](https://github.com/apache/druid/blob/-/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java#L858).

Tasks can effectively get started, shutdown, re-added, and then re-started.

Also left a note about why the timestamp ordering matters in the add() method, since it can cause this issue otherwise:

1. T1: read clock [here](https://github.com/apache/druid/blob/-/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java#L532) (T=0)
2. T2: read clock [here](https://github.com/apache/druid/blob/-/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java#L837) (T=1)
3. T2: read from DB
4. T1: commit() to DB
5. T1: add to activeTasks
6. T2: hits [this](https://github.com/apache/druid/blob/3a8d4e2a21dc2c30711ded055a6913c92207d41c/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java#L853) and we get the same issue
0<1 and not in metadata storage.

### Description

Fixes a race condition in the `TaskQueue` scheduling logic.

#### Release note

Fixes a race condition in the `TaskQueue` scheduling logic.

<hr>

##### Key changed/added classes in this PR
[...]

PR #18735, opened 2025-11-11
Add task/waiting/time metric

jtuglu1:
### Description

Adds a metric `task/waiting/time` which measures the time it takes for a task to be placed onto the task runner for scheduling+running. This time encompasses initial persistent storage insertion, lock acquisition time, any task queue-related latencies, noisy-neighbor-related slowness, and any other delays between when a task is first added and when it is scheduled for running on the task runner. Useful for debugging issues related to task startup.

`task/run/time` ~= `task/waiting/time` + `task/pending/time` + \<task run time\> .

This codepath is called in a hot-loop but emission is non-blocking so this should be alright. Manual testing indicates no material increase in overhead.

#### Release note

Add task/waiting/time metric.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/docs/operations/metrics.md
+++ b/docs/operations/metrics.md
@@ -297,8 +297,8 @@ If the JVM does not support CPU time measurement for the current thread, `ingest
 |Metric|Description|Dimensions|Normal value|
 |------|-----------|----------|------------|
 |`task/run/time`|Milliseconds taken to run a task.| `dataSource`, `taskId`, `taskType`, `groupId`, `taskStatus`, `description`, `tags`|Varies|
-|`task/pending/time`|Milliseconds taken for a task to be scheduled and start running.| `dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Varies|
-|`task/waiting/time`|Milliseconds taken for a task to wait (acquire locks, etc.) before being scheduled to run.| `dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Varies|
+|`task/pending/time`|Milliseconds taken for a task to start running after being scheduled by the Overlord.| `dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Varies|
+|`task/waiting/time`|Milliseconds taken for a task to be scheduled to run after being submitted to the Overlord.| `dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Varies|
 |`task/action/run/time`|Milliseconds taken to execute a task action.| `dataSource`, `taskId`, `taskType`, `groupId`, `taskActionType`, `tags`|Varies from subsecond to a few seconds, based on action type.|
 |`task/action/success/count`|Number of task actions that were executed successfully during the emission period. Currently only being emitted for [batched `segmentAllocate` actions](../ingestion/tasks.md#batching-segmentallocate-actions).| `dataSource`, `taskId`, `taskType`, `groupId`, `taskActionType`, `tags`|Varies|
 |`task/action/failed/count`|Number of task actions that failed during the emission period. Currently only being emitted for [batched `segmentAllocate` actions](../ingestion/tasks.md#batching-segmentallocate-actions).| `dataSource`, `taskId`, `taskType`, `groupId`, `taskActionType`, `tags`|Varies|
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
@@ -459,10 +459,8 @@ public class TaskQueue
             // Emit the waiting time for the task
             final ServiceMetricEvent.Builder metricBuilder = new ServiceMetricEvent.Builder();
             IndexTaskUtils.setTaskDimensions(metricBuilder, task);
-            emitter.emit(metricBuilder.setMetric(
-                "task/waiting/time",
-                new Duration(entry.getCreatedTime(), DateTimes.nowUtc()).getMillis())
-            );
+            final long waitDurationMillis = new Duration(entry.getTaskSubmittedTime(), DateTimes.nowUtc()).getMillis();
+            emitter.emit(metricBuilder.setMetric("task/waiting/time", waitDurationMillis));
           } else {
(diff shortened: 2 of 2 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -1163,13 +1172,15 @@ static class TaskEntry
...
     private DateTime lastUpdatedTime;
     private ListenableFuture<TaskStatus> future = null;
     private boolean isComplete = false;
 
-    TaskEntry(TaskInfo taskInfo)
+    TaskEntry(TaskInfo taskInfo, DateTime createdTime)
     {
       this.taskInfo = taskInfo;
+      this.createdTime = createdTime;
       this.lastUpdatedTime = DateTimes.nowUtc();
```
kfaraz, line 1183: Q: Should the `lastUpdatedTime` also be assigned the `createdTime` in the constructor?

I guess the difference between `createdTime` and `DateTimes.nowUtc()` is that the former is the time when the entry was _supposed_ to be updated/created and the latter is the time when it was _actually_ created.

The `lastUpdatedTime` is used to determine which of two competing updates should win in a race condition. Assigning it to `createdTime` in the constructor might ensure that the operation which actually came later is honored.

What do you think, @jtuglu1 ?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
```diff
@@ -454,6 +455,14 @@ private void startPendingTaskOnRunner(TaskEntry entry, ListenableFuture<TaskStat
...
             log.info("Asking taskRunner to run task[%s]", task.getId());
             runnerTaskFuture = taskRunner.run(task);
+
+            // Emit the waiting time for the task
+            final ServiceMetricEvent.Builder metricBuilder = new ServiceMetricEvent.Builder();
+            IndexTaskUtils.setTaskDimensions(metricBuilder, task);
+            emitter.emit(metricBuilder.setMetric(
+                "task/waiting/time",
+                new Duration(entry.getCreatedTime(), DateTimes.nowUtc()).getMillis())
+            );
```
kfaraz: Nit: Might be more readable with:


```suggestion
            final long waitDurationMillis = new Duration(entry.getCreatedTime(), DateTimes.nowUtc()).getMillis();
            emitter.emit(metricBuilder.setMetric("task/waiting/time", waitDurationMillis));
```

indexing-service/src/test/java/org/apache/druid/indexing/overlord/TaskQueueTest.java
```diff
@@ -690,6 +690,52 @@ public TaskStatus runTask(TaskToolbox toolbox)
...
+    final TestTask task1 = new TestTask("t1", Intervals.of("2021-01/P1M"));
+    prepareTaskForLocking(task1);
+    Assert.assertTrue(task1.isReady(actionClientFactory.create(task1)));
+
+    // task2 will not be ready because of task1's lock
+    final TestTask task2 = new TestTask("t2", Intervals.of("2021-01-31/P1M"));
+    taskQueue.add(task2);
+    taskQueue.manageQueuedTasks();
+
+    Thread.sleep(100);
```
kfaraz, line 706: (not required in this PR) I think we should start using the `LatchableEmitter` for these testing needs so that we can remove the sleeps from the unit tests.

docs/operations/metrics.md
```diff
@@ -302,7 +297,8 @@ If the JVM does not support CPU time measurement for the current thread, `ingest
 |Metric|Description|Dimensions|Normal value|
 |------|-----------|----------|------------|
 |`task/run/time`|Milliseconds taken to run a task.| `dataSource`, `taskId`, `taskType`, `groupId`, `taskStatus`, `description`, `tags`|Varies|
-|`task/pending/time`|Milliseconds taken for a task to wait for running.| `dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Varies|
+|`task/pending/time`|Milliseconds taken for a task to be scheduled and start running.| `dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Varies|
```
```suggestion
|`task/pending/time`|Milliseconds taken for a task to start running after being scheduled by the Overlord.| `dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Varies|
```

docs/operations/metrics.md
```diff
@@ -302,7 +297,8 @@ If the JVM does not support CPU time measurement for the current thread, `ingest
...
+|`task/waiting/time`|Milliseconds taken for a task to wait (acquire locks, etc.) before being scheduled to run.| `dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Varies|
```
```suggestion
|`task/waiting/time`|Milliseconds taken for a task to be scheduled to run after being submitted to the Overlord.| `dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Varies|
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
jtuglu1: I'll set as well in the ctor

jtuglu1: Actually, I recall there was a reason this was left as-is. We talked about it [here](https://github.com/apache/druid/pull/18020#discussion_r2097137549).

jtuglu1: I think this lastUpdatedTime should be how it currently is as it's referencing the entry's last updated time. The created time is not referring to metadata about the entry, but rather a close approximation of when the task was "submitted" to the queue (this is before task storage is updated). Let me change `createdTime` to be `taskCreatedTime` to make that explicit.

kfaraz: Thanks for clarifying. The var name could be taskSubmittedTime. Also, maybe add a 1-line comment so that we don't get confused again.

PR #18739, opened 2025-11-12
Allow developer to use system property to override default kafka image for embedded tests

capistrant:
### Description

Some docker runtime setups can struggle to run the Apache Kafka image. Switching to something like Apache kafka-native can work (it works for my apple silicon + colima docker runtime). This change allows using test configurations locally to switch the image without changing source files. It will leave the default to the same image that KafkaResource already uses `apache/kafka:4.0.0`

<hr>

* `KafkaResource`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/KafkaResource.java
+++ b/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/KafkaResource.java
@@ -37,11 +37,19 @@ import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * A Kafka container for use in embedded tests.
+ * <p>
+ * {@link #KAFKA_IMAGE} can be overriden via system property to use a different Kafka Docker image.
+ * </p>
  */
 public class KafkaResource extends TestcontainerResource<KafkaContainer>
 {
-  // Offering an opportunity to override can help some local devs whose system struggles to run apache/kafka. Overriding with apache/kafka-native can help.
-  private static final String KAFKA_IMAGE = System.getProperty("druid.kafka.test.image", "apache/kafka:4.0.0");
+  /**
+   * Kafka Docker image used in embedded tests. The image name is
+   * read from the system property {@code druid.testing.kafka.image} and
+   * defaults to {@code apache/kafka}. Environments that cannot run that
+   * image should set the system property to {@code apache/kafka-native}.
+   */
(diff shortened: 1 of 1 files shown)


jtuglu1:
Have a use for this as well – thanks.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/KafkaResource.java
```diff
@@ -40,7 +40,8 @@
  */
 public class KafkaResource extends TestcontainerResource<KafkaContainer>
 {
-  private static final String KAFKA_IMAGE = "apache/kafka:4.0.0";
+  // Offering an opportunity to override can help some local devs whose system struggles to run apache/kafka. Overriding with apache/kafka-native can help.
+  private static final String KAFKA_IMAGE = System.getProperty("druid.kafka.test.image", "apache/kafka:4.0.0");
```
kfaraz: Nit: Should we use the name `druid.testing.kafka.image` to align with the Druid image property name used in `DruidContainerResource`?

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/KafkaResource.java
```diff
@@ -40,7 +40,8 @@
  */
 public class KafkaResource extends TestcontainerResource<KafkaContainer>
 {
-  private static final String KAFKA_IMAGE = "apache/kafka:4.0.0";
+  // Offering an opportunity to override can help some local devs whose system struggles to run apache/kafka. Overriding with apache/kafka-native can help.
```
kfaraz: Let's convert this to a javadoc.
It would be nice to link it in the class level javadoc too.

Also maybe rephrase to something like:

```suggestion
/**
 * Kafka Docker image used in embedded tests. The image name is
 * read from the system property {@code druid.testing.kafka.image} and
 * defaults to {@code apache/kafka}. Environments that cannot run that
 * image should set the system property to {@code apache/kafka-native}.
 */
```

kfaraz (approved):
Thanks for adding this, @capistrant !
Left a couple of non blocking suggestions.

capistrant:
Thanks for review @kfaraz @jtuglu1 I updated the name per recommendation and have also improved javadoc comments as suggested. Plan to merge once green CI 👍

PR #18741, opened 2025-11-13
fix msq compaction to apply security policies

clintropolis:
### Description
Fixes an issue with MSQ compaction tasks can fail if a policy enforcer is enabled, since compaction generates MSQ tasks directly instead of going through the broker, it would miss out on decorating the `DataSource` as a restricted datasource with the policy set.

To fix, `MSQCompactionRunner` now uses an escalator to create an auth result and uses the policy from that to decorate the query it generates.

To test, I hijacked `CompactionSupervisorTest` and switched it to use MSQ engine and set it up with an auth config and policy enforcer (with a newly added `AllowAllWithPolicyAuthResource` which is essentially allow-all but with `NoRestrictionPolicy` permitted)

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
@@ -387,7 +387,7 @@ public class MSQCompactionRunner implements CompactionRunner
    * {@link AuthorizerMapper} and applies any resulting {@link org.apache.druid.query.policy.Policy} to it using
    * {@link DataSource#withPolicies(Map, PolicyEnforcer)}
    */
-  private DataSource getDataSource(String name)
+  private DataSource getInputDataSource(String name)
   {
     TableDataSource dataSource = new TableDataSource(name);
     final Escalator escalator = injector.getInstance(Escalator.class);
@@ -501,7 +501,7 @@ public class MSQCompactionRunner implements CompactionRunner
     RowSignature rowSignature = getRowSignature(dataSchema);
     VirtualColumns virtualColumns = VirtualColumns.create(new ArrayList<>(inputColToVirtualCol.values()));
     Druids.ScanQueryBuilder scanQueryBuilder = new Druids.ScanQueryBuilder()
-        .dataSource(getDataSource(dataSchema.getDataSource()))
+        .dataSource(getInputDataSource(dataSchema.getDataSource()))
         .columns(rowSignature.getColumnNames())
         .virtualColumns(virtualColumns)
         .columnTypes(rowSignature.getColumnTypes())
@@ -652,7 +652,7 @@ public class MSQCompactionRunner implements CompactionRunner
                             .collect(Collectors.toList());
 
     GroupByQuery.Builder builder = new GroupByQuery.Builder()
-        .setDataSource(getDataSource(compactionTask.getDataSource()))
+        .setDataSource(getInputDataSource(compactionTask.getDataSource()))
         .setVirtualColumns(virtualColumns)
         .setDimFilter(dimFilter)
         .setGranularity(new AllGranularity())
--- a/embedded-tests/pom.xml
+++ b/embedded-tests/pom.xml
@@ -547,6 +547,7 @@
       <version>5.5</version>
       <scope>test</scope>
     </dependency>
+
   </dependencies>
 
   <build>


embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
```diff
@@ -0,0 +1,183 @@
...
+  static class AllowAllWithPolicyEscalator extends NoopEscalator
+  {
+    @Override
+    public AuthenticationResult createEscalatedAuthenticationResult()
+    {
+      return AllowAllWithPolicyAuthenticator.RESULT;
+    }
+  }
+
+  @JsonTypeName(AUTH_TYPE)
```
kfaraz: Why is a json type name needed for the module?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
```diff
@@ -0,0 +1,183 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+public class AllowAllWithPolicyAuthResource implements EmbeddedResource
+{
+  private static final String AUTH_TYPE = "actually-allow-all";
```
kfaraz: Nit: any specific reason for the prefix "actually"? 🙂 Should we use some other name like `allow-all-with-policy`?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
```diff
@@ -0,0 +1,183 @@
...
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+public class AllowAllWithPolicyAuthResource implements EmbeddedResource
```
kfaraz: Please add a short javadoc.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
```diff
@@ -0,0 +1,183 @@
...
+  public static class AllowAllWithPolicyAuthModule implements DruidModule
+  {
+    @Override
+    public void configure(Binder binder)
+    {
+      final MapBinder<String, Authenticator> authenticatorMapBinder = PolyBind.optionBinder(
+          binder,
+          Key.get(Authenticator.class)
+      );
+      authenticatorMapBinder.addBinding(AUTH_TYPE)
```
kfaraz: Nit: can be chained.

```suggestion
      PolyBind.optionBinder(binder, Key.get(Authenticator.class))
              .addBinding(AUTH_TYPE)
```

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
```diff
@@ -293,6 +305,22 @@ public List<MSQControllerTask> createMsqControllerTasks(
     return msqControllerTasks;
   }
 
+  private Query<?> maybeApplyPolicy(DataSchema dataSchema, Query<?> query)
```
kfaraz: Nit: Please add a short javadoc too.

```suggestion
  private Query<?> maybeApplyPolicyOnDatasource(DataSchema dataSchema, Query<?> query)
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  @Override
+  public void beforeStart(EmbeddedDruidCluster cluster)
+  {
+    cluster.addExtension(AllowAllWithPolicyAuthModule.class)
+           .addCommonProperty("druid.auth.authenticatorChain", "[\"test\"]")
+           .addCommonProperty("druid.auth.authenticator.test.type", "actually-allow-all")
+           .addCommonProperty("druid.auth.authorizers", "[\"test\"]")
+           .addCommonProperty("druid.auth.authorizer.test.type", "actually-allow-all")
```
kfaraz: Instead of writing a new one, can't we use the `allowAll` authenticaator/authorizer policy itself alongwith `enforcer.allowedPolicies` config?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -99,7 +105,7 @@ public EmbeddedDruidCluster createCluster()
   public void enableCompactionSupervisors()
   {
     final UpdateResponse updateResponse = cluster.callApi().onLeaderOverlord(
-        o -> o.updateClusterCompactionConfig(new ClusterCompactionConfig(1.0, 100, null, true, null))
+        o -> o.updateClusterCompactionConfig(new ClusterCompactionConfig(1.0, 100, null, true, CompactionEngine.MSQ))
```
kfaraz: Maybe keep both the test cases, native and MSQ.

kfaraz (commented):
Left some minor suggestions.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
clintropolis: it is not, this was an accidental copy and paste when i was setting it on the other classes

embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
clintropolis: i named it this on discovering that `allowAll` does not actually allow everything if you have a policy enforcer configured because it does not set policies. I considered just fixing `allowAll` to be able to set the `NoRestrictionPolicy`, on the other hand there is probably limited scenario where you would have both allowAll and also a policy enforcer configured, and this has a somewhat ugly side-effect of setting a policy always (and so making a restricted datasource on the query) even if you don't have policy enforcement or policies of any sort, which seems a bit odd for out of the box behavior, so I did this instead.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
clintropolis: guilty, i just copied these from other auth modules 🙃

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
clintropolis: i didn't think this was needed since `Policy` apply to datasources so it seemed redundant, but can rename if you feel like it isn't obvious

embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
clintropolis: see https://github.com/apache/druid/pull/18741#discussion_r2528439580, allowAll doesn't work if you have an enforcer defined because it does not set any policy on the auth result (and doing so would have some undesirable side-effects)

embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/AllowAllWithPolicyAuthResource.java
gianm: Part of the purpose of the enforcer mechanism is to allow people to verify that `allowAll` is not being used by accident. So, adding a no-restriction policy to `allowAll` wouldn't be a "fix", it would partially defeat the purpose of the enforcer. It's fine to have an option for tests though that does this.

clintropolis: i changed things so that allowAll itself can have an optional policy defined, so deleted this stuff since is no longer necessary

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
clintropolis: this is changed now to just be `getInputDataSource` and is used by the 'build' query methods directly to create the `DataSource` instead of rewriting the query later

PR #18742, opened 2025-11-14
add StorageMonitor to measure storage and virtual storage usage by segment cache

clintropolis:
### Description
To use, add `org.apache.druid.server.metrics.StorageMonitor` to `druid.monitoring.monitors`.

changes:
* adds `StorageLocationStats` and `VirtualStorageLocationStats` to abstract measuring segment cache/virtual segment cache usage
* adds `StorageStats` to collect `StorageLocationStats` and `VirtualStorageLocationStats` into maps between location label and the stats of that location
* adds new method `SegmentCacheManager.getStorageStats` to expose stats collection for segment cache manager implementations
* adds new `StorageMonitor` to emit metrics for values in `StorageStats`
* metrics for `StorageLocationStats`:
  - storage/used/bytes
  - storage/load/count
  - storage/load/bytes
  - storage/drop/count
  - storage/drop/bytes
* metrics for `VirtualStorageLocationStats`:
  - storage/virtual/used/bytes
  - storage/virtual/hit/count
  - storage/virtual/load/count
  - storage/virtual/load/bytes
  - storage/virtual/evict/count
  - storage/virtual/evict/bytes
  - storage/virtual/reject/count


server/src/main/java/org/apache/druid/server/metrics/StorageMonitor.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.util.Map;
+
+@LoadScope(roles = {
+    NodeRole.BROKER_JSON_NAME,
+    NodeRole.HISTORICAL_JSON_NAME,
+    NodeRole.INDEXER_JSON_NAME,
+    NodeRole.PEON_JSON_NAME
+})
+public class StorageMonitor extends AbstractMonitor
```
kfaraz, line 44: Please add a short javadoc for this monitor. Would be nice to include it in docs too.

server/src/main/java/org/apache/druid/server/metrics/StorageMonitor.java
```diff
@@ -0,0 +1,107 @@
...
+      }
+
+      for (Map.Entry<String, VirtualStorageLocationStats> location : stats.getVirtualLocationStats().entrySet()) {
+        final VirtualStorageLocationStats weakStats = location.getValue();
+        final ServiceMetricEvent.Builder builder = new ServiceMetricEvent.Builder().setDimension(
+            LOCATION_DIMENSION,
+            location.getKey()
+        );
+        builder.setMetric(VSF_HIT_COUNT, weakStats.getHitCount());
+        emitter.emit(builder);
```
kfaraz: Nit: Maybe combine these lines for brevity (unless you prefer the current syntax more)
```suggestion
        emitter.emit(builder.setMetric(VSF_HIT_COUNT, weakStats.getHitCount()));
```

server/src/main/java/org/apache/druid/server/metrics/StorageMonitor.java
```diff
@@ -0,0 +1,107 @@
...
+
+@LoadScope(roles = {
+    NodeRole.BROKER_JSON_NAME,
+    NodeRole.HISTORICAL_JSON_NAME,
+    NodeRole.INDEXER_JSON_NAME,
+    NodeRole.PEON_JSON_NAME
+})
+public class StorageMonitor extends AbstractMonitor
+{
+  public static final String LOCATION_DIMENSION = "location";
```
kfaraz, line 46: Should we move this to `DruidMetrics` class?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
```diff
@@ -152,88 +158,141 @@ void testQueryPartials()
...
-      assertMetrics(i + 5, null);
+      assertQueryMetrics(i + 5, null);
+      long actualLoads = getMetricLatestEvent(emitter, DefaultQueryMetrics.QUERY_ON_DEMAND_LOAD_COUNT, i + 5);
+      expectedTotalLoad += actualLoads;
+      expectedTotalHits += (expectedLoads[nextQuery] - actualLoads);
     }
+
+    sleepForStorageMonitor();
+
+    emitter.waitForAnyEventWithMetricName(StorageMonitor.VSF_HIT_COUNT);
```
kfaraz: Nit: Do we need the new `waitForAnyEventWithMetricName` method?

Does the original syntax seem verbose?

```suggestion
    emitter.waitForEvent(event -> event.hasMetricName(StorageMonitor.VSF_HIT_COUNT));
```

Just wondering if we there is something we can do to make the syntax more dev-friendly for the general case too, since matching on just the metric name doesn't seem like a common use case.

server/src/main/java/org/apache/druid/segment/loading/SegmentCacheManager.java
```diff
@@ -138,4 +138,11 @@ public interface SegmentCacheManager
   void shutdownBootstrap();
 
   void shutdown();
+
+  /**
+   * Collect {@link StorageStats}, if available. Calling this method 'resets' the stats values, so that the measurements
+   * returned by this method only includes the events which have happened since the previous call to this method.
+   */
+  @Nullable
+  StorageStats getStorageStats();
```
kfaraz, line 146: Maybe rename to `getAndResetStorageStats()` to align with behaviour?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
```diff
@@ -152,88 +158,141 @@ void testQueryPartials()
...
+
+  private long getMetricLatestEvent(LatchableEmitter emitter, String metricName, int expectedCount)
+  {
+    final int lastIndex = expectedCount - 1;
+    List<ServiceMetricEvent> events = emitter.getMetricEvents(metricName);
+    Assertions.assertEquals(expectedCount, events.size());
+    return events.get(lastIndex).getValue().longValue();
+  }
+
+  private long getMetricTotal(LatchableEmitter emitter, String metricName)
```
kfaraz: I suppose this should be a utility method on `StubServiceEmitter` or `MetricsVerifier` itself. But we can do that later too.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
```diff
@@ -152,88 +158,141 @@ void testQueryPartials()
...
-        8683
-    };
+    final long[] expectedResults = new long[]{9770, 10524, 10267, 8683};
+    final long[] expectedLoads = new long[]{8L, 6L, 5L, 5L};
+
+
+    LatchableEmitter emitter = historical.latchableEmitter();
+    // sleep to clear out the pipe to get zerod out storage monitor metrics and then flush (which clears out the
+    // internal events stores in test emitter)
+    sleepForStorageMonitor();
```
kfaraz: Can we avoid the sleep somehow? Maybe by waiting for some specific metric to be emitted?

server/src/main/java/org/apache/druid/segment/loading/StorageLocationStats.java
```diff
@@ -0,0 +1,31 @@
...
+ * under the License.
+ */
+
+package org.apache.druid.segment.loading;
+
+public interface StorageLocationStats
+{
+  long getLoadCount();
+
+  long getLoadBytes();
```
kfaraz, line 37: Does this represent the load events that have occurred since the last emission period?
Or the number/bytes of segments currently loaded at a location?

The latter seems like a useful metric too. We already have the `segment/used` and `segment/usedPercent` metrics emitted by the `HistoricalMetricsMonitor`, but they are not keyed on the location as a dimension.

server/src/main/java/org/apache/druid/segment/loading/VirtualStorageLocationStats.java
```diff
@@ -0,0 +1,35 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.segment.loading;
+
+public interface VirtualStorageLocationStats
```
kfaraz, line 22: Should this extend `StorageLocationStats`?

server/src/main/java/org/apache/druid/segment/loading/StorageStats.java
```diff
@@ -0,0 +1,61 @@
...
+  private final Map<String, StorageLocationStats> stats;
+  private final Map<String, VirtualStorageLocationStats> virtualStats;
+
+  public StorageStats(
+      final Map<String, StorageLocationStats> stats,
+      final Map<String, VirtualStorageLocationStats> virtualStats
+  )
+  {
+    this.stats = stats;
+    this.virtualStats = virtualStats;
```
kfaraz, line 46: Can we ever have a case where both maps are non-empty i.e. some static storage locations and some virtual on the same service?

kfaraz (commented):
Thanks for adding the metrics, @clintropolis !
Left some suggestions/queries, mostly for my understanding.

server/src/main/java/org/apache/druid/server/metrics/StorageMonitor.java
clintropolis: can add javadocs, but i haven't documented any virtual storage stuff yet, so will leave documentation for later PR

server/src/main/java/org/apache/druid/server/metrics/StorageMonitor.java
clintropolis: just the dimension? idk, it didn't seem like it would be commonly used by anything else so just put everything here :shrug: should it be there?

server/src/main/java/org/apache/druid/segment/loading/StorageStats.java
clintropolis: not right now, we could in the future once we support defining virtual loads via load rules instead of the 'all or nothing' virtual storage mode we have right now

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
clintropolis: i had some trouble getting waiting for events from the same monitor to work, looking into the code i think i see why, waitForEvent doesn't necessarily wait for the next event, it finds any processed matching event too, i believe i misunderstood what this method did, I think i need a version that skips any already emitted events and only waits for the next one

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
clintropolis: yea, i guess this method isn't necessary, i guess i was thinking it would be common for looking for things from monitors like i'm doing here, but i guess it doesn't save all that much

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
kfaraz: > I think i need a version that skips any already emitted events and only waits for the next one

I see. That seems like a reasonable use case. I think you can add a method `waitForNextEvent` but we would need to call out that it would be susceptible to race conditions and may miss out on events and cause the test to wait forever.

server/src/main/java/org/apache/druid/segment/loading/StorageStats.java
kfaraz: Makes sense, thanks for the clarification!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
kfaraz: Yeah, I think we can leave it out for now.

server/src/main/java/org/apache/druid/server/metrics/StorageMonitor.java
kfaraz: I suppose other metrics might want to use a `location` dimension in the future, but we can leave it here too for the time being.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
clintropolis: this is more or less exactly what i have done, just haven't got around to pushing up a commit yet tidying some other stuff up

server/src/main/java/org/apache/druid/segment/loading/StorageLocationStats.java
clintropolis: right now these metrics are all just for the emission period. I was aiming for these metrics to show like the activity during the period to give a good idea of segment cache activity rather than overall usage, the static load metrics showing churn in rebalancing, while the vsf metrics could be used as an indicator that there is a lot of churn from queries targeting different segments competing for disk space.

That said, it does feel useful to have some metrics like you are suggesting as well, especially later once we can have mixed disk usage from virtual loads via load rules... will think on it a bit.

server/src/main/java/org/apache/druid/segment/loading/VirtualStorageLocationStats.java
clintropolis: they don't quite line up, there is not an explicit 'drop' in virtual storage, instead they are always evicted when something else needs the space, so i had them separate so the names are clearer to their function. I also didn't see a lot of usefulness in them being the same type, since they don't emit the same metrics, the only thing they both have in common is a location dimension and loads. I guess later when load rules would allow mixed usage of disk it might be nice to have one object per location, but it would also need to distinguish static loads from virtual loads, so need to think about it a bit.

clintropolis: went ahead and added used bytes, which unlike the others is just the level at the time of emission

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
```diff
@@ -123,6 +124,28 @@ public List<Number> getMetricValues(
...
+    for (ServiceMetricEvent event : metricEventQueue) {
+      total += event.getValue().longValue();
+    }
+    return total;
+  }
+
+  public Number getLatestMetricEventValue(String metricName)
+  {
+    final Deque<ServiceMetricEvent> metricEventQueue = metricEvents.get(metricName);
+    return metricEventQueue == null ? 0 : metricEventQueue.getLast().getValue();
```
kfaraz: Nit: should we return `null` if there has been no event?

divy-imply:
👋 Can we add a plain text mapping to metric name to meaning. 
naive example:
`storage/used/bytes` : data stored in hot tier/cache

by reading the PR description, i understand these metrics are per segment usage. how does it translate to tier labeled as hot?

clintropolis:
>👋 Can we add a plain text mapping to metric name to meaning.
naive example:
storage/used/bytes : data stored in hot tier/cache

> by reading the PR description, i understand these metrics are per segment usage. how does it translate to tier labeled as hot?

Well, there isn't a really clear way to do this since tiers are just labels, so there isn't really a concept of 'hot', that is something operators decide depending on configurations and such. Also, in the future I plan to support specifying virtual loads via load rules instead of a system level configuration, so at that point the same node could be emitting both regular and virtual metrics, so trying to frame these metrics in terms of tiers might be confusing.

PR #18743, opened 2025-11-14
protect against an illegal sublist if hydrants is empty

capistrant:
### Description

If hydrants is empty here, it can lead to an IAE creating a sublist of hydrants with invalid indices.

<hr>

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
+++ b/server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
@@ -2546,6 +2546,65 @@ public class StreamAppenderatorTest extends InitializedNullHandlingTest
     }
   }
 
+  /**
+   * Tests an edge case where a writable sink exists but has no hydrants.
+   * <p>
+   * The test scenario is somewhat unrealistic, but we are trying to prevent a regression because this issue was seen in
+   * the wild and resulted in failed indexing tasks due to an IAE.
+   * </p>
+   */
+  @Test
+  public void testPersistAllWithEmptySink() throws Exception
+  {
+    try (
+        final StreamAppenderatorTester tester =
+            new StreamAppenderatorTester.Builder().maxRowsInMemory(100)
+                                                  .basePersistDirectory(temporaryFolder.newFolder())
+                                                  .build()) {
+      final Appenderator appenderator = tester.getAppenderator();
+      final Supplier<Committer> committerSupplier = () -> new Committer()
+      {
+        @Override
+        public Object getMetadata()
+        {
+          return ImmutableMap.of("eventCount", 0);
+        }
+
+        @Override
+        public void run()
+        {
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -653,7 +653,7 @@ public ListenableFuture<Object> persistAll(@Nullable final Committer committer)
       numPersistedRows += sink.getNumRowsInMemory();
       bytesPersisted += sink.getBytesInMemory();
 
-      final int limit = sink.isWritable() ? hydrants.size() - 1 : hydrants.size();
+      final int limit = sink.isWritable() ? Integer.max(0, hydrants.size() - 1) : hydrants.size();
```
kfaraz, line 656: Is it possible to have a small UT for this?

kfaraz (approved):

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
capistrant: I couldn't figure out how to accomplish it. So for full disclosure the test case I came up with is courtesy of Claude coming up with a way to force the empty list of hydrants scenario. Confirmed that without the fix it is an IAE and with the change, no exception.

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2546,6 +2546,65 @@ public void test_dropSegment_skipsUnlockInterval_ifOverlappingSinkIsActive() thr
     }
   }
 
+  /**
+   * Tests an edge case where a writable sink exists but has no hydrants.
+   * <p>
+   * The test scenario is somewhat unrealistic, but we are trying to prevent a regression because this issue was seen in
+   * the wild and resulted in failed indexing tasks due to an IAE.
+   * </p>
+   */
```
kfaraz, line 2555: I think the javadoc here can be omitted.
Also, since it was already reported, I think it is a valid scenario.

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2546,6 +2546,65 @@ public void test_dropSegment_skipsUnlockInterval_ifOverlappingSinkIsActive() thr
...
+  @Test
+  public void testPersistAllWithEmptySink() throws Exception
```
kfaraz, line 2557: Nit: update the method name to reflect the expected behaviour with the suffix `IsNoop` or `DoesNotFail`, etc.

```suggestion
  public void testPersistAllWithEmptySinkIsNoop() throws Exception
```

PR #18745, opened 2025-11-14
Improve lag-based autoscaler config persistence

Fly-Style:
- When a supervisor is updated via API, the logic is following this order:
provided `taskCountStart` > provided `taskCount` > existing `taskCount` > provided `taskCountMin`.

 * `SupervisorManager`
 * `SeekableStreamSupervisorSpec`
 
This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/test/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResourceTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResourceTest.java
@@ -364,10 +364,8 @@ public class SupervisorResourceTest extends EasyMockSupport
     Assert.assertTrue(
         specs.stream()
              .allMatch(spec ->
-                           ("id1".equals(spec.getId()) && spec.getDataSource().equals("datasource1") && SPEC1.equals(
-                               spec.getSpec())) ||
-                           ("id2".equals(spec.getId()) && spec.getDataSource().equals("datasource2") && SPEC2.equals(
-                               spec.getSpec()))
+                           ("id1".equals(spec.getId()) && spec.getDataSource().equals("datasource1") && SPEC1.equals(spec.getSpec())) ||
+                           ("id2".equals(spec.getId()) && spec.getDataSource().equals("datasource2") && SPEC2.equals(spec.getSpec()))
              )
     );
   }
@@ -1007,9 +1005,7 @@ public class SupervisorResourceTest extends EasyMockSupport
     EasyMock.expect(taskMaster.getSupervisorManager()).andReturn(Optional.of(supervisorManager)).times(3);
     EasyMock.expect(supervisorManager.getSupervisorHistoryForId("id1", null)).andReturn(versions1).times(1);
     EasyMock.expect(supervisorManager.getSupervisorHistoryForId("id2", null)).andReturn(versions2).times(1);
-    EasyMock.expect(supervisorManager.getSupervisorHistoryForId("id3", null))
-            .andReturn(Collections.emptyList())
-            .times(1);
+    EasyMock.expect(supervisorManager.getSupervisorHistoryForId("id3", null)).andReturn(Collections.emptyList()).times(1);
     setupMockRequest();
     replayAll();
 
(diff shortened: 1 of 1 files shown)


jtuglu1:
> When the autoscaler does a scaling action, it updates ioConfig.autoScalerConfig.taskCountStart, not ioConfig.taskCount

Hi, @Fly-Style thanks for the change. Not sure the above makes sense to me. `autoScalerConfig.taskCountStart` is supposed to be an immutable value used on supervisor submit to allow the scaler to not have to start at minimum task count (this makes having a smaller minimum easier/possible). Scaling actions should not touch this `taskCountStart` value (it is meant to be immutable value in the spec to all supervisor to return to baseline on resubmit). In a way, you can expect this to be the "average" task count. From the docs:

> Optional config to specify the number of ingestion tasks to start with. When you enable the autoscaler, Druid ignores the value of taskCount in ioConfig and, if specified, starts with the taskCountStart number of tasks. Otherwise, defaults to taskCountMin.

`ioConfig.taskCount` is the currently running taskCount and should be what is updated. There are other things that rely on the `ioConfig.taskCount` value being updated/accurate, like the `stopTaskCount` calculations (both fixed and variable).

Looks like you want to have a way to not "reset" the task count during re-submits, and have the value be sticky. I would prefer this not be the default behavior, but rather opt-in since:
- Your "current" task count may not accurately reflect the true current load of the system. "Resubmitting" the supervisor should be (IMO) equivalent to "reseting" the supervisor. 
- It's nice to have a way to "reset" this value back to the expected baseline task count. For supervisors running large (~1000s) of tasks often times supervisors can become bloated and not scale down fast enough. Resubmitting them allows task count to return to baseline.

jtuglu1:
Similarly, have you tested what happens when a supervisor is terminated (tombstoned) and then resubmitted without a taskCountStart? It'd be good to make sure that we don't end up merging an old supervisor (with same ID) data with potentially unrelated, new supervisor's data.

kfaraz:
@Fly-Style , I think @jtuglu1 makes a fair point. I too would prefer that `taskCountStart` remain immutable and we retain the capability to reset the task count upon resubmission.

How about the following approach instead?

- [x] When submitting a supervisor with auto-scaler disabled, use the `taskCount` as is.
- [ ] When submitting a (new or existing) supervisor with auto-scaler enabled, set `taskCount = taskCountStart == null ? taskCountMin : taskCountStart`.
Persist the supervisor.
- [x] For any auto-scaling event, update the `taskCount` the same way that we do today.
- [ ] When the supervisor starts, it should start with the `taskCount` as the starting number of tasks rather than `taskCountStart`.

This would ensure:
- the issue with Overlord restarts is fixed
- `taskCountStart` remains an immutable config
- `taskCount` always reflects the current task count in the supervisor

What do you think?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScaler.java
```diff
@@ -190,16 +190,22 @@ private Runnable computeAndCollectLag()
...
-   * Finaly, if scaleOutThreshold/scaleInThreshold is higher than triggerScaleOutFractionThreshold/triggerScaleInFractionThreshold, scale out/in action would be triggered.
-   *
-   * @param lags the lag metrics of Stream(Kafka/Kinesis)
-   * @return Integer. target number of tasksCount, -1 means skip scale action.
-   */
+    /**
+     * This method determines whether to do scale actions based on collected lag points.
+     * The current algorithm of scale is straightforward:
+     * <ul>
+     * <li>First, compute the proportion of lag points higher/lower than <code>scaleOutThreshold</code>/<code>scaleInThreshold</code>,
```
kfaraz: You can also use the `{@code}` javadoc tag as it is less verbose.
```suggestion
     * <li>First, compute the proportion of lag points higher/lower than {@code scaleOutThreshold/scaleInThreshold},
```

gianm:
At our deployment, we do want supervisors to stick to the current task count when resubmitted. The reasons are:

- Supervisors need to be updated periodically to do stuff like change tuning configs, change schemas, etc. When we do a change like this, we don't want the task count to change from whatever is currently running.
- With the way we use autoscaling, the `taskCountStart` is just some fixed starter value that is the same for all tables, not anything actually related to that particular table. We rely on the autoscaler to find the ideal value. So, reverting back to the `taskCountStart` is not desirable.

I'm open to some way of offering both behaviors. I personally feel like sticking to the current task count on resubmitting is more intuitive as a default (since it seems odd that changing schema would lead to a reset of the task count). But I can live with the current behavior as default, as long as there's some way to get the stick-to-current behavior. Currently, there isn't.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -556,44 +566,52 @@ public String getType()
...
+                                 .setDimensionIfNotNull(
+                                     DruidMetrics.TAGS,
+                                     spec.getContextValue(DruidMetrics.TAGS)
+                                 )
+                                 .setMetric(
+                                     AUTOSCALER_SCALING_TIME_METRIC,
+                                     scaleActionStopwatch.millisElapsed()
+                                 ));
+    log.info("Changed taskCount to [%s] for supervisor[%s] for dataSource[%s].", desiredActiveTaskCount, supervisorId, dataSource);
+    return true;
```
kfaraz: I guess this is just moving the code out of the `else` block. Can we revert this change for the time being?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1161,7 +1179,7 @@ public void stop(boolean stopGracefully)
       catch (Exception e) {
         stateManager.recordThrowableEvent(e);
         log.makeAlert(e, "Exception stopping [%s]", supervisorId)
-           .emit();
+            .emit();
```
kfaraz: not needed?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1339,7 +1357,7 @@ public void tryInit()
         }
         initRetryCounter++;
         log.makeAlert(e, "Exception starting SeekableStreamSupervisor[%s]", supervisorId)
-           .emit();
+            .emit();
```
kfaraz: not needed?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -255,6 +263,32 @@ public void validateSpecUpdateTo(SupervisorSpec proposedSpec) throws DruidExcept
...
+  @Override
+  public void mergeSpecConfigs(@NotNull SupervisorSpec existingSpec)
+  {
+    AutoScalerConfig thisAutoScalerConfig = this.getIoConfig().getAutoScalerConfig();
+    // Either if autoscaler is absent or taskCountStart is specified - just return.
+    if (thisAutoScalerConfig == null || thisAutoScalerConfig.getTaskCountStart() != null) {
+      return;
+    }
+
+    // TODO[sasha]: use switch expression with pattern matching when we move to Java 21 as minimum requirement.
```
kfaraz:
```suggestion
    // Use switch expression with pattern matching when we move to Java 21 as minimum requirement.
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -255,6 +263,32 @@ public void validateSpecUpdateTo(SupervisorSpec proposedSpec) throws DruidExcept
...
+    if (existingSpec instanceof SeekableStreamSupervisorSpec) {
+      // Note: for some reason, sources are available only for bytecode version 11.
```
kfaraz: Please remove this.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -255,6 +263,32 @@ public void validateSpecUpdateTo(SupervisorSpec proposedSpec) throws DruidExcept
...
+      //noinspection PatternVariableCanBeUsed
+      var spec = (SeekableStreamSupervisorSpec) existingSpec;
```
kfaraz:
```suggestion
      SeekableStreamSupervisorSpec spec = (SeekableStreamSupervisorSpec) existingSpec;
```

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorSpec.java
```diff
@@ -113,4 +114,15 @@ default void validateSpecUpdateTo(SupervisorSpec proposedSpec) throws DruidExcep
...
     // The default implementation does not do any validation checks.
   }
+
+  /**
+   * Perform any merging of spec configurations needed after deserialization.
+   *
+   * @param existingSpec used spec to merge values from
+   * @throws DruidException if the spec update is not allowed
+   */
+  default void mergeSpecConfigs(@NotNull SupervisorSpec existingSpec)
```
kfaraz:
```suggestion
  default void merge(@NotNull SupervisorSpec existingSpec)
```

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorSpec.java
```diff
@@ -113,4 +114,15 @@ default void validateSpecUpdateTo(SupervisorSpec proposedSpec) throws DruidExcep
   {
     // The default implementation does not do any validation checks.
   }
+
+  /**
+   * Perform any merging of spec configurations needed after deserialization.
```
kfaraz:
```suggestion
   * Updates this supervisor spec by merging values from the given {@code existingSpec}.
   * This method may be used to carry forward existing spec values when a supervisor is being resubmitted.
```

kfaraz:
> Similarly, have you tested what happens when a supervisor is terminated (tombstoned) and then resubmitted without a taskCountStart? It'd be good to make sure that we don't end up merging an old supervisor (with same ID) data with potentially unrelated, new supervisor's data.

@jtuglu1 , this is already handled since the `SupervisorManager` contains only the non-tombstoned latest supervisor versions in memory. But we could add an embedded test to verify the same.

@Fly-Style , let's try to add an embedded test for `LagBasedAutoScaler` similar to what you are doing in the other PR for cost-based auto-scaler. The test could have a method which verifies that we do not pick up the `taskCountStart` from tombstoned versions, even if the supervisor id is the same.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
Fly-Style: Yeap, it is obsolete in my taste.

Fly-Style:
> let's try to add an embedded test for LagBasedAutoScaler similar to what you are doing in the other PR for cost-based auto-scaler. 

Would love to, but in separate PR if you don't mind :)

gianm:
> I'm open to some way of offering both behaviors. I personally feel like sticking to the current task count on resubmitting is more intuitive as a default (since it seems odd that changing schema would lead to a reset of the task count). But I can live with the current behavior as default, as long as there's some way to get the stick-to-current behavior. Currently, there isn't.

A possibility:

- When a supervisor is started through `supervisor.start()`, initial task count is always taken from `taskCount`. This keeps task count consistent through Overlord restarts, etc.
- When a supervisor is posted through `SupervisorResource#specPost`, `taskCount` is set to:
  1. `taskCountStart` if that is nonnull
  2. else, the user-provided `taskCount` if that is nonnull
  3. else, the `taskCount` from the previous supervisor spec (in the DB) if one exists
  4. else, the user-provided `taskCountMin`

This I think allows @jtuglu1 and me to both have the behavior we want: @jtuglu1 would set `taskCountStart` and the task count would always reset to that. I would set neither `taskCount` nor `taskCountStart`, and the task count would start out at `taskCountMin` for a fresh supervisor, then retain the current count from then on.

kfaraz:
Thanks for the suggestion, @gianm ! I think that would work nicely and meet all our needs.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/AutoScalerConfig.java
```diff
@@ -40,6 +40,7 @@ public interface AutoScalerConfig
   int getTaskCountMax();
   int getTaskCountMin();
   Integer getTaskCountStart();
+  void setTaskCountStart(int taskCountStart);
```
kfaraz: This may not be needed anymore.

jtuglu1:
SGTM

kfaraz (commented):
@Fly-Style , based on the decided approach, we should do the following:

- Continue persisting `taskCount` (and not `taskCountStart`) whenever an auto-scale event occurs.
- Leave `taskCountStart` as an immutable config
- While merging spec on resubmission, follow this order:
provided `taskCountStart` > provided `taskCount` > existing `taskCount` > provided `taskCountMin`.
- Update docs and add a release note in PR description to reflect this change in behaviour.

Please let me know if anything seems ambiguous.

Fly-Style:
@gianm, @kfaraz, @jtuglu1 thanks a lot for a productive discussion! Happy to implement this according to all requests!

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
kfaraz: Please revert this change to avoid bloating the diff.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -556,43 +566,51 @@ public String getType()
...
   }
 
-  private void changeTaskCountInIOConfig(int desiredActiveTaskCount)
+  private void changeTaskCountInAutoScalerConfig(int desiredActiveTaskCount)
   {
+    // Sanity check.
+    if (autoScalerConfig == null) {
+      log.warn("autoScalerConfig is null but scale action is submitted, how can it be ?");
+      return;
+    }
```
kfaraz: Please revert this change as well since the old method name is still relevant.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
```diff
@@ -389,7 +389,7 @@ public SeekableStreamIndexTaskClient<KafkaTopicPartition, Long> build(
 
     supervisor.start();
     int taskCountBeforeScale = supervisor.getIoConfig().getTaskCount();
-    Assert.assertEquals(1, taskCountBeforeScale);
+    Assert.assertNull(supervisor.getIoConfig().getAutoScalerConfig().getTaskCountStart());
```
kfaraz: Please revert the changes to `KafkaSupervisorTest` and `KinesisSupervisorTest`.
I don't think they are needed anymore.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -183,15 +185,16 @@ public DruidMonitorSchedulerConfig getMonitorSchedulerConfig()
...
    * @param supervisor
    * @return autoScaler
    */
   @Override
   public SupervisorTaskAutoScaler createAutoscaler(Supervisor supervisor)
   {
-    AutoScalerConfig autoScalerConfig = ingestionSchema.getIOConfig().getAutoScalerConfig();
-    if (autoScalerConfig != null && autoScalerConfig.getEnableTaskAutoScaler() && supervisor instanceof SeekableStreamSupervisor) {
-      return autoScalerConfig.createAutoScaler(supervisor, this, emitter);
+    AutoScalerConfig autoScalerCfg = ingestionSchema.getIOConfig().getAutoScalerConfig();
```
kfaraz: Please revert this rename as it does not seem required.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/SupervisorResourceConfigMergeTest.java
```diff
@@ -0,0 +1,533 @@
...
+import javax.annotation.Nullable;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.core.Response;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Set;
+
+@RunWith(EasyMockRunner.class)
+public class SupervisorResourceConfigMergeTest extends EasyMockSupport
```
kfaraz: Please move the test methods in this class to `SupervisorResourceTest` itself.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -255,6 +261,33 @@ public void validateSpecUpdateTo(SupervisorSpec proposedSpec) throws DruidExcept
...
   }
 
+  @Override
+  public void merge(@NotNull SupervisorSpec existingSpec)
+  {
+    AutoScalerConfig thisAutoScalerConfig = this.getIoConfig().getAutoScalerConfig();
+    // Either if autoscaler is absent or taskCountStart is specified - just return.
+    if (thisAutoScalerConfig == null || thisAutoScalerConfig.getTaskCountStart() != null) {
+      return;
+    }
```
kfaraz, line 271: Should we also return early if `this.ioConfig.getTaskCount()` is specified?

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/SupervisorResourceConfigMergeTest.java
```diff
@@ -0,0 +1,533 @@
...
+        auditManager
+    );
+  }
+
+  /**
+   * Tests that when a new spec is submitted without taskCountStart, the merge function
+   * follows the priority: provided taskCountStart > provided taskCount > existing taskCount > provided taskCountMin.
+   * In this test, existing spec has taskCount=5, new spec has no taskCount set,
+   * so merge should use existing taskCount (5).
+   */
```
kfaraz: Can we remove the javadocs and update the method name to include the expectation/condition?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
Fly-Style: Not really, taskCountStart has bigger priority :)

> the priority: provided `taskCountStart` > provided `taskCount` > existing `taskCount` > provided `taskCountMin`.

kfaraz (approved):
Looks good, thanks for the changes @Fly-Style ! 🚀

PR #18752, opened 2025-11-18
ITSqlQueryTest migration to embedded test suite

Fly-Style:
Part of the general effort to move as much as possible integration tests to new shiny embedded test suite, now in SQL part.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/SqlQueryHttpRequestHeadersTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/SqlQueryHttpRequestHeadersTest.java
@@ -56,7 +56,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
 {
   private static final String SQL_QUERY_ROUTE = "%s/druid/v2/sql/";
 
-  public static List<Boolean> flags = List.of(true, false);
+  public static List<Boolean> shouldUseQueryBroker = List.of(true, false);
 
   private final EmbeddedBroker broker = new EmbeddedBroker();
   private final EmbeddedRouter router = new EmbeddedRouter();
@@ -132,7 +132,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testNullContentType(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -148,7 +148,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testUnsupportedContentType(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -164,7 +164,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testTextPlain(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -181,7 +181,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testFormURLEncoded(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -198,7 +198,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testFormURLEncoded_InvalidEncoding(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -215,7 +215,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testJSON(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -244,7 +244,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testInvalidJSONFormat(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -261,7 +261,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testEmptyQuery_TextPlain(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -278,7 +278,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testEmptyQuery_UrlEncoded(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -295,7 +295,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testBlankQuery_TextPlain(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -312,7 +312,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testEmptyQuery_JSON(boolean shouldQueryBroker)
   {
     executeQuery(
@@ -329,7 +329,7 @@ public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
   }
 
   @ParameterizedTest
-  @FieldSource("flags")
+  @FieldSource("shouldUseQueryBroker")
   public void testMultipleContentType_usesFirstOne(boolean shouldQueryBroker)
   {
     executeQuery(


processing/src/main/java/org/apache/druid/query/http/ClientSqlQuery.java
```diff
@@ -39,6 +39,9 @@ public class ClientSqlQuery
   @JsonProperty
   private final String resultFormat;
 
+  /**
+   * Indicates whether the query result should include a header.
+   */
```
kfaraz, line 44: Alternative to this, we may also just rename the getter method to `includeHeader` similar to how it is done in class `SqlQuery` (which is the Broker-side equivalent of this POJO).

(this change may be done in a separate PR).

kfaraz:
Thanks for renaming the file, @Fly-Style !
But unfortunately, git still keeps treating is a new file 😅 
Probably because the file has changed too much. I think it is fine and we can review it as it is.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
```diff
@@ -0,0 +1,379 @@
...
+import javax.ws.rs.core.MediaType;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
+
+public class ITSqlQueryTest extends EmbeddedClusterTestBase
```
kfaraz: @Fly-Style , please feel free to rename this class to something more appropriate since GitHub is still treating is a separate file.
Also, maybe add a 1-line javadoc on what the test is about.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
```diff
@@ -0,0 +1,379 @@
...
+                               .addServer(broker)
+                               .addServer(router)
+                               .addServer(new EmbeddedIndexer());
+  }
+
+  @BeforeEach
+  void setUp()
+  {
+    httpClientRef = router.bindings().globalHttpClient();
+    brokerEndpoint = String.format(SQL_QUERY_ROUTE, getServerUrl(broker));
```
kfaraz: This variant of `String.format` has been marked as a forbidden API, I think since it does not explicitly specify a locale.

```suggestion
    brokerEndpoint = StringUtils.format(SQL_QUERY_ROUTE, getServerUrl(broker));
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/SqlQueryHttpRequestHeadersTest.java
```diff
@@ -0,0 +1,379 @@
...
+          assertStringCompare("Empty query", responseBody, responseBody::contains);
+        }
+    );
+  }
+
+  /**
+   * When multiple Content-Type headers are set, the first one (in this case, it's the text format) should be used.
+   */
+  @ParameterizedTest
+  @FieldSource("flags")
```
kfaraz: Rather than have all the tests be parameterized, I wonder if it would be cleaner to:
- write the main test class from point-of-view of Broker.
- then add a separate test class (using junit5 `@Nested` or even a separate class) which does testing form PoV of Router, by extending the first test class.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
```diff
@@ -0,0 +1,379 @@
...
+  public static List<Boolean> flags = List.of(true, false);
+
+  private final EmbeddedBroker broker = new EmbeddedBroker();
+  private final EmbeddedRouter router = new EmbeddedRouter();
+
+  private HttpClient httpClientRef;
+  private String brokerEndpoint;
+  private String routerEndpoint;
+
+  interface OnRequest
```
kfaraz: I think we can remove this now and just use a `Consumer<Request>`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
```diff
@@ -0,0 +1,379 @@
...
+  {
+    void on(Request request);
+  }
+
+  interface OnResponse
```
kfaraz: We can use a `BiConsumer<Integer, String>` instead. What do you think?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
```diff
@@ -0,0 +1,379 @@
...
+  {
+    URL url;
+    try {
+      url = new URL(endpoint);
+    }
+    catch (MalformedURLException e) {
+      throw new AssertionError("Malformed URL");
+    }
+
+    // TODO: instead of pure use of HttpClient, try to use onAnyBroker and create onAnyRouter instead.
```
kfaraz: We can remove this TODO and add the `onAnyRouter` method if needed in the future.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
```diff
@@ -0,0 +1,379 @@
...
+        }
+    );
+  }
+
+  /**
+   * When multiple Content-Type headers are set, the first one (in this case, it's the text format) should be used.
+   */
+  @ParameterizedTest
+  @FieldSource("flags")
+  public void testMultipleContentType(boolean shouldQueryBroker)
```
kfaraz:
```suggestion
  public void testMultipleContentType_usesFirstOne(boolean shouldQueryBroker)
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
```diff
@@ -0,0 +1,379 @@
...
+        (statusCode, responseBody) -> {
+          Assertions.assertEquals(400, statusCode);
+          assertStringCompare("Empty query", responseBody, responseBody::contains);
+        }
+    );
+  }
+
+  /**
+   * When multiple Content-Type headers are set, the first one (in this case, it's the text format) should be used.
+   */
```
kfaraz: we can remove this comment and update the test method name to capture this info.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
```diff
@@ -0,0 +1,379 @@
...
+
+  @BeforeEach
+  void setUp()
+  {
+    httpClientRef = router.bindings().globalHttpClient();
+    brokerEndpoint = String.format(SQL_QUERY_ROUTE, getServerUrl(broker));
+    routerEndpoint = String.format(SQL_QUERY_ROUTE, getServerUrl(router));
+  }
+
+  protected void assertStringCompare(String expected, String actual, Function<String, Boolean> predicate)
```
kfaraz: It seems that this method is always used to check if `responseBody.contains(testString)`. We can remove this method and just inline the call where applicable.

kfaraz (commented):
Thanks for migrating the test, @Fly-Style !
Changes look good.

I have left some minor suggestions for cleanup.
I ran the tests locally and they look good.

processing/src/main/java/org/apache/druid/query/http/ClientSqlQuery.java
Fly-Style: I would do that in separate PR, it would be slightly cleaner.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
Fly-Style: Totally agree.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
Fly-Style: Addressed in https://github.com/apache/druid/pull/18752/commits/f99bf1b71e20610cd7d6aedba2cc7457c4ab9aa2

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
Fly-Style: Addressed in https://github.com/apache/druid/pull/18752/commits/f99bf1b71e20610cd7d6aedba2cc7457c4ab9aa2

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
Fly-Style: Addressed in https://github.com/apache/druid/pull/18752/commits/f99bf1b71e20610cd7d6aedba2cc7457c4ab9aa2

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
Fly-Style: Addressed in https://github.com/apache/druid/pull/18752/commits/f99bf1b71e20610cd7d6aedba2cc7457c4ab9aa2

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
Fly-Style: Addressed in https://github.com/apache/druid/pull/18752/commits/f99bf1b71e20610cd7d6aedba2cc7457c4ab9aa2

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
Fly-Style: Addressed in https://github.com/apache/druid/pull/18752/commits/f99bf1b71e20610cd7d6aedba2cc7457c4ab9aa2

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
Fly-Style: Addressed in https://github.com/apache/druid/pull/18752/commits/f99bf1b71e20610cd7d6aedba2cc7457c4ab9aa2

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ITSqlQueryTest.java
Fly-Style: Addressed in https://github.com/apache/druid/pull/18752/commits/f99bf1b71e20610cd7d6aedba2cc7457c4ab9aa2

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/SqlQueryHttpRequestHeadersTest.java
Fly-Style: Yeap, good idea. Will do in consequent commit

kfaraz (approved):
LGTM! 🚀

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/SqlQueryHttpRequestHeadersTest.java
Fly-Style: UPD: the idea is really good, but the implementation itself became a burden, and when I tried to take a look on it with fresh eyes, the test scenery itself looks less clear.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/SqlQueryHttpRequestHeadersTest.java
```diff
@@ -0,0 +1,349 @@
...
+
+/**
+ * Suite to test various Content-Type headers attached
+ * to SQL query HTTP requests to brokers and routers.
+ */
+public class SqlQueryHttpRequestHeadersTest extends EmbeddedClusterTestBase
+{
+  private static final String SQL_QUERY_ROUTE = "%s/druid/v2/sql/";
+
+  public static List<Boolean> flags = List.of(true, false);
```
kfaraz: Nit: It would probably make sense to rename this to say `useBrokerQueryEndpoint`, so that the intent is clear.

Fly-Style: Fair enough! Will change it.

PR #18755, opened 2025-11-19
Fix off-by-two with task action retry count.

gianm:
The property druid.peon.taskActionClient.retry.maxRetryCount was being decremented prior to being passed to StandardRetryPolicy as the value of maxAttempts. It should actually be incremented: if there are N retries that means N + 1 total attempts.


kfaraz:
IT failure is unrelated and this IT is scheduled for migration anyway. Proceeding with merging this PR.

PR #18756, opened 2025-11-19
Fix computation of task slots required for MSQ compaction task

kfaraz:
### Bug

When running an MSQ compaction task using compaction supervisors on the Overlord, the number of task slots required for each compaction is incorrect as it always assumes the task to be of type native.

This means that the `maxNumTasks` specified in the `taskContext` of the compaction supervisor is ignored, and only the `maxNumConcurrentSubTasks` specified in the compaction supervisor is honored.

### Temporary workaround

Set `maxNumConcurrentSubTasks` to the same value as the desired `maxNumTasks`.

### Fix

- Add method `CompactionSlotManager.computeSlotsRequiredForTask()`
- Use method in compaction template to compute task slots based on engine type

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/client/indexing/ClientCompactionTaskQuery.java
+++ b/server/src/main/java/org/apache/druid/client/indexing/ClientCompactionTaskQuery.java
@@ -20,7 +20,6 @@
 package org.apache.druid.client.indexing;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/druid/client/indexing/ClientCompactionTaskQuery.java
```diff
@@ -161,6 +163,15 @@ public ClientCompactionRunnerInfo getCompactionRunner()
     return compactionRunner;
   }
 
+  /**
+   * @return true if this is an MSQ compaction task.
+   */
+  @JsonIgnore
```
cryptoe: Why is this annotation required ?

kfaraz: Thanks for catching this. This can be removed.

kfaraz:

PR #18751, opened 2025-11-18
Remove JDK 11 from root pom

jtuglu1:
### Description

V35 removed support for JDK 11. This ensures using proper maven compiler version.

#### Release note

Switch to using JDK 17 Maven compiler version

<hr>

This PR has:

- [ ] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [ ] been tested in a test Druid cluster.

PR #18759, opened 2025-11-19
Revert "Switch maven compiler version to 17 (#18751)"

cryptoe:
This reverts commit cf8f9958c573f929495f12e3032942b6d90a7e2b.
as this breaks hadoop ingest.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,7 @@
     </scm>
 
     <properties>
-        <java.version>17</java.version>
+        <java.version>11</java.version>
         <maven.compiler.release>${java.version}</maven.compiler.release>
         <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
         <aether.version>0.9.0.M2</aether.version>


jtuglu1:
Odd. All UT/IT tests passed. Is there a way we can somehow conditionally enable this so people who don't use hadoop ingests get the benefit of a newer compiler? Doesn't seem ideal to be stuck on JDK 11 compiler for a feature that has been deprecated. This could be done potentially with the hadoop maven profiles.

clintropolis:
>Odd. All UT/IT tests passed. Is there a way we can somehow conditionally enable this so people who don't use hadoop ingests get the benefit of a newer compiler? Doesn't seem ideal to be stuck on JDK 11 compiler for a feature that has been deprecated. This could be done potentially with the hadoop maven profiles.

If i understand things correctly, I think the issue is that it would allow using language level 17 features in Druid code which is what would actually break hadoop ingestion. We don't officially support java 11 anymore, but hadoop ingestion still works because none of the parts that run there need java 17 only stuff yet.

jtuglu1:
> > Odd. All UT/IT tests passed. Is there a way we can somehow conditionally enable this so people who don't use hadoop ingests get the benefit of a newer compiler? Doesn't seem ideal to be stuck on JDK 11 compiler for a feature that has been deprecated. This could be done potentially with the hadoop maven profiles.
> 
> If i understand things correctly, I think the issue is that it would allow using language level 17 features in Druid code which is what would actually break hadoop ingestion. We don't officially support java 11 anymore, but hadoop ingestion still works because none of the parts that run there need java 17 only stuff yet.

Is there a way to just detect to use JDK 11 compiler when a Hadoop profile is being used? This way the project doesn't stagnate and isn't pinned down by deprecated features (for example, enabling virtual thread or other JDK 17+ features). Virtual thread is 21 but same point applies I guess.

clintropolis:
>Is there a way to just detect to use JDK 11 compiler when a Hadoop profile is being used? This way the project doesn't stagnate and isn't pinned down by deprecated features (for example, enabling virtual thread or other JDK 17+ features). Virtual thread is 21 but same point applies I guess.

I think we are just doing this artificial language level restriction until Druid 37 which iirc is the current target for removing hadoop support from the dev list threads.

There is such a thing as multi-release jars which allow using new language features but still supporting older versions, but this are kind of tedious and only very very recently even supported by intellij (and I haven't had a chance to play around much with that support so unsure how usable it is yet.. also not sure about other ide support), so should be reserved for specific features where the benefits outweigh the cost in complexity I think.

PR #18760, opened 2025-11-19
compaction now only identifies multi-value dimensions for dimension schemas which can possibly produce them

clintropolis:
### Description
changes:
* added `DimensionSchema.canBeMultiValued()`, defaulting to false but true for `StringDimensionSchema` and `NewSpatialDimensionSchema`
* `CompactionTask.identifyMultiValuedDimensions` now checks `DimensionSchema.canBeMultiValued()` instead of checking for `ColumnType.STRING` from `DimensionSchema.getColumnType()`
* added test to confirm behavior using `AutoTypeColumnSchema` with a specified `castToType` of `ColumnType.STRING`, which prior to changes would identify as multi-valued despite auto never producing them

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
@@ -467,7 +467,7 @@ public class CompactionTask extends AbstractBatchIndexTask implements PendingSeg
    * <ul>
    * <li> Range partitioning is done on a possibly multi-valued string dimension or an unknown dimension
    * (since MSQ does not support partitioning on a multi-valued string dimension) </li>
-   * <li> Rollup is done on a string dimension or an unknown dimension
+   * <li> Rollup is done on a multi-valued string dimension or an unknown dimension
    * (since MSQ requires multi-valued string dimensions to be converted to arrays for rollup) </li>
    * </ul>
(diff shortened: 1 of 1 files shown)


processing/src/main/java/org/apache/druid/data/input/impl/DimensionSchema.java
```diff
@@ -176,6 +176,12 @@ public boolean hasBitmapIndex()
   @JsonIgnore
   public abstract ColumnType getColumnType();
 
+  @JsonIgnore
+  public boolean canBeMultiValued()
```
gianm, line 185: Javadoc please. It always helps even for things that may be obvious to certain eyes.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -466,13 +465,13 @@ public boolean isPerfectRollup()
    * Checks if multi-valued string dimensions need to be analyzed by downloading the segments.
    * This method returns true only for MSQ engine when either of the following holds true:
    * <ul>
-   * <li> Range partitioning is done on a string dimension or an unknown dimension
+   * <li> Range partitioning is done on a possibly multi-valued string dimension or an unknown dimension
    * (since MSQ does not support partitioning on a multi-valued string dimension) </li>
    * <li> Rollup is done on a string dimension or an unknown dimension
```
```suggestion
   * <li> Rollup is done on a possibly multi-valued string dimension or an unknown dimension
```

kfaraz:
IT failure is unrelated.

PR #18762, opened 2025-11-20
Common pattern validation for all regexp* functions

abhishekrb19:
Currently, if a user runs `regexp_like(c2, '[abc-d-12]')`, they receive a cryptic error message like "Illegal character range near index ...". When the SQL query is quite complex, this message becomes even harder to understand and debug.

Move the `RegexpExtractExprMacro.compilePattern()` utility into a shared location and have all Regexp* macros use it, so that invalid regex patterns produce nicer error messages for users.

The error would now look something like this:
```
An invalid pattern [[default-byom-gp]]] was provided for the regexp_like function, error: [Illegal character range near index 9 [default-byom-gp]] ^]
```

#### Release note

Nicer user-facing error messages for invalid patterns used in the regexp* functions.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/sql/src/test/java/org/apache/druid/sql/calcite/CalciteQueryTest.java
+++ b/sql/src/test/java/org/apache/druid/sql/calcite/CalciteQueryTest.java
@@ -8411,7 +8411,7 @@ public class CalciteQueryTest extends BaseCalciteQueryTest
         + "  REGEXP_EXTRACT(dim1, '^(.))', 1)\n"
         + "FROM foo",
         DruidExceptionMatcher.invalidInput().expectMessageContains(
-            "An invalid pattern [^(.))] was provided for the regexp_extract function, " +
+            "An invalid pattern [^(.))] was provided for the [regexp_extract] function, " +
             "error: [Unmatched closing ')' near index 3\n^(.))\n   ^]"
         )
     );


processing/src/main/java/org/apache/druid/query/expression/RegexpExprUtils.java
```diff
@@ -0,0 +1,51 @@
...
+  public static Pattern compilePattern(@Nullable String pattern, String functionName)
+  {
+    try {
+      return Pattern.compile(StringUtils.nullToEmptyNonDruidDataString(pattern));
+    }
+    catch (PatternSyntaxException e) {
+      throw InvalidInput.exception(
+          e,
+          StringUtils.format(
+              "An invalid pattern [%s] was provided for the %s function, error: [%s]",
```
jtuglu1: maybe box the function name as well?

processing/src/main/java/org/apache/druid/query/expression/RegexpReplaceExprMacro.java
```diff
@@ -96,7 +96,7 @@ private RegexpReplaceExpr(List<Expr> args)
       final String patternString = (String) patternExpr.getLiteralValue();
 
       this.arg = args.get(0);
-      this.pattern = patternString != null ? Pattern.compile(patternString) : null;
```
jtuglu1, line 99: We are moving from `Pattern.compile(patternString)` to `Pattern.compile(StringUtils.nullToEmptyNonDruidDataString(patternString))`. Is this ok?

abhishekrb19: Yeah, this is okay because the original null semantics is retained at this call site. 

`RegexpExprUtils.compilePattern()` is called only when `patternString` is not null, otherwise `pattern` continues to remain null

PR #13506, opened 2022-12-06
Sort-merge join and hash shuffles for MSQ.

gianm:
Adds a distributed sort-merge join algorithm for MSQ.

The main changes are in the processing, multi-stage-query, and sql modules.

processing module:

1) Rename SortColumn to KeyColumn, replace boolean descending with KeyOrder.
   This makes it nicer to model hash keys, which use KeyOrder.NONE.

2) Add nullability checkers to the FieldReader interface, and an
   "isPartiallyNullKey" method to FrameComparisonWidget. The join
   processor uses this to detect null keys.

3) Add WritableFrameChannel.isClosed and OutputChannel.isReadableChannelReady
   so callers can tell which OutputChannels are ready for reading and which
   aren't.

4) Specialize FrameProcessors.makeCursor to return FrameCursor, a random-access
   implementation. The join processor uses this to rewind when it needs to
   replay a set of rows with a particular key.

5) Add MemoryAllocatorFactory, which is embedded inside FrameWriterFactory
   instead of a particular MemoryAllocator. This allows FrameWriterFactory
   to be shared in more scenarios.

multi-stage-query module:

1) ShuffleSpec: Add hash-based shuffles. New enum ShuffleKind helps callers
   figure out what kind of shuffle is happening. The change from SortColumn
   to KeyColumn allows ClusterBy to be used for both hash-based and sort-based
   shuffling.

2) WorkerImpl: Add ability to handle hash-based shuffles. Refactor the logic
   to be more readable by moving the work-order-running code to the inner
   class RunWorkOrder, and the shuffle-pipeline-building code to the inner
   class ShufflePipelineBuilder.

3) Add SortMergeJoinFrameProcessor and factory.

4) WorkerMemoryParameters: Adjust logic to reserve space for output frames
   for hash partitioning. (We need one frame per partition.)

sql module:

1) Add sqlJoinAlgorithm context parameter; can be "broadcast" or
   "sortMerge". With native, it must always be "broadcast", or it's a
   validation error. MSQ supports both. Default is "broadcast" in
   both engines.

2) Validate that MSQs do not use broadcast join with RIGHT or FULL join,
   as results are not correct for broadcast join with those types. Allow
   this in native for two reasons: legacy (the docs caution against it,
   but it's always been allowed), and the fact that it actually *does*
   generate correct results in native when the join is processed on the
   Broker. It is much less likely that MSQ will plan in such a way that
   generates correct results.

3) Remove subquery penalty in DruidJoinQueryRel when using sort-merge
[...]

PR #18765, opened 2025-11-20
MSQ: Use full parallelism in localSort.

gianm:
Since localSort was introduced for sort-merge join in #13506, it has used 1 processor at a time with a max of 2 channels per processor. This is inefficient, because at the time the sorter runs, it is the only sorter running. The patch contains a comment describing the situation in more detail.

This patch has two benefits. First, the sorter should run faster if multiple processing threads are available. Second, the sorter, due to having a larger max channels per merger, will make more efficient use of intermediate channels.

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardShuffleOperations.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardShuffleOperations.java
@@ -364,6 +364,7 @@ public class StandardShuffleOperations
             nextFuture = Futures.transformAsync(
                 nextFuture,
                 ignored -> {
+                  final WorkerMemoryParameters memoryParameters = executionContext.frameContext().memoryParameters();
                   final SuperSorter sorter = new SuperSorter(
                       Collections.singletonList(channel.getReadableChannel()),
                       stageDefinition.getFrameReader(),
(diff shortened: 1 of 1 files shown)


multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/StandardShuffleOperations.java
```diff
@@ -382,8 +383,18 @@ public <T2> FrameProcessor<T2> decorate(FrameProcessor<T2> processor)
...
+                      //
+                      // Typically, nothing else is running at all. Whenever there is more than one output partition,
+                      // the step prior to this localSort (typically hashPartition) should use buffered output channels
+                      // and therefore would have exited.
+                      //
+                      // In the case where there is one output partition, the step prior to this localSort may run
+                      // concurrently with the sorter, but in that case it will only have one output channel so won't
+                      // have many frames buffered.
+                      memoryParameters.getSuperSorterConcurrentProcessors(),
+                      memoryParameters.getSuperSorterMaxChannelsPerMerger(),
```
cryptoe, line 397: There is an implicit assumption that getChannelsPerMerger >=2. 
I am not sure if that assumption is honored in the memory parameters anywhere.

gianm: Yes, it'll always be at least 2.

PR #18767, opened 2025-11-21
Don't kill tasks while a supervisor is stopping.

gianm:
Previously, if a supervisor was stopped while discovering tasks or updating their status, it could end up trying to kill those tasks because the callbacks on the requests to those tasks could fail as a result of the stopping. There should be no reason for an actively-stopping supervisor to kill a task, so this patch causes the shutdown functions to be no-ops while a supervisor is stopping.

This patch also ensures that when a supervisor transitions into the STOPPING state, that state takes priority over any other state.

### DIFF
--- a/server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorStateManager.java
+++ b/server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorStateManager.java
@@ -129,7 +129,7 @@ public class SupervisorStateManager
   {
     if (BasicState.STOPPING.equals(this.supervisorState) || BasicState.STOPPING.equals(proposedState)) {
       // STOPPING takes precedence over all other states
-      supervisorState = proposedState;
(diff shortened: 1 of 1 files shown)


gianm:
The specific scenario I saw this in had an error like:

```
Failed to return start time: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@703f49ef[Not completed, task = java.util.concurrent.Executors$RunnableAdapter@7219380a[Wrapped task = org.apache.druid.indexing.seekablestream.supervisor.SeekableStreamSupervisor$$Lambda$1573/0x000000f801db8228@6ff15c29]] rejected from java.util.concurrent.ScheduledThreadPoolExecutor@1eb21b5c[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 1]
```

It happened because the callback on `getStartTimeAsync` tried to submit a task to a `scheduledExec` that had already been stopped.

capistrant (approved):

Issue #18764, opened 2025-11-20
Segment Rebalance Race


jtuglu1:
I've identified a race in segment movement which causes temporary segment unavailability (and thus partial query results).

Scenario: Segment A is being moved from server A to server B with 1x replication (other replication factors may also be affected).

The race is as follows:

1. (Coordinator): Running dutyGroup[HistoricalManagementDuties].
2. (Coordinator): Marking segment[S] to drop on server[A] to load on server[B].
3. (Server B): Loading segment[S].
4. (Coordinator): Finished running dutyGroup[HistoricalManagementDuties].
5. (Coordinator): Running dutyGroup[HistoricalManagementDuties].
6. (Coordinator): Dropping segment [S] from server [A].
7. (Server A): Completely removing segment[S].
8. (Coordinator): Dropping segment [S] from server [B]. (RACE)

The issue is that there is a momentary lapse in time between when server B loads the segment (and it [appears](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCountMap.java#L54) as loaded in accounting) and when the callback for dropping on server A occurs. 

Because the `projectedReplicas` (calculated [here](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java#L266)) is: `loadedNotDropping` + `loading` - `max(0, moveFrom - moveTo)` = 2 (both A/B are loaded at this point) + 0 - max(0, 1 - 1) = 2, this causes the drop to be subsequently scheduled on B.

### Affected Version

All recent Druid versions.

kfaraz:
@jtuglu1 , thanks for reporting this!

IIUC, there are two parts to the problem.

### A

> The issue is that there is a momentary lapse in time between when server B loads the segment (and it [appears](https://github.com/apache/druid/blob/-/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCountMap.java#L54) as loaded in accounting) and when the callback for dropping on server A occurs.

I think this may be true but more from the Broker's perspective i.e. it is possible that the Broker sees the drop from A and then the load on B, causing the segment to be unavailable for a brief period. Since the Coordinator is not aware of the inventory of the individual Brokers, there isn't much we can do here.

Generally, the Coordinator's own inventory is meant to serve as a proxy for what the Brokers see.
[...]

PR #18769, opened 2025-11-21
[WIP]: Fix segment rebalance race

jtuglu1:
Fixes [#18764.](https://github.com/apache/druid/issues/18764)

#### Race

Segment load/drop callbacks are racing with [prepareCurrentServers](https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/server/coordinator/duty/PrepareBalancerAndLoadQueues.java#L139-L147). 

Consider the following scenario: Coordinator is moving segment `S` from server `A` to server `B`. `S` has 1x replication.

Basically, the load/drop callbacks can happen between the start/end of this function in a way where you get a 

```
SegmentReplicaCount{requiredAndLoadable=1, required=1, loaded=2, loadedNonHistorical=0, loading=0, dropping=0, movingTo=0, movingFrom=0}. 
```

Essentially:
```
T0[Coordinator]: enter prepareCurrentServers()
T1[Coordinator]: Server[B] completed request[MOVE_TO] on segment[S] with status[SUCCESS]
T2[Coordinator]: Dropping segment [S] from server[A]
T3[A]: Completely removing segment[S] in [30,000]ms.
T4[Coordinator]: Server[A] completed request[DROP] on segment[S] with status[SUCCESS].
T5[Coordinator]: exit prepareCurrentServers()
T6[Coordinator]: enter prepareCluster()
T7[Coordinator]: exit prepareCluster()
T8[Coordinator]: enter initReplicaCounts()
T9[Coordinator]: exit initReplicaCounts()
T10[Coordinator]: Segment S replica count is SegmentReplicaCount{requiredAndLoadable=1, required=1, loaded=2, loadedNonHistorical=0, loading=0, dropping=0, movingTo=0, movingFrom=0}
T11[Coordinator]: Dropping segment [S] from server[B]
```

I think what's happening is that the server loop in `prepareCurrentServers()` reads the servers in a state where LOAD has persisted in the server view (2x loaded) but the DROP has not materialized yet in the view. This causes `loaded=2`. Then, I thought the in-flight DROP (since it hasn't materialized in the view) would get picked up in the `queuedSegments` load queue peons (and show up as `dropping=1`), but I think since the DROP callback returns – and clears the entry from the old `queuedSegments` load queue peons – before `prepareCluster()` has a chance to copy over the queued action to new `queuedSegments`, we lose that important bit of information. Hence, you are left in a weird state with a "valid" queue but an invalid load state. In other words, I think we need to somehow synchronize callbacks with this `prepareCurrentServers()` and `prepareCluster()`.

#### Fix

I don't really like the idea of passing a lock through to child peons, but a solution here is to synchronize the callback executions with the `PrepareBalancerAndLoadQueues::run()`. This way, callbacks on `HttpLoadQueuePeon` can run concurrently w.r.t each other, but during the `PrepareBalancerAndLoadQueues::run()` step, callbacks must wait (and vice-versa). While this could mean "stale" accounting snapshots compared to the actual state of the cluster, this ensures *valid* accounting snapshots which means that we don't do anything crazy like drop a segment from the timeline during a rebalance (assuming the rest of the rebalancing logic is correct).

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/coordinator/duty/PrepareBalancerAndLoadQueues.java
+++ b/server/src/main/java/org/apache/druid/server/coordinator/duty/PrepareBalancerAndLoadQueues.java
@@ -81,34 +81,42 @@ public class PrepareBalancerAndLoadQueues implements CoordinatorDuty
   @Override
   public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
   {
-    List<ImmutableDruidServer> currentServers = prepareCurrentServers();
-    taskMaster.resetPeonsForNewServers(currentServers);
+    // Prevent callbacks from firing while balancer/load queue accounting is taking place.
+    // This ensures snapshot validity at the expense of potentially stale snapshots (in some cases).
+    taskMaster.getCallbackLock().writeLock().lock();
+    try {
+      List<ImmutableDruidServer> currentServers = prepareCurrentServers();
+      taskMaster.resetPeonsForNewServers(currentServers);
 
-    final CoordinatorDynamicConfig dynamicConfig = params.getCoordinatorDynamicConfig();
-    final SegmentLoadingConfig segmentLoadingConfig
-        = SegmentLoadingConfig.create(dynamicConfig, params.getUsedSegmentCount());
+      final CoordinatorDynamicConfig dynamicConfig = params.getCoordinatorDynamicConfig();
+      final SegmentLoadingConfig segmentLoadingConfig
+          = SegmentLoadingConfig.create(dynamicConfig, params.getUsedSegmentCount());
 
-    final DruidCluster cluster = prepareCluster(dynamicConfig, segmentLoadingConfig, currentServers);
-    cancelLoadsOnDecommissioningServers(cluster);
+      final DruidCluster cluster = prepareCluster(dynamicConfig, segmentLoadingConfig, currentServers);
+      cancelLoadsOnDecommissioningServers(cluster);
 
-    final CoordinatorRunStats stats = params.getCoordinatorStats();
-    collectHistoricalStats(cluster, stats);
-    collectUsedSegmentStats(params, stats);
-    collectDebugStats(segmentLoadingConfig, stats);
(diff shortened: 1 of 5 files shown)


kfaraz:
Thanks for adding the test, @jtuglu1 ! I will take a look soon.

server/src/test/java/org/apache/druid/server/coordinator/simulate/SegmentBalancingTest.java
```diff
@@ -308,4 +308,47 @@ public void testMaxSegmentsAreMovedWhenClusterIsSkewed()
...
+
+    // Physically load segments on B, but don't fire callbacks yet
+    loadQueuedSegmentsSkipCallbacks();
+    Assert.assertEquals(10, historicalT11.getTotalSegments());
+    Assert.assertEquals(5, historicalT12.getTotalSegments());
+
+    // Fire callbacks asynchronously BEFORE next coordinator run
+    // This simulates the race: callbacks fire after first build() but state is stale
+    firePendingLoadCallbacks();
+    runCoordinatorCycle();
```
kfaraz: Is this able to simulate the race condition correctly?

IIUC, as we discussed offline, the race condition gets triggered if the callback is invoked after the `HistoricalManagementDuties` group starts (so that snapshot is already taken) and before the `RunRules` duty of that group has started (so that `RunRules` ends up working on a stale snapshot).

jtuglu1: It was able to simulate, but not 100% reliably. I'll push a better test soon.

jtuglu1:
@gianm – thoughts on the approach here? One pending change is I'll need to encompass [these](https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/server/coordinator/loading/HttpLoadQueuePeon.java#L406-L407) updates to `queuedSegments` within the critical section as well to ensure an atomic snapshot of both server inventory and load queue peons.

PR #18772, opened 2025-11-21
Double check that the TaskQueue is still active before managing the queued tasks.

capistrant:
#### Fixed a bug in TaskQueue `stop()` process

`TaskQueue#stop()` starts clearing out the task queue and calling shutdownNow() on all of the executors for the `TaskQueue`. I am suspicious that I experienced an instance where `stop()` got called and before the `TaskQueue-Manager` thread was interrupted it marked all the tasks it thought existed for deletion because they were no longer in the task queue. To protect against this, A simple check to see if the `TaskQueue` is active before managing the queued tasks should prevent any issue like this.

#### Release note

Fix a bug where an overlord that is giving up leadership may erroneously kill indexing tasks

<hr>

 * `TaskQueue`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/TaskQueue.java
@@ -379,7 +379,9 @@ public class TaskQueue
   {
     startStopLock.readLock().lock();
     try {
-      startPendingTasksOnRunner();
+      if (isActive()) {
+        startPendingTasksOnRunner();
+      }
     }
     finally {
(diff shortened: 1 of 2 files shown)


capistrant:
testing a theory that once some of these ITs fail that shouldn't they will never succeed if you just retry but if you re do the whole thing they can pass. I don't know what kind of state would hold over from run to run (if any). but I'm curious  if close open results in me being able to get that IT green which I don't think had any business failing based on my change (plus it was like the cluster won't even start failure from what I can tell). It is also failing in #18771 and not passing on retries. unless master recently broke it, idk.

PR #18774, opened 2025-11-23
Bulk add to timeline in DataSourceCompactibleSegmentIterator.

gianm:
Calling "addAll" is more efficient than calling "add" in a loop, it is O(N) instead of O(N^2).

### DIFF
--- a/processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
+++ b/processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
@@ -176,14 +176,20 @@ public class VersionedIntervalTimeline<VersionType, ObjectType extends Overshado
         .toSet();
   }
 
+  /**
+   * Add a single partition chunk entry to this timeline. Avoid calling this in a loop, since it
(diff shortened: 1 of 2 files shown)


processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
```diff
@@ -176,14 +176,20 @@ public Set<ObjectType> findNonOvershadowedObjectsInInterval(Interval interval, P
         .toSet();
   }
 
+  /**
+   * Add a single partition chunk entry to this timeline. Avoid calling this in a loop, since it
+   * is O(objects in holder) and can therefore creates O(N^2) situations. Instead use {@link #addAll(Iterator)}
```
kfaraz, line 181: May be also link to `PartitionHolder.isComplete()` in this javadoc since that is really the RDS.

gianm: Follow-up here: https://github.com/apache/druid/pull/18775

PR #18777, opened 2025-11-24
Allow large partition numbers in VersionedIntervalTimeline.

gianm:
The experimental segment locking feature (`forceTimeChunkLock: false`) reserves the first 32768 partition numbers for the "root generation", and then uses an "atomic update groups" scheme to replace root segment ranges with new sets of segments during reindexing operations.

OvershadowableManager, which manages this atomic update scheme, imposes a limit of 32768 segments per time chunk. This applies even to people that are not using segment locking. In this patch, the class is now only used when segment locking is actually in play, meaning that the limit is not imposed under normal conditions.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/HighPartitionNumberTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/HighPartitionNumberTest.java
@@ -131,7 +131,7 @@ public class HighPartitionNumberTest extends EmbeddedClusterTestBase
 
   /**
    * Reinserts the segment from {@link #insertFirstSegment()} directly into metadata storage,
-   * with partition number 32767.
+   * with partition number 32766.
    */
   private void insertSecondSegment()
   {


processing/src/main/java/org/apache/druid/timeline/partition/PartitionHolder.java
```diff
@@ -41,42 +45,44 @@ public static <T extends Overshadowable<T>> PartitionHolder<T> copyWithOnlyVisib
...
+    this.contents = contents;
     this.maxMinorVersion = maxMinorVersion;
   }
 
   public boolean add(PartitionChunk<T> chunk)
   {
-    boolean added = overshadowableManager.addChunk(chunk);
+    if (chunk.getObject().getMinorVersion() != 0 && contents instanceof SimplePartitionHolderContents) {
+      // Swap simple map for an OvershadowableManager when minor versions are encountered.
+      contents = OvershadowableManager.fromSimple((SimplePartitionHolderContents<T>) contents);
```
kfaraz, line 87: Since this swap will bring about changes in the segment-holding capacity of the interval, should we add a log line here?

processing/src/main/java/org/apache/druid/timeline/partition/OvershadowableManager.java
```diff
@@ -65,7 +65,7 @@
  *
  * This class is not thread-safe.
  */
-class OvershadowableManager<T extends Overshadowable<T>>
+class OvershadowableManager<T extends Overshadowable<T>> implements PartitionHolderContents<T>
```
kfaraz: Maybe rename this to indicate that it is used only when segment locking is used.

processing/src/main/java/org/apache/druid/timeline/partition/PartitionHolderContents.java
```diff
@@ -0,0 +1,87 @@
...
+   * Whether this holder is empty.
+   */
+  boolean isEmpty();
+
+  /**
+   * Whether all visible chunks are consistent, meaning they can possibly be considered for
+   * {@link PartitionHolder#isComplete()}. When segment locking is not being used, all chunks
+   * are consistent, so this always returns true.
+   */
+  boolean visibleChunksAreConsistent();
```
kfaraz: Nit:
```suggestion
  boolean areVisibleChunksConsistent();
```

processing/src/main/java/org/apache/druid/timeline/partition/PartitionHolderContents.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Contents for {@link PartitionHolder}.
+ *
+ * @see SimplePartitionHolderContents implementation when segment locking is not in play
+ * @see OvershadowableManager implementation when segment locking is in play
+ */
+public interface PartitionHolderContents<T>
```
kfaraz, line 32: Super Nit: "Holder" + "Contents" sound like they should cancel each other out. 😛 

Should we call this class `PartitionChunks<T>` or `PartitionChunkGroup<T>` since it is basically just a group of chunks?

The implementations could be called `SimplePartitionChunkGroup` and `PartitionChunkGroupWithMinorVersion` or something.

processing/src/main/java/org/apache/druid/timeline/partition/SimplePartitionHolderContents.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Contents for {@link PartitionHolder} when segment locking was not used, and therefore no chunks have
+ * {@link Overshadowable#getMinorVersion()}.
+ */
+public class SimplePartitionHolderContents<T extends Overshadowable<T>> implements PartitionHolderContents<T>
+{
+  private final TreeMap<PartitionChunk<T>, PartitionChunk<T>> holderMap = new TreeMap<>();
+
+  /**
+   * Map of {@link PartitionChunk#getChunkNumber()} to
```
kfaraz:
```suggestion
   * Map from {@link PartitionChunk#getChunkNumber()} to the actual {@link PartitionChunk}.
```

processing/src/main/java/org/apache/druid/timeline/partition/OvershadowableManager.java
```diff
@@ -105,10 +105,23 @@ enum State
     this.overshadowedGroups = new TreeMap<>();
   }
 
-  public static <T extends Overshadowable<T>> OvershadowableManager<T> copyVisible(OvershadowableManager<T> original)
+  public static <T extends Overshadowable<T>> OvershadowableManager<T> fromSimple(
```
kfaraz, line 112: Should this method fail if the passed simple holder already has more than 32k chunks?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/config/TaskLockConfig.java
```diff
@@ -42,6 +43,13 @@ public class TaskLockConfig
   @JsonProperty
   private int batchAllocationNumThreads = 5;
 
+  /**
+   * Undocumented parameter, used for embedded tests that create high partition numbers without needing to
+   * individually create each segment. Controls the initial partition number to be used for segment allocation.
+   */
+  @JsonProperty
+  private int initialAllocationPartitionNumber = PartitionIds.ROOT_GEN_START_PARTITION_ID;
```
kfaraz: If this config is to be used only for tests, I would advise a slightly different strategy which wouldn't require touching all the production classes.

1. Use the `ClusterTestingModule` extension in the embedded test and enable it on the Overlord via `druid.unsafe.cluster.testing=true`.
2. Create a new class `FaultyIndexerSQLMetadataStorageCoordinator extends IndexerSQLMetadataStorageCoordinator` and override allocation methods.
3. In `ClusterTestingModule`, do `binder.bind(IndexerSQLMetadataStorageCoordinator.class).to(FaultyIndexer...);`
4. Pass a test config into the faulty implementation to control the `startPartitionId`.

gianm: I just pushed a different approach:

1. First insert the test data into a single segment using SQL, with partition 0.
2. Direct insert a new segment into metadata storage that is the same as that first one, but with a partition number of Short.MAX_VALUE
3. Run SQL to insert 3 more segments.
4. Verify all are loaded and queryable.

gianm: I removed this new parameter I had added.

processing/src/main/java/org/apache/druid/timeline/partition/OvershadowableManager.java
gianm: I just added a check to `addChunk`, so now it will.

processing/src/main/java/org/apache/druid/timeline/partition/PartitionHolderContents.java
gianm: "Group" unfortunately means something different in OvershadowableManager (an atomic update group).

processing/src/main/java/org/apache/druid/timeline/partition/PartitionHolder.java
gianm: I think it'll get logged too much, since timelines are created all the time.

processing/src/main/java/org/apache/druid/timeline/partition/OvershadowableManager.java
gianm: Hmm, rather than rename I added a javadoc.

gianm:
@kfaraz just updated this— TY for review.

processing/src/main/java/org/apache/druid/timeline/partition/PartitionHolder.java
kfaraz: No, I meant log something only when we change from simple to minor-version enabled. Would that be verbose too?

gianm: I think so, since timelines are created and populated as part of various operations. It would get logged semi-randomly and perhaps too frequently.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/HighPartitionNumberTest.java
```diff
@@ -0,0 +1,182 @@
...
+            overlord.bindings()
+                    .segmentsMetadataStorage()
+                    .retrieveAllUsedSegments(dataSource, Segments.ONLY_VISIBLE)
+        );
+
+    overlord.bindings().segmentsMetadataStorage().commitSegments(
+        Set.of(
+            firstSegment.withShardSpec(
+                new NumberedShardSpec(
+                    Short.MAX_VALUE - 1,
```
kfaraz, line 150: Nice trick!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/HighPartitionNumberTest.java
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  @Test
+  public void testHighPartitionNumbers()
+  {
+    insertFirstSegment();
+    insertSecondSegment();
+    insertLastSegments();
+
+
```
kfaraz: Nit: extra newline?

kfaraz (approved):

PR #18782, opened 2025-11-26
fix leaky segment reference on segment drop

clintropolis:
### Description
Fixes an unfortunate bug where we leak segment references on dropping segments, preventing them from actually being dropped  - the files are deleted on disk, but the buffer is not unmapped since it is tied to reference counter, so the fd is held open until the process restarts. Affects version 35, so probably need to do a patch release.

### DIFF
--- a/server/src/test/java/org/apache/druid/server/SegmentManagerTest.java
+++ b/server/src/test/java/org/apache/druid/server/SegmentManagerTest.java
@@ -90,7 +90,7 @@ public class SegmentManagerTest extends InitializedNullHandlingTest
   );
 
   private ExecutorService executor;
-  private SegmentLocalCacheManager  cacheManager;
+  private SegmentLocalCacheManager cacheManager;
   private SegmentManager segmentManager;
   private SegmentLocalCacheManager virtualCacheManager;
   private SegmentManager virtualSegmentManager;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/druid/server/SegmentManager.java
```diff
@@ -380,11 +380,11 @@ public void dropSegment(final DataSegment dataSegment)
               try (final Closer closer = Closer.create()) {
                 final Optional<Segment> oldSegment = cacheManager.acquireCachedSegment(oldSegmentRef);
                 long numberOfRows = oldSegment.map(segment -> {
+                  closer.register(segment);
```
cecemei, line 383: minor nit: the `oldSegment.map` here seems a bit unnecessary, could be more readable if use simple if `oldSegment.isPresent()`

processing/src/test/java/org/apache/druid/segment/TestSegmentUtils.java
```diff
@@ -126,6 +126,12 @@ public List<OrderBy> getOrdering()
     {
       return Cursors.ascendingTimeOrder();
     }
+
+    @Override
+    public int getNumRows()
+    {
+      return 1234;
+    }
```
cecemei, line 134: This seems unused?

cecemei (approved):

gianm (approved):

server/src/main/java/org/apache/druid/server/SegmentManager.java
gianm: To me the structure seems reasonable, since for regular segments it is doing `return countInspector.getNumRows()` and the return val `numberOfRows` ends up going to the `dataSourceState.removeSegment` call.

PR #18786, opened 2025-11-28
Move error and retry query integration tests into embedded suite

Fly-Style:
### Description

This patch continues the effort to move query group tests to the embedded test suite, currently, we have  migrated `ITQueryErrorTest` and `ITQueryRetryTestOnMissingSegments`
As part of this effort, we also created a base class for the convenience of the next usages. 

Note #1: the queries used in tests are not equal to the previous version, mostly because they sense just don't matter, the key thing is the errors it is throwing.

Note #2: TLS-related tests require more effort to support, so we decided to implement them in subsequent patches.

 * `ITQueryErrorTest`
 * `QueryErrorTest`
 * `ITQueryRetryTestOnMissingSegments`
 * `QueryRetryOnMissingSegmentTest`
 * `QueryTestBase`

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/integration-tests-ex/tools/src/main/java/org/apache/druid/testing/tools/QueryRetryTestCommandCreator.java
+++ b/integration-tests-ex/tools/src/main/java/org/apache/druid/testing/tools/QueryRetryTestCommandCreator.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.testing.tools;
+
+import com.github.rvesse.airline.builder.CliBuilder;
+import org.apache.druid.cli.CliCommandCreator;
+
+public class QueryRetryTestCommandCreator implements CliCommandCreator
+{
+  @SuppressWarnings({ "rawtypes", "unchecked" })
+  @Override
+  public void addCommands(CliBuilder builder)
+  {
+    builder.withGroup("server").withCommands(CliHistoricalForQueryErrorTest.class);
+  }
+}


integration-tests/src/test/java/org/apache/druid/tests/query/ITQueryRetryTestOnMissingSegments.java
```diff
@@ -52,12 +51,11 @@
...
- * all of them always. Instead, it can report missing segments for some
- * segments. See {@link ServerManagerForQueryErrorTest} for more details.
+ * all of them always. Instead, it can report missing segments for some segments.
  * <p>
  * To run this test properly, the test group must be specified as {@link TestNGGroup#QUERY_RETRY}.
  */
-@Test(groups = TestNGGroup.QUERY_RETRY)
+@Test(groups = TestNGGroup.QUERY_RETRY, enabled = false) // TODO: Temporarily disabled due to migration to embedded suite.
 @Guice(moduleFactory = DruidTestModuleFactory.class)
 public class ITQueryRetryTestOnMissingSegments
```
kfaraz: @Fly-Style , is this test not being migrated in this PR?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
```diff
@@ -0,0 +1,282 @@
...
+                               .addExtension(ServerManagerForQueryErrorTestModule.class);
+  }
+
+  @Override
+  protected void beforeAll()
+  {
+    var ingestionStatus = cluster.callApi().onAnyBroker(
+        b -> b.submitSqlTask(
+            new ClientSqlQuery(
+                "REPLACE INTO t\n"
```
kfaraz: Let's use a better datasource name. You can try to use `EmbeddedClusterApis.createTestDatasourceName()`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
```diff
@@ -0,0 +1,282 @@
...
+                               .addServer(router)
+                               .addServer(indexer)
+                               .addServer(historical)
+                               .addExtension(ServerManagerForQueryErrorTestModule.class);
+  }
+
+  @Override
+  protected void beforeAll()
+  {
+    var ingestionStatus = cluster.callApi().onAnyBroker(
```
kfaraz: Let's use concrete types instead of `var`. I think Druid has mostly stuck to the convention of concrete types so far in the code.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
```diff
@@ -0,0 +1,282 @@
...
+                false,
+                false,
+                Map.of(),
+                List.of()
+            )
+        )
+    );
+    cluster.callApi().waitForTaskToSucceed(ingestionStatus.getTaskId(), overlord);
+    try {
+      Thread.sleep(1000L);
```
kfaraz: Why is this sleep needed?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
```diff
@@ -0,0 +1,282 @@
...
+    );
+  }
+
+  @Test
+  public void testSqlValidationException()
+  {
+    MatcherAssert.assertThat(
+        Assertions.assertThrows(
+            Exception.class,
+            () -> cluster.callApi().onAnyBroker(
```
kfaraz: Consider using the shorthand `cluster.runSql()` in these cases.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
kfaraz: There is a shorthand available for this as well. You would need to create an `EmbeddedMsqApis` instance and then call `embeddedMsqApis.submitTaskSql`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
```diff
@@ -0,0 +1,282 @@
...
+ * This class tests various query failures.
+ * <p>
+ * - SQL planning failures. Both {@link org.apache.calcite.sql.parser.SqlParseException}
+ * and {@link org.apache.calcite.tools.ValidationException} are tested using SQLs that must fail.
+ * - Various query errors from historicals. These tests use {@link ServerManagerForQueryErrorTest} to make
+ * the query to always throw an exception.
+ */
+public class QueryErrorTest extends QueryTestBase
+{
+  //  TODO: introduce onAnyRouter(...) and use it; add TLS tests in the follow-up patches
```
kfaraz: Please convert this to a regular comment instead of TODO.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
```diff
@@ -0,0 +1,282 @@
...
+        ExceptionMatcher.of(HttpResponseException.class)
+                        .expectMessageContains("429")
+    );
+  }
+
+  @ParameterizedTest
+  @FieldSource("SHOULD_USE_SQL_ENGINE")
+  public void testQueryUnsupported(boolean shouldUseSqlEngine)
+  {
+    MatcherAssert.assertThat(
```
kfaraz, line 178: Rather than using a `shouldUseSqlEngine` flag, it would be better to separately test the native and SQL APIs. It is okay to have some duplication if it makes the tests more self-explanatory.

You might need to add new methods to `EmbeddedClusterApis`:
- `runSql(...)` which accepts a `Map` queryContext
- `runNativeQuery(...)` to act as shorthand for `onAnyBroker(b -> b.submitNativeQuery(...))`

For example:
```java
final Map<...> queryContext = buildTestContext(...);

// Verify behaviour of native query
MatcherAssert.assertThat(
        Assertions.assertThrows(
            Exception.class,
            () -> cluster.callApi().runNativeQuery(bulidScanQuery(queryContext));
        ),
        ExceptionMatcher.of(HttpResponseException.class).expectMessageContains("429")
);

// Verify behaviour of SQL query
MatcherAssert.assertThat(
        Assertions.assertThrows(
            Exception.class,
            () -> cluster.callApi().runSql(context, "SELECT * FROM %s", dataSource);
        ),
        ExceptionMatcher.of(HttpResponseException.class).expectMessageContains("429")
    );

```

kfaraz (commented):
Thanks for the PR, @Fly-Style !

I have left some thoughts.

kfaraz:
@Fly-Style , since we are getting rid of `CliHistoricalForQueryErrorTest`, we would also need to clean up the following:
- All the places which use the corresponding Druid command `historical-for-query-error-test`
- The TestNG groups `query-retry` and `query-error` and their mentions in the workflow files
- `docker-compose.query-retry-test.yml` and `docker-compose.query-error-test.yml`

integration-tests/src/test/java/org/apache/druid/tests/query/ITQueryRetryTestOnMissingSegments.java
Fly-Style: Nope, I think I wrote somewhere that it will be a part of the next consequent PR, if it is fine.
If not, I will expand the scope of this PR then.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
Fly-Style: Ah, a comment has been lost for some reason. Actually, it is a commit message :D 

This sleep is needed to prolong the time distance between the ingestion task **end** and test method **start**. 
Even though the method on line 98 is named `waitForTaskToSucceed`, in fact some tests are started before the task succeeds and they fail with 'data source is not available'. 
I may just assume, that it is ended on overlord, but broker/router still don't get updated metadata. 
From top of your head, is there any check/call to ensure that data source exists?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
Fly-Style: Wow nice! Thanks in advance!

integration-tests/src/test/java/org/apache/druid/tests/query/ITQueryRetryTestOnMissingSegments.java
Fly-Style: P.S. I will move `ITQueryRetryTestOnMissingSegments`  in this PR too.

kfaraz: Thanks!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
kfaraz: Ah, that makes sense. For that, you may use `cluster.callApi().waitForAllSegmentsToBeAvailable()`.

Fly-Style: Also tried it, still have same issue. Anyway, would use your suggestion.
Overall, I don't like and even hate to put thread sleep but seems `wait..` api methods don't provide me that a-la 'happens-before'

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
Fly-Style: https://github.com/apache/druid/pull/18786/commits/0d5cde4236e9d85cd6f4fb52e505a5698bcae531

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
Fly-Style: https://github.com/apache/druid/pull/18786/commits/0d5cde4236e9d85cd6f4fb52e505a5698bcae531

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
Fly-Style: Addressed in https://github.com/apache/druid/pull/18786/commits/0d5cde4236e9d85cd6f4fb52e505a5698bcae531

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
Fly-Style: https://github.com/apache/druid/pull/18786/commits/0d5cde4236e9d85cd6f4fb52e505a5698bcae531

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
Fly-Style: https://github.com/apache/druid/pull/18786/commits/0d5cde4236e9d85cd6f4fb52e505a5698bcae531

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
Fly-Style: https://github.com/apache/druid/pull/18786/commits/0d5cde4236e9d85cd6f4fb52e505a5698bcae531

integration-tests/src/test/java/org/apache/druid/tests/query/ITQueryRetryTestOnMissingSegments.java
Fly-Style: https://github.com/apache/druid/pull/18786/commits/0d5cde4236e9d85cd6f4fb52e505a5698bcae531

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -294,6 +294,8 @@ public void waitForAllSegmentsToBeAvailable(String dataSource, EmbeddedCoordinat
     final int numTombstones = (int) segments.stream().filter(DataSegment::isTombstone).count();
     final int numSegments = segments.size() - numTombstones;
 
+    // System.err.println("Segments available: " + segments.size());
```
kfaraz: Leftover from testing?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryErrorTest.java
Fly-Style: I slept a night with that proposition and ... well, it makes sense to me at the end. Will make it in the final commit soon.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
Fly-Style: Yeap. I can remove it in the consequent PR, if it is not critical.

kfaraz: I feel It would be cleaner to do it in this PR itself, so that we can keep the history clean.
The additional CI it triggers would help us have another run of our new tests. 🙂

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -435,7 +435,12 @@ public static String newTaskId(String dataSource)
   public static Map<String, Object> deserializeJsonToMap(String payload)
   {
     try {
-      return TestHelper.JSON_MAPPER.readValue(payload, new TypeReference<>() {});
+      return TestHelper.JSON_MAPPER.readValue(
+          payload,
+          new TypeReference<>()
+          {
+          }
+      );
```
kfaraz: Nit: extra change? Please revert this if not intentional.

Fly-Style: Required by checkstyle ¯\_(ツ)_/¯

kfaraz: Oh, that's odd. Did `mvn checkstyle:checkstyle` fail without this change?

Fly-Style: Yes, it did :(

kfaraz: Ah, it must be because a previous commit had already broken it up into multiple lines.
See the commit https://github.com/apache/druid/pull/18786/commits/9b43e1ce40a65be4639c3829074f1eeb7d04223f.

Leaving a suggestion that should address this.

kfaraz:
```suggestion
      return TestHelper.JSON_MAPPER.readValue(payload, new TypeReference<>() {});
```

integration-tests-ex/tools/src/main/resources/META-INF/services/org.apache.druid.initialization.DruidModule
```diff
@@ -13,4 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-org.apache.druid.testing.tools.CustomNodeRoleClientModule
+org.apache.druid.testing.tools.CustomNodeRoleClientModule
```
kfaraz, line 16: I think this change can be reverted (only if you are pushing any other changes too)

kfaraz:
Nice to finally see a successful run! 🎉 🎉

kfaraz:
@Fly-Style , I am going ahead with the merge of this PR.
Small point to note, the `ServerManagerForQueryErrorTest` has currently been duplicated in this PR.
In the follow up PRs, please remove the duplicate and use the one already available in `druid-it-tools`.

PR #18787, opened 2025-11-29
Improved failure messages for MSQCompactionRunner.

gianm:
Include the first failure message in the task status itself, so it is not necessary to fetch task logs to see the error message.

Also, don't log the entire task JSON for failed subtasks. It is logged once when a subtask is initially run, and that's enough.

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
@@ -730,23 +730,23 @@ public class MSQCompactionRunner implements CompactionRunner
       }
       try {
         if (task.isReady(toolbox.getTaskActionClient())) {
-          log.info("Running MSQControllerTask[%d]: %s", taskCnt, json);
+          log.info("Running MSQControllerTask number[%d]: %s", taskCnt, json);
           final TaskStatus taskStatus = task.run(toolbox);
           if (!taskStatus.isSuccess()) {
             failCnt++;
-            log.warn("Failed to run MSQControllerTask[%d]: %s", taskCnt, taskStatus.getErrorMsg());
+            log.warn("Failed to run MSQControllerTask number[%d]: %s", taskCnt, taskStatus.getErrorMsg());
             if (firstFailure == null) {
               firstFailure = taskStatus;
             }
           }
         } else {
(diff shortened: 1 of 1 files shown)


kfaraz (approved):
LGTM

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
```diff
@@ -717,40 +717,57 @@ private static TaskStatus runSubtasks(
...
+        log.info("%s", errMsg);
         return TaskStatus.failure(compactionTaskId, errMsg);
       }
       try {
-        if (eachTask.isReady(toolbox.getTaskActionClient())) {
-          log.info("Running MSQControllerTask: " + json);
-          final TaskStatus eachResult = eachTask.run(toolbox);
-          if (!eachResult.isSuccess()) {
+        if (task.isReady(toolbox.getTaskActionClient())) {
+          log.info("Running MSQControllerTask[%d]: %s", taskCnt, json);
```
cryptoe: Nit: It seems kind of weird that the MSQControllerTask[%d] is not the the task id but the taskcnt. 
Should be rename this to something like
"Running MSQCompaction task number[%d] with payload : %s"

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
```diff
@@ -717,40 +717,57 @@ private static TaskStatus runSubtasks(
...
+          final TaskStatus taskStatus = task.run(toolbox);
+          if (!taskStatus.isSuccess()) {
             failCnt++;
-            log.warn("Failed to run MSQControllerTask: [%s].\nTrying the next MSQControllerTask.", json);
+            log.warn("Failed to run MSQControllerTask[%d]: %s", taskCnt, taskStatus.getErrorMsg());
```
cryptoe: Similar task number comment here.

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
gianm: Updated them all to be like `MSQControllerTask number[%d]`.

PR #18790, opened 2025-11-29
Remove redundant TopNQueryConfig#minTopNThreshold

### LINKED ISSUE #16817: Remove redundant TopNQueryConfig#minTopNThreshold
### Description

Only setting which could be done in it is [minTopNThreshold](https://github.com/apache/druid/blob/954aaafe0c85c1f4967bdb5798c17d4dc813ddd4/processing/src/main/java/org/apache/druid/query/topn/TopNQueryConfig.java#L36)

However that config setting value can be overriden by the context variable [minTopNThreshold](https://druid.apache.org/docs/latest/querying/query-context/#topn) - which is applied in every cases [here](https://github.com/apache/druid/blob/954aaafe0c85c1f4967bdb5798c17d4dc813ddd4/processing/src/main/java/org/apache/druid/query/topn/TopNQueryQueryToolChest.java#L637-L638)

### Motivation

The query context approach is more flexible:
* can be changed for a single query as well
* could still be set globally via runtime properties

Possibly the whole `TopNQueryConfig` class can be removed as the class will be empty without this field.

zhan7236:
## Description

This PR removes the redundant `minTopNThreshold` field from `TopNQueryConfig` as requested in #16817.

The `minTopNThreshold` setting can already be overridden by the query context variable `minTopNThreshold`, making the config-based approach redundant. The default value (1000) is now available as a constant `TopNQueryConfig.DEFAULT_MIN_TOPN_THRESHOLD`.

### Core Changes
- Removed `minTopNThreshold` field and `getMinTopNThreshold()` method from `TopNQueryConfig`
- Updated `TopNQueryQueryToolChest.ThresholdAdjustingQueryRunner` to use the default constant directly when query context doesn't specify the value

### Test Framework Changes
- Removed `@MinTopNThreshold` annotation from `SqlTestFrameworkConfig`
- Updated `SqlTestFramework` to remove `minTopNThreshold` builder method
- Updated tests in `CalciteJoinQueryTest` to use query context (`QueryContexts.MIN_TOP_N_THRESHOLD`) instead of the annotation
- Updated `QueryStackTests` methods to deprecate the `minTopNThreshold` parameter (kept for backward compatibility)
- Updated quidem test file to use `!set minTopNThreshold 1` instead of URL parameter

### Documentation Updates
- Removed `druid.query.topN.minTopNThreshold` configuration section from `docs/configuration/index.md`
- Updated `docs/querying/topnquery.md` to remove the server parameter reference

## Release Note

Removed: The server configuration `druid.query.topN.minTopNThreshold` has been removed. Use the `minTopNThreshold` query context parameter instead to control the TopN threshold per query.

## This PR has:

- [x] been self-reviewed
- [x] added documentation for new or modified features or behaviors
- [x] a release note entry in the PR description (for user-impacting changes)
- [ ] added Javadocs for most classes and all non-trivial methods
- [x] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for coverage is met
- [ ] been tested in a test Druid cluster

Fixes #16817

### DIFF
--- a/sql/src/test/quidem/org.apache.druid.sql.calcite.DecoupledPlanningCalciteJoinQueryTest/testExactTopNOnInnerJoinWithLimit.iq
+++ b/sql/src/test/quidem/org.apache.druid.sql.calcite.DecoupledPlanningCalciteJoinQueryTest/testExactTopNOnInnerJoinWithLimit.iq
@@ -1,4 +1,4 @@
-# testExactTopNOnInnerJoinWithLimit case-crc:b5816ff1
+# testExactTopNOnInnerJoinWithLimit case-crc:f16f81b5
 # quidem testcase reason: EQUIV_PLAN
 !set debug true
 !set defaultTimeout 300000


docs/configuration/index.md
```diff
@@ -2223,9 +2223,11 @@ context). If query does have `maxQueuedBytes` in the context, then that value is
 
 ### TopN query config
 
+The `druid.query.topN.minTopNThreshold` configuration property has been deprecated. Use the `minTopNThreshold` query context parameter instead.
```
gianm: This patch appears to remove it, not just deprecate it. I think it's fine, but the docs should be removed too, and a release note should be added. (Use the "Release notes" label and draft one in the PR description.)

gianm:
Pushed an empty commit to the branch so the github actions can run.

gianm:
I see a failure in `src/test/quidem/org.apache.druid.sql.calcite.DecoupledPlanningCalciteJoinQueryTest/testExactTopNOnInnerJoinWithLimit.iq`, which uses `minTopNThreshold`. We may be able to fix it by replacing this:

```
!use druidtest:///?minTopNThreshold=1
```

With this:

```
!set minTopNThreshold 1
!use druidtest:///
```

Which would set it in the context.

gianm:
@zhan7236 would you like to make the above change?

zhan7236:
> @zhan7236 would you like to make the above change?

Sure, I'll update the PR based on your suggestions. Thanks!

zhan7236:
Fixed the CI failure in `testExactTopNOnInnerJoinWithLimit`. The root cause was an incorrect CRC hash in the quidem testcase header.

The previous commit correctly moved `!set minTopNThreshold 1` to its alphabetically sorted position (as suggested), but the CRC was computed manually without actually running the test, resulting in an incorrect hash (`b5816ff1` instead of the correct `f16f81b5`).

Regenerated the file using `-Dquidem.overwrite` and verified the test passes locally.

PR #18793, opened 2025-12-01
Bind Peon role in CliPeonTest & cleanup tests

abhishekrb19:
- The existing tests in `CliPeonTest` use the default `NodeRole`, which will allow all server modules to be loaded. This patch narrows the behavior by adding only the `Peon` role specifically to the injector that will cause it to load only peon-specific modules for this test. This will help detect any peon-specific bugs quickly because the code paths are slightly different.
- It also removes `FakeCliPeon`, which relied on reflection to manipulate private fields and to set a system property. It was a bit hacky and unnecessary, since we can simply use `CliPeon` directly along with standard `Properties` to achieve the same behavior.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/services/src/test/java/org/apache/druid/cli/CliPeonTest.java
+++ b/services/src/test/java/org/apache/druid/cli/CliPeonTest.java
@@ -97,7 +97,7 @@ public class CliPeonTest
   {
     final Properties properties = new Properties();
     properties.setProperty("druid.indexer.runner.type", "k8s");
-    final Injector peonInjector = makePeonInjector(properties, NoopTask.create());
+    final Injector peonInjector = makePeonInjector(NoopTask.create(), properties);
     final ExecutorLifecycleConfig executorLifecycleConfig = peonInjector.getInstance(ExecutorLifecycleConfig.class);
     Assert.assertFalse(executorLifecycleConfig.isParentStreamDefined());
   }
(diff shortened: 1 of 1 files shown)


services/src/test/java/org/apache/druid/cli/CliPeonTest.java
```diff
@@ -270,19 +254,23 @@ public void testTaskWithMonitorsAndMetricsSpecDoNotCauseCyclicDependency() throw
...
     final CliPeon peon = new CliPeon();
-    peon.taskAndStatusFile = ImmutableList.of(file.getParent(), "1");
+    peon.taskAndStatusFile = ImmutableList.of(taskFile.getParent(), "1");
 
     peon.configure(properties);
     peon.configure(properties, GuiceInjectors.makeStartupInjector());
     return peon.makeInjector(Set.of(NodeRole.PEON));
   }
 
+  private Injector makePeonInjector(Properties properties, Task task) throws IOException
```
jtuglu1: nit: cleaner for fn definition
```java
private Injector makePeonInjector(File taskFile, Properties properties)
```
```java
private Injector makePeonInjector(Task task, Properties properties) throws IOException
```

services/src/test/java/org/apache/druid/cli/CliPeonTest.java
```diff
@@ -97,51 +95,43 @@ public class CliPeonTest
...
-    Assert.assertNotNull(runnable.makeInjector());
+    final Properties properties = new Properties();
+    properties.setProperty("druid.indexer.runner.type", "httpRemote");
+    final Injector peonInjector = makePeonInjector(properties, NoopTask.create());
+    final ExecutorLifecycleConfig executorLifecycleConfig = peonInjector.getInstance(ExecutorLifecycleConfig.class);
+    Assert.assertTrue(executorLifecycleConfig.isParentStreamDefined());
   }
 
   @Test
   public void testCliPeonK8sANdWorkerIsK8sMode() throws IOException
```
jtuglu1: while we're here `testCliPeonK8sANdWorkerIsK8sMode` -> `testCliPeonK8sAndWorkerIsK8sMode`

jtuglu1 (approved):
LGTM – thanks

abhishekrb19:
Thanks, @jtuglu1!
The failing IT is unrelated and is fixed in https://github.com/apache/druid/pull/18807

PR #18802, opened 2025-12-03
Add new compaction policy to prioritize fragmented intervals

kfaraz:
### Description

Druid currently supports two compaction policies to prioritize intervals for compaction.
- `newestSegmentFirst`: prioritizes compaction of intervals with the latest data
  - Assumes that newer intervals are queried more frequently and thus should be picked first for the partitioning improvements provided by compaction.
  - ⚠️ Can get into a state where it keeps compacting the latest interval if new data keeps coming in or if the compaction job keeps failing
  - ⚠️ Tries to compact future intervals
  - ⚠️ Does not account for cluster stability which is a function of the number of used segments in the cluster
- `fixedIntervalOrder`: explicitly specifies the datasources and intervals which should be compacted
  - 🔧 Meant to be a test only policy and running it in production is impractical as it requires a lot of manual intervention

### Proposed policy

`mostFragmentedFirst` - prioritizes compaction of the most fragmented intervals

- Prioritizes intervals that are expected to provide the greatest reduction in segment count.
- Allows specifying thresholds which must be met in an interval to trigger compaction.
An interval is eligible for compaction if and only if:
   - The number of uncompacted segments in the interval is at least `minUncompatedCount`.
   - The total uncompacted bytes in the interval is at least `minUncompatedBytes`
   - The average size of uncompacted segments is at most `maxAverageUncompactedBytesPerSegment`

Prioritizes an interval based on the "fragmentation index" of the interval which is computed as follows:

```java
// The first term is simply the number of uncompacted segments in the interval
segmentCountTerm = numUncompactedSegments

// The second term accounts for the smallness of the segments in the interval
// Smaller segments are preferred for compaction as they are expected to provide a greater reduction
// in segment count
smallnessRatio = (maxAverageUncompactedBytesPerSegment / averageSizeOfUncompactedSegments)
smallnessTerm = minUncompactedCount * smallnessRatio

fragmentationIndex = segmentCountTerm + smallnessTerm
```

### Changes

- Update `CompactionStatus.compute()` to track the number of compacted and uncompacted segments
- Add new compaction policy `mostFragmentedFirst` which prioritizes intervals with most small uncompacted segments

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
+++ b/server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
@@ -115,7 +115,7 @@ public class MostFragmentedIntervalFirstPolicy extends BaseCandidateSearchPolicy
   {
     final double fragmentationDiff
         = computeFragmentationIndex(candidateB) - computeFragmentationIndex(candidateA);
-    return fragmentationDiff > 0 ? 1 : -1;
+    return (int) fragmentationDiff;
   }
 
   @Override
--- a/server/src/test/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicyTest.java
+++ b/server/src/test/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicyTest.java
@@ -172,6 +172,26 @@ public class MostFragmentedIntervalFirstPolicyTest
     Assertions.assertTrue(policy.compareCandidates(candidateB, candidateA) > 0);
   }
 
+  @Test
+  public void test_compareCandidates_returnsZeroIfSegmentCountAndAvgSizeScaleEquivalently()
+  {
+    final MostFragmentedIntervalFirstPolicy policy = new MostFragmentedIntervalFirstPolicy(
+        100,
+        HumanReadableBytes.valueOf(1),
+        HumanReadableBytes.valueOf(100),
+        null
+    );
+
+    final CompactionCandidate candidateA = createCandidate(100, 25);
+    final CompactionCandidate candidateB = createCandidate(400, 100);
+
+    verifyCandidateIsEligible(candidateA, policy);
+    verifyCandidateIsEligible(candidateB, policy);
+
+    Assertions.assertEquals(0, policy.compareCandidates(candidateA, candidateB));
+    Assertions.assertEquals(0, policy.compareCandidates(candidateB, candidateA));
+  }
+
   private CompactionCandidate createCandidate(int numSegments, long avgSizeBytes)
   {
     final CompactionStatistics dummyCompactedStats = CompactionStatistics.create(1L, 1L, 1L);


server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -0,0 +1,130 @@
...
+  private static final long SIZE_10_MB = 10_000_000;
+
+  private final int minUncompactedCount;
+  private final long minUncompactedBytes;
+  private final long maxUncompactedSize;
+
+  @JsonCreator
+  public MostFragmentedIntervalFirstPolicy(
+      @JsonProperty("minUncompactedCount") @Nullable Integer minUncompactedCount,
+      @JsonProperty("minUncompactedBytes") @Nullable Long minUncompactedBytes,
```
gianm: Use `HumanReadableBytes`?

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -0,0 +1,130 @@
...
+  {
+    return minUncompactedBytes;
+  }
+
+  /**
+   * Maximum average size of uncompacted segments in an interval eligible for
+   * compaction. Default value is {@link #SIZE_2_GB}.
+   */
+  @JsonProperty
+  public long getMaxUncompactedSize()
```
gianm: The names of `minUncompactedBytes` and `maxUncompactedSize` are confusing, because they look very similar but one refers to total and one refers to average. How about `minUncompactedBytes` and `maxAverageUncompactedBytesPerSegment`?

gianm: Also, I wish this could be specified in terms of row counts rather than byte counts. The target segment size `rowsPerSegment` is specified as a number of rows, so it's most natural to specify the max average uncompacted size in terms of rows as well. Like, I might say that the target is 3M rows per segment but don't bother compacting if the segments average 2.5M already.

I recognize this requires additional metadata to be available that may not currently be available, so it doesn't have to be done in this PR. I'm just having a thought.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -0,0 +1,130 @@
...
+    final CompactionStatistics compacted = candidate.getCompactedStats();
+    final CompactionStatistics uncompacted = candidate.getUncompactedStats();
+    if (uncompacted == null || compacted == null) {
+      return 0;
+    }
+
+    final long avgUncompactedSize = Math.max(1, uncompacted.getTotalBytes() / uncompacted.getNumSegments());
+
+    // Priority increases as size decreases and number increases
+    final double normalizingFactor = 1000f;
```
gianm: What's the purpose of this factor? seems like it wouldn't affect the results.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -0,0 +1,130 @@
...
+    } else {
+      return uncompacted.getNumSegments() >= minUncompactedCount
+          && uncompacted.getTotalBytes() >= minUncompactedBytes
+          && (uncompacted.getTotalBytes() / uncompacted.getNumSegments()) <= maxUncompactedSize;
+    }
+  }
+
+  /**
+   * Computes the priority of the given compaction candidate by checking the
+   * total number and average size of uncompacted segments.
```
gianm: Intuitively what does the priority "mean"? It's helpful to add that to the javadocs.

Reading through, it seems like the priority boils down to `pow(uncompacted.getNumSegments, 2) / uncompacted.getTotalBytes`. Why that particular formula?

gianm: Unit tests for this function would be helpful, to illustrate its behavior in various situations.

kfaraz: Sure, will update the javadocs.

I wanted it to be directly proportional to the number of uncompacted segments and inversely proportional to the size of uncompacted segments.

So
```
priority
= uncompactedCount / avgUncompactedSize
= uncompactedCount * (uncompactedCount / uncompactedBytes)
```

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
kfaraz: Yes, I think this doesn't serve any purpose any more.

I had started out keeping the priority as a `long` and used a multiplication factor in an effort to get whole number values. But I later realized that it would be difficult to have a universal factor that works for all cases, so best to use a `double` instead.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
kfaraz: Yeah, I agree that average number of rows per segment would be a better parameter for this function.
I have used the average bytes here as a proxy until we have the row count for each segment available in the metadata.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
kfaraz: Sure, will do.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
kfaraz: The other alternative would be to have something like

```
priority = factor1 * uncompactedCount + factor2 * avgUncompactedSize
```

but that seemed like it would be difficult to tune as no set of factors 1 and 2 would apply universally to all cases.

kfaraz: > Intuitively what does the priority "mean"? It's helpful to add that to the javadocs.

Would it also make sense to rename the method as `computeFragmentationIndex()`, since essentially that is what we are trying to achieve here?

kfaraz: Updated the javadocs. Also updated the fragmentation index formula. Details in the PR description.

kfaraz:
Thanks for the review, @gianm !

PR #18803, opened 2025-12-03
Emit supervisorId dimension with streaming task metrics

kfaraz:
### Description

With the support of multiple supervisors ingesting into the same datasource, it becomes difficult to distinguish the metrics originating from the tasks of two different supervisors.

### Changes

- Add dimension `supervisorId` to all metrics emitted by streaming tasks
- Simplify constructor of `TaskRealtimeMetricsMonitor`
- Remove `TaskRealtimeMetricsMonitorBuilder`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeQueryTest.java
@@ -145,7 +145,7 @@ public class EmbeddedMSQRealtimeQueryTest extends BaseRealtimeQueryTest
     // Wait for it to be loaded.
     indexer.latchableEmitter().waitForEventAggregate(
         event -> event.hasMetricName("ingest/events/processed")
-                      .hasDimension(DruidMetrics.DATASOURCE, Collections.singletonList(dataSource)),
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource),
         agg -> agg.hasSumAtLeast(totalRows)
     );
     broker.latchableEmitter().waitForEvent(
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeUnnestQueryTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQRealtimeUnnestQueryTest.java
@@ -37,7 +37,6 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
 
-import java.util.Collections;
 import java.util.List;
 
 public class EmbeddedMSQRealtimeUnnestQueryTest extends BaseRealtimeQueryTest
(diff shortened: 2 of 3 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTask.java
```diff
@@ -197,6 +199,16 @@ public String getCurrentRunnerStatus()
     return (status != null) ? status.toString() : null;
   }
 
+  @Override
+  protected ServiceMetricEvent.Builder getMetricBuilder()
```
gianm: Something is off here. `super.getMetricBuilder()` just returns `metricBuilder`, so what's happening in this method is redundant. It would be equivalent to do this in the constructor rather than `getMetricBuilder`. i.e. simply do `metricBuilder.setDimension(DruidMetrics.SUPERVISOR_ID, supervisorId)`.

However, this highlights a problem: the `metricBuilder` is not thread safe, yet it appears to be used from multiple threads with the pattern `emitter.emit(getMetricBuilder().setMetric(metric, value))`. For example, as far as I can tell, `ingest/segments/count` is emitted in a push thread but certain other metrics are emitted in the main thread. It probably is called infrequently enough that this doesn't cause problems in practice. But it's still a thread safety bug and we should fix it.

How about redefining this method to return a new instance of `ServiceMetricEvent.Builder` each time it's called? That should fix it, and then this implementation you have here would be correct.

indexing-service/src/main/java/org/apache/druid/indexing/common/TaskRealtimeMetricsMonitorBuilder.java
```diff
@@ -41,13 +44,26 @@ public static TaskRealtimeMetricsMonitor build(
     return new TaskRealtimeMetricsMonitor(
         metrics,
         meters,
-        ImmutableMap.of(
-            DruidMetrics.DATASOURCE, new String[]{task.getDataSource()},
-            DruidMetrics.TASK_ID, new String[]{task.getId()},
-            DruidMetrics.TASK_TYPE, new String[]{task.getType()},
-            DruidMetrics.GROUP_ID, new String[]{task.getGroupId()}
-        ),
+        getMetricDimensions(task),
```
gianm: The `Map<String, String[]> dimensions` on `TaskRealtimeMetricsMonitor` could be changed to `ServiceMetricEvent.Builder`, then it could be passed `task.getMetricBuilder()` and avoid the duplicated logic / instanceof check below.

kfaraz: Sounds good 👍🏻 .

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTask.java
kfaraz: Thanks for the suggestion, I have updated `TaskRealtimeMetricsMonitor` accordingly.

kfaraz:
Merging as the IT failure is due to the flaky test `ITNestedQueryPushDownTest`.

PR #18807, opened 2025-12-04
Migrate flaky ITNestedQueryPushDownTest to embedded test

abhishekrb19:
Migrate `ITNestedQueryPushDownTest` to the embedded test framework.

This specific test has particularly been very flaky on PRs often requiring multiple reruns . A few recent failed runs: https://github.com/apache/druid/actions/runs/19917476737/job/57101420884 and https://github.com/apache/druid/actions/runs/19696360296/job/56424984581

Kudos to @kfaraz for leading the efforts on the embedded test framework. The migration itself was quite straightforward. While porting the test, I noticed a SQL bug related to `forcePushDownNestedQuery`, please see the related disabled test.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ForcePushDownNestedQueryTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ForcePushDownNestedQueryTest.java
@@ -60,6 +60,9 @@ import java.util.Map;
  */
 public class ForcePushDownNestedQueryTest extends QueryTestBase
 {
+  private final String interval = "2015-09-12/2015-09-13";
+  private final Map<String, Object> forcePushDownNestedContext = Map.of("forcePushDownNestedQuery", "true");
+
   @Override
   public void beforeAll() throws IOException
   {
@@ -74,12 +77,8 @@ public class ForcePushDownNestedQueryTest extends QueryTestBase
   }
 
   @Test
-  public void test_forcePushDownNestedNativeQueries()
+  public void test_native_forcePushDownNestedQueryWithMultipleAggregators()
   {
-    final String interval = "2015-09-12/2015-09-13";
-    final Map<String, Object> forcePushDownNestedContext = Map.of("forcePushDownNestedQuery", "true");
-
-    // Nested group by query with aggregators in both the inner and outer queries
     verifyQuery(
         GroupByQuery
             .builder()
@@ -94,7 +93,7 @@ public class ForcePushDownNestedQueryTest extends QueryTestBase
                     )
                     .setAggregatorSpecs(new LongSumAggregatorFactory("sumAdded", "added"))
                     .setGranularity(Granularities.ALL)
-                    .setContext(forcePushDownNestedContext)
+                    .setContext(Map.of("forcePushDownNestedQuery", "true"))
                     .build()
             )
             .setInterval(interval)
@@ -110,8 +109,11 @@ public class ForcePushDownNestedQueryTest extends QueryTestBase
             )
         )
     );
+  }
 
-    // Nested group by query with force push down and renamed dimensions
+  @Test
+  public void test_native_forcePushDownNestedQueryWithAliasDimensions()
+  {
     verifyQuery(
         GroupByQuery
             .builder()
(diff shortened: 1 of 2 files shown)


kfaraz:
Thanks a lot for picking this up, @abhishekrb19 ! 🎉 🎉 
I was about to take a look but you beat me to it. Thanks for saving me the trouble!! 🥳

abhishekrb19:
Thanks, @kfaraz! I've also added the original native queries for completeness. Please let me know what you think.
While doing so, I noticed that some of the utilities can be moved into `QueryTestBase`,  so they can be shared across other embedded query tests. The other embedded query tests can also just extend `QueryTestBase`

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryTestBase.java
```diff
@@ -63,7 +64,7 @@ public abstract class QueryTestBase extends EmbeddedClusterTestBase
   /**
    * Hook for the additional setup that needs to be done before all tests.
    */
-  protected void beforeAll()
+  protected void beforeAll() throws IOException
```
kfaraz: Maybe generalize to just have `throws Exception`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ForcePushDownNestedQueryTest.java
```diff
@@ -0,0 +1,406 @@
...
+        + "  SELECT channel, \"user\", SUM(added) AS sumAdded\n"
+        + "  FROM %s\n"
+        + "  GROUP BY channel, \"user\"\n"
+        + ")",
+        dataSource,
+        "9385573"
+    );
+
+    // Nested group by query with force push down and renamed dimensions
+    cluster.callApi().verifySqlQuery(
```
kfaraz, line 277: Let's put each query into a separate test method of its own, for both native and SQL cases.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ForcePushDownNestedQueryTest.java
```diff
@@ -0,0 +1,406 @@
...
+    final File wikiFile = new File(tmpDir, "wiki.gz");
+
+    ByteStreams.copy(
+        DruidProcessingConfigTest.class.getResourceAsStream("/wikipedia/wikiticker-2015-09-12-sampled.json.gz"),
+        Files.newOutputStream(wikiFile.toPath())
+    );
+
+    final String sql = StringUtils.format(
+        "SET waitUntilSegmentsLoad = TRUE;\n"
+        + "REPLACE INTO \"%s\" OVERWRITE ALL\n"
```
kfaraz, line 379: Would it be possible to use `MoreResources.MSQ.INSERT_TINY_WIKI_JSON` here?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ForcePushDownNestedQueryTest.java
```diff
@@ -0,0 +1,406 @@
...
+        + "    )\n"
+        + "  )\n"
+        + "PARTITIONED BY DAY\n"
+        + "CLUSTERED BY channel",
+        dataSource,
+        Calcites.escapeStringLiteral(
+            broker.bindings()
+                  .jsonMapper()
+                  .writeValueAsString(new LocalInputSource(null, null, Collections.singletonList(wikiFile), null))
```
kfaraz, line 402: Can this be simplified in a manner similar to how `MoreResources.MSQ.INSERT_TINY_WIKI_JSON` does it?

kfaraz (commented):
Looks good, left some suggestions.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ForcePushDownNestedQueryTest.java
abhishekrb19: Yeah, I think so. The `tiny_wiki*` datasets contain a `timestamp` column, while the original `wikiticker` dataset used in this test has a `time` column, so switching between them breaks the test with `InsertTimeNull: Null timestamp (__time) encountered during INSERT or REPLACE.`. We could separately reconcile this by updating one of the datasets and relevant tests (probably `tiny_wiki`?)  that will let us reuse things more cleanly; for now, I've just left this wikiticker query in.

kfaraz: Thanks for checking! Yeah, this shouldn't block the PR (this change is much needed! 😄 ).
We can dedupe later.

PR #18812, opened 2025-12-05
Fix bug in Coordinator-based compaction duty and simulator

kfaraz:
### Changes

- Fix bug in `CompactionRunSimulator` so that it honors the compaction policy being verified
- Fix bug in `CompactSegments` so that the duty does not launch tasks for an interval that has already been skipped by the policy 
  - This bug does not occur in Overlord-based compaction supervisors
  - It does not affect any real scenarios currently since the only existing production policy `newestSegmentFirst` never skips intervals anyway.

<hr>

This PR has:

- [ ] been self-reviewed.


server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
```diff
@@ -245,8 +245,10 @@ private int submitCompactionTasks(
 
       if (compactionStatus.isComplete()) {
         snapshotBuilder.addToComplete(candidatesWithStatus);
+        continue;
```
cryptoe, line 248: So these 2 lines was the main fix if I understood correctly ?

kfaraz: Yes, this ensures that we skip launching tasks for intervals which have been skipped by the policy.

kfaraz:

kgyrtkirk:
it seems like `ForcePushDownNestedQueryTest` is flaky on [master](https://github.com/apache/druid/commits/master/)
the first commit I see it failing is the commit associated with [this PR](https://github.com/apache/druid/actions/runs/19955328339/job/57225328413)
the same test failed once in the CI run for this PR as well - but it passed on second run; it might be just coincidence

I wonder if this PR could be related to the stability of that test or not...

kfaraz:
Yes, @kgyrtkirk , the `ForcePushDownNestedQueryTest` does seem to be flaky.
It's unrelated to this PR though as this PR affects only the compaction duty behaviour.

Will try to investigate the failure. cc: @abhishekrb19

abhishekrb19:
https://github.com/apache/druid/pull/18816 should address the `ForcePushDownNestedQueryTest` flakiness.

PR #18813, opened 2025-12-05
Vectorize SpectatorHistogram

jtuglu1:
### Description

Vectorizes SpectatorHistogram. Based off [ApproximateHistogramVectorAggregator](https://github.com/apache/druid/blob/-/extensions-core/histogram/src/main/java/org/apache/druid/query/aggregation/histogram/ApproximateHistogramVectorAggregator.java).

#### Performance

Need to do real-world testing (TODO), but benchmarks are promising (>10x speed-up):
```
Benchmark                                                    (metricName)  (rowsPerSegment)  (vectorize)  Mode  Cnt      Score   Error  Units
SpectatorHistogramAggregatorBenchmark.queryIncrementalIndex         long1           1000000        false  avgt    2  92451.646          us/op
SpectatorHistogramAggregatorBenchmark.queryIncrementalIndex         long1           1000000         true  avgt    2  73224.875          us/op
SpectatorHistogramAggregatorBenchmark.queryQueryableIndex           long1           1000000        false  avgt    2   3776.303          us/op
SpectatorHistogramAggregatorBenchmark.queryQueryableIndex           long1           1000000         true  avgt    2    202.916          us/op
```

I will remove the benchmarks before merging.

#### Other perf ideas
- Possibly apply some fixed step loop unrolling to save even more instructions.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.


jtuglu1:
Test failure is unrelated and being fixed in https://github.com/apache/druid/pull/18814

jtuglu1:
@gianm @clintropolis do you know why many `VectorAggregator` implementations(2 examples below):
- [NullableNumericVectorAggregator](https://github.com/apache/druid/blob/-/processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericVectorAggregator.java#L116-L117)
- [DoubleHllSketchBuildVectorProcessor](https://github.com/apache/druid/blob/master/extensions-core/datasketches/src/main/java/org/apache/druid/query/aggregation/datasketches/hll/vector/DoubleHllSketchBuildVectorProcessor.java#L62-L66)

use the `rows` index to check whether the row being accessed is `null`, but [ApproximateHistogramVectorAggregator](https://github.com/apache/druid/blob/-/extensions-core/histogram/src/main/java/org/apache/druid/query/aggregation/histogram/ApproximateHistogramVectorAggregator.java#L75-L78) does not index into this null vector with the `rows` index to fetch that information (and instead uses `i` directly)?.

in other words: using `selector.getNullVector()[i]` vs `selector.getNullVector()[rows != null ? rows[i] : i]`?

benchmarks/src/test/java/org/apache/druid/benchmark/SpectatorHistogramAggregatorBenchmark.java
```diff
@@ -0,0 +1,329 @@
...
+    blackhole.consume(results);
+  }
+
+  public static void main(String[] args) throws RunnerException
+  {
+    Options opt = new OptionsBuilder()
+        .include(SpectatorHistogramAggregatorBenchmark.class.getSimpleName())
+        .forks(1)
+        .warmupIterations(1)
+        .measurementIterations(2)
```
jasonk000: A main is likely not required since we can use a compiled jmh-jar. If it is, then these numbers should be higher by default, I doubt there is enough time here for them to consistently settle down (warmup) and mitigate any variance (fork + measurement).

benchmarks/src/test/java/org/apache/druid/benchmark/SpectatorHistogramAggregatorBenchmark.java
```diff
@@ -0,0 +1,329 @@
...
+  private GeneratorSchemaInfo schemaInfo;
+  private TimeseriesQuery query;
+
+  /**
+   * Setup everything common for benchmarking both the incremental-index and the queryable-index.
+   */
+  @Setup
+  public void setup()
+  {
+    log.info("SETUP CALLED AT " + System.currentTimeMillis());
```
jasonk000: What's the purpose of this during the JMH run?

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramAggregateHelper.java
```diff
@@ -0,0 +1,132 @@
...
+  private final IdentityHashMap<ByteBuffer, Int2ObjectMap<SpectatorHistogram>> histogramCache = new IdentityHashMap<>();
+
+  public void init(ByteBuffer buffer, int position)
+  {
+    SpectatorHistogram emptyCounts = new SpectatorHistogram();
+    addToCache(buffer, position, emptyCounts);
+  }
+
+  /**
+   * Merge obj ({@link SpectatorHistogram} or {@link Object}) into {@param current}.
```
jasonk000: nit: SpectatorHistogram `or Number`

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramAggregateHelper.java
```diff
@@ -0,0 +1,132 @@
...
+    current.insert(value);
+  }
+
+  /**
+   * Fetches the SpectatorHistogram at the given buffer/position pair in the cache
+   */
+  public SpectatorHistogram get(final ByteBuffer buffer, final int position)
+  {
+    // histogramCache is an IdentityHashMap where the reference of buffer is used for equality checks.
+    // So the returned object isn't impacted by the changes in the buffer object made by concurrent threads.
```
jasonk000, line 75: I don't understand the significance of this comment. Do we expect concurrent read/write of the underlying Int2ObjectMap? (if so, does it need synchronization?).

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramAggregateHelper.java
```diff
@@ -0,0 +1,132 @@
...
+    final Int2ObjectMap<SpectatorHistogram> map = histogramCache.get(oldBuffer);
+    map.remove(oldPosition);
+    if (map.isEmpty()) {
+      histogramCache.remove(oldBuffer);
+    }
+  }
+
+  private void addToCache(final ByteBuffer buffer, final int position, final SpectatorHistogram histogram)
+  {
+    Int2ObjectMap<SpectatorHistogram> map = histogramCache.computeIfAbsent(
```
jasonk000, line 126: does this  need to be concurrent-capable?

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramBufferAggregator.java
```diff
@@ -62,70 +57,41 @@ public void aggregate(ByteBuffer buffer, int position)
...
   @Override
   public Object get(final ByteBuffer buffer, final int position)
   {
-    // histogramCache is an IdentityHashMap where the reference of buffer is used for equality checks.
-    // So the returned object isn't impacted by the changes in the buffer object made by concurrent threads.
-
-    SpectatorHistogram spectatorHistogram = histogramCache.get(buffer).get(position);
-    if (spectatorHistogram.isEmpty()) {
+    SpectatorHistogram histo = innerAggregator.get(buffer, position);
+    if (histo.isEmpty()) {
```
jasonk000: I see this is in the previous code. Regardless, this state seems odd. Should we actively remove it from the buffer if we're never going to expose it?

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramBufferAggregator.java
```diff
@@ -62,70 +57,41 @@ public void aggregate(ByteBuffer buffer, int position)
...
 
   @Override
   public Object get(final ByteBuffer buffer, final int position)
   {
-    // histogramCache is an IdentityHashMap where the reference of buffer is used for equality checks.
-    // So the returned object isn't impacted by the changes in the buffer object made by concurrent threads.
-
-    SpectatorHistogram spectatorHistogram = histogramCache.get(buffer).get(position);
-    if (spectatorHistogram.isEmpty()) {
+    SpectatorHistogram histo = innerAggregator.get(buffer, position);
```
jasonk000: It looks like `innerAggregator.get()` can return null, in which case the next line, `if (histo.isEmpty()) {` might NPE.

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramNumericVectorizedAggregator.java
```diff
@@ -0,0 +1,103 @@
...
+  @Override
+  public void aggregate(ByteBuffer buffer, int position, int startRow, int endRow)
+  {
+    boolean[] isNull = selector.getNullVector();
+    long[] vector = selector.getLongVector();
+    boolean hasNulls = isNull != null;
+    final SpectatorHistogram histogram = innerAggregator.get(buffer, position);
+
+    for (int i = startRow; i < endRow; ++i) {
+      if (!hasNulls || !isNull[i]) {
```
jasonk000: style: personally I think this (and other examples) would be more readable as

```
boolean rowIsNull = hasNulls && isNull[i];
if (!rowIsNull) {
   ...
}
```

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogram.java
```diff
@@ -276,7 +276,7 @@ byte[] toBytes()
     return Arrays.copyOf(buffer.array(), buffer.position());
   }
 
-  void insert(Number num)
+  public void insert(Number num)
```
jasonk000, line 279: We should add another `public void insert(long value)` to avoid the implicit Long/Number allocation.

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramAggregateHelper.java
```diff
@@ -0,0 +1,132 @@
...
+      );
+    }
+  }
+
+  /**
+   * Merge {@param value} into {@param current}.
+   */
+  public void merge(SpectatorHistogram current, long value)
+  {
+    current.insert(value);
```
jasonk000, line 66: (.. refer above, I think this will end up allocating on `SpectatorHistogram::insert` as there is no primitive `long` implementation).

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramNumericVectorizedAggregator.java
```diff
@@ -0,0 +1,103 @@
...
+        innerAggregator.merge(histogram, vector[i]);
+      }
+    }
+  }
+
+  @Override
+  public void aggregate(ByteBuffer buffer, int numRows, int[] positions, @Nullable int[] rows, int positionOffset)
```
jasonk000, line 70: style: would `int[] rows` as `int[] rowMap` or `int[] rowIndexMap` be better?

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramNumericVectorizedAggregator.java
```diff
@@ -0,0 +1,103 @@
...
+    boolean[] isNull = selector.getNullVector();
+    long[] vector = selector.getLongVector();
+    boolean hasNulls = isNull != null;
+
+    final Int2ObjectMap<SpectatorHistogram> histMap = innerAggregator.get(buffer);
+    for (int i = 0; i < numRows; ++i) {
+      int rowIndex = rows != null ? rows[i] : i;
+      if (!hasNulls || !isNull[rowIndex]) {
+        int position = positions[i] + positionOffset;
+        innerAggregator.merge(histMap.get(position), vector[rowIndex]);
```
jasonk000, line 86: perf: Are we likely in practice to see long runs where `position` ends up loading the same map repeatedly? Not sure if worth caching this or how likely this occurrence is.

extensions-contrib/spectator-histogram/src/test/java/org/apache/druid/spectator/histogram/SpectatorHistogramAggregateHelperTest.java
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  @Test
+  public void testMergeWithNumber()
+  {
+    SpectatorHistogram histogram = new SpectatorHistogram();
+
+    helper.merge(histogram, 100);
+    helper.merge(histogram, 200L);
+    helper.merge(histogram, 300.0);
```
jasonk000, line 125: I think this `300.0` is implicitly being cast to Number? Might be clearer as `Float.valueOf(300)`.

extensions-contrib/spectator-histogram/src/test/java/org/apache/druid/spectator/histogram/SpectatorHistogramAggregateHelperTest.java
```diff
@@ -0,0 +1,326 @@
...
+
+    SpectatorHistogram expected = new SpectatorHistogram();
+    expected.insert(100);
+    expected.insert(200L);
+    expected.insert(300.0);
+
+    Assert.assertEquals(expected, histogram);
```
jasonk000, line 132: I'm not sure this assert is testing anything. Probably the `expected` should be written with all-primitives.

jtuglu1:
@jasonk000 one note: I will not be leaving the JMH in the final PR (this is because benchmarks are intended to be Druid core pieces, and not extensions).

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramAggregateHelper.java
jtuglu1: I believe VectorAggregators are not required to be thread-safe by contract, so I didn't modify the logic to be so.

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramBufferAggregator.java
jtuglu1: I think that seems reasonable – it should get cleaned up both at relocate/end. Not sure if we can do a read prior to any writes being committed into the histogram? cc @bsyk

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramNumericVectorizedAggregator.java
jtuglu1: I had thought about this. I think `positions[]` might not be deterministic enough for that to matter and we'd likely spend more time branching on the cached value than not

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramNumericVectorizedAggregator.java
jtuglu1: I just followed the naming of the [interface](https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/query/aggregation/VectorAggregator.java#L70) – will keep as-is for the sake of uniformity.

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramAggregateHelper.java
jtuglu1: VectorAggregators are not required to be thread-safe (see [here](https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/query/aggregation/VectorAggregator.java#L34-L35)), however, the BufferAggregator may be made unsafe due it sharing some common helper methods. I'll double-check those.

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramAggregateHelper.java
jtuglu1: Also worth noting this pattern is repeated in `TDigestSketchBufferAggregator` and `DDSketchBufferAggregator`, which are required to be thread-safe.

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogram.java
bsyk: Minor: This could call the other insert method now.
`this.insert(num.longValue())`

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramBufferAggregator.java
bsyk: This was likely being defensive or copied from another example. There's no harm in returning the empty histogram, it will happily merge with others. It would be a waste of a few bytes of buffer if we had any of these. 

Could these be empty when merging query results? It's possible a segment has no results for a query. Would that return nothing, or an empty aggregation?

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramVectorizedAggregator.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public void aggregate(ByteBuffer buffer, int numRows, int[] positions, @Nullable int[] rows, int positionOffset)
+  {
+    Object[] vector = selector.getObjectVector();
+    if (vector == null) {
+      return;
+    }
+
+    final Int2ObjectMap<SpectatorHistogram> histMap = innerAggregator.get(buffer);
```
jasonk000, line 77: I think this might return null, in theory?

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramBufferAggregator.java
jtuglu1: I'd prefer to stick with the current behavior (to ensure backwards compatibility) (in other words, don't change behavior from buffer aggregator). The other classes which use this sort of bytemap approach evict in a similar fashion.

jtuglu1: cc @clintropolis if you have any thoughts, otherwise I'm inclined to leave as-is.

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramVectorizedAggregator.java
jtuglu1: In theory, yes. I'm fairly certain this is an invariant but I'll include for good measure.

jasonk000 (approved):

extensions-contrib/spectator-histogram/src/main/java/org/apache/druid/spectator/histogram/SpectatorHistogramNumericVectorizedAggregator.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    boolean[] isNull = selector.getNullVector();
+    long[] vector = selector.getLongVector();
+    boolean hasNulls = isNull != null;
+
+    for (int i = startRow; i < endRow; ++i) {
+      boolean rowIsNull = hasNulls && isNull[i];
+      if (!rowIsNull) {
+        innerAggregator.merge(histogram, vector[i]);
```
maytasm: nit: why not just do histogram.insert ?

maytasm (approved):

bsyk (approved):

gianm:
> @gianm @clintropolis do you know why many `VectorAggregator` implementations(2 examples below):
> 
>     * [NullableNumericVectorAggregator](https://github.com/apache/druid/blob/-/processing/src/main/java/org/apache/druid/query/aggregation/NullableNumericVectorAggregator.java#L116-L117)
> 
>     * [DoubleHllSketchBuildVectorProcessor](https://github.com/apache/druid/blob/master/extensions-core/datasketches/src/main/java/org/apache/druid/query/aggregation/datasketches/hll/vector/DoubleHllSketchBuildVectorProcessor.java#L62-L66)
> 
> 
> use the `rows` index to check whether the row being accessed is `null`, but [ApproximateHistogramVectorAggregator](https://github.com/apache/druid/blob/-/extensions-core/histogram/src/main/java/org/apache/druid/query/aggregation/histogram/ApproximateHistogramVectorAggregator.java#L75-L78) does not index into this null vector with the `rows` index to fetch that information (and instead uses `i` directly)?.
> 
> in other words: using `selector.getNullVector()[i]` vs `selector.getNullVector()[rows != null ? rows[i] : i]`?

It looks like a bug in `ApproximateHistogramVectorAggregator` to me.

edit: fix in https://github.com/apache/druid/pull/19086

PR #18816, opened 2025-12-05
Set the default latchable emitter timeout to 20 seconds in `ForcePushDownNestedQueryTest`

abhishekrb19:
- The `ForcePushDownNestedQueryTest` is occasionally flaky on master because the MSQ query in `loadWikipediaTable()` can take longer than the default 10 second timeout to finish and for segments to be loaded and available. Other tests such as `QueryVirtualStorageTest` and `EmbeddedDurableShuffleStorageTest` use a 20 second timeout for similar loads, so increasing the timeout here to 20 seconds should address the flakiness. See the [failed](https://productionresultssa19.blob.core.windows.net/actions-results/8a3b9883-9ea6-4cd0-8bc3-1488a1574788/workflow-job-run-1068f1c4-ce65-5618-8001-cf2bca9a5cdc/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-12-05T18%3A39%3A28Z&sig=zPTboCxyzgjf82V%2BcXOhQ2Ero7tq57HAE3r3PYQRYQc%3D&ske=2025-12-06T02%3A40%3A02Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-12-05T14%3A40%3A02Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2025-12-05T18%3A29%3A23Z&sv=2025-11-05) run: https://github.com/apache/druid/actions/runs/19955328339/job/57225328413
- Include the timeout value in the error message so the timeout is apparent without needing to inspect the stack trace

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ForcePushDownNestedQueryTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/ForcePushDownNestedQueryTest.java
@@ -40,6 +40,7 @@ import org.apache.druid.query.groupby.having.OrHavingSpec;
 import org.apache.druid.segment.TestHelper;
 import org.apache.druid.sql.calcite.planner.Calcites;
 import org.apache.druid.testing.embedded.EmbeddedClusterApis;
+import org.apache.druid.testing.embedded.EmbeddedDruidCluster;
 import org.apache.druid.testing.embedded.msq.EmbeddedMSQApis;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Disabled;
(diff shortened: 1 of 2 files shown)


kfaraz:
@abhishekrb19 , seems like it failed on this PR https://github.com/apache/druid/pull/18820 too.
https://github.com/apache/druid/actions/runs/19984107070/job/57315442102?pr=18820#step:6:8061

Should we try upping it to 30s?

abhishekrb19:
@kfaraz I think 18820 doesn’t include this patch. The log line in the run you linked doesn't include the timeout: `org.apache.druid.java.util.common.ISE: Timed out waiting for event`, which suggests it’s still using the 10s default. I think it may be worth retrying the failed test, if not, rebasing with master should help

kfaraz:
Ah, thanks for the clarification @abhishekrb19 ! Let me double check.

PR #18818, opened 2025-12-05
Add embedded test for object-storage-encoding format for nested field

cecemei:
### Description
Add embedded test for object-storage-encoding format for nested field.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
@@ -53,7 +53,7 @@ public class NestedDataFormatsTest extends EmbeddedClusterTestBase
   {
     return EmbeddedDruidCluster.withEmbeddedDerbyAndZookeeper()
                                .useLatchableEmitter()
-                               .useDefaultTimeoutForLatchableEmitter(60)
+                               .useDefaultTimeoutForLatchableEmitter(120)
                                .addServer(overlord)
                                .addServer(coordinator)
                                .addServer(new EmbeddedIndexer())
@@ -97,20 +97,29 @@ public class NestedDataFormatsTest extends EmbeddedClusterTestBase
     cluster.callApi().waitForAllSegmentsToBeAvailable(dataSource, coordinator, broker);
 
     // Test ingesting with skipping raw json smile format works, same row count, with ~20% storage saving
-    final String metadata = "select sum(num_rows), sum(size) from sys.segments where datasource = '%s' group by datasource";
-    final String defaultFormatResult = cluster.runSql(metadata, datasourceWithDefaultFormat);
-    final String noneObjectStorageFormatResult = cluster.runSql(metadata, dataSource);
+    final String metadataSql = "select sum(num_rows), sum(size) from sys.segments where datasource = '%s'";
+    final String defaultFormatResult = cluster.runSql(metadataSql, datasourceWithDefaultFormat);
+    final String noneObjectStorageFormatResult = cluster.runSql(metadataSql, dataSource);
     Assertions.assertEquals(StringUtils.format("%d,%d", 465_346, 53_000_804), defaultFormatResult);
     Assertions.assertEquals(StringUtils.format("%d,%d", 465_346, 41_938_750), noneObjectStorageFormatResult);
 
     // Test querying on a nested field works
-    final String groupByQuery = "select json_value(event, '$.type') as event_type, count(*) as total from %s group by 1 order by 2 desc, 1 asc limit 10";
+    final String groupByQuery =
+        """
+            select json_value(event, '$.type') as event_type, count(*) as total from %s
+            group by 1 order by 2 desc, 1 asc limit 10
+            """;
     final String queryResultDefaultFormat = cluster.runSql(groupByQuery, datasourceWithDefaultFormat);
     final String queryResultNoneObjectStorage = cluster.runSql(groupByQuery, dataSource);
     Assertions.assertEquals(queryResultDefaultFormat, queryResultNoneObjectStorage);
 
     // Test reconstruct json column works, the ordering of the fields has changed, but all values are perserved.
-    final String scanQuery = "select event, to_json_string(agent) as agent from %s where json_value(event, '$.type') = 'PercentClear' and json_value(agent, '$.os') = 'Android' order by __time asc limit 1";
+    final String scanQuery =
+        """
+            select event, to_json_string(agent) as agent from %s
+            where json_value(event, '$.type') = 'PercentClear' and json_value(agent, '$.os') = 'Android'
+            order by __time asc limit 1
+            """;
     final String scanQueryResultDefaultFormat = cluster.runSql(scanQuery, datasourceWithDefaultFormat);
     final String scanQueryResultNoneObjectStorage = cluster.runSql(scanQuery, dataSource);
     // CHECKSTYLE: text blocks not supported in current Checkstyle version


services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -412,11 +411,11 @@ public SupervisorStatus getSupervisorStatus(String supervisorId)
   // STATIC UTILITY METHODS
 
   /**
-   * Creates a random datasource name prefixed with {@link TestDataSource#WIKI}.
+   * Creates a random datasource name with the given prefixed.
    */
-  public static String createTestDatasourceName()
+  public static String createTestDatasourceName(String prefix)
```
kfaraz: Drive-by comment: Why do we need to specify the prefix? The datasource name is always randomly generated anyway.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
```diff
@@ -0,0 +1,155 @@
...
+                               .addServer(new EmbeddedHistorical())
+                               .addServer(broker)
+                               .addServer(new EmbeddedRouter());
+  }
+
+  @Override
+  @BeforeAll
+  protected void setup() throws Exception
+  {
+    super.setup();
```
kfaraz: It would be cleaner to use a new method annotated with `@BeforeAll` rather than override `setup`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
```diff
@@ -0,0 +1,155 @@
...
+  private final EmbeddedOverlord overlord = new EmbeddedOverlord();
+  private final EmbeddedCoordinator coordinator = new EmbeddedCoordinator();
+
+  private final String defaultFormat = "koala_default";
+
+  @Override
+  protected String getDatasourcePrefix()
+  {
+    return TestDataSource.KOALA;
+  }
```
kfaraz: If you want to use the same datasource in multiple tests, override `refreshDatasourceName()` and make it no-op (similar to `QueryVirtualStorageTest` and other tests), and assign `dataSource = createTestDatasourceName();` in the new setup method.

You shouldn't need to define a new variable for the datasource name and/or specify a specific prefix.

kfaraz (commented):
Thanks for adding the new test, @cecemei !
I have left some minor suggestions, even though the PR is still in draft.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
cecemei: not sure i follow. i want to compare the datasource with the default datasource, so do need to generate random datasource name. for the prefix, i saw that the current prefix is always wiki, and i introduced a new koala http resource and used it in my test. i guess the prefix doesnt matter, we could either just say always prefix with datasource.

processing/src/main/java/org/apache/druid/query/http/ClientSqlQuery.java
```diff
@@ -77,6 +77,11 @@ public ClientSqlQuery(
     this.parameters = parameters;
   }
 
+  public static ClientSqlQuery simple(String query)
```
kfaraz:
```suggestion
  public static ClientSqlQuery forSql(String query)
```

services/src/test/java/org/apache/druid/testing/embedded/indexing/Resources.java
```diff
@@ -131,6 +131,23 @@ public static HttpInputSource wikipedia1Day()
         throw new RuntimeException(e);
       }
     }
+
+    public static HttpInputSource kttm1Day()
```
kfaraz: Maybe include "nested" in the resource name so that it reflects that the file contains some nested data.
```suggestion
    public static HttpInputSource kttmNested1Day()
```

processing/src/test/java/org/apache/druid/segment/TestDataSource.java
```diff
@@ -24,6 +24,7 @@
  */
 public final class TestDataSource
 {
+  public static final String PREFIX = "datasource";
```
kfaraz: I don't think this needs to be a constant, as it is unlikely to be used anywhere else.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
```diff
@@ -412,11 +412,11 @@ public SupervisorStatus getSupervisorStatus(String supervisorId)
   // STATIC UTILITY METHODS
 
   /**
-   * Creates a random datasource name prefixed with {@link TestDataSource#WIKI}.
+   * Creates a random datasource name prefixed with {@link TestDataSource#PREFIX}.
    */
   public static String createTestDatasourceName()
   {
-    return TestDataSource.WIKI + "_" + IdUtils.getRandomId();
+    return TestDataSource.PREFIX + "_" + IdUtils.getRandomId();
```
kfaraz: Just curious, any specific reason to update this?
Does the datasource name in the tests matter?

pom.xml
```diff
@@ -1572,7 +1572,7 @@
                     <consoleOutput>true</consoleOutput>
                     <failsOnError>true</failsOnError>
                     <excludes>
-                        *com/fasterxml/jackson/databind/*
+                        *com/fasterxml/jackson/databind/*,**/NestedDataFormatsTest.java
```
kfaraz, line 1575: Why is the new test excluded from checkstyle?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
```diff
@@ -0,0 +1,146 @@
...
+  private final EmbeddedCoordinator coordinator = new EmbeddedCoordinator();
+
+  private final String defaultFormat = "koala_default";
+
+  @Override
+  protected EmbeddedDruidCluster createCluster()
+  {
+    return EmbeddedDruidCluster.withEmbeddedDerbyAndZookeeper()
+                               .useLatchableEmitter()
+                               .useDefaultTimeoutForLatchableEmitter(60)
```
kfaraz: Does the default timeout of 10s not suffice for this test?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
```diff
@@ -0,0 +1,146 @@
...
+    final String groupByQuery = "select json_value(event, '$.type') as event_type, count(*) as total from %s group by 1 order by 2 desc, 1 asc limit 10";
+    final String queryResultDefaultFormat = cluster.callApi().runSql(groupByQuery, defaultFormat);
+    final String queryResultNoneObjectStorage = cluster.callApi().runSql(groupByQuery, dataSource);
+    Assertions.assertEquals(queryResultDefaultFormat, queryResultNoneObjectStorage);
+
+    // Test reconstruct json column works, the ordering of the fields has changed, but all values are perserved.
+    final String scanQuery = "select event, to_json_string(agent) as agent from %s where json_value(event, '$.type') = 'PercentClear' and json_value(agent, '$.os') = 'Android' order by __time asc limit 1";
+    final String scanQueryResultDefaultFormat = cluster.callApi().runSql(scanQuery, defaultFormat);
+    final String scanQueryResultNoneObjectStorage = cluster.callApi().runSql(scanQuery, dataSource);
+    // CHECKSTYLE: text blocks not supported in current Checkstyle version
```
kfaraz, line 125: Let's consider updating the checkstyle version in that case.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
```diff
@@ -0,0 +1,146 @@
...
+                   .jsonInputFormat()
+                   .inputSource(Resources.HttpData.kttm1Day())
+                   .schemaDiscovery()
+                   .tuningConfig(t -> t.withIndexSpec(IndexSpec.builder().withAutoColumnFormatSpec(spec).build()));
+    final String taskId = EmbeddedClusterApis.newTaskId(dataSource);
+    cluster.callApi().runTask(indexTask.withId(taskId), overlord);
+    cluster.callApi().waitForAllSegmentsToBeAvailable(dataSource, coordinator, broker);
+
+    final String resultAsJson =
+        cluster.callApi().onAnyBroker(b -> b.submitSqlQuery(ClientSqlQuery.simple("select * from sys.segments")));
```
kfaraz: Why not use a 
```sql
SELECT COUNT(*), SUM(num_rows) FROM sys.segments WHERE datasource = '%s'
```

You can fire a separate query for each datasource. I feel the code would be cleaner that way.

kfaraz: You can also just do `cluster.runSql(sqlString)` to get back a CSV response.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Embedded tests for nested data, ingested in different {@link NestedCommonFormatColumnFormatSpec}.
+ */
+public class NestedDataFormatsTest extends EmbeddedClusterTestBase
+{
+  private final EmbeddedBroker broker = new EmbeddedBroker();
+  private final EmbeddedOverlord overlord = new EmbeddedOverlord();
+  private final EmbeddedCoordinator coordinator = new EmbeddedCoordinator();
+
+  private final String defaultFormat = "koala_default";
```
kfaraz:
```suggestion
  private final String datasourceWithDefaultFormat = EmbeddedClusterApis.createTestDatasourceName();
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
```diff
@@ -0,0 +1,146 @@
...
+    final String scanQueryResultNoneObjectStorage = cluster.callApi().runSql(scanQuery, dataSource);
+    // CHECKSTYLE: text blocks not supported in current Checkstyle version
+    Assertions.assertEquals(
+        """
+            "{""type"":""PercentClear"",""percentage"":85}","{""type"":""Mobile Browser"",""category"":""Smartphone"",""browser"":""Chrome Mobile"",""browser_version"":""50.0.2661.89"",""os"":""Android"",""platform"":""Android""}"
+            """.trim(), scanQueryResultDefaultFormat);
+    Assertions.assertEquals(
+        """
+            "{""percentage"":85,""type"":""PercentClear""}","{""browser"":""Chrome Mobile"",""browser_version"":""50.0.2661.89"",""category"":""Smartphone"",""os"":""Android"",""platform"":""Android"",""type"":""Mobile Browser""}"
+            """.trim(), scanQueryResultNoneObjectStorage);
```
kfaraz: style:
```suggestion
            """.trim(),
            scanQueryResultNoneObjectStorage
        );
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
kfaraz: Ah, I see, thanks for the clarification. I thought you meant to use a single datasource in all the tests.
Left an updated suggestion [here](https://github.com/apache/druid/pull/18818#discussion_r2600935709).

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
cecemei: it doesnt matter, but since this is not wikipedia data sometimes, it just seems not right to default to use wiki as prefix.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
cecemei: only later version of checkstyle support the """ style (java 15+), and had to exclude this file since the error thrown was parsing error. in the later version of checkstyle a bunch of other things failed as well, so prefer to make checkstyle a separate pr besides this one.

services/src/test/java/org/apache/druid/testing/embedded/EmbeddedClusterApis.java
kfaraz: sure, that's reasonable. You may just use the string `datasource`. I think we can skip adding it as a constant.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
kfaraz: Okay, thanks for the details.

I think you can also use `// CHECKSTYLE.OFF` to turn off checkstyle without having to exclude the entire file in pom. e.g. see `StringUtils` class.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
cecemei: yea, it takes almost 1 min, more than 30s

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
cecemei: it still complains about error, i think parsing was before style check

>Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:checkstyle (default-cli) on project druid-embedded-tests: An error has occurred in Checkstyle report generation.: Failed during checkstyle configuration: Exception was thrown while processing /Users/cecemei/codespace/bitmap/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java: IllegalStateException occurred while parsing file /Users/cecemei/codespace/bitmap/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java. /Users/cecemei/codespace/bitmap/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java:118:5: unexpected token: Assertions -> [Help 1]

pom.xml
cecemei: parsing error

cecemei:
> Thanks for adding the new test, @cecemei ! I have left some minor suggestions, even though the PR is still in draft.

Updated, please take another look, thx!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
kfaraz: Oh, in that case, let's use `120` to be on the safe side.
Otherwise, the test might end up becoming flaky.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
kfaraz: I see, thanks for checking. Let's just exclude it in the pom until we upgrade the checkstyle version in the follow up PR.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
```diff
@@ -0,0 +1,130 @@
...
+                   .jsonInputFormat()
+                   .inputSource(Resources.HttpData.kttmNested1Day())
+                   .schemaDiscovery()
+                   .tuningConfig(t -> t.withIndexSpec(IndexSpec.builder().withAutoColumnFormatSpec(spec).build()));
+    final String taskId = EmbeddedClusterApis.newTaskId(dataSource);
+    cluster.callApi().runTask(indexTask.withId(taskId), overlord);
+    cluster.callApi().waitForAllSegmentsToBeAvailable(dataSource, coordinator, broker);
+
+    // Test ingesting with skipping raw json smile format works, same row count, with ~20% storage saving
+    final String metadata = "select sum(num_rows), sum(size) from sys.segments where datasource = '%s' group by datasource";
```
kfaraz: I think the `GROUP BY` clause can be omitted.

```suggestion
    final String metadataSql = "select sum(num_rows), sum(size) from sys.segments where datasource = '%s'";
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/NestedDataFormatsTest.java
```diff
@@ -0,0 +1,130 @@
...
+    Assertions.assertEquals(StringUtils.format("%d,%d", 465_346, 41_938_750), noneObjectStorageFormatResult);
+
+    // Test querying on a nested field works
+    final String groupByQuery = "select json_value(event, '$.type') as event_type, count(*) as total from %s group by 1 order by 2 desc, 1 asc limit 10";
+    final String queryResultDefaultFormat = cluster.runSql(groupByQuery, datasourceWithDefaultFormat);
+    final String queryResultNoneObjectStorage = cluster.runSql(groupByQuery, dataSource);
+    Assertions.assertEquals(queryResultDefaultFormat, queryResultNoneObjectStorage);
+
+    // Test reconstruct json column works, the ordering of the fields has changed, but all values are perserved.
+    final String scanQuery = "select event, to_json_string(agent) as agent from %s where json_value(event, '$.type') = 'PercentClear' and json_value(agent, '$.os') = 'Android' order by __time asc limit 1";
```
kfaraz: Please try to break the SQL string into multiple lines.

kfaraz (approved):
LGTM, thanks for the updates, @cecemei !

jtuglu1:
Hey folks, looks like this updated to JDK 17 compiler. This was a change previously reverted here: https://github.com/apache/druid/pull/18759 as it breaks a few things.

PR #18824, opened 2025-12-08
Change segment state on HttpLoadQueuePeon from "moving from" to "dropping" atomically

kfaraz:
### Description

Based on discussions with @jtuglu1 in #18764 , there seem to be some race conditions in segment balancing between the duty run thread and the segment drop callback thread (triggered by segment balancing).

#### Race condition 1

Duty takes snapshot of peon when segment is in the middle of being transitioned from state `MOVING_FROM` to `DROPPING`.

- The result is that the `ServerHolder` assumes that this segment actually has state `LOADED` (when in fact it should be `DROPPING`)
- This leads the `StrategicSegmentAssigner` to assume that the segment is over-replicated (being loaded on both A and B) and then trigger a drop from either A (no-op since we were going to start a drop from A anyway) or B (segment becomes briefly unavailable).

#### Race condition 2

Duty takes snapshot of peon after the `DROP` operation has succeeded but it is not yet reflected in the inventory (since the inventory was snapshotted earlier in the flow).

- This would again cause the `StrategicSegmentAssigner` to consider this segment as over-replicated.
- Something similar may happen with a load operation as well.

### Changes

- Use a synchronized block to ensure that a segment is moved from the set `segmentsMarkedToDrop` to `segmentsToDrop` atomically
- Use `@GuardedBy` to ensure that these fields remain thread-safe
- Pass the list of segments already loaded on a server to `LoadQueuePeon.getSegmentsInQueue()` so that we track completed requests inside the peon until the inventory has confirmed success of the same.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/coordinator/loading/HttpLoadQueuePeon.java
+++ b/server/src/main/java/org/apache/druid/server/coordinator/loading/HttpLoadQueuePeon.java
@@ -74,6 +74,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 /**
  *
@@ -106,6 +107,9 @@ public class HttpLoadQueuePeon implements LoadQueuePeon
    */
   @GuardedBy("lock")
   private final Set<SegmentHolder> queuedSegments = new TreeSet<>();
+
+  @GuardedBy("lock")
+  private final Set<SegmentHolder> recentlySucceededActions = new TreeSet<>();
 
   /**
    * Set of segments for which requests have been sent to the server and can
@@ -420,6 +424,9 @@ public class HttpLoadQueuePeon implements LoadQueuePeon
             } else {
               onRequestCompleted(holder, RequestStatus.SUCCESS, status);
             }
+
+            holder.markRequestSucceeded();
+            recentlySucceededActions.add(holder);
           }
         }, null
     );
@@ -474,6 +481,7 @@ public class HttpLoadQueuePeon implements LoadQueuePeon
       segmentsToLoad.clear();
       queuedSegments.clear();
       activeRequestSegments.clear();
+      recentlySucceededActions.clear();
       queuedSize.set(0L);
       loadingRateTracker.stop();
       stats.get().clear();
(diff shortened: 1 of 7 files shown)


jtuglu1:
Hi @kfaraz, thanks for the patch. While I think this does potentially fix some other issues (and is just generally good to have),  I don't think this fully fixes the aforementioned issue:

- The server inventory view is updated to show `loaded=2` (callback on B is showing).
- We take a copy of the server inventory view in `prepareServers()` which is then incrementing loaded here.
- Then, load on B + drop on A callbacks happen.
- We take a copy of the `queuedSegments` in all the HttpLoadQueuePeon in `prepareCluster()` routine. Since the callbacks have already completed, the copy doesn't see any dropping or loading, etc. (it's effectively empty for that segment). But the old "stale" copy of the server inventory view shows `loaded=2`. 

While this is still improvement, I don't think it fixes the core issue that the server inventory and the httploadqueue peons can be "snapshotted" non-atomically, allowing for how we [build](https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCountMap.java#L52-L61) the SegmentReplicaCount to be messed up. We need to have a way to ensure the `ServerHolder` and the `HttpLoadQueuePeon` can be atomically snapshotted without any callback interleavings happening.

kfaraz:
Thanks for the clarification, @jtuglu1 !
It makes sense to have the inventory view be consistent with the peons.
I also encountered another possible race condition related `segmentsMarkedToDrop`.
In `ServerHolder.initializeQueuedSegments()`, we make a call to `peon.getSegmentsInQueue()` and `peon.getSegmentsMarkedToDrop()` in quick succession, but it is possible that between these calls, the state of the peon changes.

I have tried to address this race condition and the reconciliation with the server inventory snapshot in the latest commit.
Let me know if it makes sense and addresses your issue.

kfaraz:
Currently squashing some more potential race conditions in the flow. Will update this PR soon.

jtuglu1:
Thanks. I'm doing a bit of simulation testing to see if I can squeeze some more out. I've added the other potential one I uncovered here:
```
T0[Coordinator]: Start DutyGroup[HistoricalManagementDuties]
T1[Coordinator]: Start PrepareBalancerAndLoadQueues::run()
T2[Coordinator]: Move initiated of segment[S] from Server[A] to Server[B] (no callbacks yet)
T3[Coordinator]: End PrepareBalancerAndLoadQueues::run()
T4[Coordinator]: End DutyGroup[HistoricalManagementDuties]
T5[Coordinator]: Start DutyGroup[HistoricalManagementDuties]
T6[Coordinator]: Start PrepareBalancerAndLoadQueues::run()
T7[Coordinator]: enter prepareCurrentServers()
T8[Coordinator]: exit prepareCurrentServers()
T9[Coordinator]: enter prepareCluster()
T10[Coordinator-callback]: Server[B] completed request[MOVE_TO] on segment[S] with status[SUCCESS]
T11[Coordinator-callback]: Dropping segment [S] from server[A]
T12[A]: Completely removing segment[S] in [30,000]ms.
T13[Coordinator-callback]: Server[A] completed request[DROP] on segment[S] with status[SUCCESS].
T14[Coordinator]: exit prepareCluster()
// at this point, I think somehow the segmentreplicacount struct has loaded=0 (somehow which causes the other load)
T15[Coordinator-callback]: Server[C] completed request[LOAD] on segment[S] with status[SUCCESS]
T16[Coordinator]: End DutyGroup[HistoricalManagementDuties]
```
This causes a 2nd load

server/src/main/java/org/apache/druid/server/coordinator/loading/HttpLoadQueuePeon.java
```diff
@@ -410,6 +424,9 @@ private void updateSuccessOrFailureInHolder(SegmentHolder holder, SegmentChangeS
             } else {
               onRequestCompleted(holder, RequestStatus.SUCCESS, status);
             }
+
+            holder.markRequestSucceeded();
+            recentlySucceededActions.add(holder);
```
FrankChen021, line 429: [P2] Failed requests are retained as recently succeeded

updateSuccessOrFailureInHolder adds every completed holder to recentlySucceededActions even after SegmentChangeStatus.State.FAILED. getSegmentsInQueue then reports a failed load/drop as still pending until timeout when the inventory does not match the successful end state, so the coordinator can suppress retries or project a failed drop/load incorrectly for the whole load timeout. Only successful statuses should enter this reconciliation set.

server/src/main/java/org/apache/druid/server/coordinator/loading/HttpLoadQueuePeon.java
```diff
@@ -410,6 +424,9 @@ private void updateSuccessOrFailureInHolder(SegmentHolder holder, SegmentChangeS
             } else {
               onRequestCompleted(holder, RequestStatus.SUCCESS, status);
             }
+
+            holder.markRequestSucceeded();
```
FrankChen021, line 428: [P1] Only retain successful requests

updateSuccessOrFailureInHolder is reached for both SUCCESS and FAILED statuses, but this bookkeeping runs after the FAILED branch too. A failed LOAD normally leaves the segment absent from the inventory, so the next getSegmentsInQueue call sees holder.isLoad() != isSegmentLoaded and re-adds this failed holder as a queued action for up to getLoadTimeout() (15 minutes); a failed DROP is similarly treated as freeing space while the segment is still loaded. This suppresses retry/replica correction and can overcommit disk. Add to recentlySucceededActions only for SUCCESS responses, and cover failed load/drop responses with regression tests.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Static review found one P1 correctness issue, attached inline.



---

jtuglu1:
Was working on https://github.com/apache/druid/pull/19889 @kfaraz

kfaraz:
Thanks, @jtuglu1 , I will take a look soon.

kfaraz:
@jtuglu1 , does #19889 also cover the case handled in this PR and can we close this one off?

P.S. This PR is so old now that I don't even recall what other race conditions I was trying to address in this PR. I will need to another look some time soon.

jtuglu1:
> Does https://github.com/apache/druid/pull/19889 also cover the case handled in this PR and can we close this one off?

I think so yeah.

> P.S. This PR is so old now that I don't even recall what other race conditions I was trying to address in this PR. I will need to another look some time soon.

Haha, I may also need a refresher.

PR #18825, opened 2025-12-08
Increase timeout for query to be registered and processed in `SqlQueryCancelTest`

Fly-Style:
This patch fixes occasional failures of `SqlQueryCancelTest`. Unfortunately, the query is not detectable via REST API and currently it is an only good way (known to me) to improve the situation. 

This PR has:

- [x] been self-reviewed.


kfaraz:
I feel it would be nicer if we could wait on a metric instead. Will check if there is any existing metric that we could use for this purpose.

kfaraz (approved):

PR #18830, opened 2025-12-10
Web console: Improve explore max time cancelation

vogievetsky:
Right now cancelation does not propagate to the max time query that the Explore view makes. Also the grouping table module has cancelation broken in general.

As a side benefit: I got Claude to expand the `wait` helper function to be cancelable which is hugely helpful when testing cancelation.

### DIFF
--- a/web-console/src/views/explore-view/modules/grouping-table-module/grouping-table-module.tsx
+++ b/web-console/src/views/explore-view/modules/grouping-table-module/grouping-table-module.tsx
@@ -332,7 +332,6 @@ ModuleRepository.registerModule<GroupingTableParameterValues>({
           <Loader
             cancelText="Cancel query"
             onCancel={() => {
-              console.log('Cancel query clicked');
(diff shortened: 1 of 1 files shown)


web-console/src/views/explore-view/modules/grouping-table-module/grouping-table-module.tsx
```diff
@@ -328,7 +329,14 @@ ModuleRepository.registerModule<GroupingTableParameterValues>({
           />
         ) : undefined}
         {resultState.loading && (
-          <Loader cancelText="Cancel query" onCancel={() => queryManager.cancelCurrent()} />
+          <Loader
+            cancelText="Cancel query"
+            onCancel={() => {
+              console.log('Cancel query clicked');
```
clintropolis: is this supposed to be here'?

vogievetsky: woops. no. good spot

PR #18831, opened 2025-12-10
Revert java 17 compiler

cecemei:
### Description
- revert back to java 11 compiler for hadoop compatibility in root pom
- in `embedded-test` module, force use java 17 to support the multi-line string syntax, and exclude it from strict compilation. 

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,11 @@
     </scm>
 
     <properties>
-        <!--  java 11 is no longer supported, but for backward compatibility reasons we plan to stay on 11 until druid37, see https://github.com/apache/druid/pull/18759  -->
+        <!--  java 11 is no longer supported, but for backward compatibility reasons we plan to stay on 11 until druid37, see https://github.com/apache/druid/pull/18759.
+          Note for upgrade:
+          - remove the maven-compiler-plugin in embedded-tests
+          - remove exclusion of embedded-tests in strict compilation check
+          -->
         <java.version>11</java.version>
         <maven.compiler.release>${java.version}</maven.compiler.release>
         <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>


embedded-tests/pom.xml
```diff
@@ -29,6 +29,7 @@
 
   <properties>
     <surefire.rerunFailingTestsCount>0</surefire.rerunFailingTestsCount>
+    <maven.compiler.release>17</maven.compiler.release>
```
clintropolis, line 34: why is this needed? seems worth a comment

jtuglu1: Maybe:

```
Using JDK 11 language level restriction until Druid 37 which (as of 2025-12-10) is the current target for removing Hadoop support from the dev list threads. Hadoop relies upon JDK 8/11 language level which breaks when upgraded to anything higher.
```

from: https://github.com/apache/druid/pull/18759#issuecomment-3554561707

jtuglu1 (approved):

embedded-tests/pom.xml
cecemei: Thanks all, I added some explanation.

clintropolis: we will also need to remember to remove the changes here and in static-checks.yml in the future, not sure how to best manage that debt, i guess we could put a note in the root pom.xml...

imo we should try to avoid diverging like this in any other modules and try to stay consistent with the root pom unless there is a seriously good reason so that we don't have to keep track of stuff like this.

kfaraz: I think we may fix it up in this patch itself. The only requirement for 17 (AFAIK) is the doc-style strings added in a recent test `NestedDataFormatsTest`. For the time being, we can just stick to regular strings.

cecemei: update the root pom to include things to check for upgrade to 17.

imo using 17 in embedded-test is safe and multi-line string syntax helps with readability a lot, so hopefully we dont let this java 11 thing drag up back.

jtuglu1:
@cecemei good to merge?

PR #18836, opened 2025-12-11
docs: update zookeeper version

317brian:
Fixes the outdated version. We shouldn't have to keep updating it now. 

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [x] been self-reviewed.


docs/design/zookeeper.md
```diff
@@ -27,12 +27,7 @@ Apache Druid uses [Apache ZooKeeper](http://zookeeper.apache.org/) (ZK) for mana
 
 ## Minimum ZooKeeper versions
 
-Apache Druid supports ZooKeeper versions 3.5.x and above.
-
-:::info
- Note: Starting with Apache Druid 0.22.0, support for ZooKeeper 3.4.x has been removed
- Starting with Apache Druid 31.0.0, support for Zookeeper-based segment loading has been removed.
-:::
+Apache Druid supports all stable versions of ZooKeeper. For information about ZooKeeper's stable version, see [ZooKeeper releases](https://zookeeper.apache.org/releases.html).
```
capistrant, line 30: hmmm. this feels like it is true until it isn't, but I guess is better than what it used to say since 3.7 is about to be EOL so 3.5 probably has been for long time. I assume 3.8 clients will be compatible with 3.9 servers which will be next release so this should hold for a while at least.

capistrant (approved):

PR #18838, opened 2025-12-12
Change appender-ref to the more updated AppenderRef

GWphua:
### Description

According to the [log4j2 javadocs](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/config/AppenderRef.html), `Appender-ref` is an alias for `AppenderRef`. 

I think we should standardize to use `AppenderRef` in our examples, so as not to confuse the doc readers.

<hr>

 * `log4j2.xml` configs under `examples/`

<hr>

This PR has:

- [x] been self-reviewed.


GWphua:
Thanks for the catch!

Do take a look again and merge if no other issues. Thanks!

PR #18841, opened 2025-12-12
feat: log total bytes gathered when max scatter-gather bytes limit is reached

### LINKED ISSUE #18840: Log the number of bytes returned from historical when max scatter-gather is reached
### Description

Currently, when the query returned data is bigger than the max scatter-gather, we just log that the query's max scatter-gather bytes limit is reached, but we don't have visibility on how big is the data returned from the historical which makes Druid administration hard.

### Motivation

We have some queries that are different in their nature and a max scatter-gather value can fit one query but not all. We want to this value on the context of the query so it's pretty customized. And for that, we need to know how much each query needed when it fails so we can adjust this value.

TessaIO:
Fixes #18840.

### Description

log total bytes gathered when max scatter-gather bytes limit is reached

#### Release note

<hr> log total bytes gathered when max scatter-gather bytes limit is reached

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/client/DirectDruidClient.java
+++ b/server/src/main/java/org/apache/druid/client/DirectDruidClient.java
@@ -437,11 +437,14 @@ public class DirectDruidClient<T> implements QueryRunner<T>
 
         private void checkTotalBytesLimit(long bytes)
         {
-          if (maxScatterGatherBytes < Long.MAX_VALUE && totalBytesGathered.addAndGet(bytes) > maxScatterGatherBytes) {
+          final long currentTotalBytesGathered = totalBytesGathered.addAndGet(bytes);
+          if (currentTotalBytesGathered > maxScatterGatherBytes) {
             String msg = StringUtils.format(
-                "Query[%s] url[%s] max scatter-gather bytes limit reached.",
+                "Query[%s] url[%s] total bytes gathered[%,d] exceeds maxScatterGatherBytes[%,d]",
                 query.getId(),
-                url
+                url,
+                currentTotalBytesGathered,
+                maxScatterGatherBytes
             );
             setupResponseReadFailure(msg, null);
             throw new ResourceLimitExceededException(msg);
--- a/server/src/test/java/org/apache/druid/query/QueryRunnerBasedOnClusteredClientTestBase.java
+++ b/server/src/test/java/org/apache/druid/query/QueryRunnerBasedOnClusteredClientTestBase.java
@@ -42,7 +42,6 @@ import org.apache.druid.java.util.common.granularity.Granularities;
 import org.apache.druid.java.util.common.io.Closer;
 import org.apache.druid.java.util.metrics.StubServiceEmitter;
 import org.apache.druid.query.aggregation.CountAggregatorFactory;
-import org.apache.druid.query.context.ConcurrentResponseContext;
 import org.apache.druid.query.context.ResponseContext;
 import org.apache.druid.query.timeseries.TimeseriesResultValue;
 import org.apache.druid.query.topn.TopNQueryConfig;
@@ -228,9 +227,7 @@ public abstract class QueryRunnerBasedOnClusteredClientTestBase
 
   protected static ResponseContext responseContext()
   {
-    final ResponseContext responseContext = ConcurrentResponseContext.createEmpty();
-    responseContext.initializeRemainingResponses();
-    return responseContext;
+    return DirectDruidClient.makeResponseContextForQuery();
   }
 
   protected static DataSegment newSegment(


server/src/main/java/org/apache/druid/client/DirectDruidClient.java
```diff
@@ -439,9 +439,10 @@ private void checkTotalBytesLimit(long bytes)
         {
           if (maxScatterGatherBytes < Long.MAX_VALUE && totalBytesGathered.addAndGet(bytes) > maxScatterGatherBytes) {
             String msg = StringUtils.format(
-                "Query[%s] url[%s] max scatter-gather bytes limit reached.",
+                "Query[%s] url[%s] max scatter-gather bytes limit reached. Total bytes gathered are equal to: %d",
                 query.getId(),
-                url
+                url,
+                totalBytesGathered.get()
             );
```
abhishekrb19: We can assign the result of total bytes gathered by this query and also log `maxScatterGatherBytes` while at it. Something like: 
```suggestion
          final long currentTotalBytesGathered = totalBytesGathered.addAndGet(bytes);
          if (maxScatterGatherBytes < Long.MAX_VALUE && currentTotalBytesGathered > maxScatterGatherBytes) {
            String msg = StringUtils.format(
                "Query[%s] url[%s] maxScatterGatherBytes[%,d] limit reached. Total bytes gathered: [%,d]",
                query.getId(),
                url,
                maxScatterGatherBytes,
                currentTotalBytesGathered
            );
```

TessaIO: The condition is updated by also removing `maxScatterGatherBytes < Long.MAX_VALUE` as the condition `currentTotalBytesGathered > maxScatterGatherBytes` already covers it.

TessaIO:
@abhishekrb19 thanks for the suggestion, I think it would be cleaner if the test coverage is addressed in a separate PR as this one is solely focused on improving the logs. wdyt?

server/src/main/java/org/apache/druid/client/DirectDruidClient.java
```diff
@@ -437,11 +437,13 @@ private long checkQueryTimeout()
...
+          final long currentTotalBytesGathered = totalBytesGathered.addAndGet(bytes);
+          if (currentTotalBytesGathered > maxScatterGatherBytes) {
             String msg = StringUtils.format(
-                "Query[%s] url[%s] max scatter-gather bytes limit reached.",
+                "Query[%s] url[%s] max scatter-gather bytes limit reached. Total bytes gathered: %d",
                 query.getId(),
-                url
+                url,
+                currentTotalBytesGathered
             );
```
abhishekrb19, line 448: While at it, let's also include the `maxScatterGatherBytes` value in the exception message.

nit: use `[]` to interpolate per Druid [style-guide](https://github.com/apache/druid/blob/master/dev/style-conventions.md#message-formatting-for-logs-and-exceptions). 
```
Query[%s] url[%s] total bytes gathered[%,d] exceeds maxScatterGatherBytes[%,d]
```

abhishekrb19 (commented):
@TessaIO could you look into the CI test failures? The [test client](https://github.com/apache/druid/blob/master/server/src/test/java/org/apache/druid/query/QueryRunnerBasedOnClusteredClientTestBase.java#L231-L233) doesn't call `responseContext.initialize()` - we can replace the code block as follows:
```java
  protected static ResponseContext responseContext()
  {
    return DirectDruidClient.makeResponseContextForQuery();
  }
```
The condition `maxScatterGatherBytes < Long.MAX_VALUE` was previously masking this behavior.

TessaIO:
@abhishekrb19 thanks for the review, updated!

server/src/main/java/org/apache/druid/client/DirectDruidClient.java
```diff
@@ -437,11 +437,14 @@ private long checkQueryTimeout()
...
+          final long currentTotalBytesGathered = totalBytesGathered.addAndGet(bytes);
+          if (currentTotalBytesGathered > maxScatterGatherBytes) {
             String msg = StringUtils.format(
-                "Query[%s] url[%s] max scatter-gather bytes limit reached.",
+                "Query[%s] url[%s] total bytes gathered[%,d] exceeds maxScatterGatherBytes[%,d]",
                 query.getId(),
-                url
+                url,
+                maxScatterGatherBytes,
+                currentTotalBytesGathered
```
abhishekrb19: @TessaIO  the arguments need to be swapped to match the error message order - total bytes first, then max total bytes

server/src/test/java/org/apache/druid/query/QueryRunnerBasedOnClusteredClientTestBase.java
```diff
@@ -228,9 +228,7 @@ protected static List<Result<TimeseriesResultValue>> expectedTimeseriesResult(in
 
   protected static ResponseContext responseContext()
   {
-    final ResponseContext responseContext = ConcurrentResponseContext.createEmpty();
-    responseContext.initializeRemainingResponses();
-    return responseContext;
+    return DirectDruidClient.makeResponseContextForQuery();
```
abhishekrb19, line 230: could you address the checkstyle unused import errors?

TessaIO:
@abhishekrb19 done

server/src/main/java/org/apache/druid/client/DirectDruidClient.java
TessaIO: thanks for spotting

PR #18847, opened 2025-12-16
Update surefire to 3.5.4 ; upgrade NestedDataScanQueryTest to use junit5

kgyrtkirk:
The changes of #18722 have triggered a bug in surefire which in turn started to write invalid report files.

It also passed thru the CI as surefire was failing incorrectly - but as a sideeffet it started causing interesting [annotations](https://github.com/implydata/druid/actions/runs/20258072960) which I've noticed today.

The test should have failed in that PR - however it hit this bug and hidden the problem...

With surefire 4.5.2 the issue happens "less" ; but it still fails to execute the `NestedDataScanQueryTest`...most likely the `JUnitParamsRunner` doesn't work well with recent surefire and/or junit5...

This PR upgraded surefire to 4.5.2 and disables the problematic testcases of `NestedDataScanQueryTest`.


FrankChen021 (approved):

kgyrtkirk:
thank you @FrankChen021 for taking a look! :)

PR #18849, opened 2025-12-16
Increase timeout for Docker Test

abhishekrb19:
The `test_runIndexTask_andKillData` Docker test has been quite [flaky](https://github.com/apache/druid/actions/workflows/unit-and-integration-tests-unified.yml?query=is%3Afailure+event%3Apush) due to timeouts, often requiring multiple re-runs.

```
Error:  test_runIndexTask_andKillData  Time elapsed: 127.754 s  <<< ERROR!
org.apache.druid.java.util.common.ISE: Timed out waiting for event after [120,000]ms
```

We’ll need to inspect the Docker logs to identify the source of timeout/flakiness, but the logs appear to be missing because they get cleaned up after the test completes? In the meantime, try increasing the timeout from 2 minutes to 3 minutes to see if it reduces or eliminates the flakiness.


abhishekrb19:
With a 3 minute timeout, the Docker test has been successful for five [consecutive](https://github.com/apache/druid/actions/runs/20283420795/job/58265857364?pr=18849) runs

PR #18851, opened 2025-12-18
perf: HttpRemoteTaskRunner enhancements

jtuglu1:
### Description

Clone of https://github.com/apache/druid/pull/18729 but merged into current runner per @kfaraz request.

I've seen on the giant lock in `HttpRemoteTaskRunner` cause severe performance degradation under heavy load(200-500ms per acquisition with 1000s of activeTasks can slow down the startPendingTasks loop in TaskQueue). This leads to scheduling delays, which leads to more lag, which auto-scales more tasks, ..., etc. The runner also has a few (un)documented races abundant in the code. This overhead also slows down query tasks under load (e.g. MSQE and others) which utilize the scheduler for execution.

I'm attempting a rewrite of this class to optimize for throughput and safety.

Apart from the performance improvements/bug fixes, this will also include some new features:
- Simpler code. The old task runner had old, legacy ZK references dangling around as well as a pretty complicated scheduling loop.

I would ultimately like to make this the default `HttpRemoteTaskRunner` and have it run in all tests/production clusters, etc. as I think that would help catch more bugs/issues. 

#### Performance Testing

Test results thus far have shown ~100-300ms speed up per task runner operation (`add()`, etc.). Over 1000s of tasks, this amounts to minutes of delay saved.

#### Release note

Speed up throughput and improve thread safety of HttpRemoteTaskRunner

<hr>

This PR has:

- [ ] been self-reviewed.


kfaraz:
Thanks for creating this PR, @jtuglu1 ! The patch seems much simpler now.
I should be able to complete an initial review today.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1197,109 +1159,82 @@ private void startPendingTaskHandling()
...
         }
 
+        final String workerHost = workerToAssign.getWorker().getHost();
         try {
-          // this will send HTTP request to worker for assigning task
-          if (!runTaskOnWorker(taskItem, immutableWorker.getWorker().getHost())) {
-            if (taskItem.getState() == HttpRemoteTaskRunnerWorkItem.State.PENDING_WORKER_ASSIGN) {
-              taskItem.revertStateFromPendingWorkerAssignToPending();
-            }
+          if (!runTaskOnWorker(taskId, workerHost)) {
```
jtuglu1: A quick note: @kfaraz I was tempted to place this inside the above .compute() to avoid the race, but given it could be holding the lock for a while (and concurrent hashmap locks a range of keys, not necessarily just the key), I opted to unlock (similar to what the existing runner does). Additionally, if something like a removeWorker callback occurs, I'd prefer to update the state right away, rather than wait for an assignment timeout, etc.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -112,71 +107,66 @@
...
- * 1. POST request for assigning a task
- * 2. POST request for shutting down a task
- * 3. GET request for getting list of assigned, running, completed tasks on Middle Manager and its enable/disable status.
- * This endpoint is implemented to support long poll and holds the request till there is a change. This class
- * sends the next request immediately as the previous finishes to keep the state up-to-date.
- * <p>
- * ZK_CLEANUP_TODO : As of 0.11.1, it is required to cleanup task status paths from ZK which are created by the
- * workers to support deprecated RemoteTaskRunner. So a method "scheduleCompletedTaskStatusCleanupFromZk()" is added'
- * which should be removed in the release that removes RemoteTaskRunner legacy ZK updation WorkerTaskMonitor class.
+ * HTTP-based distributed task scheduler that manages assignment of tasks to slots on workers (MiddleManagers).
```
kfaraz:
```suggestion
 * HTTP-based distributed task scheduler that manages assignment of tasks to slots on workers (MiddleManagers or Indexers).
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1521,206 +1456,215 @@ public void taskAddedOrUpdated(final TaskAnnouncement announcement, final Worker
     final Worker worker = workerHolder.getWorker();
 
     log.debug(
-        "Worker[%s] wrote [%s] status for task [%s] on [%s]",
+        "Worker[%s] wrote status[%s] for task[%s] on [%s]",
```
kfaraz:
```suggestion
        "Worker[%s] wrote status[%s] for task[%s] on location[%s]",
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -186,39 +176,28 @@ public class HttpRemoteTaskRunner implements WorkerTaskRunner, TaskLogStreamer,
 
   private final DruidNodeDiscoveryProvider druidNodeDiscoveryProvider;
   private final HttpClient httpClient;
-  private final ObjectMapper smileMapper;
+  private final ObjectMapper objectMapper;
```
kfaraz: Looking at the factory class, we are still passing `@Smile` annotated mapper here. Let's retain the original name of this field to avoid ambiguity with the other mapper (typically injected without an annotation or with `@Json`).

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -402,31 +296,54 @@ private ImmutableWorkerInfo findWorkerToRunTask(Task task)
   }
 
   private boolean runTaskOnWorker(
```
kfaraz, line 305: Please add a javadoc to this method, which clarifies what the return value/excpetion in this method means.
- true: successfully assigned
- false: failed to assign, retry later
- exception: failed to assign, mark task as completed

I wonder if this tri-state should be captured in the return value itself rather than throwing varying exceptions like ISE, IllegalState (using Preconditions). You could return a value which has a `success` boolean, a `retry` boolean and a String `failureReason`, which can be used for alerting/task completion message.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -112,71 +107,66 @@
...
+ *  Task state machine is as follows:
+ *    1. PENDING – Task has been submitted to the scheduler.
+ *    2. PENDING_ASSIGN – Task has been assignment to a worker, but has not started running yet.
+ *    3. EXECUTING – Task is running on a worker.
+ *    4. COMPLETE – Task has completed (success/fail).
+ *  Worker state machine is as follows:
+ *    1. READY – Worker is online and ready to receive new tasks.
+ *    2. PENDING_ASSIGN – A task has been submitted to this worker, but has not started running yet.
+ *    3. BLACKLISTED – Worker has too many failed tasks.
+ *    4. LAZY – Worker has no more task running and been marked as reapable by the worker auto-scaler.
```
kfaraz: This is not needed here and should be moved to the respective enum values.
Just link them here using `@see WorkerHolder.State` and `@see HttpRemoteTaskRunnerWorkItem.State`

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -466,60 +377,82 @@ private boolean runTaskOnWorker(
...
-              taskStatus.getStatusCode(),
-              lastKnownState
-          );
+    AtomicBoolean taskCompleted = new AtomicBoolean(false);
+
+    tasks.compute(
+        taskId,
+        (key, taskEntry) -> {
+          Preconditions.checkState(taskEntry != null, "Expected task[%s] to exist", taskId);
+          if (taskEntry.getResult().isDone()) {
```
kfaraz: Don't use a future here. Instead, keep an additional `TaskStatus resultStatus` or similar field in `HttpRemoteTaskRunnerWorkItem`, which gets updated on calling `taskEntry.setStatus()`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -466,60 +377,82 @@ private boolean runTaskOnWorker(
...
-              + "New state[%s], last known state[%s]",
-              taskStatus.getStatusCode(),
-              lastKnownState
-          );
+    AtomicBoolean taskCompleted = new AtomicBoolean(false);
+
+    tasks.compute(
+        taskId,
+        (key, taskEntry) -> {
+          Preconditions.checkState(taskEntry != null, "Expected task[%s] to exist", taskId);
```
kfaraz: If we find the entry to be null, just set some flag (similar to `taskCompleted`) and return null rather than throwing exception. We can throw the exception outside the `compute`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -466,60 +377,82 @@ private boolean runTaskOnWorker(
...
+            catch (InterruptedException e) {
+              log.warn(e, "Interrupted while getting the last known task status.");
+              Thread.currentThread().interrupt();
+            }
+            catch (ExecutionException e) {
+              // This case should not really happen.
+              log.warn(e, "Failed to get the last known task status. Ignoring this failure.");
+            }
+          } else {
+            // Notify interested parties
```
kfaraz: We are not notifying here.
```suggestion           
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -466,60 +377,82 @@ private boolean runTaskOnWorker(
...
+                    lastKnownState
+                );
+              }
+            }
+            catch (InterruptedException e) {
+              log.warn(e, "Interrupted while getting the last known task status.");
+              Thread.currentThread().interrupt();
+            }
+            catch (ExecutionException e) {
+              // This case should not really happen.
```
kfaraz: This try-catch will not be needed at all, if we don't use a future.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -466,60 +377,82 @@ private boolean runTaskOnWorker(
...
+        (key, taskEntry) -> {
+          Preconditions.checkState(taskEntry != null, "Expected task[%s] to exist", taskId);
+          if (taskEntry.getResult().isDone()) {
+            // This is not the first complete event.
+            try {
+              TaskState lastKnownState = taskEntry.getResult().get().getStatusCode();
+              if (taskStatus.getStatusCode() != lastKnownState) {
+                log.warn(
+                    "The state of the new task complete event is different from its last known state. "
+                    + "New state[%s], last known state[%s]",
```
kfaraz:
```suggestion
                    "Ignoring update to status[%s] as task[%s] has already completed with status[%s].",
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -466,60 +377,82 @@ private boolean runTaskOnWorker(
...
+        workers.compute(
+            workerHost,
+            (key, workerHolder) -> {
+              if (workerHolder != null) {
+                log.info(
+                    "Worker[%s] completed task[%s] with status[%s]",
+                    workerHolder.getWorker().getHost(),
+                    taskStatus.getId(),
+                    taskStatus.getStatusCode()
+                );
```
kfaraz: Nit: Can be 1-lined, I feel.
```suggestion
                log.info("Worker[%s] completed task[%s] with status[%s].", workerHost, taskId, taskStatus);
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -466,60 +377,82 @@ private boolean runTaskOnWorker(
...
+                // Worker is done with this task
+                workerHolder.setLastCompletedTaskTime(DateTimes.nowUtc());
+                blacklistWorkerIfNeeded(taskStatus, workerHolder);
+              } else {
+                log.warn("Could not find worker[%s]", workerHost);
```
kfaraz: Is this really a warning? It is possible that the task finished on the worker and then the worker went away.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: Also, should we keep this log line outside the `compute`? It doesn't need to be inside the compute block.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -466,60 +377,82 @@ private boolean runTaskOnWorker(
...
-      // Notify interested parties
-      taskRunnerWorkItem.setResult(taskStatus);
-      TaskRunnerUtils.notifyStatusChanged(listeners, taskStatus.getId(), taskStatus);
+    );
 
-      // Update success/failure counters, Blacklist node if there are too many failures.
-      if (workerHolder != null) {
-        blacklistWorkerIfNeeded(taskStatus, workerHolder);
+    if (workerHost != null) {
+      synchronized (workerStateLock) {
```
kfaraz, line 410: Doesn't seem like we need to acquire this lock.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -564,20 +498,27 @@ public void nodeViewInitializedTimedOut()
...
-    log.info("Workers have sync'd state successfully.");
+    workers.forEach((workerHost, workerEntry) -> {
+      try {
+        workerEntry.waitForInitialization();
+      }
+      catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
+    });
+    log.info("Workers have synced state successfully.");
```
kfaraz: I don't think changing from the regular for loop to a `forEach` makes much of a difference. It is still possible for new items to get added to the `workers` map after we started the `forEach` iteration.

kfaraz (commented):
Leaving a partial review, will try to finish going through the rest of the changes today.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: Actually, with this one, we ensure that acquiring a key reference (while iterating through it instead of keySet), we also get a synchronized scope to the value to do what we want. I found this to be a better model that looping through a keySet, and then indexing with those (potentially stale) keys.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: I use the `workerStateLock` as a global synchronization point for the worker state. This ensures that scheduling loop doesn't accidentally schedule a task on a worker that is being killed, just been scheduled, etc. It also helps notify sleeping scheduling threads that potential slots are ready due to a worker state change. In this case, a worker has completed a task; this frees up a slot so we should notify interested parties.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: I guess it's not unexpected behavior, but it's certainly not normal behavior (these are MMs, not Peons). I can change to info.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: Ah – I think that's a stale comment. See [here](https://github.com/apache/druid/blob/f1b210a55425459861bccccb4233d21c56e5aaba/indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java#L455) – I just moved this down. Will delete the comment.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: Thanks! Could you also update the log message to include more info e.g. `Could not find worker[%s] while marking task[%s] as complete`?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: Oh, I see.

Can that ever cause a deadlock?

Say a worker is yet to be initialized and a thread is waiting on `waitForInitialization()` while holding the lock for the corresponding key in the map. Will any other thread be able to mark this worker as "initialized" without acquiring that lock?

> I found this to be a better model that looping through a keySet, and then indexing with those (potentially stale) keys.

Hmm, I was hoping we use `.keySet()` followed by `.compute()` itself and move the `.compute()` part into a separate method (similar to how `TaskQueue` does it). We could use that method wherever we add/update the entry for a worker so that we always perform actions using a `compute()` ensuring thread safety.

What do you think?

kfaraz: Also, `.forEach()` doesn't seem to be mutually exclusive with `.compute()` on the same key.
Could you please double check this?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -603,58 +544,62 @@ private Worker toWorker(DiscoveryDruidNode node)
...
-                      null
-                  )
+    log.info("Adding worker[%s]", worker.getHost());
+    synchronized (workerStateLock) {
+      workers.compute(
+          worker.getHost(), (key, workerEntry) -> {
+            cancelWorkerCleanup(worker.getHost());
+
+            // There cannot be any new tasks assigned to this worker as the entry has not been published yet.
+            // That being said, there can be callbacks in taskAddedOrUpdated() where some task suddenly begins running
```
kfaraz:
```suggestion
            // However, there can be callbacks in taskAddedOrUpdated() where some task suddenly begins running
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -603,58 +544,62 @@ private Worker toWorker(DiscoveryDruidNode node)
...
-                      workItem.getTaskType(),
-                      null,
-                      TaskStatus.running(workItem.getTaskId()),
-                      workItem.getLocation(),
-                      null
-                  )
+    log.info("Adding worker[%s]", worker.getHost());
+    synchronized (workerStateLock) {
+      workers.compute(
+          worker.getHost(), (key, workerEntry) -> {
```
kfaraz: Nit:
```suggestion
          worker.getHost(),
          (key, workerEntry) -> {
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -603,58 +544,62 @@ private Worker toWorker(DiscoveryDruidNode node)
...
+            // There cannot be any new tasks assigned to this worker as the entry has not been published yet.
+            // That being said, there can be callbacks in taskAddedOrUpdated() where some task suddenly begins running
+            // on this worker. That method still blocks on this key lock, so it will occur strictly before/after this insertion.
+            if (workerEntry == null) {
+              log.info("Unrecognized worker[%s], rebuilding task mapping", worker.getHost());
+              final List<TaskAnnouncement> expectedAnnouncements = new ArrayList<>();
+              // It might be a worker that existed before, temporarily went away and came back. We might have a set of
+              // tasks that we think are running on this worker. Provide that information to WorkerHolder that
+              // manages the task syncing with that worker.
+              tasks.forEach((taskId, taskEntry) -> {
```
kfaraz: Iterating through all tasks here seems problematic. The original code was doing this too, but now with the introduction of two separate locks and two separate concurrent hash maps, we need to take extra care that we don't end up in a deadlock.

Maybe build the expected announcements before acquiring the `workerStateLock`, since the announcements deal with only task state and not worker state.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -112,71 +107,66 @@
...
-  // This is the list of pending tasks in the order they arrived, exclusively manipulated/used by thread that
-  // gives a new task to this class and threads in pendingTasksExec that are responsible for assigning tasks to
-  // workers.
-  @GuardedBy("statusLock")
-  private final List<String> pendingTaskIds = new ArrayList<>();
+  private final PriorityBlockingQueue<PendingTaskQueueItem> pendingTasks = new PriorityBlockingQueue<>();
 
   // All discovered workers, "host:port" -> WorkerHolder
-  private final ConcurrentMap<String, WorkerHolder> workers = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<String, WorkerHolder> workers = new ConcurrentHashMap<>();
```
kfaraz, line 148: This should probably be `@GuardedBy("workerStateLock")` since it is always meant to be accessed within that lock.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -705,31 +650,28 @@ private void scheduleTasksCleanupForWorker(final String workerHostAndPort)
...
-            Set<HttpRemoteTaskRunnerWorkItem> tasksToFail = new HashSet<>();
-            synchronized (statusLock) {
-              for (Map.Entry<String, HttpRemoteTaskRunnerWorkItem> e : tasks.entrySet()) {
-                if (e.getValue().getState() == HttpRemoteTaskRunnerWorkItem.State.RUNNING) {
-                  Worker w = e.getValue().getWorker();
-                  if (w != null && w.getHost().equals(workerHostAndPort)) {
-                    tasksToFail.add(e.getValue());
-                  }
+            final Set<HttpRemoteTaskRunnerWorkItem> tasksToFail = new HashSet<>();
+            tasks.forEach((taskId, taskEntry) -> {
```
kfaraz: Note that `.forEach()` since it is not mutually exclusive with `.compute()` and essentially just provides iteration over a snapshot. But I suppose it is okay here?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -705,31 +650,28 @@ private void scheduleTasksCleanupForWorker(final String workerHostAndPort)
...
               if (!taskItem.getResult().isDone()) {
                 log.warn(
-                    "Failing task[%s] because worker[%s] disappeared and did not report within cleanup timeout[%s].",
+                    "Failing task[%s] because worker[%s] disappeared and did not report within cleanup timeout[%s]",
                     taskItem.getTaskId(),
                     workerHostAndPort,
                     config.getTaskCleanupTimeout()
                 );
-                // taskComplete(..) must be called outside of statusLock, see comments on method.
+                // taskComplete(..) must be called outside workerStatusLock, see comments on method.
```
kfaraz, line 671: Stale comment?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -807,17 +745,15 @@ void syncMonitoring()
...
-          if (workers.containsKey(e.getKey())) {
-            log.makeAlert(
-                "Worker[%s] is not syncing properly. Current state is [%s]. Resetting it.",
-                workerHolder.getWorker().getHost(),
-                workerHolder.getUnderlyingSyncer().getDebugInfo()
-            ).emit();
-            removeWorker(workerHolder.getWorker());
-            addWorker(workerHolder.getWorker());
-          }
+        // TODO: do we want to make this remove/add atomic (e.g. acquire workerLock before)
```
kfaraz: I think it is fine even if we don't acquire the lock since `removeWorker` and `addWorker` would independently acquire the lock as needed. But let's capture that in the comment here.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -844,33 +780,41 @@ Map<String, Object> getWorkerSyncerDebugInfo()
...
+      for (final String workerHost : workers.keySet()) {
+        workers.computeIfPresent(
+            workerHost,
+            (workerHostKey, workerEntry) -> {
+              if (workerEntry.getState() == WorkerHolder.State.BLACKLISTED) {
+                if (shouldRemoveNodeFromBlackList(workerEntry)) {
+                  log.debug("Removing worker[%s] from blacklist", workerHost);
+                  workerEntry.resetContinuouslyFailedTasksCount();
+                  workerEntry.setBlacklistedUntil(null);
+                  workerEntry.setState(WorkerHolder.State.READY);
```
kfaraz, line 791: should we decrement the `blackListedWorkersCount` here?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -844,33 +780,41 @@ Map<String, Object> getWorkerSyncerDebugInfo()
...
+   * This method should be called under the corresponding worker key lock.
+   */
   private boolean shouldRemoveNodeFromBlackList(WorkerHolder workerHolder)
   {
-    if (!workers.containsKey(workerHolder.getWorker().getHost())) {
-      return true;
-    }
-
-    if (blackListedWorkers.size() > workers.size() * (config.getMaxPercentageBlacklistWorkers() / 100.0)) {
+    if (blackListedWorkersCount.get() > workers.size() * (config.getMaxPercentageBlacklistWorkers() / 100.0)) {
```
kfaraz, line 809: Can we put this condition in a method, so that it can be reused by `blacklistWorkerIfNeeded` too?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -890,103 +842,118 @@ private boolean shouldRemoveNodeFromBlackList(WorkerHolder workerHolder)
...
   public Collection<ImmutableWorkerInfo> getWorkers()
   {
-    return workers.values().stream().map(worker -> worker.toImmutable()).collect(Collectors.toList());
-  }
-
-  @VisibleForTesting
-  ConcurrentMap<String, WorkerHolder> getWorkersForTestingReadOnly()
-  {
-    return workers;
+    return workers.values().stream().map(WorkerHolder::toImmutable).collect(Collectors.toList());
```
kfaraz: Let's access this within `workerStateLock` too, for homogeneity. We can reconsider the `workerStateLock` later.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -890,103 +842,118 @@ private boolean shouldRemoveNodeFromBlackList(WorkerHolder workerHolder)
...
   }
 
   @Override
   public Collection<Worker> getLazyWorkers()
   {
-    return lazyWorkers.values().stream().map(holder -> holder.getWorker()).collect(Collectors.toList());
+    // Want this synchronized with lazy-worker routine
```
kfaraz:
```suggestion
    // synchronized since this needs to be mutually exclusive with markWorkersLazy()
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -890,103 +842,118 @@ private boolean shouldRemoveNodeFromBlackList(WorkerHolder workerHolder)
...
+  @GuardedBy("workerStateLock")
+  private boolean isWorkerOkForMarkingLazy(WorkerHolder workerHolder)
   {
-    // Check that worker is not running any tasks and no task is being assigned to it.
-    synchronized (statusLock) {
-      if (workersWithUnacknowledgedTask.containsKey(worker.getHost())) {
-        return false;
-      }
+    // Check that worker is not already lazy, and does not have any in-flight tasks being assigned to it.
+    if (workerHolder.getState() == WorkerHolder.State.LAZY
```
kfaraz, line 939: Shouldn't this apply only to READY workers? Can BLACKLISTED workers also be marked lazy?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -890,103 +842,118 @@ private boolean shouldRemoveNodeFromBlackList(WorkerHolder workerHolder)
...
+        || workerHolder.getState() == WorkerHolder.State.PENDING_ASSIGN) {
+      log.debug(
+          "Skipping marking worker[%s] with incompatiable state[%s]",
```
kfaraz:
```suggestion
          "Skipping marking worker[%s] as lazy since it is in state[%s]",
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -890,103 +842,118 @@ private boolean shouldRemoveNodeFromBlackList(WorkerHolder workerHolder)
...
-        if (e.getValue().getState() == HttpRemoteTaskRunnerWorkItem.State.RUNNING) {
-          Worker w = e.getValue().getWorker();
-          if (w != null && w.getHost().equals(worker.getHost())) {
-            return false;
-          }
+    // Check that worker has no in-flight/running tasks associated with it
+    final AtomicBoolean inProgress = new AtomicBoolean(false);
+    tasks.forEach((taskId, taskEntry) -> {
+      if (taskEntry.getState().inProgress()) {
+        if (taskEntry.getWorker() != null && taskEntry.getWorker()
```
kfaraz: Use `&&` to avoid the nesting.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -890,103 +842,118 @@ private boolean shouldRemoveNodeFromBlackList(WorkerHolder workerHolder)
...
 
-      for (Map.Entry<String, HttpRemoteTaskRunnerWorkItem> e : tasks.entrySet()) {
-        if (e.getValue().getState() == HttpRemoteTaskRunnerWorkItem.State.RUNNING) {
-          Worker w = e.getValue().getWorker();
-          if (w != null && w.getHost().equals(worker.getHost())) {
-            return false;
-          }
+    // Check that worker has no in-flight/running tasks associated with it
+    final AtomicBoolean inProgress = new AtomicBoolean(false);
+    tasks.forEach((taskId, taskEntry) -> {
```
kfaraz: You could probably also do `tasks.values().stream().anyMatch()`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1108,20 +1071,18 @@ public void registerListener(TaskRunnerListener listener, Executor executor)
...
-    synchronized (statusLock) {
-      for (Map.Entry<String, HttpRemoteTaskRunnerWorkItem> entry : tasks.entrySet()) {
-        if (entry.getValue().getState() == HttpRemoteTaskRunnerWorkItem.State.RUNNING) {
-          TaskRunnerUtils.notifyLocationChanged(
-              ImmutableList.of(listenerPair),
-              entry.getKey(),
-              entry.getValue().getLocation()
-          );
-        }
+    tasks.forEach((taskId, taskEntry) -> {
```
kfaraz: Avoid using `.forEach()` since it is not mutually exclusive with `.compute()` even on the same key.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1108,20 +1071,18 @@ public void registerListener(TaskRunnerListener listener, Executor executor)
...
+      if (taskEntry.getState() == HttpRemoteTaskRunnerWorkItem.State.RUNNING) {
+        TaskRunnerUtils.notifyLocationChanged(
```
kfaraz, line 1077: Do not invoke the listeners while holding the locks. Maybe populate some other set or map containing all the notifications that need to be sent out and then trigger them outside the `tasks.compute()` or `tasks.forEach()`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1139,37 +1100,38 @@ public void unregisterListener(String listenerId)
...
-            task,
-            task.getType(),
-            HttpRemoteTaskRunnerWorkItem.State.PENDING
-        );
-        tasks.put(task.getId(), taskRunnerWorkItem);
-        pendingTaskIds.add(task.getId());
+    log.info("Adding task[%s]", task.getId());
+
+    tasks.compute(
+        task.getId(), (id, entry) -> {
```
kfaraz, line 1111:
```suggestion
        task.getId(),
        (id, entry) -> {
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1977,4 +1917,73 @@ private void setStateUnconditionally(State state)
...
+    }
+
+    /**
+     * Compares this item to another for priority queue ordering.
+     * Items are ordered by task priority, then ticket # (fifo).
+     */
+    @Override
+    public int compareTo(PendingTaskQueueItem other)
+    {
+      int priorityComparison = Integer.compare(other.task.getPriority(), this.task.getPriority());
```
kfaraz: Since priority scheduling is a new feature, it would make sense to have it in a separate patch rather than clubbed with the locking improvements in this PR.

Once this is removed, the `PendingTaskQueueItem` will be simplified as we wouldn't need the ticket generator logic. Pending tasks should simply be picked in the order in which they arrived.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1876,14 +1823,19 @@ boolean isPending()
...
       RunnerTaskState toRunnerTaskState()
       {
         return runnerTaskState;
       }
     }
 
-    private Task task;
-    private State state;
+    private volatile Task task;
+    private volatile State state;
```
kfaraz, line 1861: Since the `HttpRemoteTaskRunnerWorkItem` class is always accessed within the required concurrenthashmap lock, I don't think these fields need to be marked as volatile.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1876,14 +1823,19 @@ boolean isPending()
...
 
     HttpRemoteTaskRunnerWorkItem(
```
kfaraz, line 1865: Future work: Can we simplify this class name to just `WorkItem`?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1197,109 +1159,82 @@ private void startPendingTaskHandling()
...
-        String taskId = taskItem.getTaskId();
+          // Mark this worker as unassignable while task is being assigned
+          workers.compute(
+              workerToAssign.getWorker().getHost(), (key, entry) -> {
+                Preconditions.checkState(
+                    entry != null,
+                    "Expected selected worker[%s] to be available",
+                    entry.getWorker().getHost()
+                );
+                Preconditions.checkState(
```
kfaraz, line 1202: These checks should be done outside the `.compute()` since we are already holding the `workerStateLock`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1197,109 +1159,82 @@ private void startPendingTaskHandling()
...
-        }
 
-        if (immutableWorker == null) {
-          throw new ISE("Unexpected state: null immutableWorker");
+          // Mark this task as pending worker assign
+          tasks.compute(
+              taskId,
+              (key, entry) -> {
+                Preconditions.checkState(entry != null, "Expected task[%s] to be in tasks set", taskId);
+                Preconditions.checkState(
```
kfaraz: Can we just log an error / raise an alert and continue with the loop?
That's what would happen anyway. It feels weird to throw these exceptions inside the lambda and have the outermost catch process them as generic unknown exceptions.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: Thanks! Could you add a 1-line comment on why it is okay to do this.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1197,109 +1159,82 @@ private void startPendingTaskHandling()
...
+        final String workerHost = workerToAssign.getWorker().getHost();
         try {
-          // this will send HTTP request to worker for assigning task
-          if (!runTaskOnWorker(taskItem, immutableWorker.getWorker().getHost())) {
-            if (taskItem.getState() == HttpRemoteTaskRunnerWorkItem.State.PENDING_WORKER_ASSIGN) {
-              taskItem.revertStateFromPendingWorkerAssignToPending();
-            }
+          if (!runTaskOnWorker(taskId, workerHost)) {
+            log.warn("Failed to assign task[%s] to worker[%s]. Sending to back of queue", taskId, workerHost);
+            pendingTasks.put(taskItem.withFreshSequenceNumber());
```
kfaraz: Elsewhere, we have used `pendingTasks.offer()`. If there isn't a specific reason to use one or the other, let's keep it consistent and use the same method in all the places.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: Nit: Let's invert the condition.

```
if (runTaskOnWorker(taskId, workerHost)) {
   // do positive things
} else {
   // do negative things
}
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1311,76 +1246,87 @@ private void pendingTasksExecutionLoop()
              .addData("taskId", taskId)
              .emit();
 
-          // taskComplete(..) must be called outside of statusLock, see comments on method.
+          // taskComplete(..) must be called outside workerStatusLock, see comments on method.
```
kfaraz, line 1257: Let's get rid of this comment, and convert the comment on the `taskComplete()` method into a javadoc.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1311,76 +1246,87 @@ private void pendingTasksExecutionLoop()
...
           taskComplete(
-              taskItem,
+              taskId,
               null,
-              TaskStatus.failure(taskId, "Failed to assign this task. See overlord logs for more details.")
+              TaskStatus.failure(
+                  taskId,
+                  StringUtils.format(
```
kfaraz, line 1263: it might be worth updating the `TaskStatus.failure()` method to accept a string message argument and args.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1311,76 +1246,87 @@ private void pendingTasksExecutionLoop()
...
 
-  /**
-   * Must not be used outside of this class and {@link HttpRemoteTaskRunnerResource}
-   */
-  List<String> getPendingTasksList()
-  {
-    synchronized (statusLock) {
-      return ImmutableList.copyOf(pendingTaskIds);
-    }
+    log.warn("Pending tasks execution loop exited");
```
kfaraz, line 1309:
```suggestion
    log.info("Stopped processing of pending tasks");
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1197,109 +1159,82 @@ private void startPendingTaskHandling()
...
-          // taskComplete(..) must be called outside of statusLock, see comments on method.
-          taskComplete(
-              taskItem,
-              null,
-              TaskStatus.failure(
-                  taskId,
-                  "No payload found for this task. "
-                  + "See overlord logs and middleManager/indexer logs for more details."
-              )
+                entry.setState(WorkerHolder.State.PENDING_ASSIGN);
```
kfaraz, line 1209: Does this mean that while we are submitting a task to a worker (using `runTaskOnWorker`), no other task can be submitted to that worker? Do we do this to ensure that we do not over-assign the worker?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1311,76 +1246,87 @@ private void pendingTasksExecutionLoop()
...
-    synchronized (statusLock) {
-      log.info("Shutdown [%s] because: [%s]", taskId, reason);
-      HttpRemoteTaskRunnerWorkItem taskRunnerWorkItem = tasks.get(taskId);
-      if (taskRunnerWorkItem != null) {
-        if (taskRunnerWorkItem.getState() == HttpRemoteTaskRunnerWorkItem.State.RUNNING) {
-          workerHolderRunningTask = workers.get(taskRunnerWorkItem.getWorker().getHost());
-          if (workerHolderRunningTask == null) {
-            log.info("Can't shutdown! No worker running task[%s]", taskId);
+    AtomicReference<WorkerHolder> workerHolderRunningTaskRef = new AtomicReference<>();
+    final AtomicBoolean wasComplete = new AtomicBoolean(false);
```
kfaraz, line 1318:
```suggestion
    final AtomicBoolean isComplete = new AtomicBoolean(false);
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1311,76 +1246,87 @@ private void pendingTasksExecutionLoop()
...
           taskId,
-          workerHolderRunningTask.getWorker().getHost()
+          workerHolderRunningTaskRef.get().getWorker().getHost()
       );
-      workerHolderRunningTask.shutdownTask(taskId);
+      workerHolderRunningTaskRef.get().shutdownTask(taskId);
+    } else if (wasComplete.get()) {
+      log.debug("Task[%s] already completed, no shutdown needed.", taskId);
+    } else {
+      log.debug("Task[%s] not found or not running, no shutdown needed.", taskId);
```
kfaraz: Are these debug logs really needed? `shutdown()` is expected to be idempotent anyway.

This class is already flooded with barely useful debug logs. Let's try to limit the logging to error scenarios, which would be difficult to figure out without the logs.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1408,16 +1354,14 @@ public void stop()
...
-        workers.values().forEach(w -> {
-          try {
-            w.stop();
-          }
-          catch (Exception e) {
-            log.error(e, e.getMessage());
-          }
-        });
-      }
+      workers.forEach((workerHost, workerHolder) -> {
```
kfaraz: `workers` accessed outside `synchronized (workerStateLock)`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1521,206 +1456,215 @@ public void taskAddedOrUpdated(final TaskAnnouncement announcement, final Worker
...
 
   @Override
   public void stateChanged(boolean enabled, WorkerHolder workerHolder)
   {
-    synchronized (statusLock) {
-      statusLock.notifyAll();
+    notifyWatchers();
+  }
+
+  private void notifyWatchers()
```
kfaraz, line 1685:
```suggestion
  private void notifyWorkerAndTaskStateWatchers()
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1521,206 +1456,215 @@ public void taskAddedOrUpdated(final TaskAnnouncement announcement, final Worker
...
-      log.warn("Killing task[%s] on worker[%s].", taskId, worker.getHost());
+    if (shouldShutdownTask.get()) {
+      log.warn("Killing task[%s] on worker[%s]", taskId, worker.getHost());
       workerHolder.shutdownTask(taskId);
     }
 
-    synchronized (statusLock) {
-      statusLock.notifyAll();
-    }
+    // Notify interested parties
```
kfaraz, line 1675: Comment can be removed since the method name is self-explanatory.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1824,13 +1773,11 @@ public int getMaximumCapacityWithAutoscale()
     if (workerBehaviorConfig == null) {
       // Auto scale not setup
       log.debug("Cannot calculate maximum worker capacity as worker behavior config is not configured");
-      maximumCapacity = -1;
```
kfaraz, line 1827: Nit: To ensure that we assign this variable correctly, make it a `final` variable and retain these `-1` assignments.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1950,18 +1902,6 @@ public void setState(State state)
       setStateUnconditionally(state);
     }
 
-    public void revertStateFromPendingWorkerAssignToPending()
```
kfaraz, line 1953: I suppose we don't need this anymore since we simply add the task item back to the queue, right?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: The task entry in the `tasks` map would now have state PENDING_WORKER_ASSIGN. Do we need to revert it back to `PENDING` (same as old code)?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1521,206 +1456,215 @@ public void taskAddedOrUpdated(final TaskAnnouncement announcement, final Worker
...
+    final AtomicBoolean isTaskCompleted = new AtomicBoolean(false);
 
-                // fall through
-              case RUNNING:
-                if (worker.getHost().equals(taskItem.getWorker().getHost())) {
-                  if (!announcement.getTaskLocation().equals(taskItem.getLocation())) {
+    tasks.compute(
+        taskId,
+        (key, taskEntry) -> {
+          if (taskEntry == null) {
```
kfaraz, line 1496: Can we move this entire lambda into a new method?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1521,206 +1456,215 @@ public void taskAddedOrUpdated(final TaskAnnouncement announcement, final Worker
...
+            // Try to find information about it in the TaskStorage
+            Optional<TaskStatus> knownStatusInStorage = taskStorage.getStatus(taskId);
```
kfaraz, line 1498: Maybe perform this before the `.compute()`?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1521,206 +1456,215 @@ public void taskAddedOrUpdated(final TaskAnnouncement announcement, final Worker
...
+                switch (taskEntry.getState()) {
+                  case PENDING:
+                  case PENDING_WORKER_ASSIGN:
+                    taskEntry.setWorker(worker);
+                    taskEntry.setState(HttpRemoteTaskRunnerWorkItem.State.RUNNING);
+                    log.info("Task[%s] started RUNNING on worker[%s].", taskId, worker.getHost());
+
+                    final ServiceMetricEvent.Builder metricBuilder = new ServiceMetricEvent.Builder();
+                    IndexTaskUtils.setTaskDimensions(metricBuilder, taskEntry.getTask());
+                    emitter.emit(metricBuilder.setMetric(
```
kfaraz, line 1560: Let's do metric emission, listener notification, etc. outside the `.compute()` block.

kfaraz (commented):
Finished going through the bulk of the changes.

On the whole, the patch looks good. I have these major suggestions:

- For the time being, it would be cleaner to use `workerStateLock` consistently whenever accessing the `workers` map. We can try to improve this later.
- Avoid use of `.forEach()` and use `.compute()` instead, preferably encasing it in an `addOrUpdate` method similar to `TaskQueue`.
- Do not perform any heavy operation like metadata store access, metric emission, listener notification, etc. inside the `.compute()` lambda.
- Avoid throwing exceptions inside the lambda, if they are just to be caught back in the same method/loop. Instead, log an error and continue with the loop.
- Remove the priority scheduling changes for now.
- Reduce debug logging.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: Not object to changing, but any reason for changing this now? This is the current method used in the current runner.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: While I think that would help moderately with performance, since we're using key-based locking it's not likely to help that much. I can do it, but also nice way (in the logs) to validate that nothing else was modifying this simulatenously.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
kfaraz: Seems unnecessary to invoke `.get()` on a future and then handle the IO exceptions. The alternative would be much simpler.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: Sure, I was trying to keep as similar to original structure (then future iterations could pull more parts out of critical section).

jtuglu1: Let's leave as-is for now. Metric creation + is low overhead/asynchronous and listener notification isn't too much overhead (top-level submission is with a direct executor, but the execution spills into an async threadpool).

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: I was a bit in favor of making this initial rewrite as conservative as the original in terms of most operations being the same (with the exception of a few major ones: add(), etc.). This is mainly just so we can ensure things work properly, then iteratively adjust the remaining parts. Thoughts?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: Yes. Currently the only way to speed up the tput of the scheduler is to make it faster or add more scheduling threads via a config. Ideally there could be some sort of oversubscription mechanism, but I think that's overkill for what this is.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: Yes

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: Sure, I had considered this but opted for parity with old design.

kgyrtkirk:
I will move this out from 36.0.0 for now - it doesn't seem like something which should block the release.
If it gets merged in the upcoming days it could still be ported over to the release branch - and thus be part of it!

jtuglu1:
@gianm any thoughts here?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerResource.java
```diff
@@ -72,7 +72,7 @@ public Response getPendingTasksQueue()
       return Response.status(Response.Status.FORBIDDEN).entity("HttpRemoteTaskRunner is NULL.").build();
     }
 
-    return Response.ok().entity(httpRemoteTaskRunner.getPendingTasksList()).build();
+    return Response.ok().entity(httpRemoteTaskRunner.getPendingTasks()).build();
```
gianm, line 75: This would be a breaking change in the API response. Please avoid it by introducing a new API or something like `@QueryParam("full")`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerResource.java
```diff
@@ -124,7 +124,7 @@ public Response getWorkersWithUnacknowledgedTasks()
       return Response.status(Response.Status.FORBIDDEN).entity("HttpRemoteTaskRunner is NULL.").build();
     }
 
-    return Response.ok().entity(httpRemoteTaskRunner.getWorkersWithUnacknowledgedTasks()).build();
+    return Response.ok().entity(httpRemoteTaskRunner.getPendingAssignWorkers()).build();
```
gianm, line 127: Also a breaking API change, please avoid it by introducing a new API or a `@QueryParam`.

gianm:
> @gianm any thoughts here?

I will try to take a look. It may take some time to get to it, since the changes look quite extensive.

Have you run this on a real production at-scale cluster yet (something with hundreds or thousands of tasks running simultaneously, ideally)? If so, that's always helpful to know.

jtuglu1:
> > @gianm any thoughts here?
> 
> I will try to take a look. It may take some time to get to it, since the changes look quite extensive.
> 
> Have you run this on a real production at-scale cluster yet (something with hundreds or thousands of tasks running simultaneously, ideally)? If so, that's always helpful to know.

Yes, no observed issues. We run with close to 10k tasks at peak per cluster.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -463,63 +365,77 @@ private boolean runTaskOnWorker(
   }
 
   // CAUTION: This method calls RemoteTaskRunnerWorkItem.setResult(..) which results in TaskQueue.notifyStatus() being called
-  // because that is attached by TaskQueue to task result future. So, this method must not be called with "statusLock"
+  // because that is attached by TaskQueue to task result future. So, this method must not be called with "workerStatusLock"
```
gianm, line 366: Should refer to `workerStateLock`?

jtuglu1:
@kfaraz @gianm thoughts here?

jtuglu1:
@kfaraz @gianm any thoughts here?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1521,206 +1480,212 @@ public void taskAddedOrUpdated(final TaskAnnouncement announcement, final Worker
...
-                // fall through
-              case RUNNING:
-                if (worker.getHost().equals(taskItem.getWorker().getHost())) {
-                  if (!announcement.getTaskLocation().equals(taskItem.getLocation())) {
+    tasks.compute(
+        taskId,
+        (key, taskEntry) -> {
+          if (taskEntry == null) {
+            // Try to find information about it in the TaskStorage
+            Optional<TaskStatus> knownStatusInStorage = taskStorage.getStatus(taskId);
```
gianm, line 1498: This is going to need to do a metadata call while holding a (partial) lock on `tasks`. I see the old code did it under `statusLock`, and also there's a resolved conversation about keeping this here. It's fine to keep it here, I suppose, but please include a comment about how this does a metadata call and may cause contention on `tasks`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1190,197 +1153,197 @@ private void startPendingTaskHandling()
...
+        ImmutableWorkerInfo workerToAssign;
+        // Set to false inside tasks.compute() if the task is no longer PENDING when we attempt
+        // to transition it to PENDING_WORKER_ASSIGN. A task can complete concurrently (e.g. via
+        // cancellation) without holding workerStateLock, so we cannot rely on a pre-check.
+        final AtomicBoolean taskWasPending = new AtomicBoolean(true);
+
+        synchronized (workerStateLock) {
+          workerToAssign = findWorkerToRunTask(taskItem.getTask());
+
+          if (workerToAssign == null) {
```
gianm, line 1184: It looks like this code will park and wait if the next task from `pendingTasks` can't be assigned. But there are situations where that task A can't be assigned, but another, later task B can be assigned. For example, if there is 1 free slot, and task A has `requiredCapacity: 2` while task B has `requiredCapacity: 1`. Another example: if strong worker affinity is configured, and none of the affinity workers for task A are available, but affinity workers for task B *are* available.

The old logic would potentially iterate the entire `pendingTaskIds` looking for an assignable task, essentially allowing tasks to skip the line in case they required different capacity or different affinity workers. Please update the new logic to handle this case.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
```diff
@@ -1190,197 +1153,197 @@ private void startPendingTaskHandling()
...
-    synchronized (statusLock) {
-      return ImmutableList.copyOf(pendingTaskIds);
-    }
+    log.warn("Pending tasks execution loop exited");
   }
 
   @Override
   public void shutdown(String taskId, String reason)
   {
-    if (!lifecycleLock.awaitStarted(1, TimeUnit.SECONDS)) {
```
gianm, line 1352: Why remove this guard?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: > The old logic would potentially iterate the entire pendingTaskIds looking for an assignable task, essentially allowing tasks to skip the line in case they required different capacity or different affinity workers. Please update the new logic to handle this case.

Yes, I thought about this. The older logic was a bit cumbersome and hard to read and was overly-conservative (slow) in the locking behavior; would you be opposed to simply rescheduling this task? I was thinking of extending this to some sort of priority/backoff queue to address this problem.

gianm: What do you mean by "rescheduling this task"?

The thing I'm worried about is that we need line-skipping behavior. Especially with strong worker affinity, it's important for tasks to be able to skip the line. For example: a typical config would have a set of affinity workers for batch tasks and a set for realtime tasks. When the batch affinity workers are full we want to continue to assign realtime tasks to the realtime affinity workers.

So, if the solution does give tasks the ability to skip the line, it should be OK.

jtuglu1: > What do you mean by "rescheduling this task"?

Send the task to the back of the queue and effectively just filter through the queue until you can find a task that's runnable, or do a timed wait backoff if none are found after full iteration. This preserves FIFO ordering while still not causing HOL-blocking.

jtuglu1: I think we can address FIFO behavior in a follow-up. That is, prioritizing tasks' priority in the queue based on `Task::getPriority()` for example.

gianm: > Send the task to the back of the queue (effectively just filter through the queue until you can find a task that's runnable). This preserves FIFO ordering while still not causing HOL-blocking.

Sure, that's fine. But be careful to avoid a spin loop of rescheduling if *no* task is currently-schedulable.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunner.java
jtuglu1: > I see the old code did it under statusLock, and also there's a resolved conversation about keeping this here. It's fine to keep it here, I suppose, but please include a comment about how this does a metadata call and may cause contention on tasks.

Yes, I can add a comment. The key point is this will only lock a (hopefully small depending on how ConcurrentHashmap determines the range size) subset of the task keys, allowing other tasks to continue their work.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerResource.java
jtuglu1: Hmm – why would this break the API? The returned container is semantically the same?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerResource.java
jtuglu1: Same – they are just renamed methods, they shouldn't be returning different sets/container types?

gianm: Am I reading it right? Looks like the old `getWorkersWithUnacknowledgedTasks` returns `Map<String, String>`, the new `getPendingAssignWorkers` returns `Collection<ImmutableWorkerInfo>`. The response would change from a key-value object to an array of info objects.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerResource.java
gianm: The old `getPendingTasksList` is `List<String>`, the new `getPendingTasks` is `Collection<? extends TaskRunnerWorkItem>`. The response would change from a list of strings to a list of objects. Right?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerResource.java
jtuglu1: This task runner API is meant to be internal (not user-facing and and meant for debugging). It already currently has unsafe access(es) which IMO are not proper to expose to users. Is it necessary to adhere to this?

gianm: Oh, I see, this resource isn't documented and it's in `/druid-internal`. Nevermind, it's fine to change it, then.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/hrtr/HttpRemoteTaskRunnerResource.java
gianm: Nevermind, per other thread in the same file.

cecemei:
Hello @jtuglu1 , whats the status for this PR, i'm cutting branch for druid 37 soon, should this PR move to next release since it's now in draft state?

jtuglu1:
> Hello @jtuglu1 , whats the status for this PR, i'm cutting branch for druid 37 soon, should this PR move to next release since it's now in draft state?

Hi, let's leave out of v37 release as I don't think this will be merged soon.

PR #18852, opened 2025-12-18
Reduce nesting in `SegmentLocalCacheManager.addFilesToCachedSegments`

GWphua:
#### Add an early return for the if-else block

Before
```
boolean removeInfo = false;
if (!segment.getId().toString().equals(file.getName())) {  
  // Do something without changing removeInfo
} else {  
  removeInfo = true; 
  // All processing logic nested inside the else block, may change removeInfo...
}

if (removeInfo) {
  // Do something... then returns
}
```

After:
```
if (!segment.getId().toString().equals(file.getName())) {  
  // Early Return
}

// Declare removeInfo = true and move on.
```

No functional changes; this is a pure readability improvement

 * `SegmentLocalCacheManager`

<hr>

This PR has:

- [x] been self-reviewed.


kfaraz:
Thanks for the cleanup, @GWphua !

PR #18853, opened 2025-12-18
Fix serde of UpdateLocationAction used by K8s-based tasks

kfaraz:
### Bug

`AbstractTask.setup()` fires an `UpdateLocationAction` when running a K8s-based encapsulated task to notify the Overlord of the task location. This object gets serialized incorrectly causing the location to be null.

### Fix

- Fix serialization of `UpdateLocationAction`
- Add test for serde
- Add null check and throw InvalidInput exception

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/UpdateLocationAction.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/UpdateLocationAction.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.google.common.base.Optional;
+import org.apache.druid.error.InvalidInput;
 import org.apache.druid.indexer.TaskLocation;
 import org.apache.druid.indexing.common.task.Task;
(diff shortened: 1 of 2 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/common/actions/UpdateLocationAction.java
```diff
@@ -23,26 +23,34 @@
...
-  private final TaskLocation taskLocation;
+  private final TaskLocation location;
 
   @JsonCreator
   public UpdateLocationAction(
       @JsonProperty("location") TaskLocation location
   )
   {
-    this.taskLocation = location;
+    InvalidInput.conditionalException(location != null, "No task location specified");
```
GWphua, line 46: What is the difference between `InvalidInput` and `Preconditions.checkNotNull`?

I'm thinking it may be neater to standardise the usage in the codebase?

kfaraz: Using `InvalidInput` throws a `DruidException` which is more user-friendly. `Preconditions` just throws a generic NPE, IIRC.

Yes, I agree that we should standardise it. We can probably address it in a follow up.

kfaraz:
Thanks for the review, @GWphua !

PR #18855, opened 2025-12-18
Add drop reason dimension to ingest/events/thrownAway metric

jtuglu1:
### Description

I've found it hard to build observability on logically-dropped events (either due to transform, null, or temporal conditions). 
This adds:
- Better logging in task logs indicating the breakdown of thrown away events by reason.
- A `reason` dimension to the `ingest/events/thrownAway` metric for aggregating on thrown away reason.
- A `thrownAwayByReason` map to row statistics task API response payload, so future consumers can make use of it should they need to.
- Better interface for row filters, making it easy to adding more filtering reasons in the future.

#### Release note

Add reason dimension to ingest/events/thrownAway metric. This allows for increased observability on why certain events are being logically excluded from ingest.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/stats/DropwizardRowIngestionMeters.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/stats/DropwizardRowIngestionMeters.java
@@ -39,7 +39,7 @@ public class DropwizardRowIngestionMeters implements RowIngestionMeters
   private final Meter processedBytes;
   private final Meter processedWithError;
   private final Meter unparseable;
-  private final Meter[] thrownAwayByReason = new Meter[InputRowFilterResult.NUM_FILTER_RESULT];
+  private final Meter[] thrownAwayByReason = new Meter[InputRowFilterResult.numValues()];
 
   public DropwizardRowIngestionMeters()
   {
@@ -120,9 +120,12 @@ public class DropwizardRowIngestionMeters implements RowIngestionMeters
   @Override
   public Map<String, Long> getThrownAwayByReason()
   {
-    Map<String, Long> result = InputRowFilterResult.buildRejectedCounterMap();
+    Map<String, Long> result = new HashMap<>();
     for (InputRowFilterResult reason : InputRowFilterResult.rejectedValues()) {
-      result.put(reason.getReason(), thrownAwayByReason[reason.ordinal()].getCount());
+      long count = thrownAwayByReason[reason.ordinal()].getCount();
+      if (count > 0) {
+        result.put(reason.getReason(), count);
+      }
     }
     return result;
   }
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/stats/TaskRealtimeMetricsMonitor.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/stats/TaskRealtimeMetricsMonitor.java
@@ -23,6 +23,7 @@ import org.apache.druid.java.util.emitter.EmittingLogger;
 import org.apache.druid.java.util.emitter.service.ServiceEmitter;
 import org.apache.druid.java.util.emitter.service.ServiceMetricEvent;
 import org.apache.druid.java.util.metrics.AbstractMonitor;
+import org.apache.druid.query.DruidMetrics;
 import org.apache.druid.segment.incremental.InputRowFilterResult;
 import org.apache.druid.segment.incremental.RowIngestionMeters;
 import org.apache.druid.segment.incremental.RowIngestionMetersTotals;
@@ -37,7 +38,6 @@ import java.util.Map;
 public class TaskRealtimeMetricsMonitor extends AbstractMonitor
 {
   private static final EmittingLogger log = new EmittingLogger(TaskRealtimeMetricsMonitor.class);
-  private static final String REASON_DIMENSION = "reason";
 
   private final SegmentGenerationMetrics segmentGenerationMetrics;
   private final RowIngestionMeters rowIngestionMeters;
@@ -76,7 +76,7 @@ public class TaskRealtimeMetricsMonitor extends AbstractMonitor
       if (delta > 0) {
         deltaThrownAwayByReason.put(reason.getReason(), delta);
         emitter.emit(
-            builder.setDimension(REASON_DIMENSION, reason.getReason())
+            builder.setDimension(DruidMetrics.REASON, reason.getReason())
                    .setMetric("ingest/events/thrownAway", delta)
         );
       }
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/task/InputRowFilter.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/task/InputRowFilter.java
@@ -34,7 +34,7 @@ public interface InputRowFilter
   /**
    * Tests whether the given row should be accepted.
    *
-   * @return {@link InputRowFilterResult#ACCEPTED} if the row should be accepted, or another {@link InputRowFilterResult} value if the row should be rejected
+   * @return {@link InputRowFilterResult#ACCEPTED} only if the row should be accepted, otherwise another {@link InputRowFilterResult} value.
    */
   InputRowFilterResult test(InputRow row);
 
@@ -44,7 +44,7 @@ public interface InputRowFilter
    */
   static InputRowFilter fromPredicate(Predicate<InputRow> predicate)
   {
-    return row -> predicate.test(row) ? InputRowFilterResult.ACCEPTED : InputRowFilterResult.FILTERED;
+    return row -> predicate.test(row) ? InputRowFilterResult.ACCEPTED : InputRowFilterResult.CUSTOM_FILTER;
   }
 
   /**
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
@@ -66,7 +66,6 @@ import org.apache.druid.indexing.common.actions.SegmentLockAcquireAction;
 import org.apache.druid.indexing.common.actions.TaskLocks;
 import org.apache.druid.indexing.common.actions.TimeChunkLockAcquireAction;
 import org.apache.druid.indexing.common.stats.TaskRealtimeMetricsMonitor;
-import org.apache.druid.indexing.common.task.InputRowFilter;
 import org.apache.druid.indexing.input.InputRowSchemas;
 import org.apache.druid.indexing.seekablestream.common.OrderedPartitionableRecord;
 import org.apache.druid.indexing.seekablestream.common.OrderedSequenceNumber;
@@ -2147,10 +2146,8 @@ public abstract class SeekableStreamIndexTaskRunner<PartitionIdType, SequenceOff
   }
 
   /**
-   * Returns the filter result for a row.
    * Returns {@link InputRowFilterResult#ACCEPTED} if the row should be accepted,
    * or a rejection reason otherwise.
-   * This method is used as a {@link InputRowFilter} for the {@link StreamChunkParser}.
    */
   InputRowFilterResult ensureRowIsNonNullAndWithinMessageTimeBounds(@Nullable InputRow row)
   {
--- a/indexing-service/src/test/java/org/apache/druid/indexing/common/TaskRealtimeMetricsMonitorTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/common/TaskRealtimeMetricsMonitorTest.java
@@ -125,10 +125,10 @@ public class TaskRealtimeMetricsMonitorTest
     realMeters.incrementThrownAway(InputRowFilterResult.BEFORE_MIN_MESSAGE_TIME);
     realMeters.incrementThrownAway(InputRowFilterResult.BEFORE_MIN_MESSAGE_TIME);
     realMeters.incrementThrownAway(InputRowFilterResult.AFTER_MAX_MESSAGE_TIME);
-    realMeters.incrementThrownAway(InputRowFilterResult.FILTERED);
-    realMeters.incrementThrownAway(InputRowFilterResult.FILTERED);
-    realMeters.incrementThrownAway(InputRowFilterResult.FILTERED);
-    realMeters.incrementThrownAway(InputRowFilterResult.FILTERED);
+    realMeters.incrementThrownAway(InputRowFilterResult.CUSTOM_FILTER);
+    realMeters.incrementThrownAway(InputRowFilterResult.CUSTOM_FILTER);
+    realMeters.incrementThrownAway(InputRowFilterResult.CUSTOM_FILTER);
+    realMeters.incrementThrownAway(InputRowFilterResult.CUSTOM_FILTER);
 
     TaskRealtimeMetricsMonitor monitor = new TaskRealtimeMetricsMonitor(
         segmentGenerationMetrics,
@@ -145,8 +145,8 @@ public class TaskRealtimeMetricsMonitorTest
     }
 
     Assert.assertEquals(Long.valueOf(2), thrownAwayByReason.get("null"));
-    Assert.assertEquals(Long.valueOf(3), thrownAwayByReason.get("beforeMinMessageTime"));
-    Assert.assertEquals(Long.valueOf(1), thrownAwayByReason.get("afterMaxMessageTime"));
+    Assert.assertEquals(Long.valueOf(3), thrownAwayByReason.get("beforeMinimumMessageTime"));
+    Assert.assertEquals(Long.valueOf(1), thrownAwayByReason.get("afterMaximumMessageTime"));
     Assert.assertEquals(Long.valueOf(4), thrownAwayByReason.get("filtered"));
   }
 
@@ -155,7 +155,7 @@ public class TaskRealtimeMetricsMonitorTest
   {
     SimpleRowIngestionMeters realMeters = new SimpleRowIngestionMeters();
     realMeters.incrementThrownAway(InputRowFilterResult.NULL_OR_EMPTY_RECORD);
-    realMeters.incrementThrownAway(InputRowFilterResult.FILTERED);
+    realMeters.incrementThrownAway(InputRowFilterResult.CUSTOM_FILTER);
 
     TaskRealtimeMetricsMonitor monitor = new TaskRealtimeMetricsMonitor(
         segmentGenerationMetrics,
@@ -175,8 +175,8 @@ public class TaskRealtimeMetricsMonitorTest
     Assert.assertEquals(2, thrownAwayByReason.size());
     Assert.assertTrue(thrownAwayByReason.containsKey("null"));
     Assert.assertTrue(thrownAwayByReason.containsKey("filtered"));
-    Assert.assertFalse(thrownAwayByReason.containsKey("beforeMinMessageTime"));
-    Assert.assertFalse(thrownAwayByReason.containsKey("afterMaxMessageTime"));
+    Assert.assertFalse(thrownAwayByReason.containsKey("beforeMinimumMessageTime"));
+    Assert.assertFalse(thrownAwayByReason.containsKey("afterMaximumMessageTime"));
   }
 
   @Test
@@ -204,8 +204,8 @@ public class TaskRealtimeMetricsMonitorTest
 
     emitter.flush();
     realMeters.incrementThrownAway(InputRowFilterResult.NULL_OR_EMPTY_RECORD);
-    realMeters.incrementThrownAway(InputRowFilterResult.FILTERED);
-    realMeters.incrementThrownAway(InputRowFilterResult.FILTERED);
+    realMeters.incrementThrownAway(InputRowFilterResult.CUSTOM_FILTER);
+    realMeters.incrementThrownAway(InputRowFilterResult.CUSTOM_FILTER);
     monitor.doMonitor(emitter);
 
     // Find counts from second call - should be deltas only
--- a/processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
+++ b/processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
@@ -20,8 +20,6 @@
 package org.apache.druid.segment.incremental;
 
 import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
 
 /**
  * Result of filtering an input row during ingestion.
@@ -40,17 +38,17 @@ public enum InputRowFilterResult
   /**
    * The row's timestamp is before the minimum message time (late message rejection).
    */
-  BEFORE_MIN_MESSAGE_TIME("beforeMinMessageTime"),
+  BEFORE_MIN_MESSAGE_TIME("beforeMinimumMessageTime"),
 
   /**
    * The row's timestamp is after the maximum message time (early message rejection).
    */
-  AFTER_MAX_MESSAGE_TIME("afterMaxMessageTime"),
+  AFTER_MAX_MESSAGE_TIME("afterMaximumMessageTime"),
 
   /**
    * The row was filtered out by a transformSpec filter or other row filter.
    */
-  FILTERED("filtered"),
+  CUSTOM_FILTER("filtered"),
 
   /**
    * A backwards-compatible value for tracking filter reasons for ingestion tasks using older Druid versions without filter reason tracking.
@@ -60,7 +58,6 @@ public enum InputRowFilterResult
   private static final InputRowFilterResult[] REJECTED_VALUES = Arrays.stream(InputRowFilterResult.values())
                                                                       .filter(InputRowFilterResult::isRejected)
                                                                       .toArray(InputRowFilterResult[]::new);
-  public static final int NUM_FILTER_RESULT = InputRowFilterResult.values().length;
 
   private final String reason;
 
@@ -87,24 +84,18 @@ public enum InputRowFilterResult
   }
 
   /**
-   * Public utility for building a mutable frequency map over the possible rejection {@link InputRowFilterResult} values.
-   * Keys on {@link InputRowFilterResult#getReason()} rather than the enum name as the latter is more likely to change longer-term.
-   * It is also easier to have stats payload keys match what is being emitted in metrics.
-   */
-  public static Map<String, Long> buildRejectedCounterMap()
-  {
-    final Map<String, Long> result = new HashMap<>();
-    for (InputRowFilterResult reason : InputRowFilterResult.rejectedValues()) {
-      result.put(reason.getReason(), 0L);
-    }
-    return result;
-  }
-
-  /**
-   * Returns {@link InputRowFilterResult} that are not rejection states.
+   * Returns {@link InputRowFilterResult} that are rejection states.
    */
   public static InputRowFilterResult[] rejectedValues()
   {
     return REJECTED_VALUES;
+  }
+
+  /**
+   * Returns total number of {@link InputRowFilterResult} values.
+   */
+  public static int numValues()
+  {
+    return values().length;
   }
 }
--- a/processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
+++ b/processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.druid.common.config.Configs;
 
 import javax.annotation.Nullable;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 
@@ -56,11 +57,11 @@ public class RowIngestionMetersTotals
   }
 
   public RowIngestionMetersTotals(
-      @JsonProperty("processed") long processed,
-      @JsonProperty("processedBytes") long processedBytes,
-      @JsonProperty("processedWithError") long processedWithError,
-      @JsonProperty("thrownAway") long thrownAway,
-      @JsonProperty("unparseable") long unparseable
+      long processed,
+      long processedBytes,
+      long processedWithError,
+      long thrownAway,
+      long unparseable
   )
   {
     this(
@@ -73,11 +74,11 @@ public class RowIngestionMetersTotals
   }
 
   public RowIngestionMetersTotals(
-      @JsonProperty("processed") long processed,
-      @JsonProperty("processedBytes") long processedBytes,
-      @JsonProperty("processedWithError") long processedWithError,
-      @JsonProperty("thrownAwayByReason") Map<String, Long> thrownAwayByReason,
-      @JsonProperty("unparseable") long unparseable
+      long processed,
+      long processedBytes,
+      long processedWithError,
+      Map<String, Long> thrownAwayByReason,
+      long unparseable
   )
   {
     this.processed = processed;
@@ -167,8 +168,10 @@ public class RowIngestionMetersTotals
    */
   private static Map<String, Long> getBackwardsCompatibleThrownAwayByReason(long thrownAway)
   {
-    Map<String, Long> results = InputRowFilterResult.buildRejectedCounterMap();
-    results.put(InputRowFilterResult.UNKNOWN.getReason(), thrownAway);
+    Map<String, Long> results = new HashMap<>();
+    if (thrownAway > 0) {
+      results.put(InputRowFilterResult.UNKNOWN.getReason(), thrownAway);
+    }
     return results;
   }
 }
--- a/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaIndexTaskTest.java
+++ b/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaIndexTaskTest.java
@@ -129,7 +129,6 @@ import org.apache.druid.server.security.Resource;
 import org.apache.druid.server.security.ResourceAction;
 import org.apache.druid.server.security.ResourceType;
 import org.apache.druid.timeline.DataSegment;
-import org.apache.druid.utils.CollectionUtils;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.header.Header;
@@ -1663,8 +1662,7 @@ public class KafkaIndexTaskTest extends SeekableStreamIndexTaskTestBase
     Assert.assertNull(reportData.getErrorMsg());
 
     // Jackson will serde numerics ≤ 32bits as Integers, rather than Longs
-    Map<String, Integer> expectedThrownAwayByReason = CollectionUtils.mapValues(InputRowFilterResult.buildRejectedCounterMap(), Long::intValue);
-    expectedThrownAwayByReason.put(InputRowFilterResult.NULL_OR_EMPTY_RECORD.getReason(), 1);
+    Map<String, Integer> expectedThrownAwayByReason = Map.of(InputRowFilterResult.NULL_OR_EMPTY_RECORD.getReason(), 1);
     Map<String, Object> expectedMetrics = ImmutableMap.of(
         RowIngestionMeters.BUILD_SEGMENTS,
         ImmutableMap.of(
@@ -1752,7 +1750,7 @@ public class KafkaIndexTaskTest extends SeekableStreamIndexTaskTestBase
     Assert.assertNotNull(reportData.getErrorMsg());
 
     // Jackson will serde numerics ≤ 32bits as Integers, rather than Longs
-    Map<String, Integer> expectedThrownAwayByReason = CollectionUtils.mapValues(InputRowFilterResult.buildRejectedCounterMap(), Long::intValue);
+    Map<String, Integer> expectedThrownAwayByReason = Map.of();
     Map<String, Object> expectedMetrics = ImmutableMap.of(
         RowIngestionMeters.BUILD_SEGMENTS,
         ImmutableMap.of(
--- a/extensions-core/kinesis-indexing-service/src/test/java/org/apache/druid/indexing/kinesis/KinesisIndexTaskTest.java
+++ b/extensions-core/kinesis-indexing-service/src/test/java/org/apache/druid/indexing/kinesis/KinesisIndexTaskTest.java
@@ -89,7 +89,6 @@ import org.apache.druid.segment.indexing.DataSchema;
 import org.apache.druid.segment.transform.ExpressionTransform;
 import org.apache.druid.segment.transform.TransformSpec;
 import org.apache.druid.timeline.DataSegment;
-import org.apache.druid.utils.CollectionUtils;
 import org.easymock.EasyMock;
 import org.joda.time.Duration;
 import org.joda.time.Period;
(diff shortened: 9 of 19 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -2144,26 +2146,36 @@ private void refreshMinMaxMessageTime()
     );
   }
 
-  public boolean withinMinMaxRecordTime(final InputRow row)
+  /**
+   * Returns the rejection reason for a row, or null if the row should be accepted.
+   * This method is used as a {@link RowFilter} for the {@link StreamChunkParser}.
+   */
+  @Nullable
+  ThrownAwayReason getRowRejectionReason(final InputRow row)
```
abhishekrb19:
```suggestion
  ThrownAwayReason getRowRejectionReason(@Nullable final InputRow row)
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -2144,26 +2146,36 @@ private void refreshMinMaxMessageTime()
...
             minMessageTime
         );
-      } else if (afterMaximumMessageTime) {
+      }
+      return ThrownAwayReason.BEFORE_MIN_MESSAGE_TIME;
+    }
+    if (maxMessageTime.isBefore(row.getTimestamp())) {
+      if (log.isDebugEnabled()) {
         log.debug(
             "CurrentTimeStamp[%s] is after MaximumMessageTime[%s]",
```
abhishekrb19:
```suggestion
            "CurrentTimeStamp[%s] is after maximumMessageTime[%s]",
```

docs/operations/metrics.md
```diff
@@ -270,7 +270,7 @@ batch ingestion emit the following metrics. These metrics are deltas for each em
 |`ingest/events/processed`|Number of events processed per emission period.|`dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Equal to the number of events per emission period.|
 |`ingest/events/processedWithError`|Number of events processed with some partial errors per emission period. Events processed with partial errors are counted towards both this metric and `ingest/events/processed`.|`dataSource`, `taskId`, `taskType`, `groupId`, `tags`|0|
 |`ingest/events/unparseable`|Number of events rejected because the events are unparseable.|`dataSource`, `taskId`, `taskType`, `groupId`, `tags`|0|
-|`ingest/events/thrownAway`|Number of events rejected because they are null, or filtered by `transformSpec`, or outside one of `lateMessageRejectionPeriod`, `earlyMessageRejectionPeriod`.|`dataSource`, `taskId`, `taskType`, `groupId`, `tags`|0|
+|`ingest/events/thrownAway`|Number of events rejected because they are null, or filtered by `transformSpec`, or outside one of `lateMessageRejectionPeriod`, `earlyMessageRejectionPeriod`. The `reason` dimension indicates why the event was thrown away.|`dataSource`, `taskId`, `taskType`, `groupId`, `tags`, `reason`|0|
```
kfaraz, line 273: I would advise using the `description` dimension instead, since it is pretty much a free form dimension already baked into Druid, and suits this use case well.

indexing-service/src/main/java/org/apache/druid/indexing/common/stats/TaskRealtimeMetricsMonitor.java
```diff
@@ -25,21 +25,27 @@
...
   private static final EmittingLogger log = new EmittingLogger(TaskRealtimeMetricsMonitor.class);
+  private static final String REASON_DIMENSION = "reason";
 
   private final SegmentGenerationMetrics segmentGenerationMetrics;
   private final RowIngestionMeters rowIngestionMeters;
   private final ServiceMetricEvent.Builder builder;
 
   private SegmentGenerationMetrics previousSegmentGenerationMetrics;
   private RowIngestionMetersTotals previousRowIngestionMetersTotals;
+  private Map<ThrownAwayReason, Long> previousThrownAwayByReason;
```
kfaraz: I feel it would be cleaner to have this map live inside `RowIngestionMetersTotals` itself, since it already captures all the aggregates, it might as well contain the numbers by category.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/RowFilter.java
```diff
@@ -0,0 +1,76 @@
...
+
+import javax.annotation.Nullable;
+import java.util.function.Predicate;
+
+/**
+ * A filter for input rows during ingestion that can report the reason for rejection or null for acceptance.
+ * This is similar to {@link Predicate} but returns the rejection reason instead of just a boolean.
+ */
+@FunctionalInterface
+public interface RowFilter
```
kfaraz:
```suggestion
public interface InputRowFilter
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/RowFilter.java
```diff
@@ -0,0 +1,76 @@
...
+   */
+  static RowFilter fromPredicate(Predicate<InputRow> predicate)
+  {
+    return row -> predicate.test(row) ? null : ThrownAwayReason.FILTERED;
+  }
+
+  /**
+   * Fully-permissive {@link RowFilter} used mainly for tests.
+   */
+  static RowFilter allow()
```
kfaraz:
```suggestion
  static RowFilter allowAll()
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/RowFilter.java
```diff
@@ -0,0 +1,76 @@
...
+public interface RowFilter
+{
+  /**
+   * Tests whether the given row should be accepted.
+   *
+   * @param row the input row to test
+   * @return null if the row should be accepted, or the {@link ThrownAwayReason} if the row should be rejected
+   */
+  @Nullable
+  ThrownAwayReason test(InputRow row);
```
kfaraz: Would this be cleaner?


```suggestion
  RowFilterTestResult test(InputRow row);
```

where `RowFilterTestResult` contains a boolean `isRejected` and a `reason`.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -2144,26 +2146,36 @@ private void refreshMinMaxMessageTime()
...
   {
-    final boolean beforeMinimumMessageTime = minMessageTime.isAfter(row.getTimestamp());
-    final boolean afterMaximumMessageTime = maxMessageTime.isBefore(row.getTimestamp());
-
-    if (log.isDebugEnabled()) {
-      if (beforeMinimumMessageTime) {
+    if (row == null) {
+      return ThrownAwayReason.NULL;
+    }
+    if (minMessageTime.isAfter(row.getTimestamp())) {
```
kfaraz: Cleaner to chain these using `else`.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
kfaraz: The method name should reflect when it rejects something.


```suggestion
  ThrownAwayReason ensureRowIsNonNullAndWithinMessageTimeBounds(final InputRow row)
```

processing/src/main/java/org/apache/druid/segment/incremental/ThrownAwayReason.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Reasons why an input row may be thrown away during ingestion.
+ */
+public enum ThrownAwayReason
+{
+  /**
+   * The row was null or the input record was empty.
+   */
+  NULL,
```
kfaraz:
```suggestion
  NULL_OR_EMPTY_RECORD,
```

indexing-service/src/test/java/org/apache/druid/indexing/common/task/RowFilterTest.java
```diff
@@ -0,0 +1,122 @@
...
+import org.junit.Test;
+
+import java.util.List;
+
+public class RowFilterTest
+{
+  private static final List<String> DIMENSIONS = ImmutableList.of("dim1");
+
+  @Test
+  public void testFromPredicateAccept()
```
kfaraz: Nit: Different style which makes the tested method name more apparent, but it is okay if you prefer the current style itself.


```suggestion
  public void test_fromPredicate_whichAllowsAll()
```

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMeters.java
```diff
@@ -73,7 +73,16 @@ default long getProcessedBytes()
...
+
+  /**
+   * Increments the thrown away counter for the specified reason.
+   */
+  void incrementThrownAway(ThrownAwayReason reason);
+
+  /**
+   * Returns the count of thrown away events for each reason.
+   */
+  Map<ThrownAwayReason, Long> getThrownAwayByReason();
```
kfaraz: Is this method used anywhere directly?
I feel it should just be included inside the result of `getTotals()` and need not be exposed as a separate map.

processing/src/main/java/org/apache/druid/segment/incremental/ThrownAwayReason.java
```diff
@@ -0,0 +1,64 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.segment.incremental;
+
+/**
+ * Reasons why an input row may be thrown away during ingestion.
+ */
+public enum ThrownAwayReason
```
kfaraz: Should we align this enum name with `InputRowFilter`, e.g. `RowFilterReason`?

processing/src/main/java/org/apache/druid/segment/incremental/ThrownAwayReason.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * The row's timestamp is after the maximum message time (early message rejection).
+   */
+  AFTER_MAX_MESSAGE_TIME,
+
+  /**
+   * The row was filtered out by a transformSpec filter or other row filter.
+   */
+  FILTERED;
```
kfaraz:
```suggestion
  FILTERED("Filtered by transform spec etc");
```

processing/src/main/java/org/apache/druid/segment/incremental/ThrownAwayReason.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * The row's timestamp is before the minimum message time (late message rejection).
+   */
+  BEFORE_MIN_MESSAGE_TIME,
+
+  /**
+   * The row's timestamp is after the maximum message time (early message rejection).
+   */
+  AFTER_MAX_MESSAGE_TIME,
```
kfaraz:
```suggestion
  AFTER_MAX_MESSAGE_TIME("Record after max message time"),
```

processing/src/main/java/org/apache/druid/segment/incremental/ThrownAwayReason.java
```diff
@@ -0,0 +1,64 @@
...
+{
+  /**
+   * The row was null or the input record was empty.
+   */
+  NULL,
+
+  /**
+   * The row's timestamp is before the minimum message time (late message rejection).
+   */
+  BEFORE_MIN_MESSAGE_TIME,
```
kfaraz:
```suggestion
  BEFORE_MIN_MESSAGE_TIME("Record before min message time"),
```

processing/src/main/java/org/apache/druid/segment/incremental/ThrownAwayReason.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Pre-computed metric dimension values, indexed by ordinal.
+   */
+  private static final String[] METRIC_VALUES = {
+      "null",
+      "beforeMinMessageTime",
+      "afterMaxMessageTime",
+      "filtered"
+  };
```
kfaraz: The dimension name should be passed inside the constructor of each enum category.

processing/src/main/java/org/apache/druid/segment/incremental/ThrownAwayReason.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Returns the value to be used as the dimension value in metrics.
+   */
+  public String getMetricValue()
```
kfaraz:
```suggestion
  public String emittedDimensionValue()
```

docs/operations/metrics.md
jtuglu1: I considered this, but ultimately chose against it in the spirit of https://netflix.github.io/atlas-docs/concepts/naming/#summary. I feel like `description` ends up being a "kitchen sink" string which you need to regex matching on to do any kind of programmatic alerting; I'd actually like to move away from using this excessively. 

IMO `description` should be used if we're not 100% sure what the value is (e.g. an exception is thrown) if at all. Even then, if a user wants a full trace, they can turn on parse exception logs in the task to have higher fidelity insight into what went wrong. IMO, metrics should be reserved for compact slice/dice metrics/dimensions which have fixed length/pre-defined values; anything else should be in an alert or log.

In our case, we have a fixed set of possible cases that  we can assign proper values. This makes things like building alerts, etc. much easier. Thoughts?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
jtuglu1: SGTM – I was just following pattern of prior code to keep diff easier to reason about.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/RowFilter.java
jtuglu1: I made sure we return a primitive-compiled enum for a reason (same as the prior stack-allocated boolean) – I would prefer to not do object allocation (heap) on the hotpath. This code is called in a tight-loop in latency-critical places like streaming ingestion. The more allocs we do in that loop slows things down quite a bit. IMO as long as the contract is clear in the `RowFilter`, we can keep this a zero-cost abstraction over the current setup while getting better observability.

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMeters.java
jtuglu1: It's used in `TaskRealtimeMetricsMonitor::doMonitor`.

processing/src/main/java/org/apache/druid/segment/incremental/ThrownAwayReason.java
jtuglu1: See comment about `description`.

docs/operations/metrics.md
kfaraz: Yeah, I guess that sounds reasonable from an alerting point of view.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/RowFilter.java
kfaraz: That makes sense, thanks!

indexing-service/src/test/java/org/apache/druid/indexing/common/task/RowFilterTest.java
jtuglu1: We should pick a test naming convention style and commit it to the codestyle thing. Is the logic behind the suggested change formalizable?

kfaraz: Yeah, that's a good idea! I think it should be possible to do it via checkstyle or similar.

indexing-service/src/main/java/org/apache/druid/indexing/common/stats/TaskRealtimeMetricsMonitor.java
jtuglu1: Do we want to be passing this in the serialized `RowIngestionMetersTotals` payload back-and-forth? AFAIK this is used to populate UI, etc. For backwards compatibility, I can keep both `thrownAway` and `thrownAwayByReason` as parameters to a `RowIngestionMetersTotals`, but just `@JsonIgnore` it for now. If needed, it can be added to the payload.

kfaraz: It would be nice to include in the serialized payload too, since the map is going to be small anyway and the info might be handy while looking at task live reports.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
abhishekrb19:
```suggestion
  ThrownAwayReason getRowRejectionReason(@Nullable final InputRow row)
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -2144,26 +2146,36 @@ private void refreshMinMaxMessageTime()
...
+      if (log.isDebugEnabled()) {
         log.debug(
             "CurrentTimeStamp[%s] is before MinimumMessageTime[%s]",
```
abhishekrb19: nit - while at it:
```suggestion
            "CurrentTimeStamp[%s] is before minimumMessageTime[%s]",
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
abhishekrb19:
```suggestion
            "CurrentTimeStamp[%s] is after maximumMessageTime[%s]",
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -2144,26 +2146,36 @@ private void refreshMinMaxMessageTime()
...
         log.debug(
             "CurrentTimeStamp[%s] is after MaximumMessageTime[%s]",
             row.getTimestamp(),
             maxMessageTime
         );
       }
+      return ThrownAwayReason.AFTER_MAX_MESSAGE_TIME;
     }
-    return !beforeMinimumMessageTime && !afterMaximumMessageTime;
+    return null;
```
abhishekrb19: I wonder if it'd be cleaner to also capture "Accepted" as an enum state rather than null. Perhaps an enum like: RowFilterResult.Accepted, RowFilterResult.Null, RowFilterResult.Before_MaxTime, etc.

Then the callers can test and filter out the specific enum predicate to get the ThrownAway dimension as needed. What do you think?

jtuglu1: Sure.

Edit: actually, I'd like to leave this as-is. I don't see a reason to use this `RowFilterResult.Accepted` anywhere else except the conditional to check whether it is indeed accepted/rejected. We already have metrics for ingest row counters.

jtuglu1: Actually, I'd like to leave this as-is. I don't see a reason to use this `RowFilterResult.Accepted` anywhere else other than to actually check whether the result was accepted/denied (for incrementing a thrown away counter). That's really no better than checking against `null`. We wouldn't be tracking counters for this enum value as we already have metrics for ingest row counters. It's not a breaking change on user-side, so if this turns out to be the wrong approach, we can always switch it out under-the-hood (the change is simple).

abhishekrb19: Sounds good to me
EDIT: there's also one more usage of Accepted as noted in this [comment](https://github.com/apache/druid/pull/18855#discussion_r2636537612)

indexing-service/src/test/java/org/apache/druid/indexing/common/TaskRealtimeMetricsMonitorTest.java
```diff
@@ -127,6 +131,151 @@ public void testMessageGapAggStats()
...
+    ServiceEmitter captureEmitter = Mockito.mock(ServiceEmitter.class);
+    Mockito.doCallRealMethod().when(captureEmitter).emit(ArgumentMatchers.any(ServiceEventBuilder.class));
+    Mockito
+        .doAnswer(invocation -> {
+          ServiceMetricEvent e = invocation.getArgument(0);
+          allEmittedEvents.add(e);
+          return null;
+        })
+        .when(captureEmitter).emit(ArgumentMatchers.any(Event.class));
+
```
abhishekrb19: Could we use `StubServiceEmitter` instead of a mocked `ServiceEmitter` for these new tests? It should simplify this test code & make it more readable. We can migrate the existing tests that use mocked emitter separately.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/RowFilter.java
abhishekrb19: The API contract is slightly strange in that `test()` returns `ThrownAwayReason`. Instead of the boolean in `RowFilterTestResult`, if we did make it a broader enum to encapsulate Accepted as a state, would that satisfy most requirements? I don't feel too strongly about this suggestion, but there are multiple places where we implicitly interpret null as "Accepted", so it's worth considering to just operate on a single unified object

jtuglu1: Sure – I can update this.

jtuglu1: It's a bit awkward to keep building frequency maps that need to keep excluding the `ACCEPTED` state. I understand why this might seem "cleaner," but from the actual implementation IMO it's more confusing.

jtuglu1: I've added and we can discuss after.

indexing-service/src/main/java/org/apache/druid/indexing/common/stats/DropwizardRowIngestionMeters.java
```diff
@@ -100,13 +104,27 @@ public void incrementUnparseable()
...
+  public void incrementThrownAway(InputRowFilterResult reason)
   {
-    thrownAway.mark();
+    thrownAwayByReason[reason.ordinal()].mark();
+  }
+
+  @Override
+  public Map<String, Long> getThrownAwayByReason()
+  {
+    Map<String, Long> result = InputRowFilterResult.buildRejectedCounterMap();
```
kfaraz: Why can't we just initialize this to an empty hash map? It is odd to have the map be constructed elsewhere. Is it just so that we have 0 values for the missing reasons?

indexing-service/src/main/java/org/apache/druid/indexing/common/stats/TaskRealtimeMetricsMonitor.java
```diff
@@ -23,16 +23,21 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Emits metrics from {@link SegmentGenerationMetrics} and {@link RowIngestionMeters}.
  */
 public class TaskRealtimeMetricsMonitor extends AbstractMonitor
 {
   private static final EmittingLogger log = new EmittingLogger(TaskRealtimeMetricsMonitor.class);
+  private static final String REASON_DIMENSION = "reason";
```
kfaraz: Move this to `DruidMetrics`?

indexing-service/src/main/java/org/apache/druid/indexing/common/task/InputRowFilter.java
```diff
@@ -0,0 +1,73 @@
...
+ * A filter for input rows during ingestion that returns a {@link InputRowFilterResult} per row.
+ * This is similar to {@link Predicate} but returns an {@link InputRowFilterResult} instead of a boolean.
+ */
+@FunctionalInterface
+public interface InputRowFilter
+{
+  /**
+   * Tests whether the given row should be accepted.
+   *
+   * @return {@link InputRowFilterResult#ACCEPTED} if the row should be accepted, or another {@link InputRowFilterResult} value if the row should be rejected
```
kfaraz:
```suggestion
   * @return {@link InputRowFilterResult#ACCEPTED} only if the row should be accepted, otherwise another {@link InputRowFilterResult} value.
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -2144,26 +2146,35 @@ private void refreshMinMaxMessageTime()
     );
   }
 
-  public boolean withinMinMaxRecordTime(final InputRow row)
+  /**
+   * Returns the filter result for a row.
+   * Returns {@link InputRowFilterResult#ACCEPTED} if the row should be accepted,
+   * or a rejection reason otherwise.
+   * This method is used as a {@link InputRowFilter} for the {@link StreamChunkParser}.
```
kfaraz: I think this can be shortened, as the other lines don't add any extra info.
```suggestion
   * Returns {@link InputRowFilterResult#ACCEPTED} if the row should be accepted,
   * or a rejection reason otherwise.
```

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
```diff
@@ -0,0 +1,110 @@
...
+   * The row passed the filter and should be processed.
+   */
+  ACCEPTED("accepted"),
+  /**
+   * The row was null or the input record was empty.
+   */
+  NULL_OR_EMPTY_RECORD("null"),
+
+  /**
+   * The row's timestamp is before the minimum message time (late message rejection).
```
kfaraz, line 39:
```suggestion
   * The timestamp of the row is before the minimum message time (late message rejection).
```

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * The row's timestamp is after the maximum message time (early message rejection).
+   */
+  AFTER_MAX_MESSAGE_TIME("afterMaxMessageTime"),
+
+  /**
+   * The row was filtered out by a transformSpec filter or other row filter.
+   */
+  FILTERED("filtered"),
```
kfaraz:
```suggestion
  REJECTED_BY_FILTER("rejectedByFilter"),
```

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
```diff
@@ -0,0 +1,110 @@
...
+  {
+    return this != ACCEPTED;
+  }
+
+  /**
+   * Public utility for building a mutable frequency map over the possible rejection {@link InputRowFilterResult} values.
+   * Keys on {@link InputRowFilterResult#getReason()} rather than the enum name as the latter is more likely to change longer-term.
+   * It is also easier to have stats payload keys match what is being emitted in metrics.
+   */
+  public static Map<String, Long> buildRejectedCounterMap()
```
kfaraz: I don't think we need this utility method.

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
```diff
@@ -38,13 +42,49 @@ public RowIngestionMetersTotals(
...
+        processed,
+        processedBytes,
+        processedWithError,
+        Configs.valueOrDefault(thrownAwayByReason, getBackwardsCompatibleThrownAwayByReason(thrownAway)),
+        unparseable
+    );
+  }
+
+  public RowIngestionMetersTotals(
+      @JsonProperty("processed") long processed,
```
kfaraz: Since this is not the `@JsonCreator` constructor, we don't need the `@JsonProperty` annotations here.

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
```diff
@@ -38,13 +42,49 @@ public RowIngestionMetersTotals(
...
+    this(
+        processed,
+        processedBytes,
+        processedWithError,
+        getBackwardsCompatibleThrownAwayByReason(thrownAway),
+        unparseable
+    );
+  }
+
+  public RowIngestionMetersTotals(
```
kfaraz, line 76: Do we need these extra constructors?

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
```diff
@@ -109,7 +156,19 @@ public String toString()
...
+
+  /**
+   * For backwards compatibility, key by {@link InputRowFilterResult} in case of lack of thrownAwayByReason input during rolling Druid upgrades.
+   * This can occur when tasks running on older Druid versions return ingest statistic payloads to an overlord running on a newer Druid version.
+   */
+  private static Map<String, Long> getBackwardsCompatibleThrownAwayByReason(long thrownAway)
+  {
+    Map<String, Long> results = InputRowFilterResult.buildRejectedCounterMap();
+    results.put(InputRowFilterResult.UNKNOWN.getReason(), thrownAway);
+    return results;
```
kfaraz:
```suggestion
    return Map.of(InputRowFilterResult.UNKNOWN.getReason(), thrownAway);
```

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
jtuglu1: I opted for this because I liked the contract of reporting all stats to the emitter, irrespective if they are 0 or not. This allows the end emitter to do what it wants with all values, rather than fight with logic on the aggregation side or treat omission of key-value pairs as 0s. Moreover, it prevents repetitive code patterns to initialize a frequency map with possible thrown away values. This ensures we never have a case where consumers of these metrics (logs, emitters, callers of APIs, etc.) see some subset of the available throw away reasons. Let me know what you think.

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
jtuglu1: I think REJECTED_BY is implicit in the `isRejected()`. I've renamed to `CUSTOM_FILTER`.

indexing-service/src/main/java/org/apache/druid/indexing/common/stats/DropwizardRowIngestionMeters.java
jtuglu1: See https://github.com/apache/druid/pull/18855#discussion_r2639933353.

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
jtuglu1: This would not work as the code expects a frequency map containing the full set of rejected reasons. As I mentioned above, I want to *always* emit a frequency map that covers all possible keys. IMO, omission of a key isn't as clear as reporting the `key:0` (e.g. did the event not exist, did the filter reason not exist, did the task not check for that reason?) and is easier for clients to build logic around. Additionally, this is already the pattern with the rest of the counters (processed, etc.) where they are still returned (but maybe not emitted) if they are 0. Open to suggestions here, but I think this fits with the current accounting contracts.

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
jtuglu1: Not necessarily in the future, but I thought it made things cleaner from client perspective. Reasons for each:
1. The main JSON ctor (needs both nullable `thrownAwayByReason` and `thrownAway` for backwards compatibility).
2. Ctor for internal test usage that just uses thrownAway for backwards compatibility. IMO was cleaner than passing `null` for `thrownAwayByReason` to #1.
3. New ctor (what +2 versions of Druid can default to) where `thrownAwayReason` is only key.

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
```diff
@@ -0,0 +1,110 @@
...
+  {
+    final Map<String, Long> result = new HashMap<>();
+    for (InputRowFilterResult reason : InputRowFilterResult.rejectedValues()) {
+      result.put(reason.getReason(), 0L);
+    }
+    return result;
+  }
+
+  /**
+   * Returns {@link InputRowFilterResult} that are not rejection states.
```
abhishekrb19:
```suggestion
   * Returns {@link InputRowFilterResult} that are rejection states.
```

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
```diff
@@ -0,0 +1,110 @@
...
+  ACCEPTED("accepted"),
+  /**
+   * The row was null or the input record was empty.
+   */
+  NULL_OR_EMPTY_RECORD("null"),
+
+  /**
+   * The row's timestamp is before the minimum message time (late message rejection).
+   */
+  BEFORE_MIN_MESSAGE_TIME("beforeMinMessageTime"),
```
abhishekrb19: Should we call these `afterMaximumMessageTime` and `beforeMinimumMessageTime` consistent with the corresponding `ioConfig` properties?

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
abhishekrb19: nit: `filteredByTransformSpec` to be explicit

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
```diff
@@ -0,0 +1,110 @@
...
+   * Keys on {@link InputRowFilterResult#getReason()} rather than the enum name as the latter is more likely to change longer-term.
+   * It is also easier to have stats payload keys match what is being emitted in metrics.
+   */
+  public static Map<String, Long> buildRejectedCounterMap()
+  {
+    final Map<String, Long> result = new HashMap<>();
+    for (InputRowFilterResult reason : InputRowFilterResult.rejectedValues()) {
+      result.put(reason.getReason(), 0L);
+    }
+    return result;
```
abhishekrb19: Similar to `REJECTED_VALUES`, I think this empty map could be initialized once and returned in this method (no need to lazily build every time this method is invoked)

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * A backwards-compatible value for tracking filter reasons for ingestion tasks using older Druid versions without filter reason tracking.
+   */
+  UNKNOWN("unknown");
+
+  private static final InputRowFilterResult[] REJECTED_VALUES = Arrays.stream(InputRowFilterResult.values())
+                                                                      .filter(InputRowFilterResult::isRejected)
+                                                                      .toArray(InputRowFilterResult[]::new);
+  public static final int NUM_FILTER_RESULT = InputRowFilterResult.values().length;
```
abhishekrb19: nit: consider adding a static getter for this rather than accessing the variable directly

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
jtuglu1: Ah – I've left it as such because it is not necessarily the transformSpec that filters out the row. An example is [here](https://github.com/apache/druid/blob/master/indexing-service/src/main/java/org/apache/druid/indexing/common/task/AbstractBatchIndexTask.java#L260-L268). The filter is using granularity spec's intervals to filter out by time (it's not transform). I'd prefer to keep it as `filtered`.

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
jtuglu1: Hmm. The method is a builder pattern to ensure all possible thrown away values have an entry in the map (again using value `0` instead of key omission to signify lack of events). Returning a reference to a static map doesn't make much sense IMO because callers would need to make a copy anyways (they're using to assign their counter values). I'd like to keep as-is as this is mainly a way for code that needs to emit a frequency map to not have to worry about including the correct keys, etc.

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
kfaraz: Hmm. Even if we wanted to always send back a map containing all the reasons, it would be a concern of the API layer. It doesn't seem appropriate to initialize the map here and then populate it elsewhere.

Let's remove this for now. It is okay to duplicate the logic if required.

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
kfaraz: Sure, sounds good.

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
jtuglu1: This is used pretty extensively in around ~17 files  (3 code, 14 test) – are we sure we don't want to use this?

processing/src/main/java/org/apache/druid/segment/incremental/RowIngestionMetersTotals.java
kfaraz: In the future, we may have a large number of reasons. It wouldn't make sense to have a zero entry for all of them in the payload or to emit a zero metric for all of them. It would just bloat up the payload.

> Additionally, this is already the pattern with the rest of the counters (processed, etc.)

It makes sense for `processed` to always be reported (even if zero), since it indicates that processing was actually attempted. It is a "success" metric in the sense that the absence of the metric also indicates some erratic behaviour. This is generally true for gauge metrics too.

`thrownAway` on the other hand is more of an "error" count metric that need not be emitted if nothing was thrown away. This metric is typically useful only when we are looking for the reason of why certain records didn't make it into the system, or _did_ make it into the system even when we expected them to get discarded.

In both the cases, the absence of the metric is equivalent to a zero value as both indicate that nothing was thrown away. Downstream applications could treat absence/null the same as zero in this case.

> (e.g. did the event not exist, did the filter reason not exist, did the task not check for that reason?)

Yes, the `thrownAway` metric is not equipped to answer these questions and should not try to either. By populating the values for all possible reasons as 0, we are still not answering these questions.
Same as any other error metric, the absence of the metric should simply indicate that the error did not occur.
Any further reason should be determined with the help of other success metrics (`processed` etc).

processing/src/main/java/org/apache/druid/segment/incremental/InputRowFilterResult.java
kfaraz: You can add a utility method in the test classes, to allow reuse.
I feel the usages in the main code should go away anyway, since the default zero value map should be avoided.

PR #18819, opened 2025-12-05
Introduce cost-based tasks autoscaler for streaming ingestion

Fly-Style:
### Overview

Implements a cost-based autoscaling algorithm for seekable stream supervisor tasks that optimizes task count by balancing lag reduction against resource efficiency.

Note: this patch doesn't support autoscaling (down) during task rollover. Temporarily, it scales down in the same manner as scales up.
Introduces `WeightedCostFunction` for cost-based autoscaling decisions. The function computes a cost score (in seconds) for each candidate task count, balancing lag recovery time against idle resource waste.

### Cost Formula
```
totalCost = lagWeight × lagRecoveryTime + idleWeight × idlenessCost
```

- **lagRecoveryTime** = `aggregateLag / (taskCount × avgProcessingRate)` — time to clear backlog
- **idlenessCost** = `taskCount × taskDuration × predictedIdleRatio` — wasted compute time

### Idle Prediction Model

Uses capacity-based linear scaling:
```
predictedIdle = 1 - (1 - currentIdle) / (proposedTasks / currentTasks)
```

More tasks → more idle per task; fewer tasks → busier tasks.

### Ideal Idle Range

Defines optimal utilization as idle ratio within [0.2, 0.6]:
- Below 0.2: overloaded → scale up
- Within range: optimal → no action
- Above 0.6: underutilized → scale down

### Conservative Cold Start Behavior

When processing rate is unavailable (cold start, new tasks):
- Current task count: cost = 0.01 (allowed)
- Any scaling: cost = +∞ (prohibited)

This prevents scaling decisions based on incomplete data.

Additionally, we add reading `poll-idle ratio-avg` from `/rowStats` task endpoint. 

<hr>

This PR has:

- [x] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [x] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [x] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [x] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [x] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
[...]

PR #18860, opened 2025-12-19
Inject new task count calculation during the rollover

Fly-Style:
As a follow-up PR for #18819, the patch fixes temporal behaviour when scale down happens in the same manner, as scaleup, by injecting the new task count calculation logic during the task rollover time,

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
@@ -3444,7 +3444,7 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
       activelyReadingTaskGroups.remove(groupId);
     }
 
-    maybeScaleUpDuringTaskRollover();
+    maybeScaleDuringTaskRollover();
   }
 
   /**
@@ -3453,7 +3453,8 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
    * This method is invoked to determine whether a task count adjustment is needed
    * during a task rollover based on the recommendations from the task auto-scaler.
  */
-  private void maybeScaleUpDuringTaskRollover()
+  @VisibleForTesting
+  void maybeScaleDuringTaskRollover()
   {
     if (taskAutoScaler != null && activelyReadingTaskGroups.isEmpty()) {
       int rolloverTaskCount = taskAutoScaler.computeTaskCountForRollover();
--- a/indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorScaleDuringTaskRolloverTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorScaleDuringTaskRolloverTest.java
@@ -1,0 +1,612 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.indexing.seekablestream.supervisor;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.apache.druid.data.input.impl.ByteEntity;
+import org.apache.druid.data.input.impl.DimensionSchema;
+import org.apache.druid.data.input.impl.JsonInputFormat;
+import org.apache.druid.data.input.impl.StringDimensionSchema;
+import org.apache.druid.data.input.impl.TimestampSpec;
+import org.apache.druid.indexer.granularity.UniformGranularitySpec;
+import org.apache.druid.indexing.common.task.Task;
+import org.apache.druid.indexing.overlord.DataSourceMetadata;
+import org.apache.druid.indexing.overlord.IndexerMetadataStorageCoordinator;
+import org.apache.druid.indexing.overlord.TaskMaster;
+import org.apache.druid.indexing.overlord.TaskStorage;
+import org.apache.druid.indexing.overlord.supervisor.SupervisorStateManagerConfig;
+import org.apache.druid.indexing.overlord.supervisor.autoscaler.LagStats;
+import org.apache.druid.indexing.overlord.supervisor.autoscaler.SupervisorTaskAutoScaler;
+import org.apache.druid.indexing.seekablestream.SeekableStreamDataSourceMetadata;
+import org.apache.druid.indexing.seekablestream.SeekableStreamEndSequenceNumbers;
+import org.apache.druid.indexing.seekablestream.SeekableStreamIndexTask;
+import org.apache.druid.indexing.seekablestream.SeekableStreamIndexTaskClientFactory;
+import org.apache.druid.indexing.seekablestream.SeekableStreamIndexTaskIOConfig;
+import org.apache.druid.indexing.seekablestream.SeekableStreamIndexTaskTuningConfig;
+import org.apache.druid.indexing.seekablestream.SeekableStreamStartSequenceNumbers;
+import org.apache.druid.indexing.seekablestream.common.OrderedSequenceNumber;
+import org.apache.druid.indexing.seekablestream.common.RecordSupplier;
+import org.apache.druid.indexing.seekablestream.supervisor.autoscaler.CostBasedAutoScalerConfig;
+import org.apache.druid.java.util.common.granularity.Granularities;
+import org.apache.druid.java.util.common.parsers.JSONPathSpec;
+import org.apache.druid.java.util.emitter.service.ServiceEmitter;
+import org.apache.druid.java.util.metrics.StubServiceEmitter;
+import org.apache.druid.query.aggregation.CountAggregatorFactory;
+import org.apache.druid.segment.TestHelper;
+import org.apache.druid.segment.incremental.RowIngestionMetersFactory;
+import org.apache.druid.segment.indexing.DataSchema;
+import org.easymock.EasyMock;
+import org.easymock.EasyMockSupport;
+import org.joda.time.DateTime;
+import org.joda.time.Duration;
+import org.joda.time.Period;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import javax.annotation.Nullable;
+import java.io.File;
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.concurrent.ScheduledExecutorService;
+
+public class SeekableStreamSupervisorScaleDuringTaskRolloverTest extends EasyMockSupport
+{
+  private static final ObjectMapper OBJECT_MAPPER = TestHelper.makeJsonMapper();
+  private static final String STREAM = "stream";
+  private static final String DATASOURCE = "testDS";
+  private static final String SUPERVISOR = "supervisor";
+  private static final int DEFAULT_TASK_COUNT = 10;
+
+  private TaskStorage taskStorage;
+  private TaskMaster taskMaster;
+  private IndexerMetadataStorageCoordinator indexerMetadataStorageCoordinator;
+  private ServiceEmitter emitter;
+  private RowIngestionMetersFactory rowIngestionMetersFactory;
+  private SeekableStreamIndexTaskClientFactory taskClientFactory;
+  private SeekableStreamSupervisorSpec spec;
+  private SupervisorStateManagerConfig supervisorConfig;
+
+  @Before
+  public void setUp()
+  {
+    taskStorage = EasyMock.mock(TaskStorage.class);
+    taskMaster = EasyMock.mock(TaskMaster.class);
+    indexerMetadataStorageCoordinator = EasyMock.mock(IndexerMetadataStorageCoordinator.class);
+    emitter = new StubServiceEmitter();
+    rowIngestionMetersFactory = EasyMock.mock(RowIngestionMetersFactory.class);
+    taskClientFactory = EasyMock.mock(SeekableStreamIndexTaskClientFactory.class);
+    spec = EasyMock.mock(SeekableStreamSupervisorSpec.class);
+    supervisorConfig = new SupervisorStateManagerConfig();
+
+    // Common taskMaster setup - used by all tests
+    EasyMock.expect(taskMaster.getTaskRunner()).andReturn(Optional.absent()).anyTimes();
+    EasyMock.expect(taskMaster.getSupervisorManager()).andReturn(Optional.absent()).anyTimes();
+    EasyMock.replay(taskMaster);
+  }
+
+  @Test
+  public void test_maybeScaleDuringTaskRollover_noAutoScaler_doesNotScale()
+  {
+    // Given
+    setupSpecExpectations(getIOConfigWithoutAutoScaler(5));
+    EasyMock.expect(spec.createAutoscaler(EasyMock.anyObject())).andReturn(null).anyTimes();
+    EasyMock.replay(spec);
+
+    TestSeekableStreamSupervisor supervisor = new TestSeekableStreamSupervisor(10);
+    supervisor.start();
+
+    int beforeTaskCount = supervisor.getIoConfig().getTaskCount();
+
+    // When
+    supervisor.maybeScaleDuringTaskRollover();
+
+    // Then
+    Assert.assertEquals(
+        "Task count should not change when taskAutoScaler is null",
+        beforeTaskCount,
+        (int) supervisor.getIoConfig().getTaskCount()
+    );
+  }
+
+  @Test
+  public void test_maybeScaleDuringTaskRollover_rolloverCountNonPositive_doesNotScale()
+  {
+    // Given
+    setupSpecExpectations(getIOConfigWithCostBasedAutoScaler());
+    EasyMock.expect(spec.createAutoscaler(EasyMock.anyObject()))
+            .andReturn(createMockAutoScaler(-1))
+            .anyTimes();
+    EasyMock.replay(spec);
+
+    TestSeekableStreamSupervisor supervisor = new TestSeekableStreamSupervisor(100);
+    supervisor.start();
+    supervisor.createAutoscaler(spec);
+
+    int beforeTaskCount = supervisor.getIoConfig().getTaskCount();
+
+    // When
+    supervisor.maybeScaleDuringTaskRollover();
+
+    // Then
+    Assert.assertEquals(
+        "Task count should not change when rolloverTaskCount <= 0",
+        beforeTaskCount,
+        (int) supervisor.getIoConfig().getTaskCount()
+    );
+  }
+
+  @Test
+  public void test_maybeScaleDuringTaskRollover_rolloverCountPositive_performsScaling()
+  {
+    // Given
+    final int targetTaskCount = 5;
+
+    setupSpecExpectations(getIOConfigWithCostBasedAutoScaler());
+    EasyMock.expect(spec.createAutoscaler(EasyMock.anyObject()))
+            .andReturn(createMockAutoScaler(targetTaskCount))
+            .anyTimes();
+    EasyMock.replay(spec);
+
+    TestSeekableStreamSupervisor supervisor = new TestSeekableStreamSupervisor(100);
+    supervisor.start();
+    supervisor.createAutoscaler(spec);
+
+    Assert.assertEquals(1, (int) supervisor.getIoConfig().getTaskCount());
+
+    // When
+    supervisor.maybeScaleDuringTaskRollover();
+
+    // Then
+    Assert.assertEquals(
+        "Task count should be updated to " + targetTaskCount + " when rolloverTaskCount > 0",
+        targetTaskCount,
+        (int) supervisor.getIoConfig().getTaskCount()
+    );
+  }
+
+  @Test
+  public void test_maybeScaleDuringTaskRollover_rolloverCountZero_doesNotScale()
+  {
+    // Given
+    setupSpecExpectations(getIOConfigWithCostBasedAutoScaler());
+    EasyMock.expect(spec.createAutoscaler(EasyMock.anyObject()))
+            .andReturn(createMockAutoScaler(0))
+            .anyTimes();
+    EasyMock.replay(spec);
+
+    TestSeekableStreamSupervisor supervisor = new TestSeekableStreamSupervisor(100);
+    supervisor.start();
+    supervisor.createAutoscaler(spec);
+
+    int beforeTaskCount = supervisor.getIoConfig().getTaskCount();
+
+    // When
+    supervisor.maybeScaleDuringTaskRollover();
+
+    // Then
+    Assert.assertEquals(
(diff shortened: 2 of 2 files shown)


embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/TaskScaleDownRolloverTest.java
```diff
@@ -0,0 +1,232 @@
...
+   *   <li>Verify that after rollover, fewer tasks are running due to cost-based autoscaler (no ingestion at all)</li>
+   * </ol>
+   *
+   * <p>Scale down during rollover is triggered in {@code SeekableStreamSupervisor.checkTaskDuration()}
+   * when all task groups have rolled over and the autoscaler recommends a lower task count.</p>
+   */
+  @SuppressWarnings("resource")
+  @Test
+  @Timeout(300)
+  void test_scaleDownDuringTaskRollover()
```
kfaraz: Can this test not be moved to the `CostBasedAutoScalerIntegrationTest` itself?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -204,50 +218,54 @@ public int computeOptimalTaskCount(CostMetrics metrics)
...
-          "Idle ratio [%.3f] is in ideal range for supervisorId [%s], no scaling needed",
-          currentIdleRatio,
-          supervisorId
-      );
-      return -1;
-    }
-
     int optimalTaskCount = -1;
-    double optimalCost = Double.POSITIVE_INFINITY;
+    Tuple3<Double, Double, Double> optimalCost = new Tuple3<>(Double.POSITIVE_INFINITY,
```
kfaraz: Better use a dedicated class than a Tuple or Pair.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -184,7 +198,7 @@ private CostMetrics collectMetrics()
    *
    * @return optimal task count for scale-up, or -1 if no scaling action needed
    */
-  public int computeOptimalTaskCount(CostMetrics metrics)
+  public int computeOptimalTaskCount(CostMetrics metrics, CostComputeMode costComputeMode)
```
kfaraz: Instead of keeping a cost compute mode, please keep two separate methods that may have some internal common implementation. Methods may be named something like:
`computeOptimalTaskCount` and `computeOptimalTaskCountOnRollover`.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -596,7 +601,12 @@ private boolean changeTaskCount(int desiredActiveTaskCount)
                                          AUTOSCALER_SCALING_TIME_METRIC,
                                          scaleActionStopwatch.millisElapsed()
                                      ));
-      log.info("Changed taskCount to [%s] for supervisor[%s] for dataSource[%s].", desiredActiveTaskCount, supervisorId, dataSource);
+      log.info(
+          "Changed taskCount to [%s] for supervisor[%s] for dataSource[%s].",
+          desiredActiveTaskCount,
+          supervisorId,
+          dataSource
+      );
```
kfaraz: Nit: @Fly-Style , could we please revert all the formatting changes from this PR, to keep the focus on the actual changes?

I think we can leave out the formatting changes for now.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/TaskScaleDownRolloverTest.java
Fly-Style: I decided to make it separate, because we're testing abstract autoscaler actions during the task rollover, no specifically the cost-based one. We might create lets say CPU-based autoscaler later with the requirement to scale-down during the rollover.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: +

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/TaskScaleDownRolloverTest.java
kfaraz: > because we're testing abstract autoscaler actions during the task rollover, no specifically the cost-based one.

I am not sure if this is entirely true, since cost-based auto-scaler is the only one that supports task count change on rollover right now.

The new test should be in the existing `CostBasedAutoScalerIntegrationTest` since it uses the same cluster setup (AFAICT) and verifies an important aspect of the cost-based auto-scaler.

In the future, when we add more auto-scalers, we can add separate tests for them.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
Fly-Style: https://github.com/apache/druid/pull/18860/commits/d6894c7a962328efa477aeb24686e5a9720b6042

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: https://github.com/apache/druid/pull/18860/commits/d6894c7a962328efa477aeb24686e5a9720b6042

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: https://github.com/apache/druid/pull/18860/commits/d6894c7a962328efa477aeb24686e5a9720b6042

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostResult.java
```diff
@@ -0,0 +1,58 @@
...
+ * under the License.
+ */
+
+package org.apache.druid.indexing.seekablestream.supervisor.autoscaler;
+
+/**
+ * Holds the result of a cost computation from {@link WeightedCostFunction#computeCost}.
+ * All costs are measured in seconds.
+ */
+public class CostResult
```
kfaraz, line 26: Thanks for adding this.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -504,6 +504,11 @@ private boolean createAndStartSupervisorInternal(SupervisorSpec spec, boolean pe
       supervisor = spec.createSupervisor();
       autoscaler = spec.createAutoscaler(supervisor);
 
+      // Wire autoscaler back to supervisor for rollover-based scale-down
+      if (supervisor instanceof SeekableStreamSupervisor && autoscaler != null) {
```
kfaraz: It feels weird to first create the auto-scaler and then inject it back into the supervisor.

How about we add a `createTaskAutoScaler()` method on `Supervisor` interface itself.
Internally, this method will simply call `spec.createAutoscaler(this)` and will initialize its own auto-scaler field if needed.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -3339,43 +3346,43 @@ private void checkTaskDuration() throws ExecutionException, InterruptedException
     final AtomicInteger numStoppedTasks = new AtomicInteger();
     // Sort task groups by start time to prioritize early termination of earlier groups, then iterate for processing
     activelyReadingTaskGroups.entrySet().stream().sorted(
-            Comparator.comparingLong(
```
kfaraz, line 3342: Please revert all the formatting changes in this file.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
```diff
@@ -161,6 +155,20 @@ public void test_autoScaler_computesOptimalTaskCountAndProduceScaleDown()
...
+   * <ol>
+   *   <li>Start supervisor with 10 tasks and 50 partitions, minimal data (500 records)</li>
+   *   <li>Wait for initial tasks to start running</li>
+   *   <li>Wait for the first task rollover to complete (task duration is 8 seconds)</li>
+   *   <li>Verify that after rollover, fewer tasks are running due to cost-based autoscaler (no ingestion at all)</li>
+   * </ol>
+   *
+   * <p>Scale down during rollover is triggered in {@code SeekableStreamSupervisor.checkTaskDuration()}
+   * when all task groups have rolled over and the autoscaler recommends a lower task count.</p>
+   */
```
kfaraz: I feel we should omit the javadoc. The test itself should be readable enough to follow through the details.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
```diff
@@ -258,7 +324,7 @@ private KafkaSupervisorSpec createKafkaSupervisorWithAutoScaler(
             ioConfig -> ioConfig
                 .withConsumerProperties(kafkaServer.consumerProperties())
                 .withTaskCount(taskCount)
-                .withTaskDuration(Seconds.THREE.toPeriod())
+                .withTaskDuration(Seconds.parseSeconds("PT7S").toPeriod())
```
kfaraz:
```suggestion
                .withTaskDuration(Period.seconds(7))
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
```diff
@@ -215,6 +223,64 @@ public void test_autoScaler_computesOptimalTaskCountAndProducesScaleUp()
...
+
+    // Wait for the task running for the datasource managed by a supervisor.
+    overlord.latchableEmitter().waitForEvent(e -> e.hasMetricName("task/run/time")
+                                                   .hasDimension(DruidMetrics.DATASOURCE, dataSource));
+
+    // After rollover, verify that the running task count has decreased
+    // The autoscaler should have recommended fewer tasks due to high idle time
+    final int postRolloverRunningTasks = cluster.callApi().getTaskCount("running", dataSource);
+
+    Assertions.assertTrue(
```
kfaraz, line 255: Thanks for adding this verification.

Can we update the existing scale down test to verify that scaling was actually skipped in that case?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -129,46 +136,15 @@ public void reset()
     // No-op.
   }
 
-  private CostMetrics collectMetrics()
```
kfaraz, line 132: Nit: Can we retain the original order of methods? It might help clean up the patch.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -184,7 +160,7 @@ private CostMetrics collectMetrics()
    *
    * @return optimal task count for scale-up, or -1 if no scaling action needed
    */
-  public int computeOptimalTaskCount(CostMetrics metrics)
+  int computeOptimalTaskCount(CostMetrics metrics, CostComputeMode costComputeMode)
```
kfaraz: Nit: Rather than using a `ComputeMode` or even a boolean flag, it would be cleaner to just add two new methods, both of which call the `computeOptimalTaskCount` method.

```java
public int computeTaskCountOnRollover(int currentTaskCount)
{
    // Perform both scale downs and scale ups
    return computeOptimalTaskCount();
}

{
    final CostMetrics metrics = collectMetrics();
    final int currentTaskCount = metrics.currentTaskCount();
    final int optimalTaskCount = computeOptimalTaskCount(metrics);

    // Perform only scale up actions
    return optimalTaskCount >= currentTaskCount ? optimalTaskCount : -1;
}
```

This would significantly simplify the diff and clarify the intent better.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: To be honest, I prefer to keep it changed, because it follows the correct order.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
Fly-Style: Here it is actually should not be skipped :)

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
Fly-Style: Done.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -3338,6 +3352,7 @@ private void checkTaskDuration() throws ExecutionException, InterruptedException
 
     final AtomicInteger numStoppedTasks = new AtomicInteger();
     // Sort task groups by start time to prioritize early termination of earlier groups, then iterate for processing
+    // Sort task groups by start time to prioritize early termination of earlier groups, then iterate for processing
```
kfaraz: duplicate comment.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -3376,7 +3391,6 @@ private void checkTaskDuration() throws ExecutionException, InterruptedException
             }
           }
         });
-
```
kfaraz, line 3379: Please retain the newline for clean separation of code.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -3428,6 +3442,17 @@ private void checkTaskDuration() throws ExecutionException, InterruptedException
       // remove this task group from the list of current task groups now that it has been handled
       activelyReadingTaskGroups.remove(groupId);
     }
+
+    if (taskAutoScaler != null && activelyReadingTaskGroups.isEmpty()) {
+      int rolloverTaskCount = taskAutoScaler.computeTaskCountForRollover();
+      if (rolloverTaskCount > 0 && rolloverTaskCount < ioConfig.getTaskCount()) {
+        log.info("Cost-based autoscaler recommends scaling down to [%d] tasks during rollover", rolloverTaskCount);
```
```suggestion
        log.info("Autoscaler recommends scaling down to [%d] tasks during rollover.", rolloverTaskCount);
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -3428,6 +3442,17 @@ private void checkTaskDuration() throws ExecutionException, InterruptedException
       // remove this task group from the list of current task groups now that it has been handled
       activelyReadingTaskGroups.remove(groupId);
     }
+
+    if (taskAutoScaler != null && activelyReadingTaskGroups.isEmpty()) {
+      int rolloverTaskCount = taskAutoScaler.computeTaskCountForRollover();
+      if (rolloverTaskCount > 0 && rolloverTaskCount < ioConfig.getTaskCount()) {
```
kfaraz: Should we also allow scale up on task rollover?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4330,6 +4355,15 @@ public SeekableStreamSupervisorIOConfig getIoConfig()
     return ioConfig;
   }
 
+  /**
+   * Sets the autoscaler reference for rollover-based scale-down decisions.
+   * Called by {@link SupervisorManager} after supervisor creation.
+   */
+  public void setTaskAutoScaler(@Nullable SupervisorTaskAutoScaler taskAutoScaler)
```
kfaraz: Is this still needed?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -204,50 +179,46 @@ public int computeOptimalTaskCount(CostMetrics metrics)
...
         supervisorId,
         metrics.getCurrentTaskCount(),
         optimalTaskCount,
-        optimalCost,
+        optimalCost.totalCost(),
         metrics.getAvgPartitionLag(),
         metrics.getPollIdleRatio()
     );
 
+
```
kfaraz: Nit: extra newline?

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/Supervisor.java
```diff
@@ -86,6 +87,11 @@ default Boolean isHealthy()
     return null; // default implementation for interface compatability; returning null since true or false is misleading
   }
 
+  default SupervisorTaskAutoScaler createAutoscaler()
+  {
+    return null;
+  }
```
kfaraz: It seems a little untidy but the default impl should do the same thing that the existing impl does, so that we do not break extensions that use auto-scalers.

```suggestion
  default SupervisorTaskAutoScaler createAutoscaler(SupervisorSpec spec)
  {
    return spec.createAutoscaler(this);
  }
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -129,46 +125,25 @@ public void reset()
...
-      return null;
-    }
-
-    final int currentTaskCount = supervisor.getIoConfig().getTaskCount();
-    final int partitionCount = supervisor.getPartitionCount();
-
-    final Map<String, Map<String, Object>> taskStats = supervisor.getStats();
-    final double movingAvgRate = extractMovingAverage(taskStats, DropwizardRowIngestionMeters.ONE_MINUTE_NAME);
-    final double pollIdleRatio = extractPollIdleRatio(taskStats);
+    return computeOptimalTaskCount(collectMetrics());
```
kfaraz: I wonder if for this method we shouldn't just reuse the metrics collected in the last cycle.
Metrics collection may be slow since the supervisor might need to contact all the running tasks.
This would slow down the task rollover process causing ingestion lag.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -3428,6 +3442,17 @@ private void checkTaskDuration() throws ExecutionException, InterruptedException
       // remove this task group from the list of current task groups now that it has been handled
       activelyReadingTaskGroups.remove(groupId);
     }
+
+    if (taskAutoScaler != null && activelyReadingTaskGroups.isEmpty()) {
```
kfaraz, line 3459: Please move this entire new logic into a new private method and add a short javadoc to it.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
Fly-Style: Merge issue, fixed.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
Fly-Style: Ah, I thought checkstyle catch unused methods. Sorry, was removed.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
Fly-Style: Done.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Good idea, fixed

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
Fly-Style: done!

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Removed

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorScaleDuringTaskRolloverTest.java
```diff
@@ -0,0 +1,612 @@
...
+    supervisorConfig = new SupervisorStateManagerConfig();
+
+    // Common taskMaster setup - used by all tests
+    EasyMock.expect(taskMaster.getTaskRunner()).andReturn(Optional.absent()).anyTimes();
+    EasyMock.expect(taskMaster.getSupervisorManager()).andReturn(Optional.absent()).anyTimes();
+    EasyMock.replay(taskMaster);
+  }
+
+  @Test
+  public void test_maybeScaleDuringTaskRollover_noAutoScaler_doesNotScale()
```
kfaraz, line 113: Can we keep these test methods in the `SupervisorSpecTest` itself? Seems like we have copied over a lot of the boiler plate code that could have been avoided by keeping these methods there.

Alternatively, have these two tests extend some common base test class or use common test fixtures.

Fly-Style: Not really, I don't want to break incapsulation of `maybeScaleDuringTaskRollover` method for that purpose.
That was a point to introduce such test class - to be present under same package name.

kfaraz: Oh, that should be easy to fix. The old `SeekableStreamSupervisorSpecTest` is itself in the wrong pacakge. 😂 

Please move the existing test class to `org.apache.druid.indexing.seekablestream.supervisor` where both `SeekableStreamSupervisor` and `SeekableStreamSupervisorSpec` already live. Would this address your issue?

PR #18861, opened 2025-12-19
Increase code coverage for `DirectDruidClientTest`

abhishekrb19:
Test-only change:
- Increases test coverage for the DirectDruidClient class from 57%, 35%, 16% to 76%, 73%, 52% for method / line / branch coverage respectively. We can add more tests as we evolve the code and/or do bug fixes.
- Replaces mocks with more concrete classes and test helpers, allowing the internals of the class, including `HttpResponseHandler` to be exercised.
- Also added a test for https://github.com/apache/druid/pull/18841 for which coverage had to be skipped

This PR has:
- [x] been self-reviewed.

### DIFF
--- a/server/src/test/java/org/apache/druid/client/DirectDruidClientTest.java
+++ b/server/src/test/java/org/apache/druid/client/DirectDruidClientTest.java
@@ -26,6 +26,7 @@ import com.google.common.util.concurrent.SettableFuture;
 import org.apache.druid.data.input.ResourceInputSource;
 import org.apache.druid.jackson.DefaultObjectMapper;
 import org.apache.druid.java.util.common.DateTimes;
+import org.apache.druid.java.util.common.ISE;
 import org.apache.druid.java.util.common.StringUtils;
 import org.apache.druid.java.util.common.guava.Sequence;
 import org.apache.druid.java.util.http.client.HttpClient;
@@ -67,9 +68,11 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.TimeUnit;
 
 public class DirectDruidClientTest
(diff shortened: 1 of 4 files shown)


server/src/test/java/org/apache/druid/client/DirectDruidClientTest.java
```diff
@@ -241,244 +191,199 @@ public void testRun() throws Exception
...
+  {
+    return initHttpClientWithFuture(new TestHttpClient(objectMapper), false);
+  }
+
+  private HttpClient initHttpClientWithFuture(ListenableFuture future)
+  {
+    return initHttpClientWithFuture(new TestHttpClient(objectMapper, future), false);
+  }
+
+  private HttpClient initHttpClientWithFuture(TestHttpClient httpClient, boolean throwQueryError)
```
jtuglu1: can we name this something different? It has the same name as the above method but different signature. Maybe something like `initHttpClientFromExistingClient`.

server/src/test/java/org/apache/druid/client/DirectDruidClientTest.java
```diff
@@ -241,244 +191,199 @@ public void testRun() throws Exception
...
+    QueryInterruptedException actualException =
+        Assert.assertThrows(QueryInterruptedException.class, () -> results.toList());
     Assert.assertEquals("testing1", actualException.getErrorCode());
     Assert.assertEquals("testing2", actualException.getMessage());
     Assert.assertEquals(hostName, actualException.getHost());
-    EasyMock.verify(httpClient);
   }
 
   @Test
   public void testQueryTimeoutBeforeFuture() throws IOException, InterruptedException
```
jtuglu1: Removing the mocks I believe removes `InterruptedException` – let's clean that up too while we're here.

Come to think of it, I'm not quite sure why a linter wouldn't catch this. We should add this as a rule.

server/src/test/java/org/apache/druid/client/DirectDruidClientTest.java
```diff
@@ -151,86 +110,77 @@ public void testRun() throws Exception
...
-        httpClient.go(
-            EasyMock.capture(capturedRequest),
-            EasyMock.<HttpResponseHandler>anyObject(),
-            EasyMock.anyObject(Duration.class)
-        )
-    )
-            .andReturn(SettableFuture.create())
-            .atLeastOnce();
+    // A simple HttpClient that records requests and returns queued futures per call.
+    class SequencedHttpClient implements HttpClient
```
jtuglu1: Can we move this to member class? This might be helpful for other tests.

jtuglu1 (approved):
LGTM – once the minor comments are handled.

server/src/test/java/org/apache/druid/client/DirectDruidClientTest.java
abhishekrb19: Removed `InterruptedException` from this test method.  

>Come to think of it, I'm not quite sure why a linter wouldn't catch this. We should add this as a rule.

hmm, I vaguely recall that we did have IntelliJ checks to catch this kind of issue, but they were removed at some point. I'm not sure if pmc/checkstyle can be configured to catch this.

server/src/test/java/org/apache/druid/client/DirectDruidClientTest.java
abhishekrb19: Sure, I’ve moved it to a separate test class under the same root as `TestHttpClient`.

PR #18866, opened 2025-12-23
Emit missing `ingest/merge/time`, `ingest/merge/cpu` and `ingest/persists/cpu` metrics

abhishekrb19:
The metrics `ingest/merge/time`, `ingest/merge/cpu` and `ingest/persists/cpu` metrics have [been](https://druid.apache.org/docs/latest/operations/metrics/#:~:text=0-,ingest/merge/time,-Milliseconds%20spent%20merging) documented but were previously reported as zero because they were not set in the streaming and batch appendators (it probably regressed in a refactor). 

These metrics are now set by the appendators and will be reported during the persist and merge phases for realtime and batch ingestion. Updated unit tests to verify that these values are emitted (non-zero values).

Also, cleaned up `KafkaIndexTaskTest` and `KinesisIndexTaskTest` by initializing a `StubServiceEmitter` as a non-static member in the base class `SeekableStreamIndexTaskTestBase` so it can be used by each unit test as needed.

### Release note

Correctly report `ingest/merge/time`, `ingest/merge/cpu` and `ingest/persists/cpu` metrics for streaming and batch ingestion tasks, which were previously always misreported as zero values.

This PR has:

- [x] been self-reviewed.
(https://github.com/apache/druid/blob/master/dev/license.md)


server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -748,14 +750,15 @@ public Object call() throws IOException
             }
             finally {
               metrics.incrementNumPersists();
```
jtuglu1: nit: can we move the increments below the time-sensitive metric emissions? Given we are reporting at nano-scale, chance of interrupt and/or delay due to atomic load+store can skew the nano count by X-XXns.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -991,8 +998,8 @@ private DataSegmentWithMetadata mergeAndPush(
           identifier,
           segment.getSize(),
           indexes.size(),
-          (mergeFinishTime - startTime) / 1000000,
```
jtuglu1, line 994: 👍 We should probably make sure this is not being done elsewhere. For future readers, `System.nanoTime()` is actually more expensive than `System.getCurrentTimeMillis()` so if you need milli-level precision, just use the latter from the start. Milli itself is also expensive: https://pzemtsov.github.io/2017/07/23/the-slow-currenttimemillis.html

jtuglu1 (approved):
LGTM – left minor comments.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
abhishekrb19: Yeah, there are a few other places where this nanos -> millis conversion happens, even though only millis is really required. We can clean those up together in a separate change.

PR #18871, opened 2025-12-29
Add virtualStorageFabricEvictImmediately config.

gianm:
The purpose of this config is to enable using SegmentLocalCacheManager for loading segments on MSQ worker tasks, where segments are not assigned by load/drop rules, and where there is not generally a specific maxSize configured for the local cache. We need to evict segments immediately so local disks don't fill up.

The main changes:

1) In StorageLocation, update the `releaseHold` runnable to check for `evictImmediately`. If it is set, unmount the cache entry if all holds have been released.

2) In SegmentLocalCacheManager, when `evictImmediately` is set, `mount` sets an `onUnmount` handler to delete the info file.

### DIFF
--- a/server/src/main/java/org/apache/druid/segment/loading/SegmentLoaderConfig.java
+++ b/server/src/main/java/org/apache/druid/segment/loading/SegmentLoaderConfig.java
@@ -77,8 +77,13 @@ public class SegmentLoaderConfig
   @JsonProperty("virtualStorageLoadThreads")
   private int virtualStorageLoadThreads = 2 * runtimeInfo.getAvailableProcessors();
 
-  @JsonProperty("virtualStorageFabricEvictImmediately")
-  private boolean virtualStorageFabricEvictImmediately = false;
+  /**
+   * When enabled, weakly-held cache entries are evicted immediately upon release of all holds, rather than
+   * waiting for space pressure to trigger eviction. This setting is not intended to be configured directly by
+   * administrators. Instead, it is expected to be set when appropriate via {@link #setVirtualStorage}.
+   */
+  @JsonProperty("virtualStorageFabricEvictImmediatelyOnHoldRelease")
+  private boolean virtualStorageFabricEvictImmediatelyOnHoldRelease = false;
 
   private long combinedMaxSize = 0;
 
@@ -157,9 +162,9 @@ public class SegmentLoaderConfig
     return virtualStorageLoadThreads;
   }
 
-  public boolean isVirtualStorageFabricEvictImmediately()
+  public boolean isVirtualStorageFabricEvictImmediatelyOnHoldRelease()
   {
-    return virtualStorageFabricEvictImmediately;
+    return virtualStorageFabricEvictImmediatelyOnHoldRelease;
   }
 
   public SegmentLoaderConfig withLocations(List<StorageLocationConfig> locations)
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/druid/segment/loading/StorageLocation.java
```diff
@@ -444,6 +435,52 @@ public void release(CacheEntry entry)
...
+      lock.writeLock().lock();
+      try {
+        weakCacheEntries.computeIfPresent(
+            weakEntry.cacheEntry.getId(),
+            (cacheEntryIdentifier, weakCacheEntry) -> {
+              // If we never successfully mounted, go ahead and remove so we don't have a dead entry.
+              // Futhermore, if evictImmediately is set, evict on release if all holds are gone.
+              final boolean isMounted = weakCacheEntry.cacheEntry.isMounted();
+              if ((isNewEntry && !isMounted)
+                  || (evictImmediately && !weakCacheEntry.isHeld())) {
```
clintropolis: i guess the implication here is that segments can only be used once per download (or like, concurrently with multiple holds) if `evictImmediately` is set, this feels worth a comment somewhere (or rebranding slightly like suggested in other comment)

server/src/main/java/org/apache/druid/segment/loading/StorageLocation.java
```diff
@@ -106,9 +106,7 @@ public class StorageLocation
   @GuardedBy("lock")
   private WeakCacheEntry hand;
 
-  /**
-   * Current total size of files in bytes, including weak entries.
-   */
+  private volatile boolean evictImmediately = false;
```
clintropolis: nit: maybe call this `evictImmediatelyOnHoldRelease` or like.. `burnAfterReading` or something to make it more obvious what the behavior is?

server/src/main/java/org/apache/druid/segment/loading/SegmentLoaderConfig.java
```diff
@@ -77,6 +77,9 @@ public class SegmentLoaderConfig
   @JsonProperty("virtualStorageLoadThreads")
   private int virtualStorageLoadThreads = 2 * runtimeInfo.getAvailableProcessors();
 
+  @JsonProperty("virtualStorageFabricEvictImmediately")
+  private boolean virtualStorageFabricEvictImmediately = false;
```
clintropolis: i wonder if this should this just be set automatically by a module or something instead of a user facing config that needs to be set manually, but this is fine for now we can worry about that later

gianm: In #18873, which uses this, it is set automatically for tasks. I don't expect an admin to ever want or need to configure this. Will add a comment about that.

server/src/main/java/org/apache/druid/segment/loading/StorageLocation.java
gianm: Yeah, the idea is that segments are evicted as soon as they have no holds. I'll do the rebrand.

server/src/main/java/org/apache/druid/segment/loading/StorageLocation.java
gianm: I thought `evictImmediately` was clear, since what else could "immediately" mean beyond "as soon as holds are released"? Will do the rebrand though.

clintropolis: i think i was mostly worried about this causing confusion with `reclaim` which is an entirely separate eviction process to drop weak references when something else needs the space

clintropolis: ah, how about 'ephemeral'? maybe something like `areWeakEntriesEphemeral` or something? also this is fine too though, naming is the worst

gianm: To me they seem about as good as each other, but if one seems a lot better to you, lmk and I can change it.

gianm: I'll merge this but LMK if you want to change it (or just do change it).

PR #18873, opened 2025-12-29
MSQ: Use VSF, give stages control over inputs.

gianm:
This patch integrates MSQ with virtual storage. It also refactors how MSQ reads inputs to give stages more control over how inputs are read and merged. In particular, stages are now able to fully control merging logic.

The main changes:

1) Integrate with virtual storage. Removed DataSegmentProvider, replaced it with direct usage of SegmentManager by SegmentsInputSliceReader. The SegmentManager reference ends up wrapped into RegularLoadableSegment, which provides methods `acquire()` and `acquireIfCached()` to the query logic.

2) Give stages control over input merging: rework InputSliceReader to return ReadablePartitions directly, without embedding any merging logic. Break out StandardPartitionReader as a separate class.

Other changes:

1) Move ReadableInput to the querykit package. It is no longer specific to the MSQ framework.

2) Remove StandardStageProcessor, refactoring dependent code to not require it.

3) Remove ExternalColumnSelectorFactory wrapper. Type casting is now handled directly by RowBasedColumnSelectorFactory.

4) Include full query context in worker context, rather than just a subset.

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/querykit/ReadableInputQueue.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/querykit/ReadableInputQueue.java
@@ -364,6 +364,10 @@ public class ReadableInputQueue implements Closeable
       for (AcquireSegmentAction acquireSegmentAction : loadingSegments) {
         CloseableUtils.closeAndSuppressExceptions(
             acquireSegmentAction,
+
+            // AcquireSegmentAction currently doesn't have a meaningful toString method, so if this message
+            // ever actually gets logged, it won't mention the specific segment that had a problem. Perhaps
+            // one day this will change.
             e -> log.warn(e, "Failed to close loadingSegment[%s]", acquireSegmentAction)
         );
       }
--- a/processing/src/main/java/org/apache/druid/segment/SegmentReference.java
+++ b/processing/src/main/java/org/apache/druid/segment/SegmentReference.java
@@ -32,10 +32,15 @@ import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
- * Wrapper for a {@link SegmentDescriptor} and {@link Optional<Segment>}, the latter being created by a
- * {@link SegmentMapFunction} being applied to a {@link ReferenceCountedSegmentProvider}. Closing this object closes
- * both the {@link #segmentReference} and any closeables attached from the process of creating this object, such as
- * from {@link AcquireSegmentAction}
+ * Wrapper for a {@link SegmentDescriptor} and {@link Optional<Segment>}. For regular segments, the latter is typically
+ * created by a {@link ReferenceCountedSegmentProvider} that may have a {@link SegmentMapFunction} applied to it.
+ *
+ * If the {@link SegmentMapFunction} you want to apply is not available at the time the {@link SegmentReference}
+ * is created, use {@link #map(SegmentMapFunction)} to apply it.
+ *
+ * Closing this object closes both the {@link #getSegmentReference()} and any closeables attached from the process of
+ * creating this object, such as from {@link AcquireSegmentAction}. The object from {@link #getSegmentReference()}
+ * should not be closed directly by callers.
  */
 public class SegmentReference implements Closeable
 {


indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentCacheManagerFactory.java
```diff
@@ -52,11 +52,12 @@ public SegmentCacheManagerFactory(
...
 
-  public SegmentCacheManager manufacturate(File storageDir)
+  public SegmentCacheManager manufacturate(File storageDir, boolean virtualStorage)
   {
-    final SegmentLoaderConfig loaderConfig = new SegmentLoaderConfig().withLocations(
-        Collections.singletonList(new StorageLocationConfig(storageDir, null, null))
-    );
+    final SegmentLoaderConfig loaderConfig =
+        new SegmentLoaderConfig()
+            .setLocations(Collections.singletonList(new StorageLocationConfig(storageDir, null, null)))
```
clintropolis, line 59: i know this isn't new, but should this have a size so like... there is a guard rail to not just wreck things if something tries to load too much? There is `tmpStorageBytesPerTask` already on the task config, but that is used for tmp sorter output it looks like so.. maybe those files should be managed with StorageLocation too. This doesn't need to be done in this PR, just thinking

multi-stage-query/src/main/java/org/apache/druid/msq/input/external/ExternalInputSliceReader.java
```diff
@@ -66,48 +65,63 @@
...
-                counters.warnings(),
-                warningPublisher
-            ),
-            ReadableInput::segment
-        )
-    );
+    for (final InputSource inputSource : externalInputSlice.getInputSources()) {
+      // The LoadableSegment generated here does not acquire a real hold, and ends up loading the external data in a
+      // processing thread (when the cursor is created). Ideally, this would be better integrated with the virtual
+      // storage system, giving us storage holds and the ability to load data outside of a processing thread.
```
clintropolis, line 97: :+1: it feels like this should be possible, `StorageLocation` is pretty generic, but agree we can worry about this later

processing/src/main/java/org/apache/druid/segment/SegmentReference.java
```diff
@@ -52,23 +57,60 @@ public SegmentReference(
...
   {
     return segmentReference;
   }
 
+  /**
+   * Maps the wrapped segment and returns a reference to it. Closes the reference if the {@link SegmentMapFunction}
+   * throws an exception. Regardless of success or failure of this method, the old reference should be discarded.
+   * Do not call {@link #close()} on the old reference, only call it on the new one.
+   */
+  public SegmentReference map(final SegmentMapFunction segmentMapFn)
```
clintropolis, line 87: could you update the javadoc at the top of this class to indicate that callers can call this if they don't have a mapFn to apply to the segment at the time of creation of the `SegmentReference`?

multi-stage-query/src/main/java/org/apache/druid/msq/querykit/ReadableInputQueue.java
```diff
@@ -0,0 +1,366 @@
...
+    synchronized (this) {
+      readablePartitions.clear();
+      queryableServers.clear();
+      loadableSegments.clear();
+
+      // Cancel all pending segment loads.
+      for (AcquireSegmentAction acquireSegmentAction : loadingSegments) {
+        CloseableUtils.closeAndSuppressExceptions(
+            acquireSegmentAction,
+            e -> log.warn(e, "Failed to close loadingSegment[%s]", acquireSegmentAction)
```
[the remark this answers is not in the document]
clintropolis: this looks legit, maybe we should decorate `AcquireSegmentAction` to have a segmentId or something and implement toString

multi-stage-query/src/main/java/org/apache/druid/msq/input/PhysicalInputSlice.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.druid.msq.exec.DataServerQueryHandler;
+import org.apache.druid.msq.input.stage.ReadablePartitions;
+
+import java.util.List;
+
+/**
+ * An {@link InputSlice} that has been prepared for reading by an {@link InputSliceReader}. Nothing contained in
+ * this class references open resources, so this class is not closeable and does not need to be closed.
+ */
+public class PhysicalInputSlice
```
clintropolis, line 32: what does 'physical' mean here? obligatory 'naming is hard' comment. Is it basically 'slice that gets processed to kick everything off by a leaf processor'? If so, should it be `LeafInputSlice` to match other terminology? though i guess i could see how that could be confusing too

multi-stage-query/src/main/java/org/apache/druid/msq/input/RegularLoadableSegment.java
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * Fetches the {@link DataSegment}, either returning it immediately if cached or fetching from the Coordinator.
+   */
+  private ListenableFuture<DataSegment> fetchDataSegment()
+  {
+    if (cachedDataSegment != null) {
+      return Futures.immediateFuture(cachedDataSegment);
+    } else if (coordinatorClient != null) {
+      return coordinatorClient.fetchSegment(
```
clintropolis, line 190: i find myself wondering if maybe we can just do something to make `SegmentManager` more flexible (like say backed with a coordinator client to download datasegments on the fly) to make it so we can just do all the stuff with a segmentManager, but haven't really thought about this and maybe is too weird, either way, i think this is fine for this PR

indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentCacheManagerFactory.java
gianm: I was hoping to manage more stuff through the storage locations as time goes on and that would be a good time to unify things. The other thing that isn't currently tracked is temporary storage of external data that has been downloaded.

processing/src/main/java/org/apache/druid/segment/SegmentReference.java
gianm: Updated to:

```
/**
 * Wrapper for a {@link SegmentDescriptor} and {@link Optional<Segment>}. For regular segments, the latter is typically
 * created by a {@link ReferenceCountedSegmentProvider} that may have a {@link SegmentMapFunction} applied to it.
 *
 * If the {@link SegmentMapFunction} you want to apply is not available at the time the {@link SegmentReference}
 * is created, use {@link #map(SegmentMapFunction)} to apply it.
 *
 * Closing this object closes both the {@link #getSegmentReference()} and any closeables attached from the process of
 * creating this object, such as from {@link AcquireSegmentAction}. The object from {@link #getSegmentReference()}
 * should not be closed directly by callers.
 */
```

multi-stage-query/src/main/java/org/apache/druid/msq/querykit/ReadableInputQueue.java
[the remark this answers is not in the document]
gianm: I thought about that but didn't want to add a field to `AcquireSegmentAction` just for this. I was going to leave it the way it is.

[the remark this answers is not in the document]
gianm: I'll add a comment about this though.

multi-stage-query/src/main/java/org/apache/druid/msq/input/PhysicalInputSlice.java
gianm: It's definitely a "naming is hard" thing.

It's a somewhat-more-physical version of `InputSlice`. `InputSlice` is always purely logical (it just has the identifiers / paths of things that should be read, as strings and POJOs) but `PhysicalInputSlice` may contain the ability to actually read external data or actually load segments.

It's not just "leaf" since it includes partition inputs, which would be used by non-leaf stages.

multi-stage-query/src/main/java/org/apache/druid/msq/input/RegularLoadableSegment.java
gianm: That would be nice.

gianm: Along with the ability to manage non-segment data, like external files from cloud storage.

PR #18876, opened 2025-12-29
Add task dimensions to `ServiceEmitter`

abhishekrb19:
This patch adds task dimensions to the `ServiceEmitter` so individual monitors no longer need to pass the task holder to retrieve task dimensions and append them on every monitor call. Some of the monitors were even missing these task dimensions. 

The added dimensions include `taskId`, `dataSource`, `taskType` and `groupId`.  This change ensures that all task metrics are emitted with these task dimensions by default and in a consistent format.

Changes:
- Plumb `TaskHolder` into the `ServiceEmitter` via the `EmitterModule`.
- Add a new method, `TaskHolder.getMetricDimensions()` and override it as appropriate.
- Initialize the dimensions once during the ServiceEmitter’s `start()` by calling `TaskHolder.getMetricDimensions()`. This is done during the `start()` explicitly to ensure all the dependencies are ready.
- Move the `TaskHolder` and `NoopTaskHolder` classes from the `server` module to the `processing` module for better reuse.
- Remove and cleanup all usages of the `dimension` map from the individual monitor implementation since this is taken care of already.
- Add `TestLoadSpecHolder`; this isn’t really in scope for this change, so I’m happy to move it out to simplify review if needed.
- The remainder of the changes are primarily tests and related refactoring.

#### Release note

All task metrics are emitted with the following dimensions: `taskId`, `dataSource`, `taskType`, `groupId`, and `id` (for backward compatibility; `id` will be removed in favor of the `taskId` dimension in a future release).

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceEmitter.java
+++ b/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceEmitter.java
@@ -32,7 +32,7 @@ import java.io.IOException;
 
 public class ServiceEmitter implements Emitter
 {
-  protected final Emitter emitter;
+  private final Emitter emitter;
   private final String service;
   private final ImmutableMap<String, String> otherServiceDimensions;
   private final String host;
--- a/processing/src/main/java/org/apache/druid/java/util/metrics/TaskHolder.java
+++ b/processing/src/main/java/org/apache/druid/java/util/metrics/TaskHolder.java
@@ -24,7 +24,7 @@ import java.util.Map;
 
 /**
  * Provides identifying information for a task. Implementations return {@code null}
- * when used in server processes that are not {@code CliPeon}. Note that t
+ * when used in server processes that are not {@code CliPeon}.
  */
 public interface TaskHolder
 {
@@ -41,13 +41,13 @@ public interface TaskHolder
   String getTaskId();
 
   /**
-   * @return the taskId, or {@code null} if called from a server that is not {@code CliPeon}.
+   * @return the type name of this task, or {@code null} if called from a server that is not {@code CliPeon}.
    */
   @Nullable
   String getTaskType();
 
   /**
-   * @return the taskId, or {@code null} if called from a server that is not {@code CliPeon}.
+   * @return the group ID of this task, or {@code null} if called from a server that is not {@code CliPeon}.
    */
   @Nullable
   String getGroupId();
--- a/processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
+++ b/processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
@@ -55,21 +55,21 @@ public class StubServiceEmitter extends ServiceEmitter implements MetricsVerifie
     this("testing", "localhost");
   }
 
-  /**
-   * Initialize a stub service emitter and auto-{@link #start()}  it for test convenience.
-   */
   public StubServiceEmitter(String service, String host)
   {
     this(service, host, new NoopTaskHolder());
-    super.start();
   }
 
-  /**
-   * Initialize a stub service emitter. Tests must explicitly call {@link #start()}.
-   */
   public StubServiceEmitter(String service, String host, TaskHolder taskHolder)
   {
     super(service, host, new NoopEmitter(), ImmutableMap.of(), taskHolder);
+  }
+
+  public static StubServiceEmitter createStarted()
+  {
+    final StubServiceEmitter stubServiceEmitter = new StubServiceEmitter();
+    stubServiceEmitter.start();
+    return stubServiceEmitter;
   }
 
   @Override
@@ -192,7 +192,7 @@ public class StubServiceEmitter extends ServiceEmitter implements MetricsVerifie
   public void close()
   {
     try {
-      emitter.close();
+      super.close();
     }
     catch (IOException e) {
       throw new RuntimeException(e);
--- a/server/src/test/java/org/apache/druid/server/emitter/EmitterModuleTest.java
+++ b/server/src/test/java/org/apache/druid/server/emitter/EmitterModuleTest.java
@@ -118,7 +118,7 @@ public class EmitterModuleTest
 
     ImmutableSet<NodeRole> nodeRoles = ImmutableSet.of();
 
-    TestTaskHolder testTaskHolder = new TestTaskHolder("d", "e", "a", "w");
+    TestTaskHolder testTaskHolder = new TestTaskHolder("wiki", "id1", "type1", "group1");
     Injector injector = Guice.createInjector(
         new JacksonModule(),
         new LifecycleModule(),
@@ -156,16 +156,16 @@ public class EmitterModuleTest
     Assert.assertEquals(1, events.size());
     ServiceMetricEvent event = (ServiceMetricEvent) events.get(0);
     EventMap map = event.toMap();
-    Assert.assertEquals("e", map.get(DruidMetrics.TASK_ID));
-    Assert.assertEquals("e", map.get(DruidMetrics.ID));
-    Assert.assertEquals("a", map.get(DruidMetrics.TASK_TYPE));
-    Assert.assertEquals("w", map.get(DruidMetrics.GROUP_ID));
-    Assert.assertEquals("d", map.get(DruidMetrics.DATASOURCE));
+    Assert.assertEquals("id1", map.get(DruidMetrics.TASK_ID));
+    Assert.assertEquals("id1", map.get(DruidMetrics.ID));
+    Assert.assertEquals("type1", map.get(DruidMetrics.TASK_TYPE));
+    Assert.assertEquals("group1", map.get(DruidMetrics.GROUP_ID));
+    Assert.assertEquals("wiki", map.get(DruidMetrics.DATASOURCE));
     stubEmitter.flush();
 
     // Override a dimension and verify that is emitted
     final ServiceMetricEvent.Builder builder2 = new ServiceMetricEvent.Builder();
-    builder2.setDimension("taskId", "xyz");
+    builder2.setDimension("taskId", "id2");
     builder2.setMetric("metric2", 1);
     instance.emit(builder2);
 
@@ -173,11 +173,11 @@ public class EmitterModuleTest
     Assert.assertEquals(1, events2.size());
     ServiceMetricEvent event2 = (ServiceMetricEvent) events2.get(0);
     EventMap map2 = event2.toMap();
-    Assert.assertEquals("xyz", map2.get(DruidMetrics.TASK_ID));
-    Assert.assertEquals("e", map2.get(DruidMetrics.ID));
-    Assert.assertEquals("a", map2.get(DruidMetrics.TASK_TYPE));
-    Assert.assertEquals("w", map2.get(DruidMetrics.GROUP_ID));
-    Assert.assertEquals("d", map2.get(DruidMetrics.DATASOURCE));
+    Assert.assertEquals("id2", map2.get(DruidMetrics.TASK_ID));
+    Assert.assertEquals("id1", map2.get(DruidMetrics.ID));
+    Assert.assertEquals("type1", map2.get(DruidMetrics.TASK_TYPE));
+    Assert.assertEquals("group1", map2.get(DruidMetrics.GROUP_ID));
+    Assert.assertEquals("wiki", map2.get(DruidMetrics.DATASOURCE));
   }
 
   private Injector makeInjectorWithProperties(final Properties props)
--- a/server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
+++ b/server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
@@ -116,17 +116,21 @@ public class GroupByStatsMonitorTest
     // Trigger metric emission
     monitor.doMonitor(emitter);
 
-    final Map<String, Object> dimFilters = Map.of(DruidMetrics.DATASOURCE, dataSource, DruidMetrics.TASK_ID, taskId,
-        DruidMetrics.ID, taskId, DruidMetrics.TASK_TYPE, taskType, DruidMetrics.GROUP_ID, groupId
+    final Map<String, Object> dimFilters = Map.of(
+        DruidMetrics.DATASOURCE, dataSource,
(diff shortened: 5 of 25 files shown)


processing/src/main/java/org/apache/druid/java/util/metrics/Monitors.java
```diff
@@ -20,43 +20,39 @@
...
    * monitors. Emitted events have default feed {@link FeedDefiningMonitor#DEFAULT_METRICS_FEED}
-   * See: {@link Monitors#createCompoundJvmMonitor(Map, String)}
-   *
-   * @param dimensions common dimensions to configure the JVM monitor with
+   * See: {@link Monitors#createCompoundJvmMonitor(String)}
    *
    * @return a universally useful JVM-wide monitor
    */
-  public static Monitor createCompoundJvmMonitor(Map<String, String[]> dimensions)
+  public static Monitor createCompoundJvmMonitor()
```
kfaraz, line 33: Is this class/method used anywhere? I couldn't find any usage.

processing/src/main/java/org/apache/druid/java/util/metrics/TaskHolder.java
```diff
@@ -17,13 +17,14 @@
...
-package org.apache.druid.server.metrics;
+package org.apache.druid.java.util.metrics;
 
 import javax.annotation.Nullable;
+import java.util.Map;
 
 /**
  * Provides identifying information for a task. Implementations return {@code null}
- * when used in server processes that are not {@code CliPeon}.
+ * when used in server processes that are not {@code CliPeon}. Note that t
```
kfaraz: Incomplete javadoc?

processing/src/main/java/org/apache/druid/java/util/metrics/TaskHolder.java
```diff
@@ -38,4 +39,21 @@ public interface TaskHolder
    */
   @Nullable
   String getTaskId();
+
+  /**
+   * @return the taskId, or {@code null} if called from a server that is not {@code CliPeon}.
```
kfaraz:
```suggestion
   * @return the type name of this task, or {@code null} if called from a server that is not {@code CliPeon}.
```

processing/src/main/java/org/apache/druid/java/util/metrics/TaskHolder.java
```diff
@@ -38,4 +39,21 @@ public interface TaskHolder
...
+   */
+  @Nullable
+  String getTaskType();
+
+  /**
+   * @return the taskId, or {@code null} if called from a server that is not {@code CliPeon}.
```
kfaraz:
```suggestion
   * @return the group ID of this task, or {@code null} if called from a server that is not {@code CliPeon}.
```

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
```diff
@@ -41,18 +44,32 @@
...
   private final ConcurrentHashMap<String, Deque<ServiceMetricEvent>> metricEvents = new ConcurrentHashMap<>();
 
   public StubServiceEmitter()
   {
-    super("testing", "localhost", null);
+    this("testing", "localhost");
   }
 
+  /**
+   * Initialize a stub service emitter and auto-{@link #start()}  it for test convenience.
```
kfaraz: Instead of doing the auto-start in the constructor, add a static utility method that may be used something like
```
ServiceEmitter emitter = StubServiceEmitter.createStarted();
```

server/src/test/java/org/apache/druid/server/emitter/EmitterModuleTest.java
```diff
@@ -87,6 +108,78 @@ public void testInvalidEmitterType()
...
+  public void testEmitterForTaskContainsAllTaskDimensions()
+  {
+    Properties props = new Properties();
+    props.setProperty("druid.emitter", "stub");
+    EmitterModule emitterModule = new EmitterModule();
+    emitterModule.setProps(props);
+
+    ImmutableSet<NodeRole> nodeRoles = ImmutableSet.of();
+
+    TestTaskHolder testTaskHolder = new TestTaskHolder("d", "e", "a", "w");
```
kfaraz: Maybe use strings that are easier to follow, e.g. 
`datasource = wiki`, `taskID = id1`, `taskType = type1`, `groupId = group1`, etc.

server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
```diff
@@ -98,30 +103,31 @@ public void testMonitorWithServiceDimensions()
...
+    final StubServiceEmitter emitter = new StubServiceEmitter("service", "host", new TestTaskHolder(dataSource, taskId, taskType, groupId));
+    emitter.start();
     monitor.doMonitor(emitter);
     emitter.flush();
     // Trigger metric emission
     monitor.doMonitor(emitter);
 
-    final Map<String, Object> dimFilters = Map.of(
-        "taskId", List.of(taskId), "dataSource", List.of(dataSource), "id", List.of(taskId)
+    final Map<String, Object> dimFilters = Map.of(DruidMetrics.DATASOURCE, dataSource, DruidMetrics.TASK_ID, taskId,
```
kfaraz: Please put each key-value pair in a separate line.

server/src/test/java/org/apache/druid/server/metrics/GroupByStatsMonitorTest.java
```diff
@@ -174,4 +180,17 @@ public void testMonitoringMergeBuffer_pendingRequests()
       // do nothing
     }
   }
+
+
+  private void verifyTaskServiceDimensions(StubServiceEmitter emitter, String metricName, Map<String, Object> dimFilters, Number expectedValue)
```
kfaraz:
```suggestion
  private void verifyMetricValue(StubServiceEmitter emitter, String metricName, Map<String, Object> dimFilters, Number expectedValue)
```

services/src/main/java/org/apache/druid/cli/PeonTaskHolder.java
```diff
@@ -59,4 +62,34 @@ public String getTaskId()
...
+   * @return a map of all task-specific dimensions applicable to this peon.
+   * The task ID ({@link TaskHolder#getTaskId()}) is added to both {@link DruidMetrics#TASK_ID}
+   * {@link DruidMetrics#ID} dimensions to the map for backward compatibility. {@link DruidMetrics#ID} is
+   * deprecated because it's ambiguous and can be removed in a future release.</p>
+   */
+  @Override
+  public Map<String, String> getMetricDimensions()
+  {
+    return Map.of(
+        DruidMetrics.DATASOURCE, getDataSource(),
```
kfaraz: Nit: Each getter invokes the `taskProvider` separately. Maybe just invoke it once and assign the result to a `Task task` variable.

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
```diff
@@ -41,18 +44,32 @@
...
   {
-    super(service, host, null);
+    this(service, host, new NoopTaskHolder());
+    super.start();
+  }
+
+  /**
+   * Initialize a stub service emitter. Tests must explicitly call {@link #start()}.
+   */
+  public StubServiceEmitter(String service, String host, TaskHolder taskHolder)
```
kfaraz, line 63: We should update the `LatchableEmitter` constructor to accept a `TaskHolder` too.
Otherwise, the task dimensions will not show up in embedded tests.
You could also add a short test method in any of the existing tests to verify the new dimensions.

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
```diff
@@ -173,5 +191,11 @@ public void flush()
   @Override
   public void close()
   {
+    try {
+      emitter.close();
```
kfaraz: Isn't the `emitter` always a `NoopEmitter`, why do we need to invoke `start` and `close` on it?

processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceEmitter.java
```diff
@@ -25,39 +25,56 @@
...
+import org.apache.druid.java.util.metrics.NoopTaskHolder;
+import org.apache.druid.java.util.metrics.TaskHolder;
 
 import java.io.IOException;
 
 public class ServiceEmitter implements Emitter
 {
-  private final ImmutableMap<String, String> serviceDimensions;
-  private final Emitter emitter;
+  protected final Emitter emitter;
```
kfaraz: Why not private?

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitterModule.java
```diff
@@ -0,0 +1,43 @@
...
+package org.apache.druid.java.util.metrics;
+
+import com.google.inject.Binder;
+import com.google.inject.Module;
+import com.google.inject.Provides;
+import com.google.inject.name.Named;
+import org.apache.druid.guice.ManageLifecycle;
+import org.apache.druid.java.util.emitter.core.Emitter;
+
+public class StubServiceEmitterModule implements Module
```
kfaraz, line 29: Instead of adding this, I wonder if we shouldn't just move `LatchableEmitterModule` to `processing/src/test/`.

kfaraz:
> Some of the monitors were even missing these task dimensions.

@abhishekrb19 , did you mean monitors like `CgroupCpuMonitor` which do have a `Map<String, String[]> dimensions` as a constructor arg but were still not wired up correctly? I couldn't figure out how these dimensions were getting injected in the original code.

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
kfaraz: > You could also add a short test method in any of the existing tests to verify the new dimensions.

Oh, I suppose this would require MMs. You could add a metric verification in the `KubernetesTaskRunnerDockerTest` since that test runs peons and also uses an event collector + latchable emitter combo to wait on metrics.

processing/src/main/java/org/apache/druid/java/util/metrics/TaskHolder.java
abhishekrb19: Whoops, removed.

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
abhishekrb19: Good idea, done

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
abhishekrb19: Ah, I intended to delegate this to the base class via `super.close()` here, similar to how `start()` invokes `super.start()` to delegate to the base class, which takes care of additional initialization. We could skip this in `close()` as it doesn't do much, but it feels cleaner to do the same way here as well for completeness.

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitterModule.java
abhishekrb19: Yeah, I considered doing this. However, all the unit tests use `StubServiceEmitter` directly and moving `LatchableEmitterModule` to processing would also require moving non-test classes like `DruidNode` and other dependencies from the server module into processing. Given that, I’ve left this test stub module in place for now and we can refactor these test dependencies later if needed.

I’ve also wired TaskHolder into LatchableEmitterModule so the embedded tests can make use of it as needed.

processing/src/main/java/org/apache/druid/java/util/metrics/Monitors.java
abhishekrb19: I don’t think they’re used anywhere. Opened a separate PR to remove this unused class altogether: https://github.com/apache/druid/pull/18877/

abhishekrb19:
@kfaraz thanks for taking a look!

> @abhishekrb19 , did you mean monitors like CgroupCpuMonitor which do have a Map<String, String[]> dimensions as a constructor arg but were still not wired up correctly? I couldn't figure out how these dimensions were getting injected in the original code.

That's right, all the `Cgroup*` monitors were previously initialized with an empty dimensions map via the default constructor.

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
abhishekrb19: >We should update the LatchableEmitter constructor to accept a TaskHolder too.
Otherwise, the task dimensions will not show up in embedded tests.

Sounds good, updated.

>Oh, I suppose this would require MMs. You could add a metric verification in the KubernetesTaskRunnerDockerTest since that test runs peons and also uses an event collector + latchable emitter combo to wait on metrics.

I’m trying to figure out how to run the Docker tests locally to verify this...I’m looking at the GHA Docker tests to see what I’m missing in my setup. It would probably be a good idea to add a brief README or update the javadocs, to document how devs can run these tests locally, since it requires some additional setup like installing dependencies like k3s, etc.

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitterModule.java
kfaraz: > Given that, I’ve left this test stub module in place for now and we can refactor these test dependencies later if needed.

Makes sense, thanks!

processing/src/test/java/org/apache/druid/java/util/metrics/StubServiceEmitter.java
kfaraz: Yeah, that's fair, let me create a PR to add the steps for running these tests.
Meanwhile, we can merge this PR and you can add the tests in a follow up.

> since it requires some additional setup like installing dependencies like k3s, etc.

I might need to check but AFAIR, the tests just need docker to be running. There are some steps required though, like setting the Druid image name etc.

PR #18879, opened 2025-12-31
Fix segment metadata query inconsistency bug with HLLSketch columns

### LINKED ISSUE #14315: HLLSketchBuilder causes segment metadata queries to produce an error during realtime ingestion
### Affected Version

26.0.0

### Description

Thoughts on how to address this issue.
It seems like a bug in either Druid metadata or DataSketches implementation.

Query for up-to-the-minute segment metadata that has realtime ingestion using `HLLSketchBuild` yields this error
`error:cannot_merge_diff_types: [HLLSketch] and [HLLSketchBuild]`

`HLLSketchBuild` is the required aggregator for ingestion. `HLLSketch` is the type of the column produced by the builder when persisting a segment.
If we restrict the segment metadata query up to the prior hour (segment granularity is HOUR), we see the correct type reflected: `"type": "HLLSketch"`

Snippet of ingestion spec
```
"dataSchema": {
      "dataSource": "uniq_test",
      "timestampSpec": null,
      "dimensionsSpec": null,
      "metricsSpec": [
        {
          "type": "count",
          "name": "count"
        },
        {
          "type": "HLLSketchBuild",
          "name": "uniq_account_id",
          "fieldName": "user_account_id",
          "lgK": 12,
          "tgtHllType": "HLL_4"
        }
      ],
      "granularitySpec": {
        "type": "uniform",
        "segmentGranularity": "HOUR",
        "queryGranularity": "MINUTE",
        "rollup": true,
        "intervals": []
      },
...
```

Metadata query
```
{
   "queryType": "segmentMetadata",
   "dataSource": "uniq_test",
   "intervals": ["2000-01-01T00:00:00Z/2999-12-31T23:59:59Z"],
   "merge": true,
   "lenientAggregatorMerge": true,
   "analysisTypes": []
 }
```

Metadata response snippet
```
"uniq_account_id": {
  "typeSignature": "STRING",
  "type": "STRING",
  "hasMultipleValues": false,
  "hasNulls": false,
  "size": -1,
  "cardinality": null,
  "minValue": null,
  "maxValue": null,
  "errorMessage": "error:cannot_merge_diff_types: [HLLSketch] and [HLLSketchBuild]"
}
```

jtuglu1:
Fixes #14315.

### Description

As outlined in https://github.com/apache/druid/issues/14315, segment metadata queries issued against complex columns served by realtime data nodes (`IncrementalIndex`) can potentially clash with type names of the same column (in `QueryableIndex`) served by historical nodes. I thought the best way to solve this was ensuring that the column type used is consistent across both data serving nodes (e.g. realtime and historical), rather than attempting to do some case-by-case coverage. Keeping as draft until I can confirm this doesn't break any other codepaths.

#### Release note

Fix segment metadata query type inconsistency bug when querying between realtime and historical sketch columns.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/datasketches/src/test/java/org/apache/druid/query/aggregation/datasketches/BaseSketchBuildSegmentMetadataQueryTest.java
+++ b/extensions-core/datasketches/src/test/java/org/apache/druid/query/aggregation/datasketches/BaseSketchBuildSegmentMetadataQueryTest.java
@@ -65,6 +65,7 @@ import org.junit.rules.TemporaryFolder;
 
 import java.io.File;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 
@@ -75,7 +76,7 @@ import java.util.Map;
  * IncrementalIndex (realtime) and QueryableIndex (historical) segments that were created with various sketches,
  * ensuring that both segment types report the same column type (COMPLEX<HLLSketch>) by using the serde's normal type.
  */
-public abstract class SketchBuildSegmentMetadataQueryTestBase extends InitializedNullHandlingTest
+public abstract class BaseSketchBuildSegmentMetadataQueryTest extends InitializedNullHandlingTest
 {
   protected static final String DATA_SOURCE = "test_datasource";
   protected static final String SKETCH_COLUMN = "sketch";
@@ -117,14 +118,18 @@ public abstract class SketchBuildSegmentMetadataQueryTestBase extends Initialize
 
   /**
    * @return an aggregator that builds a sketch from raw input (its intermediate type may be a non-canonical
-   *         "build" type name).
+   * "build" type name).
    */
   protected abstract AggregatorFactory buildSketchAggregatorFactory(String sketchColumn, String inputFieldName);
 
-  /** @return the canonical complex type that should be reported by SegmentMetadataQuery. */
+  /**
+   * @return the canonical complex type that should be reported by SegmentMetadataQuery.
+   */
   protected abstract ColumnType expectedCanonicalColumnType();
 
-  /** Validate the combining/merged aggregator returned by SegmentMetadataQuery. */
+  /**
+   * Validate the combining/merged aggregator returned by SegmentMetadataQuery.
+   */
   protected abstract void assertMergedSketchAggregator(AggregatorFactory aggregator, String sketchColumn);
 
   @Test
@@ -190,15 +195,17 @@ public abstract class SketchBuildSegmentMetadataQueryTestBase extends Initialize
     }
 
     // Make sure the merge went through ok
-    Assert.assertEquals(ColumnType.LONG, mergedAnalysis.getColumns().get("__time").getTypeSignature());
-    Assert.assertEquals(ColumnType.STRING, mergedAnalysis.getColumns().get(DIM_COLUMN).getTypeSignature());
-    Assert.assertEquals(ColumnType.LONG, mergedAnalysis.getColumns().get("count").getTypeSignature());
+    ColumnAnalysis timeColumnAnalysis = mergedAnalysis.getColumns().get("__time");
+    Assert.assertEquals(ColumnType.LONG, timeColumnAnalysis.getTypeSignature());
+    ColumnAnalysis dimColumnAnalysis = mergedAnalysis.getColumns().get(DIM_COLUMN);
+    Assert.assertEquals(ColumnType.STRING, dimColumnAnalysis.getTypeSignature());
+    ColumnAnalysis countColumnAnalysis = mergedAnalysis.getColumns().get("count");
+    Assert.assertEquals(ColumnType.LONG, countColumnAnalysis.getTypeSignature());
+    ColumnAnalysis sketchColumnAnalysis = mergedAnalysis.getColumns().get(SKETCH_COLUMN);
     Assert.assertEquals(
         expectedCanonicalColumnType(),
-        mergedAnalysis.getColumns().get(SKETCH_COLUMN).getTypeSignature()
+        sketchColumnAnalysis.getTypeSignature()
     );
-
-    ColumnAnalysis sketchColumnAnalysis = mergedAnalysis.getColumns().get(SKETCH_COLUMN);
     Assert.assertFalse("Sketch column should not have multiple values", sketchColumnAnalysis.isHasMultipleValues());
 
     Assert.assertNotNull("Aggregators should be present", mergedAnalysis.getAggregators());
@@ -259,6 +266,10 @@ public abstract class SketchBuildSegmentMetadataQueryTestBase extends Initialize
                                        .dataSource(DATA_SOURCE)
                                        .intervals(Collections.singletonList(Intervals.ETERNITY))
                                        .merge(true)
+                                       .analysisTypes(EnumSet.of(
+                                           SegmentMetadataQuery.AnalysisType.MINMAX,
+                                           SegmentMetadataQuery.AnalysisType.CARDINALITY
+                                       ))
                                        .build();
     QueryToolChest<SegmentAnalysis, SegmentMetadataQuery> toolChest = queryRunnerFactory.getToolchest();
     QueryRunner<SegmentAnalysis> mergedRunner = toolChest.mergeResults(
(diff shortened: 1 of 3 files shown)


processing/src/main/java/org/apache/druid/segment/incremental/IncrementalIndex.java
```diff
@@ -1029,15 +1029,19 @@ public MetricDesc(int index, AggregatorFactory factory)
...
-          this.type = serde.getTypeName();
-        } else {
+        if (serde == null) {
           throw new ISE("Unable to handle complex type[%s]", valueType);
         }
+        this.type = serde.getTypeName();
+        // The serde's type name represents the canonical storage type (e.g., "HLLSketch"),
+        // while the aggregator's intermediate type may be more specific (e.g., "HLLSketchBuild").
+        // Using the serde's type ensures that segment metadata queries return consistent types
+        // across realtime (IncrementalIndex) and historical (QueryableIndex) segments.
```
abhishekrb19: Would this be applicable to all aggregators that have different build vs outputTypes like `thetaSketchMerge` vs `thetaSketch` where we want to homogenize things to the outputType? Not just `HLLSketch` type as the original issue reports?

extensions-core/datasketches/src/test/java/org/apache/druid/query/aggregation/datasketches/hll/HllSketchBuildSegmentMetadataQueryTest.java
```diff
@@ -0,0 +1,432 @@
...
+  }
+
+  /**
+   * Verifies fix for issue #14315: SegmentMetadataQuery should work correctly when
+   * querying across IncrementalIndex (realtime) and QueryableIndex (historical) segments
+   * that were created with HLLSketchBuild aggregator.
+   * <p>
+   * Both segment types should report the same column type (COMPLEX<HLLSketch>) and
+   * the merged metadata query should succeed without errors.
+   */
```
abhishekrb19: nit: can be removed as it's redundant with the class-level javadoc

processing/src/main/java/org/apache/druid/segment/incremental/IncrementalIndex.java
```diff
@@ -1029,15 +1029,19 @@ public MetricDesc(int index, AggregatorFactory factory)
...
+        // See https://github.com/apache/druid/issues/14315.
```
abhishekrb19, line 1041: Would this actually be applicable to all aggregators that have different build vs outputTypes like `thetaSketchMerge` vs `thetaSketch` as well? I think those aggregators will also require homogenizing the type to the outputType that we're doing for HLLSketch thing reported in the original issue.

extensions-core/datasketches/src/test/java/org/apache/druid/query/aggregation/datasketches/hll/HllSketchBuildSegmentMetadataQueryTest.java
```diff
@@ -0,0 +1,432 @@
...
+ * Test case for issue https://github.com/apache/druid/issues/14315
+ * <p>
+ * This test verifies that SegmentMetadataQuery works correctly when querying across
+ * IncrementalIndex (realtime) and QueryableIndex (historical) segments that were created
+ * with HLLSketchBuild aggregator.
+ * <p>
+ * The fix ensures that both segment types report the same column type (COMPLEX<HLLSketch>)
+ * by using the serde's normal type name in IncrementalIndex.
+ */
+public class HllSketchBuildSegmentMetadataQueryTest extends InitializedNullHandlingTest
```
abhishekrb19: Related to the other comment, I think it'd make sense to make this be more general as I think the issue is also applicable to other aggs like `thetaSketch`, `quantilesDoublesSketch`, etc.

If so, could we rename this test class to something more general and maybe include a test for at least one other aggregator that isn’t `HLLSketch`?

extensions-core/datasketches/src/test/java/org/apache/druid/query/aggregation/datasketches/hll/HllSketchBuildSegmentMetadataQueryTest.java
abhishekrb19: nit: can be removed as it's redundant with the class-level javadoc

extensions-core/datasketches/src/test/java/org/apache/druid/query/aggregation/datasketches/hll/HllSketchBuildSegmentMetadataQueryTest.java
```diff
@@ -0,0 +1,432 @@
...
+    );
+    Assert.assertEquals(
+        "stringEncoding should be default value",
+        StringEncoding.UTF16LE,
+        hllAggregator.getStringEncoding()
+    );
+
+    incrementalIndex.close();
+    indexToPersist.close();
+    queryableIndex.close();
```
abhishekrb19: For all these tests, you could create a `Closer` and register the indices with it as they're created. For example:
`IncrementalIndex incrementalIndex = closer.register(createIncrementalIndex(aggregators));`

extensions-core/datasketches/src/test/java/org/apache/druid/query/aggregation/datasketches/hll/HllSketchBuildSegmentMetadataQueryTest.java
```diff
@@ -0,0 +1,432 @@
...
+                .build()
+        )
+        .setMaxRowCount(1000)
+        .build();
+  }
+
+  private void addRows(IncrementalIndex index, int startId, int endId)
+  {
+    for (int i = startId; i < endId; i++) {
+      Map<String, Object> event = ImmutableMap.of(
```
abhishekrb19: nit: switch to `Map.of()` and `List.of()` for newer usages

processing/src/main/java/org/apache/druid/segment/incremental/IncrementalIndex.java
jtuglu1: Yeah – I was going to look into this. I presume it's a similar story there, I'll add another test.

jtuglu1: Added a base class which other sketches can inherit/implement as needed.

extensions-core/datasketches/src/test/java/org/apache/druid/query/aggregation/datasketches/BaseSketchBuildSegmentMetadataQueryTest.java
```diff
@@ -0,0 +1,404 @@
...
+    SegmentAnalysis mergedAnalysis = resultList.get(0);
+
+    // Verify row count: 100 rows per segment, no rollup due to unique timestamps
+    Assert.assertEquals("Merged row count should be 200", 200, mergedAnalysis.getNumRows());
+
+    Assert.assertEquals("Should have 4 columns", 4, mergedAnalysis.getColumns().size());
+    Assert.assertTrue("Should contain __time column", mergedAnalysis.getColumns().containsKey("__time"));
+    Assert.assertTrue("Should contain dim column", mergedAnalysis.getColumns().containsKey(DIM_COLUMN));
+    Assert.assertTrue("Should contain count column", mergedAnalysis.getColumns().containsKey("count"));
+    Assert.assertTrue("Should contain sketch column", mergedAnalysis.getColumns().containsKey(SKETCH_COLUMN));
```
abhishekrb19, line 187: nit: these assertion messages are redundant. We can remove them or to make it more useful by including more context on what was actually present `mergedAnalysis.getColumns()` (for `assertTrue` in particular, that will be more useful)

abhishekrb19 (approved):
Thanks! Could you also add a note to the release notes mentioning that this fix applies to other sketch aggregators as well?

extensions-core/datasketches/src/test/java/org/apache/druid/query/aggregation/datasketches/BaseSketchBuildSegmentMetadataQueryTest.java
jtuglu1: Sure – I can make some more assertions on returned `mergedAnalysis` payload.

PR #18886, opened 2026-01-06
Dart: Serve reports for running and recently-finished queries.

gianm:
This patch adds query reports, and a mapping from SQL query ID to Dart query ID, to DartControllerRegistry. It is made available through a new getQueryReport method on SqlEngine, and through a new `/druid/v2/sql/queries/<sqlQueryId>/reports` endpoint on SqlResource.

The format of the response is a JSON object with two keys, "query" and "report". The "query" key is the same info that is available from the existing `/druid/v2/sql/queries` endpoint. The "report" key is a report map including an MSQ report.

New Broker configurations control retention of reports:

- `druid.msq.dart.controller.maxRetainedReportCount` (default `0`, meaning no reports are retained)
- `druid.msq.dart.controller.maxRetainedReportDuration` (default `PT0S`, meaning time-based expiration is disabled)

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDartReportApiTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDartReportApiTest.java
@@ -35,7 +35,7 @@ import org.apache.druid.msq.indexing.report.MSQTaskReportPayload;
 import org.apache.druid.query.QueryContexts;
 import org.apache.druid.query.http.ClientSqlQuery;
 import org.apache.druid.server.metrics.LatchableEmitter;
-import org.apache.druid.sql.http.GetReportResponse;
+import org.apache.druid.sql.http.GetQueryReportResponse;
 import org.apache.druid.testing.embedded.EmbeddedBroker;
 import org.apache.druid.testing.embedded.EmbeddedClusterApis;
 import org.apache.druid.testing.embedded.EmbeddedCoordinator;
@@ -156,7 +156,7 @@ public class EmbeddedDartReportApiTest extends EmbeddedClusterTestBase
     Assertions.assertEquals("10", result);
 
     // Now fetch the report using the SQL query ID
-    final GetReportResponse reportResponse = msqApis.getDartQueryReport(sqlQueryId, broker1);
+    final GetQueryReportResponse reportResponse = msqApis.getDartQueryReport(sqlQueryId, broker1);
 
     // Verify the report response
     Assertions.assertNotNull(reportResponse, "Report response should not be null");
@@ -179,7 +179,7 @@ public class EmbeddedDartReportApiTest extends EmbeddedClusterTestBase
   public void test_getQueryReport_notFound()
   {
     // Try to get a report for a non-existent query
-    final GetReportResponse reportResponse = msqApis.getDartQueryReport("nonexistent-query-id", broker1);
+    final GetQueryReportResponse reportResponse = msqApis.getDartQueryReport("nonexistent-query-id", broker1);
 
     // Verify the response is null (not found)
     Assertions.assertNull(reportResponse, "Report response should be null for non-existent query");
@@ -220,11 +220,11 @@ public class EmbeddedDartReportApiTest extends EmbeddedClusterTestBase
     Assertions.assertEquals(1, sqlClients2.getAllClients().size(), "Broker2 should have 1 client (broker1)");
 
     // Fetch the report from both brokers, to verify cross-broker lookup is working
-    final GetReportResponse reportFromBroker1 = msqApis.getDartQueryReport(sqlQueryId, broker1);
-    final GetReportResponse reportFromBroker2 = msqApis.getDartQueryReport(sqlQueryId, broker2);
+    final GetQueryReportResponse reportFromBroker1 = msqApis.getDartQueryReport(sqlQueryId, broker1);
+    final GetQueryReportResponse reportFromBroker2 = msqApis.getDartQueryReport(sqlQueryId, broker2);
 
     // Verify the report content
-    for (GetReportResponse report : Arrays.asList(reportFromBroker1, reportFromBroker2)) {
+    for (GetQueryReportResponse report : Arrays.asList(reportFromBroker1, reportFromBroker2)) {
       Assertions.assertNotNull(report);
       final DartQueryInfo queryInfo = (DartQueryInfo) report.getQueryInfo();
       Assertions.assertEquals(sqlQueryId, queryInfo.getSqlQueryId());
@@ -251,7 +251,7 @@ public class EmbeddedDartReportApiTest extends EmbeddedClusterTestBase
         msqApis.submitDartSqlAsync(sql, Map.of(QueryContexts.CTX_SQL_QUERY_ID, sqlQueryId), broker1);
 
     // Step 2: Get the report.
-    final GetReportResponse runningReport = waitForReport(sqlQueryId);
+    final GetQueryReportResponse runningReport = waitForReport(sqlQueryId);
 
     Assertions.assertNotNull(runningReport, "Report should be available for running query");
     Assertions.assertNotNull(runningReport.getQueryInfo(), "Query info should not be null");
@@ -288,7 +288,7 @@ public class EmbeddedDartReportApiTest extends EmbeddedClusterTestBase
     );
 
     // Step 5: Fetch the report again - should now be in FAILED state
-    final GetReportResponse canceledReport = msqApis.getDartQueryReport(sqlQueryId, broker1);
+    final GetQueryReportResponse canceledReport = msqApis.getDartQueryReport(sqlQueryId, broker1);
 
     Assertions.assertNotNull(canceledReport, "Report should be available for canceled query");
     Assertions.assertNotNull(canceledReport.getReportMap(), "Report map should not be null");
@@ -319,12 +319,12 @@ public class EmbeddedDartReportApiTest extends EmbeddedClusterTestBase
   /**
    * Polls the report API until a report is available.
    */
-  private GetReportResponse waitForReport(String sqlQueryId)
+  private GetQueryReportResponse waitForReport(String sqlQueryId)
   {
     final long timeout = 30_000;
     final long deadline = System.currentTimeMillis() + timeout;
     while (System.currentTimeMillis() < deadline) {
-      final GetReportResponse report = msqApis.getDartQueryReport(sqlQueryId, broker1);
+      final GetQueryReportResponse report = msqApis.getDartQueryReport(sqlQueryId, broker1);
       if (report != null) {
         return report;
       }
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQApis.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQApis.java
@@ -22,12 +22,7 @@ package org.apache.druid.testing.embedded.msq;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.util.concurrent.ListenableFuture;
-import org.apache.druid.client.broker.BrokerClient;
-import org.apache.druid.client.broker.BrokerClientImpl;
-import org.apache.druid.common.guava.FutureUtils;
-import org.apache.druid.discovery.NodeRole;
 import org.apache.druid.error.DruidException;
-import org.apache.druid.guice.annotations.EscalatedGlobal;
 import org.apache.druid.indexer.TaskState;
 import org.apache.druid.indexer.report.TaskReport;
 import org.apache.druid.java.util.common.StringUtils;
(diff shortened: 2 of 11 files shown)


gianm:
Added `jacoco:skip` because most of the files that are being complained about, are being tested with embedded tests.

multi-stage-query/src/main/java/org/apache/druid/msq/counters/CounterSnapshotsTree.java
```diff
@@ -109,6 +109,31 @@ private void putAll(final Map<Integer, Map<Integer, CounterSnapshots>> otherMap)
...
+  {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    final CounterSnapshotsTree that = (CounterSnapshotsTree) o;
+    synchronized (snapshotsMap) {
+      synchronized (that.snapshotsMap) {
```
clintropolis: is this deadlocky if multiple things are calling equals? i'm nervous at least

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/sql/DartSqlClientFactoryImpl.java
```diff
@@ -55,10 +55,16 @@ public DartSqlClient makeClient(DruidNode node)
...
-        new FixedServiceLocator(ServiceLocation.fromDruidNode(node).withBasePath(SqlResource.PATH)),
+        new FixedServiceLocator(ServiceLocation.fromDruidNode(node)
+                                               .withBasePath(removeTrailingSlash(SqlResource.PATH))),
         StandardRetryPolicy.noRetries()
     );
 
     return new DartSqlClientImpl(client, jsonMapper);
   }
+
+  private static String removeTrailingSlash(String path)
```
clintropolis: we also have `StringUtils.maybeRemoveTrailingSlash`

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/sql/DartQueryMaker.java
```diff
@@ -563,7 +579,8 @@ public void onResultsComplete()
     public void onQueryComplete(MSQTaskReportPayload report)
     {
       try {
-        complete = true;
+        this.complete = true;
+        this.report = report;
```
clintropolis: do the order these are set in matter? I can't see any overlap, but .. didn't follow all around everything that is using complete so just wanting to make sure nothing will try to get the report after checking completed

multi-stage-query/src/main/java/org/apache/druid/msq/counters/CounterSnapshotsTree.java
gianm: Hmm, I think in theory yes. `equals` is only going to be used in unit tests though so it probably doesn't matter (this was added to make a test case easier to write). But, still not great, so I'll change it to `copyMap().equals(that.copyMap())`.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/sql/DartSqlClientFactoryImpl.java
gianm: Ha, I thought there was something but couldn't find it. Updated.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/sql/DartQueryMaker.java
gianm: I don't believe it matters, but I switched the order anyway because that's easier than trying to prove it doesn't matter.

clintropolis (approved):

sql/src/main/java/org/apache/druid/sql/http/GetReportResponse.java
```diff
@@ -0,0 +1,82 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.druid.indexer.report.TaskReport;
+
+import java.util.Objects;
+
+/**
+ * Class returned by {@link SqlResource#doGetQueryReport}, the "get query report" API.
+ */
+public class GetReportResponse
```
kfaraz: Nit: rename to `GetQueryReportResponse` instead?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedMSQApis.java
```diff
@@ -150,4 +168,124 @@ public MSQTaskReportPayload runTaskSqlAndGetReport(String sql, Object... args)
...
+  public boolean cancelDartQuery(String sqlQueryId, EmbeddedBroker targetBroker)
+  {
+    final BrokerClient brokerClient = createBrokerClientForBroker(targetBroker);
+    return FutureUtils.getUnchecked(brokerClient.cancelSqlQuery(sqlQueryId), true);
+  }
+
+  /**
+   * Creates a {@link BrokerClient} that targets a specific broker using a {@link FixedServiceLocator}.
+   */
+  private static BrokerClient createBrokerClientForBroker(EmbeddedBroker targetBroker)
```
kfaraz: Nit: I think it may be useful to move this method to `EmbeddedServiceClient` as a variation of `onAnyBroker()`. The new method could be `onTargetBroker()` which accepts the target `EmbeddedBroker` and the API lambda to be called on the underlying client.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDartReportApiTest.java
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  /**
+   * Polls the report API until a report is available.
+   */
+  private GetReportResponse waitForReport(String sqlQueryId)
+  {
+    final long timeout = 30_000;
+    final long deadline = System.currentTimeMillis() + timeout;
+    while (System.currentTimeMillis() < deadline) {
```
kfaraz, line 326: Is there any "query has started running" metric that we could wait for instead of sleeping?
Nit: You could also use `Stopwatch.hasMillisElapsed()` instead of `System.currentTimeMillis()`.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/QueryInfoAndReport.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.druid.msq.dart.controller.http.DartQueryInfo;
+import org.joda.time.DateTime;
+
+import javax.annotation.Nullable;
+import java.util.Objects;
+
+/**
+ * Object returned by {@link DartControllerRegistry#getQueryInfoAndReport(String)}.
+ */
+public class QueryInfoAndReport
```
kfaraz, line 32: This class seems very similar to `GetReportResponse`. Do we need the two to be distinct?

multi-stage-query/src/main/java/org/apache/druid/msq/dart/guice/DartControllerConfig.java
```diff
@@ -42,6 +43,12 @@ public class DartControllerConfig
   @JsonProperty("heapFraction")
   private double heapFraction = DEFAULT_HEAP_FRACTION;
 
+  @JsonProperty("maxRetainedReportCount")
+  private int maxRetainedReportCount = 0;
```
kfaraz, line 51: Should we default to a small positive number, say 10? This seems like a useful feature to have out of the box.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerRegistry.java
```diff
@@ -19,54 +19,237 @@
...
+    this.config = config;
+  }
+
+  @LifecycleStart
+  public void start()
+  {
+    // Schedule periodic cleanup of expired reports.
+    if (!config.getMaxRetainedReportDuration().equals(Period.ZERO)) {
+      final String threadNameFormat = StringUtils.format("%s-ReportCleanupExec-%%s", getClass().getSimpleName());
+      final long cleanupPeriodMs = Math.max(
```
kfaraz, line 97: Just curious, is it necessary to do periodic cleanup?
Isn't it sufficient to just have a limit on the number of reports?

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerRegistry.java
```diff
@@ -19,54 +19,237 @@
...
     // Remove only if the current mapping for the queryId is this specific controller.
-    controllerMap.remove(holder.getController().queryId(), holder);
+    final boolean didRemove = controllerMap.remove(dartQueryId, holder);
+
+    // Add completeReport to completeReports, if present, and if we actually did deregister this specific controller.
+    if (didRemove && completeReport != null && config.getMaxRetainedReportCount() > 0) {
+      synchronized (completeReports) {
+        // Remove reports if size is greater than maxRetainedReportCount - 1.
+        int reportsToRemove = completeReports.size() - config.getMaxRetainedReportCount() + 1;
+        if (reportsToRemove > 0) {
```
kfaraz, line 149: We might be able to get rid of this logic by overriding the `removeEldestEntry` method on the `LinkedHashMap completeReports`.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerRegistry.java
```diff
@@ -19,54 +19,237 @@
...
+
     // Remove only if the current mapping for the queryId is this specific controller.
-    controllerMap.remove(holder.getController().queryId(), holder);
+    final boolean didRemove = controllerMap.remove(dartQueryId, holder);
+
+    // Add completeReport to completeReports, if present, and if we actually did deregister this specific controller.
+    if (didRemove && completeReport != null && config.getMaxRetainedReportCount() > 0) {
+      synchronized (completeReports) {
+        // Remove reports if size is greater than maxRetainedReportCount - 1.
+        int reportsToRemove = completeReports.size() - config.getMaxRetainedReportCount() + 1;
```
kfaraz, line 148: Since we are acquiring a lock before this, I suppose this will always be <= 1.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/msq/EmbeddedDartReportApiTest.java
gianm: Not really, the metrics are emitted when a query finishes. This could likely be refactored to use the `waitForResult` utility in https://github.com/apache/druid/pull/18766#issuecomment-3722903833 once that PR is merged.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/QueryInfoAndReport.java
gianm: I didn't want to have an unnecessary `timestamp` field polluting the object that represents what the API returns. This class has the `timestamp` so it can be used for expiration of reports.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/guice/DartControllerConfig.java
gianm: I was being conservative here with the new functionality, and also thinking that a small number might not be that useful. A cluster that gets real usage will probably go through 10 queries pretty quickly. In a production situation it probably makes sense to set it much higher and to have a time-based expiration of at least a few minutes.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerRegistry.java
gianm: I was thinking that in a production scenario, an admin probably wants to think about things in terms of time-based expiration (i.e. I'm going to have reports available for 5 minutes). I was thinking the count-based retention is more just to protect against the broker going OOM if there is a burst of queries. It would be better for the limit to be footprint-based, but the code for that is more complex :|

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerRegistry.java
gianm: True but I always find that approach ugly.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerRegistry.java
gianm: I think so too, just being defensive.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/DartControllerRegistry.java
kfaraz: That makes sense. thanks!

PR #18891, opened 2026-01-08
Upgrade to AWS SDK V2

### LINKED ISSUE #16903: AWS SDK 1.x EOL - Migrate AWS SDK for Java from 1.x to 2.x
### Description

Druid uses a deprecated version of the AWS SDK Java, [1.x](https://github.com/aws/aws-sdk-java). AWS is requesting developers to move to the new version, [2.x](https://github.com/aws/aws-sdk-java-v2).

Those are two different packages: 1.x is `com.amazonaws` and 2.x is software.amazon.awssdk.

This issue isn't a detailed design, rather a placeholder to raise awareness. There will need to be additional work to come up with the required design and code changes.

### Motivation

The AWS SDK for Java 1.x is in maintenance mode since July 31, 2024, and will enter End-of-Support on December 31, 2025.
Moreover, new AWS SDK features are no longer implemented for 1.x. One example is [Kinesis dedicated throughput (enhanced fan-out)](https://docs.aws.amazon.com/streams/latest/dev/building-enhanced-consumers-kcl.html).

jtuglu1:
Closes #16903.

### Description

Upgrade to AWS SDK v2.40.0. Looks like dependencies hdfs-storage + ranger have transitive dependencies on V1. Those will need to be excluded or upgraded as well. Core modules/extensions required for runtime are not reliant on V1.

Migrated to strictly synchronous APIs (to maintain backwards-compatibility with V1). A follow-up PR should be able to make use of async APIs where suitable. S3TransferManager is the only case where an async client may be used.

#### New Tests
- This adds `IngestionRollingUpgradeDockerTest` which simulates ingestion of Druid segments using differing Druid versions, verifying that indexing processes can talk to each other and brokers/historicals can read the segments.

#### Hiccups

1. `KinesisAggregatorV2` not available in Maven – https://github.com/awslabs/kinesis-aggregation/issues/120. Doesn't look like anything uses `aggregation=true`, so added a warning message. Seems like https://github.com/apache/flink-connector-aws/pull/188 also ran into similar issues, but implemented their own unmerged approach. Per the warning on the https://github.com/awslabs/kinesis-aggregation repo, looks like data loss can occur anyways when using this mode: https://github.com/awslabs/kinesis-aggregation/blob/master/potential_data_loss.md so don't think that's really a desired feature anyways.

#### Release note

Upgrade to AWS SDK v2.40.0

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/KubernetesClusterDockerTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/k8s/KubernetesClusterDockerTest.java
@@ -36,7 +36,8 @@ public class KubernetesClusterDockerTest extends IngestionSmokeTest implements L
   protected EmbeddedDruidCluster addServers(EmbeddedDruidCluster cluster)
   {
     final K3sDruidService brokerService = new K3sDruidService(DruidCommand.Server.BROKER)
-        .addProperty("druid.sql.planner.metadataRefreshPeriod", "PT1s");
+        .addProperty("druid.sql.planner.metadataRefreshPeriod", "PT1s")
+        .addProperty("druid.sql.planner.metadataSegmentPollPeriod", "1000");
 
     // Create a K3s cluster with all the required services
     final K3sClusterResource k3sCluster = new K3sClusterResource()


pom.xml
```diff
@@ -123,6 +123,7 @@
         <graaljs.version>22.3.5</graaljs.version>
         <mockito.version>5.14.2</mockito.version>
         <aws.sdk.version>1.12.784</aws.sdk.version>
+        <aws.sdk.v2.version>2.28.28</aws.sdk.v2.version>
```
cryptoe: Should we upgrade to the latest aws v2 sdk version ?

jtuglu1: Yes I will bump it to 2.40 minor release. Wanted to minimize chances of upstream bugs by using a stable version for testing.

cloud/aws-common/src/main/java/org/apache/druid/common/aws/AWSCredentialsUtils.java
```diff
@@ -19,26 +19,28 @@
...
+import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.InstanceProfileCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.ProfileCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.SystemPropertyCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.WebIdentityTokenFileCredentialsProvider;
 
 public class AWSCredentialsUtils
 {
-  public static AWSCredentialsProviderChain defaultAWSCredentialsProviderChain(final AWSCredentialsConfig config)
+  public static AwsCredentialsProvider defaultAWSCredentialsProviderChain(final AWSCredentialsConfig config)
```
Fly-Style, line 33: Nit: there is also a `AwsCredentialsProviderChain.of(...)` method, it might look cleaner.

cloud/aws-common/src/test/java/org/apache/druid/common/aws/AWSClientUtilTest.java
```diff
@@ -33,75 +31,79 @@ public class AWSClientUtilTest
...
   }
 
   @Test
   public void testRecoverableException_RequestTimeout()
   {
-    AmazonServiceException ex = new AmazonServiceException(null);
-    ex.setErrorCode("RequestTimeout");
+    AwsServiceException ex = AwsServiceException.builder()
+        .message("RequestTimeout")
+        .awsErrorDetails(software.amazon.awssdk.awscore.exception.AwsErrorDetails.builder()
```
Fly-Style: Nit: is that prefix really necessary?

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3ServerSideEncryption.java
```diff
@@ -19,29 +19,20 @@
...
   @Override
-  public CopyObjectRequest decorate(CopyObjectRequest request)
+  public CopyObjectRequest.Builder decorate(CopyObjectRequest.Builder builder)
   {
-    final ObjectMetadata objectMetadata = request.getNewObjectMetadata() == null ?
-                                          new ObjectMetadata() :
-                                          request.getNewObjectMetadata().clone();
-    objectMetadata.setSSEAlgorithm(ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
-    return request.withNewObjectMetadata(objectMetadata);
+    return builder.serverSideEncryption(software.amazon.awssdk.services.s3.model.ServerSideEncryption.AES256);
```
Fly-Style: Nit: do we need this prefix?

Fly-Style:
Very good job @jtuglu1! I have no objections, PR looks good from my perspective!

jtuglu1:
👍 I'd still like to add 1-2 ITs that fully cover the newer S3 changes, but overall think it's basically there.

cryptoe:
Due to the criticality of the change, I think we should get 2 approvals for this PR. 
cc @jtuglu1

cloud/aws-common/src/main/java/org/apache/druid/common/aws/FileSessionCredentialsProvider.java
```diff
@@ -53,13 +53,12 @@ public FileSessionCredentialsProvider(String sessionCredentialsFile)
...
   @Override
-  public AWSCredentials getCredentials()
+  public AwsCredentials resolveCredentials()
   {
     return awsSessionCredentials;
   }
 
-  @Override
-  public void refresh()
+  private void refresh()
```
cryptoe, line 71: Who calls the refresh method now ?

extensions-core/kinesis-indexing-service/src/main/java/org/apache/druid/indexing/kinesis/KinesisRecordSupplier.java
```diff
@@ -474,45 +459,88 @@ public KinesisRecordSupplier(
...
+    }
+
+    return builder.build();
   }
 
+  /**
+   * Parse region from a Kinesis endpoint URL.
+   * Expected format: https://kinesis.{region}.amazonaws.com
+   */
+  private static Region parseRegionFromEndpoint(String endpoint)
```
cryptoe: We should probably add tests for this function.

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/ObjectSummaryWithBucketIterator.java
```diff
@@ -0,0 +1,170 @@
...
+      final int maxListingLength,
+      final int maxRetries
+  )
+  {
+    this.s3Client = s3Client;
+    this.prefixesIterator = prefixes.iterator();
+    this.maxListingLength = maxListingLength;
+    this.maxRetries = maxRetries;
+
+    prepareNextRequest();
```
cryptoe: Feels a bit weird that in the constructor we are calling these methods. 
Should we call these in hasNext() ?

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/ObjectSummaryWithBucketIterator.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  private static boolean isDirectoryPlaceholder(final S3Object objectSummary)
+  {
+    // Recognize "standard" directory place-holder indications used by Amazon's AWS Console and Panic's Transmit.
+    if (objectSummary.key().endsWith("/") && objectSummary.size() == 0) {
+      return true;
+    }
+
+    // Recognize s3sync.rb directory placeholders by MD5/ETag value.
```
cryptoe: Could you please share relevant docs for these placeholders ?

jtuglu1:
Hi @cryptoe. Yes, I'm testing this on our production clusters running Kafka + S3. We don't have production clusters using Kinesis. I've only been able to run tests locally. I wonder if you folks might have some Kinesis clusters to test the new logic out on?

cryptoe:
> Hi @cryptoe. Yes, I'm testing this on our production clusters running Kafka + S3. We don't have production clusters using Kinesis. I've only been able to run tests locally.


Thanks for this. Do keep us posted about the findings. 

>  I wonder if you folks might have some Kinesis clusters to test the new logic out on?

We also don't have active kinesis clusters but let me get back to you on this.

jtuglu1:
Hi, update here:

I've left this internally on a few of our clusters and things are working fine. Confirmed I've tested a superset of the following:

| # | Test | Status |
|---|------|--------|
| 1 | Native Batch Ingestion (index_parallel) | DONE |
| 2 | Streaming Ingestion (index_kafka) | DONE |
| 3 | Streaming Ingestion (index_kinesis) | TODO |
| 4 | MSQ SELECT with S3 Durable Storage | DONE |
| 5 | MSQ INSERT with S3 Shuffle Storage | DONE |
| 6 | Segment Mark as Unused + Kill Task + Nuke Segments | DONE |
| 7 | Task Log Deletion from S3 | DONE |

I will see if I can get the kinesis tested, but otherwise things are looking good. I will address rest of comments and then should be ready to go.

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/ObjectSummaryWithBucketIterator.java
jtuglu1: Sure, I believe Gian added these a while back: https://github.com/apache/druid/pull/9098#discussion_r362036658

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/ObjectSummaryWithBucketIterator.java
jtuglu1: I've moved out of ctor and into an initialize method. Iterators are expected to be single-threaded access, I'll make the initialized indicator thread-safe just in case.

cloud/aws-common/src/main/java/org/apache/druid/common/aws/FileSessionCredentialsProvider.java
jtuglu1: This refresh() is done now solely through a preexisting scheduled on the class. This is equivalent behavior as before, where `FileSessionCredentialsProvider` handles its own refresh internally on its scheduler. I don't believe the standalone method was used explicitly either; the v1 `AWSCredentialsProviderChain` didn't proactively call refresh() on providers during getCredentials() resolution — it just iterated through providers catching exceptions (from what I can tell).

cryptoe:
Thanks @jtuglu1 for your comments. 
While thinking through this more, I was also thinking if we have covered rolling upgrade tests. 

* ie Ingestion running on newer aws sdk and historicals running with older sdk are able to pull segments created with the new s3 client
* MSQ tasks do share shuffle files via s3. So what happens if task (on old aws sdk) tries to write data and task (on new sdk) tries to read data.

jtuglu1:
> ie Ingestion running on newer aws sdk and historicals running with older sdk are able to pull segments created with the new s3 client

Yes – I have tested both v2-written segments being read by v1 and vice-versa.

> MSQ tasks do share shuffle files via s3. So what happens if task (on old aws sdk) tries to write data and task (on new sdk) tries to read data.

I have not tested this yet – we currently deploy in a R/B fashion where there are never old version/new version tasks running at the same time. I have tested S3 shuffle via MSQ queries (both select and insert) on the same version tasks, as well as S3 shuffle for compaction tasks. I can write some backwards-compatibility docker (embedded) tests for this?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionRollingUpgradeDockerTest.java
```diff
@@ -0,0 +1,278 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Tests that ingestion works correctly during a rolling upgrade, where the
+ * cluster has a mix of old-version and new-version indexers (the assumption is that job logic is shared betwen indexers and MM/Peons, indexers are just easier to use in tests).
```
kfaraz:
```suggestion
 * Tests that ingestion works correctly during a rolling upgrade, by using a
 * cluster which has a Historical and an Indexer on old version and
 * another Indexer and the other services on the latest version.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionRollingUpgradeDockerTest.java
```diff
@@ -0,0 +1,278 @@
...
+ * Validates that segments and intermediate shuffle data written to S3 by one version can be
```
kfaraz: Please remove the rest of this javadoc as all of that information is already captured in the tests themselves.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionRollingUpgradeDockerTest.java
```diff
@@ -0,0 +1,278 @@
...
+        .addExtensions(S3StorageConnectorModule.class)
+        .addResource(new PostgreSQLMetadataResource())
+        .addResource(storageResource)
+        .addResource(durableStorageResource)
+        .addResource(containerIndexer)
+        .addServer(new EmbeddedCoordinator())
+        .addServer(overlord)
+        .addServer(embeddedIndexer)
+        .addServer(broker)
+        .addServer(new EmbeddedHistorical())
```
kfaraz: I feel Historical should also be on Apache 33 version since in a rolling upgrade, historicals are upgraded first.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionRollingUpgradeDockerTest.java
```diff
@@ -0,0 +1,278 @@
...
+        .usingImage(DruidContainer.Image.APACHE_33)
+        .addMount(testDataMount);
+
+    return EmbeddedDruidCluster
+        .withZookeeper()
+        .useContainerFriendlyHostname()
+        .useLatchableEmitter()
+        .useDefaultTimeoutForLatchableEmitter(60)
+        .addExtensions(S3StorageConnectorModule.class)
+        .addResource(new PostgreSQLMetadataResource())
```
kfaraz: you could probably get away with the embedded Derby itself since coordinator and overlord are still embedded.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionRollingUpgradeDockerTest.java
```diff
@@ -0,0 +1,278 @@
...
+    createdDataSources.clear();
+  }
+
+  /**
+   * Submits two native batch IndexTasks concurrently to separate datasources.
+   * With {@code EqualDistributionWorkerSelectStrategy} and equal capacity on
+   * both indexers, the two tasks are guaranteed to land on different indexers.
+   * Verifies that segments produced by both old and new indexers are loaded
+   * and queryable by the Historical.
+   */
```
kfaraz: Nit: try to include (atleast some of) this info in the test name itself.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionRollingUpgradeDockerTest.java
```diff
@@ -0,0 +1,278 @@
...
+    cluster.callApi().verifySqlQuery("SELECT * FROM %s", ds1, Resources.InlineData.CSV_10_DAYS);
+    cluster.callApi().verifySqlQuery("SELECT * FROM %s", ds2, Resources.InlineData.CSV_10_DAYS);
+  }
+
+  /**
+   * Submits an {@code index_parallel} task with 3 local input files and
+   * {@code maxNumConcurrentSubTasks=3}. Using {@code MaxSizeSplitHintSpec(1)}
+   * forces one subtask per file. The supervisor occupies one slot on one
+   * indexer, leaving 3 slots across 2 indexers for the subtasks -- guaranteeing
+   * that subtasks land on both the old and new indexer.
```
kfaraz: Same here, it is nicer to have the test names be more descriptive rather than putting all the details in javadocs.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/StreamIndexDataFormatsTestBase.java
```diff
@@ -84,6 +84,7 @@
  *
  * This tests both InputFormat and Parser. Parser is deprecated for Streaming Ingestion,
  * and those tests will be removed in the future.
+ *
```
kfaraz: Nit: extra newline

pom.xml
```diff
@@ -119,7 +119,8 @@
         <hadoop.compile.version>3.3.6</hadoop.compile.version>
         <graaljs.version>22.3.5</graaljs.version>
         <mockito.version>5.14.2</mockito.version>
-        <aws.sdk.version>1.12.784</aws.sdk.version>
+        <aws.sdk.v1.version>1.12.784</aws.sdk.v1.version>
```
kfaraz: Please add a comment on where v1 is still used and if there is a plan to eventually get rid of it.
I guess it is only used with Hadoop, which we should be able to remove soon in #19109 .

jtuglu1:
@kfaraz any thoughts here?

kfaraz:
Yes, @jtuglu1 , I am almost done going through the files. A few remain, after which I will approve and merge this PR tomorrow.

Everything looks good so far.

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3Utils.java
```diff
@@ -65,6 +64,42 @@ public class S3Utils
   private static final Joiner JOINER = Joiner.on("/").skipNulls();
   private static final Logger log = new Logger(S3Utils.class);
 
+  /**
+   * A holder for S3Object with its associated bucket name.
+   * In AWS SDK v2, S3Object doesn't include the bucket name, so we need to track it separately.
+   */
+  public static class S3ObjectWithBucket
```
kfaraz: This should be in a separate file and not inside the `S3Utils` class.

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3Utils.java
```diff
@@ -229,11 +288,42 @@ static String constructSegmentBasePath(String baseKey, String storageDir)
...
   {
-    final AccessControlList acl = s3Client.getBucketAcl(bucket);
-    acl.grantAllPermissions(new Grant(new CanonicalGrantee(acl.getOwner().getId()), Permission.FullControl));
-    return acl;
+    final String ownerId = s3Client.getBucketAcl(bucket).owner().id();
+    return Grant.builder()
+        .grantee(Grantee.builder()
+            .type(Type.CANONICAL_USER)
+            .id(ownerId)
+            .build())
```
kfaraz: nit: formatting here

jtuglu1:
Checking in on this – @kfaraz the failing tests pass locally for me. Going to see if these are flakes.

jtuglu1:
@kfaraz bumping this

jtuglu1:
@gianm would you be able to take a look at this? Would like to get this merged as this branch is starting to accumulate conflicts

kfaraz:
I have had this PR half-reviewed for more than a week now.
My backlog of reviews is pretty much done now, I should be able to finally get back to this.

I will go through the major files tomorrow and most likely approve the PR.
Any other minor improvement can always be done later.

cc: @jtuglu1

gianm:
> @gianm would you be able to take a look at this? Would like to get this merged as this branch is starting to accumulate conflicts

I am hoping @kfaraz can finish out the review.

cloud/aws-common/src/main/java/org/apache/druid/common/aws/AWSClientConfig.java
```diff
@@ -56,6 +72,21 @@ public boolean isForceGlobalBucketAccessEnabled()
     return forceGlobalBucketAccessEnabled;
   }
 
+  public int getConnectionTimeout()
+  {
+    return connectionTimeout;
+  }
+
+  public int getSocketTimeout()
```
kfaraz: Maybe rename these to `getXyzTimeoutMillis()` to avoid ambiguity.
Other millis-related configs typically use the type `long` but I guess these fields will never actually need to be that large.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/minio/ITS3ToS3ParallelIndexTest.java
```diff
@@ -40,7 +39,11 @@ public void testS3IndexData(Pair<String, List<?>> s3InputSource) throws Exceptio
   @MethodSource("resources")
   public void testS3IndexData_withTempCredentials(Pair<String, List<?>> s3InputSource) throws Exception
   {
-    final S3InputSourceConfig inputSourceConfig = minIOStorageResource.createTempCredentialsForInputSource();
-    doTest(s3InputSource, new Pair<>(false, false), "s3", inputSourceConfig);
+    doTestWithEndpointConfig(
```
kfaraz, line 42: We should probably add a separate test method instead of updating this one.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/minio/MinIOStorageResource.java
```diff
@@ -126,62 +127,72 @@ public String getEndpointUrl()
...
 
-  private AmazonS3 createS3Client()
+  private S3Client createS3Client()
   {
-    return AmazonS3Client
-        .builder()
-        .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(getEndpointUrl(), "us-east-1"))
-        .withCredentials(new AWSStaticCredentialsProvider(new BasicAWSCredentials(getAccessKey(), getSecretKey())))
-        .withPathStyleAccessEnabled(true)
+    return S3Client.builder()
```
kfaraz: Nit: Please move `.builder()` to the next line in all such places. This currently does not adhere to the Druid style formatting.

extensions-core/druid-aws-rds-extensions/src/main/java/org/apache/druid/aws/rds/AWSRDSTokenPasswordProvider.java
```diff
@@ -98,18 +99,16 @@ public String getRegion()
   public String getPassword()
   {
     try {
-      RdsIamAuthTokenGenerator generator = RdsIamAuthTokenGenerator
-          .builder()
-          .credentials(awsCredentialsProvider)
-          .region(region)
+      RdsUtilities rdsUtilities = RdsUtilities.builder()
```
kfaraz: Nit: `.builder()` to next line to avoid unnecessary reformats later.

extensions-core/kinesis-indexing-service/src/test/java/org/apache/druid/indexing/kinesis/KinesisRecordSupplierTest.java
```diff
@@ -140,37 +138,34 @@ private static ByteBuffer jb(String timestamp, String dim1, String dim2, String
...
+  {
+    return Record.builder()
+        .data(SdkBytes.fromByteBuffer(data.duplicate()))
+        .sequenceNumber(sequenceNumber)
+        .partitionKey("key")
+        .approximateArrivalTimestamp(Instant.now())
+        .build();
+  }
+
+  private static KinesisClientRecord toKinesisClientRecord(Record v2Record)
```
kfaraz: Nit: Calling the record v2Record might be confusing since we are not using v1 anywhere anymore (are we?).

extensions-core/kinesis-indexing-service/src/test/java/org/apache/druid/indexing/kinesis/KinesisRecordSupplierTest.java
```diff
@@ -616,132 +559,49 @@ public void testSeekUnassigned() throws InterruptedException
     recordSupplier.seekToEarliest(Collections.singleton(shard0));
   }
 
-
-  @Test
-  public void testPollAfterSeek()
```
kfaraz, line 621: Is this test not applicable anymore with v2?

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3TaskLogs.java
```diff
@@ -128,16 +129,18 @@ private Optional<InputStream> streamTaskFile(final long offset, String taskKey)
...
       }
 
-      final GetObjectRequest request = new GetObjectRequest(config.getS3Bucket(), taskKey)
-          .withMatchingETagConstraint(ensureQuotated(objectMetadata.getETag()))
-          .withRange(start, end);
+      GetObjectRequest.Builder requestBuilder = GetObjectRequest.builder()
+          .bucket(config.getS3Bucket())
+          .key(taskKey)
+          .ifMatch(ensureQuotated(objectMetadata.eTag()))
+          .range(StringUtils.format("bytes=%d-%d", start, end));
```
kfaraz: For the range stuff, I think it would be useful to write an `AwsBytesRange` utility class that takes optional start, end as inputs and exposes a method `String getBytesRange()`. That would help avoid doing the String formatting in multiple places and keep the code less error prone.

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
```diff
@@ -147,65 +152,69 @@ public void kill(List<DataSegment> segments) throws SegmentLoadingException
...
             "Deleting the following segment files from S3 bucket[%s]: [%s]",
             s3Bucket,
             keysToDeleteStrings
         );
         S3Utils.retryS3Operation(
             () -> {
-              s3Client.deleteObjects(deleteObjectsRequest);
+              DeleteObjectsResponse response = s3Client.deleteObjects(deleteObjectsRequest);
+              // Check for errors in the response
+              if (response.hasErrors()) {
```
gianm: The old code for `deleteKeysForBucket`, if there was an error in a multi-delete, would set `hadException = true` and then return `true` (meaning there was an issue deleting keys). That would then cause the caller (`kill`) to throw a `SegmentLoadingException`. Now, the exception is suppressed and `kill` will treat it as a success. Please restore the old behavior.

There's a retry issue too. The old code would sometimes retry `MultiObjectDeleteException` (see my comment in `S3Utils`) due to some logic for retrying it in `retryS3Operation`. The new code does not retry, just silently treats it as a success (after logging some warnings).

It may be best to focus on fixing `S3Utils.deleteBucketKeys` and then use that here. Then we only need to get the error/retry logic right in one place.

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3Utils.java
```diff
@@ -328,18 +418,21 @@ public static void deleteObjectsInPath(
...
                                       .collect(Collectors.toList());
       log.debug("Deleting keys from bucket: [%s], keys: [%s]", bucket, keys);
     }
-    DeleteObjectsRequest deleteRequest = new DeleteObjectsRequest(bucket).withKeys(keysToDelete);
+    DeleteObjectsRequest deleteRequest = DeleteObjectsRequest.builder()
+        .bucket(bucket)
+        .delete(Delete.builder().objects(keysToDelete).build())
+        .build();
     S3Utils.retryS3Operation(() -> {
       s3Client.deleteObjects(deleteRequest);
```
gianm: Need to check `hasErrors` in the response. Related: the old `S3RETRY` checked for `MultiObjectDeleteException` and would retry in certain cases. That's lost now, please restore it. The best way I can think of to do that is to introduce our own version of `MultiObjectDeleteException` and throw it when `hasErrors` is true, and then add our version to `S3RETRY`.

Please add a test for this scenario too.

For extra credit only retry the objects that failed to be deleted!

extensions-core/s3-extensions/src/test/java/org/apache/druid/storage/s3/S3DataSegmentKillerTest.java
```diff
@@ -438,20 +453,20 @@ public void test_kill_listOfSegments_amazonServiceExceptionExceptionIsThrown()
   @Test
   public void test_kill_listOfSegments_retryableExceptionThrown() throws SegmentLoadingException
```
gianm, line 501: This test was formerly testing specifically the partial failure case. Please add back a test that focuses on this case. In v2 it will show up as `hasErrors` on the response.

extensions-core/s3-extensions/src/test/java/org/apache/druid/storage/s3/S3DataSegmentKillerTest.java
```diff
@@ -359,16 +365,19 @@ public void test_kill_listOfSegments() throws SegmentLoadingException
...
     s3Client.deleteObjects(EasyMock.anyObject(DeleteObjectsRequest.class));
-    MultiObjectDeleteException.DeleteError deleteError = new MultiObjectDeleteException.DeleteError();
-    deleteError.setKey(KEY_1_PATH);
-    MultiObjectDeleteException multiObjectDeleteException = new MultiObjectDeleteException(
-        ImmutableList.of(deleteError),
-        ImmutableList.of());
-    EasyMock.expectLastCall().andThrow(multiObjectDeleteException).once();
+    // In v2 SDK, multi-object delete errors are returned in the response, not thrown as exceptions
+    // For testing purposes, we simulate an S3Exception being thrown
+    S3Exception s3Exception = (S3Exception) S3Exception.builder()
```
gianm: This isn't right, we should test what the v2 SDK actually does. The test should use a regular `DeleteObjectsResponse` with errors set.

extensions-core/s3-extensions/src/test/java/org/apache/druid/storage/s3/S3DataSegmentKillerTest.java
```diff
@@ -383,22 +392,26 @@ public void test_kill_listOfSegments_multiDeleteExceptionIsThrown()
   @Test
   public void test_kill_listOfSegments_multiDeleteExceptionIsThrownMultipleTimes()
   {
-    DeleteObjectsRequest deleteObjectsRequest = new DeleteObjectsRequest(TEST_BUCKET);
```
gianm, line 386: Same issue as `test_kill_listOfSegments_multiDeleteExceptionIsThrown`. We should use a regular `DeleteObjectsResponse` with errors set.

gianm:
The latest updates look good to me as far as resolving the concerns from the limited pass I did-- thanks @jtuglu1! I am stopping short of doing a full ✅ because I only did that limited pass. The full review I am leaving to @kfaraz.

jtuglu1:
> The latest updates look good to me as far as resolving the concerns from the limited pass I did-- thanks @jtuglu1! I am stopping short of doing a full ✅ because I only did that limited pass. The full review I am leaving to @kfaraz.

Thanks. I addressed Kashif's comments as well. I just need to get docker tests to pass then I will merge.

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
```diff
@@ -94,32 +97,34 @@ public void kill(List<DataSegment> segments) throws SegmentLoadingException
...
           k -> new ArrayList<>()
       );
       if (path.endsWith("/")) {
         // segment is not compressed, list objects and add them all to delete list
-        final ListObjectsV2Result list = s3Client.listObjectsV2(
-            new ListObjectsV2Request().withBucketName(s3Bucket).withPrefix(path)
-        );
-        for (S3ObjectSummary objectSummary : list.getObjectSummaries()) {
-          keysToDelete.add(new DeleteObjectsRequest.KeyVersion(objectSummary.getKey()));
+        ListObjectsV2Request request = ListObjectsV2Request.builder()
```
kfaraz: Nit: The `.builder()` in several of the files will lead to unnecessary reformats later. Please check if they can be moved to the next line in all such places.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/minio/ITS3ToS3ParallelIndexTest.java
kfaraz: Please avoid resolving comments if they are not being addressed. 🙂 
It makes it difficult for the reviewer to figure out why a change should not be made.

In such cases, it is best to add a reply clarifying why the change is not needed and leave the comment unresolved to keep the discussion open.

jtuglu1: Oops – missed this one. There's a bunch of changes in this PR so been a bit annoying to keep track of them, apologies.

> We should probably add a separate test method instead of updating this one.

The original test would break w/o this update. The reason is SDK v2 requires some ARN and MinIO will basically just check whether the ARN is valid or not (which includes needing a signing URL).

kfaraz: No worries! Thanks for the clarification!

Issue #7124, opened 2019-02-22
rework segment metadata query "size" analysis


clintropolis:
# Motivation

Currently segment metadata query has a `size` analysis type that the documentation describes as "estimated byte size for the segment columns if they were stored in a flat format" and "estimated total segment byte size in if it was stored in a flat format", but this doesn't really have any practical value as far as I can tell. I think this value is confusing, and should be replaced with a `size` value that represents the actual segment and column sizes for mapped segments, or the estimated size in memory for incremental segments. 

This allows size analysis to be useful for finding which columns are the heavy hitters in terms of overall segment size, observe fluctuations in segment size over time, and could also be of aid in capacity planning.

Alternatively, if this doesn't seem useful enough, or more trouble than it's worth, I would instead propose that the value just be removed completely because it's confusing and expensive to run.

# Proposed changes

Making the value meaningful likely doesn't take much effort. For mapped columns, it would just be preserving the byte size of the columns when initially loading segments and making it accessible through the `BaseColumn` or related interface (if it isn't already and I just missed it). For incremental index, it would involve modifying the size functions to report the estimated size in memory of the values where necessary. No surface level changes would be necessary for this approach, but we would need to call out in the release notes that the meanings of values has changed and update documentation accordingly.

# Rationale

We already need to know this size information to properly map columns from the 'smoosh' file, so preserving it and offering it up for segment metadata queries should be rather straightforward.

# Operational impact

Size analysis as is looks rather expensive, if we switch to column "in segment" size, then this operation can become constant for mapped segments resulting in cheaper overall segment metadata queries using this analysis on mapped segments. I expect little change for the case of incremental indexes.

drcrallen:
Can such a feature allow comparisons to https://cloud.google.com/bigquery/pricing ?

clintropolis:
> Can such a feature allow comparisons to https://cloud.google.com/bigquery/pricing ?

I'm not sure, I think it would depend on which of "size data requires to be loaded on historicals" or "size data requires to be at rest in deep storage" is the more appropriate metric to compare to the big query thing. My suggestion would provide a means to get the former value, which I think is maybe the better thing to use in Druid world to compare here since data needs to be loaded to be useful, so ... yes I guess? 😅That said, you might need both values if you also wanted to also meter segments that are in deep storage and _not_ loaded, or are trying to bill for all costs.

gianm:
[...]

PR #18892, opened 2026-01-08
Add totalRows to DataSegment

cecemei:
Description
  - Add `totalRows` field to `DataSegment`, and use it as source of truth for `totalRows`. Not using `numRows` since it's already used in `SegmentStatusInCluster` and it unwraps `DataSegment` when serialized. 
  - Refactor `DataSegment` construction to use builder pattern for better maintainability.

Key Changes
  - `DataSegment` 
  - `DataSegmentWithLocation`
  - `LoadableDataSegment`
  - `SystemSchema`
  - `BatchAppenderator`
  - `StreamAppenderator`
  - `DataSegmentWithLocationTest`
  - `LoadableDataSegmentTest`
  - `SegmentStatusInClusterTest`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/segment/realtime/appenderator/BatchAppenderator.java
+++ b/server/src/main/java/org/apache/druid/segment/realtime/appenderator/BatchAppenderator.java
@@ -841,7 +841,7 @@ public class BatchAppenderator implements Appenderator
         mergedSegment =
             sink.getSegment()
                 .toBuilder()
-                .dimensions(IndexMerger.getMergedDimensionsFromQueryableIndexes(indexes, schema.getDimensionsSpec()))
+                .dimensions(Lists.newArrayList(index.getAvailableDimensions().iterator()))
                 .totalRows(index.getNumRows())
                 .build();
       }
--- a/server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
+++ b/server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
@@ -976,7 +976,7 @@ public class StreamAppenderator implements Appenderator
         mergedSegment =
             sink.getSegment()
                 .toBuilder()
-                .dimensions(IndexMerger.getMergedDimensionsFromQueryableIndexes(indexes, schema.getDimensionsSpec()))
+                .dimensions(Lists.newArrayList(index.getAvailableDimensions().iterator()))
                 .totalRows(index.getNumRows())
                 .build();
       }
--- a/processing/src/test/java/org/apache/druid/timeline/SegmentStatusInClusterTest.java
+++ b/processing/src/test/java/org/apache/druid/timeline/SegmentStatusInClusterTest.java
@@ -41,7 +41,7 @@ public class SegmentStatusInClusterTest
   private static final ImmutableMap<String, Object> LOAD_SPEC = ImmutableMap.of("something", "or_other");
   private static final boolean OVERSHADOWED = true;
   private static final Integer REPLICATION_FACTOR = 2;
-  private static final Integer NUM_ROWS = 10;
+  private static final Integer TOTAL_ROWS = 10;
   private static final boolean REALTIME = true;
   private static final int TEST_VERSION = 0x9;
   private static final SegmentStatusInCluster SEGMENT = createSegmentForTest();
@@ -65,7 +65,7 @@ public class SegmentStatusInClusterTest
                                          .loadSpec(LOAD_SPEC)
                                          .binaryVersion(TEST_VERSION)
                                          .size(1)
-                                         .totalRows(NUM_ROWS)
+                                         .totalRows(TOTAL_ROWS)
                                          .build();
     return new SegmentStatusInCluster(dataSegment, OVERSHADOWED, REPLICATION_FACTOR, null, REALTIME);
   }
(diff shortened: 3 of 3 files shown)


processing/src/main/java/org/apache/druid/segment/SegmentReference.java
```diff
@@ -49,23 +51,51 @@ public static SegmentReference missing(SegmentDescriptor segmentDescriptor)
...
       SegmentDescriptor segmentDescriptor,
       Optional<Segment> segmentReference,
       @Nullable Closeable cleanupHold
   )
   {
+    this(null, segmentDescriptor, segmentReference, cleanupHold);
+  }
+
+  private SegmentReference(
+      @Nullable DataSegment segment,
```
clintropolis: I'm unsure that this should be here, at least this change should be made in a separate PR since it doesn't seem directly related to the row count changes on `DataSegment`

processing/src/main/java/org/apache/druid/query/metadata/SegmentMetadataQueryRunnerFactory.java
```diff
@@ -67,6 +68,59 @@ public SegmentMetadataQueryRunnerFactory(
...
+      public Sequence<SegmentAnalysis> run(QueryPlus<SegmentAnalysis> inQ, ResponseContext responseContext)
+      {
+        SegmentMetadataQuery updatedQuery = ((SegmentMetadataQuery) inQ.getQuery()).withFinalizedAnalysisTypes(toolChest.getConfig());
+        final SegmentAnalyzer analyzer = new SegmentAnalyzer(updatedQuery.getAnalysisTypes());
+
+        Integer numRows = segmentRef.getDataSegment().getTotalRows();
+        if (numRows == null) {
+          numRows = analyzer.numRows(segment);
+        }
+        long totalSize = analyzer.analyzingSize() ? segmentRef.getDataSegment().getSize() : 0L;
```
clintropolis: this is is a breaking change to size analysis. While the current size analysis is arguably nonsense, I don't think we want to do an incompatible change like this, see discussion in https://github.com/apache/druid/issues/7124, and so this change should be reverted.

I'm also not terribly sure the total size isn't particularly useful, I think the size analysis we want to replace this with would be a breakdown of column parts since that is actionable (e.g. can decide to not have indexes or whatever else). In other words, I don't think `DataSegment` is particularly useful to use here since we already have the `Segment`, so it doesn't save us anything, and costs us dragging around the overhead of an extra memory reference linking back to the `DataSegment` for anything a query engine processes

processing/src/main/java/org/apache/druid/segment/SegmentReference.java
clintropolis: this change feels controversial to me, I don't love the idea of dragging around a reference to `DataSegment` everywhere we are using segments, what is this for. just for the segment metadata query changes?

server/src/main/java/org/apache/druid/segment/realtime/appenderator/BatchAppenderator.java
```diff
@@ -835,6 +836,14 @@ private DataSegmentWithMetadata mergeAndPush(
         metrics.setMergeTime(mergeTimeMillis);
 
         log.debug("Segment[%s] built in %,dms.", identifier, mergeTimeMillis);
+        QueryableIndex index = indexIO.loadIndex(mergedFile);
+        closer.register(index);
```
clintropolis, line 840: can this just use `sink.getNumRows()` or is that not cool for some reason?

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -970,6 +971,14 @@ private DataSegmentWithMetadata mergeAndPush(
         metrics.setMergeTime(mergeTimeMillis);
 
         log.debug("Segment[%s] built in %,dms.", identifier, mergeTimeMillis);
+        QueryableIndex index = indexIO.loadIndex(mergedFile);
+        closer.register(index);
```
clintropolis, line 975: same comment about `sink.getNumRows()`

server/src/main/java/org/apache/druid/segment/realtime/appenderator/BatchAppenderator.java
clintropolis: oh i guess that is before merging, and would be wrong in the case of rollup, still wonder if there is some way to get it without mapping new segment since the merger knows it when writing, will look at some stuff

clintropolis: i guess we would need to change the signatures of merge/persist methods to return some `MergeResult` or something that has both the file and extra stuff like row count, i don't think we need to worry about that in this PR, though maybe something to consider in a follow-up to save us from having to load the segment

processing/src/main/java/org/apache/druid/segment/SegmentReference.java
cecemei: yea we can probably do this a better way, reverted this change

processing/src/main/java/org/apache/druid/query/metadata/SegmentMetadataQueryRunnerFactory.java
cecemei: makes sense! it'd probably be more useful when we add all the columns size breakdown stuff in the analysis and deprecate the SIZE analysis type, reverted this change.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/BatchAppenderator.java
```diff
@@ -835,6 +836,14 @@ private DataSegmentWithMetadata mergeAndPush(
...
+        mergedSegment =
+            sink.getSegment()
+                .toBuilder()
+                .dimensions(IndexMerger.getMergedDimensionsFromQueryableIndexes(indexes, schema.getDimensionsSpec()))
```
clintropolis: Oh, i think this could just use `index.getAvailableDimensions` since we are loading the merged segment now

PR #18893, opened 2026-01-08
Various CI enhancements

kgyrtkirk:
* move static-checks to separate scripts (runnable locally) + use `worker.yml` - this is to prepare them to be run them alongside with the main ci
* remove now not necessary `build` stage as we are down to just a few old style integration tests
* feels like creating a report at the run level doesn't fully match how people expect it to behave...creating one in every split I guess will feel more straightforward
* reorg a bit how additional tests are run
* default to run things with jdk21
* fix jacoco to not load 2 different versions
* upgrade commons-compress to 1.27.0 to run all the unit-tests/etc
* skip jdk17 unless `test-jdk17` label is present

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        jdk: [ "21" ]
         include:
           - name: "QTest 0/4"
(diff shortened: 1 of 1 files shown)


cryptoe:
Thanks for the change @kgyrtkirk !!

kgyrtkirk:
thank you @cryptoe for reviewing the changes!

PR #18898, opened 2026-01-09
Fix vuln-scan in cron-jobs

kgyrtkirk:
This should fix the cron job scan which was running into exceptions 
```
AnalysisException: OSS Index rate limit exceeded, disabling the analyzer
```
https://github.com/apache/druid/actions/runs/20840285873/job/59872999425


kgyrtkirk:
thank you @capistrant for taking a look!
I hope it will work :)

PR #18899, opened 2026-01-09
Update master to 37.0.0-SNAPSHOT

kgyrtkirk:
updates
* pom.xml-s
* referenced docker image version
* web-console

I don't see any matches with `git grep -F 36.0.0` 

followed the https://github.com/apache/druid/blob/master/distribution/asf-release-process-guide.md about releasing
updated  `dev/update-version-master`

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -27,7 +27,7 @@
   <parent>
     <groupId>org.apache.druid</groupId>
     <artifactId>druid</artifactId>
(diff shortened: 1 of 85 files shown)


jtuglu1:
Once this is merged we should remove JDK 17 support from automated tests, etc. (see [here](https://github.com/apache/druid/pull/18759#issuecomment-3554561707)) as Hadoop will no longer be supported.

kgyrtkirk:
thank you @capistrant for reviewing the changes!

PR #18906, opened 2026-01-12
pass test framework jsonMapper to QueryComponentSupplier.addSegmentsToWalker so IndexBuilder can use the same mapper

clintropolis:
Was doing an experiment with a test using v10 format and was running into some issues like this when using `IndexBuilder` in `addSegmentsToWalker` method:
```
java.lang.RuntimeException: com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) [Guice/ErrorInCustomProvider]: RuntimeException: InvalidTypeIdException: Please make sure to load all the necessary extensions and jars with type 'HllSketchMergeAggregatorFactory'. Could not resolve type id 'HllSketchMergeAggregatorFactory' as a subtype of `AggregatorFactory` known type ids = [cardinality, count, doubleAny, doubleFirst, doubleLast, doubleMax, doubleMean, doubleMin, doubleSum, expression, filtered, floatAny, floatFirst, floatLast, floatMax, floatMin, floatSum, grouping, histogram, hyperUnique, javascript, longAny, longFirst, longLast, longMax, longMin, longSum, singleValue, stringAny, stringFirst, stringFirstFold, stringLast, stringLastFold] (for POJO property 'aggregators')
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 5012] (through reference chain: SegmentFileMetadata["projections"]->ArrayList[0]->ProjectionMetadata["schema"]->TableProjectionSchema["aggregators"]->Object[][1])
```

This is because by default IndexBuilder just uses a test object mapper, but it is different from the mapper of the test framework that has all of the extension modules and such loaded, so to fix I just pass this mapper into the method so the `IndexBuilder` can use it


extensions-core/druid-bloom-filter/src/test/java/org/apache/druid/query/aggregation/bloom/sql/BloomFilterSqlAggregatorTest.java
```diff
@@ -78,10 +78,13 @@ public DruidModule getCoreModule()
     }
 
     @Override
-    public SpecificSegmentsQuerySegmentWalker addSegmentsToWalker(SpecificSegmentsQuerySegmentWalker walker)
+    public SpecificSegmentsQuerySegmentWalker addSegmentsToWalker(
+        SpecificSegmentsQuerySegmentWalker walker,
+        ObjectMapper jsonMapper
```
kgyrtkirk, line 83: I think it would be better to give access to the walker thru the injector 
`createEmptyWalker#createEmpty`

clintropolis: i thought about that, but also it seemed kind of weird since `SpecificSegmentsQuerySegmentWalker` doesn't otherwise need a jsonMapper and it isn't very obvious to callers that there is a mapper that they should be using for everything their extension might have set up in the component supplier compared to doing it like this where it is sort of in your face

PR #18909, opened 2026-01-12
WireTransferable result batches.

gianm:
This patch introduces a "WireTransferable" interface that is used to serialize batches of results represented by RowsAndColumns objects. This allows query implementations to control the exact manner in which their result batches are modeled, serialized, and deserialized, which enables greater efficiency.

For example, this would allow an MSQ implementation of TopN (which does not currently exist) to use a data structure between leaf and merge stages that matches what it uses in its native implementation. It would also allow this data structure to be serialized efficiently, without needing to use JSON/Smile.

WireTransferable is used in two places:

1) In ObjectMappers, when they read or write RowsAndColumns objects.

2) In MSQ channels, when frames or other batches are read or written to disk or network.

Prior to this patch, the ObjectMapper path only handled FrameRowsAndColumns, and the channel path only handled Frames, in both cases using a Frame-specific serialization format. These Frame-specific serialization formats are still used by default when the result batches are Frames. The new FrameWireTransferable format can be enabled in both places by setting
`druid.serde.rac.useLegacyFrameSerialization = false`.

As part of this change, the ReadableFrameChannels and WritableFrameChannels used by MSQ are generalized to allow transmitting any RowsAndColumns, not just Frames. When Frames are transmitted, they are enclosed in a thin FrameRowsAndColumns wrapper. In the interest of minimizing the size of the diff, the names of the classes are not changed.


processing/src/main/java/org/apache/druid/query/rowsandcols/semantic/WireTransferable.java
```diff
@@ -0,0 +1,157 @@
...
+   * A holder class for a byte array, offset and length, this exists to minimize copies.  It's pretty similar to a
+   * ByteBuffer, but is separate because jackson cannot write a ByteBuffer anyway so might as well wrap in this.
+   */
+  class ByteArrayOffsetAndLen
+  {
+    private final byte[] array;
+    private final int offset;
+    private final int length;
+
+    public ByteArrayOffsetAndLen(byte[] array, int offset, int length)
```
clintropolis, line 89: what is the use case for `offset`? Afaict it seems like it always set to 0, but trying to wrap my head around what it would mean/how it would be used if it was set to non-zero, like i guess packing multiple `WireTransferrable` into the same blob?

processing/src/main/java/org/apache/druid/frame/file/FrameFileWriter.java
```diff
@@ -95,18 +107,70 @@ public static FrameFileWriter open(
...
+   * converted to a Frame, it will be written as raw frame bytes with {@link #MARKER_FRAME}. Otherwise, it will be
+   * written using {@link WireTransferable} with {@link #MARKER_RAC}.
    *
-   * @param frame     the frame
+   * @param rac       the RowsAndColumns to write
    * @param partition partition number for a partitioned frame file, or {@link #NO_PARTITION} for an unpartitioned file.
    *                  Must be monotonically increasing.
    */
-  public void writeFrame(final Frame frame, final int partition) throws IOException
+  public void writeRAC(final RowsAndColumns rac, final int partition) throws IOException
```
clintropolis: any reason not to just call this `write`?

processing/src/main/java/org/apache/druid/frame/channel/ReadableFrameChannel.java
```diff
@@ -20,54 +20,80 @@
...
    *
-   * Before calling this method, you should check {@link #canRead()} to ensure there is a frame or
+   * Before calling this method, you should check {@link #canRead()} to ensure there is a batch of data or
    * error available.
    *
-   * @throws java.util.NoSuchElementException if there is no frame currently available
+   * @throws NoSuchElementException if there is no batch currently available
    */
-  Frame read();
+  RowsAndColumns readRAC();
```
clintropolis: it seems like this method should be called `read` and the other should be `readAsFrame` or `readFrame`

gianm: This was about diff-minimization. Changing the name of `read()` adds another chunk of changed lines at various call sites that only deal with Frames.

processing/src/main/java/org/apache/druid/frame/file/FrameFileWriter.java
gianm: I named it `writeRAC` because the old method was called `writeFrame`, but I suppose it could be just `write`.

gianm: Renamed it `write`.

processing/src/main/java/org/apache/druid/query/rowsandcols/semantic/WireTransferable.java
gianm: It's meant to be something that can be created from an onheap `ByteBuffer` without doing any copies, so it supports offset/length like position/limit. Although those aren't currently used.

clintropolis (approved):

processing/src/main/java/org/apache/druid/frame/channel/ReadableFrameChannel.java
clintropolis: yea fair, it still seems worth changing since this is now the real 'read' method; it looks like only 37 callers of `read` after the changes in this PR, so maybe not too disruptive? fine to do as a follow-up too though

gianm: Hmm, I'll just change it now.

PR #17443, opened 2024-11-04
MSQ WF: Pass a flag from broker to determine operator chain transformation

Akshat-Jain:
### Description

In https://github.com/apache/druid/pull/17433, we had to move the operator-chain-transformation logic to the MSQ worker layer to ensure backward compatibility. But it wasn't ideal since such logic shouldn't be a responsibility of the worker.

On further offline discussion, an alternative was suggested that we can pass a flag from the broker during query execution.

If MSQ controller sees the flag, then it does the operator transformation. Otherwise it doesn't do the operator transformation.

This relies on the upgrade order of the different services. This allows us to achieve backward compatibility since brokers are upgraded after indexers/middlemanagers, so passing a flag is essentially passing info whether all indexers/middlemanagers have been upgraded or not - hence helps us achieve backward compatibility.

<hr>

This PR has:

- [x] been self-reviewed.
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [ ] been tested in a test Druid cluster.

PR #18911, opened 2026-01-13
MSQ: Fix WindowOperatorQuery planning.

gianm:
PR #18875 caused a regression in WindowOperatorQuery planning, by causing non-outermost WindowOperatorQuery to be planned without `windowFunctionOperatorTransformation` (see #17443). This happened because only the outermost query has the `windowFunctionOperatorTransformation` parameter.

This patch fixes the problem by propagating the context parameter from the outermost queries to any inner WindowOperatorQuery instances.


kgyrtkirk:
close/open to get the flaky fix from master

PR #18918, opened 2026-01-14
Fix another race in getWorkerFor.

gianm:
This patch is covering another case, beyond the one covered in #18910. There was still a race when multiple workers were running and all trying to contact each other, if worker A tried to contact worker B before worker B had been set up.

The fix is to have newWorker return null rather than throw an error in the base class, and update getWorkerFor to retry on null.


jtuglu1 (approved):
Can we add a comment in the code for why this busy wait is being done (for future tests/modifications)?

gianm:
> Can we add a comment in the code for why this busy wait is being done (for future tests/modifications)?

Added a comment like this:

```
    // Wait for the worker to exist. It may not have been created or started up yet, especially if this is
    // a worker trying to contact another worker.
```

PR #18922, opened 2026-01-16
Update `TableInputSpec` to be able to handle specific segments.

cecemei:
### Description
Update `TableInputSpec` to be able to handle specific segments.

<hr>

 * `TableInputSpec`
 * `TableInputSpecTest`
 * `IndexerTableInputSpecSlicer`

<hr>

This PR has:

- [x] been self-reviewed.


multi-stage-query/src/main/java/org/apache/druid/msq/indexing/IndexerTableInputSpecSlicer.java
```diff
@@ -157,24 +161,25 @@ private Set<DataSegmentWithInterval> getPrunedSegmentSet(final TableInputSpec ta
...
         getTimeline(tableInputSpec.getDataSource(), tableInputSpec.getIntervals());
+    final Predicate<SegmentDescriptor> segmentFilter = tableInputSpec.getSegments() != null
+                                                       ? Set.copyOf(tableInputSpec.getSegments())::contains
+                                                       : Predicates.alwaysTrue();
 
     if (timeline == null) {
       return Collections.emptySet();
     } else {
+      // A segment can overlap with multiple search intervals, or even outside search intervals.
+      // The same segment can appear multiple times or 0 time, but each is also bounded within the overlapped search interval
```
capistrant, line 172: I don't think I fully follow this comment.

> or even outside search intervals

A segment outside search intervals. Is that referring to a segment that is in `tableInputSpec.getSegments()` but ends up not overlapping any search intervals and thus not get found?

> The same segment can appear multiple times or 0 time

Same idea regarding the `0 time` comment. Is that just saying that even though a segment was in `tableInputSpec.getSegments()`, it may not appear in the iterator?

multi-stage-query/src/test/java/org/apache/druid/msq/input/table/IndexerTableInputSpecSlicerTest.java
capistrant, line 1: For the case when `TableInputSpec` is instantiated with non-null intervals **and** non-null segments - For segments who are not in the interval, I understand it that they will **not** end up in the produced "slice", correct? Is there need for a UT of that? `SEGMENT4` with an interval that doesn't overlap with "2000/2001" and confirm that if we provide the interval with `SEGMENT4` in segments list, `SEGMENT4` is not in the output?

multi-stage-query/src/main/java/org/apache/druid/msq/input/table/TableInputSpec.java
```diff
@@ -58,6 +63,8 @@ public class TableInputSpec implements InputSpec
    *                     meaning that when this spec is sliced and read, the returned {@link LoadableSegment}
    *                     from {@link PhysicalInputSlice#getLoadableSegments()} are clipped to these intervals using
    *                     {@link LoadableSegment#descriptor()}.
+   * @param segments     specific segments to read, or null to read all segments in the intervals. If provided,
```
capistrant, line 66: does there need to be documentation around what happens with non-null intervals` and non-null `segments` and some/all of those segments not overlapping with anything in `intervals`? I think that is the most interesting part of adding this support for specific segments, ant potentially the most confusing

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/IndexerTableInputSpecSlicer.java
cecemei: think of it as a combination of `MultipleIntervalSegmentSpec` and `MultipleSpecificSegmentSpec`, on top of it this spec also has filter stuff. 

the 0 time is actually not a change. even before this change think if we have two intervals day0 and day1, we can have segments from both days, so we could build a `DataSegmentWithInterval` with segment from day0 and interval day1, but it doesnt matter at the end. although we usually only have 1 interval so this dont happen often.

i guess the change here is that since now segments can be user input now, so they could be anything.

multi-stage-query/src/main/java/org/apache/druid/msq/input/table/TableInputSpec.java
cecemei: this would just be the same if request year2025 data when there's only year2026 is available. msq input allows it, although compaction io config doesnt allow this (this is not in this pr though).

multi-stage-query/src/test/java/org/apache/druid/msq/input/table/IndexerTableInputSpecSlicerTest.java
cecemei: do you mean `test_sliceStatic_segmentAndIntervalFilter`, it doesnt produce slices while `test_sliceStatic_segmentFilter` does provide slice.

capistrant: ok ya my bad! I was not properly getting the assertion. 🤦‍♂️ test_sliceStatic_segmentAndIntervalFilter does the testing that I though was missing

cecemei:
> Sorry about the confusion on the test cases for the input slicer. I think this looks good

thanks for the review!

PR #18923, opened 2026-01-16
Add sys.queries table.

gianm:
The `sys.queries` table provides insight into currently-running queries. It provides the same information as the `/druid/v2/sql/queries` API. As such, it currently only works with Dart.

In this patch the table is documented, but off by default. It can be enabled by setting `druid.sql.planner.enableSysQueriesTable = true`.

This patch additionally adds an `includeComplete` parameter to `/druid/v2/sql/queries`, which is used by the implementation of the sys.queries table, to allow it to show information for recently-completed queries.

### DIFF
--- a/docs/querying/sql-metadata-tables.md
+++ b/docs/querying/sql-metadata-tables.md
@@ -341,8 +341,8 @@ SELECT * FROM sys.server_properties WHERE server='192.168.1.1:8081'
 
 :::info
  The `sys.queries` table is an experimental feature. You must enable it by setting the runtime property
- `druid.sql.planner.enableSysQueriesTable=true` on Broker processes. Currently, this table only shows
- queries from the Dart (MSQ) engine.
+ `druid.sql.planner.enableSysQueriesTable=true` on Broker processes. The main reason this table is experimental
+  is that it only shows queries from the [Dart](dart.md) engine, which is also experimental.
 :::
 
 The queries table provides information about currently running and recently completed SQL queries.
--- a/sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
+++ b/sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
@@ -1307,7 +1307,7 @@ public class SystemSchema extends AbstractSchema
       final Object[] row = new Object[QUERIES_SIGNATURE.size()];
       row[0] = queryInfo.executionId();
       row[1] = queryInfo.engine();
-      row[2] = queryInfo.state().toString();
+      row[2] = queryInfo.state();
 
       // Only serialize info if it's in the projection
       if (includeInfo) {


embedded-tests/src/test/java/org/apache/druid/testing/embedded/auth/EmbeddedBasicAuthResource.java
```diff
@@ -65,19 +75,100 @@ public void stop()
...
+   * Creates a user with specified permissions using the basic auth security API.
+   *
+   * @param adminClient  HTTP client authenticated as admin
+   * @param coordinator  the coordinator server to make API calls against
+   * @param username     the username to create
+   * @param password     the password for the user
+   * @param roleName     the role name to create and assign
+   * @param permissions  the permissions to grant to the role
+   */
+  public static void createUserWithPermissions(
```
capistrant, line 104: nice

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -1184,4 +1217,136 @@
...
+    private static Object[] buildQueryRow(
+        final QueryInfo queryInfo,
+        final boolean includeInfo,
+        final ObjectMapper jsonMapper
+    )
+    {
+      final Object[] row = new Object[QUERIES_SIGNATURE.size()];
+      row[0] = queryInfo.executionId();
+      row[1] = queryInfo.engine();
+      row[2] = queryInfo.state().toString();
```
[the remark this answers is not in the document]
capistrant: +1 to the bot

docs/querying/sql-metadata-tables.md
```diff
@@ -335,4 +335,37 @@ For example, to retrieve properties for a specific server, use the query
 
 ```sql
 SELECT * FROM sys.server_properties WHERE server='192.168.1.1:8081'
-```
\ No newline at end of file
+```
+
+### QUERIES table
```
capistrant, line 340: With this being off by default, I guess it would make me think as an operator that there is either a cost or a risk to turning it on. I don't think risk, but cost, maybe? Or is this just how we have operated with new sys tables in past, feature flag them, and once non-experimental remove the flag?

If there is a consideration an operator should take beyond "this is off by default cuz it is experimental" then I think we should add such a disclaimer here

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
```diff
@@ -1587,6 +1595,76 @@ public void testPropertiesTable()
 
   }
 
+  @Test
+  public void testQueriesTable()
+  {
+    // Create mock SqlEngine that returns test queries
+    final SqlEngine mockEngine = EasyMock.createMock(SqlEngine.class);
+    EasyMock.expect(mockEngine.name()).andReturn("native").anyTimes();
```
capistrant, line 1603: just showing the native engine some love by using it in the mocks so it isn't feeling left out by not being in the real thing?

docs/querying/sql-metadata-tables.md
gianm: I was thinking it's because it only works with Dart, and Dart is experimental, so this should be experimental too by transitivity. I will add that to the docs.

sql/src/test/java/org/apache/druid/sql/calcite/schema/SystemSchemaTest.java
gianm: Maybe it'd support `native` one day 😄

PR #18926, opened 2026-01-17
MSQFaultsTest: Speed up TooManyClusteredByColumns test.

gianm:
Speed up the test by introducing a context parameter maxClusteredByColumns, and use it to lower the threshold needed to generate the fault. The test formerly took over 30s, and becomes less than a second.

### DIFF
--- a/multi-stage-query/src/test/java/org/apache/druid/msq/exec/MSQFaultsTest.java
+++ b/multi-stage-query/src/test/java/org/apache/druid/msq/exec/MSQFaultsTest.java
@@ -594,7 +594,7 @@ public class MSQFaultsTest extends MSQTestBase
             new DruidExceptionMatcher(
                 DruidException.Persona.USER,
                 DruidException.Category.INVALID_INPUT,
-                "generic"
+                "general"
             ).expectMessageContains(
                 "SQL requires union between two tables and column names queried for each table are different "
                 + "Left: [dim2, dim1, m1], Right: [dim1, dim2, m1]."))


multi-stage-query/src/main/java/org/apache/druid/msq/util/MultiStageQueryContext.java
```diff
@@ -370,6 +371,20 @@ public static int getMaxPartitions(final QueryContext queryContext)
...
+  public static int getMaxClusteredByColumns(final QueryContext queryContext)
+  {
+    final Integer value = queryContext.getInt(CTX_MAX_CLUSTERED_BY_COLUMNS);
+    if (value == null) {
+      return Limits.MAX_CLUSTERED_BY_COLUMNS;
+    }
+    if (value <= 0) {
+      throw DruidException.forPersona(DruidException.Persona.USER)
+                          .ofCategory(DruidException.Category.INVALID_INPUT)
+                          .build("%s must be a positive integer, got[%d]", CTX_MAX_CLUSTERED_BY_COLUMNS, value);
```
abhishekrb19: shorthand nit: `throw InvalidInput.exception()`

multi-stage-query/src/main/java/org/apache/druid/msq/util/MultiStageQueryContext.java
```diff
@@ -145,6 +145,7 @@ public class MultiStageQueryContext
   public static final String CTX_MAX_INPUT_BYTES_PER_WORKER = "maxInputBytesPerWorker";
   public static final String CTX_MAX_INPUT_FILES_PER_WORKER = "maxInputFilesPerWorker";
   public static final String CTX_MAX_PARTITIONS = "maxPartitions";
+  public static final String CTX_MAX_CLUSTERED_BY_COLUMNS = "maxClusteredByColumns";
```
cryptoe, line 154: Nit: 
We add the dev documentation of context flags not released externally here : 
https://github.com/apache/druid/pull/18926/files#diff-5c9b518168571acb24b0a3c8c9e7213b960e839e38d105c148927a32fc64d65cL72
Also we might want to mention the motivation of this context parameter for code maintainability. I am really not sure if setting this in any prod workflows would be useful.

multi-stage-query/src/main/java/org/apache/druid/msq/util/MultiStageQueryContext.java
gianm: Updated this and also other occurrences in the same file.

multi-stage-query/src/main/java/org/apache/druid/msq/util/MultiStageQueryContext.java
gianm: Added a comment.

PR #18930, opened 2026-01-19
lz4j: use safeDecompressor instead of fastDecompressor

kgyrtkirk:
* the fastDecompressor method is deprecated
* according to lz4 apidoc-s the `fast` could be even slower

https://github.com/yawkat/lz4-java/blob/384ade4e11ca33002e600449682e5ebb63b85ac8/src/java/net/jpountz/lz4/LZ4Factory.java#L82-L89


cryptoe (commented):
Just wanted to check if there are any backward compatibility concerns during rolling upgrade etc that we need to be aware about, or reading older data.

kgyrtkirk:
It seems to me that `lz4` is only used in `memcached` related classes. But he decoders should work the same - there shouldn't be any compatibility issues.

cryptoe (approved):

PR #18936, opened 2026-01-21
Adjust cost-based autoscaler algorithm

Fly-Style:
## Core ideas around the change
- Scale up more aggressively when per-partition lag is meaningful.
- Relax the PPT (partitions-per-task) increase limit based on lag severity and headroom.
- Keep behavior conservative near `taskCountMax` and avoid negative headroom effects.

## Key changes
- **Lag-proportional PPT relaxation**: add `rawExtra` when avg lag exceeds thresholds, plus a step function for higher lag.
- **Headroom-aware scaling**: multiply by `max(0, 1 - currentTasks/maxTasks)` to reduce aggressiveness near max and avoid negative scaling.
- **Lag-amplified idle decay**: when lag/partition exceeds the activation threshold, the lag-based busy factor is multiplied 
    by a ramped multiplier (1.0x at 50K to 2.0x at 500K) to reduce predicted idle more aggressively during high lag.

## Expected behavior (for example, 30 partitions, `taskCountMax`=30)
- 3 tasks: 50K lag/partition → 8 tasks; 300K → 15 tasks; 500K+ → 30 tasks.
- 10 tasks: 100K → 15; 300K → 30; 500K → 30.
- 20/25 tasks (PPT=1): scale-up capped only by `taskCountMax`.
Note: examples assume lag-dominant conditions (low/mide processing rate, low idle).

## Open questions / risks
- **Lag skew**: using `aggregateLag / partitionCount` can under-react if a few partitions are far behind. Consider max/percentile lag if available.

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
@@ -3403,7 +3403,6 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
         group.completionTimeout = DateTimes.nowUtc().plus(ioConfig.getCompletionTimeout());
         pendingCompletionTaskGroups.computeIfAbsent(groupId, k -> new CopyOnWriteArrayList<>()).add(group);
 
-
         boolean endOffsetsAreInvalid = false;
         for (Entry<PartitionIdType, SequenceOffsetType> entry : endOffsets.entrySet()) {
           if (entry.getValue().equals(getEndOfPartitionMarker())) {
@@ -3453,18 +3452,26 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
    * This method is invoked to determine whether a task count adjustment is needed
    * during a task rollover based on the recommendations from the task auto-scaler.
  */
+
   @VisibleForTesting
   void maybeScaleDuringTaskRollover()
   {
     if (taskAutoScaler != null && activelyReadingTaskGroups.isEmpty()) {
       int rolloverTaskCount = taskAutoScaler.computeTaskCountForRollover();
-      if (rolloverTaskCount > 0) {
+      if (rolloverTaskCount > 0 && rolloverTaskCount != getIoConfig().getTaskCount()) {
         log.info("Autoscaler recommends scaling down to [%d] tasks during rollover", rolloverTaskCount);
         changeTaskCountInIOConfig(rolloverTaskCount);
-        autoScalerConfig.setTaskCountStart(rolloverTaskCount);
         // Here force reset the supervisor state to be re-calculated on the next iteration of runInternal() call.
         // This seems the best way to inject task amount recalculation during the rollover.
         clearAllocationInfo();
+
+        ServiceMetricEvent.Builder event = ServiceMetricEvent
+            .builder()
+            .setDimension(DruidMetrics.SUPERVISOR_ID, supervisorId)
+            .setDimension(DruidMetrics.DATASOURCE, dataSource)
+            .setDimension(DruidMetrics.STREAM, getIoConfig().getStream());
+
+        emitter.emit(event.setMetric(AUTOSCALER_REQUIRED_TASKS_METRIC, rolloverTaskCount));
       }
     }
   }
@@ -4049,7 +4056,7 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
           builder.put(partitionId, makeSequenceNumber(sequence, useExclusiveStartSequenceNumberForNonFirstSequence()));
         }
       } else {
-        // if we don't have a startingOffset (first run or we had some previous failures and reset the sequences) then
+        // if we don't have a startingOffset (first run, or we had some previous failures and reset the sequences) then
         // get the sequence from metadata storage (if available) or Kafka/Kinesis (otherwise)
         OrderedSequenceNumber<SequenceOffsetType> offsetFromStorage = getOffsetFromStorageForPartition(
             partitionId,
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
@@ -19,6 +19,8 @@
 
 package org.apache.druid.indexing.seekablestream.supervisor.autoscaler;
 
+import it.unimi.dsi.fastutil.ints.IntArraySet;
+import it.unimi.dsi.fastutil.ints.IntSet;
 import org.apache.druid.indexing.common.stats.DropwizardRowIngestionMeters;
 import org.apache.druid.indexing.overlord.supervisor.SupervisorSpec;
 import org.apache.druid.indexing.overlord.supervisor.autoscaler.LagStats;
@@ -33,9 +35,7 @@ import org.apache.druid.java.util.emitter.service.ServiceMetricEvent;
 import org.apache.druid.query.DruidMetrics;
 import org.apache.druid.segment.incremental.RowIngestionMeters;
 
-import java.util.HashSet;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
@@ -57,10 +57,24 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
 
   private static final int MAX_INCREASE_IN_PARTITIONS_PER_TASK = 2;
   private static final int MAX_DECREASE_IN_PARTITIONS_PER_TASK = MAX_INCREASE_IN_PARTITIONS_PER_TASK * 2;
+  /**
+   * Defines the step size used for evaluating lag when computing scaling actions.
+   * This constant helps control the granularity of lag considerations in scaling decisions,
+   * ensuring smoother transitions between scaled states and avoiding abrupt changes in task counts.
+   */
   private static final int LAG_STEP = 100_000;
+  /**
+   * This parameter fine-tunes autoscaling behavior by adding extra flexibility
+   * when calculating maximum allowable partitions per task in response to lag,
+   * which must be processed as fast, as possible.
+   * It acts as a foundational factor that balances the responsiveness and stability of autoscaling.
+   */
   private static final int BASE_RAW_EXTRA = 5;
-  static final int LAG_ACTIVATION_THRESHOLD = 50_000;
-  static final int LAG_AGGRESSIVE_THRESHOLD = 100_000;
+  // Base PPT lag threshold allowing to activate a burst scaleup to eliminate high lag.
+  static final int EXTRA_SCALING_LAG_PER_PARTITION_THRESHOLD = 50_000;
+  // Extra PPT lag threshold allowing activation of even more aggressive scaleup to eliminate high lag,
+  // also enabling lag-amplified idle calculation decay in the cost function (to reduce idle weight).
+  static final int AGGRESSIVE_SCALING_LAG_PER_PARTITION_THRESHOLD = 100_000;
 
   public static final String LAG_COST_METRIC = "task/autoScaler/costBased/lagCost";
   public static final String IDLE_COST_METRIC = "task/autoScaler/costBased/idleCost";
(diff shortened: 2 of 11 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -57,6 +57,11 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
 
   private static final int MAX_INCREASE_IN_PARTITIONS_PER_TASK = 2;
   private static final int MAX_DECREASE_IN_PARTITIONS_PER_TASK = MAX_INCREASE_IN_PARTITIONS_PER_TASK * 2;
+  private static final int LAG_STEP = 100_000;
```
gianm, line 65: There's a lot of magic-looking constants here. Please add javadocs to these constants describing what they do, and consider whether any of them should be configurable to allow for experimentation without changing the code. (Of course, ideally, users do not need to configure anything.)

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -230,32 +240,77 @@ int computeOptimalTaskCount(CostMetrics metrics)
...
     );
 
     for (int partitionsPerTask = maxPartitionsPerTask; partitionsPerTask >= minPartitionsPerTask; partitionsPerTask--) {
       final int taskCount = (partitionCount + partitionsPerTask - 1) / partitionsPerTask;
-      result.add(taskCount);
+      if (taskCount <= taskCountMax) {
+        result.add(taskCount);
+      }
+    }
+    return result.stream().mapToInt(Integer::intValue).sorted().toArray();
```
gianm: Could use an `IntSet` for `result` to avoid the boxing and stream usage. (It has a method that directly returns an `int[]`)

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -3461,6 +3461,7 @@ void maybeScaleDuringTaskRollover()
       if (rolloverTaskCount > 0) {
         log.info("Autoscaler recommends scaling down to [%d] tasks during rollover", rolloverTaskCount);
         changeTaskCountInIOConfig(rolloverTaskCount);
+        autoScalerConfig.setTaskCountStart(rolloverTaskCount);
```
gianm: What's this needed for? I recall in the discussion on #18745 we ended up deciding that `taskCountStart` would be immutable, and the autoscaler would change only `taskCount`.

Fly-Style: Yes, but seems it doesn't work in rollover scenario, which I was not aware about at that time.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Will add javadocs; I am following common sense here in terms of specific numbers.

> Of course, ideally, users do not need to configure anything.

That's what I'm looking for.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
gianm: What's the issue?

Fly-Style: Config persistence works fine when on the supervisor start/stop level, but doesn't work on the task level. Will update that too in the current PR, if that's possible to do painlessly.

UPD: the code flow around it a bit complicated, thinking how to resolve it properly...

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -230,32 +240,77 @@ int computeOptimalTaskCount(CostMetrics metrics)
...
+  }
+
+  /**
+   * Computes extra allowed increase in partitions-per-task in scenarios when the average
+   * per-partition lag is relatively high.
```
capistrant: relatively high compared to what? I think I'm failing to grasp how the constants for lag thresholds can be so generally applicable that we don't need to open them up to configuration. I think this idea of allowing for a burst in PPT scale factor is reasonable, but is this specific impl coming from tuning for too specific of a case? I think this goes along with Gian's question on the constants above.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -29,7 +29,11 @@
 public class WeightedCostFunction
 {
   private static final Logger log = new Logger(WeightedCostFunction.class);
-
+  private static final double HIHG_LAG_SCALE_FACTOR = 100_000.0;
```
capistrant: is this supposed to be `HIGH` or am I unfamiliar with what `HIHG` may stand for in math terms? I tried to google it in case it is an abbreviation I don't know, but didn't find anything so I tend to think maybe just misspelled?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -106,10 +110,14 @@ public CostResult computeCost(CostMetrics metrics, int proposedTaskCount, CostBa
 
 
   /**
-   * Estimates the idle ratio for a given task count using a capacity-based linear model.
+   * Estimates the idle ratio for a proposed task count.
+   * Includes lag-based adjustment to eliminate high lag and
```
capistrant, line 110: I think here again I am wondering about how generally applicable a constant for per partition lag in the real world. Same/similar questions as from the PPT scale limits computed in `CostBasedAutoScaler`. 

Also, in addition to the above, I think adding in this lag consideration does add some complexity here. Mainly it generally starts us down the path of making the cost function harder to easily and quickly understand for a newcomer, IMO. If this added complexity is considered a negative or "cost", the positive of improved behavior should outweigh it. So I guess that begs the question, how did we or are we going to measure the improvement that this additional logic/computation provides?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: It is `HIGH`, just typo.

Fly-Style:
@capistrant thanks for the review. 
About the documentation: check https://github.com/apache/druid/pull/18819, the algorigthm is described; this patch introduces a tweaks to the autoscaler about being too conservative on high lag scenarios.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Eventually, we may make those constants configurable, but now they are baked only by my observations as 'good' defaults.

I thought even more and I make some constants configurable (like lag threshold) even in the current PR.

capistrant:
> @capistrant thanks for the review. About the documentation: check #18819, the algorigthm is described; this patch introduces a tweaks to the autoscaler about being too conservative on high lag scenarios.

thanks ya I was looking through that PR description to get a grasp on things. plus your javadocs are informative which is good. I guess lets table docs/blog talk until future, I got ahead of myself there. Math stresses me out too easily 😆

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: > Also, in addition to the above, I think adding in this lag consideration does add some complexity here. Mainly it generally starts us down the path of making the cost function harder to easily and quickly understand for a newcomer, IMO. 

Sometimes you want to have complex things in the project, because they make some things work slightly better. A good example is query planner / query optimizer, which we have from the Calcite side. It's not easy to enter, hard to master, but with complexity it brings a good framework to start using SQL for your database.
Same here: in order to make supervisor autoscaling work well, we need to introduce a level of complexity baked by math (the formulas are described here: https://github.com/apache/druid/pull/18819). During the testing, I realized it is too conservative in the high lag scenarios, and it is an attempt to tweak it a bit.

_I hook up your question from general comment:_
> I also wonder if a more in depth technical writeup once this feature is stabilized is in order. Something that explains the method to the madness and a bit of the math. Perhaps in a brief blog post or docs page within the apache Druid website?

We must do it, but the feature is not finally stabilized; anyway, it already has a decent base.

> So I guess that begs the question, how did we or are we going to measure the improvement that this additional logic/computation provides?

That's a very good question, and I would answer in the following manner: the less time we spend scaling supervisors manually / fine-tuning an autoscaler, the better the result we will receive.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
gianm: I don't quite understand what you mean by "it doesn't work on the task level". The autoscaler config is a supervisor-only concept so it should only need to be persisted with the supervisor. I'm asking about this because I thought the conclusion in #18745 did make sense, and I feel the current PR deviates from that approach. In particular, in #18745 we arrived at an approach where `taskCount` is always the "current" task count and is updated by the autoscaler as the task count changes. `taskCountStart` is provided by the user only, and is optional. If present, it overwrites `taskCount` on supervisor submit (but not stop/start or failover).

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
gianm: Ideally nothing needs to be configured although I also feel that it may be useful for now to allow the lag threshold to be configured. The reason is that it's expressed in number of messages, but in practice, processing time of a message can vary quite a lot. Messages can be 100 bytes and processed very quickly, or they can be 10KB and processed more slowly. The difference between these two extremes is 100x so one number does not fit all. A lag of 100,000 messages could represent less than a second of processing, or could represent a minute of processing.

Ultimately I think we want the decision-making to be based on the system's assessment of how many messages it can process per unit of time. But that doesn't need to be done now. Probably for now we can set the initial thresholds assuming some medium rate of processing, like assuming we can process 5,000 or 10,000 messages per second per task.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
Fly-Style: @gianm I spent some time debugging and realized that the issue mentioned above is not an issue at all. I already rolled back `AutoScalerConfig` related changes. What have you written is fully correct and there is no intention to change it now, it was my false assumption.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: > Ultimately I think we want the decision-making to be based on the system's assessment of how many messages it can process per unit of time

That's an ideal scenario, and actually what I am looking for too in the long run. 

Now I want to have some good defaults to start from, and your assumptions are close to what is in the codebase already: assuming that we're processing ~10.000 messages from ~10 partitions (typical for mid-load Kafka topic). 
Lag of 100.000 messages per partition is already inappropriate and may cause serious inconvenience to Druid users.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: Fixed: https://github.com/apache/druid/pull/18936/commits/caf6aa20375c5178c00d16edbf7e059f86568bed

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Fixed: https://github.com/apache/druid/pull/18936/commits/caf6aa20375c5178c00d16edbf7e059f86568bed

PR #18938, opened 2026-01-21
bugfix: Create tombstones when needed while doing REPLACE mode with range partitioning plus parallel indexing

capistrant:
##### Bug Description

Tombstones are not properly created when in `REPLACE` indexing mode and ParallelIndexSupervisorTask range partitioning gets no reports for partition boundaries due to all rows being filtered out of the underlying source data. This results in existing segments remaining available that should have been overshadowed.

##### Reproduction Steps

Compaction with range partitioning and a compaction transform filter that filters all rows out of the segments being compacted. 

Expected behavior: tombstones created for the interval, and the old segments being overshadowed.

Actual behavior: Task completes with a warning that there were no valid rows for range partitioning. This leaves the old segments in place and will lead to an infinite loop of compaction

#### Release note

Fix bug for native compaction engine where tasks using range partitioning and a filter that filters out all rows in an interval would not create Tombstones to overshadow existing segments when `dropExisting=true` is set in the IOConfig. This leads to intervals stuck in an infinite loop of running compaction because the underlying segments that are having all their rows filtered are not properly overshadowed and marked unused.

<hr>

 * `ParallelIndexSupervisorTask`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/test/java/org/apache/druid/indexing/common/task/CompactionTaskParallelRunTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/common/task/CompactionTaskParallelRunTest.java
@@ -1003,6 +1003,108 @@ public class CompactionTaskParallelRunTest extends AbstractParallelIndexSupervis
     }
   }
 
+  @Test
+  public void testRunParallelWithRangePartitioningFilteringAllRows() throws Exception
+  {
+    allowSegmentFetchesByCompactionTask = true;
+
+    // Range partitioning is not supported with segment lock yet
+    Assume.assumeFalse(lockGranularity == LockGranularity.SEGMENT);
+
+    runIndexTask(null, true);
+
+    Collection<DataSegment> usedSegments = getCoordinatorClient().fetchUsedSegments(
+        DATA_SOURCE,
+        ImmutableList.of(INTERVAL_TO_INDEX)
+    ).get();
+    Assert.assertEquals(3, usedSegments.size());
+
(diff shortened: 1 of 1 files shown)


capistrant:
I thought the hashed partitions phase would face the same issue, but it does not. The test I added flexes the code path that would have been buggy if it did, but tombstones still end up being created.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -374,4 +496,22 @@ public static List<CompactionEngine> getEngine()
...
+    for (CompactionEngine engine : List.of(CompactionEngine.NATIVE, CompactionEngine.MSQ)) {
+      for (String partitionType : List.of("range", "hash")) {
+        if (engine == CompactionEngine.MSQ && "hash".equals(partitionType)) {
+          // MSQ does not support hash partitioning in this context
+          continue;
+        }
+        params.add(new Object[]{engine, partitionType});
+      }
+    }
+    return params;
```
abhishekrb19: This really is {"MSQ", "range"}, {"NATIVE", "range"}, {"NATIVE", "hash"}, we can just inline it I think

capistrant:
I forgot that embedded tests coverage doesn't count for jacoco. will have to look at what existing test structure there is in the package

PR #18844, opened 2025-12-15
Implement a fingerprinting mechanism to track compaction states in a more efficient manner

capistrant:
**disclaimer: This PR begins the migration away from always using the term compaction. It may create some short term friction/frustration. But the goal is that longer term we refer to things that are not pure compaction in a more ambiguous way (indexing state versus compaction state, for example)**

### Description

Add new functionality to Compaction Supervisors. Instead of storing compaction state for segments individually, centralize the state storage in a new metadata table. Individual segments will store a computed fingerprint that references an indexing state (aka compaction state) in the new metadata table. Since many segments will eventually end up sharing common indexing states, this should greatly reduce duplication in metadata storage.

**note:** This applies only to Compaction Supervisors. Scheduled compaction on the coordinator will not use fingerprinting. This is intentional as a part of the roadmap to sunsetting coordinator duty compaction in the future.

####  Indexing State Fingerprinting

Instead of storing serialized `CompactionState` objects as the `lastCompactionState` field in every compacted segment, generate a fingerprint for a `CompactionState` and attach that to compacted segments. Add new centralized storage for serialized `CompactionState` objects where individual states can be looked up by the aforementioned fingerprint. Since it is common for many segments in a data source to share a single `CompactionState`, this greatly reduces the metadata storage overhead for storing indexing states in the metadata store.

###### `druid_segments`

Add new column `indexing_state_fingerprint` that stores the fingerprint representation of the segments current `CompactionState`. It can be `null` if no compaction has taken place.

###### `druid_indexingStates`

New metadata table that stores the full serialized `CompactionState` associated with a fingerprint. Segments can look up their detailed state here by using the `indexing_state_fingerprint` that they are associated with to pull the full state payload.

###### `SqlIndexingStateStorage`

The `IndexingStateStorage` interface + implementation(s) is responsible for managing the persistence and lifecycle of indexing states. It stores unique compaction configurations (identified by fingerprints) in the metadata database. The manager tracks which compaction states are actively referenced by segments, marking unreferenced states as unused and periodically cleaning up old unused states. This fingerprinting approach allows Druid to efficiently store and retrieve compaction metadata without duplicating identical compaction configurations across multiple segments.

###### `HeapMemoryIndexingStateStorage`

Meant to serve as a mechanism for testing and simulations where metadata persistence may not be available/needed

##### `IndexingStateCache`

[...]

Issue #9768, opened 2020-04-24
Allow minor compaction for non-consecutive segments


jihoonson:
Each segment has a partitionId which uniquely identifies the segment in a time chunk. Currently, you cannot compact segments with minor compaction (which uses the segment lock) if their partitionIds are not consecutive. For example, you cannot compact the segments of the partitionIds 0, 1, 10 together because the partitionId 1 and the partitionId 10 are not consecutive. 

This is an expected limitation of the minor compaction by its design. (It is for reducing memory footprint. See https://github.com/apache/druid/issues/7491 for more details.) However, in practice, it would be nice if the minor compaction can compact non-consecutive segments. This will be nice especially if there are some expected but transient task failures in streaming ingestion because those task failures can cause the non-consecutive segment IDs.

The minor compaction can support this if it's guaranteed that no new segments have the partitionId which falls in the overlapping root partition range of the existing segments.

jihoonson:
Assuming that we will keep the current segment ID allocation protocol that monotonically increases the partition ID on task failures, the problem we want to solve is, given a missing partitionId, how we would know whether the segment of that ID really doesn't exist or it is being created by some other task. One way to do is modifying the compaction task to as below.

1) When some missing partitionIds are found, the compaction task tries to lock them using the regular locking mechanism.
2) If the locking succeeds, the compaction task can safely assume that those partitionIds will never be used since there is no ingestion task creating segments of those partitionIds. In this case, the compaction task can simply ignore those missing partitionIds and compact the given segments all together.
3) If the locking fails, there should be some ingestion task creating the segments of those partitionIds. In this case, the compaction task can split the input segments into multiple groups where each group has only consecutive partitionIds, and compact each group separately. Those segments that are being created by other task can be compacted later using another compaction task.

Issue #9712, opened 2020-04-16
Enable auto minor compaction


yuanlihan:
### Description

Note that @jihoonson introduced the minor compaction feature to Druid at version 0.16.x, which helps to effectively compact small segments. As we know, the cluster may suffer from the too many small segments issue if there are big pipelines that have a long trickle of late data. 

But after upgrading to version 0.16.1, I found that the `auto minor compaction` was not implemented then. But still I rushed to the feature by making an adaptation based on the implementation of minor compaction and it works well till now. Thanks @jihoonson and all others involved with this greate feature.

Nowadays, we decide to upgrade to newer version, but find that the `auto minor compaction` is still not available. May I know if @jihoonson or anyone else WIP about this feature now? And if not, maybe I can have a try to raise a PR about the adaptation.

PS: there was a detail discuss https://github.com/apache/druid/issues/8489 between @jihoonson and @himanshug about `CompactionTask` and the `auto compaction`.

### Motivation
- It will be more convenient to issue minor compaction tasks if enabling `auto minor compaction`.

jihoonson:
Hi @yuanlihan, thank you for trying out minor compaction!

The thing is that you can enable auto minor compaction now by setting `forceTimeChunkLock = false` in the tuningConfig of coordinator compaction configuration. However, with auto compaction, you can use only either minor compaction or major compaction for now. To support both major and minor compactions, `CompactSegments` should be improved to be able to run both of them. 

Also, please note that the segment lock which is used in minor compaction is still experimental and there are a couple of known issues such as https://github.com/apache/druid/issues/9571. 

Probably I could work on fixing those bugs with segment lock and improving auto compaction to support both major and minor compactions at some point this year, but am unsure when it could be yet.

yuanlihan:
Thanks @jihoonson for the explanation.

Now technically, the auto minor compaction is available by setting `forceTimeChunkLock = false` in the `context` property of both coordinator compaction configuration and the supervisor spec.(using version 0.17.x)
But it seems that currently it does not work perfectly in my case. The minor compaction tasks issued by coordinator always grab all segments in the given time interval, which seems not minor enough to fully leverage the power of minor compaction.

Here is a typical arrangement of segments within an hour in my case:
<details>
  <summary>click to expand list</summary>
  
| segmentId    | size |
| ----- | ---- |
| ds_seg_id | 129 KB  | 
| ds_seg_id_1 | 132 KB  | 
| ds_seg_id_2 | 192 KB  | 
[...]

Issue #9571, opened 2020-03-26
Race in SegmentAllocateAction with segment lock


jihoonson:
### Affected Version

0.16, 0.17

### Description

Before 0.16, everything was simple and there was only one type of task lock, i.e., time chunk lock. With time chunk lock, only one task can work on the same datasource and the same time period at any time. As a result, `SegmentAllocateAction` could safely assume that there is only one task asking a new segment allocation for the same datasource and the same time period.

However, with segment lock, multiple tasks can work on the same datasource and the same time period simultaneously. Multiple tasks can request new segment allocations at the same time, which introduces a race condition in segment allocation. 

In `SegmentAllocateAction`, it first looks for segments of overlapping intervals with the request interval populated based on the given segment granularity. This is to avoid creating segments of overlapping intervals. To do so, [it first retrieves all overlapping _used segments_ with the request interval from the metadata store](https://github.com/apache/druid/blob/master/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentAllocateAction.java#L195-L196). However, this is not enough with segment lock because there could be other tasks running which created _pending segments_ of overlapping intervals. If there were such tasks, they could end up creating overlapping segments which breaks the timeline of brokers and the coordinator.

I think this check probably should include pending segments as well, and be done in `TaskLockbox` with a proper synchronization.

PR #18939, opened 2026-01-21
Reindexing rule providers with cascading interval based reindexing

capistrant:
follow up to #18844 ... at least in terms of the quest to begin the transition from the term compaction to reindexing. More info can be found in that PR description about the naming change and new centralized indexing state storage that the supervisor uses to determine if segments need to be reindexed (replacement for lastCompactionState stored per segment). In this PR I will use the term reindexing whenever possible. When the term compaction is used it will only be to refer to an actual Java class that is yet to be refactored.

### Description

Extend reindexing supervisors (AKA compaction supervisors) to allow for a single Druid data source to apply different reindexing configurations to different segments depending on what "reindexing rules" are defined for the data source that apply to the time interval of the segment being reindexed.

#### Value Proposition

Timeseries data often provides value in different ways over time. Data for the last 7 days is often interacted with differently than data for the last 30 days and once again for data from some number of years ago. In Druid, we should give data owners the ability to use reindexing supervisors (AKA compaction supervisors) to change the state of a single datasource as the data ages. Operators should have the ability to define a data lifecycle of sorts for their datasources and allow reindexing supervisors to dynamically apply that definition to the underlying segments as they age. A great, and simple, example is query granularity. 

> realtime use cases often call for fine grained query granularity as data is flowing into druid. I may want to analyze metrics by minute or even second. As that data ages to be days or months old, I may still want to be able to check trends for that data over time, but it is often highly unlikely that I care about second to second analysis for data that is past a certain age. And beyond that, there may be tiers of value. For instance, where I want second granularity for the last 24 hours, minute granularity for everything in the past week, and hour granularity for everything older than that. The new functionality in this PR intends to make achieving this as simple as defining a few rules for a datasource using reindexing supervisors, and allowing Druid to reindex the data as segments age out of one time interval and into another that calls for less finely grained query granularity.

#### CompactionConfigBasedJobTemplate underpins all of this

The existing `CompactionConfigBasedJobTemplate` underpins all of the new functionality. At the end of the day, the cascading reindexing template uses per-datasource `ReindexingRule`s that are supplied by a `ReindexingRuleProvider` to create multiple non-overlapping search intervals that each apply a distinct `InlineSchemaDataSourceCompactionConfig`.

#### `ReindexingRule`

A good place to start review is to take a quick look at the `ReindexingRule` concept and the different implementations. The basic idea is that the components of the existing `CompactionState` concept have been broken out into multiple rule types:

##### `ReindexingDataSchemaRule`

* Query Granularity and rollup bool
* DimensionsSpec
* MetricsSpec
* Projections

##### `ReindexingDeletionRule`

`CompactionState#TransformSpec` maps to what we call `ReindexingDeletionRule` which is a `DimFilter` under the hood, and all rows matching are deleted during reindexing.

##### Rule Anatomy

```JAVA
private final String id;
private final String description;
private final Period olderThan;
//...
//The special bits of an individual rule. Like a `Granularity` for a ReindexingSegmentGranularityRule
```

The key concept to take away is that `olderThan` reads as: "Apply this rule to Druid segments who have intervals ending at or before `now - olderThan`"

###### Rule selection when `> 1` rules of the same type exist for a datasource

The whole point of this PR is so a datasource can have different reindexing configuration at different points in its timeline. That means there will inevitably multiple rules of the same type that exist for a dataSource. For all but one rule type, only a single rule can be selected per search interval and its associated `InlineSchemaDataSourceCompactionConfig`. When this happens, it is the `ReindexingRuleProvider` implementation that makes the selection of which rule to pick for an interval. In this PR, the only concrete implementation is `InlineReindexingRuleProvider` and the logic for selection is simple:

> If N rules of type T apply to the search interval `[2025-01-01/2025-01-02)` select the one most specific to that interval. "most specific" being the rule with an `olderThan` period that when subtracted from the current time UTC when the supervisor is running, is closest to the interval end (but never before it, we don't apply rules to intervals they partially overlap with the inline provider)

####### Special Case: `ReindexingDeletionRule`

The `ReindexingDeletionRule` is the only rule type that is additive. All matching rules for a search interval become part of the `InlineSchemaDataSourceCompactionConfig` for that interval. The take the form of:

`NOT(A OR B OR C OR D)` 

where A, B, C, ... are individual `ReindexingDeletionRule`s

######## `ReindexingDeletionRule` optimization

We avoid applying `ReindexingDeletionRule`s to `CompactionCandidate`s whose segments have all applied the rule already. This rule pruning is done using an optimizer before the underling `CompactionConfigBasedJobTemplate`  creates the `CompactionTask`s. This is done to avoid wasting work during re-indexing.

#### `ReindexingRuleProvider`

The next logical place to go from here in review is the rule provider itself. This is meant to be extensible, with this PR providing one concrete implementation, `InlineReindexingRuleProvider`, that defines all of the rules within the supervisor spec itself.

Classes to check out are: `ReindexingRuleProvider` and `InlineReindexingRuleProvider`

A note on `ComposingReindexingRuleProvider`: This guy is a little funny in this PR as it is meant to give you power once there are multiple `ReindexingRuleProvider` implementations in Druid. You can chain them together, with their ordering mattering in determining what provider provides what rules. 

##### `InlineReindexingRuleProvider` rule application

A rule applies to an interval IF the `reference time - olderThan period` for the rule is explicitly AFTER the end time of that interval. This means that rules only apply to intervals that are fully before their threshold.

#### `CascadingReindexingTemplate`

This is the brains of the operation when it comes to how the supervisor takes a collection of rules that exist for a datasource and turns them into `CompactionTask`s that can be run.

##### Non-Overlapping Timeline Generation

The most interesting thing about this is generating the non-overlapping timeline of `CompactionConfigBasedJobTemplate` objects where the non-overlapping terminology pertains to the search intervals that each template will cover.

###### Target SegmentGranularity Requirement

It is mandatory that every `CompactionConfigBasedJobTemplate` has a target `SegmentGranularity` defined in its config. This target granularity comes from either an explicit `ReindexingSegmentGranularityRule` that applies to the search interval or the default segment granularity that the `cascadeReindex` must provide in the spec to ensure that even if no rules exist, the default will stand in to fulfill this requirement.

###### Search interval boundary adjustment to match target segment granularity

I will use an example to show this in action as I think it is easier than words.

* Supervisor Run Time: `2024-02-04T22:12:04.873Z`
* Supervisor Segment Granularity Rules:
  * `olderThan=P7D segmentGranularity=DAY`
  * `olderThan=P1M segmentGranularity=MONTH`
  * `olderThan=P1Y segmentGranularity=YEAR`

Timeline with no adjustments to align search intervals with target granularity:

```
[-INF/2023-02-04T22:12:04.873)
[2023-02-04T22:12:04.873/2024-01-04T22:12:04.873)
[2024-01-04T22:12:04.873/2024-01-28T22:12:04.873)
```

Timeline with our adjustments to align search intervals with target granularity:

```
[-INF/2023-01-01T00:00:00.000Z)
[2023-01-01T00:00:00.000Z/2024-01-01T00:00:00.000Z)
[2024-01-01T00:00:00.000Z/2024-01-28T00:00:00.00)
```

A side effect of this is that rules will technically be delayed in their being applied, but it will ensure our search intervals align nicely with the segment granularity that they are targeting. 

Two unfortunate constraints:
* The source segments being re-indexed need to be at the same or finer grain than the target granularity of the search interval covering them. If not, there will be segments that overlap the interval boundaries in our timeline and get ignored.
* The target granularities in the search intervals generated need to the same or getting more coarse as you get further in the past. Not doing this will cause issues with granularity aligning the search intervals while generating the timeline.
  * I think it is important to strictly enforce this. I have added a change that during the creation of the initial granularity aligned timeline for creating reindexing search intervals, an IAE is thrown if the granularity is not stable or coarsening over time into the past. 

###### Splitting the timeline built over the Granularity rules to accommodate the non-segment gran rules

We still have tons of rule types beyond these segment granularity rules that we fussed over to create our initial timeline. The `reference_time - olderThan` thresholds for each of these additional rules will fall within an existing interval in our timeline (except for one case we will cover at the end of this section). For each of these rules, we first use the segment granularity of the interval they fall in to granularity align their threshold time `Granularity.bucketStart(threshold)`. If this adjusted threshold still falls within the interval, we need to split the interval at this time to allow precise application of the rule. We repeat this with all of the distinct `olderThan` periods that exist for our non segment granularity rules.

_Edge cases_

* No Segment Granularity Rule exists
  * This is what the default granularity is for. We create a synthetic granularity aligned timeline using the smallest period from our non segment granularity rules and then split this synthetic interval just as we would for the normal case.
* Segment granularity rules exist, but the smallest `olderThan` period for a segment granularity rule is larger than the smallest period for non segment granularity rules
  * we again use the default segment granularity and the smallest `olderThan` period to create a synthetic interval to make sure all of our search intervals have a segment granularity.  

### Example Inline Rule Provider Spec

```json
{
  "type": "autocompact",
  "spec": {
    "type": "reindexCascade",
    "dataSource": "wikipedia",
    "defaultSegmentGranularity": "HOUR",
    "taskPriority": 25,
    "inputSegmentSizeBytes": 500000000,
    "skipOffsetFromLatest": "P1D",
    "engine": "msq",
    "taskContext": {
      "maxNumTasks": 10
    },
    "ruleProvider": {
      "type": "inline",
      "segmentGranularityRules": [
        {
          "id": "recent-fine-grain",
          "olderThan": "P1D",
          "segmentGranularity": "HOUR"
        },
        {
          "id": "week-old-medium-grain",
          "olderThan": "P7D",
          "segmentGranularity": "DAY"
        },
        {
          "id": "old-coarse-grain",
          "olderThan": "P30D",
          "segmentGranularity": "MONTH"
        }
      ],
      "deletionRules": [
        {
          "id": "remove-bots",
          "description": "Remove robot traffic from old data",
          "olderThan": "P30D",
          "deleteWhere": {
            "type": "equals",
            "column": "is_bot",
            "matchValueType": "STRING",
            "matchValue": "true"
          }
        },
        {
          "id": "remove-test-users",
          "description": "Remove test users from very old data",
          "olderThan": "P1Y",
          "deleteWhere": {
            "type": "equals",
            "column": "userType",
            "matchValueType": "STRING",
            "matchValue": "test"
          }
        }
      ],
      "tuningConfigRules": [
        {
          "id": "partition-by-country",
          "description": "Partition data by country for better query performance for common use cases",
          "olderThan": "P1D",
          "tuningConfig": {
            "partitionsSpec": {
              "type": "range",
              "maxRowsPerSegment": 10000000,
              "partitionDimensions": [
                "countryName"
              ]
            }
          }
        }
      ],
      "dataSchemaRules": [
        {
          "id": "base-data-schema-rule",
          "olderThan": "P1D",
          "dimensionsSpec": {
            "dimensions": [
              "channel",
              "cityName",
              "comment",
              "countryIsoCode",
              "countryName",
              "namespace",
              "page",
              "regionIsoCode",
              "regionName",
              "user",
              "isAnonymous",
              "isMinor",
              "isNew",
              "isRobot",
              "isUnpatrolled",
              "metroCode"
            ]
          },
          "metricsSpec": [
            {
              "type": "longSum",
              "name": "added",
              "fieldName": "added"
            },
            {
              "type": "longSum",
              "name": "deleted",
              "fieldName": "deleted"
            },
            {
              "type": "longSum",
              "name": "delta",
              "fieldName": "delta"
            }
          ],
          "queryGranularity": "MINUTE",
          "rollup": true
        }
      ]
    }
  },
  "suspended": true
}

```

This spec will:
* Apply the following for all data older than 1 day
  * Use range partitioning on countryName
  * pass through base metrics and dimensions specs for all time
* Apply the following for data 1 to 7 days old
  * Reindex to HOUR segment gran
* Apply the following for data 1 to 30 days old
  * Reindex to MINUTE query granularity
* Apply the following for data 7 to 30 days old
  * Reindex to DAY segment gran
* Apply the following for data older than 30 days old
 * Reindex to MONTH segment gran
 * Reindex to HOUR query granularity
 * Remove robot traffic when data
* Apply the following for data older than 1 year
  * Remove test users when data is older than 1 year

#### documentation

I think this should come with the console addition PR since that will be when this really becomes realistic to operate without getting too confused with what is going on under the hood for a config.

#### Console (Future Addition)

Edit: After discussions we decided this is better fit for a follow up PR. I will leave this section as preview for now. but the console changes will actually come in 2nd PR

Based of review with others, adding visualizations of the reindexing timeline in the console seemed like a nice to have for operators. And after experimenting with adding it, I think it is actually a critical piece for sane operations. Using an AI coding agent (I'm know nothing about front end dev), I iterated on console changes and came up with a proposal for a console view of the reindexing timeline so an operator can see how the underlying rules break up into search intervals and what configuration will be applied to each search interval. Below are some pictures using the example inline supervisor above

image

image

image

#### catalog provider?

Would be cool to tie this into the catalog definitions of tables

#### Release note

<hr>

 * `ReindexingConfigOptimizer`
 * `ReindexingRuleProvider`
   * `InlineReindexingRuleProvider`
   * `ComposingReindexingRuleProvider` 
 * `ReindexingRule` + `AbstractReindexingRule`
   *  `ReindexingDimensionsRule`
   * `ReindexingMetricsRule`
   * `ReindexingSegmentGranularityRule`
   * `ReindexingQueryGranularityRule`
   * `ReindexingIOConfigRule`
   * `ReindexingProjectionRule`
   * `ReindexingDeletionRule`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
+++ b/server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
@@ -34,6 +34,7 @@ import org.apache.druid.java.util.common.logger.Logger;
 import org.apache.druid.query.aggregation.AggregatorFactory;
 import org.apache.druid.segment.IndexSpec;
 import org.apache.druid.segment.metadata.IndexingStateFingerprintMapper;
+import org.apache.druid.segment.transform.CompactionTransformSpec;
 import org.apache.druid.server.coordinator.DataSourceCompactionConfig;
 import org.apache.druid.server.coordinator.UserCompactionTaskGranularityConfig;
 import org.apache.druid.timeline.CompactionState;
@@ -752,14 +753,20 @@ public class CompactionStatus
 
     private CompactionStatus transformSpecIsUpToDate(CompactionState lastCompactionState)
     {
-      if (compactionConfig.getTransformSpec() == null) {
+      final CompactionTransformSpec configuredSpec = compactionConfig.getTransformSpec();
+      if (configuredSpec == null
+          || (configuredSpec.getFilter() == null && configuredSpec.getVirtualColumns().isEmpty())) {
         return COMPLETE;
       }
 
+      final CompactionTransformSpec existingSpec = Configs.valueOrDefault(
+          lastCompactionState.getTransformSpec(),
+          new CompactionTransformSpec(null, null)
+      );
       return CompactionStatus.completeIfNullOrEqual(
           "transformSpec",
-          compactionConfig.getTransformSpec(),
-          lastCompactionState.getTransformSpec(),
+          configuredSpec,
+          existingSpec,
           String::valueOf
       );
     }


server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
```diff
@@ -251,6 +258,113 @@ public static CompactionStatus running(String message)
...
+      // Extract applied filters from the CompactionState into a Set
+      Set<DimFilter> appliedFilters = extractAppliedFilters(state);
+
+      // If transform spec or filter for the CompactionState is null, return all expected filters eagerly
+      if (appliedFilters == null) {
+        return expectedFilter;
+      }
+
+      // Check which expected filters are NOT in the applied set and add them to unappliedRules
+      for (DimFilter expected : expectedFilters) {
```
capistrant: possible to extract expectedFilters into a set and do a type of set operation here?

server/src/main/java/org/apache/druid/server/compaction/InlineReindexingRuleProvider.java
```diff
@@ -0,0 +1,390 @@
...
+  private final List<ReindexingDimensionsRule> compactionDimensionsRules;
+  private final List<ReindexingIOConfigRule> compactionIOConfigRules;
+  private final List<ReindexingProjectionRule> compactionProjectionRules;
+  private final List<ReindexingGranularityRule> compactionGranularityRules;
+  private final List<ReindexingTuningConfigRule> compactionTuningConfigRules;
+
+
+  @JsonCreator
+  public InlineReindexingRuleProvider(
+      @JsonProperty("compactionFilterRules") @Nullable List<ReindexingFilterRule> compactionFilterRules,
```
capistrant: these are all out of date naming wise

FrankChen021:
So, in future, there's no compaction term?

However, I have a different view. IMO, the compaction and re-indexing should be separated from each other, the should serve complete different purposes

Compaction should only performs the merge of small segments without any schema changs(query granularity, segment granularity). The compaction should perform eagerly and aggresively especially for kafka ingestion to reduce number of segment. There're many problems/limitation around this feature that have not solved. for example, the compaction now performs compaction on a whole interval, if there're many segments, it takes very long time(and sometimes it's not realistic to complete) to finish the job. This kind of compaction was originally named as 'Majar compaction', while a 'minor compaction' is there to allow us to compact given segments, but it's buggy now, even we give 2 segments for example, the task will still fetch all segments in that interval. And another problem is that the minor compaction only accepts segments with consecutive segments. these problems are states in: #9712 , #9768, #9571
However, these problems are not solved, and we still experience the large number of small segments for long. 
Apply the re-indexing term for this use case, I think the term itself does not reflect its feature but introduces confusion.

capistrant:
> So, in future, there's no compaction term?
> 
> However, I have a different view. IMO, the compaction and re-indexing should be separated from each other, the should serve complete different purposes
> 
> Compaction should only performs the merge of small segments without any schema changs(query granularity, segment granularity). The compaction should perform eagerly and aggresively especially for kafka ingestion to reduce number of segment. There're many problems/limitation around this feature that have not solved. for example, the compaction now performs compaction on a whole interval, if there're many segments, it takes very long time(and sometimes it's not realistic to complete) to finish the job. This kind of compaction was originally named as 'Majar compaction', while a 'minor compaction' is there to allow us to compact given segments, but it's buggy now, even we give 2 segments for example, the task will still fetch all segments in that interval. And another problem is that the minor compaction only accepts segments with consecutive segments. these problems are states in: #9712 , #9768, #9571 However, these problems are not solved, and we still experience the large number of small segments for long. Apply the re-indexing term for this use case, I think the term itself does not reflect its feature but introduces confusion.

I appreciate the thoughts @FrankChen021. In general, this push away from using the term compaction for everything that re-processes existing druid segments is long needed. But, I do agree that pure compaction like you spec out with minor compactions does still warrant being called "compaction". Whether that be as a subset of the "reindexing" space or its own separate concept entirely, I guess I don't know. Overall, we have lots of robust production ready code for "compaction" already that I could not justify re-building anything for "reindexing" specifically. That is the genesis of trying to generalize the name as I do think it does make more sense to call pure compaction, reindexing, than it does to call activity that changes the underlying data definition, compaction. I do want to work towards a naming scheme and code base that is logical and reasonable though, so I am open to considering how we can best navigate to a world where only stuff that is legitimately compaction is called compaction.

capistrant:
I think it is a bad assumption by me that I don't allow rules to have a period of `PT0s` for "apply this to everything". Since currently compaction does allow a skip offset from latest to be `PT0s`, I think it makes sense to support

gianm:
> for example, the compaction now performs compaction on a whole interval, if there're many segments, it takes very long time(and sometimes it's not realistic to complete) to finish the job. This kind of compaction was originally named as 'Majar compaction', while a 'minor compaction' is there to allow us to compact given segments, but it's buggy now, even we give 2 segments for example, the task will still fetch all segments in that interval.

As to minor compaction, IMO it should come back and use the mechanisms that exist for concurrent compaction in order to carry-through already-compacted segments unchanged (new version, same fie on s3) while only performing the physical compaction operation on the new segments.

It could be a reshuffle of the new segments, which would require some evolving of the shard specs to be able to attach pruning metadata outside the core set. It could also be a simple merge with no shuffle, which wouldn't be as good for pruning but would execute much more quickly. I lean towards a simple merge being a good idea, with major compaction being used to attach the pruning information.

I believe an in depth discussion of minor compaction is off-topic for this PR, though, except to acknowledge that if we are having a unified concept of compaction and reindexing, then that unified concept should be broad enough to include minor compaction down the road.

capistrant:
looking into those test failures after merging master in now. That compaction task test framework was recently refactored to work with MSQ so this is first time they have ran against this PR

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
```diff
@@ -645,7 +653,7 @@ private Query<?> buildGroupByQuery(
     List<PostAggregator> postAggregators =
         inputColToVirtualCol.entrySet()
                             .stream()
-                            .filter(entry -> !entry.getKey().equals(ColumnHolder.TIME_COLUMN_NAME))
+                            .filter(entry -> entry.getKey().startsWith(ARRAY_VIRTUAL_COLUMN_PREFIX))
```
capistrant: there is a bug introduced here.

capistrant: From looking at this, I think my change to the filter was just blatantly wrong and now got caught since we started running MSQ compaction tests on a much wider scope

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NativeCompactionRunner.java
```diff
@@ -89,6 +89,14 @@ public CompactionConfigValidationResult validateCompactionTask(
       Map<Interval, DataSchema> intervalDataSchemaMap
   )
   {
+    // Virtual columns in filter rules are not supported by native compaction
+    if (compactionTask.getTransformSpec() != null
+        && compactionTask.getTransformSpec().getVirtualColumns() != null
+        && compactionTask.getTransformSpec().getVirtualColumns().getVirtualColumns().length > 0) {
```
clintropolis, line 95: could also consider using `isEmpty()`

processing/src/main/java/org/apache/druid/segment/transform/CompactionTransformSpec.java
```diff
@@ -46,17 +47,20 @@ public static CompactionTransformSpec of(@Nullable TransformSpec transformSpec)
...
   }
 
   @Nullable private final DimFilter filter;
+  @Nullable private final VirtualColumns virtualColumns;
 
   @JsonCreator
   public CompactionTransformSpec(
-      @JsonProperty("filter") final DimFilter filter
+      @JsonProperty("filter") final DimFilter filter,
+      @JsonProperty("virtualColumns") final VirtualColumns virtualColumns
```
clintropolis: this feels like a weird place for this to me but maybe i keep getting hung up on this just because of how stuff is laid out internally where transform spec filters and transforms happen at a layer below where virtual columns operate.

i think really my main beef is really with DataSourceCompactionConfig, i'd like to change it eventually with something that looks more like the v10 layout, where the base table schema is also defined as a projection, so that base table schema would be where the filter and virtual column live in that world and the stuff would be the set of projections to build and then the context/segment granularity/ioconfig/tuningconfig. 

All that said, this is fine for now i think...

server/src/main/java/org/apache/druid/server/compaction/InlineReindexingRuleProvider.java
```diff
@@ -0,0 +1,421 @@
...
+  public InlineReindexingRuleProvider(
+      @JsonProperty("reindexingDeletionRules") @Nullable List<ReindexingDeletionRule> reindexingDeletionRules,
+      @JsonProperty("reindexingMetricsRules") @Nullable List<ReindexingMetricsRule> reindexingMetricsRules,
+      @JsonProperty("reindexingDimensionsRules") @Nullable List<ReindexingDimensionsRule> reindexingDimensionsRules,
+      @JsonProperty("reindexingIOConfigRules") @Nullable List<ReindexingIOConfigRule> reindexingIOConfigRules,
+      @JsonProperty("reindexingProjectionRules") @Nullable List<ReindexingProjectionRule> reindexingProjectionRules,
+      @JsonProperty("reindexingSegmentGranularityRules") @Nullable List<ReindexingSegmentGranularityRule> reindexingSegmentGranularityRules,
+      @JsonProperty("reindexingQueryGranularityRules") @Nullable List<ReindexingQueryGranularityRule> reindexingQueryGranularityRules,
+      @JsonProperty("reindexingTuningConfigRules") @Nullable List<ReindexingTuningConfigRule> reindexingTuningConfigRules
+  )
```
clintropolis, line 105: also, should we just drop the `reindexing` prefix from all these names?

server/src/main/java/org/apache/druid/server/compaction/InlineReindexingRuleProvider.java
```diff
@@ -0,0 +1,421 @@
...
+  @JsonCreator
+  public InlineReindexingRuleProvider(
+      @JsonProperty("reindexingDeletionRules") @Nullable List<ReindexingDeletionRule> reindexingDeletionRules,
+      @JsonProperty("reindexingMetricsRules") @Nullable List<ReindexingMetricsRule> reindexingMetricsRules,
+      @JsonProperty("reindexingDimensionsRules") @Nullable List<ReindexingDimensionsRule> reindexingDimensionsRules,
+      @JsonProperty("reindexingIOConfigRules") @Nullable List<ReindexingIOConfigRule> reindexingIOConfigRules,
+      @JsonProperty("reindexingProjectionRules") @Nullable List<ReindexingProjectionRule> reindexingProjectionRules,
+      @JsonProperty("reindexingSegmentGranularityRules") @Nullable List<ReindexingSegmentGranularityRule> reindexingSegmentGranularityRules,
+      @JsonProperty("reindexingQueryGranularityRules") @Nullable List<ReindexingQueryGranularityRule> reindexingQueryGranularityRules,
+      @JsonProperty("reindexingTuningConfigRules") @Nullable List<ReindexingTuningConfigRule> reindexingTuningConfigRules
```
clintropolis: while very flexible, i think this might be a bit overly expressive to the point that things might be hard to reason about. How about consolidating `ReindexingDimensionsRule`, `ReindexingMetricsRule`, `ReindexingProjectionRule`, and `ReindexingQueryGranularityRule` into a single `ReindexingSchemaRule` that has the option to set all of these things in one shot. I think that will make it a lot easier for operators to keep on top of what is actually going to happen during reindexing, and still should be adequately expressive.

server/src/main/java/org/apache/druid/server/compaction/ReindexingConfigBuilder.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * Builds compaction configs by applying reindexing rules.
+ * Encapsulates the logic for combining additive rules and applying all rule types.
+ */
+public class ReindexingConfigBuilder
```
clintropolis: I sort of wonder if we need this, or if it should just be built into a apply method of ReindexingRuleProvider so that it can fully decide how to transform the InlineSchemaDataSourceCompactionConfig.Builder. Then again, maybe it is fine if we imagine that all rule providers will always do the transform in the same way?

server/src/main/java/org/apache/druid/server/compaction/ReindexingRule.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Defines a reindexing configuration that applies to data based on age thresholds.
+ * <p>
+ * Rules encapsulate specific aspects of reindexing (granularity, filters, tuning, etc.)
+ * and specify when they should apply via a {@link Period} which defines the age threshold for applicability.
+ * <p>
+ * Rules conditionally apply to data "older than" the rules threshold relative to the time of rule evaluation.
+ */
+public interface ReindexingRule
```
clintropolis, line 36: gut reaction here was is any reason for 'apply' to not live on this interface to do the decoration of the config builder directly? On further thought, I guess it doesn't provide a lot of benefit since we would still need to like, group them by rule type so that we can choose which one(s) of each type to apply, and also probably encode whether or not they are additive to decide how to apply each type

server/src/main/java/org/apache/druid/server/compaction/ReindexingRule.java
```diff
@@ -0,0 +1,73 @@
...
+    PARTIAL,
+    FULL,
+    NONE
+  }
+
+  String getId();
+
+  String getDescription();
+
+  Period getOlderThan();
```
clintropolis, line 71: nit: javadocs, at least for `getOlderThan`

indexing-service/src/main/java/org/apache/druid/indexing/compact/ReindexingConfigFinalizer.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Functional interface for customizing a {@link DataSourceCompactionConfig} for a specific
+ * {@link CompactionCandidate} before creating a reindexing job. This allows template-specific
+ * logic to be injected without hardcoding behavior in {@link CompactionConfigBasedJobTemplate}.
+ * <p>
+ * For example, cascading reindexing templates can use this to optimize filter rules based on
+ * the candidate's indexing state, while simpler templates can use the identity finalizer.
+ */
+@FunctionalInterface
+public interface ReindexingConfigFinalizer
```
clintropolis: (naming is the worst) i wonder if 'specialize' is a better way to refer to this than 'finalizer'. 

Also, I can't help but wonder if this should be part of the `ReindexingRuleProvider` since it also handles doing transforms to compaction configs, and it would allow implementations the ability to do per candidate transforms that are beyond what is built-in to the cascading template.

That said, if we think that we will always want to do the same set of rules for cascading template, this is probably fine too

indexing-service/src/main/java/org/apache/druid/indexing/compact/ReindexingDeletionRuleOptimizer.java
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Optimization utilities for applying {@link ReindexingDeletionRule}s during reindexing
+ * <p>
+ * When reindexing with {@link ReindexingDeletionRule}s, it is possible that candidate
+ * segments have already applied some or all of the deletion rules in previous reindexing runs. Reapplying such rules would
+ * be wasteful and redundant. This class provides funcionality to optimize the set of rules to be applied by
+ * any given reindexing task.
+ */
+public class ReindexingDeletionRuleOptimizer
```
clintropolis: the naming of this makes it more sound like a 'thing' instead of utility methods. I wonder should this just be reworked into an implementation of `ReindexingConfigFinalizer` instead of a separate class?

clintropolis:
we discussed this a bit offline, but to add here for more visibility - I think it would be quite useful to have an API available that could resolve a timeline of all the effective compaction configs for a datasource, to help operators visualize what is going to happen the the table over time (and could also serve as the basis of a UI to do the same in some future). It would probably be fine if this was a follow-up PR too if you prefer. I think the logic would basically be like creating jobs except just returning the config and interval instead of an actual compaction jobs

server/src/main/java/org/apache/druid/server/compaction/InlineReindexingRuleProvider.java
capistrant: ya, seems redundant to have now that you call it out

indexing-service/src/main/java/org/apache/druid/indexing/compact/ReindexingDeletionRuleOptimizer.java
capistrant: fair point. This was the child of me wanting to make CascadingReindexingTemplate less dense and I didn't scrutinize my approach very well. My thinking here is to actually make this all an implementation of what is previously called `ReindexingConfigFinalizer` The CascadingReindexingTemplate creates one of these Optimizers and passes it to the underlying job template to call optimize on. In the future if the cascading template had more optimizations, you could like add different optimizer implementations and chain them?

indexing-service/src/main/java/org/apache/druid/indexing/compact/ReindexingConfigFinalizer.java
capistrant: hmm. What about 'optimizer'

I haven't really thought too much about this in the rule provider. But in general I think the rule provider should be as light weight as possible in that it is just a vessel to pass rules from some source location to a template, so it can create configs based on the rules (in this case the cascading template). They all supply the same rule types by contract, and like if we hypothetically had a catalog provider and a raw rdbms provider, I think all 3 would want the same optimization done for deletion rules. So overall at least as of now, I think the optimization implementation ties closest to the template type versus the rule provider

server/src/main/java/org/apache/druid/server/compaction/ReindexingConfigBuilder.java
capistrant: It's definitely possible that this is not the right approach, ya. it was another thing born out of trying to take some of the cognitive load and complexity of of the `CascadingRedindexingTemplate` because that is originally where it lived. I'll think a bit about the alternatives such as being in the provider. But part of me doesn't really want the rule provider to know anything other than what rules it servers and just offer them up to a caller to do something with them

capistrant:
> we discussed this a bit offline, but to add here for more visibility - I think it would be quite useful to have an API available that could resolve a timeline of all the effective compaction configs for a datasource, to help operators visualize what is going to happen the the table over time (and could also serve as the basis of a UI to do the same in some future). It would probably be fine if this was a follow-up PR too if you prefer. I think the logic would basically be like creating jobs except just returning the config and interval instead of an actual compaction jobs

After toying around with a coding agent I feel like I was able to drum up a surprisingly nice UI in the supervisor view (for cascading reindex type only). The code is currently in this PR, but I'm down to move it to a follow in terms of sanity for review.

server/src/main/java/org/apache/druid/server/compaction/InlineReindexingRuleProvider.java
capistrant: latest changes consolidate into a single schema rule

capistrant:
@clintropolis TY for the review. I made some changes based off our conversation and off some findings during further testing. rough summary is:
* consolidated those rule types we talked about into a single schema rule type
* did some refactoring of the "finalizer", now called "optimizer". No more utility class full of static helpers, but rather just a single implementation of the deletion rule optimizer which is all the cascading template needs (for now).
* moved and changed visibility of the builder of the builder that you questioned the need for. Still haven't thought of a better way to achieve what it is doing, but hoping this way is easier to refactor out of in the future if we don't replace it in this PR specifically
* ~added a supervisor api for getting the timeline of search intervals and applied rules~
  * ~added a UI for this api in the console~

not directly related to your review:
* Fixed a bug in the building of the configs where the wrong segment granularity rule was getting applied to intervals
* Changed the logic for skipOffsets, where if a skip offset calculation lands in the middle of a search interval, it just skips compaction of that whole interval instead of trying to truncate the interval and process part of it. I think this simplification of that bit takes some complexity out of this code creating the search intervals, making it a little easier to reason about.
* I also tried to break down some of the dense methods with lots of logic in CascadingReindexingTemplate by splitting out some private methods. Some of this was driven by the need for some re-use now that the template services the timeline creation request from the supervisor API

server/src/main/java/org/apache/druid/server/compaction/ReindexingConfigBuilder.java
capistrant: Took step to move this to indexing-service alongside the CascadingReindexingTemplate as a package private class since it really is tied to that template at this point. I think this will make it easier to refactor if we decide a better way/place for this.

processing/src/main/java/org/apache/druid/segment/transform/CompactionTransformSpec.java
capistrant: one thought I guess is that CompcationTransformSpec actually just becomes a baseTransformSpec with transforms that are ExpressionTransform? Just popped in my head. I do want to try to avoid going down a path where we are using something that is ok for now but needs changing eventually, cuz that would end up changing the indexing state fingerprint and cause the supervisor to have to re-process everything that used this feature

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -274,6 +245,191 @@ public void test_compaction_withPersistLastCompactionStateFalse_storesOnlyFinger
...
+    Assertions.assertEquals(4, getNumSegmentsWith(Granularities.FIFTEEN_MINUTE));
+    Assertions.assertEquals(5, getNumSegmentsWith(Granularities.HOUR));
+    Assertions.assertEquals(4, getNumSegmentsWith(Granularities.DAY));
+    verifyEventCountOlderThan(Period.days(7), "item", "hat", 0);
+  }
+
+  @Test
+  public void test_cascadingReindexing_withVirtualColumnOnNestedData_filtersCorrectly()
+  {
+    // Virtual Collumns on nested data is only supported with MSQ compaction engine right now.
```
clintropolis: nit: 
```suggestion
    // Virtual Columns on nested data is only supported with MSQ compaction engine right now.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -274,6 +245,191 @@ public void test_compaction_withPersistLastCompactionStateFalse_storesOnlyFinger
...
+        .schemaDiscovery()
+        .segmentGranularity("DAY");
+
+    cluster.callApi().runTask(task.withId(IdUtils.getRandomId()), overlord);
+    cluster.callApi().waitForAllSegmentsToBeAvailable(dataSource, coordinator, broker);
+
+    Assertions.assertEquals(4, getTotalRowCount());
+
+    VirtualColumns virtualColumns = VirtualColumns.create(
+        ImmutableList.of(
```
clintropolis: nit: i think can leave out the list part and use the other 'create'

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
```diff
@@ -0,0 +1,803 @@
...
+      @JsonProperty("ruleProvider") ReindexingRuleProvider ruleProvider,
+      @JsonProperty("engine") @Nullable CompactionEngine engine,
+      @JsonProperty("taskContext") @Nullable Map<String, Object> taskContext,
+      @JsonProperty("skipOffsetFromLatest") @Nullable Period skipOffsetFromLatest,
+      @JsonProperty("skipOffsetFromNow") @Nullable Period skipOffsetFromNow,
+      @JsonProperty("defaultSegmentGranularity") Granularity defaultSegmentGranularity
+  )
+  {
+    this.dataSource = Objects.requireNonNull(dataSource, "'dataSource' cannot be null");
+    this.ruleProvider = Objects.requireNonNull(ruleProvider, "'ruleProvider' cannot be null");
```
clintropolis: should these use `InvalidInput.conditionalException` to throw a proper `DruidException` to signal user error? (same comments for other validations)

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionConfigBasedJobTemplate.java
```diff
@@ -48,10 +48,20 @@
 public class CompactionConfigBasedJobTemplate implements CompactionJobTemplate
 {
   private final DataSourceCompactionConfig config;
+  private final ReindexingConfigOptimizer configFinalizer;
```
clintropolis: nit: variable name no longer matches

processing/src/main/java/org/apache/druid/segment/transform/CompactionTransformSpec.java
```diff
@@ -66,6 +71,14 @@ public DimFilter getFilter()
     return filter;
   }
 
+  @JsonProperty
+  @JsonInclude(JsonInclude.Include.NON_NULL)
```
clintropolis: i think if we modified this to 
```
@JsonInclude(JsonInclude.Include.NON_DEFAULT)
```

we could modify the constructor to initialize to `VirtualColumns.EMPTY` and save ourselves some null checks and not risk messing up fingerprinting stuff for things without virtual columns defined

server/src/main/java/org/apache/druid/server/compaction/DataSourceCompactibleSegmentIterator.java
```diff
@@ -509,7 +509,7 @@ skipInterval, new Interval(remainingStart, remainingEnd)
       }
     }
 
-    if (!remainingStart.equals(remainingEnd)) {
```
clintropolis, line 512: was this a bug?

indexing-service/src/main/java/org/apache/druid/indexing/compact/SegmentGranularityTimelineValidationException.java
```diff
@@ -0,0 +1,81 @@
...
+      Granularity olderGranularity,
+      Interval newerInterval,
+      Granularity newerGranularity
+  )
+  {
+    super(
+        "Invalid segment granularity timeline for dataSource[%s]: "
+        + "Interval[%s] with granularity[%s] is more recent than "
+        + "interval[%s] with granularity[%s], but has a coarser granularity. "
+        + "Segment granularity must stay the same or become finer as data ages from present to past.",
```
clintropolis: >Segment granularity must stay the same or become finer as data ages from present to past.

reading this sort of sounds backwards from what the code says, the segment granularity of newer data must be finer than the older data

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -0,0 +1,1318 @@
...
+   *   <li>Sort rules by period: P90D→DAY (oldest), P30D→HOUR (newest)</li>
+   *   <li>P7D metrics is finer than P30D, so prepend interval with default MONTH granularity</li>
+   *   <li>Timeline would be: [-∞, DAY_boundary) DAY, [DAY_boundary, HOUR_boundary) HOUR, [HOUR_boundary, MONTH_boundary) MONTH</li>
+   *   <li>Validation: HOUR → MONTH progression means granularity is getting COARSER toward present</li>
+   * </ol>
+   * <p>
+   * EXPECTED: IllegalArgumentException with message about invalid granularity timeline
+   */
+  @Test
+  public void test_generateAlignedSearchIntervals_failsWhenDefaultGranularityIsCoarserThanMostRecentSegmentGranRule()
```
clintropolis, line 1408: nit: would be nicer to keep this grouped with other `@Test` methods instead of interleaved with private methods and class definitions

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -0,0 +1,1318 @@
...
+    SegmentTimeline timeline = createTestTimeline(referenceTime.minusDays(90), referenceTime.minusDays(10));
+    ReindexingRuleProvider mockProvider = createMockProvider(List.of(Period.days(7), Period.days(30)));
+    CompactionJobParams mockParams = createMockParams(referenceTime, timeline);
+    DruidInputSource mockSource = createMockSource();
+
+    TestCascadingReindexingTemplate template = new TestCascadingReindexingTemplate(
+        "testDS", null, null, mockProvider, null, null, null, null
+    );
+
+    template.createCompactionJobs(mockSource, mockParams);
```
clintropolis, line 284: i guess we ignore the results here because the test class just returns empty instead of a job to assert stuff about?

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -0,0 +1,1318 @@
...
+import org.apache.druid.testing.InitializedNullHandlingTest;
+import org.apache.druid.timeline.DataSegment;
+import org.apache.druid.timeline.SegmentTimeline;
+import org.easymock.EasyMock;
+import org.joda.time.DateTime;
+import org.joda.time.Interval;
+import org.joda.time.Period;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
```
clintropolis: nit: new test, maybe use junit5? (`Assertions` instead of `Assert`, etc)

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -0,0 +1,1318 @@
...
+    Assert.assertEquals(3, intervals.size());
+
+    Assert.assertEquals(DateTimes.MIN, intervals.get(0).getInterval().getStart());
+    Assert.assertEquals(DateTimes.of("2024-10-01T00:00:00Z"), intervals.get(0).getInterval().getEnd());
+
+    Assert.assertEquals(DateTimes.of("2024-10-01T00:00:00Z"), intervals.get(1).getInterval().getStart());
+    Assert.assertEquals(DateTimes.of("2024-12-29T00:00:00Z"), intervals.get(1).getInterval().getEnd());
+
+    Assert.assertEquals(DateTimes.of("2024-12-29T00:00:00Z"), intervals.get(2).getInterval().getStart());
+    Assert.assertEquals(DateTimes.of("2025-01-22T16:00:00Z"), intervals.get(2).getInterval().getEnd());
```
clintropolis: nit: this would probably be slightly nicer if it was just making an expected list and comparing to `intervals` instead of comparing all the starts and ends separately (same for other tests).

it would also be pretty cool if it was validating the segment granularities are expected

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -0,0 +1,1318 @@
...
+            new ReindexingSegmentGranularityRule("hour-rule", null, Period.days(7), Granularities.HOUR),
+            new ReindexingSegmentGranularityRule("day-rule", null, Period.months(1), Granularities.DAY),
+            new ReindexingSegmentGranularityRule("month-rule", null, Period.months(3), Granularities.MONTH)
+        ))
+        .dataSchemaRules(List.of(
+            new ReindexingDataSchemaRule("metrics-8d", null, Period.days(8), null, new AggregatorFactory[0], null, null, null),
+            new ReindexingDataSchemaRule("metrics-14d", null, Period.days(14), null, new AggregatorFactory[0], null, null, null),
+            new ReindexingDataSchemaRule("metrics-45d", null, Period.days(45), null, new AggregatorFactory[0], null, null, null),
+            new ReindexingDataSchemaRule("metrics-100d", null, Period.days(100), null, new AggregatorFactory[0], null, null, null)
+        ))
```
clintropolis, line 564: took me a second to realize these are to break up the segment granularity intervals into separate portions even though they don't actually change the dataschema, perhaps worth a comment (also maybe a utility method to shrink it down to a name and period)

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -0,0 +1,1318 @@
...
+   *   <li>[2024-01-21T00:00:00, 2024-01-28T00:00:00) - DAY</li>
+   *   <li>[2024-01-28T00:00:00, 2024-02-03T22:00:00) - HOUR (prepended, note non-midnight end)</li>
+   * </ol>
+   */
+  @Test
+  public void test_generateAlignedSearchIntervals()
+  {
+    DateTime referenceTime = DateTimes.of("2024-02-04T22:12:04.873Z");
+
+    ReindexingRuleProvider provider = InlineReindexingRuleProvider.builder()
```
clintropolis: nit: worth moving this first line onto a new line to shift everything over to the left to make more readable

server/src/test/java/org/apache/druid/server/compaction/ReindexingTuningConfigRuleTest.java
```diff
@@ -0,0 +1,191 @@
...
+  private final ReindexingTuningConfigRule rule = new ReindexingTuningConfigRule(
+      "test-tuning-rule",
+      "Custom tuning config",
+      PERIOD_21_DAYS,
+      createTestTuningConfig()
+
+  );
+
+  @Test
+  public void test_appliesTo_intervalFullyBeforeThreshold_returnsFull()
```
clintropolis, line 47: on the one hand i wonder if some of these tests should be consolidated since these all share an implementation of `appliesTo` afaict, but i guess it could be nicer than it is tedious if we ever split up implementations or something...

server/src/main/java/org/apache/druid/server/compaction/ComposingReindexingRuleProvider.java
```diff
@@ -0,0 +1,239 @@
...
+                    .orElse(Collections.emptyList());
+  }
+
+  @Override
+  public List<ReindexingDeletionRule> getDeletionRules(Interval interval, DateTime referenceTime)
+  {
+    return providers.stream()
+                    .map(p -> p.getDeletionRules(interval, referenceTime))
+                    .filter(rules -> !rules.isEmpty())
+                    .findFirst()
```
clintropolis, line 124: i wonder if it is worth warning or.. something if there are providers that have rules that are overshadowed by not being first. i don't think it needs to block this PR, we could wait and see how this is used in practice more to determine if we need additional assistance to operators wondering why rules resolved the way they did

server/src/main/java/org/apache/druid/server/compaction/ReindexingDataSchemaRule.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.druid.java.util.common.granularity.Granularity;
+import org.apache.druid.query.aggregation.AggregatorFactory;
+import org.apache.druid.server.coordinator.UserCompactionTaskDimensionsConfig;
+import org.joda.time.Period;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+
+public class ReindexingDataSchemaRule extends AbstractReindexingRule
```
clintropolis, line 52: may not matter, but this is missing equals/hashcode/tostring (provider uses them in its implementation of these methods, other rules missing these methods too)

also worth a small javadoc

server/src/main/java/org/apache/druid/server/compaction/AbstractReindexingRule.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+
+  /**
+   * Checks if a period with variable-length components (months/years) is positive.
+   *
+   * @param period the period to check
+   * @return true if any component is positive and no components are negative
+   */
+  private static boolean isPeriodPositive(Period period)
```
clintropolis: is there a case for having a rule that should always apply? (e.g. i would think period of 0?)

server/src/main/java/org/apache/druid/server/compaction/ReindexingSegmentGranularityRule.java
```diff
@@ -0,0 +1,71 @@
...
+public class ReindexingSegmentGranularityRule extends AbstractReindexingRule
+{
+  private final Granularity segmentGranularity;
+
+  @JsonCreator
+  public ReindexingSegmentGranularityRule(
+      @JsonProperty("id") @Nonnull String id,
+      @JsonProperty("description") @Nullable String description,
+      @JsonProperty("olderThan") @Nonnull Period olderThan,
+      @JsonProperty("segmentGranularity") @Nonnull Granularity segmentGranularity
```
clintropolis, line 72: granularity allows a lot of weird stuff, should we really allow all kinds of granularities here?

server/src/main/java/org/apache/druid/server/compaction/ReindexingRule.java
```diff
@@ -0,0 +1,87 @@
...
+   * Indicates how a rule applies to a given time interval based on the rule's period threshold.
+   * <ul>
+   * <li>PARTIAL: The rule applies to part of the interval.</li>
+   * <li>FULL: The rule applies to the entire interval.</li>
+   * <li>NONE: The rule does not apply to the interval at all.</li>
+   * </ul>
+   */
+  enum AppliesToMode
+  {
+    PARTIAL,
```
clintropolis, line 48: partial seems underutilized since afaict we only pick rules for FULL, what is the plan for it?

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
```diff
@@ -0,0 +1,803 @@
...
+        .collect(Collectors.toList());
+
+    // Build base timeline with granularities tracked
+    List<IntervalGranularityInfo> baseTimeline = new ArrayList<>();
+    DateTime previousAlignedEnd = null;
+
+    for (ReindexingSegmentGranularityRule rule : sortedRules) {
+      DateTime rawEnd = referenceTime.minus(rule.getOlderThan());
+      DateTime alignedEnd = rule.getSegmentGranularity().bucketStart(rawEnd);
+      DateTime alignedStart = (previousAlignedEnd != null) ? previousAlignedEnd : DateTimes.MIN;
```
clintropolis, line 584: you can get some pretty weird stuff going on here, perhaps we just encourage people not to create unhinged combinations that don't make sense?

LIke, I guess I wonder if we should take an opinionated stance and add some additional validation to make sure that the granularity is itself also like aligned with the interval?

Like I was curious what happens with a total freedom here, so for example imagine i have gone mad and define 3 segment granularity reindexing rules:
* 'hour-rule' - olderThan p1d, hourly granularity
* '3day-rule' - olderThan p2d, period granularity of p3d
* '5day-rule' - olderThan p3d, period granularity of p5d

Using the 'reference time' of the tests, this resolves into something like
```
IntervalGranularityInfo{interval=-146136543-09-08T08:23:32.096Z/2025-01-22T00:00:00.000Z, granularity={type=period, period=P5D, timeZone=UTC, origin=null}},
IntervalGranularityInfo{interval=2025-01-22T00:00:00.000Z/2025-01-27T00:00:00.000Z, granularity={type=period, period=P3D, timeZone=UTC, origin=null}},
IntervalGranularityInfo{interval=2025-01-27T00:00:00.000Z/2025-01-28T16:00:00.000Z, granularity={type=period, period=PT1H, timeZone=UTC, origin=null}}
```

seems strange because there is a 5 day interval of 3 day segments, what does that mean? :shrug:

I didn't even try adding timezones on period granularities or using like, duration granularity, but I suspect they can make stuff even weirder.

Is this ok? (i think the answer is maybe not, since i tried running a weird configuration like this on CompactionSupervisorTest.test_cascadingCompactionTemplate_multiplePeriodsApplyDifferentCompactionRules and MSQ got stuck in some weird state if i specified a timezone.. i didn't dig very deep into it yet)

server/src/main/java/org/apache/druid/server/compaction/DataSourceCompactibleSegmentIterator.java
capistrant: ya. I think it is/was generally impossible to hit. but you can't create an interval if start is after end and this wasn't checking that. It relied on the caller not passing in params that would cause it to happen. at one point my code was causing it to happen. But now my safety checks in cascading reindexing before creating some search intervals that make no sense should prevent it, and I guess existing compaction code doesn't hit it our we'd know about it

indexing-service/src/main/java/org/apache/druid/indexing/compact/SegmentGranularityTimelineValidationException.java
capistrant: yes

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
capistrant: ya the dummy template is just there to intercept and record what search intervals come out of the createCompactionJobs call to confirm the timeline is properly formed

server/src/main/java/org/apache/druid/server/compaction/ReindexingSegmentGranularityRule.java
capistrant: I guess I am not sure? I was basing off of what the old compaction granularity allowed and that was just `Granularity` there

server/src/main/java/org/apache/druid/server/compaction/ReindexingRule.java
capistrant: good question. I was under the impression when I initially was writing this that it may be useful, but from where I am now after dealing with trying to make all of this possible to reason about as an operator, I think a rule provider trying to manage partial overlaps somehow instead of forcing full coverage, would be a nightmare. so maybe I should delete?

server/src/main/java/org/apache/druid/server/compaction/AbstractReindexingRule.java
capistrant: ya. I think I accidentally pulled it from the description that period of 0 and handling of future data need to be addressed. Perhaps this PR warrants supporting P0D. but future data should be left for follow on if needed

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
capistrant: I almost want to say we take a hard stance and limit segment granularity rules to something like FIFTEEN_MINUTE, THIRTY_MINUTE, HOUR, DAY, MONTH, YEAR. at this point. The management of this timeline is proving to a real pain

server/src/main/java/org/apache/druid/server/compaction/ReindexingSegmentGranularityRule.java
capistrant: per your comment below I am trending towards just laying the hammer and having a very narrow set of allowed values here to prevent people from causing themselves harm

processing/src/main/java/org/apache/druid/segment/transform/CompactionTransformSpec.java
capistrant: thanks for the tip. I was definitely confused about what to do here to make it so everything that is already fingerprinted didn't end up being re-processed

server/src/main/java/org/apache/druid/server/compaction/ComposingReindexingRuleProvider.java
capistrant: This is a fair point, definitely could be annoying to troubleshoot

server/src/test/java/org/apache/druid/server/compaction/ReindexingTuningConfigRuleTest.java
capistrant: ya they all share. I didn't know what appropriate way was to handle this when a bunch of concrete classes share an abstract impl. cuz like you said, if someone added an impl to a concrete class, it would then be untested if we didn't have all these... but I guess that is why we have jacoco and code reviews.

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
capistrant: I guess I probably didn't do this for all the new test files at least in this module. If I do it for one, I guess I should do it for all but want to check with you before I do since it will create quite the diff for you to sort through.

capistrant:
thanks again for review round @clintropolis I think a lot makes sense and I've got most the changes locally. The segment granularity thing I'm at the point where it is so ripe for edge cases and errors where an operator may get themselves in trouble and regret it that I think the static set of standard granularities I called out in the comment is the way. my take is that it still gives people a lot more flexibility than operators have today, while trying to make our lives easier coding it, and prevent them (operators) from doing something crazy that they hate. what do you think?

processing/src/main/java/org/apache/druid/segment/transform/CompactionTransformSpec.java
capistrant: I went down a little bit of a rabbit hole and found that ` @JsonInclude(JsonInclude.Include.NON_EMPTY)` maybe covers this more robustly?

clintropolis: oh, since it just serializes the array as the value :+1:

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
clintropolis: its up to you if you want to do now or later. i think we eventually want to migrate them all over, i've personally just been doing it opportunistically whenever i touch a test if it isn't too much trouble, but i think its fine either way

server/src/main/java/org/apache/druid/server/compaction/ReindexingRule.java
clintropolis: i'm not completely sure, it seems semi useful to track, if for nothing else from a debugging standpoint in case it is useful later

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
clintropolis: i think it would be fine to start out super restrictive, we can always revisit this part later, probably easier to do that than start out super permissive and try to rein it in later

server/src/main/java/org/apache/druid/server/compaction/ComposingReindexingRuleProvider.java
clintropolis: i think this is fine for now, we can worry about in a follow-up if it proves hard to work with

server/src/test/java/org/apache/druid/server/compaction/ReindexingTuningConfigRuleTest.java
clintropolis: :shrug: this was just thinking out loud, is fine either way i think since i doubt they are very slow tests

server/src/main/java/org/apache/druid/server/compaction/ReindexingDeletionRule.java
```diff
@@ -0,0 +1,157 @@
...
+  public ReindexingDeletionRule(
+      @JsonProperty("id") @Nonnull String id,
+      @JsonProperty("description") @Nullable String description,
+      @JsonProperty("olderThan") @Nonnull Period olderThan,
+      @JsonProperty("deleteWhere") @Nonnull DimFilter deleteWhere,
+      @JsonProperty("virtualColumns") @Nullable VirtualColumns virtualColumns
+  )
+  {
+    super(id, description, olderThan);
+    this.deleteWhere = Objects.requireNonNull(deleteWhere, "deleteWhere cannot be null");
```
clintropolis: nit: this and a few other rules have places which could also be using `InvalidInput`

clintropolis (approved):
:metal:

indexing-service/src/main/java/org/apache/druid/indexing/compact/SegmentGranularityTimelineValidationException.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.apache.druid.java.util.common.IAE;
+import org.apache.druid.java.util.common.granularity.Granularity;
+import org.joda.time.Interval;
+
+/**
+ * Exception thrown when segment granularity timeline validation fails when using {@link CascadingReindexingTemplate}.
+ * Contains structured information about the conflicting intervals.
+ */
+public class SegmentGranularityTimelineValidationException extends IAE
```
kfaraz, line 30: Do we mean to catch this exception somewhere in the code and then throw a more user-friendly error? If this exception is itself propagated to the user, then we might as well use `InvalidInput.exception()`.

capistrant: This was born when I was working on the UI part of this. But even so, it may be not needed for that too. That code is a follow up PR so maybe this should also be sidelined and brought back in that PR if it makes sense.

server/src/main/java/org/apache/druid/server/compaction/ReindexingDeletionRule.java
capistrant: ah, will address this in this PR so it doesn't get forgotten.

PR #12279, opened 2022-02-24
Store null columns in the segments

jihoonson:
### Terminology

- Column: a logical column that can be stored across multiple segments.
- Null column: a column that has only nulls in it. Druid is aware of this column.
- Unknown column: a column that Druid is not aware of. In other words, it is a column that Druid is not tracking at all via segment metadata or any other methods.

### Description

Today, null columns are not stored at ingestion time and thus they become unknown columns once the ingestion job is done. The druid native query engine uses segment-level schema and treats unknown columns as if they were null columns; reading unknown columns returns only nulls.

Druid SQL is different. Druid is using Calcite SQL planner which requires valid column information at planning time. The column information is retrieved from datasource-level schema which is dynamically discovered by merging segment schemas. As a result, users cannot query unknown columns using SQL. This is causing a couple of issues. One of the main issues is the SQL queries failing against stream ingestion from time to time. While it creates segments, the realtime task announces a realtime segment that has all columns in the ingestion spec. This realtime task thus reports the broker even null columns for the realtime segment which can be used by the Druid SQL planner. Once the segment is handed off to a historical, it announces a historical segment that does not store any null columns in it. As a result, the same SQL query will no longer work after the segment is handed off.

### Proposed solution

To make the SQL planner be aware of null columns, Druid needs to track of them. This PR proposes to store those null columns in the segment just like normal columns.

#### Feature flag

`druid.index.task.storeEmptyColumns` is added. This is on by default. A new task context, `storeEmptyColumns`, is added which can override the system property.

#### Ingestion tasks

When `storeEmptyColumns` is set, the task stores every column specified in `DimensionsSpec` in the segments that it creates. This applies to all kinds of ingestion except for Hadoop ingestion and Tranquility.

#### Segment writes/reads

For null columns, Druid stores a column name, column type, number of rows, and bitmapSerdeFactory. The first two are stored in `ColumnDescriptor` and the last two are stored in `NullColumnPartSerde`. `NullColumnPartSerde` has a no-op serializer and a deserializer that can dynamically create a bitmap index and a dictionary. Finally, the null column names are stored at the end of `index.drd` separately from normal columns. This is for the compatibility for older historicals. When they read a segment that has null column stored, they won't be aware of those columns but will just ignore them without exploding. 

#### Test plan

- Unit tests are added in this PR to verify the compatibility for older historicals.
- Unit tests are added in this PR to verify that null columns are stored in segments.
- Integration tests will be added in https://github.com/apache/druid/pull/12268.

#### Future work

[...]

PR #18944, opened 2026-01-22
Fix Hadoop multi-value string null value handling to match native batch

jtuglu1:
### Description

1. Doing some more digging, I found another unfortunate data difference between native batch (on-cluster) and Hadoop batch ingest. Ingesting a multi-value string `["a","b",null]` with Hadoop is treated as `["a","b","null"]` and in native batch, this correctly ingests to `["a","b",null]`. This is difference appears to be a bug in all Druid versions(even latest). While this will not affect the current null handling migration, this will affect the future Hadoop -> native batch ingestion migration that will also need to take place.
2. Hadoop doesn't allow for all-null columns in segments, it simply excludes them from the segment. I've updated the Hadoop job to support running `druid.indexer.task.storeEmptyColumns=true`, which allows us to store all `NULL` columns (how native/streaming ingest work today).

Related to: 
- https://github.com/apache/druid/pull/15190.
- https://github.com/apache/druid/pull/12279.

#### Release note

Fix Hadoop null value handling to match native batch and allow v10 segment creation.

##### BREAKING CHANGES
1. Hadoop ingests will now process multi-value string inputs like `["a","b",null]` -> `["a","b",null]` instead of `["a","b","null"]` to match native batch ingestion.
2. Hadoop ingests will now by default keep columns with all `NULL` values, instead of excluding them from the segment.
3. `useStringValueOfNullInLists` parameter in RowBasedColumnSelectorFactory.java‎ has been removed.

<hr>

This PR has:

- [x] been self-reviewed.


indexing-hadoop/src/main/java/org/apache/druid/indexer/InputRowSerde.java
```diff
@@ -372,9 +372,9 @@ private static void writeBytes(@Nullable byte[] value, ByteArrayDataOutput out)
     }
   }
 
-  private static void writeString(String value, ByteArrayDataOutput out) throws IOException
+  private static void writeString(@Nullable String value, ByteArrayDataOutput out) throws IOException
   {
-    writeBytes(StringUtils.toUtf8(value), out);
+    writeBytes(value == null ? null : StringUtils.toUtf8(value), out);
```
gianm: could use `StringUtils.toUtf8Nullable`

indexing-hadoop/src/test/java/org/apache/druid/indexer/InputRowSerdeTest.java
```diff
@@ -281,4 +281,180 @@ public void testDimensionNullOrDefaultForNumerics()
...
+    List<String> dimValues = out.getDimension("d1");
+    Assert.assertEquals(3, dimValues.size());
+    Assert.assertTrue(dimValues.contains("a"));
+    Assert.assertTrue(dimValues.contains("b"));
+    Assert.assertTrue(dimValues.contains(null)); // null is preserved
+    Assert.assertFalse(dimValues.contains("null"));
+  }
+
+  @Test
+  public void testMultiValueDimensionWithNulls_sqlCompatibleMode()
```
gianm: I guess this is talking about SQL-compatible null-handling mode, which is no longer a thing, since we're now always in what used to be called "SQL-compatible null-handling mode". The references to that being a specific mode should be removed.

processing/src/main/java/org/apache/druid/data/input/Rows.java
```diff
@@ -74,7 +76,7 @@ public static List<String> objectToStrings(final Object inputValue)
       return Collections.emptyList();
     } else if (inputValue instanceof List) {
       // guava's toString function fails on null objects, so please do not use it
-      return ((List<?>) inputValue).stream().map(String::valueOf).collect(Collectors.toList());
+      return ((List<?>) inputValue).stream().map(v -> v == null ? null : String.valueOf(v)).collect(Collectors.toList());
```
gianm: Should change the `Object[]` path too. Also, could use `Evals::asString`.

jtuglu1:
@gianm another thing I discovered investigating this patch is that Hadoop by default does not create all-null columns in a segment(`-Ddruid.indexer.task.storeEmptyColumns=false` by default). Native batch in latest version does... . This is the key difference that showed up in the segment diff. https://github.com/apache/druid/pull/12279. Do you know why this is?

For example, if you were to ingest `[null]` multi-value string, native batch would see this correctly as `null`, whereas Hadoop would not even include this column in the segment which causes headaches when switching from Hadoop to native batch.

gianm:
> @gianm another thing I discovered investigating this patch is that Hadoop by default does not create all-null columns in a segment(`-Ddruid.indexer.task.storeEmptyColumns=false` by default). Native batch in latest version does... . This is the key difference that showed up in the segment diff. #12279. Do you know why this is?

There's a comment in `IndexMergerV9` that says Hadoop indexing uses a constructor that doesn't support the `storeEmptyColumns` configuration yet. In that spot it's hard-coded to `false`. I suppose it would make more sense for that to be hard-coded to `TaskConfig.DEFAULT_STORE_EMPTY_COLUMNS`, i.e., `true`. It would make even more sense for it to respect the actual configuration, which would mean switching to some logic like the `TaskToolboxFactory` uses:

```java
config.buildV10()
            ? indexMergerV10Factory.create()
            : indexMergerV9Factory.create(
                task.getContextValue(Tasks.STORE_EMPTY_COLUMNS_KEY, config.isStoreEmptyColumns())
            )
```

Rather than injecting the `IndexMergerV9` directly.

jtuglu1:
@gianm thanks, that's what I thought. I guess I just wanted to make sure there wasn't any critical piece of null merging that was missing/incompatible with Hadoop for both V9/V10.

extensions-core/avro-extensions/src/test/java/org/apache/druid/data/input/AvroStreamInputRowParserTest.java
```diff
@@ -366,8 +367,9 @@ static void assertInputRowCorrect(InputRow inputRow, List<String> expectedDimens
           Lists.transform(SOME_INT_ARRAY_VALUE, String::valueOf),
           inputRow.getDimension("someIntArray")
       );
+      // For string array, nulls are preserved so use ArrayList (ImmutableList doesn't support nulls)
       Assert.assertEquals(
-          Lists.transform(SOME_STRING_ARRAY_VALUE, String::valueOf),
+          SOME_STRING_ARRAY_VALUE.stream().map(v -> v == null ? null : String.valueOf(v)).collect(Collectors.toList()),
```
clintropolis: nit: this can use `Evals.asString`

indexing-hadoop/src/main/java/org/apache/druid/indexer/HadoopDruidIndexerConfig.java
```diff
@@ -129,10 +132,17 @@ public class HadoopDruidIndexerConfig
     );
     JSON_MAPPER = INJECTOR.getInstance(ObjectMapper.class);
     INDEX_IO = INJECTOR.getInstance(IndexIO.class);
-    INDEX_MERGER_V9 = INJECTOR.getInstance(IndexMergerV9.class);
     HADOOP_KERBEROS_CONFIG = INJECTOR.getInstance(HadoopKerberosConfig.class);
     DATA_SEGMENT_PUSHER = INJECTOR.getInstance(DataSegmentPusher.class);
     PROPERTIES = INJECTOR.getInstance(Properties.class);
+
+    boolean buildV10 = Boolean.parseBoolean(PROPERTIES.getProperty(BUILD_V10_KEY, "false"));
```
clintropolis, line 139: neat, :+1: I ignored hadoop for v10 format stuff, thanks for adding

indexing-hadoop/src/main/java/org/apache/druid/indexer/InputRowSerde.java
```diff
@@ -389,15 +389,20 @@ private static void writeStringArray(List<String> values, ByteArrayDataOutput ou
     }
   }
 
+  @Nullable
   private static String readString(DataInput in) throws IOException
   {
     byte[] result = readBytes(in);
-    return StringUtils.fromUtf8(result);
+    return result == null ? null : StringUtils.fromUtf8(result);
```
clintropolis: i wonder if we should make a `StringUtils.fromUtf8Nullable` that takes `byte[]` (we already have one for `ByteBuffer`)

processing/src/main/java/org/apache/druid/data/input/Rows.java
```diff
@@ -74,15 +77,15 @@ public static List<String> objectToStrings(final Object inputValue)
       return Collections.emptyList();
     } else if (inputValue instanceof List) {
       // guava's toString function fails on null objects, so please do not use it
-      return ((List<?>) inputValue).stream().map(String::valueOf).collect(Collectors.toList());
+      return ((List<?>) inputValue).stream().map(Evals::asString).collect(Collectors.toList());
```
clintropolis, line 80: i think this is a good change because the old behavior was wack, but I'm still tracing through to try to determine the actual impacts of this change.

Besides the hadoop impact, which you fix in this PR, this method seems like it will mostly impact callers of `Row.getDimension` as well as the `toGroupKey` method of this class since it calls `getDimension`.

luckily there are a relatively small number of 'production' callers of these methods

`Row.getDimension`:
Image

`Rows.toGroupKey`:
Image

which look mostly related to partitioning. I think we need to determine if the `null` -> `'null'` coercion is important for these callers, and if so, do the coercion there. I'm uncertain currently but will keep trying to figure it out.

jtuglu1:
@clintropolis @gianm Question: do you know why [this](https://github.com/apache/druid/blob/master/indexing-hadoop/src/main/java/org/apache/druid/indexer/IndexGeneratorJob.java#L622) is `null`? I had to update this to get it to work properly (actually ingest `null`-only columns).

indexing-hadoop/src/main/java/org/apache/druid/indexer/InputRowSerde.java
jtuglu1: I can add that

processing/src/main/java/org/apache/druid/data/input/Rows.java
jtuglu1: @clintropolis I can scope this PR to hadoop-only by creating a separate implementation of the Rows.* methods?

clintropolis: it feels worth figuring this out since the old code seems quite odd to be doing what it is at this layer, so I want to keep looking. It probably would be fine though if we can't figure it out?

jtuglu1: I've summarized the breaking changes in the release notes section.

gianm:
> @clintropolis @gianm Question: do you know why [this](https://github.com/apache/druid/blob/master/indexing-hadoop/src/main/java/org/apache/druid/indexer/IndexGeneratorJob.java#L622) is `null`? I had to update this to get it to work properly (actually ingest `null`-only columns).

I do not know. It's the DimensionsSpec parameter, but I don't recall what happens when that isn't provided.

jtuglu1:
@gianm @clintropolis Looks like with this change there's a single byte difference between native/hadoop-outputted segments in handling of this multi-value string: `[null, "a", "b"]`. With this change, Hadoop properly encodes the dictionary cardinality as 3, whereas native seems to incorrectly encode it as 2. The byte difference is in the dictionary cardinality (hadoop represents as 3, while native represents as 2). It doesn't show up in queries, but I believe difference is in `VSizeColumnarMultiInts`.

processing/src/main/java/org/apache/druid/data/input/Rows.java
jtuglu1: I've gone through the callers of these methods and provided some extra tests where I can. From what I can tell, things should work.

processing/src/main/java/org/apache/druid/segment/RowBasedColumnSelectorFactory.java
```diff
@@ -411,15 +411,14 @@ private void updateCurrentValues()
                 for (final Object item : ((List) rawValue)) {
                   final String itemString;
 
+                  // When useStringValueOfNullInLists is true, null items are converted to the string "null".
+                  // When false, null items are preserved as actual null values, consistent with Rows.objectToStrings.
                   if (useStringValueOfNullInLists) {
```
clintropolis: hmm, pretty sure `useStringValueOfNullInLists` exists for compatibility with `Rows.objectToStrings`, maybe we need to remove this now?

jtuglu1: Yeah I meant to ask about that – I think so. IMO we shouldn't be keeping any of these older flags since they just cause more migration pain down-the-line.

jtuglu1: Removed

processing/src/main/java/org/apache/druid/segment/RowBasedColumnSelectorFactory.java
```diff
@@ -411,15 +403,8 @@ private void updateCurrentValues()
                 for (final Object item : ((List) rawValue)) {
                   final String itemString;
 
-                  if (useStringValueOfNullInLists) {
-                    itemString = String.valueOf(item);
-                  } else {
-                    itemString = item == null ? null : String.valueOf(item);
-                  }
+                  itemString = item == null ? null : String.valueOf(item);
```
clintropolis: nit: heh, i suppose this is another thing that can use `Evals.asString`

jtuglu1: Just removed these refs + one more in Rows.java

PR #18948, opened 2026-01-23
fix json column isNumeric check to properly consider array element selector types

clintropolis:
This was a regression of sorts from refactoring in #18589, the `isNumeric` check was always returning `true` when it really should have only been returning `true` if the field is a numeric type or non-existent. Improved the check to return true if the field is a numeric array with an array element path.

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
+++ b/processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
@@ -1053,7 +1053,10 @@ public abstract class CompressedNestedDataComplexColumn<TKeyDictionary extends I
       if (elementType != null) {
         return elementType.isNumeric();
       }
+      // if element type is null, the field was not an array, so don't consider it as numeric
+      return false;
     }
+    // a non-existent field can be considered numeric via a nil selector
     return field == null;
   }
 


processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -1043,8 +1044,17 @@ public boolean isNumeric(List<NestedPathPart> path)
...
+      final TypeSignature<ValueType> elementType = getColumnHolder(
+          element.nestedField.fieldName,
+          element.nestedField.fieldIndex
+      ).getCapabilities().getElementType();
+      if (elementType != null) {
+        return elementType.isNumeric();
+      }
     }
-    return true;
+    return field == null;
```
cecemei, line 1060: would be more readable if this just returns true (and return false as i commented above). also might be helpful adding explanation like this field doesnt exist, thus we can vectorized....

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
```diff
@@ -1043,8 +1044,17 @@ public boolean isNumeric(List<NestedPathPart> path)
...
       return getColumnHolder(nestedField.fieldName, nestedField.fieldIndex).getCapabilities().isNumeric();
+    } else if (field instanceof NestedArrayElement) {
+      final NestedArrayElement element = (NestedArrayElement) field;
+      final TypeSignature<ValueType> elementType = getColumnHolder(
+          element.nestedField.fieldName,
+          element.nestedField.fieldIndex
+      ).getCapabilities().getElementType();
+      if (elementType != null) {
+        return elementType.isNumeric();
+      }
```
cecemei, line 1055: would be helpful to add some explanation here say if elementType not derived, we dont know this field is numeric or not... also might be just more readable return false directly after the if statement.

processing/src/main/java/org/apache/druid/segment/nested/CompressedNestedDataComplexColumn.java
clintropolis: I guess i was thinking that if we ever add any other types of path parts here then they will default to 'not numeric' instead of 'numeric'

PR #14127, opened 2023-04-20
Compaction: Block input specs not aligned with segmentGranularity.

gianm:
When input intervals are not aligned with segmentGranularity, data may be overshadowed if it lies in the space between the input intervals and the output segmentGranularity.

In MSQ REPLACE, this is a validation error. IMO the same behavior makes sense for compaction tasks. In case anyone was depending on the ability to compact nonaligned intervals, a configuration parameter `allowNonAlignedInterval` is provided. I don't expect it to be used much.

PR #18950, opened 2026-01-24
Fix MSQ compaction state and native interval locking, add test coverage

cecemei:
### Description
This PR fixes several issues with MSQ compaction state and compaction interval locking, and adds comprehensive test coverage for MSQ compaction tasks.

#### Fixed compaction state issues
 - Filter propagation: MSQ compaction runner now propagates filter to compaction state
 - Query granularity propagation: MSQ compaction runner now reads query granularity from inputs and propagates it to compaction state

 #### Fixed compaction interval locking
  - Compaction task now locks umbrella interval of input interval from CompactionTask and existing overlap segments

 #### Added MSQCompactionTaskRunTest
  - New test suite covering MSQ compaction task execution
  - Tests various compaction scenarios including partition handling, granularity changes, and metrics aggregation

 #### Refactored test infrastructure
  - Extracted CompactionTaskRunBase to support both native and MSQ compaction tests
  - Added NativeCompactionTaskRunTest to provide dedicated test coverage for native compaction

#### Added builder pattern support
  - Added toBuilder() method to CompactionState with a new Builder class
  - Added toBuilder() method to DimensionsSpec

###  Key changed/added classes in this PR
  - MSQCompactionTaskRunTest
  - NativeCompactionTaskRunTest
  - CompactionTaskRunBase
  - CompactionState.Builder
  - DimensionsSpec.toBuilder()

### Release note
- Fixed an issue in MSQ compaction where transform spec could be lost in compacted segments.
- Fixed an issue in compaction task where interval locking could be incorrect.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/test/java/org/apache/druid/indexing/common/task/NativeCompactionTaskRunTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/common/task/NativeCompactionTaskRunTest.java
@@ -42,7 +42,7 @@ public class NativeCompactionTaskRunTest extends CompactionTaskRunBase
 
     for (LockGranularity lockGranularity : new LockGranularity[]{LockGranularity.TIME_CHUNK, LockGranularity.SEGMENT}) {
       for (boolean useCentralizedDatasourceSchema : new boolean[]{true}) {
-        for (boolean batchSegmentAllocation : new boolean[]{true}) {
+        for (boolean batchSegmentAllocation : new boolean[]{false, true}) {
           for (boolean useSegmentMetadataCache : new boolean[]{false, true}) {
             for (boolean useConcurrentLocks : new boolean[]{false, true}) {
               for (Interval inputInterval : new Interval[]{TEST_INTERVAL, TEST_INTERVAL_DAY}) {


indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -246,6 +245,15 @@ public CompactionTask(
...
                             ? new NativeCompactionRunner(segmentCacheManagerFactory)
                             : compactionRunner;
+    if (this.compactionRunner.requireAlignedInterval() && this.ioConfig.isAllowNonAlignedInterval()) {
+      throw new IAE(
+          "Invalid config: allowNonAlignedInterval is not allowed by runner[%s]",
+          this.compactionRunner.getClass()
+      );
+    }
+    if (this.compactionRunner.forceDropExisting() && !this.ioConfig.isDropExisting()) {
+      throw new IAE("Invalid config: runner[%s] must run with dropExisting", this.compactionRunner.getClass());
```
capistrant: nit

```suggestion
      throw new IAE("Invalid config: runner[%s] must run with dropExisting=true", this.compactionRunner.getClass());
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -246,6 +245,15 @@ public CompactionTask(
     this.compactionRunner = compactionRunner == null
                             ? new NativeCompactionRunner(segmentCacheManagerFactory)
                             : compactionRunner;
+    if (this.compactionRunner.requireAlignedInterval() && this.ioConfig.isAllowNonAlignedInterval()) {
+      throw new IAE(
+          "Invalid config: allowNonAlignedInterval is not allowed by runner[%s]",
```
capistrant: nit

```suggestion
          "Invalid config: allowNonAlignedInterval=true is not allowed by runner[%s]",
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionRunner.java
```diff
@@ -42,6 +42,28 @@ public interface CompactionRunner
...
+
+  /**
+   * Returns whether this runner requires dropExisting to be enabled for compaction.
+   * When true, the compaction task will fail if dropExisting is not set to true in the IOConfig.
+   *
+   * @return true if dropExisting must be enabled, false otherwise. Default is true.
+   */
+  default boolean forceDropExisting()
+  {
+    return true;
```
capistrant: Was this a nasty bug that MSQ compaction didn't enforce dropExisting being true in the past? cuz like I'm sure there are people using it in production and not setting their io config explicitly.. so I guess their tasks would now fail until they set the config? I always thought dropExisting was a native only config since MSQ always creates tombstones regardless of this value. But I guess if `REPLACE_LEGACY` is used to make any other decisions behind the scenes it could be doing something unexpected?

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -440,8 +448,7 @@ public int getPriority()
   @Override
   public boolean isReady(TaskActionClient taskActionClient) throws Exception
   {
-    final List<DataSegment> segments = segmentProvider.findSegments(taskActionClient);
-    return determineLockGranularityAndTryLockWithSegments(taskActionClient, segments, segmentProvider::checkSegments);
+    return determineLockGranularityAndTryLock(taskActionClient, List.of(segmentProvider.interval));
```
capistrant: Is this what you refer to in `Native runner's index task now locks intervals from CompactionTask instead of existing segments' intervals`? Isn't this code on the path for both native and MSQ compaction? that is at least how I interpret it. Not that I think that makes this wrong, but I'm trying to make sure I understand the scope of the change.

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/destination/SegmentGenerationUtils.java
```diff
@@ -95,12 +100,30 @@ public static DataSchema makeDataSchemaForIngestion(
...
                      .withDimensions(dimensionsAndAggregators.lhs)
                      .withAggregators(dimensionsAndAggregators.rhs.toArray(new AggregatorFactory[0]))
-                     .withGranularity(makeGranularitySpecForIngestion(query, querySpec.getColumnMappings(), isRollupQuery, jsonMapper))
+                     .withGranularity(makeGranularitySpecForIngestion(
+                         query,
+                         querySpec.getColumnMappings(),
+                         isRollupQuery,
+                         jsonMapper
+                     ))
+                     .withTransform(transformSpec)
```
capistrant, line 126: crazy that this has gone so long unnoticed by anyone in the community. Maybe compaction filters are rare (or I guess MSQ compaction adoption may be low). either way, good find

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/destination/SegmentGenerationUtils.java
```diff
@@ -95,12 +100,30 @@ public static DataSchema makeDataSchemaForIngestion(
             destination.getDimensionSchemas()
         );
 
+    final TransformSpec transformSpec;
+    if (query.getFilter() != null) {
+      List<Transform> transforms = new ArrayList<>();
+      for (VirtualColumn vc : query.getVirtualColumns().getVirtualColumns()) {
```
capistrant, line 106: Since CompactionTransformSpec doesn't support transforms, these wouldn't actually end up in lastCompactionState for a data segment even if they existed. I don't know if the answer is to consider supporting transforms for CompactionState so this would flow through or to not bother with this code here, cuz it is confusing to extract the virtual columns and add them to the data schema only for them to get ignored.

I actually was toying with the idea of modifying CompactionTransformSpec in https://github.com/apache/druid/pull/18939 to support virtual columns being passed through for cascading deletion filters that are using virtual columns to extract information from nested fields. Seeing this code and seeing how I was going to accomplish this, I wonder if it is time to support the full TransformSpec in compaction instead of just filter? That is probably out of scope for this PR.

I guess right now we have to decide:
1. leave this code even though it ends up ignored (at least for now)
2. remove it and just extract the filter to pass to the DataSchema builder

what do you think?

capistrant: I thought about it more while I was reviewing the rest of the PR and think that it makes sense to keep the code. This code need not care how it is consumed, creating the accurate dataschema feels right. If the transforms end up getting ignored later for lastCompactionState, so be it.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
capistrant: an aside, I think this would make `determineLockGranularityAndTryLockWithSegments` unused now. do we want to leave that around for future use or remove it?

processing/src/main/java/org/apache/druid/timeline/CompactionState.java
```diff
@@ -79,6 +79,93 @@ public CompactionState(
     this.projections = projections;
   }
 
+  public Builder toBuilder()
+  {
+    return new Builder(this);
+  }
+
+  public static class Builder
```
kfaraz: Nit: It would be cleaner to have the `Builder` class and the `toBuilder` method at the end of this file.

processing/src/main/java/org/apache/druid/timeline/CompactionState.java
```diff
@@ -79,6 +79,93 @@ public CompactionState(
...
+    private GranularitySpec granularitySpec;
+    private List<AggregatorFactory> metricsSpec;
+    @Nullable
+    private List<AggregateProjectionSpec> projections;
+
+    public Builder()
+    {
+    }
+
+    public Builder(CompactionState compactionState)
```
kfaraz: I think this can be private.

processing/src/main/java/org/apache/druid/java/util/common/granularity/Granularities.java
```diff
@@ -57,6 +57,7 @@ public class Granularities
   public static final Granularity FIFTEEN_MINUTE = GranularityType.FIFTEEN_MINUTE.getDefaultGranularity();
   public static final Granularity THIRTY_MINUTE = GranularityType.THIRTY_MINUTE.getDefaultGranularity();
   public static final Granularity HOUR = GranularityType.HOUR.getDefaultGranularity();
+  public static final Granularity THREE_HOUR = GranularityType.THREE_HOUR.getDefaultGranularity();
```
kfaraz: Why is this new granularity needed?

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionRunner.java
cecemei: MSQ actually never read this config, but instead just run with dropExisting, even though it's set to false (which is also the default) in the config. I thought this would be somewhat confusing if ppl really look into what this field means. 

So it's not a real change for compaction runner, the result would still be the same but instead we just want the config to be honest. I was also hesistant about this might break anything, so also considering maybe just revert this change in case it breaks something.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: Yes this is on both path. This would make a different say compaction is for a day, but maybe we only have segments for 1 hour. This actually broke msq runner since the task lock is not that smart to figure out it's just the same thing, but then i felt this might be just before concurrent append & replace times that we try to lock less, we should probably move things to the concurrent append & replace world.

cecemei: removed determineLockGranularityAndTryLockWithSegments

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/destination/SegmentGenerationUtils.java
cecemei: i think there're some tests covering the case of virtual column thats how i discovered this. but yea dataschema is probably only used in compactionstate and it doesnt get lots of attention.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionRunner.java
capistrant: In my test cluster, my msq compact supervisor does start failing to create tasks after deploying your feature branch. I think this would be a sad change to force people into making.

`Caused by: org.apache.druid.java.util.common.IAE: Invalid config: runner[class org.apache.druid.msq.indexing.MSQCompactionRunner] must run with dropExisting`

cecemei: yea i made the dropExisting change in another pr: https://github.com/apache/druid/pull/18968/changes#diff-280dab320bc1341105272f99d3ec8e7f8626c9d117f7affa933bc1a522bd465aL368, but the check here is probably not all that necessary. anyway i removed this check here so the task wont fail.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionRunner.java
```diff
@@ -42,6 +42,17 @@ public interface CompactionRunner
...
 
+  /**
+   * Returns whether this runner requires aligned intervals for compaction.
+   * When true, the compaction task will throw an error if the IOConfig has allowNonAlignedInterval enabled.
+   *
+   * @return true if aligned intervals are required by this runner, false otherwise. Default is true.
+   */
+  default boolean requireAlignedInterval()
+  {
+    return true;
```
kfaraz: I don't think we need a default implementation for this method. There are only 2 concrete classes and we might as well implement this method for both of them for clarity.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -1146,7 +1146,8 @@ private void processProjections(final QueryableIndex index)
           final QueryableIndex projectionIndex = Preconditions.checkNotNull(
               index.getProjectionQueryableIndex(schema.getName())
           );
-          final List<DimensionSchema> columnSchemas = Lists.newArrayListWithExpectedSize(schema.getGroupingColumns().size());
+          final List<DimensionSchema> columnSchemas = Lists.newArrayListWithExpectedSize(schema.getGroupingColumns()
+                                                                                               .size());
```
```suggestion
          final List<DimensionSchema> columnSchemas = Lists.newArrayListWithExpectedSize(
              schema.getGroupingColumns().size()
          );
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
kfaraz: I am not sure if it safe to remove this. IIUC, the original logic was to fetch all the segments that _overlap_ (even if partially) with the `semgentProvider.interval`, then try to lock the umbrella interval of the segments. So the final interval locked may be bigger than the original interval. The idea is that a task will be able to replace only those segments which are fully contained in the interval that is locked by that task.

@cecemei , could you share some more details on the issue that you encountered with the MSQ runner and the hour granularity? Maybe there could be an alternative solution.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -635,12 +640,7 @@ static Map<Interval, DataSchema> createDataSchemasForIntervals(
           toolbox.getEmitter(),
           metricBuilder,
           segmentProvider.dataSource,
-          JodaUtils.umbrellaInterval(
```
kfaraz, line 638: Not sure if it is safe to do this, the umbrella interval need not be the same as `segmentProvider.interval` since there may be segments that have partial overlap with the `segmentProvider.interval`, so the umbrella interval would be bigger.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: ah i was thinking the case when the umbrella interval is smaller than the original interval (which is the case in CompactionTaskRunBase, the segment is hourly and there's only segments in hour 0 - hour 3, so compaction task only locks for 3 hours, but the interval can be all day (see NativeCompactionTaskRunTest). MSQ runner failed when getting a task lock with the following change:

[diff.patch](https://github.com/user-attachments/files/25053997/diff.patch)

> Cannot create a new taskLockPosse for request[TimeChunkLockRequest{lockType=REPLACE, groupId='compact_test_mgmlcmch_2026-02-03T18:37:51.661Z', dataSource='test', interval=2014-01-01T00:00:00.000Z/2014-01-02T00:00:00.000Z, preferredVersion='null', priority=25, revoked=false}] because existing locks[[TaskLockPosse{taskLock=TimeChunkLock{type=REPLACE, groupId='compact_test_mgmlcmch_2026-02-03T18:37:51.661Z', dataSource='test', interval=2014-01-01T00:00:00.000Z/2014-01-01T03:00:00.000Z, version='2026-02-03T18:37:51.673Z', priority=25, revoked=false}, taskIds=[compact_test_mgmlcmch_2026-02-03T18:37:51.661Z]}]] have same or higher priorities

processing/src/main/java/org/apache/druid/java/util/common/granularity/Granularities.java
cecemei: added this for the test because compaction task automatically picks the finest gran (hourly) if it's not specified, but msq runner can only process one interval per task (there's 3 hours).

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: i'm not sure i quite understand the part replace only those segments which are fully contained in the locked interval? do you mean when append segments get upgrade? for compaction, i think testPartialIntervalCompactWithFinerSegmentGranularity proves that a smaller interval lock works?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -131,6 +132,7 @@ public void test_ingestDayGranularity_andCompactToMonthGranularity_andCompactToY
             .withGranularitySpec(
                 new UserCompactionTaskGranularityConfig(Granularities.MONTH, null, null)
             )
+            .withIoConfig(new UserCompactionTaskIOConfig(true))
```
kfaraz: Why was this needed explicitly?

Does this mean that existing supervisor specs which do not specify the `ioConfig` would fail?

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -246,6 +246,12 @@ public CompactionTask(
     this.compactionRunner = compactionRunner == null
                             ? new NativeCompactionRunner(segmentCacheManagerFactory)
                             : compactionRunner;
+    if (this.compactionRunner.requireAlignedInterval() && this.ioConfig.isAllowNonAlignedInterval()) {
+      throw new IAE(
```
kfaraz: Please use `InvalidInput.exception()` instead.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -656,19 +661,32 @@ static Map<Interval, DataSchema> createDataSchemasForIntervals(
...
   }
 
-  private static Iterable<DataSegment> retrieveRelevantTimelineHolders(
-      TaskToolbox toolbox,
+  /**
+   * Retrieves data segments and their umbrella interval from the timeline for the given segment provider.
+   *
+   * @return a pair of umbrella interval and data segments
+   */
+  private static Pair<Interval, Iterable<DataSegment>> retrieveRelevantTimelineHolder(
```
kfaraz: Please avoid using pairs as it sacrifices readability. Instead, let this method return `Iterabe<DataSegment>` and add a new method to compute the umbrella interval given the list of segments, and the `segmentProvider.interval`.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionRunner.java
```diff
@@ -42,6 +42,14 @@ public interface CompactionRunner
 {
   String TYPE_PROPERTY = "type";
 
+  /**
+   * Returns whether this runner requires aligned intervals for compaction.
+   * When true, the compaction task will throw an error if the IOConfig has allowNonAlignedInterval enabled.
+   *
+   * @return true if aligned intervals are required by this runner, false otherwise.
+   */
+  boolean requireAlignedInterval();
```
kfaraz: Nit: I wonder if we ever allow non-aligned intervals even for native compaction.
I don't see that field being set while creating auto-compaction tasks.

I feel adding this new method to the `CompactionRunner` interface for a field which is never used is overkill.
Let's just continue ignoring this field in the MSQ compaction runner as we do today and just log a warning message in MSQ compaction runner if it is ever set to `true`.

I think we will just go ahead and deprecate this field and remove it altogether in a future release.

kfaraz: The docs seem to think so too:
https://druid.apache.org/docs/latest/data-management/manual-compaction/#compaction-io-configuration

kfaraz: https://github.com/apache/druid/pull/14127

I think we have had the parameter for backwards compat for a while, we can probably deprecate it now since auto-compaction does not use it and MSQ does not use it either.

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
```diff
@@ -66,7 +68,46 @@ public class TaskActionTestKit extends ExternalResource
...
+  public TaskActionTestKit setUseCentralizedDatasourceSchema(boolean useCentralizedDatasourceSchema)
+  {
+    if (configFinalized.get()) {
+      throw new IllegalStateException("Test config already finalized");
+    }
+    this.useCentralizedDatasourceSchema = useCentralizedDatasourceSchema;
+    return this;
+  }
+
+  public TaskActionTestKit setBatchSegmentAllocation(boolean batchSegmentAllocation)
```
kfaraz, line 94: Just curious, have we added new tests which need this?

indexing-service/src/test/java/org/apache/druid/indexing/common/task/NativeCompactionTaskRunTest.java
```diff
@@ -0,0 +1,98 @@
...
+  @Parameterized.Parameters(name = "name={0}, inputInterval={5}, segmentGran={6}")
+  public static Iterable<Object[]> constructorFeeder()
+  {
+    final List<Object[]> constructors = new ArrayList<>();
+
+    for (LockGranularity lockGranularity : new LockGranularity[]{LockGranularity.TIME_CHUNK, LockGranularity.SEGMENT}) {
+      for (boolean useCentralizedDatasourceSchema : new boolean[]{true, false}) {
+        for (boolean useSegmentMetadataCache : new boolean[]{false, true}) {
+          for (boolean useConcurrentLocks : new boolean[]{false, true}) {
+            for (Interval inputInterval : new Interval[]{TEST_INTERVAL, TEST_INTERVAL_DAY}) {
```
kfaraz: Instead of nesting a large number of `for` loops, please list out the test cases separately even if it spans multiple lines and seems verbose.

processing/src/main/java/org/apache/druid/java/util/common/granularity/Granularities.java
kfaraz: Let's use one of the existing standard granularities then, like SIX_HOUR or EIGHT_HOUR and update the tests accordingly. Would that work?

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
cecemei: ah i think i had some issues with batch segment allocation with segment lock, and some pending segments cache was not removed properly or something (maybe i should have written it down), and have decided to disable the batch allocation to avoid this issue

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: synced offline, the data schema & lock would use the umbrella interval of segments and input interval

indexing-service/src/test/java/org/apache/druid/indexing/common/task/NativeCompactionTaskRunTest.java
cecemei: this has been updated

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
cecemei: yea so it's set to false in CompactionTaskRunBase

kfaraz: Oh, if that is the case, we would need to dig deeper into it since it sounds like a bug and batch segment allocation is the default. It should work for all cases.

cecemei: it's only for segment lock, and i wasn't sure it's due to test setup or anything.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
cecemei: reverted this change, this is because i added dropExisting validation previously

processing/src/main/java/org/apache/druid/java/util/common/granularity/Granularities.java
cecemei: updated to use SIX_HOUR and reverted THREE_HOUR change

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionIOConfig.java
```diff
@@ -60,6 +60,7 @@ public CompactionInputSpec getInputSpec()
     return inputSpec;
   }
 
+  @Deprecated
```
kfaraz, line 66: Nit: Adding a short javadoc indicating why this is deprecated would be nice.

multi-stage-query/src/main/java/org/apache/druid/msq/guice/MSQIndexingModule.java
```diff
@@ -239,4 +248,20 @@ public void configure(Binder binder)
               .to(WindowOperatorQueryKit.class);
     binder.bind(WindowOperatorQueryKit.class).in(LazySingleton.class);
   }
+
+  @Provides
+  IndexerControllerContext.Builder providesContextBuilder(Injector injector)
```
kfaraz: We need not use the `Injector` directly here.

```suggestion
  IndexerControllerContext.Builder getControllerContextBuilder(
      @EscalatedGlobal ServiceClientFactory serviceClientFactory,
      OverlordClient overlordClient
  )
```

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/IndexerControllerContext.java
```diff
@@ -75,6 +75,11 @@
  */
 public class IndexerControllerContext implements ControllerContext
 {
+  public interface Builder
```
kfaraz: It might be cleaner to move this into a separate file and call it `IndexerControllerContextFactory` to align with other similar factory classes in Druid.

multi-stage-query/src/main/java/org/apache/druid/msq/guice/MSQIndexingModule.java
kfaraz: Also, is there any functional change here or is the code just being moved from `MSQControllerTask`?

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/IndexerControllerContext.java
```diff
@@ -108,7 +113,10 @@ public IndexerControllerContext(
     this.clientFactory = clientFactory;
     this.overlordClient = overlordClient;
     this.memoryIntrospector = injector.getInstance(MemoryIntrospector.class);
-    final StorageConnectorProvider storageConnectorProvider = injector.getInstance(Key.get(StorageConnectorProvider.class, MultiStageQuery.class));
+    final StorageConnectorProvider storageConnectorProvider = injector.getInstance(Key.get(
```
kfaraz: Suggestion: Change doesn't seem necessary. It's best to avoid formatting changes unless they really help with readability.

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
kfaraz: Okay, in that case, let's skip the test case for segment lock and add a comment or add a test case and mark it disabled. That way, we would be aware that there is a bug and can address it later.

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
```diff
@@ -66,7 +68,46 @@ public class TaskActionTestKit extends ExternalResource
...
+  {
+    if (configFinalized.get()) {
```
kfaraz, line 96: Why do we want to impose this limit? It should be okay to be able to change the config between tests.
The `before` method would be invoked before every test, right?

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
```diff
@@ -66,7 +68,46 @@ public class TaskActionTestKit extends ExternalResource
...
+  public TaskActionTestKit setUseSegmentMetadataCache(boolean useSegmentMetadataCache)
+  {
+    if (configFinalized.get()) {
+      throw new IllegalStateException("Test config already finalized");
+    }
+    this.useSegmentMetadataCache = useSegmentMetadataCache;
+    return this;
+  }
+
+  public TaskActionTestKit setUseCentralizedDatasourceSchema(boolean useCentralizedDatasourceSchema)
```
kfaraz, line 85: Did centralized schema also run into an issue similar to batch allocation?

processing/src/main/java/org/apache/druid/timeline/CompactionState.java
```diff
@@ -206,4 +206,92 @@ public static Function<Set<DataSegment>, Set<DataSegment>> addCompactionStateToS
         .map(s -> s.withLastCompactionState(compactionState))
         .collect(Collectors.toSet());
   }
+
+
+  public Builder toBuilder()
```
kfaraz, line 211: Nit: Would it be to make this method static?

```
public static Builder builder(CompactionState state) {...}
```

multi-stage-query/src/test/java/org/apache/druid/msq/exec/MSQReplaceTest.java
```diff
@@ -1012,7 +1016,10 @@ public void testReplaceOnFoo1WithWhereExtern(String contextName, Map<String, Obj
                              Collections.emptyList(),
                              Collections.singletonList(new StringDimensionSchema("user")),
                              GranularityType.HOUR,
-                             Intervals.of("2016-06-27T01:00:00.000Z/2016-06-27T02:00:00.000Z")
+                             Intervals.of("2016-06-27T01:00:00.000Z/2016-06-27T02:00:00.000Z"),
+                             new CompactionTransformSpec(
```
kfaraz, line 1020: Do all the tests now have a transform spec? Should we retain some tests without transform spec too?

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
```diff
@@ -411,12 +411,18 @@ private static List<DimensionSpec> getAggregateDimensions(
   {
     List<DimensionSpec> dimensionSpecs = new ArrayList<>();
 
-    if (isQueryGranularityEmptyOrNone(dataSchema)) {
-      // Dimensions in group-by aren't allowed to have time column name as the output name.
-      dimensionSpecs.add(new DefaultDimensionSpec(ColumnHolder.TIME_COLUMN_NAME, TIME_VIRTUAL_COLUMN, ColumnType.LONG));
-    } else {
-      // The changed granularity would result in a new virtual column that needs to be aggregated upon.
-      dimensionSpecs.add(new DefaultDimensionSpec(TIME_VIRTUAL_COLUMN, TIME_VIRTUAL_COLUMN, ColumnType.LONG));
+    if (!dataSchema.getDimensionsSpec().getDimensionNames().contains(ColumnHolder.TIME_COLUMN_NAME)) {
```
kfaraz, line 415: Please add a 1-line comment before this `if` clause clarifying the reason behind the check.

indexing-service/src/test/java/org/apache/druid/indexing/common/task/TestSpyTaskActionClient.java
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Test utility that wraps a {@link TaskActionClient} to track published segments and their schema mappings.
+ * <p>
+ * This spy intercepts {@link SegmentTransactionalInsertAction}, {@link SegmentTransactionalReplaceAction},
+ * and {@link SegmentTransactionalAppendAction} submissions to collect the segments and schema mappings being
+ * published. All other task actions are delegated to the wrapped client without modification.
+ * <p>
+ * Useful for verifying that tasks publish the expected segments and schemas in integration tests.
+ */
+public class TestSpyTaskActionClient implements TaskActionClient
```
kfaraz: This action client is very specific to segment transactions and might not have a wide usage.

I would suggest the following:
- Move this test class to `CompactionTaskRunBase` itself
- Rename it to something simpler `TestTaskActionClient` or `WrappingTaskActionClient`. (The "spy" is a little misleading since it suggests some usage of Mockito spy utilities).
- Avoid javadocs since this code is unit-test-only and seems self explanatory. Add only 1-line javadocs or regular comments where necessary.

multi-stage-query/src/test/java/org/apache/druid/msq/exec/MSQCompactionTaskRunTest.java
```diff
@@ -0,0 +1,600 @@
...
+  public static Iterable<Object[]> constructorFeeder()
+  {
+    final List<Object[]> constructors = new ArrayList<>();
+
+    for (LockGranularity lockGranularity : new LockGranularity[]{LockGranularity.TIME_CHUNK}) {
+      for (boolean useCentralizedDatasourceSchema : new boolean[]{false}) {
+        for (boolean useSegmentMetadataCache : new boolean[]{false, true}) {
+          for (boolean useConcurrentLocks : new boolean[]{false, true}) {
+            for (Interval inputInterval : new Interval[]{TEST_INTERVAL}) {
+              for (Granularity segmentGran : new Granularity[]{Granularities.SIX_HOUR}) {
```
kfaraz: I think only 2 for loops (for `useSegmentMetadataCache` and `useConcurrentLocks`) should be enough here. The other parameters seem to be taking a single value only.

multi-stage-query/src/test/java/org/apache/druid/msq/exec/MSQCompactionTaskRunTest.java
```diff
@@ -0,0 +1,600 @@
...
+                String name = StringUtils.format(
```
kfaraz: This should be done in the `@Paramaters` annotation itself.

indexing-service/src/test/java/org/apache/druid/indexing/common/task/CompactionTaskRunBase.java
```diff
@@ -571,513 +479,277 @@ public void testRunIndexAndCompactAtTheSameTimeForDifferentInterval() throws Exc
...
-            segments.get(i).getShardSpec()
-        );
-      } else {
-        Assert.assertEquals(new NumberedShardSpec(0, 1), segments.get(i).getShardSpec());
-      }
-    }
-  }
-
-  @Test
-  public void testWithSegmentGranularity() throws Exception
```
kfaraz, line 610: Is this test not needed anymore?

indexing-service/src/test/java/org/apache/druid/indexing/common/task/CompactionTaskRunBase.java
```diff
@@ -137,30 +138,35 @@
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
-@RunWith(Parameterized.class)
-public class CompactionTaskRunTest extends IngestionTestBase
+public abstract class CompactionTaskRunBase
```
kfaraz, line 148: The diff for this class seems big and it is a little difficult to ensure that no assertions have actually changed.
Since we are changing some core logic in this PR, I would advise minimizing the changes to this test class so that we can be certain that the new code works correctly with all existing tests.

If any refactor is needed in this test, we can do it in a follow up PR.
For the time being, if some methods need to be reused for the `MSQCompactionTaskRunTest`, they may be copied over or we may make them `public static` where applicable.

kfaraz (commented):
Thanks for your patience, @cecemei !
I have left some final thoughts. We should be good to merge once these are addressed.

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
cecemei: it's not supported by msq engine, schema is not saved

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
cecemei: the before method was called before @test, wont that mean we can't change the config between tests, but in setup only?

processing/src/main/java/org/apache/druid/timeline/CompactionState.java
cecemei: ah i thought this toBuilder() is more fluent style? like:
> getDefaultCompactionState().toBuilder().dimensionSpec().build()

multi-stage-query/src/test/java/org/apache/druid/msq/exec/MSQReplaceTest.java
cecemei: tests without filters are still the same (without transform spec).

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/MSQCompactionRunner.java
cecemei: added

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/IndexerControllerContext.java
cecemei: updated

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/IndexerControllerContext.java
cecemei: this has been reverted

multi-stage-query/src/main/java/org/apache/druid/msq/guice/MSQIndexingModule.java
cecemei: there's no functional change, it's just easier for testing so that we can inject the context in tests

indexing-service/src/test/java/org/apache/druid/indexing/common/task/TestSpyTaskActionClient.java
cecemei: this is done

multi-stage-query/src/main/java/org/apache/druid/msq/guice/MSQIndexingModule.java
cecemei: controller context need the injector... i originally added separate class as well but i think some tests breaks since the guice module asks for class instance even before it asks to provide context instance. it's not ideal but injector is in many places.

indexing-service/src/test/java/org/apache/druid/indexing/common/task/CompactionTaskRunBase.java
cecemei: motivation to change this test is that it doesnt cover msq runner, and while re-writing it i find some tests too long to read, and some issues in msq runner have been discovered due to making it run with msq runner. the test coverage has improved a lot in this class, priori to this pr, this test only tests native compaction runner and also dont use a real task action client. i was surprised by how little test coverage we have on msq compaction runner, wanted to use this test file for any msq compaction runner related change in the future.

the changes in this pr actually was mostly not covered by this test (before this pr), only the interval locking is related. i re-ran the old test with the interval lock change, the failures are due to the interval diff in compaction state, and another test failed due to lock interval covers the input interval now. so all seems expected.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionIOConfig.java
cecemei: added

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
cecemei: i could not reproduce the test cases any more, maybe it disappeared after we switched to use the widen interval. anyways, so i added the segmentQueue test param.

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
cecemei: well this is just to prevent ppl from accidently calling it inside test method

indexing-service/src/test/java/org/apache/druid/indexing/common/task/CompactionTaskRunBase.java
cecemei: segment granularity has been tested very throughly in this test after this change, so we dont need it any more.

indexing-service/src/test/java/org/apache/druid/indexing/common/task/CompactionTaskRunBase.java
kfaraz: Cool, thanks for sharing the details.

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
kfaraz: Why do we need to protect devs against that? This is test-only code.

> wont that mean we can't change the config between tests, but in setup only?

If the setup method (i.e. `before`) is called before every `@Test`, then each test can potentially have its own setup with a different set of configs if needed. Changing the config while a `@Test` method is in progress will have no effect anyway.

multi-stage-query/src/main/java/org/apache/druid/msq/guice/MSQIndexingModule.java
kfaraz: It would still be desirable to pass `OverlordClient` and `ServiceClientFactory` as method args so that the dependencies used by this method are clearly visible.

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/TaskActionTestKit.java
kfaraz: Oh, okay. So do the tests now use batch segment allocation or not?

indexing-service/src/test/java/org/apache/druid/indexing/common/task/NativeCompactionTaskRunTest.java
```diff
@@ -0,0 +1,109 @@
...
+  {
+    final List<Object[]> constructors = new ArrayList<>();
+
+    for (LockGranularity lockGranularity : new LockGranularity[]{LockGranularity.TIME_CHUNK, LockGranularity.SEGMENT}) {
+      for (boolean useCentralizedDatasourceSchema : new boolean[]{true}) {
+        for (boolean batchSegmentAllocation : new boolean[]{true}) {
+          for (boolean useSegmentMetadataCache : new boolean[]{false, true}) {
+            for (boolean useConcurrentLocks : new boolean[]{false, true}) {
+              for (Interval inputInterval : new Interval[]{TEST_INTERVAL, TEST_INTERVAL_DAY}) {
+                for (Granularity segmentGran : new Granularity[]{null, Granularities.HOUR, Granularities.SIX_HOUR}) {
```
kfaraz, line 49: The 7-level for loops look really weird and make reading the test cases difficult.

Given that some params are deprecated and some params are independent of each other, you can greatly reduce the number of test cases to keep only those that add real value.
For example, you need not test out segment locking with concurrent locks or segment metadata cache.
Then you could just list out the test cases in separate lines.

---

Alternatively, you could consider adding a small utility class which generates the parameters, but I suppose this need not be done in this PR itself.

It could have a syntax like (I don't know if JUnit already has a class like this):

```java
final List<Object[]> testCases = new CartesianTestCases()
      .withParam("batchSegmentAllocation", true)
      .withParam("useSegmentMetadataCache", true, false)
      .withParam("inputInterval", INTERVAL_1, INTERVAL_2)
      .generateWithName();
```

PR #18952, opened 2026-01-25
Faster bucket search in ByteBufferHashTable

puzpuzpuz:
### Description

Adds hash code comparison for large enough keys to `ByteBufferHashTable#findBucket()`. Also, changes key comparison to use long/int/byte instead of byte-only comparison (thus, the comparison is now closer to `HashTableUtils#memoryEquals()` used in `MemoryOpenHashTable`). These changes are aimed to speed-up bucket search in `ByteBufferHashTable`, especially in high-collision cases.

#### Microbenchmarks

Environment: Ryzen 7900x, Ubuntu 24.04, OpenJDK 64-bit 17.0.17

Before:
```
Benchmark                                (keySize)  Mode  Cnt   Score   Error  Units
ByteBufferHashTableBenchmark.findBucket          8  avgt    5  19.965 ± 0.212  ns/op
ByteBufferHashTableBenchmark.findBucket         16  avgt    5  26.816 ± 1.282  ns/op
ByteBufferHashTableBenchmark.findBucket         32  avgt    5  36.174 ± 0.337  ns/op
ByteBufferHashTableBenchmark.findBucket         64  avgt    5  49.581 ± 0.482  ns/op
ByteBufferHashTableBenchmark.findBucket        128  avgt    5  72.990 ± 1.429  ns/op
```

After:
```
Benchmark                                (keySize)  Mode  Cnt   Score   Error  Units
ByteBufferHashTableBenchmark.findBucket          8  avgt    5   5.502 ± 0.338  ns/op
ByteBufferHashTableBenchmark.findBucket         16  avgt    5  11.830 ± 0.046  ns/op
ByteBufferHashTableBenchmark.findBucket         32  avgt    5  15.965 ± 0.135  ns/op
ByteBufferHashTableBenchmark.findBucket         64  avgt    5  20.522 ± 0.069  ns/op
ByteBufferHashTableBenchmark.findBucket        128  avgt    5  29.035 ± 1.806  ns/op
```

#### Release note

Speed-up bucket search in hash table used by GROUP BY

<hr>

 * `ByteBufferHashTable`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
+++ b/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
@@ -202,7 +202,7 @@ public class ByteBufferHashTable
         keyBuffer.limit(entryBuffer.position() + HASH_SIZE + keySize);
         keyBuffer.position(entryBuffer.position() + HASH_SIZE);
 
-        final int keyHash = entryBuffer.getInt(entryBuffer.position()) & 0x7fffffff;
+        final int keyHash = entryBuffer.getInt(entryBuffer.position()) & Groupers.USED_FLAG_MASK;
         final int newBucket = findBucket(true, newBuckets, newTableBuffer, keyBuffer, keyHash);
 
         if (newBucket < 0) {
@@ -308,7 +308,7 @@ public class ByteBufferHashTable
       final int bucketOffset = bucket * bucketSizeWithHash;
       final int storedHashWithUsedFlag = targetTableBuffer.getInt(bucketOffset);
 
-      if ((storedHashWithUsedFlag & 0x80000000) == 0) {
+      if ((storedHashWithUsedFlag & Groupers.USED_FLAG_BIT) == 0) {
         // Found unused bucket before finding our key
         return allowNewBucket ? bucket : -1;
       }
--- a/benchmarks/src/test/java/org/apache/druid/benchmark/ByteBufferHashTableBenchmark.java
+++ b/benchmarks/src/test/java/org/apache/druid/benchmark/ByteBufferHashTableBenchmark.java
@@ -115,7 +115,7 @@ public class ByteBufferHashTableBenchmark
       ByteBuffer keyBuffer = insertedKeysBuffer.slice();
       insertedKeysBuffer.clear(); // Reset limit for next iteration
 
-      int keyHash = Groupers.smear(keyBuffer.getInt(0)) & 0x7FFFFFFF;
+      int keyHash = Groupers.smear(keyBuffer.getInt(0)) & Groupers.USED_FLAG_MASK;
       insertedKeyHashes[i] = keyHash;
 
       int bucket = hashTable.findBucket0(true, keyBuffer, keyHash);
@@ -150,7 +150,7 @@ public class ByteBufferHashTableBenchmark
       lookupKeys[i] = lookupKeysBuffer.slice();
       lookupKeysBuffer.clear();
 
-      lookupKeyHashes[i] = Groupers.smear(lookupKeys[i].getInt(0)) & 0x7FFFFFFF;
+      lookupKeyHashes[i] = Groupers.smear(lookupKeys[i].getInt(0)) & Groupers.USED_FLAG_MASK;
     }
 
     // Fill the rest with existing keys (copy from insertedKeysBuffer)
--- a/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/Groupers.java
+++ b/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/Groupers.java
@@ -48,7 +48,17 @@ public class Groupers
       + "for details."
   );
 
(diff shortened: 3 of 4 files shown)


jtuglu1:
Thanks! Can we include a performance test (see `benchmarks` folder)?

puzpuzpuz:
> Thanks! Can we include a performance test (see `benchmarks` folder)?

@jtuglu1 done in 85869a8fd205e8bc367771925e8920417227c232 - the measurements on my box are in the benchmark description. I've also made hash code check mandatory (previously it was disabled for keys <= 8 bytes).

puzpuzpuz:
Sorry, I accidentally broke the empty bucket check with the earlier commit - fixed it in 2d9520ffbefce78013067d94d52745fc01d0419a. Also updated the benchmark results with the measurements obtained on the latest commit.

puzpuzpuz:
More updates. I've noticed that the change introduced implicit endianess dependency since it now checks an `int` for the empty bit instead of the previous single `byte` check. This should be fixed in 952fab383867443c5126b407be80a212e6525cfe and 7ea333929a62f7a607853b43cb3070b59938e716. No more changes from my side.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
```diff
@@ -300,35 +300,81 @@ protected int findBucket(
...
+      }
+      tableOffset++;
+      keyOffset++;
+      length--;
+    }
+
+    return true;
   }
 
   protected boolean canAllowNewBucket()
```
jtuglu1: Is it possible to add these same byte -> int-based comparisons to `isBucketUsed` and `isOffsetUsed`?

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
```diff
@@ -300,35 +300,81 @@ protected int findBucket(
...
+    // Pre-compute hash with used flag for comparison.
+    final int keyHashWithUsedFlag = Groupers.getUsedFlag(keyHash);
+    final int keyBufferPosition = keyBuffer.position();
+
     while (true) {
       final int bucketOffset = bucket * bucketSizeWithHash;
+      final int storedHashWithUsedFlag = targetTableBuffer.getInt(bucketOffset);
 
-      if ((targetTableBuffer.get(bucketOffset) & 0x80) == 0) {
+      if ((storedHashWithUsedFlag & 0x80000000) == 0) {
```
jtuglu1: nit: while we're here, can we name this mask? It's used in other places below (byte-level mask) and makes it easier to read potentially.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
```diff
@@ -300,35 +300,81 @@ protected int findBucket(
...
+  private static boolean keysEqual(
+      final ByteBuffer tableBuffer,
+      int tableOffset,
+      final ByteBuffer keyBuffer,
+      int keyOffset,
+      int length
+  )
+  {
+    // Compare 8 bytes at a time
+    while (length >= Long.BYTES) {
```
jtuglu1, line 348: Maybe we can save a comparison by switching to a do/while loop since I believe `length` will always be ≥ 8. This likely will not show up in the benchmark, however. Unfortunately we cannot do something like `[[likely]]` in Java I don't think.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
puzpuzpuz: Addressed in 5fb014f77031c39e49a0f4e57c1ef2e0a38a1420

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
puzpuzpuz: I'd rather keep the method correct in the face of smaller keys, if this ever changes in the future. Also, if the keys are always >=8, the first branch will be always taken, so CPU's branch predictor should make it very cheap.

puzpuzpuz:
@jtuglu1 @gianm thanks for the reviews!

FrankChen021:
@puzpuzpuz Thanks for the change. I'm wondering how much does this change improve(like the CPU usage) in a real cluster?

puzpuzpuz:
> @puzpuzpuz Thanks for the change. I'm wondering how much does this change improve(like the CPU usage) in a real cluster?

This is a small change and unlikely it's a significant bottleneck in typical workloads, but I'm guessing here. BTW are there any public benchmarks in which Druid actively participates? If so, it's a good idea to check those to be able to make more educated optimizations (if any required/possible).

PR #18953, opened 2026-01-25
Allow failing on residual for Iceberg filters on non-partition cols

jtuglu1:
### Description

Currently Iceberg ingest extension may ingest more data than is necessary due to residual data occurring from an Iceberg filter on non-partition columns. This adds an option to ignore + log a warning or fail on filters that result in residual so users are aware of this extra data and can action on it.

#### Release note

Allow ignoring/warning/failing on residual for Iceberg filters on non-partition columns.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
+++ b/extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
@@ -183,7 +183,6 @@ public class IcebergInputSource implements SplittableInputSource<List<String>>
     return snapshotTime;
   }
 
-  @Nullable
   @JsonProperty
   public ResidualFilterMode getResidualFilterMode()
   {
--- a/extensions-contrib/druid-iceberg-extensions/src/test/java/org/apache/druid/iceberg/input/IcebergInputSourceTest.java
+++ b/extensions-contrib/druid-iceberg-extensions/src/test/java/org/apache/druid/iceberg/input/IcebergInputSourceTest.java
@@ -265,59 +265,52 @@ public class IcebergInputSourceTest
   @Test
   public void testResidualFilterModeFailWithPartitionedTable() throws IOException
   {
+    // Cleanup default table first
+    tearDown();
     // Create a partitioned table and filter on the partition column
-    TableIdentifier partitionedTableId = TableIdentifier.of(Namespace.of(NAMESPACE), "partitionedTable");
-    createAndLoadPartitionedTable(partitionedTableId);
+    tableIdentifier = TableIdentifier.of(Namespace.of(NAMESPACE), "partitionedTable");
+    createAndLoadPartitionedTable(tableIdentifier);
 
-    try {
-      // Filter on partition column with FAIL mode should succeed (no residual)
-      IcebergInputSource inputSource = new IcebergInputSource(
-          "partitionedTable",
-          NAMESPACE,
-          new IcebergEqualsFilter("id", "123988"),
-          testCatalog,
-          new LocalInputSourceFactory(),
-          null,
-          ResidualFilterMode.FAIL
-      );
-      Stream<InputSplit<List<String>>> splits = inputSource.createSplits(null, new MaxSizeSplitHintSpec(null, null));
-      Assert.assertEquals(1, splits.count());
-    }
-    finally {
-      dropTableFromCatalog(partitionedTableId);
-    }
+    IcebergInputSource inputSource = new IcebergInputSource(
+        "partitionedTable",
+        NAMESPACE,
+        new IcebergEqualsFilter("id", "123988"),
+        testCatalog,
+        new LocalInputSourceFactory(),
+        null,
+        ResidualFilterMode.FAIL
+    );
+    Stream<InputSplit<List<String>>> splits = inputSource.createSplits(null, new MaxSizeSplitHintSpec(null, null));
+    Assert.assertEquals(1, splits.count());
   }
 
   @Test
   public void testResidualFilterModeFailWithPartitionedTableNonPartitionColumn() throws IOException
   {
+    // Cleanup default table first
+    tearDown();
     // Create a partitioned table and filter on a non-partition column
-    TableIdentifier partitionedTableId = TableIdentifier.of(Namespace.of(NAMESPACE), "partitionedTable2");
-    createAndLoadPartitionedTable(partitionedTableId);
+    tableIdentifier = TableIdentifier.of(Namespace.of(NAMESPACE), "partitionedTable2");
+    createAndLoadPartitionedTable(tableIdentifier);
 
-    try {
-      // Filter on non-partition column with FAIL mode should throw exception
-      IcebergInputSource inputSource = new IcebergInputSource(
-          "partitionedTable2",
-          NAMESPACE,
-          new IcebergEqualsFilter("name", "Foo"),
-          testCatalog,
-          new LocalInputSourceFactory(),
-          null,
-          ResidualFilterMode.FAIL
-      );
-      DruidException exception = Assert.assertThrows(
-          DruidException.class,
-          () -> inputSource.createSplits(null, new MaxSizeSplitHintSpec(null, null))
-      );
-      Assert.assertTrue(
-          "Expect residual error to be thrown",
-          exception.getMessage().contains("residual")
-      );
-    }
-    finally {
-      dropTableFromCatalog(partitionedTableId);
-    }
+    // Filter on non-partition column with FAIL mode should throw exception
+    IcebergInputSource inputSource = new IcebergInputSource(
+        "partitionedTable2",
+        NAMESPACE,
+        new IcebergEqualsFilter("name", "Foo"),
+        testCatalog,
+        new LocalInputSourceFactory(),
+        null,
+        ResidualFilterMode.FAIL
+    );
+    DruidException exception = Assert.assertThrows(
+        DruidException.class,
+        () -> inputSource.createSplits(null, new MaxSizeSplitHintSpec(null, null))
+    );
+    Assert.assertTrue(
+        "Expect residual error to be thrown",
+        exception.getMessage().contains("residual")
+    );
   }
 
   @After


extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -70,12 +74,35 @@ public List<String> extractSnapshotDataFiles(
       IcebergFilter icebergFilter,
       DateTime snapshotTime
   )
+  {
+    return extractSnapshotDataFiles(tableNamespace, tableName, icebergFilter, snapshotTime, null);
+  }
```
abhishekrb19: This old signature appears to be unused, can be removed?

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -100,12 +127,47 @@ public List<String> extractSnapshotDataFiles(
...
-                       .forEach(dataFile -> dataFilePaths.add(dataFile.path().toString()));
+
+      Expression detectedResidual = null;
+      for (FileScanTask task : tasks) {
+        dataFilePaths.add(task.file().path().toString());
+
+        // Check for residual filters if mode is not IGNORE
+        if (effectiveMode != ResidualFilterMode.IGNORE && detectedResidual == null) {
+          Expression residual = task.residual();
+          if (residual != null && !residual.equals(Expressions.alwaysTrue())) {
```
abhishekrb19, line 118: Is the residual expression from task guaranteed to be null for filters on partition columns? (i.e., confirming there would be no false negatives)

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -100,12 +127,47 @@ public List<String> extractSnapshotDataFiles(
...
+        String message = StringUtils.format(
+            "Iceberg filter produced residual expression that requires row-level filtering. "
+            + "This typically means the filter is on a non-partition column. "
+            + "Residual rows may be ingested unless filtered by transformSpec. "
+            + "Residual filter: [%s]",
+            detectedResidual
+        );
+
+        if (effectiveMode == ResidualFilterMode.FAIL) {
+          throw new IAE(
```
abhishekrb19: nit: throw `DruidException`

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -100,12 +127,47 @@ public List<String> extractSnapshotDataFiles(
...
+          log.warn(message);
+        }
+      }
 
       long duration = System.currentTimeMillis() - start;
       log.info("Data file scan and fetch took [%d ms] time for [%d] paths", duration, dataFilePaths.size());
     }
+    catch (IAE e) {
+      throw e;
+    }
```
abhishekrb19: Is this catch block needed? It can just fall through to the generic block below.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -70,12 +74,35 @@ public List<String> extractSnapshotDataFiles(
...
+      DateTime snapshotTime,
+      @Nullable ResidualFilterMode residualFilterMode
+  )
   {
     Catalog catalog = retrieveCatalog();
     Namespace namespace = Namespace.of(tableNamespace);
     String tableIdentifier = tableNamespace + "." + tableName;
 
     List<String> dataFilePaths = new ArrayList<>();
+    ResidualFilterMode effectiveMode = residualFilterMode != null ? residualFilterMode : ResidualFilterMode.IGNORE;
```
abhishekrb19: This initialization when the value is null can be done in the `IcebergInputSource` constructor and then we can just pass a non-null mode everywhere.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/ResidualFilterMode.java
```diff
@@ -0,0 +1,77 @@
...
+  public static ResidualFilterMode fromString(String value)
+  {
+    for (ResidualFilterMode mode : values()) {
+      if (mode.value.equalsIgnoreCase(value)) {
+        return mode;
+      }
+    }
+    throw new IllegalArgumentException(
+        "Unknown residualFilterMode: " + value + ". Valid values are: ignore, warn, fail"
+    );
```
abhishekrb19, line 69: `DruidException` for better UX

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -100,12 +127,47 @@ public List<String> extractSnapshotDataFiles(
...
+        if (effectiveMode == ResidualFilterMode.FAIL) {
+          throw new IAE(
+              "%s To allow residual rows, set residualFilterMode to 'ignore' or 'warn', "
+              + "or add a corresponding filter in transformSpec.",
+              message
+          );
+        } else if (effectiveMode == ResidualFilterMode.WARN) {
+          log.warn(message);
+        }
+      }
```
abhishekrb19: IMO it’d be better to always log a warning when a residual filter is detected (regardless of mode), not sure anyone would want just `warn`.

 What do you think about simplifying this to two just two modes: `ignore` (default) and `fail`? (we cadn still keep it as an enum for extensibility)

docs/development/extensions-contrib/iceberg.md
```diff
@@ -139,6 +139,37 @@ java \
...
+## Residual filter handling
+
+When an Iceberg filter is applied on a non-partition column, the filtering happens at the file metadata level only (using column statistics). Files that might contain matching rows are returned, but these files may include "residual" rows that don't actually match the filter. These residual rows would be ingested unless filtered by a `transformSpec` filter on the Druid side.
+
+To control this behavior, you can set the `residualFilterMode` property on the Iceberg input source:
+
+| Mode | Description |
+|------|-------------|
+| `ignore` | Default. Residual rows are ingested unless filtered by `transformSpec`. |
+| `warn` | Log a warning when residual filters are detected, but continue with ingestion. |
```
abhishekrb19: I feel having a separate mode for `warn` is unnecessary. Can we just fold a warning log in both the `ignore` and `fail` modes?

docs/ingestion/input-sources.md
```diff
@@ -1063,6 +1063,7 @@ The following is a sample spec for a S3 warehouse source:
 |icebergCatalog|The JSON Object used to define the catalog that manages the configured Iceberg table.|yes|
 |warehouseSource|The JSON Object that defines the native input source for reading the data files from the warehouse.|yes|
 |snapshotTime|Timestamp in ISO8601 DateTime format that will be used to fetch the most recent snapshot as of this time.|no|
+|residualFilterMode|Controls how residual filters are handled when filtering on non-partition columns. When an Iceberg filter targets a non-partition column, files may contain rows that don't match the filter (residual rows). Valid values are: `ignore` (default, ingest all rows), `warn` (log a warning but continue), `fail` (fail the ingestion job). Use `fail` to ensure filters only target partition columns.|no|
```
abhishekrb19: fwiw, this is also the same behavior with Delta lake filtering, where the filter predicates are pushed down to partition columns; for filtering on non-partition columns, it's best-effort

---

It might also make sense to update `icebergFilter` in the docs to clarify how filtering on partition columns vs non-partition columns behave and perhaps point to this new property `residualFilterMode`

jtuglu1: Sure – I think this is already clear in the iceberg.md changes:

```
When an Iceberg filter is applied on a non-partition column, the filtering happens at the file metadata level only (using column statistics). Files that might contain matching rows are returned, but these files may include "residual" rows that don't actually match the filter. These residual rows would be ingested unless filtered by a `transformSpec` filter on the Druid side.

To control this behavior, you can set the `residualFilterMode` property on the Iceberg input source:
```

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
jtuglu1: That's fine with me. It's only a single log and reduces branching.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
jtuglu1: I don't think so. If you change partitions, etc. I think it can still be non-null. Partitions created via void transforms are another possibility. If there is residual, it should show up regardless of filtered-column partitioning.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -177,6 +183,13 @@ public DateTime getSnapshotTime()
     return snapshotTime;
   }
 
+  @Nullable
```
abhishekrb19:
```suggestion
```

extensions-contrib/druid-iceberg-extensions/src/test/java/org/apache/druid/iceberg/input/IcebergInputSourceTest.java
```diff
@@ -215,6 +222,104 @@ public void testCaseInsensitiveFiltering() throws IOException
...
+        "Expect residual error to be thrown",
+        exception.getMessage().contains("residual")
+    );
+  }
+
+  @Test
+  public void testResidualFilterModeFailWithPartitionedTable() throws IOException
+  {
+    // Create a partitioned table and filter on the partition column
+    TableIdentifier partitionedTableId = TableIdentifier.of(Namespace.of(NAMESPACE), "partitionedTable");
```
abhishekrb19: Could assign it to `tableIdentifier` and I think we can remove the try - finally block

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -100,12 +107,44 @@ public List<String> extractSnapshotDataFiles(
...
+        String message = StringUtils.format(
+            "Iceberg filter produced residual expression that requires row-level filtering. "
+            + "This typically means the filter is on a non-partition column. "
+            + "Residual rows may be ingested unless filtered by transformSpec. "
+            + "Residual filter: [%s]",
+            detectedResidual
+        );
+
+        if (residualFilterMode == ResidualFilterMode.FAIL) {
+          throw DruidException.forPersona(DruidException.Persona.DEVELOPER)
```
abhishekrb19, line 135: Should this be `Admin` persona rather than `Developer`? (developer is for unexpected things / bugs)

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
abhishekrb19: no longer Nullable:
```suggestion
```

extensions-contrib/druid-iceberg-extensions/src/test/java/org/apache/druid/iceberg/input/IcebergInputSourceTest.java
abhishekrb19: Could assign this to `tableIdentifier` and I think we can remove the try - finally block along with cleanup `dropTableFromCatalog(tableIdentifier)` since it's already part of  `teardown()`?

abhishekrb19 (approved):
Lgtm, thanks!

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
jtuglu1: Considering a developer is the person running the ingests (same as with queries), I figured we'd surface it there. IMO Admin should not be concerned with the state of data, but rather how/whether the data can be ingested at all.

extensions-contrib/druid-iceberg-extensions/src/test/java/org/apache/druid/iceberg/input/IcebergInputSourceTest.java
jtuglu1: We can, I just did separate tables as it means the other table set in setup() would not get dropped without a preemptive cleanup.

PR #18954, opened 2026-01-26
Enhance autoscaling during task rollover

Fly-Style:
This patch improves the autoscaling during task rollover. Previously, despite the alorithmic correctness, the emptyness of`activelyReadingTaskGroups`  was a very rare event in reality (but that accidentally happened to me during the manual testing of this feature), and the task scaling down happened very rarely as well, which is inappropriate given the facts that:
- most tasks' duration is 1 hour+;
- in neighbouring patch https://github.com/apache/druid/pull/18936 are making scaleup for high lag scenarios more aggressive.

**UPD**: to not to block cost-based autoscaler to be used in production, there is a parralel patch allowing the scaledown during task runtime: https://github.com/apache/druid/pull/18958

### Changeset

  - Introduced a two-phase rollover scaling flow: detect/prepare during task duration checks and apply only after all active tasks with the old partition assignment stop.
  - This stopping logic respects `maxAllowedStops` to avoid worker exhaustion and a disruptive ingestion stop.
  - Add missed autoscaler-required task metrics once rollover scaling is applied.
  - Added a testing hook for autoscaler metrics and new unit tests covering rollover scale-down behavior.

#### Key changed class in this PR
`SeekableStreamSupervisor`

### Algorithm

- Phase 1: Gradual stop (in `checkTaskDuration()`) :
  - At least one task hits end of life -> autoscaler recommends a `taskCount`, if needed
  - Set `pendingRolloverTaskCount`, stop tasks (we respect `maxAllowedStops`)
  - The process itself takes multiple cycles to stop all tasks
- Phase 2: Apply scale (in `maybeApplyPendingScaleRollover()`)
  - By reaching conditions, change taskCount, clear allocation info, cleanup.

Full flow is available on the diagram under the spoiler.

<details>

```mermaid
sequenceDiagram
    participant S as Supervisor
    participant ATG as activelyReadingTaskGroups
    participant PCG as pendingCompletionTaskGroups
    participant AS as AutoScaler

    Note over S: Cycle 1 - Scale triggered
    S->>ATG: 8 task groups running
    S->>AS: computeTaskCountForRollover()
    AS-->>S: returns 4
    S->>S: pendingRolloverTaskCount = 4
    S->>ATG: Stop 3 groups (maxAllowedStops)
    S->>PCG: Move 3 groups to publishing
    Note over ATG: 5 groups remain

    Note over S: Cycle 2 - Continue stopping
    S->>PCG: 1 publishing complete
    S->>ATG: Stop 3 more groups
[...]

PR #18958, opened 2026-01-27
Add configurable option to scale-down during task run time for cost-based autoscaler

Fly-Style:
To test the cost-based autoscaler in real-world scenarios, we temporarily allow scale-downs during task runtime. In case if we don't want to scale down immediately.  
Two configuration options has been added:
- `scaleDownDuringTaskRolloverOnly` indicates if task scaling down is limited to periods during task rollovers only. `false` by default, allows scaling down during normal task run time.
- `scaleDownBarrier`, A number of successful scale down attempts which should be skipped to allow prevention of the auto-scaler from scaling down tasks immediately when the computed cost-based metrics fall below this barrier. Setting it to `0` removes the threshold.

There is a parallel PR that tries to implement the most correct behaviour for scaling during task rollover: https://github.com/apache/druid/pull/18954

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
@@ -170,8 +170,9 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
       taskCount = optimalTaskCount;
       scaleDownCounter = 0; // Nullify the scaleDown counter after a successful scaleup too.
       log.info("New task count [%d] on supervisor [%s], scaling up", taskCount, supervisorId);
-    } else if (!config.isScaleDownOnTaskRolloverOnly() && optimalTaskCount < currentTaskCount &&
-               (config.getScaleDownBarrier() == 0 || ++scaleDownCounter == config.getScaleDownBarrier())) {
+    } else if (!config.isScaleDownOnTaskRolloverOnly()
+               && optimalTaskCount < currentTaskCount
+               && ++scaleDownCounter >= config.getScaleDownBarrier()) {
       taskCount = optimalTaskCount;
       scaleDownCounter = 0;
       log.info("New task count [%d] on supervisor [%s], scaling down", taskCount, supervisorId);
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
@@ -43,12 +43,12 @@ import java.util.Objects;
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class CostBasedAutoScalerConfig implements AutoScalerConfig
 {
-  private static final long DEFAULT_SCALE_ACTION_PERIOD_MILLIS = 10 * 60 * 1000; // 10 minutes
-  private static final long DEFAULT_MIN_TRIGGER_SCALE_ACTION_FREQUENCY_MILLIS = 5 * 60 * 1000; // 5 minutes
-  private static final double DEFAULT_LAG_WEIGHT = 0.25;
-  private static final double DEFAULT_IDLE_WEIGHT = 0.75;
-  private static final double DEFAULT_PROCESSING_RATE = 1000.0; // 1000 records/sec per task as default
-  private static final int DEFAULT_SCALE_DOWN_BARRIER = 5; // We delay scale down by 5 * DEFAULT_SCALE_ACTION_PERIOD_MILLIS
+  static final long DEFAULT_SCALE_ACTION_PERIOD_MILLIS = 10 * 60 * 1000; // 10 minutes
+  static final long DEFAULT_MIN_TRIGGER_SCALE_ACTION_FREQUENCY_MILLIS = 5 * 60 * 1000; // 5 minutes
+  static final double DEFAULT_LAG_WEIGHT = 0.25;
+  static final double DEFAULT_IDLE_WEIGHT = 0.75;
+  static final double DEFAULT_PROCESSING_RATE = 1000.0; // 1000 records/sec per task as default
+  static final int DEFAULT_SCALE_DOWN_BARRIER = 5; // We delay scale down by 5 * DEFAULT_SCALE_ACTION_PERIOD_MILLIS
 
   private final boolean enableTaskAutoScaler;
   private final int taskCountMax;
@@ -218,7 +218,7 @@ public class CostBasedAutoScalerConfig implements AutoScalerConfig
     return scaleDownBarrier;
   }
 
-  @JsonProperty
+  @JsonProperty("scaleDownDuringTaskRolloverOnly")
   public boolean isScaleDownOnTaskRolloverOnly()
   {
     return scaleDownDuringTaskRolloverOnly;
--- a/indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfigTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfigTest.java
@@ -24,6 +24,13 @@ import org.apache.druid.jackson.DefaultObjectMapper;
 import org.junit.Assert;
 import org.junit.Test;
 
+import static org.apache.druid.indexing.seekablestream.supervisor.autoscaler.CostBasedAutoScalerConfig.DEFAULT_IDLE_WEIGHT;
+import static org.apache.druid.indexing.seekablestream.supervisor.autoscaler.CostBasedAutoScalerConfig.DEFAULT_LAG_WEIGHT;
+import static org.apache.druid.indexing.seekablestream.supervisor.autoscaler.CostBasedAutoScalerConfig.DEFAULT_MIN_TRIGGER_SCALE_ACTION_FREQUENCY_MILLIS;
+import static org.apache.druid.indexing.seekablestream.supervisor.autoscaler.CostBasedAutoScalerConfig.DEFAULT_PROCESSING_RATE;
+import static org.apache.druid.indexing.seekablestream.supervisor.autoscaler.CostBasedAutoScalerConfig.DEFAULT_SCALE_ACTION_PERIOD_MILLIS;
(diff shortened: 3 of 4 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -157,11 +164,17 @@ public int computeTaskCountForScaleAction()
...
-    // Perform only scale-up actions
+    // Perform scale-up actions; scale-down actions only if configured.
     int taskCount = -1;
     if (optimalTaskCount > currentTaskCount) {
       taskCount = optimalTaskCount;
-      log.info("New task count [%d] on supervisor [%s]", taskCount, supervisorId);
+      scaleDownCounter = 0; // Nullify the scaleDown counter after a successful scaleup too.
+      log.info("New task count [%d] on supervisor [%s], scaling up", taskCount, supervisorId);
+    } else if (!config.isScaleDownOnTaskRolloverOnly() && optimalTaskCount < currentTaskCount &&
+               (config.getScaleDownBarrier() == 0 || ++scaleDownCounter == config.getScaleDownBarrier())) {
```
capistrant: nit suggestion: might be easier cognitively to drop the `||`?

```suggestion
    } else if (!config.isScaleDownOnTaskRolloverOnly() && optimalTaskCount < currentTaskCount && ++scaleDownCounter >= config.getScaleDownBarrier()) {
```

docs/ingestion/supervisor.md
```diff
@@ -208,6 +208,8 @@ The following table outlines the configuration properties related to the `costBa
 |`lagWeight`|The weight of extracted lag value in cost function.| No| 0.25|
 |`idleWeight`|The weight of extracted poll idle value in cost function. | No | 0.75 |
 |`defaultProcessingRate`|A planned processing rate per task, required for first cost estimations. | No | 1000 |
+|`scaleDownBarrier`| A number of successful scale down attempts which should be skipped  to prevent the auto-scaler from scaling down tasks immediately.  | No | 5 |
+|`scaleDownDuringTaskRolloverOnly`| Indicates whether task scaling down is limited to periods during task rollovers only. | No | False |
```
capistrant, line 212: Is it possible that we want this to be `true` as default, if that is like the production behavior we think will be preferred? I say `if that is like the production behavior we think will be preferred` because of this in the description: 
> To test the cost-based autoscaler in real-world scenarios, we temporarily allow scale-downs during task runtime

like instead of temporarily defaulting for testing, default for long term productionalization plans and we can override for folks testing behavior more rapidly?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -90,6 +105,8 @@ public CostBasedAutoScalerConfig(
     this.lagWeight = Configs.valueOrDefault(lagWeight, DEFAULT_LAG_WEIGHT);
     this.idleWeight = Configs.valueOrDefault(idleWeight, DEFAULT_IDLE_WEIGHT);
     this.defaultProcessingRate = Configs.valueOrDefault(defaultProcessingRate, DEFAULT_PROCESSING_RATE);
+    this.scaleDownBarrier = Configs.valueOrDefault(scaleDownBarrier, DEFAULT_SCALE_DOWN_BARRIER);
+    this.scaleDownDuringTaskRolloverOnly = Configs.valueOrDefault(scaleDownDuringTaskRolloverOnly, false);
```
capistrant, line 109: nit: should this follow pattern of default at top of file or are we only doing that for magic numbers with context added in comments?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -43,11 +43,12 @@
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class CostBasedAutoScalerConfig implements AutoScalerConfig
 {
-  private static final long DEFAULT_SCALE_ACTION_PERIOD_MILLIS = 15 * 60 * 1000; // 15 minutes
-  private static final long DEFAULT_MIN_TRIGGER_SCALE_ACTION_FREQUENCY_MILLIS = 1200000; // 20 minutes
+  private static final long DEFAULT_SCALE_ACTION_PERIOD_MILLIS = 10 * 60 * 1000; // 10 minutes
```
capistrant: looks like this caused the default serde test to fail. I assume the one below would also fail assertions if not fixed up as well.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Yep, good catch!

docs/ingestion/supervisor.md
Fly-Style: Actually, that's the case. Scale down during task rollover is now in unsatisfactory condition, and the existence of that PR with `scaleDownDuringTaskRolloverOnly` set to `false` is required for smooth autoscaler testing.
There is a parallel PR: https://github.com/apache/druid/pull/18954, I tried to integrate a new way to gradually stop tasks and scale down, but it is very risky and need a very careful review and testing.

Well, by bad, I did not produce a good PR description. I updated the it.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
Fly-Style: Sorry, did not get your point :)

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
Fly-Style: Fixed: https://github.com/apache/druid/pull/18958/commits/688833e86e53d55ae94b82184e4d7188cb4f2e13

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Fixed: https://github.com/apache/druid/pull/18958/commits/688833e86e53d55ae94b82184e4d7188cb4f2e13

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
capistrant: Just wondering if inlining the default when the rest are constants at the top of the file for just this one config is intentional. I think it is no big deal

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfigTest.java
```diff
@@ -81,10 +93,15 @@ public void testSerdeWithDefaults() throws Exception
...
-    Assert.assertEquals(900000L, config.getScaleActionPeriodMillis());
-    Assert.assertEquals(1200000L, config.getMinTriggerScaleActionFrequencyMillis());
-    Assert.assertEquals(0.25, config.getLagWeight(), 0.001);
-    Assert.assertEquals(0.75, config.getIdleWeight(), 0.001);
+    Assert.assertEquals(DEFAULT_SCALE_ACTION_PERIOD_MILLIS, config.getScaleActionPeriodMillis());
+    Assert.assertEquals(DEFAULT_MIN_TRIGGER_SCALE_ACTION_FREQUENCY_MILLIS, config.getMinTriggerScaleActionFrequencyMillis());
+    Assert.assertEquals(DEFAULT_LAG_WEIGHT, config.getLagWeight(), 0.001);
+    Assert.assertEquals(DEFAULT_IDLE_WEIGHT, config.getIdleWeight(), 0.001);
+    Assert.assertEquals(DEFAULT_PROCESSING_RATE, config.getDefaultProcessingRate(), 0.001);
+    Assert.assertEquals(DEFAULT_SCALE_DOWN_BARRIER, config.getScaleDownBarrier());
```
kfaraz, line 101: Rather than importing the constants, it would be nice to verify the numerical values directly here, so that when the constants change, we consciously update the tests as well.

docs/ingestion/supervisor.md
kfaraz, line 212: Thanks for adding this config, it makes sense to have this option. But I agree with @capistrant in that we would want the default of this flag to be `true` so that scale down happens only on task rollover by default. But we can fix it up once we have addressed the issues with the task rollover bit.

docs/ingestion/supervisor.md
```diff
@@ -208,6 +208,8 @@ The following table outlines the configuration properties related to the `costBa
 |`lagWeight`|The weight of extracted lag value in cost function.| No| 0.25|
 |`idleWeight`|The weight of extracted poll idle value in cost function. | No | 0.75 |
 |`defaultProcessingRate`|A planned processing rate per task, required for first cost estimations. | No | 1000 |
+|`scaleDownBarrier`| A number of successful scale down attempts which should be skipped  to prevent the auto-scaler from scaling down tasks immediately.  | No | 5 |
```
kfaraz, line 211: This config does not seem to be aligned with the whole cost-based story.
It is a little counterintuitive as it essentially negates the decision taken by the auto-scaler.

IIUC, this has been put in as a hack until we have fixed the issue with scaling on task rollover.

Please remove this config and we should try to evaluate the following instead:
- increase the window over which we compute the average lag
- AND/OR add more terms to the cost function
- AND/OR add a config like `minScaleDownDelay` (which serves a role similar to the `scaleDownBarrier` but in terms of time rather than count)

docs/ingestion/supervisor.md
Fly-Style: That's a reason for this option's existence, actually :D

docs/ingestion/supervisor.md
Fly-Style: `minScaleDelay` idea looks better, I like it more 👍🏻

PR #18962, opened 2026-01-28
add tier/storage/capacity metric to make actual tier disk size metrics available for historicals in vsf mode

clintropolis:
This PR adds a new `tier/storage/capacity` metric for the coordinator to emit that is guaranteed to reflect the total `StorageLocation` size configured across all historicals in a tier. Where the existing `tier/total/capacity` is computed based on `maxSize` advertised by the data node servers, `tier/storage/capacity` is computed on a new `storageSize` field that has been added. Mechanically this means `storageSize` has been added to `DruidServer`/`ImmutableDruidServer`/`DruidServerMetadata`/`DataNodeService`, and is populated with `segmentLoaderConfig.getCombinedSize` as an alternative to `maxSize`, which is possibly set by config or computed to a large value when in virtual storage mode.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
@@ -226,6 +226,11 @@ class QueryVirtualStorageTest extends EmbeddedClusterTestBase
         HumanReadableBytes.parse("1MiB"),
         coordinatorEmitter.getLatestMetricEventValue(Stats.Tier.STORAGE_CAPACITY.getMetricName())
     );
+    coordinatorEmitter.waitForEvent(event -> event.hasMetricName(Stats.Tier.TOTAL_CAPACITY.getMetricName()));
+    Assertions.assertEquals(
+        HumanReadableBytes.parse("100MiB"),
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/druid/server/coordination/DruidServerMetadata.java
```diff
@@ -120,11 +130,6 @@ public boolean isSegmentBroadcastTarget()
     return type.isSegmentBroadcastTarget();
   }
 
-  public boolean isSegmentReplicationOrBroadcastTarget()
```
capistrant, line 123: I am assuming this is just drive by cleanup on unused method

capistrant (approved):
looks good. Are you leaving it intentionally undocumented until VSF is officially documented? I guess if we put it in the metrics table now without mentioning vsf people would be confused thinking it is a duplicate. And if you did mention vsf they'd be confused cuz vsf isn't fully documented? lol

server/src/main/java/org/apache/druid/server/coordination/DruidServerMetadata.java
clintropolis: yea

clintropolis:
>looks good. Are you leaving it intentionally undocumented until VSF is officially documented? I guess if we put it in the metrics table now without mentioning vsf people would be confused thinking it is a duplicate. And if you did mention vsf they'd be confused cuz vsf isn't fully documented? lol

yea, i am planning to document all of this stuff when we document vsf

PR #18965, opened 2026-01-29
SQL: Use specialized virtual columns for expression filters.

gianm:
This patch adjusts planning for expression filters to use specialized virtual columns when they exist. This allows them to take advantage of optimizations, such as the ones that are available for JSON_VALUE, even when the overall expression is complicated.

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/join/filter/JoinFilterAnalyzer.java
+++ b/processing/src/main/java/org/apache/druid/segment/join/filter/JoinFilterAnalyzer.java
@@ -361,7 +361,14 @@ public class JoinFilterAnalyzer
     boolean retainRhs = false;
 
     for (Filter filter : orFilter.getFilters()) {
-      if (!joinFilterPreAnalysis.getJoinableClauses().areSomeColumnsFromJoin(filter.getRequiredColumns())) {
+      // The filter must not reference any column from the join, or any post-join virtual columns.
+      final boolean canUseFilter =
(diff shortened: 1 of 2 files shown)


sql/src/main/java/org/apache/druid/sql/calcite/expression/Expressions.java
```diff
@@ -831,25 +831,43 @@ private static DimFilter toSimpleLeafFilter(
...
+      return null;
+    }
+
+    if (virtualColumnRegistry != null && druidExpression.containsSpecializedNodes()) {
+      // Use a virtual column so we can take advantage of specialization.
+      final String virtualColumnName = virtualColumnRegistry.getOrCreateVirtualColumnForExpression(
+          druidExpression,
+          ColumnType.LONG
+      );
+      final String vcExpression = "\"" + virtualColumnName + "\"";
```
clintropolis, line 857: first impression is it looks a bit weird that the expression filter is just an identifier expr a pointer to a vc, but internally that is basically what happens anyway so it seems fine to me after getting past the strangeness, and this i think would also allow expr re-use between filters and other things so :+1:

clintropolis (approved):

sql/src/main/java/org/apache/druid/sql/calcite/expression/Expressions.java
gianm: I did think about using an `equals` filter in this case (like, `equals` with `1L` value) but decided against it because of the fancy coercing that the `expression` does.

PR #18966, opened 2026-01-29
Add metrics and improve logging for row signature flapping.

abhishekrb19:
We keep encountering different row signatures on the Brokers for the same datasource even when the schema has changed. We’ll need to root cause why the schema “flaps” periodically, but my current suspicion is that this originates on the ingestion path when data is sparse + complex<json> columns - when this happens the columns in the row signature gets reordered.

The side effect of this row signature flapping is that SQL queries can be planned into incorrect native queries until the broker’s schema cache eventually reconciles back to the “correct” signature. This leads to sporadic missing data.

To this end, add some metrics and improve logging to include dimension counts in the old vs new signatures (we have really large schemas, so the actual signature get truncated) whenever a row signature change occurs.

Release note
Add  `segment/schemaCache/rowSignature/changed` and `segment/schemaCache/rowSignature/column/count` metrics to get visibility into when the Broker's segment metadata cache's row signature for each datasource is initialized and updated.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/docs/operations/metrics.md
+++ b/docs/operations/metrics.md
@@ -462,9 +462,8 @@ These metrics are emitted by the Druid Coordinator in every run of the correspon
 |`segment/schemaCache/dataSource/removed`|Emitted when a datasource is removed from the Broker cache due to segments being marked as unused.|`dataSource`||
 |`segment/schemaCache/refresh/time`|Time taken to refresh segments in coordinator segment schema cache.|`dataSource`||
 |`segment/schemaCache/backfill/count`|Number of segments for which schema was back filled in the database.|`dataSource`||
-|`segment/schemaCache/rowSignature/initialized`|Emitted when a datasource row signature is initialized for the first time in the Broker's segment metadata cache.|`dataSource`||
 |`segment/schemaCache/rowSignature/changed`|Emitted when the cached row signature on the Broker's segment metadata cache for a datasource changes, indicating schema evolution or some form of flapping.|`dataSource`||
-|`segment/schemaCache/rowSignature/column/count`|Number of columns in the updated row signature for a datasource.|`dataSource`||
+|`segment/schemaCache/rowSignature/column/count`|Number of columns in the row signature on the Broker's segment metadata cache for a datasource when it's initialized or updated.|`dataSource`||
 |`segment/schemaCache/realtime/count`|Number of realtime segments for which schema is cached.||Depends on the number of realtime segments in the cluster.|
 |`segment/schemaCache/used/count`|Number of published used segments for which schema is cached.||Depends on the number of segments in the cluster.|
 |`segment/schemaCache/usedFingerprint/count`|Number of unique schema fingerprints cached for published used segments.||Depends on the number of distinct schema in the cluster.|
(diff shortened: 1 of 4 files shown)


docs/operations/metrics.md
```diff
@@ -462,6 +462,9 @@ These metrics are emitted by the Druid Coordinator in every run of the correspon
 |`segment/schemaCache/dataSource/removed`|Emitted when a datasource is removed from the Broker cache due to segments being marked as unused.|`dataSource`||
 |`segment/schemaCache/refresh/time`|Time taken to refresh segments in coordinator segment schema cache.|`dataSource`||
 |`segment/schemaCache/backfill/count`|Number of segments for which schema was back filled in the database.|`dataSource`||
+|`segment/schemaCache/rowSignature/initialized`|Emitted when a datasource row signature is initialized for the first time in the Broker's segment metadata cache.|`dataSource`||
+|`segment/schemaCache/rowSignature/changed`|Emitted when the cached row signature on the Broker's segment metadata cache for a datasource changes, indicating schema evolution or some form of flapping.|`dataSource`||
+|`segment/schemaCache/rowSignature/column/count`|Number of columns in the updated row signature for a datasource.|`dataSource`||
```
jtuglu1: nit: `Number of columns in the initial/updated row signature for a datasource.`

sql/src/main/java/org/apache/druid/sql/calcite/schema/BrokerSegmentMetadataCache.java
```diff
@@ -358,10 +358,34 @@ private Map<String, PhysicalDatasourceMetadata> queryDataSourceInformation(Set<S
...
+    final ServiceMetricEvent.Builder builder =
+        new ServiceMetricEvent.Builder().setDimension(DruidMetrics.DATASOURCE, dataSource);
+
+    if (oldTable == null) {
+      log.info(
+          "Row signature for datasource[%s] initialized with [%d] columns - signature[%s]",
+          dataSource, newColumnCount, newRowSignature
+      );
+
+      emitMetric(Metric.SCHEMA_ROW_SIGNATURE_INITIALIZED, 1, builder);
```
jtuglu1: Do we really need this `SCHEMA_ROW_SIGNATURE_INITIALIZED` metric? I wonder if it suffices to just look at the earliest `SCHEMA_ROW_SIGNATURE_COLUMN_COUNT` for a given deployment (since it's emitted in both cases). For "one-off" events I tend to lean towards logging them (and then for repeated events use metrics), although I suppose this would catch bugs where the table schema was getting cleared somehow. I'm not sure how much value adding the initialized metric provides given you would be checking logs anyways most likely to see old column count, etc.

abhishekrb19: Right, I mainly added the `SCHEMA_ROW_SIGNATURE_INITIALIZED` metric to differentiate between the different code paths. I've removed it for now since the updated logs and the signature update &. column count metrics should provide better insights as you note.

PR #18968, opened 2026-01-30
introduce clear separation between proposed compaction, eligibility status, and compaction mode, along with improved task state tracking

cecemei:
This PR refactors compaction eligibility evaluation by introducing clear separation between **proposed compaction** (what segments to compact), **eligibility status** (whether compaction is needed), and **compaction mode** (how to compact), along with improved task state tracking.

   ### Motivation

   The previous `CompactionStatus` implementation mixed multiple concerns:
   1. **Eligibility checks**: Is compaction needed based on config?
   2. **Policy decisions**: Should we compact this interval now?
   3. **Task state tracking**: Is a compaction task already running?

This made it difficult to reason about compaction decisions and track the state of compaction candidates through the system.

   ### Key Changes

   #### 1. Introduced `CompactionMode` Enum

   New enum to represent different compaction strategies:
   - `FULL_COMPACTION`: Compact all segments in the interval
   - `NOT_APPLICABLE`: No compaction should be performed

   The mode determines whether a candidate will be queued for compaction or skipped.

   #### 2. Refactored `CompactionCandidate`

   **New structure:**
   - `ProposedCompaction`: Nested class containing the segment list, intervals, and statistics
     - Represents the raw "what should we compact?" without policy decisions
   - `CompactionStatus`: Eligibility evaluation result (COMPLETE, ELIGIBLE, NOT_ELIGIBLE)
     - Tracks compacted vs uncompacted segments and statistics
     - Contains reason for ineligibility when applicable
   - `CompactionMode`: How to perform compaction (FULL_COMPACTION or NOT_APPLICABLE)
   - `policyNote`: Optional note from the policy explaining why this candidate was chosen/skipped

   **Benefits:**
   - Clear separation between the proposed segments and the decision to compact them
   - Easier to track why candidates are skipped or accepted
   - More maintainable and testable code structure

   #### 3. Updated `CompactionStatus`

   **New responsibilities:**
   - Evaluates if segments need compaction based on config settings
   - Returns state: `COMPLETE`, `ELIGIBLE`, or `NOT_ELIGIBLE`
   - For eligible intervals, tracks:
     - Compacted segment statistics
     - Uncompacted segment statistics
   - Provides detailed reason when not eligible

   **Separation from policy:**
   - `CompactionStatus` checks config-based requirements (granularity, rollup, etc.)
   - Policy then decides whether to actually compact eligible intervals

   #### 4. Simplified `CompactionCandidateSearchPolicy`

   **Changed interface:**
   - Removed: `checkEligibilityForCompaction(candidate, taskStatus) → Eligibility`
   - Added: `createCandidate(proposedCompaction, eligibility) → CompactionCandidate`

   **Benefits:**
   - Policies now create the final candidate after receiving eligibility info
   - Can add policy-specific notes and select compaction mode
   - Cleaner separation: status checks config compliance, policy makes scheduling decisions

   #### 5. Improved Task State Tracking

   **New `TaskState` enum in `CompactionCandidate`:**
   - `READY`: No task running, can start compaction
   - `TASK_IN_PROGRESS`: Compaction task already running
   - `RECENTLY_COMPLETED`: Task recently finished, segments not yet updated

   **Benefits:**
   - Clear distinction between eligibility and task state
   - Prevents launching duplicate tasks for the same interval
   - Easier to understand candidate lifecycle

   ### Code Flow Example

   ```
   1. DataSourceCompactibleSegmentIterator finds segments in a time chunk
      → Creates ProposedCompaction

   2. CompactionStatus evaluates config compliance
      → Returns ELIGIBLE with compacted/uncompacted stats

   3. Policy receives ProposedCompaction + CompactionStatus
      → Calls createCandidate() to produce final CompactionCandidate
      → Sets mode = FULL_COMPACTION if should compact, NOT_APPLICABLE if skip
      → Adds policyNote explaining decision

   4. CompactionStatusTracker checks task state
      → Returns READY/TASK_IN_PROGRESS/RECENTLY_COMPLETED

   5. Scheduler decides whether to launch task based on mode + task state
   ```

   ### Testing

   - Added `CompactionCandidateTest` for new candidate structure
   - Added `CompactionStatusBuilderTest` for eligibility evaluation
   - Updated existing tests to work with new architecture

This PR has:

- [x] been self-reviewed.


server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
```diff
@@ -64,18 +62,27 @@ Eligibility checkEligibilityForCompaction(
    */
   class Eligibility
   {
-    public static final Eligibility OK = new Eligibility(true, null);
+    public enum PolicyEligibility
```
kfaraz: Maybe we should call this enum `State` (since this represents the eligibility status) to avoid confusion with `Eligibility` class itself.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
```diff
@@ -64,18 +62,27 @@ Eligibility checkEligibilityForCompaction(
...
+    {
+      FULL_COMPACTION,
+      INCREMENTAL_COMPACTION,
+      NOT_ELIGIBLE,
+      NOT_APPLICABLE
```
kfaraz: How is "not applicable" different from "not eligible"?
I think eligibility should have only 3 possible states - not eligible, incremental and full.

server/src/main/java/org/apache/druid/client/indexing/ClientCompactionIntervalSpec.java
```diff
@@ -38,18 +42,37 @@ public class ClientCompactionIntervalSpec
...
   private final Interval interval;
   @Nullable
+  private final List<SegmentDescriptor> uncompactedSegments;
+  @Nullable
   private final String sha256OfSortedSegmentIds;
 
   @JsonCreator
   public ClientCompactionIntervalSpec(
       @JsonProperty("interval") Interval interval,
+      @JsonProperty("uncompactedSegments") @Nullable List<SegmentDescriptor> uncompactedSegments,
```
kfaraz: It is a little confusing to add segments here since there is already a `SpecificSegmentsSpec` which specifies the set of segments that should be compacted.

I feel we should add a new type of `CompactionInputSpec` which will be used for incremental compaction only. It would have both a non-null interval as well as a non-empty set of segment IDs to compact incrementally. The `CompactionTask` would handle the different input specs accordingly.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
```diff
@@ -64,18 +62,27 @@ Eligibility checkEligibilityForCompaction(
    */
   class Eligibility
```
kfaraz, line 63: Since this class is going to have a wider usage now, let's move it out into a separate file of its own.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -160,21 +169,122 @@ public CompactionStatus getCurrentStatus()
...
+   * <li><b>INCREMENTAL_COMPACTION</b>: Returns a new candidate containing only
+   *     the uncompacted segments (as determined by the evaluator), with status
+   *     PENDING for incremental compaction</li>
+   * </ul>
+   *
+   * @param config       the compaction configuration for the datasource
+   * @param searchPolicy the policy used to determine compaction eligibility
+   * @return a CompactionCandidate with updated status and potentially filtered segments
+   */
+  public CompactionCandidate evaluate(
```
kfaraz: This method should not live here. The computation of the eligiblity should be done in `CompactionStatusTracker`.

The overall workflow would be like this:
- identify a `CompactionCandidate`
- determine its `CompactionStatus`
- determine the eligiblity via `CompactionStatusTracker.computeEligibility()`

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -185,8 +215,18 @@ public Eligibility checkEligibilityForCompaction(
           "Average size[%,d] of uncompacted segments in interval must be at most [%,d]",
           avgSegmentSize, maxAverageUncompactedBytesPerSegment.getBytes()
       );
+    }
+
+    final double uncompactedBytesRatio = (double) uncompacted.getTotalBytes() /
+                                         (uncompacted.getTotalBytes() + candidate.getCompactedStats().getTotalBytes());
+    if (uncompactedBytesRatio < incrementalCompactionUncompactedBytesRatioThreshold) {
+      return Eligibility.incrementalCompaction(
```
kfaraz: Let's not return `incrementalCompaction` here until incremental compaction is actually implemented.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -47,13 +47,16 @@ public class MostFragmentedIntervalFirstPolicy extends BaseCandidateSearchPolicy
...
   private final HumanReadableBytes maxAverageUncompactedBytesPerSegment;
+  private final double incrementalCompactionUncompactedBytesRatioThreshold;
 
   @JsonCreator
   public MostFragmentedIntervalFirstPolicy(
       @JsonProperty("minUncompactedCount") @Nullable Integer minUncompactedCount,
       @JsonProperty("minUncompactedBytes") @Nullable HumanReadableBytes minUncompactedBytes,
       @JsonProperty("maxAverageUncompactedBytesPerSegment") @Nullable
       HumanReadableBytes maxAverageUncompactedBytesPerSegment,
+      @JsonProperty("incrementalCompactionUncompactedBytesRatioThreshold") @Nullable
```
kfaraz: Let's not add this user-facing config until incremental compaction has been implemented.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
```diff
@@ -344,7 +335,7 @@ static DimensionRangePartitionsSpec getEffectiveRangePartitionsSpec(DimensionRan
    * Evaluates {@link #CHECKS} to determine the compaction status of a
    * {@link CompactionCandidate}.
    */
-  private static class Evaluator
+  static class Evaluator
```
kfaraz: Let's keep the evaluator private as before.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
```diff
@@ -357,29 +348,52 @@ private static class Evaluator
...
+     * <ol>
+     *   <li>Validates input bytes are within limits</li>
+     *   <li>Categorizes segments by compaction state (fingerprinted, uncompacted, or unknown)</li>
+     *   <li>Performs fingerprint-based validation if available (fast path)</li>
+     *   <li>Runs detailed checks against unknown states via {@link #CHECKS}</li>
+     * </ol>
+     *
+     * @return Pair of eligibility status and compaction status with reason for first failed check
+     */
+    Pair<CompactionCandidateSearchPolicy.Eligibility, CompactionStatus> evaluate()
```
kfaraz: Since the `CompactionStatus` and eligibility are going to be two separate things now, this class should not deal with eligibility at all. We can retain the old code in this class.

Once the `CompactionStatus` has been computed, that should be passed to `CompactionStatusTracker.computeEligiblity()` to determine the eligibility.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusTracker.java
```diff
@@ -80,10 +80,7 @@ public Set<String> getSubmittedTaskIds()
    * This method assumes that the given candidate is eligible for compaction
    * based on the current compaction config/supervisor of the datasource.
    */
-  public CompactionStatus computeCompactionStatus(
-      CompactionCandidate candidate,
-      CompactionCandidateSearchPolicy searchPolicy
-  )
+  public CompactionStatus computeCompactionStatus(CompactionCandidate candidate)
```
kfaraz: Since we are now decoupling eligibility (i.e. whether or not to run compaction) from the current compaction status (i.e. current degree of compaction in the interval), it would make sense to do the following:
- Repurpose this method to compute the eligibility since the eligibility can be affected by the `lastTaskStatus` but the `CompactionStatus` should not depend on it.
- `CompactionStatus` need not have the `RUNNING` state anymore since it is really associated with the running state of a compaction task.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
```diff
@@ -357,29 +348,52 @@ private static class Evaluator
...
     {
-      final CompactionStatus inputBytesCheck = inputBytesAreWithinLimit();
-      if (inputBytesCheck.isSkipped()) {
-        return inputBytesCheck;
+      final CompactionCandidateSearchPolicy.Eligibility inputBytesCheck = inputBytesAreWithinLimit();
+      if (inputBytesCheck != null) {
```
kfaraz: I think the input bytes check should now be moved out of this class and used in `CompactionStatusTracker.computeEligibility()` instead.

server/src/main/java/org/apache/druid/server/compaction/DataSourceCompactibleSegmentIterator.java
```diff
@@ -329,17 +331,17 @@ private void findAndEnqueueSegmentsToCompact(CompactibleSegmentIterator compacti
         continue;
       }
 
-      final CompactionCandidate candidates = CompactionCandidate.from(segments, config.getSegmentGranularity());
-      final CompactionStatus compactionStatus = CompactionStatus.compute(candidates, config, fingerprintMapper);
-      final CompactionCandidate candidatesWithStatus = candidates.withCurrentStatus(compactionStatus);
+      final CompactionCandidate candidatesWithStatus =
```
kfaraz: I feel this class should remain untouched since we are only computing the `CompactionStatus` here. The eligiblity computation should happen later in the `CompactionJobQueue` only.

server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
```diff
@@ -460,14 +462,27 @@ private static ClientCompactionTaskQuery compactSegments(
     context.put("priority", compactionTaskPriority);
 
     final String taskId = IdUtils.newTaskId(TASK_ID_PREFIX, ClientCompactionTaskQuery.TYPE, dataSource, null);
+    final ClientCompactionIntervalSpec clientCompactionIntervalSpec;
+    Preconditions.checkArgument(entry.getPolicyEligibility() != null, "Must have a policy eligibility");
+    switch (entry.getPolicyEligibility().getPolicyEligibility()) {
+      case FULL_COMPACTION:
+        clientCompactionIntervalSpec = new ClientCompactionIntervalSpec(entry.getCompactionInterval(), null, null);
+        break;
+      case INCREMENTAL_COMPACTION:
```
kfaraz: Let's update this part only when incremental compaction has been implemented on the task side. Also, we should support incremental compaction with `CompactionJobQueue` (i.e. Overlord-based compaction) only.

kfaraz (commented):
Thanks for the PR, @cecemei !

I like the idea of separating the `CompactionStatus` (i.e. the current degree of compaction of an interval) from the `Eligibility` (i.e. whether an interval should be picked for compaction or not).

I have left some suggestions to aid with the separation. Let me know if they make sense.

kfaraz:
> Currently, CompactionStatus conflates two concerns:
    Status: What's the task status of a CompactionCandidate?
    Eligibility: Should we compact this interval now, and how?

@cecemei , just so that we are on the same page, the `CompactionStatus` does not exactly contain information of the compaction "task" (except for the RUNNING state, which we can now remove in this PR). It contains info of the current level of compaction in the interval, i.e. number of compacted segments, number of uncompacted segments and overall state (COMPLETE/PENDING/SKIPPED).

It would be nice to get rid of the SKIPPED state too since it is more of an eligibility thing, but I guess we can leave it for now.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
cecemei: this is done

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
cecemei: this is done

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
cecemei: they are tracked differently in `AutoCompactionSnapshot`, my original implementation only has three states and it failed: https://github.com/apache/druid/actions/runs/21509820801/job/61973864065. eventually i'd want evaluate() to only return eligibility and not status, so complete/pending status should have different eligibility

server/src/main/java/org/apache/druid/server/compaction/DataSourceCompactibleSegmentIterator.java
cecemei: why should it be jobQueue only? both jobQueue and not as a coordinator duty?

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
cecemei: synced offline, DataSourceCompactibleSegmentIterator would be responsible for filtering out ineligible candidates

server/src/main/java/org/apache/druid/server/compaction/DataSourceCompactibleSegmentIterator.java
kfaraz: Coordinator-based compaction will be deprecated soon.
We want to support incremental compaction on Overlord-based compaction only so that it becomes an incentive for users to migrate to compaction supervisors. We followed the same approach with indexing fingerprints as well.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
kfaraz: +1

But this method should still not live here.
We should just have a `withEligibility()` method so that this class remains a bean for the most part.

This method can be moved to `CompactionEligibility.compute()` similar to `CompactionStatus.compute()`.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
kfaraz: @cecemei , I think this class should remain untouched so that `CompactionStatus` can remain completely agnostic of `CompactionEligibility`.

server/src/main/java/org/apache/druid/server/compaction/CompactionEligibility.java
```diff
@@ -0,0 +1,98 @@
...
+package org.apache.druid.server.compaction;
+
+import org.apache.druid.java.util.common.StringUtils;
+
+import java.util.Objects;
+
+/**
+ * Describes the eligibility of an interval for compaction.
+ */
+public class CompactionEligibility
```
kfaraz: We should add a method to this class:

```
public static CompactionEligibility compute(CompactionStatus status, CompactionCandidateSearchPolicy policy) {
   // determine the eligibility here
}
```

server/src/main/java/org/apache/druid/server/compaction/DataSourceCompactibleSegmentIterator.java
```diff
@@ -329,17 +331,17 @@ private void findAndEnqueueSegmentsToCompact(CompactibleSegmentIterator compacti
...
+          CompactionCandidate.from(segments, config.getSegmentGranularity())
+                             .evaluate(config, searchPolicy, fingerprintMapper);
```
kfaraz: This part should probably look more like the following:


```suggestion
      final CompactionCandidate candidates = CompactionCandidate.from(segments, config.getSegmentGranularity());
      final CompactionStatus compactionStatus = CompactionStatus.compute(candidates, config, fingerprintMapper);
      final CompactionEligibility eligibility = CompactionEligibility.compute(compactionStatus, searchPolicy);
      final CompactionCandidate candidatesWithStatus = candidates.withCurrentStatus(compactionStatus).withEligibility(eligibility);
```

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -160,21 +169,123 @@ public CompactionStatus getCurrentStatus()
...
+        segments,
+        umbrellaInterval,
+        compactionInterval,
+        numIntervals,
+        policyEligiblity,
+        status
+    );
+  }
+
+  public CompactionCandidate withPolicyEligibility(CompactionEligibility eligibility)
```
kfaraz: Nit: Thinking about this, I think it would be best to have a single method (e.g. `withStatusAndEligibility()`) that creates a copy of the `CompactionCandidate` with both the status and eligibility populated.
Otherwise, we would always create an intermediate copy unnecessarily.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusTracker.java
kfaraz: I suppose we can now remove this method altogether, since we do not want to use the `lastTaskStatus` anyway.

kfaraz: Not needed anymore.

kfaraz (commented):
Thanks for the update, @cecemei !
I have left some follow-up suggestions to clean up the PR.

server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
kfaraz: Please revert this change since we need not support incremental compaction with Coordinator-based compaction duty.

server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
```diff
@@ -366,6 +364,10 @@ public static ClientCompactionTaskQuery createCompactionTask(
     }
 
     final CompactionEngine compactionEngine = config.getEngine() == null ? defaultEngine : config.getEngine();
+    if (CompactionEngine.MSQ.equals(compactionEngine) && !Boolean.TRUE.equals(dropExisting)) {
```
kfaraz, line 383: Seems like this should be included in #18950 instead.

server/src/main/java/org/apache/druid/client/indexing/ClientCompactionIntervalSpec.java
cecemei: SpecificSegmentsSpec is not supported by msq, and it somewhat felt a bit deprecated to me, maybe because of the segment lock stuff. i kinda like to specify an interval since it gives some certainty, also wonder maybe we should check non-null for interval here? i didnt see any instance with null interval.

server/src/main/java/org/apache/druid/server/compaction/CompactionEligibility.java
cecemei: added a new evaluate static method

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusTracker.java
cecemei: not sure i understand, dont we still track pending/running/completed status?

server/src/main/java/org/apache/druid/server/compaction/DataSourceCompactibleSegmentIterator.java
cecemei: hmm still thinking about this.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
cecemei: there's another pr that'd almost ready, i could also move this incremental part to a separate pr if you feel strong about this need to be separate from eligibility refactor.

cecemei:
> Thanks for the PR, @cecemei !
> 
> I like the idea of separating the `CompactionStatus` (i.e. the current degree of compaction of an interval) from the `Eligibility` (i.e. whether an interval should be picked for compaction or not).
> 
> I have left some suggestions to aid with the separation. Let me know if they make sense.

I addressed most comments except for the coordinator based compaction and incremental compaction feature change. 

The major change is i moved a lot of CompactionStatus stuff to CompactionEligibility, and update the CHECK to return String (null for previous CompactionStatus.COMPLETE, and non-null for previous CompactionStatus.pending...), it seems simpler since after all we're just looking for a reason to do compaction.

server/src/main/java/org/apache/druid/client/indexing/ClientCompactionIntervalSpec.java
kfaraz: > i kinda like to specify an interval since it gives some certainty

Yes, I agree, in the new scheme of things, it is always nice to specify the interval.
That's why I advise adding a new implementation of `CompactionInputSpec` which will be used only for incremental compaction so that it is easy to distinguish the two on the task side as well. We need not piggyback on the existing `CompactionIntervalSpec`.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
kfaraz: Yes, it is best to add this config in the same PR that adds the incremental compaction feature.

server/src/main/java/org/apache/druid/server/compaction/CompactionEligibility.java
```diff
@@ -0,0 +1,932 @@
...
+        throw DruidException.defensive("Unexpected eligibility state[%s]", state);
+    }
+  }
+
+  /**
+   * Evaluates a compaction candidate to determine its eligibility and compaction status.
+   * <p>
+   * This method performs a two-stage evaluation:
+   * <ol>
+   * <li>First, uses {@link Evaluator} to check if the candidate needs compaction
```
capistrant: did we make any changes to the Evaluator logic in moving it to this file or is it just lift and place in new location?

server/src/main/java/org/apache/druid/server/compaction/DataSourceCompactibleSegmentIterator.java
capistrant: For reference, Cece. Changes I have been working on have also been only adding functionality to the supervisor. The IndexingState fingerprinting that compaction uses is only available for supervisors and rule based compaction/reindexing will also be supervisor only. Maybe it is time we officially do a separate PR and mark this duty as deprecated and update docs as such since so much is happening in this space these days

kfaraz:
Had a brief discussion with @cecemei on this. I think there has been some confusion regarding the roles of `CompactionStatus`, `CompactionEligibility` and the task status. It is in part due to the fact that `CompactionStatus` was originally serving all of these roles but not perfectly.

In short, we shouldn't need to make any major modification to any class in this PR except `CompactionEligibility`.

I have shared some details offline. We should be able to get this resolved by Monday.

cecemei:
Hi @kfaraz @capistrant , I made some major changes to this pr, and hopefully it's in a much better state, please take a look when you're available, thanks! The incremental compaction related change has been removed from this pr, i'm going to make another one for that.

server/src/main/java/org/apache/druid/server/compaction/CompactionMode.java
capistrant, line 1: I think some javadocs in this class would be helpful. A class level one + at least a brief sentence of each mode, even if they are a bit self explanatory

server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
```diff
@@ -371,8 +369,11 @@ public static ClientCompactionTaskQuery createCompactionTask(
     }
     final Map<String, Object> autoCompactionContext = newAutoCompactionContext(config.getTaskContext());
 
-    if (candidate.getCurrentStatus() != null) {
-      autoCompactionContext.put(COMPACTION_REASON_KEY, candidate.getCurrentStatus().getReason());
+    if (CompactionMode.NOT_APPLICABLE.equals(candidate.getMode())) {
```
capistrant: Is this supposed to be `if (!CompactionMode.NOT_APPLICABLE.equals(candidate.getMode()))`?

Seems to me like for compaction jobs the reason will be missing now, and for NOT_APPLICABLE a task wouldn't even be started anyways, as an exception would get thrown by `compactSegments`

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
```diff
@@ -346,8 +361,10 @@ static DimensionRangePartitionsSpec getEffectiveRangePartitionsSpec(DimensionRan
    */
   private static class Evaluator
```
capistrant, line 362: I'm not sure how much I like moving away from a more descriptive return type for the individual checks. It feels like a nitpick since functionally everything is the same. And I guess I understand why you wanted to not be using `CompactionStatus` all over the place in the Evaluator like we were. My only alternative suggestion would be yet another simple record like object or something signifying a check with a result and reason (populated if the check failed). The "return null if check passed" just feels a little unintuitive. Maybe a compromise would be explicitly stating how the checks work in javadoc for the CHECKS and FINGERPRINT_CHECKS lists (or in the evalaute method javadoc)

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
capistrant: I guess this conversation is out of date after latest refactorings?

server/src/test/java/org/apache/druid/server/coordinator/duty/CompactSegmentsTest.java
```diff
@@ -1545,17 +1550,20 @@ private void verifySnapshot(
...
-    Assert.assertEquals(expectedByteCountAwaitingCompaction, snapshot.getBytesAwaitingCompaction());
-    Assert.assertEquals(expectedByteCountCompressed, snapshot.getBytesCompacted());
-    Assert.assertEquals(expectedByteCountSkipped, snapshot.getBytesSkipped());
-    Assert.assertEquals(expectedIntervalCountAwaitingCompaction, snapshot.getIntervalCountAwaitingCompaction());
-    Assert.assertEquals(expectedIntervalCountCompressed, snapshot.getIntervalCountCompacted());
-    Assert.assertEquals(expectedIntervalCountSkipped, snapshot.getIntervalCountSkipped());
-    Assert.assertEquals(expectedSegmentCountAwaitingCompaction, snapshot.getSegmentCountAwaitingCompaction());
-    Assert.assertEquals(expectedSegmentCountCompressed, snapshot.getSegmentCountCompacted());
-    Assert.assertEquals(expectedSegmentCountSkipped, snapshot.getSegmentCountSkipped());
+    Assert.assertEquals(new AutoCompactionSnapshot(
```
capistrant, line 1554: will this make it less clear what is causing the difference than having all the individual assertions? your way looks better on the eyes, but I wonder if it makes it more annoying for a dev to identify what exactly they broke if it starts failing?

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
```diff
@@ -282,18 +281,17 @@ private boolean startJobIfPendingAndReady(
     }
 
     // Check if the job is already running, completed or skipped
-    final CompactionStatus compactionStatus = getCurrentStatusForJob(job, policy);
-    switch (compactionStatus.getState()) {
-      case RUNNING:
+    final CompactionCandidate.TaskState candidateState = getCurrentTaskStateForJob(job);
```
kfaraz: Please do not add a new task state. Use the existing `CompactionTaskStatus` class.

server/src/main/java/org/apache/druid/server/compaction/CompactionMode.java
```diff
@@ -0,0 +1,92 @@
...
+
+  public CompactionCandidate createCandidate(
+      CompactionCandidate.ProposedCompaction proposedCompaction,
+      CompactionStatus eligibility
+  )
+  {
+    return createCandidate(proposedCompaction, eligibility, null);
+  }
+
+  public CompactionCandidate createCandidate(
```
kfaraz: We shouldn't really have this method here. The `CompactionMode` is an attribute that we associate to a `CompactionCandidate`. One should not be responsible for creating the other.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
```diff
@@ -48,74 +45,11 @@ public interface CompactionCandidateSearchPolicy
...
-    @Override
-    public String toString()
-    {
-      return "Eligibility{" +
-             "eligible=" + eligible +
-             ", reason='" + reason + '\'' +
-             '}';
-    }
-  }
+  CompactionCandidate createCandidate(CompactionCandidate.ProposedCompaction candidate, CompactionStatus eligibility);
```
kfaraz: It doesn't make sense for the policy to be creating candidates since the candidates are identified much earlier by the `DatasourceCompactibleSegmentIterator` itself. The policy can only check the eligibility of a candidate for compaction.  (I suppose the name `CompactionCandidateSearchPolicy` might be misleading here, since it is really the `CompactionCandidatePickAndPriorizePolicy` but it can be a mouthful, so we can stick with what we have.)

We should retain the original method here. The caller (`CompactionJobQueue`) should just check the eligibility and then decide whether to launch a job or not. If yes, then whether full or incremental.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -37,76 +39,200 @@
...
-      Interval umbrellaInterval,
-      Interval compactionInterval,
-      int numDistinctSegmentIntervals,
-      @Nullable CompactionStatus currentStatus
-  )
+  /**
+   * Used by {@link CompactionStatusTracker#computeCompactionTaskState(CompactionCandidate)}.
+   * The callsite then determines whether to launch compaction task or not.
+   */
+  public enum TaskState
```
kfaraz: Please do not add a new `TaskState`, use `CompactionTaskStatus` instead. If it doesn't contain any info, let's add it there.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -37,76 +39,200 @@
...
-
-  public static CompactionCandidate from(
-      List<DataSegment> segments,
-      @Nullable Granularity targetSegmentGranularity
-  )
+  /**
+   * Non-empty list of segments of a datasource being proposed for compaction.
+   * A proposed compaction typically contains all the segments of a single time chunk.
+   */
+  public static class ProposedCompaction
```
kfaraz: What we need is a way to distinguish between a candidate before and after the status has been computed, so that we can avoid the null checks on `currentStatus`. I don't think this change fully achieves that.

Also, the name `CompactionCandidate` captures the intent better as it defines a potential "candidate" for compaction. So let's stick with it.

Instead of this change, you may consider adding a new class `CompactionCandidateAndStatus` which will be used by both the methods of the search policy. The new class will contain only a `candidate` field and a `status` field.

I don't think we will need this class to contain the `Eligibility` (or the `CompactionMode`) since it will be computed and used only by the `CompactionJobQueue`.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
```diff
@@ -48,74 +45,11 @@ public interface CompactionCandidateSearchPolicy
...
-      CompactionTaskStatus latestTaskStatus
-  );
-
-  /**
-   * Describes the eligibility of an interval for compaction.
+   * @param candidate the proposed compaction
+   * @param eligibility initial eligibility from compaction config checks
+   * @return final compaction candidate
    */
-  class Eligibility
```
kfaraz, line 66: Let's retain the `Eligibility` class since it answers the question:

_"Does the search policy consider this `CompactionCandidate` eligible for compaction?"_

The `Eligibility` class should now also contain an enum field for `CompactionMode`. The rest of the class can remain unchanged.

kfaraz (changes requested):
@cecemei , I feel a lot of changes here are really not needed and probably make the definitions of various compaction entities confusing.
- The only change that is really needed here is adding a `CompactionMode` enum to `Eligibility`.
- The existing terminology `CompactionCandidate`, `CompactionStatus`, `Eligibility`, `CompactionTaskStatus` captures the respective purposes nicely and also provides a clear distinction. Let's stick with the existing terminology unless there is some real gap in the current nomenclature/functionality.
- If you feel it needs some explanation, we may add/update javadocs.
- You may either remove all the refactors from this PR (so that we can quickly unblock incremental compaction in #18996) OR update this PR to only make suggested modifications that do not break the existing definitions.

<hr>

P.S. Existing definitions
|Class|Definition|
|-----|----------|
|`CompactionCandidate`| A potential __candidate__ for compaction, identified by an interval and some segment IDs|
|`CompactionStatus`| Current __status__ of a `CompactionCandidate` indicated by number of compacted/uncompacted segments/bytes in the interval.|
|`Eligibility`| Given a `CompactionCandidate` and `CompactionStatus`, does the policy think it is __eligible__ for compaction?|
|`CompactionTaskStatus`| Status of past/ongoing compaction tasks for a `CompactionCandidate`|

All of these are distinct from each other and should preferably remain so.
There is room for improvement in these classes, but we should do them in a manner that does not completely change the existing semantics.

server/src/main/java/org/apache/druid/server/compaction/DataSourceCompactibleSegmentIterator.java
```diff
@@ -121,9 +125,11 @@ private void populateQueue(SegmentTimeline timeline, List<Interval> skipInterval
           if (!partialEternitySegments.isEmpty()) {
             // Do not use the target segment granularity in the CompactionCandidate
             // as Granularities.getIterable() will cause OOM due to the above issue
-            CompactionCandidate candidatesWithStatus = CompactionCandidate
-                .from(partialEternitySegments, null)
-                .withCurrentStatus(CompactionStatus.skipped("Segments have partial-eternity intervals"));
+            CompactionCandidate candidatesWithStatus =
```
kfaraz: This class probably doesn't need to change in this PR.
The `Eligibility` should be computed and used only by the `CompactionJobQueue` (and maybe `CompactSegments` for backward compatibility).

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusTracker.java
```diff
@@ -80,43 +80,48 @@ public Set<String> getSubmittedTaskIds()
    * This method assumes that the given candidate is eligible for compaction
    * based on the current compaction config/supervisor of the datasource.
    */
-  public CompactionStatus computeCompactionStatus(
```
kfaraz, line 89: The preferable approach here is to completely remove this method and use the existing method `getLatestTaskStatus()`.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
kfaraz: Yeah, I don't like the null returns much myself.
We should update each CHECK to return a `CheckResult` object instead of a `CompactionStatus`.
The rest of this class should probably remain unchanged as it seems to be bloating up the PR without much value added.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
cecemei: I've considered use `org.apache.druid.indexer.TaskState`(the one used in `CompactionTaskStatus` and not quite like it, because it doesn't have READY state (only running/failed/success), we could 1. use a null value for ready state, but then the simulator uses a map of TaskState; 2. or add a new READY/ NOT_RUN state, but then the existing `isRunnable` and `isComplete` method in `org.apache.druid.indexer.TaskState` is somewhat confusing, which we might need to update it if add a new enum value. Given that this is just an enum, i figured it might be ok to use a new class.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
cecemei: The motivation of this new `ProposedCompaction` is exactly to avoid null `CompactionStatus eligibility`, after this PR, the `CompactionCandidate` would always have an eligibility. 

don't we need the uncompacted & compacted stats in `CompactionStatus` to sort `CompactionCandidate`?

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
cecemei: gotcha - can just return `CompactionMode` here and let callsite handle creation of candidate.

server/src/test/java/org/apache/druid/server/coordinator/duty/CompactSegmentsTest.java
cecemei: i find it captures all the diff in the snapshot better, e.x. one less segment in skipped, one more segment in compressed, thus motivates me to make this change.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
kfaraz: Why do you need a ready state? Everything that is currently in the `CompactionJobQueue` is ready anyway.

Having multiple differing states for task makes the definitions very confusing.

The compaction simulator is broken and has never actually been used.
We will either remove it completely and put up an improved version that can work with cascading reindexers too.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidateSearchPolicy.java
kfaraz: No we should continue to use the existing method in `CompactionCandidateSearchPolicy` i.e. `checkEligibility()`.

Just add a new enum inside the `Eligibility` class.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
kfaraz: > don't we need the uncompacted & compacted stats in CompactionStatus to sort CompactionCandidate?

We do, that's why we shouldn't be touching the `CompactionStatus` class at all.

The `Eligibility` is its own thing. It does not need be put inside any other class.
In `CompactionJobQueue`, we will pick a `CompactionCandidate` from the queue and then ask the search policy on whether that candidate is eligible or not. Based on the eligibility, we will launch the job (info of job type i.e incremental or not) will also be included inside the eligibility object returned by the policy.

Similarly, `CompactSegments` can also check the eligibility of a candidate.

So, effectively, we are moving the call `searchPolicy.checkEligibility()` from `CompactionStatusTracker` to `CompactSegments` and `CompactionJobQueue`. We need to do this because we are getting rid of the method `CompactionStatusTracker.computeCompactionStatus()`.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusTracker.java
cecemei: dont we still want to check `final DateTime snapshotTime = segmentSnapshotTime.get()`, and trigger compaction after new segment?

server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
cecemei: true updated

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
cecemei: added a new CheckResult class

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
cecemei: updated to use option 1, and add queuedIntervals to CompactionSimulateResult

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
cecemei: moved search policy check outside of data source iterator to job queue

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusTracker.java
kfaraz: No, we were snapshot time earlier when we didn't have compactedStats and uncompactedStats.
Now, since we have the stats, we can already make better decisions in the policy itself.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
```diff
@@ -91,117 +93,164 @@ public enum State
...
+
   private final State state;
   private final String reason;
-  private final CompactionStatistics compactedStats;
-  private final CompactionStatistics uncompactedStats;
+
+  @Nullable
+  private final CompactionStatistics compacted;
+  @Nullable
+  private final CompactionStatistics uncompacted;
```
kfaraz: Let's keep them non-nullable and init them to zero if needed.
I think this PR is trying to get rid of the null checks (at least in `CompactionCandidate`, so let's stick with that).

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
```diff
@@ -778,22 +808,60 @@ private CompactionStatus transformSpecIsUpToDate(CompactionState lastCompactionS
...
+
+      @Nullable
+      String reason;
+
+      private CheckResult(@Nullable String reason)
+      {
+        this.reason = reason;
+      }
+
+      static CheckResult fail(String reason)
```
kfaraz: to avoid callers having to do `StringUtils.format` repeatedly.
```suggestion
      static CheckResult fail(String reasonFormat, Object... args)
```

server/src/main/java/org/apache/druid/server/compaction/CompactionStatus.java
```diff
@@ -91,117 +93,164 @@ public enum State
...
+           + ", compacted=" + compacted
+           + ", uncompacted=" + uncompacted
+           + '}';
   }
 
   /**
-   * Computes compaction status for the given field. The status is assumed to be
-   * COMPLETE (i.e. no further compaction is required) if the configured value
-   * of the field is null or equal to the current value.
+   * Returns a 'mismatch' reason to be eligible for compaction if config doesn't match, NULL if config matches.
```
kfaraz:
```suggestion
   * Returns a mismatch reason to be eligible for compaction if config doesn't match, PASS if config matches.
```

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusTracker.java
```diff
@@ -80,43 +81,42 @@ public Set<String> getSubmittedTaskIds()
...
-      CompactionCandidate candidateSegments,
+  public void onSkippedCandidate(
+      CompactionCandidateAndStatus candidateSegments,
+      @Nullable String policyNote
+  )
+  {
+    // Nothing to do, used by simulator
+  }
+
+  public void onCompactionTaskStateComputed(
```
kfaraz: Same, let's remove this method too.

PR #18969, opened 2026-01-30
Adjust costs for burst scaleup during heavy lag for cost-based autoscaler

Fly-Style:
Also add an invariant check for scale-down during task runtime


capistrant (commented):
looks fine, but are we sure we don't want to make these magic numbers configurable (even if undocumented) so you can iterate with them faster and then once you find something that seems right, make them constants for the druid 37 release?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -172,6 +172,7 @@ public int computeTaskCountForScaleAction()
       log.info("New task count [%d] on supervisor [%s], scaling up", taskCount, supervisorId);
     } else if (!config.isScaleDownOnTaskRolloverOnly()
                && optimalTaskCount < currentTaskCount
+               && optimalTaskCount > 0
```
capistrant, line 175: good sanity check, but is it purely for safety or is it actually possible for optimalTask count to be invalid in the code as written? If yes, why isn't that having a lower bound of 1 task or something to make this impossible?

Fly-Style: `optimalTaskCount` may be -1 (which means no changes in task count), and we don't want to count it as successful scaledown

capistrant: Oh lol I shoulda remembered that from the last pr I reviewed

PR #18972, opened 2026-01-31
update copyright year to 2026

clintropolis:


capistrant (approved):

clintropolis:
>Should we update the years in licenses.yaml too?
i think we don't directly reference the druid notice in there so those are just the years from whatever notices from the version we depend on

PR #18976, opened 2026-02-02
Add optional plugins to basic cost function in CostBasedAutoScaler

Fly-Style:
## Changes

- separate the logic of pure cost function, making all additional logic opt-in in config;
- `scaleDownBarrier` has been changed to `minScaleDownDelay`, which is now `Duration`;
- changes to high lag fast scaleup: logarithmic scaling formula for idle decay on high lag and task boundaries.

### Details
This change replaces the sqrt-based scaling formula with a logarithmic formula that provides more aggressive emergency recovery at low task counts and millions of lag.

Idle decay: ` ln(lagSeverity) / ln(maxSeverity)`. Less aggressive, scales well with lag growth.

Formula `K = P/(6.4*sqrt(C))` means small task counts get massive K values (emergency recovery), while large task counts get smaller K values (stability).

More details under the hood:

<details>

#### Idle decay on high lag

No decay: 
cost_based_case1_baseline

New idle decay:
cost_based_case1_ln_severity

#### Task boundaries: formula

```
deltaTasks = K * ln(lagSeverity)

where:
  lagSeverity = (aggregateLag / partitionCount) / lagThreshold
  K = (partitionCount / 6.4) / sqrt(currentTaskCount)
```

Constant **6.4** was carefully chosen as the best 'good' multiplier during multiple analysis attempts for different cluster sizes and situations.

##### Details

| Property             | Old (sqrt-based)              | New (logarithmic)        |
|----------------------|-------------------------------|--------------------------|
| Small cluster (C=1)  | Conservative (~4-6 tasks max) | Controlled (~6-12 tasks) |
| Large cluster (C=24) | Moderate                      | Moderate                 |
| Lag response         | Saturates via `x/(x+1)`       | Unbounded via `ln(x)`    |
| Growth factor K      | Increases with sqrt(C)        | Decreases with sqrt(C)   |

### Example Behavior (48 Partitions, threshold=50K)

| Current | Lag | K   | Delta | Target | Valid Range |
|---------|-----|-----|-------|--------|-------------|
| 1       | 5M  | 7.5 | 5.5   | 6-7    | 1-8         |
| 1       | 10M | 7.5 | 10.7  | 12     | 1-12        |
| 3       | 10M | 4.3 | 6.2   | 10     | 3-10        |
| 12      | 10M | 2.2 | 3.1   | 16     | 12-16       |

Plot:

image

</details>

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
@@ -219,20 +219,24 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
     int optimalTaskCount = -1;
     CostResult optimalCost = new CostResult();
 
+    log.info(
+        "Current metrics: avgPartitionLag[%.1f], pollIdleRatio[%.1f], lagWeight[%.1f], idleWeight[%.1f]",
+        metrics.getAggregateLag(),
+        metrics.getPollIdleRatio(),
+        config.getLagWeight(),
+        config.getIdleWeight()
+    );
+
     for (int taskCount : validTaskCounts) {
       CostResult costResult = costFunction.computeCost(metrics, taskCount, config);
       double cost = costResult.totalCost();
+
       log.info(
-          "Proposed task count[%d] has total cost[%.4f] = lagCost[%.4f] + idleCost[%.4f]."
-          + " Stats: avgPartitionLag[%.1f], pollIdleRatio[%.1f], lagWeight[%.1f], idleWeight[%.1f]",
+          "Proposed task count[%d] has total cost[%.4f] = lagCost[%.4f] + idleCost[%.4f].",
           taskCount,
           cost,
           costResult.lagCost(),
-          costResult.idleCost(),
-          metrics.getAggregateLag(),
-          metrics.getPollIdleRatio(),
-          config.getLagWeight(),
-          config.getIdleWeight()
+          costResult.idleCost()
       );
       if (cost < optimalCost.totalCost()) {
         optimalTaskCount = taskCount;
@@ -287,8 +291,8 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
     final int currentPartitionsPerTask = partitionCount / currentTaskCount;
 
     // Minimum partitions per task correspond to the maximum number of tasks (scale up) and vice versa.
-    int minPartitionsPerTask = partitionCount / taskCountMax;
-    int maxPartitionsPerTask = partitionCount / taskCountMin;
+    int minPartitionsPerTask = Math.min(1, partitionCount / taskCountMax);
+    int maxPartitionsPerTask = Math.max(partitionCount, partitionCount / taskCountMin);
 
     if (isTaskCountBoundariesEnabled) {
       maxPartitionsPerTask = Math.min(
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
@@ -62,25 +62,9 @@ public class CostBasedAutoScalerConfig implements AutoScalerConfig
   private final double lagWeight;
   private final double idleWeight;
   private final double defaultProcessingRate;
-  /**
-   * Enables or disables {@code OptimalTaskCountBoundariesPlugin} which allows
-   * considering only task counts within a certain PPT-based window around the current PPT.
-   */
   private final boolean useTaskCountBoundaries;
-  /**
-   * Per-partition lag threshold allowing to activate a burst scaleup to eliminate high lag.
-   */
   private final int highLagThreshold;
-  /**
-   * Represents the minimum duration between successful scale actions.
-   * A higher value implies a more conservative scaling behavior, ensuring that tasks
-   * are not scaled too frequently during workload fluctuations.
-   */
   private final Duration minScaleDownDelay;
-  /**
-   * Indicates whether task scaling down is limited to periods during task rollovers only.
-   * If set to {@code false}, allows scaling down during normal task run time.
-   */
   private final boolean scaleDownDuringTaskRolloverOnly;
 
   @JsonCreator
@@ -225,24 +209,39 @@ public class CostBasedAutoScalerConfig implements AutoScalerConfig
     return defaultProcessingRate;
   }
 
+  /**
+   * Enables or disables the use of task count boundaries derived from the current partitions-per-task (PPT) ratio.
+   */
   @JsonProperty("useTaskCountBoundaries")
   public boolean shouldUseTaskCountBoundaries()
   {
     return useTaskCountBoundaries;
   }
 
+  /**
+   * Per-partition lag threshold allowing to activate a burst scaleup to eliminate high lag.
+   */
   @JsonProperty("highLagThreshold")
   public int getHighLagThreshold()
   {
     return highLagThreshold;
   }
 
+  /**
+   * Represents the minimum duration between successful scale actions.
+   * A higher value implies a more conservative scaling behavior, ensuring that tasks
+   * are not scaled too frequently during workload fluctuations.
+   */
   @JsonProperty
   public Duration getMinScaleDownDelay()
   {
     return minScaleDownDelay;
   }
 
+  /**
+   * Indicates whether task scaling down is limited to periods during task rollovers only.
+   * If set to {@code false}, allows scaling down during normal task run time.
+   */
   @JsonProperty("scaleDownDuringTaskRolloverOnly")
   public boolean isScaleDownOnTaskRolloverOnly()
   {


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -57,24 +58,19 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
...
+
   /**
-   * Defines the step size used for evaluating lag when computing scaling actions.
-   * This constant helps control the granularity of lag considerations in scaling decisions,
-   * ensuring smoother transitions between scaled states and avoiding abrupt changes in task counts.
+   * Controls how fast the additional tasks grow with the square root of current tasks.
+   * This allows bigger jumps when under-provisioned, but growth slows down as the task count increases.
    */
-  private static final int LAG_STEP = 50_000;
+  private static final int SQRT_TASK_COUNT_SCALE_FACTOR = 5;
```
kfaraz: @Fly-Style , I wonder if we shouldn't just remove this limit altogether, at least for the time being. We already have the `maxTaskCount` and `minTaskCount` guardrails. I feel they should be enough for now so that we can see how reactive the auto-scaler really is.

The user should be able to tweak the `idleWeight` and `lagWeight` to ensure that the auto-scaler doesn't always scale up to the `maxTaskCount`.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -57,24 +58,19 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
...
    */
-  private static final int BASE_RAW_EXTRA = 5;
+  private static final int MAX_JUMP = 12;
+
   // Base PPT lag threshold allowing to activate a burst scaleup to eliminate high lag.
-  static final int EXTRA_SCALING_LAG_PER_PARTITION_THRESHOLD = 25_000;
-  // Extra PPT lag threshold allowing activation of even more aggressive scaleup to eliminate high lag,
-  // also enabling lag-amplified idle calculation decay in the cost function (to reduce idle weight).
-  static final int AGGRESSIVE_SCALING_LAG_PER_PARTITION_THRESHOLD = 50_000;
+  static final int EXTRA_SCALING_LAG_PER_PARTITION_THRESHOLD = 50_000;
```
kfaraz: Let's remove this for now to keep things simple.

The high lag would already increase the lag cost which should cause the auto-scaler to choose a high task count anyway. For the time being, let's just focus on the cost function and tweaking the `idleWeight` and `lagWeight` to get the auto-scaler to do the right thing.

The more parameters we add, the less deterministic the entire setup becomes.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -231,7 +227,7 @@ int computeOptimalTaskCount(CostMetrics metrics)
     for (int taskCount : validTaskCounts) {
       CostResult costResult = costFunction.computeCost(metrics, taskCount, config);
       double cost = costResult.totalCost();
-      log.debug(
+      log.info(
           "Proposed task count: %d, Cost: %.4f (lag: %.4f, idle: %.4f)",
```
kfaraz:
```suggestion
          "Proposed task count[%d] has total Cost[%.4f] = lagCost[%.4f] + idleCost[%.4f]",
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -231,7 +227,7 @@ int computeOptimalTaskCount(CostMetrics metrics)
     for (int taskCount : validTaskCounts) {
```
kfaraz, line 230: Let's also add a log line here which prints out the `idleWeight`, `lagWeight`, `aggregateLag` and `poll-to-idle-ratio`.

docs/ingestion/supervisor.md
```diff
@@ -208,7 +208,10 @@ The following table outlines the configuration properties related to the `costBa
 |`lagWeight`|The weight of extracted lag value in cost function.| No| 0.25|
 |`idleWeight`|The weight of extracted poll idle value in cost function. | No | 0.75 |
 |`defaultProcessingRate`|A planned processing rate per task, required for first cost estimations. | No | 1000 |
-|`scaleDownBarrier`| A number of successful scale down attempts which should be skipped  to prevent the auto-scaler from scaling down tasks immediately.  | No | 5 |
+|`useTaskCountBoundaries`|Enables the bounded partitions-per-task window when selecting task counts.|No|`false`|
```
kfaraz: Instead of a boolean flag, should we just make this an integer for the value of partitions-per-task window?

docs/ingestion/supervisor.md
```diff
@@ -208,7 +208,10 @@ The following table outlines the configuration properties related to the `costBa
...
+|`useBurstScaleOnHeavyLag`|Enables burst scale-up when per-partition lag is high.|No|`false`|
+|`highLagThreshold`|Per-partition lag threshold that triggers burst scale-up when `useBurstScaleOnHeavyLag` is enabled.|No|50000|
```
kfaraz: Feels like these two can be merged into a single config for simplicity.
Maybe use a config name like `lagThresholdForBurstScaleUp`.
By default, it can be -1 i.e. no burst scaling.
Any value >= 0 would cause burst scaling to get enabled at that lag threshold.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -88,9 +71,11 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
   private final ServiceMetricEvent.Builder metricBuilder;
   private final ScheduledExecutorService autoscalerExecutor;
   private final WeightedCostFunction costFunction;
+  private OptimalTaskCountBoundariesPlugin boundariesPlugin = null;
```
kfaraz: This is nice! I like the plugin approach.

Couple of suggestions:
- Can we think of each plugin as being a `CostFunction` itself?
- The `WeightedCostFunction` could also implement the `CostFunction` interface.
- The `CostFunction` interface would have a single method:
```java
CostResult computeCost(
    CostMetrics metrics,
    int proposedTaskCount,
    CostBasedAutoScalerConfig config
);
```
- Then we could do something like `f(g(h(weightedCost())))`, where `f`, `g` and `h` are various plugins on top of the cost function.
```java
CostFunction costFunction = new WeightedCostFunction();
if (burstEnabled) {
   costFunction = new BurstFunction(costFunction);
}
if (taskLimitEnabled) {
   costFunction = new TaskBoundariesFunction(costFunction);
}
CostResult costResult = costFunction.computeCost();
```
- This approach would allow `WeightedCostFunction` to remain agnostic of all plugins and it would make adding new plugins much simpler.

docs/ingestion/supervisor.md
Fly-Style: Also thought about that... Well, why not :)

docs/ingestion/supervisor.md
Fly-Style: No, I won't say so. This option be explained better ("intention of this option to make autoscaler more conservative, yada-yada"), but without specific details. The naming definitely may be better.

docs/ingestion/supervisor.md
Fly-Style: Done: https://github.com/apache/druid/pull/18976/commits/e3b27330cd516ecccd06f85b1781e1addf115c07

docs/ingestion/supervisor.md
kfaraz: No, I mean that since we are still in the validation phase, we should not freeze the value of the SQRT_TASK_INCREASE constant. Instead, we should allow it to be configurable for ease of testing, since we are adding a config to enable/disable it anyway.

docs/ingestion/supervisor.md
```diff
@@ -202,13 +202,15 @@ Note: Kinesis is not supported yet, support is in progress.
 
 The following table outlines the configuration properties related to the `costBased` autoscaler strategy:
 
-| Property|Description|Required|Default|
-|---------|---------------------------------------------------|---|-----|
+| Property|Description|Required| Default |
+|---------|---------------------------------------------------|---|--|
```
kfaraz: To adhere to the existing style used in the Druid docs:

```suggestion
| Property|Description|Required|Default|
|---------|-----------|--------|-------|
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/plugins/OptimalTaskCountBoundariesPlugin.java
```diff
@@ -0,0 +1,33 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.indexing.seekablestream.supervisor.autoscaler.plugins;
+
+/**
+ *
+ */
+public class OptimalTaskCountBoundariesPlugin
```
kfaraz: Do you intend to add some methods to this class?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/plugins/BurstScaleUpOnHighLagPlugin.java
```diff
@@ -0,0 +1,89 @@
...
+   * This uses a logarithmic formula for consistent absolute growth:
+   * {@code deltaTasks = K * ln(lagSeverity)}
+   * where {@code K = (partitionCount / 6.4) / sqrt(currentTaskCount)}
+   * <p>
+   * This ensures:
+   * 1. Partition-aware scaling: larger datasets get more aggressive scaling.
+   * 2. Small taskCount's get a massive relative boost, while large taskCount's receive more measured, stable increases.
+   * 3. Logarithmic lag response: diminishing returns at extreme lag values.
+   */
+  public int computeScaleUpBoost(
```
kfaraz: In the current code flow, this class doesn't really behave very much like a plugin. For the time being, we might as well just move this method to `WeightedCostFunction` since that class has to be aware of the plugin anyway.

Once we add the `CostFunction` interface and its implementations, we can move out the method.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
kfaraz: As discussed offline, we are postponing this for a future patch.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Added.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/plugins/OptimalTaskCountBoundariesPlugin.java
Fly-Style: Nope, removed after offline discussion.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/plugins/BurstScaleUpOnHighLagPlugin.java
Fly-Style: Done: https://github.com/apache/druid/pull/18976/commits/de7f15f243af94fa7b499962275fa82fbae85d3e

docs/ingestion/supervisor.md
Fly-Style: It was reworked in https://github.com/apache/druid/pull/18976/commits/75b15d67b903d2b8541721c599c5e44a1a44ee85

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -231,12 +222,17 @@ int computeOptimalTaskCount(CostMetrics metrics)
     for (int taskCount : validTaskCounts) {
       CostResult costResult = costFunction.computeCost(metrics, taskCount, config);
       double cost = costResult.totalCost();
-      log.debug(
-          "Proposed task count: %d, Cost: %.4f (lag: %.4f, idle: %.4f)",
+      log.info(
+          "Proposed task count[%d] has total cost[%.4f] = lagCost[%.4f] + idleCost[%.4f]."
+          + " Stats: avgPartitionLag[%.1f], pollIdleRatio[%.1f], lagWeight[%.1f], idleWeight[%.1f]",
```
kfaraz: Please don't log this here. This line will be logged for every proposed task count.
We should log the stats only once.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -314,32 +322,46 @@ static int[] computeValidTaskCounts(
...
+   * <p>
+   * This uses a logarithmic formula for consistent absolute growth:
+   * {@code deltaTasks = K * ln(lagSeverity)}
+   * where {@code K = (partitionCount / 6.4) / sqrt(currentTaskCount)}
+   * <p>
+   * This ensures that small taskCount's get a massive relative boost,
+   * while large taskCount's receive more measured, stable increases.
    */
-  static int computeExtraMaxPartitionsPerTaskIncrease(
+  static int computeExtraPPTIncrease(
```
kfaraz, line 338: Please simplify this method by returning the max allowed task count itself.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -288,22 +285,33 @@ static int[] computeValidTaskCounts(
...
-    final int effectiveMaxIncrease = MAX_INCREASE_IN_PARTITIONS_PER_TASK + extraIncrease;
 
     // Minimum partitions per task correspond to the maximum number of tasks (scale up) and vice versa.
-    final int minPartitionsPerTask = Math.max(1, currentPartitionsPerTask - effectiveMaxIncrease);
-    final int maxPartitionsPerTask = Math.min(
-        partitionCount,
-        currentPartitionsPerTask + MAX_DECREASE_IN_PARTITIONS_PER_TASK
-    );
+    int minPartitionsPerTask = partitionCount / taskCountMax;
+    int maxPartitionsPerTask = partitionCount / taskCountMin;
```
kfaraz: Should we clamp these to the limits of [1, partitionCount]. Otherwise, they may overflow those bounds.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -62,12 +63,20 @@ public class CostBasedAutoScalerConfig implements AutoScalerConfig
...
+   * considering only task counts within a certain PPT-based window around the current PPT.
    */
-  private final int scaleDownBarrier;
+  private final boolean useTaskCountBoundaries;
+  /**
+   * Per-partition lag threshold allowing to activate a burst scaleup to eliminate high lag.
+   */
+  private final int highLagThreshold;
+  /**
+   * Represents the minimum duration between successful scale actions.
```
kfaraz: Please move these javadocs to the getters instead. That way, it would be easier for callers to look up the javadocs.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Would leave it for some follow-up PR, actually want to start testing that stuff soon.

PR #18977, opened 2026-02-02
Upgrade compiler version to JDK 17

jtuglu1:
Fully deprecates JDK 11 per https://github.com/apache/druid/pull/18759#issuecomment-3554561707. Ideally v37 can remove JDK 17 support altogether and JDK 25 will be the newer supported version (with JDK 21 being deprecated).

#### Release note

Upgrade compiler version to JDK 17. This removes compiler compatibility for indexing-hadoop (no longer supported extension).

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/.github/scripts/strict-compilation.sh
+++ b/.github/scripts/strict-compilation.sh
@@ -18,4 +18,4 @@
 set -e
 set -x
 
-mvn -B clean -DstrictCompile -pl '!embedded-tests' compile test-compile --fail-at-end -P skip-tests -Dweb.console.skip=true -T1C
+mvn -B clean -DstrictCompile compile test-compile --fail-at-end -P skip-tests -Dweb.console.skip=true -T1C
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ Find articles written by community members and a calendar of upcoming events on 
 
 ### Building from source
 
-Please note that JDK 11 or JDK 17 is required to build Druid.
+Please note that JDK 17 or JDK 21 is required to build Druid.
 
 See the latest [build guide](https://druid.apache.org/docs/latest/development/build.html) for instructions on building Apache Druid from source.
 
--- a/embedded-tests/pom.xml
+++ b/embedded-tests/pom.xml
@@ -29,9 +29,6 @@
 
   <properties>
     <surefire.rerunFailingTestsCount>0</surefire.rerunFailingTestsCount>
-    <!--  Hadoop relies upon JDK 11 language level which breaks when upgraded to anything higher, but we can use 17 in \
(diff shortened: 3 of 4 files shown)


FrankChen021:
> and JDK 24 will be the newer supported version.

why not jdk25 which is the LTS after jdk 21?

FrankChen021 (approved):

jtuglu1:
> why not jdk25 which is the LTS after jdk 21?

SGTM 👍.

clintropolis:
>why not jdk25 which is the LTS after jdk 21?

supporting 25 is a bit non-trivial at this point, any extension that uses hadoopy stuff (hdfs storage, parquet, avro, orc readers) will fail with an error related to https://docs.oracle.com/en/java/javase/24/security/security-manager-is-permanently-disabled.html, when i was testing i saw something like

```
java.lang.UnsupportedOperationException: getSubject is not supported
	at java.base/javax.security.auth.Subject.getSubject(Subject.java:277) ~[?:?]
	at org.apache.hadoop.security.UserGroupInformation.getCurrentUser(UserGroupInformation.java:577) ~[?:?]
	at org.apache.hadoop.fs.FileSystem$Cache$Key.<init>(FileSystem.java:3884) ~[?:?]
	at org.apache.hadoop.fs.FileSystem$Cache$Key.<init>(FileSystem.java:3874) ~[?:?]
	at org.apache.hadoop.fs.FileSystem$Cache.get(FileSystem.java:3662) ~[?:?]
	at org.apache.hadoop.fs.FileSystem.get(FileSystem.java:557) ~[?:?]
	at org.apache.hadoop.fs.FileSystem.get(FileSystem.java:290) ~[?:?]
	at org.apache.druid.storage.hdfs.HdfsStorageDruidModule.configure(HdfsStorageDruidModule.java:96) ~[?:?]
	at 
```

I believe hadoop 3.5.0 will fix this if i understand https://issues.apache.org/jira/browse/HADOOP-19744 correctly, so i suppose it is possible we could have this fixed in time for 37... though that would also i think technically eliminate the need to actually remove hadoop indexing from druid (though i think if we don't delete it we should move it to be an actual contrib extension so it isn't bundled by default, which hopefully shouldn't be all that much work after #18247). 

Regardless of 25 support, I think it is fine to move forward with this PR, we are just making a commitment that either we will delete hadoop indexing before 37 (i totally volunteer to do this... it sounds very satisfying), or if 3.5.0 is released and fixes the problems.

I created #18980 to help track 25 support.

jtuglu1:
@clintropolis I also suspect AWS SDK V2 (or libs that depend on those APIs) will be a blocker as well but I intend to get that merged by V37.

kfaraz (approved):

PR #18982, opened 2026-02-04
Support lz4 compression in hdfs

FrankChen021:
### Description

Currently zip(default level 6) is used to compress all files when uploading to hdfs.
the zip compression takes cpu resources and time to upload large files. Another consequence is that it significantly increases task duration and may introduce task pending during task rolling over, which also introduces high lag in kafka.
This PR adds support to using lz4 compression(fastest level) when uploading and decompression during downloading from hdfs.

## Metrics Comparison

- because task duration is reduced, we didn't observe task pending
image

- push duration is significantly reduced from up to 2 minutes to 30 seconds while the pull duration(pull+decompression) keep the same
image

- file size 
although the lz4 may increase the size of segment stored in hdfs, maybe 10% - 20%, but storage cost is acceptable

- task hand off 
task hand off was reduced because task takes fewer time to compress the data

image

- cpu resources 
    during the push phase of ingestion tasks, the lz4 does not take more cpu resource compared to zip
image

and for historicals, we didn't observe any cpu resources increase in a cluster where lz4 formatted segments are frequently downloaded

## New Configuration 
- druid.storage.compressionFormat is added, if not given, it defaults to zip to keep current behaviour

## Metrics

Added the following metrics:

- hdfs/pull/size
- hdfs/pull/duration
- hdfs/push/size
- hdfs/push/duration

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/docs/operations/metrics.md
+++ b/docs/operations/metrics.md
@@ -616,3 +616,14 @@ These metrics are available on operating systems with the cgroup kernel feature.
 |`cgroup/cpuset/effective_cpu_count`|Total number of active CPUs available to the process. Derived from `cpuset.effective_cpus`.||Varies|
 |`cgroup/cpuset/mems_count`|Total number of memory nodes available to the process. Derived from `cpuset.mems`.||Varies|
 |`cgroup/cpuset/effective_mems_count`|Total number of active memory nodes available to the process. Derived from `cpuset.effective_mems`.||Varies|
+
+## HDFS
+
+These metrics are available when the `druid-hdfs-storage` extension is used to push/pull segment files.
+
+| Metric               | Description                                                                   | Dimensions                                  | Normal value |
+|----------------------|-------------------------------------------------------------------------------|---------------------------------------------|--------------|
+| `hdfs/pull/size`     | Total bytes of decompressed segment files pulled from HDFS.                   | `format`: compression format (`ZIP`, `LZ4`) | Varies       |
+| `hdfs/pull/duration` | Time in milliseconds spent decompressing and pulling segment files from HDFS. | `format`: compression format (`ZIP`, `LZ4`) | Varies       |
+| `hdfs/push/size`     | Total bytes compressed segment files pushed to HDFS.                          | `format`: compression format (`ZIP`, `LZ4`) | Varies       |
+| `hdfs/push/duration` | Time in milliseconds spent compressing and pushing segment files to HDFS.     | `format`: compression format (`ZIP`, `LZ4`) | Varies       |


clintropolis:
quite a lot of the segment already uses lz4 by default so curious how effective running the whole thing through lz4 again would be, did you do any experiments to compare with just not 'externally' compressing at all in deep storage?

S3 support for honoring `druid.storage.zip` was added in #18544, and we've been using it in some of our clusters alongside the virtual storage functionality introduced in #18176, since not having to decompress the segments to load them speeds up things quite a lot. It trades some extra space, but has been worth it for our use case.

Also, the new 'V10' segment format introduced in #18880 was built around some future ideas i have about improving the virtual storage functionality to start only downloading the v10 metadata to enable partial downloads of only the parts of the segment that are needed to take part in a query, which will require the segment not be 'externally' compressed in deep storage (columns inside the segment can obviously still be compressed).

Fwiw, I'm not necessarily opposed to making this 'external' segment compression stuff more configurable (but it is certainly a bit tedious with the current interfaces since it needs to be handled by each implementation of segment pusher/puller separately).

FrankChen021:
@clintropolis thanks for the comments.

> quite a lot of the segment already uses lz4 by default so curious how effective running the whole thing through lz4 again would be, did you do any experiments to compare with just not 'externally' compressing at all in deep storage?

I think you're ask for 2 questions here.
the first is, after lz4 compression on columns, will the compression gain? here's a table showing the raw segments and the size under zip/lz4 compression again

image

You can see that decompressed segment has a size of 11GB, in the HDFS(by the current default zip), it has a size of 2.48GB

Another question I think you're asking is wheter we did experiments to upload the raw files into deep storage?
No. 
1st, above table shows the compression over the dir has gain, 
2nd, HDFS deep storage currently does not support upload file by file.

As for v10, I do know that we can support partial download, but HDFS as deep storage is different from object storage, 
for HDFS, it can't support too many files as object storage, and it can't provide high concurrent access as object storage, querying data directly from hdfs is much slower than querying data from object storage. For us, hdfs is main storage and I don't see there's way for us to migrate from hdfs to object storage, we will stay on hdfs for very long time.


> since it needs to be handled by each implementation of segment pusher/puller

No need to worry about it. the compression configuration is only provided on hdfs, and it's implemented in the hdfs, not on a higher level which requires all deep storage to do so.

clintropolis:
>You can see that decompressed segment has a size of 11GB, in the HDFS(by the current default zip), it has a size of 2.48GB

interesting, that is a lot bigger of a difference than i expected, though perhaps if there are a lot of complex columns that are not using compression (off by default added in #16863) then a difference of that size makes sense. Basically where I'm thking things are heading is moving away from generic compression in favor of all of the contents of the segment file being compressed.

>Another question I think you're asking is wheter we did experiments to upload the raw files into deep storage?

Ah, this was mostly just me wondering about uncompressed sizing, though i expect most of the perf stuff would look better not having to do any extra compression/decompression, for your segments at least it seems like some additional stuff would need to happen to make that viable.

>As for v10, I do know that we can support partial download, but HDFS as deep storage is different from object storage,
for HDFS, it can't support too many files as object storage, and it can't provide high concurrent access as object storage, querying data directly from hdfs is much slower than querying data from object storage. For us, hdfs is main storage and I don't see there's way for us to migrate from hdfs to object storage, we will stay on hdfs for very long time.

V10 segments store everything in a single file, `druid.segment`, so in terms of count it should be no different than having a single .zip or whatever that there is today with externally compressed v9 segments. Though it is fair that partial downloads would potentially increase concurrent access, however with smaller fetches so maybe not quite as bad. None of the partial stuff exists yet though, and virtual storage mode is optional.

processing/src/main/java/org/apache/druid/utils/CompressionUtils.java
```diff
@@ -212,6 +286,106 @@ public static long zip(File directory, OutputStream out) throws IOException
...
+  /**
+   * Decompresses LZ4-compressed directory archive
+   */
+  public static FileUtils.FileCopyResult lz4DecompressDirectory(InputStream in, File outDir) throws IOException
+  {
+    if (!(outDir.exists() && outDir.isDirectory())) {
+      throw new ISE("outDir[%s] must exist and be a directory", outDir);
+    }
+
+    final LZ4BlockInputStream lz4In = new LZ4BlockInputStream(in);
```
GWphua, line 364: Deprecated API, maybe try 

LZ4BlockInputStream.newBuilder().build(in);

processing/src/main/java/org/apache/druid/utils/CompressionUtils.java
```diff
@@ -622,6 +796,8 @@ public static InputStream decompress(final InputStream in, final String fileName
   {
     if (fileName.endsWith(Format.GZ.getSuffix())) {
       return gzipInputStream(in);
+    } else if (fileName.endsWith(Format.LZ4.getSuffix())) {
+      return new LZ4BlockInputStream(in);
```
GWphua, line 800: Same as above

extensions-core/hdfs-storage/src/main/java/org/apache/druid/storage/hdfs/HdfsDataSegmentPuller.java
```diff
@@ -235,27 +248,16 @@ FileUtils.FileCopyResult getSegmentFiles(final Path path, final File outDir) thr
...
-        );
+        final FileUtils.FileCopyResult result = format.decompressDirectory(getInputStream(path), outDir);
 
-        log.info(
-            "Unzipped %d bytes from [%s] to [%s]",
-            result.size(),
-            path.toString(),
-            outDir.getAbsolutePath()
-        );
+        emitMetrics(format, result.size(), System.currentTimeMillis() - startTime);
```
GWphua, line 260: Any chance that `emitMetrics` can also apply for Gz / Directory? Since we are emitting for `hdfs/pull/duration` + `hdfs/pull/size`. 

There is an expectation for metrics to be emitted for non-zip/lz4 too.

extensions-core/hdfs-storage/src/main/java/org/apache/druid/storage/hdfs/HdfsDataSegmentPuller.java
```diff
@@ -292,6 +294,17 @@ public InputStream openStream() throws IOException
...
 
+  private void emitMetrics(CompressionUtils.Format format, long size, long duration)
+  {
+    if (emitter == null) {
+      return;
+    }
+    ServiceMetricEvent.Builder metricBuilder = ServiceMetricEvent.builder();
+    metricBuilder.setDimension("format", format);
+    emitter.emit(metricBuilder.setMetric("hdfs/pull/size", size));
+    emitter.emit(metricBuilder.setMetric("hdfs/pull/duration", duration));
```
GWphua, line 305: Can add docs for these metrics.

extensions-core/hdfs-storage/src/main/java/org/apache/druid/storage/hdfs/HdfsDataSegmentPuller.java
```diff
@@ -235,27 +248,16 @@ FileUtils.FileCopyResult getSegmentFiles(final Path path, final File outDir) thr
         catch (Exception e) {
           throw new RuntimeException(e);
         }
-      } else if (CompressionUtils.isZip(path.getName())) {
+      }
 
-        // --------    zip     ---------
```
GWphua, line 240: Out of scope of this PR, but seems like it may be better to use a factory class instead to handle this logic... 

Maybe something to do in the future.

GWphua (commented):
Overall looks ok, can address the comments + some warnings from the `github-advanced-security`

processing/src/main/java/org/apache/druid/utils/CompressionUtils.java
FrankChen021: not true

extensions-core/hdfs-storage/src/main/java/org/apache/druid/storage/hdfs/HdfsDataSegmentPuller.java
FrankChen021: currently only zip is saved in hdfs, gz and dir should be old code path. no need to support it.

extensions-core/hdfs-storage/src/main/java/org/apache/druid/storage/hdfs/HdfsDataSegmentPuller.java
FrankChen021: sure

extensions-core/hdfs-storage/src/main/java/org/apache/druid/storage/hdfs/HdfsDataSegmentPuller.java
FrankChen021: it's better to do refactoring in a separated pr

processing/src/main/java/org/apache/druid/utils/CompressionUtils.java
GWphua: Seems like the deprecated API comes from lz4-java 1.10.1, which is not introduced into the repo yet.

GWphua (approved):
Changes LGTM

PR #18984, opened 2026-02-04
switch default front-coding format to v1, drop experimental from docs

clintropolis:
This PR bumps the default front-coded format to V1, which has a number of improvements over V0, but couldn't immediately be the default due to backwards compatibility concerns. V1 has been in Druid since 26 release, so I think it is plenty safe to switch the default at this point, however anyone upgrading from a version older than this and also using front-coding might want to take note of this and plan accordingly for upgrade to make downgrades not be tricky.

The docs are also updated to drop the 'experimental' wording around this feature, as it has been included and in use for quite some time. I've also removed some stuff in the release/upgrade notes about it being on by default, since it is not. We can add that info there at some future time, if that ever comes to pass where we make it the default.

With regards to making it the default, I think my preference would be to have an 'auto' mode where it takes statistics during indexing to determine if front-coding would be effective, and if so, automatically pick an ideal bucket size.

### DIFF
--- a/docs/release-info/migr-front-coded-dict.md
+++ b/docs/release-info/migr-front-coded-dict.md
@@ -23,9 +23,6 @@ sidebar_label: Front-coded dictionaries
   ~ under the License.
 -->
 
-:::info
-Front coding is an [experimental feature](../development/experimental.md) introduced in Druid 25.0.0.
-:::
 Apache Druid stores STRING columns using dictionary encoding for better compression, where each string value is added to a lexicographically sorted dictionary and the actual column just stores a pointer to a dictionary entry.
 Front coding is an optional incremental encoding strategy that lets you further compress STRING and [COMPLEX&lt;json&gt;](../querying/nested-columns.md) columns in Druid with minimal performance impact.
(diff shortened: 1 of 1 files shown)


docs/release-info/migr-front-coded-dict.md
```diff
@@ -27,12 +27,10 @@ sidebar_label: Front-coded dictionaries
 Front coding is an [experimental feature](../development/experimental.md) introduced in Druid 25.0.0.
 :::
```
capistrant:
```suggestion
```

capistrant (commented):
left suggestion to nuke another experimental reference

docs/release-info/migr-front-coded-dict.md
clintropolis: oops, missed that one, thanks

capistrant:
Added upgrade note label to remind ourselves to callout for folks who may be using v0 and upgrading from < druid 26

capistrant:
> Added upgrade note label to remind ourselves to callout for folks who may be using v0 and upgrading from < druid 26

which would be an insane jump, so I'd bet very unlikely.

ps sorry about the close open. clicking buttons to fast I guess

PR #18985, opened 2026-02-04
Fix resourceName used in auth check for SupervisorsTable in SystemSchema

maytasm:
### Description

SupervisorsTable in SystemSchema (powering queries to the sys.supervisors table) was using the topic/stream of the supervisorSpec as the resourceName to authenticate against. This is wrong and is inconsistent with other APIs that uses the datasource name of the supervisor (i.e. updating supervisor spec, suspending/starting, querying, etc). This change fixes this to use the datasourceName and add test to verify that only datasourceName is use as resourceName in authentication.  

This PR has:

- [x] been self-reviewed.


sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -1097,7 +1097,7 @@ private CloseableIterator<SupervisorStatus> getAuthorizedSupervisors(
       );
 
       Function<SupervisorStatus, Iterable<ResourceAction>> raGenerator = supervisor -> Collections.singletonList(
-          AuthorizationUtils.DATASOURCE_READ_RA_GENERATOR.apply(supervisor.getSource()));
+          AuthorizationUtils.DATASOURCE_READ_RA_GENERATOR.apply(supervisor.getDataSource()));
```
FrankChen021, line 1100: the java doc of orignial getSource says:
```
  /**
   * This API is only used for informational purposes in
   * org.apache.druid.sql.calcite.schema.SystemSchema.SupervisorsTable
   *
   * @return source like stream or topic name
   */
  String getSource();
```

Since here we' re using `getDataSource` I think we should remove the original getSource method definition. I check the code, besides the SystemSchema, there're few references in the the kafka supervisor spec, which can be replaced by calling `getIoConfig().getStream()`

maytasm: I don't think that's possible. getIoConfig is not a method of SupervisorSpec. It is only available in some (but not all) sub classes like KafkaSupervisorSpec but not CompactionSupervisorSpec. I think we still need this method

FrankChen021: at least we should update the javadoc

maytasm: Done

maytasm:
@jtuglu1 @FrankChen021 Please take a look when you can. Thanks!

jtuglu1 (approved):
LGTM

maytasm:
CI failure seems unrelated (but also failing a lot, ive retried many times)

kfaraz:
> CI failure seems unrelated (but also failing a lot, ive retried many times)

Yeah, something seems to have gotten broken in the docker-tests.
I am trying to investigate it in #19032

PR #18986, opened 2026-02-04
Update guice to 6.0.0

kgyrtkirk:
Running with jdk25 leads to guice errors suppressed by class version exceptions.

### DIFF
--- a/extensions-core/kubernetes-extensions/pom.xml
+++ b/extensions-core/kubernetes-extensions/pom.xml
@@ -100,10 +100,6 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>javax.inject</groupId>
-      <artifactId>javax.inject</artifactId>
-    </dependency>
-    <dependency>
       <groupId>com.google.inject</groupId>
       <artifactId>guice</artifactId>
       <scope>provided</scope>


FrankChen021:
This upgrade is smooth, no code change required unlike previous upgrade to 5.x

FrankChen021 (approved):

kgyrtkirk:
thank you @FrankChen021and @cryptoe for taking a look!

I wasn't expecting it to be this simple either :)
but I did choose to tradeoff something: `6.0` is a transition release - which enables to use the relocated jakarta api while still retains the javax one... 
upgrading fully to the jakarta api is not yet possible - it seemed like jersey needs to be upgraded first

capistrant:
> thank you @FrankChen021and @cryptoe for taking a look!
> 
> I wasn't expecting it to be this simple either :) but I did choose to tradeoff something: `6.0` is a transition release - which enables to use the relocated jakarta api while still retains the javax one... upgrading fully to the jakarta api is not yet possible - it seemed like jersey needs to be upgraded first

IIRC Jersey and Guice have stopped getting along well on newer versions so I'm interested to see how things go when we try to upgrade, if it will be straightforward to continue using both, a nightmare, or impossible

PR #18991, opened 2026-02-06
Get rid of dominant task duration in cost-based autoscaler

Fly-Style:
Also, this patch replaces idle decay with logarithmic lag acceleration, providing smoother and less complicating scaling logic under loaded lag conditions.

Example under the hood.

<details>

Plot provides information about the scale-up aggressiveness from 1 task, having a 500k total lag.
 
**p** = partition count
**rate** = processing rate, given from metrics
**idle** = `poll-avg-idle` metric

image

</details>

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
@@ -29,12 +29,7 @@ import org.apache.druid.java.util.common.logger.Logger;
 public class WeightedCostFunction
 {
   private static final Logger log = new Logger(WeightedCostFunction.class);
-  /**
-   * The lag severity at which lagBusyFactor reaches 1.0 (full idle suppression).
-   * lagSeverity is defined as lagPerPartition / highLagThreshold.
-   * At severity=1 (threshold), lagBusyFactor=0. At severity=MAX, lagBusyFactor=1.0.
-   */
-  private static final int LAG_AMPLIFICATION_MAX_SEVERITY = 5;
+  private static final double LAG_AMPLIFICATION_MULTIPLIER = 0.2;
 
   /**
    * Computes cost for a given task count using compute time metrics.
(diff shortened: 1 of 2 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -73,12 +68,19 @@ public CostResult computeCost(
...
+        final double lagPerPartition = metrics.getAggregateLag() / metrics.getPartitionCount();
+        final double amplification = Math.max(1.0, 1.0 + LAG_AMPLIFICATION_MULTIPLIER * Math.log(lagPerPartition));
+        lagRecoveryTime = metrics.getAggregateLag() * amplification / (proposedTaskCount * avgProcessingRate);
+      }
     }
 
-    final double predictedIdleRatio = estimateIdleRatio(metrics, proposedTaskCount, config.getHighLagThreshold());
-    final double idleCost = proposedTaskCount * metrics.getTaskDurationSeconds() * predictedIdleRatio;
+    final double predictedIdleRatio = estimateIdleRatio(metrics, proposedTaskCount);
+    final double idleCost = proposedTaskCount * predictedIdleRatio;
```
kfaraz, line 83: Getting rid of the time component here makes it difficult to compare `lagCost` and `idleCost` since they are not relatable entities anymore. The former deals with "time" whereas the latter is dimension-less (or maybe uses an implicit duration of 1 second).

I guess it would be more meaningful to use something like:
```suggestion
    final double idleCost = proposedTaskCount * lagRecoveryTime * predictedIdleRatio;
```

This way the `idleCost` would denote the seconds wasted being idle while lag was being recovered. I suppose this is more relatable than multiplying by task duration.

Could you please try the calculations with this update?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -73,12 +68,19 @@ public CostResult computeCost(
...
     } else {
       // Lag recovery time is decreasing by adding tasks and increasing by ejecting tasks.
+      // In case of increasing lag, we apply an amplification factor to reflect the urgency of addressing lag.
       // Caution: we rely only on the metrics, the real issues may be absolutely different, up to hardware failure.
-      lagRecoveryTime = metrics.getAggregateLag() / (proposedTaskCount * avgProcessingRate);
+      if (metrics.getAggregateLag() <= 0) {
+        lagRecoveryTime = 0;
+      } else {
+        final double lagPerPartition = metrics.getAggregateLag() / metrics.getPartitionCount();
+        final double amplification = Math.max(1.0, 1.0 + LAG_AMPLIFICATION_MULTIPLIER * Math.log(lagPerPartition));
```
kfaraz, line 77: Why do we need the amplication exactly?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -73,12 +68,19 @@ public CostResult computeCost(
...
+        lagRecoveryTime = metrics.getAggregateLag() * amplification / (proposedTaskCount * avgProcessingRate);
```
kfaraz, line 78: Maybe rename this field to `lagRecoveryTimeInSeconds` so that we are always able to think in terms of the relevant units.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: It is mentioned in the description:

>  this patch replaces idle decay with logarithmic lag acceleration, providing smoother and less complicating scaling logic under loaded lag conditions.

kfaraz: I see. Would it be possible to see how the autoscaler does without the amplification?
Could that be an option?

Fly-Style: >  Would it be possible to see how the autoscaler does without the amplification?

Both the math research and testing showed that it behaves poorly under several lag conditions. Generally, it scales, but not enough and then lag becames stale. 
For close-to-zero lag that amplifier almost not exists.

Fly-Style: Plot below the hood. `mult = 0` is a plot with removed lag amplification.

<details>

image

</details>

kfaraz (approved):
LGTM

kgyrtkirk:
thank you @Fly-Style  for the changes and for the review @kfaraz !

PR #18993, opened 2026-02-06
Use LinkedHashMap for EventMap to preserve ordering of Event dimensions

abhishekrb19:
Currently, when we use LoggingEmitter and/or LoggingRequestLogger, long, complex SQL queries make up the majority of the log volume, so we truncate logs to some extent. With these events, this can lead to the "value" part, timestamp among other getting truncated since the map's ordering is arbitrary. So this patch switches the EventMap to a LinkedHashMap so critical info (e.g., metric name and value) are logged first, followed by fields like service dimensions, user dimensions and the SQL query itself. This consistent order should also make it easy on the eyes.

Same idea for the request logger: we log the query context before the SQL query itself, which lets us preserve any useful/debuggable information deterministically even under truncation.

Given these maps typically have at most ~20–30 key/value pairs, the additional overhead of using a LinkedHashMap should be okay.

For example:

```
 org.apache.druid.java.util.emitter.core.LoggingEmitter - [metrics] {"feed":"metrics","timestamp":"2026-02-06T21:43:41.033Z","metric":"query/bytes","value":138,"service":"druid/historical","host":"localhost:8083","version":"37.0.0-SNAPSHOT","context":{"applyLimitPushDown":true,"defaultTimeout":180000,"engine":"native","finalize":false,"fudgeTimestamp":"1454795020784","groupByOutermost":false,"maxQueuedBytes":5242880,"maxScatterGatherBytes":9223372036854775807,"queryFailTime":1770414400967,"queryId":"379c91c7-40b1-4d40-b86f-340fcde5177b","queryResourceId":"c5f4256c-f84f-40c4-964a-840f1505d9a3","resultAsArray":true,"sliceIndex":23,"sqlOuterLimit":1001,"sqlQueryId":"yay","sqlStringifyArrays":false,"timeout":179998},"dataSource":"foo1","duration":"PT2592000S","hasFilters":"false","id":"379c91c7-40b1-4d40-b86f-340fcde5177b","interval":["2025-09-01T00:00:00.000Z/2025-10-01T00:00:00.000Z"],"numComplexMetrics":"0","numDimensions":"3","numMetrics":"0","remoteAddress":"127.0.0.1","statusCode":200,"success":"true","type":"groupBy"}
```

instead of 

```
org.apache.druid.java.util.emitter.core.LoggingEmitter - [metrics] {"server":"localhost:8083","numComplexMetrics":"0","type":"groupBy","version":"37.0.0-SNAPSHOT","duration":"PT2592000S","feed":"metrics","numDimensions":"3","metric":"query/node/ttfb","hasFilters":"false","service":"druid/broker","host":"localhost:8082","context":{"applyLimitPushDown":true,"defaultTimeout":180000,"engine":"native","finalize":false,"fudgeTimestamp":"1675477114564","groupByOutermost":false,"maxQueuedBytes":5242880,"maxScatterGatherBytes":9223372036854775807,"queryFailTime":1770171694634,"queryId":"51152d78-80c6-452b-ba1f-604c2edde74a","queryResourceId":"b5c87e25-8368-4c1c-b974-8fc6e272e54b","resultAsArray":true,"sliceIndex":23,"sqlOuterLimit":1001,"sqlQueryId":"foo","sqlStringifyArrays":false},"interval":["2025-09-01T00:00:00.000Z/2025-10-01T00:00:00.000Z"],"id":"51152d78-80c6-452b-ba1f-604c2edde74a","numMetrics":"0","value":17,"dataSource":"foo1","timestamp":"2026-02-06T02:18:34.654Z"}
```

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/test/java/org/apache/druid/java/util/emitter/service/ServiceMetricEventTest.java
+++ b/processing/src/test/java/org/apache/druid/java/util/emitter/service/ServiceMetricEventTest.java
@@ -366,13 +366,13 @@ public class ServiceMetricEventTest
   @Test
   public void testSerializedServiceMetricEventIsOrdered() throws JsonProcessingException
   {
-    ServiceMetricEvent event = ServiceMetricEvent.builder()
-                                                  .setCreatedTime(DateTimes.of("2026-01-01"))
-                                                 .setMetric("m1", 42)
-                                                 .setDimension("dim1", "xyz")
-                                                 .setFeed("test_feed")
-                                                 .setDimension("dim2", "xyz")
-                                                 .build("broker", "hostA");
+    final ServiceMetricEvent event = ServiceMetricEvent.builder()
+                                                       .setCreatedTime(DateTimes.of("2026-01-01"))
+                                                       .setMetric("m1", 42)
+                                                       .setDimension("dim1", "xyz")
+                                                       .setFeed("test_feed")
+                                                       .setDimension("dim2", "xyz")
+                                                       .build("broker", "hostA");
 
 
     Assert.assertEquals(
--- a/server/src/main/java/org/apache/druid/server/RequestLogLine.java
+++ b/server/src/main/java/org/apache/druid/server/RequestLogLine.java
@@ -93,7 +93,6 @@ public class RequestLogLine
 
   public String getSqlQueryLine(ObjectMapper objectMapper) throws JsonProcessingException
   {
-
     final Map<String, Object> queryMap = new LinkedHashMap<>();
     queryMap.put("context", sqlQueryContext);
     queryMap.put("query", sql == null ? "<unavailable>" : sql);
--- a/server/src/test/java/org/apache/druid/server/log/DefaultRequestLogEventTest.java
+++ b/server/src/test/java/org/apache/druid/server/log/DefaultRequestLogEventTest.java
@@ -165,14 +165,14 @@ public class DefaultRequestLogEventTest
     expected.put("remoteAddr", host);
     expected.put("queryStats", queryStats);
 
-    EventMap observedMap = defaultRequestLogEvent.toMap();
-    Assert.assertEquals(expected, observedMap);
+    final EventMap observedEventMap = defaultRequestLogEvent.toMap();
+    Assert.assertEquals(expected, observedEventMap);
     Assert.assertEquals(
         StringUtils.format(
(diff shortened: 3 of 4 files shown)


server/src/main/java/org/apache/druid/server/log/DefaultRequestLogEvent.java
```diff
@@ -63,16 +63,15 @@ public EventMap toMap()
...
         .put("service", getService())
         .put("host", getHost())
+        .put("remoteAddr", getRemoteAddr())
+        .put("queryStats", getQueryStats())
         .putNonNull("query", getQuery());
 
     if (getSql() != null) {
       builder.put("sql", getSql())
              .put("sqlQueryContext", getSqlQueryContext());
     }
```
aho135, line 73: Can we log the sqlQueryContext before the query itself here?

processing/src/main/java/org/apache/druid/java/util/emitter/core/EventMap.java
```diff
@@ -21,17 +21,18 @@
...
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 /**
- * EventMap is a hash map implementation. It can be safely serialzed to JSON using Jackson serializer as it
+ * EventMap is a linked hash map implementation where the insertion order of key-value pairs is maintained.
+ * It can be safely serialzed to JSON using Jackson serializer as it
```
aho135:
```suggestion
 * It can be safely serialized to JSON using Jackson serializer as it
```

server/src/main/java/org/apache/druid/server/log/DefaultRequestLogEvent.java
abhishekrb19: yes, good point will do that

processing/src/test/java/org/apache/druid/java/util/emitter/service/ServiceMetricEventTest.java
```diff
@@ -359,4 +362,22 @@ public void test_builder_throwsException_ifDimNameOrValueIsNull()
         () -> eventBuilder.setDimension(null, new String[]{"a"})
     );
   }
+
+  @Test
+  public void testSerializedServiceMetricEventIsOrdered() throws JsonProcessingException
+  {
+    ServiceMetricEvent event = ServiceMetricEvent.builder()
+                                                  .setCreatedTime(DateTimes.of("2026-01-01"))
```
aho135:
```suggestion
                                                 .setCreatedTime(DateTimes.of("2026-01-01"))
```

server/src/main/java/org/apache/druid/server/RequestLogLine.java
aho135, line 175: Do you think it's worth re-ordering toString to match the new ordering?

abhishekrb19: I don’t think so. The `toString()` and `hashCode()` are typically IDE-generated based on the order of the arguments defined in the constructor. It’s probably only used for debugging and likely predates the `getSqlQueryLine()` and `getNativeQueryLine()` implementations

abhishekrb19:
Thanks for the review, @aho135

PR #18996, opened 2026-02-08
add incremental compaction support

cecemei:
### Description

  This PR builds on #18968 to add incremental compaction support, allowing compaction to operate on only uncompacted segments within an interval while upgrading already-compacted segments to maintain consistency.

 ### Key Changes

  1. Incremental Compaction Mode

  Extended CompactionMode enum with INCREMENTAL_COMPACTION:
  - FULL_COMPACTION: Compacts all segments (existing behavior)
  - INCREMENTAL_COMPACTION: Compacts only uncompacted segments, upgrades compacted segments
  - NOT_APPLICABLE: Skip compaction

  2. SegmentUpgradeAction

  New task action that upgrades segment metadata without rewriting data:
  - Updates partition numbers and shard specs of already-compacted segments
  - Ensures compatibility between old and newly compacted segments

  3. ShardSpec Compatibility

  Extended ShardSpec interface with mutation methods to support incremental compaction:
  - withPartitionNum(int): Update partition number
  - withCorePartitions(int): Update core partition count
  - isNumChunkSupported(): Check numbered partition chunk support

  4. CompactionIntervalSpec Enhancements

  Added optional uncompactedSegments field:
  - When null: Compact all segments in interval (full compaction)
  - When non-null: Compact only specified segments (incremental compaction)

  5. Policy-Level Control

  Added incrementalCompactionUncompactedRatioThreshold to MostFragmentedIntervalFirstPolicy:
  - When uncompactedBytes / compactedBytes < threshold, use incremental compaction
  - When >= threshold, use full compaction
  - Default: 0.0 (always full compaction, backward compatible)
  - Example: 0.5 uses incremental when uncompacted data < 50% of compacted data

  6. Builder Pattern for UserCompactionTaskQueryTuningConfig

  Replaced verbose 19-parameter constructors with builder pattern for better readability.

###  Release Note

  Added incremental compaction mode that compacts only newly ingested segments while preserving already-compacted data. Configure using incrementalCompactionUncompactedRatioThreshold in MostFragmentedIntervalFirstPolicy to control when incremental compaction is used based on the ratio of uncompacted to compacted data. Only supported by compaction supervisors on Overlord with MSQ engine.

 #### Key changed/added classes

  - CompactionMode - Added INCREMENTAL_COMPACTION
  - SegmentUpgradeAction - New task action for metadata upgrades
  - ShardSpec - Added mutation methods
  - CompactionIntervalSpec - Added uncompactedSegments field
  - MostFragmentedIntervalFirstPolicy - Added threshold configuration
  - UserCompactionTaskQueryTuningConfig - Added builder pattern

<hr>

This PR has:

- [ ] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentUpgradeAction.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.druid.indexing.common.task.Task;
+import org.apache.druid.java.util.common.IAE;
+import org.apache.druid.metadata.ReplaceTaskLock;
+import org.apache.druid.timeline.DataSegment;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class SegmentUpgradeAction implements TaskAction<Integer>
```
capistrant: javadoc seems in order here

gianm:
Wanted to point out that #19016 is aiming at similar goals but is taking a different approach. The big one is that #19016 only works with non-MSQ compaction and this one only works with MSQ compaction. Longer comment on the other PR here: https://github.com/apache/druid/pull/19016#issuecomment-3892042166

FrankChen021:
@gianm is there a plan that the msq based compaction replaces the old one? otherwise we have two different mechanisms, which not only confuses users but also increases the effort to maintain the code base.

gianm:
> @gianm is there a plan that the msq based compaction replaces the old one? otherwise we have two different mechanisms, which not only confuses users but also increases the effort to maintain the code base.

For us, internally, it already has. We have migrated our internal use cases to MSQ compaction. I think it would be nice if at some point, only MSQ compaction was supported. But, I never like removing things that people are using without a long warning / deprecation period, so I think it would be some time before we considered removing native compaction.

cecemei:
mark this PR as draft since it has some refactor changes, please see #19059 instead.

PR #19000, opened 2026-02-09
bump bytebuddy dep flagged by upper bound check

capistrant:
### Description

dependency bump. Seems to have been flagged by new strict dep checks in  #18990

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
         <scala.library.version>2.13.14</scala.library.version>
         <avatica.version>1.27.0</avatica.version>
         <avro.version>1.11.4</avro.version>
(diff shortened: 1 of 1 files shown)


abhishekrb19 (approved):
Thanks! Is this a dev blocker or are only the master builds broken?

capistrant:
> Thanks! Is this a dev blocker or are only the master builds broken?

seems like a dev blocker if by that you mean all feature branches seem like they are going to have to pull this in to start passing CI :( I noticed my other PR failing CI cuz of this bytebuddy thing earlier, and now after retrying failed CI jobs on that PR (without merging master in) it still fails

PR #15585, opened 2023-12-15
expression virtual column indexes

clintropolis:
### Description
This PR adds support for using underlying column indexes for `ExpressionVirtualColumn` which have a single input column, similar to what `ExpressionFilter` can do in the same scenario. It achieves this a bit differently however. `ExpressionFilter` can make the filter expression itself into a predicate and push it down as a `DruidPredicateFactory` to the underlying `DruidPredicateIndexes`, if the expression evaluates to true then the value matches, if not then it doesn't. `ExpressionVirtualColumn` however is making arbitrary values which can be matched by _other_ filters such as equality or ranges, and so instead needs a mechanism to supply its own indexes. So, we expose an implementation of `DruidPredicateIndexes` that can accept any `DruidPredicateFactory`, and then use the `DictionaryEncodedValueIndex` which exposes low level access to the dictionary and value indexes of the underlying column so that we can scan the values of the dictionary to use as inputs to the expression and then feed the transformed values to the appropriate type of predicate from the `DruidPredicateFactory` based on the output type of the expression. 

I decided to push this all the way down to `Expr`, using the predicate index implementation as the default, but also allow individual expressions to produce more optimized index suppliers. I have currently only wired this up to `IdentifierExpr`, which can simply delegate directly to the underlying `ColumnIndexSupplier` because it is a direct column access expression. This means that an expression virtual column which just has an identifier should now be able to produce equivalent performance during filtering as if the column was used directly.

This measures pretty well, using a regex expression as an example, which in the first example uses no indexes and the second uses the new expression predicate indexes showed a decent performance boost (the virtual column performance should be approximately the same to using an expression filter or extractionFn directly, which also use predicate indexes).

```
SELECT string4, COUNT(*) FROM foo WHERE REGEXP_EXTRACT(string1, '^1') IS NOT NULL OR REGEXP_EXTRACT('Z' || string2, '^Z2') IS NOT NULL GROUP BY 1
```

before:
```
Benchmark                        (query)  (rowsPerSegment)  (schema)  (vectorize)  Mode  Cnt    Score    Error  Units
SqlExpressionBenchmark.querySql       40           5000000  explicit        false  avgt    5  279.224 ±  3.353  ms/op
SqlExpressionBenchmark.querySql       40           5000000      auto        false  avgt    5  274.911 ±  5.642  ms/op
```

after:
```
Benchmark                        (query)  (rowsPerSegment)  (schema)  (vectorize)  Mode  Cnt    Score   Error  Units
[...]

PR #15838, opened 2024-02-06
adaptive filter partitioning

clintropolis:
### Description
This PR reworks how filter partitioning is performed, pushing it into `Filter` to "bundle" both index and value matcher creation, allowing filter partitioning to be done adaptively based on the selectivity of prior filters. This was alluded to during discussions in #12388, and contains and enhances the functionality added in #13977 (as well as removes the need to merge #15551), and aims to solve the problems originally described in #3878.

The main change centers around a new addition to the `Filter` interface,

```
default <T> FilterBundle makeFilterBundle(
      ColumnIndexSelector columnIndexSelector,
      BitmapResultFactory<T> bitmapResultFactory,
      int selectionRowCount,
      int totalRowCount,
      boolean includeUnknown,
      boolean allowPartialIndex
  )
```

for which a default implementation is provided that should be sufficient in most cases, and produces a new container type `FilterBundle`, 

Production of the `FilterBundle` is assisted by a new method on `BitmapColumnIndex` 

```
  @Nullable
  default <T> T computeBitmapResult(
      BitmapResultFactory<T> bitmapResultFactory,
      int selectionRowCount,
      int totalRowCount,
      boolean includeUnknown
  )
```
which allows index computation to factor `selectionRowCount` and `totalRowCount` into whether or not they should compute indexes, gauged based on the amount of work they need to do vs the amount of work it would take to just use a `ValueMatcher`/`VectorValueMatcher` and scan the 'remaining' rows. 

The primary beneficiary of this is `AND` and `OR` filters, which can now partition not just on whether or not the sub-filter supports indexes at all, but also based on whether or not they _should_ use those indexes based on the selectivity of sub-filters whose indexes have already been computed, which itself is delegated to the sub-filters via the new `BitmapColumnIndex` method.

This can have a pretty dramatic performance impact:

query:
```
"SELECT string2, SUM(long1) FROM foo WHERE string1 = '1000' AND string5 LIKE '%1%' AND (string3 in ('1', '10', '20', '22', '32') AND long2 IN (1, 19, 21, 23, 25, 26, 46) AND double3 < 1010.0 AND double3 > 1000.0 AND (string4 = '1' OR REGEXP_EXTRACT(string1, '^1') IS NOT NULL OR REGEXP_EXTRACT('Z' || string2, '^Z2') IS NOT NULL)) GROUP BY 1 ORDER BY 2"
```

before:
```
[...]

PR #19004, opened 2026-02-10
Configurable index disabling for virtual columns

nozjkoitop:
### Description

In some of our environments we observed a significant query performance regression after upgrading to Druid 29.0. This regression appears to be related to the virtual column bitmap indexing optimization introduced in https://github.com/apache/druid/pull/15585 and https://github.com/apache/druid/pull/15633.

While the change improves performance for many cases by enabling bitmap index creation for expression-based virtual columns, we found that in certain workloads the index computation becomes unexpectedly expensive. In our deployments, building these indices accounted for more than 90% of the CPU usage on Historical nodes during query execution, leading to severe degradation in overall query latency.

For one representative workload, the average query runtime increased from 10.01 seconds to 150.18 seconds compared to earlier versions. Once expression-based virtual column bitmap index creation was disabled, the regression disappeared and query performance returned to expected levels.

This PR addresses this issue by preventing the optimization from triggering in scenarios where the cost of index computation outweighs its benefit, avoiding major regressions in affected environments.

Added a new query context parameter, `maxVirtualColumnsForBitmapIndexing` (default: `Integer.MAX_VALUE`), which sets the virtual-column count threshold beyond which Druid stops using bitmap indexes for filters on virtual columns.

#### Release note

Added a safeguard to skip virtual-column bitmap indexing when it is likely to be counterproductive, falling back to non-indexed filtering to preserve expected performance. 

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/QueryableIndexCursorHolder.java
+++ b/processing/src/main/java/org/apache/druid/segment/QueryableIndexCursorHolder.java
@@ -84,6 +84,7 @@ public class QueryableIndexCursorHolder implements CursorHolder
   private final QueryContext queryContext;
   private final int vectorSize;
   private final Supplier<CursorResources> resourcesSupplier;
+  private final boolean disableVirtualColumnBitmapIndexes;
 
   public QueryableIndexCursorHolder(
       QueryableIndex index,
@@ -108,6 +109,8 @@ public class QueryableIndexCursorHolder implements CursorHolder
     this.queryContext = cursorBuildSpec.getQueryContext();
     this.vectorSize = cursorBuildSpec.getQueryContext().getVectorSize();
     this.metrics = cursorBuildSpec.getQueryMetrics();
+    this.disableVirtualColumnBitmapIndexes = queryContext.getMaxVirtualColumnsForBitmapIndexing()
+                                             < virtualColumns.getColumnNames().size();
     this.resourcesSupplier = Suppliers.memoize(
         () -> new CursorResources(
             index,
@@ -117,7 +120,8 @@ public class QueryableIndexCursorHolder implements CursorHolder
             interval,
             filter,
             cursorBuildSpec.getQueryContext().getBoolean(QueryContexts.CURSOR_AUTO_ARRANGE_FILTERS, true),
-            metrics
+            metrics,
+            disableVirtualColumnBitmapIndexes
         )
     );
   }
@@ -672,7 +676,8 @@ public class QueryableIndexCursorHolder implements CursorHolder
         Interval interval,
         @Nullable Filter filter,
         boolean cursorAutoArrangeFilters,
-        @Nullable QueryMetrics<? extends Query<?>> metrics
+        @Nullable QueryMetrics<? extends Query<?>> metrics,
+        boolean disableVirtualColumnBitmapIndexes
     )
     {
       this.closer = Closer.create();
@@ -680,6 +685,10 @@ public class QueryableIndexCursorHolder implements CursorHolder
       this.timeBoundaryInspector = timeBoundaryInspector;
       try {
         this.numRows = index.getNumRows();
+        final ColumnIndexSelector indexSelectorForFilters =
+            disableVirtualColumnBitmapIndexes
+            ? new VirtualColumnBitmapDisablingIndexSelector(columnCache, virtualColumns)
+            : columnCache;
         this.filterBundle = makeFilterBundle(
             computeFilterWithIntervalIfNeeded(
                 timeBoundaryInspector,
@@ -688,7 +697,7 @@ public class QueryableIndexCursorHolder implements CursorHolder
                 filter
             ),
             cursorAutoArrangeFilters,
-            columnCache,
+            indexSelectorForFilters,
             numRows,
             metrics
         );
--- a/benchmarks/src/test/java/org/apache/druid/benchmark/query/SqlExpressionBenchmark.java
+++ b/benchmarks/src/test/java/org/apache/druid/benchmark/query/SqlExpressionBenchmark.java
@@ -159,7 +159,9 @@ public class SqlExpressionBenchmark extends SqlBaseQueryBenchmark
       "SELECT CASE WHEN MOD(long1, 2) = 0 THEN -1 WHEN MOD(long1, 2) = 1 THEN long2 / MOD(long1, 2) ELSE long3 END FROM expressions GROUP BY 1",
       // cast
       "SELECT CAST(string1 as BIGINT) + CAST(string3 as DOUBLE) + long3, COUNT(*) FROM expressions GROUP BY 1 ORDER BY 2",
-      "SELECT COUNT(*), SUM(CAST(string1 as BIGINT) + CAST(string3 as BIGINT)) FROM expressions WHERE double3 < 1010.0 AND double3 > 100.0"
+      "SELECT COUNT(*), SUM(CAST(string1 as BIGINT) + CAST(string3 as BIGINT)) FROM expressions WHERE double3 < 1010.0 AND double3 > 100.0",
+      // virtual-column bitmap-index
+      "SELECT COUNT(*) FROM expressions WHERE __time >= TIMESTAMP '2000-01-01 00:00:00' AND __time < TIMESTAMP '2000-01-02 00:00:00' AND (UPPER(COALESCE(string3,'')) LIKE '1%' OR TRIM(UPPER(COALESCE(string3,''))) LIKE '1%' OR SUBSTRING(UPPER(COALESCE(string3,'')),1,1) IN ('1','2','3','4','5') OR ('X' || UPPER(COALESCE(string3,''))) LIKE 'X1%') AND (UPPER(COALESCE(string5,'')) LIKE '2%' OR TRIM(UPPER(COALESCE(string5,''))) LIKE '2%' OR SUBSTRING(UPPER(COALESCE(string5,'')),1,1) IN ('1','2','3','4','5') OR ('Y' || UPPER(COALESCE(string5,''))) LIKE 'Y2%') AND CAST(double4 * 1000 AS BIGINT) BETWEEN -850000000 AND 850000000"
   );
 
   @Param({
@@ -233,9 +235,16 @@ public class SqlExpressionBenchmark extends SqlBaseQueryBenchmark
       "57",
       "58",
       "59",
-      "60"
+      "60",
+      "61"
   })
   private String query;
+
+  @Param({
+      "0",
+      "100"
+  })
+  private String maxVirtualColumnsForBitmapIndexing;
 
   @Override
   public String getQuery()
@@ -255,7 +264,8 @@ public class SqlExpressionBenchmark extends SqlBaseQueryBenchmark
     final Map<String, Object> context = ImmutableMap.of(
         QueryContexts.VECTORIZE_KEY, vectorize,
         QueryContexts.VECTORIZE_VIRTUAL_COLUMNS_KEY, vectorize,
-        GroupByQueryConfig.CTX_KEY_DEFER_EXPRESSION_DIMENSIONS, deferExpressionDimensions
+        GroupByQueryConfig.CTX_KEY_DEFER_EXPRESSION_DIMENSIONS, deferExpressionDimensions,
+        QueryContexts.CTX_MAX_VIRTUAL_COLUMNS_FOR_BITMAP, maxVirtualColumnsForBitmapIndexing
     );
     return context;
   }
--- a/docs/querying/query-context-reference.md
+++ b/docs/querying/query-context-reference.md
@@ -72,6 +72,7 @@ Unless otherwise noted, the following parameters apply to all query types, and t
 |`sqlPlannerBloat`|`1000`|Calcite parameter which controls whether to merge two Project operators when inlining expressions causes complexity to increase. Implemented as a workaround to exception `There are not enough rules to produce a node with desired properties: convention=DRUID, sort=[]` thrown after rejecting the merge of two projects.|
 |`cloneQueryMode`|`excludeClones`| Indicates whether clone Historicals should be queried by brokers. Clone servers are created by the `cloneServers` Coordinator dynamic configuration. Possible values are `excludeClones`, `includeClones` and `preferClones`. `excludeClones` means that clone Historicals are not queried by the broker. `preferClones` indicates that when given a choice between the clone Historical and the original Historical which is being cloned, the broker chooses the clones. Historicals which are not involved in the cloning process will still be queried. `includeClones` means that broker queries any Historical without regarding clone status. This parameter only affects native queries. MSQ does not query Historicals directly.|
 |`realtimeSegmentsOnly` |`false`| When set to true, only query realtime segments. Historical segments are excluded. |
+|`maxVirtualColumnsForBitmapIndexing`|`Integer.MAX_VALUE`| Sets the virtual-column count threshold beyond which Druid stops using bitmap indexes for filters on virtual columns.|
 
 ## Parameters by query type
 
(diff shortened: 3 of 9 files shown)


clintropolis:
>In some of our environments we observed a significant query performance regression after upgrading to Druid 29.0. This regression appears to be related to the virtual column bitmap indexing optimization introduced in https://github.com/apache/druid/pull/15585 and https://github.com/apache/druid/pull/15633.

have you tried in versions newer than 29? #15838 was in 30, and #17055/#17125 in 31, which should make it so that expression indexes are only used in cases where there would be less work to do than a full scan (which has to compute the expression value and perform the match for every row), which i think should reduce problems like you are seeing. 

I would be very interested if there are still cases where expression indexes cause a slowdown after those changes, since it would indicate that perhaps the cost estimate needs adjustment.

nozjkoitop:
Thanks for the references, unfortunately this still reproduces for us on 31. 
To put numbers behind this, i can try adding a benchmark that shows the impact

clintropolis:
> Thanks for the references, unfortunately this still reproduces for us on 31.
To put numbers behind this, i can try adding a benchmark that shows the impact

that would be great, there is [SqlExpressionBenchmark](https://github.com/apache/druid/blob/master/benchmarks/src/test/java/org/apache/druid/benchmark/query/SqlExpressionBenchmark.java) which uses a [data generator to have a collection of columns with various types and data distributions](https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/generator/GeneratorBasicSchemas.java#L331) maybe you can add something that looks approximately like the problem you're running into.

I would like to try to see if this is something we can improve without adding a new manual parameter

nozjkoitop:
added a case with broad, expression-heavy virtual-column filters
there were likely other improvements in later versions, but I still saw ~12% better performance with indexing disabled

clintropolis:
hmm, I still show the added benchmark query as faster with indexes, especially for vectorized processing. I didn't pull your branch, just used apache master with the query added and to test no indexes modified [this line ExpressionVirtualColumn](https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java#L274) to return `NoIndexesColumnIndexSupplier.getInstance()`


current master:
```
Benchmark                        (complexCompression)  (deferExpressionDimensions)  (jsonObjectStorageEncoding)  (query)  (rowsPerSegment)  (schemaType)  (storageType)  (stringEncoding)  (vectorize)  Mode  Cnt    Score    Error  Units
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP              UTF8        false  avgt    5  437.811 ± 57.658  ms/op
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP              UTF8        force  avgt    5  371.432 ±  2.740  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP              UTF8        false  avgt    5  417.081 ± 10.159  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP              UTF8        force  avgt    5  368.717 ±  8.266  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP              UTF8        false  avgt    5  417.167 ±  4.085  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP              UTF8        force  avgt    5  369.220 ± 10.618  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP              UTF8        false  avgt    5  417.864 ± 13.421  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP              UTF8        force  avgt    5  344.602 ±  9.518  ms/op
```

modified so that no indexes are used:
```
Benchmark                        (complexCompression)  (deferExpressionDimensions)  (jsonObjectStorageEncoding)  (query)  (rowsPerSegment)  (schemaType)  (storageType)  (stringEncoding)  (vectorize)  Mode  Cnt     Score    Error  Units
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP              UTF8        false  avgt    5   505.174 ±  7.250  ms/op
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP              UTF8        force  avgt    5  1165.574 ± 37.120  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP              UTF8        false  avgt    5   501.502 ± 33.384  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP              UTF8        force  avgt    5  1160.577 ± 43.142  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP              UTF8        false  avgt    5   501.388 ± 24.289  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP              UTF8        force  avgt    5  1198.882 ± 57.135  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP              UTF8        false  avgt    5   508.851 ±  4.995  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP              UTF8        force  avgt    5  1206.142 ± 37.526  ms/op
```

nozjkoitop:
Interesting, although I was using lz4 compression + FRONT_CODED_16_V1 encoding, ur results surprised me :) lemme give it another try

Could u share the JMH flags used please?

clintropolis:
```
DRUID_BENCHMARK_CACHE_DIR=./tmp java --add-exports=java.base/jdk.internal.misc=ALL-UNNAMED --add-exports=java.base/jdk.internal.ref=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED --add-opens=java.base/sun.nio.ch=ALL-UNNAMED --add-opens=java.base/jdk.internal.ref=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=jdk.management/com.sun.management.internal=ALL-UNNAMED -server -jar benchmarks/target/benchmarks.jar -p query=61 -p stringEncoding=UTF8 -p complexCompression=NONE -p schemaType=explicit -p storageType=MMAP org.apache.druid.benchmark.query.SqlExpressionBenchmark
```

I'm using java 21 on a m1 mac for additional context.

>Interesting, although I was using lz4 compression + FRONT_CODED_16_V1 encoding

i think by default everything here would be using lz4 since that parameter is only for complex columns to measure #16863 and has no impact if not using complex columns, but I will try with front coding though, since significantly slower perf there would also be quite interesting to look into

nozjkoitop:
thanks, i'll return it with matching params and let u know about results

clintropolis:
a neat, I do show front-coded as a bit slower for non-vectorized, which is curious, will look into that a bit more

with index:
```
Benchmark                        (complexCompression)  (deferExpressionDimensions)  (jsonObjectStorageEncoding)  (query)  (rowsPerSegment)  (schemaType)  (storageType)   (stringEncoding)  (vectorize)  Mode  Cnt    Score    Error  Units
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5  522.387 ± 22.942  ms/op
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  501.122 ± 17.559  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5  547.506 ± 15.055  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  446.650 ±  5.308  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5  572.099 ± 67.823  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  499.534 ± 19.926  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5  549.607 ± 25.846  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  496.660 ± 16.439  ms/op
```

without:
```
Benchmark                        (complexCompression)  (deferExpressionDimensions)  (jsonObjectStorageEncoding)  (query)  (rowsPerSegment)  (schemaType)  (storageType)   (stringEncoding)  (vectorize)  Mode  Cnt     Score    Error  Units
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5   516.153 ± 18.747  ms/op
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  1154.914 ± 34.838  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5   438.316 ± 24.513  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  1204.622 ± 57.132  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5   431.339 ± 23.974  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  1199.698 ± 34.916  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5   518.188 ± 19.511  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  1235.943 ± 38.730  ms/op
```

clintropolis:
re front-coding, i think i see what is going on, [the `ExpressionPredicateIndexSupplier` that is computing the indexes](https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/math/expr/ExpressionPredicateIndexSupplier.java#L262) is scanning the dictionary in order to find all of the values that match, but is using random access to get the values which for front coding means seeking in the bucket to the position we actually need which is basically where all of the cost of using it is (in exchange for the smaller sizes). I suspect when it is not using the indexes that perhaps the matcher must be able to rule out a match earlier and so fewer overall calls to [`FrontCodedIndexed.get`](https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/data/FrontCodedIndexed.java#L208) are happening.

I think since we are checking every dictionary value, it would be a lot more chill for front-coding if use used the dictionary iterator instead of calling get, it needs to be exposed on `DictionaryEncodedValueIndex` so that the expression predicate indexes could use it. Will look into this :+1:

clintropolis:
ah yea, it is totally that, using the iterator improves the measurement on using the indexes quite a lot

with indexes using iterator:
```
Benchmark                        (complexCompression)  (deferExpressionDimensions)  (jsonObjectStorageEncoding)  (query)  (rowsPerSegment)  (schemaType)  (storageType)   (stringEncoding)  (vectorize)  Mode  Cnt    Score     Error  Units
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5  428.333 ±  14.320  ms/op
SqlExpressionBenchmark.querySql                  NONE                 singleString                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  364.073 ±   5.671  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5  423.951 ±  12.710  ms/op
SqlExpressionBenchmark.querySql                  NONE                   fixedWidth                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  371.926 ±   5.133  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5  424.357 ±  10.445  ms/op
SqlExpressionBenchmark.querySql                  NONE         fixedWidthNonNumeric                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  419.708 ±  71.678  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        false  avgt    5  444.724 ± 112.962  ms/op
SqlExpressionBenchmark.querySql                  NONE                       always                        SMILE       61           1500000      explicit           MMAP  FRONT_CODED_16_V1        force  avgt    5  373.843 ±   8.409  ms/op
```

thanks for bringing this to attention :+1:

clintropolis:
opened #19023 to help perf for front-coding + expression indexes (and maybe a few other things)

nozjkoitop:
appreciate the input, that's a clean fix, nice

processing/src/main/java/org/apache/druid/segment/QueryableIndexCursorHolder.java
```diff
@@ -672,14 +676,19 @@ private CursorResources(
...
+        @Nullable QueryMetrics<? extends Query<?>> metrics,
+        boolean disableVirtualColumnBitmapIndexes
     )
     {
       this.closer = Closer.create();
       this.columnCache = new ColumnCache(index, virtualColumns, closer);
       this.timeBoundaryInspector = timeBoundaryInspector;
       try {
         this.numRows = index.getNumRows();
+        final ColumnIndexSelector indexSelectorForFilters =
```
FrankChen021, line 688: [P2] Search queries still bypass this virtual-column index guard

This only wraps the selector used by cursor filter bundles, but `SearchQuery` has a separate index-only path in `UseIndexesStrategy` that still constructs plain `ColumnCache` selectors for `partitionDimensionList`, `makeTimeFilteredBitmap`, and `IndexOnlyExecutor`. A search query with `maxVirtualColumnsForBitmapIndexing` set low and a virtual search dimension or virtual-column filter can still choose and execute the bitmap-index plan, so the new query context is not honored for one of the query types that accepts virtual columns. Please apply the same disabling selector in the search index strategy, or document/limit the parameter to cursor-backed query paths.

gianm:
Is this PR still being pursued now that https://github.com/apache/druid/pull/19023 is in?

nozjkoitop:
yeah, there is no need in this one anymore, closing it now

PR #19005, opened 2026-02-10
[grpc-query] Respect client cancellations/disconnections

bsmithgall:
### Description

Add support for in-flight Druid query cancellation after a gRPC client disconnects or cancels. Previously, these events were ignored, leading to queries continuing to run. Now, these cancellations forward that signal into Druid by:

1. Registering a [`Context.CancellationListener`](https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.CancellationListener.html)
2. Invoking `DirectStatement.cancel()` for SQL queries and `QueryScheduler.cancelQuery()` for native queries
3. Mapping `QueryInterruptedException` to appropriate gRPC status codes

This change uses an `AtomicReference` to handle a race condition where the client cancels after the listener is registered but before the callback is set. We use `getAndSet()` to ensure cancellation is fired at most once.

#### Release note

The gRPC query extension now cancels in-flight queries when clients cancel/disconnect.

<hr>

 * `QueryService`
 * `QueryDriver`
 * `GrpcEndpointInitializer`
 * `QueryServiceTest`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-contrib/grpc-query/src/main/java/org/apache/druid/grpc/server/QueryDriver.java
+++ b/extensions-contrib/grpc-query/src/main/java/org/apache/druid/grpc/server/QueryDriver.java
@@ -175,6 +175,7 @@ public class QueryDriver
 
     final org.apache.druid.server.QueryResponse queryResponse;
     final String currThreadName = Thread.currentThread().getName();
+    Throwable caught = null;
     try {
       queryLifecycle.initialize(query);
       AuthorizationResult authorizationResult = queryLifecycle.authorize(authResult);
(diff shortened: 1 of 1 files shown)


extensions-contrib/grpc-query/src/main/java/org/apache/druid/grpc/server/QueryDriver.java
```diff
@@ -163,6 +190,7 @@ private QueryResponse runNativeQuery(QueryRequest request, AuthenticationResult
 
       Thread.currentThread().setName(StringUtils.format("grpc-native[%s]", query.getId()));
       final ByteString results = encodeNativeResults(request, sequence, rowSignature);
+      // TODO: should we call queryLifecycle.emitLogsAndMetrics() here?
```
gianm: Please don't commit TODO comments. In general `queryLifecycle.emitLogsAndMetrics()` should be called any time a query is done, regardless of whether it completes successfully, with an error, or with a cancellation.

bsmithgall: Oops sorry I thought I had flagged this in the PR description but just realized I didn't; the existing behavior doesn't call this, so I wasn't sure if that was intentional. It seems like it was an oversight so I will fix.

PR #19009, opened 2026-02-11
s3 killer: paginate to find complete list of files to delete

abannon:
### Description

For segments stored in s3, the segment killer looks for files to remove within the segment prefix by calling listObjectsV2, which returns at most 1000 items. In the case where there are more than 1000 files per segment prefix, the segment killer did not find and delete all of the files.  The fix is to add pagination by reusing existing 'deleteSegmentFilesFromS3' function.

#### Release note

Fixed: Kill task for s3 will remove all segment files, instead of up to 1000 files.

<hr>

 * Changed S3DataSegmentKiller

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
+++ b/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
@@ -21,10 +21,7 @@ package org.apache.druid.storage.s3;
 
 import com.amazonaws.AmazonServiceException;
 import com.amazonaws.services.s3.model.DeleteObjectsRequest;
-import com.amazonaws.services.s3.model.ListObjectsV2Request;
-import com.amazonaws.services.s3.model.ListObjectsV2Result;
 import com.amazonaws.services.s3.model.MultiObjectDeleteException;
-import com.amazonaws.services.s3.model.S3ObjectSummary;
 import com.google.common.base.Predicates;
 import com.google.common.base.Supplier;
 import com.google.common.collect.Lists;
@@ -95,6 +92,8 @@ public class S3DataSegmentKiller implements DataSegmentKiller
 
     // create a map of bucket to keys to delete
     Map<String, List<DeleteObjectsRequest.KeyVersion>> bucketToKeysToDelete = new HashMap<>();
+    boolean shouldThrowException = false;
+
     for (DataSegment segment : segments) {
       String s3Bucket = MapUtils.getString(segment.getLoadSpec(), S3DataSegmentPuller.BUCKET);
       String path = MapUtils.getString(segment.getLoadSpec(), S3DataSegmentPuller.KEY);
@@ -103,16 +102,17 @@ public class S3DataSegmentKiller implements DataSegmentKiller
           k -> new ArrayList<>()
       );
       if (path.endsWith("/")) {
-        // segment is not compressed, list all objects with pagination and add them to delete list
-        listAllKeysWithPrefix(s3Client, s3Bucket, path, inputDataConfig.getMaxListingLength())
-            .forEach(key -> keysToDelete.add(new DeleteObjectsRequest.KeyVersion(key)));
+        // segment is not compressed, list objects and add them all to delete list
+        boolean hadException = deleteSegmentFilesFromS3(s3Bucket, path);
+        if (hadException) {
+          shouldThrowException = true;
+        }
       } else {
         keysToDelete.add(new DeleteObjectsRequest.KeyVersion(path));
         keysToDelete.add(new DeleteObjectsRequest.KeyVersion(DataSegmentKiller.descriptorPath(path)));
       }
     }
 
-    boolean shouldThrowException = false;
     for (Map.Entry<String, List<DeleteObjectsRequest.KeyVersion>> bucketToKeys : bucketToKeysToDelete.entrySet()) {
       String s3Bucket = bucketToKeys.getKey();
       List<DeleteObjectsRequest.KeyVersion> keysToDelete = bucketToKeys.getValue();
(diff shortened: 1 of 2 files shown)


extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
```diff
@@ -135,6 +131,40 @@ public void kill(List<DataSegment> segments) throws SegmentLoadingException
...
+   * so segment directories with more than 1000 files would previously be only partially deleted. This method
+   * follows continuation tokens to ensure all objects are listed and can be removed from deep storage.
+   *
+   * @param s3Client        client used to communicate with s3
+   * @param s3Bucket        the bucket name
+   * @param prefix          the key prefix (e.g. segment path ending with /)
+   * @param maxListingLength maximum keys per list request (e.g. 1000)
+   * @return list of all object keys under the prefix
+   */
+  private List<String> listAllKeysWithPrefix(
```
kfaraz: I think the `ObjectSummaryIterator` performs the same list operation in batches. Do you think it would be possible to reuse that class here instead of duplicating the logic?

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
```diff
@@ -103,13 +103,9 @@ public void kill(List<DataSegment> segments) throws SegmentLoadingException
...
       );
       if (path.endsWith("/")) {
-        // segment is not compressed, list objects and add them all to delete list
-        final ListObjectsV2Result list = s3Client.listObjectsV2(
-            new ListObjectsV2Request().withBucketName(s3Bucket).withPrefix(path)
-        );
-        for (S3ObjectSummary objectSummary : list.getObjectSummaries()) {
-          keysToDelete.add(new DeleteObjectsRequest.KeyVersion(objectSummary.getKey()));
-        }
+        // segment is not compressed, list all objects with pagination and add them to delete list
```
kfaraz: On a second glance, @abannon , I wonder if this change is even necessary. I am not sure if there can be a case where we have more than 1000 files for a single segment. Have you encountered such a case?

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
abannon: Yes, and actually, it might make sense to reuse even more by calling S3Utils.deleteObjectsInPath() 

nevermind, that would disrupt the delete batching.

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
abannon: Yes. (I am wondering if that suggests something is misconfigured, but regardless, I would like to get the files cleaned up)

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
```diff
@@ -212,6 +208,24 @@ private boolean deleteKeysForBucket(
...
+          s3ClientSupplier.get(),
+          inputDataConfig.getMaxListingLength(),
+          s3Bucket,
+          s3Path,
+          Predicates.alwaysTrue()
+      );
+    }
+    catch (Exception e) {
+      log.error("Error occurred while deleting segment files from s3. Error: %s", e.getMessage());
+      return true;
```
kfaraz, line 224: Please invert these return values as it is more natural to return `true` for success scenarios and `false` for failures.

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
```diff
@@ -212,6 +208,24 @@ private boolean deleteKeysForBucket(
...
+      S3Utils.deleteObjectsInPath(
+          s3ClientSupplier.get(),
+          inputDataConfig.getMaxListingLength(),
+          s3Bucket,
+          s3Path,
+          Predicates.alwaysTrue()
+      );
+    }
+    catch (Exception e) {
+      log.error("Error occurred while deleting segment files from s3. Error: %s", e.getMessage());
```
kfaraz, line 223: This would log the stack trace as well.

```suggestion
      log.error(e, "Error occurred while deleting segment files from s3.");
```

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
```diff
@@ -223,12 +237,17 @@ public void kill(DataSegment segment) throws SegmentLoadingException
...
+          S3Utils.deleteObjectsInPath(
+              s3ClientSupplier.get(),
+              inputDataConfig.getMaxListingLength(),
+              s3Bucket,
+              s3Path,
+              Predicates.alwaysTrue()
+          );
+        }
+        catch (Exception e) {
+          log.error("Error occurred while deleting segment files from s3. Error: %s", e.getMessage());
```
kfaraz, line 250:
```suggestion
          log.error(e, "Error occurred while deleting segment files from s3.");
```

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
```diff
@@ -104,19 +103,16 @@ public void kill(List<DataSegment> segments) throws SegmentLoadingException
       );
       if (path.endsWith("/")) {
         // segment is not compressed, list objects and add them all to delete list
-        final ListObjectsV2Result list = s3Client.listObjectsV2(
-            new ListObjectsV2Request().withBucketName(s3Bucket).withPrefix(path)
-        );
-        for (S3ObjectSummary objectSummary : list.getObjectSummaries()) {
-          keysToDelete.add(new DeleteObjectsRequest.KeyVersion(objectSummary.getKey()));
+        boolean hadException = deleteSegmentFilesFromS3(s3Bucket, path);
+        if (hadException) {
```
kfaraz, line 107:
```suggestion
        boolean success = deleteSegmentFilesFromS3(s3Bucket, path);
        if (!success) {
```

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentKiller.java
```diff
@@ -223,12 +237,17 @@ public void kill(DataSegment segment) throws SegmentLoadingException
 
       if (s3Path.endsWith("/")) {
         // segment is not compressed, list objects and delete them all
-        final ListObjectsV2Result list = s3Client.listObjectsV2(
-            new ListObjectsV2Request().withBucketName(s3Bucket).withPrefix(s3Path)
-        );
-        for (S3ObjectSummary objectSummary : list.getObjectSummaries()) {
-          log.info("Removing index file[s3://%s/%s] from s3!", s3Bucket, objectSummary.getKey());
-          s3Client.deleteObject(s3Bucket, objectSummary.getKey());
+        try {
```
kfaraz, line 240: Nit: Maybe just call the new method `deleteSegmentFilesFromS3` here, since the logic seems to be the same.

PR #19011, opened 2026-02-11
Broker Config and Query Blocklist

### LINKED ISSUE #18964: Broker-level dynamic config + query blacklist
### Motivation

Currently, Druid has a coordinator config but no broker-level dynamic config which can be appropriate for features like query blacklist enforcement, data aliases, routing rules, feature flags etc. The primary motivation for having a broker config is for blacklisting queries dynamically in the situation there is a rogue app/user spamming the cluster without relying on static configs/restarts. 

#### Broker config
Similar to CoordinatorDynamicConfig, introduce a broker-level dynamic config persisted in metadata store, polled periodically by brokers. Admin APIs to GET and POST configs (UI can be added later).

#### Query Blacklist
Enforce at broker query lifecycle, blacklist based on datasource in query or additionally, query context matches, querytype, remote address etc. 

### Rationale

Alternatively, we could reuse coordinator dynamic config for query blacklisting but it is not the appropriate enforcement point and would require broker to pull coordinator config. Moreover, broker config can be used for other future features such as datasource aliasing, routing rules or feature flags.

mshahid6:
Fixes #18964

### Description

Added a new **BrokerDynamicConfig** to enable dynamic, runtime configuration of Druid brokers without restarts and a query blocklist for dynamically blocking queries in the situation there is a rogue app/user spamming the cluster without relying on static configs/restarts.

#### QueryBlocklistRule

Enforced early in QueryLifecycle (after init) and throws DruidException when a query matches a rule i.e. if ALL specified criteria match (AND logic). Null or empty criteria act as wildcards (match everything):                                                                                                                                            
  - `dataSources`: Matches if ANY datasource in the query intersects with the rule's datasources                                                          
  - `queryTypes`: Matches if the query type is in the rule's query types                                                                                  
  - `contextMatches`: Matches if ALL key-value pairs in the rule match the query context (exact string match)  

Query blocklist is managed via the existing coordinator config APIs:
```
GET /druid/coordinator/v1/config/broker

POST /druid/coordinator/v1/config/broker
  {
    "queryBlocklist": [
      {
        "ruleName": "block-wikipedia-groupbys",
        "dataSources": ["wikipedia"],
        "queryTypes": ["groupBy"],
        "contextMatches": {"priority": "0"}
      }
    ]
  }

GET /druid/coordinator/v1/config/broker/history
```

#### Considerations
Creating a separate broker-level dynamic config was considered which can also be used for other future features such as datasource aliasing, routing rules or feature flags. However, Coordinator handles communications with the metadata store and syncs to brokers so that there are fewer servers communicating with the metadata store and future built-in metadata store operations go through central leaders

 The query blocklist follows a "best effort" approach:
  - If BrokerViewOfCoordinatorConfig is null, blocklist check is skipped
  - If BrokerDynamicConfig is null (config not yet loaded), queries are allowed to proceed
  - Once config is loaded, blocklist enforcement begins

Rationale: The query blocklist is an operational safety feature, not a security control. Blocking all queries when config is unavailable would be more disruptive than the problem it's trying to solve. Operators can still use coordinator/historical unavailability to enforce hard blocks if needed.

#### Release note
Added query blocklist feature for dynamically blocking queries without restarts. Operators can block queries by datasource, query type, or query context using the new /druid/coordinator/v1/config/broker API. Rules use AND logic (all criteria must match) and are stored in the metadata database.

 - BrokerDynamicConfig - New broker-specific dynamic config with queryBlocklist field
 - BrokerDynamicConfigSyncer - Coordinator component that syncs config to brokers
 - BrokerViewOfBrokerConfig - Broker component that fetches config from coordinator
- BaseDynamicConfigSyncer<T> - Base class for config syncers (extracted)
- BaseBrokerViewOfConfig<T> - Base class for broker config views (extracted)
- QueryBlocklistRule - Rule-based query matching with datasource/type/context criteria
- CoordinatorBrokerConfigsResource - REST endpoints for managing broker config
- QueryLifecycle - Added checkQueryBlocklist() method to enforce rules
- QueryLifecycleFactory - Injects BrokerViewOfBrokerConfig for accessing blocklist
- CoordinatorDynamicConfigSyncer - Refactored to extend BaseDynamicConfigSyncer
- BrokerViewOfCoordinatorConfig - Refactored to extend BaseBrokerViewOfConfig
- CliBroker - Added Guice binding for BrokerViewOfBrokerConfig
- CliCoordinator - Added Guice binding for BrokerDynamicConfigSyncer
- DruidCoordinator - Added lifecycle management for BrokerDynamicConfigSyncer
- BrokerDynamicConfigTest
- BrokerDynamicConfigResourceTest 
- CoordinatorBrokerConfigsResourceTest
- QueryBlocklistRuleTest
- QueryLifecycleTest 
- BrokerViewOfBrokerConfigTest 

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/http/BrokerDynamicConfigSyncer.java
+++ b/server/src/main/java/org/apache/druid/server/http/BrokerDynamicConfigSyncer.java
@@ -59,7 +59,7 @@ public class BrokerDynamicConfigSyncer extends BaseDynamicConfigSyncer<BrokerDyn
     this.currentConfig = configManager.watch(
         BrokerDynamicConfig.CONFIG_KEY,
         BrokerDynamicConfig.class,
-        new BrokerDynamicConfig(null)
+        BrokerDynamicConfig.builder().build()
     );
   }
 
--- a/server/src/main/java/org/apache/druid/server/http/CoordinatorBrokerConfigsResource.java
+++ b/server/src/main/java/org/apache/druid/server/http/CoordinatorBrokerConfigsResource.java
@@ -66,7 +66,7 @@ public class CoordinatorBrokerConfigsResource
     this.currentConfig = configManager.watch(
         BrokerDynamicConfig.CONFIG_KEY,
         BrokerDynamicConfig.class,
-        new BrokerDynamicConfig(null)
+        BrokerDynamicConfig.builder().build()
     );
     this.brokerDynamicConfigSyncer = brokerDynamicConfigSyncer;
   }
--- a/server/src/test/java/org/apache/druid/server/BrokerDynamicConfigResourceTest.java
+++ b/server/src/test/java/org/apache/druid/server/BrokerDynamicConfigResourceTest.java
@@ -50,7 +50,7 @@ public class BrokerDynamicConfigResourceTest
   @Test
   public void testGetBrokerDynamicConfig()
   {
-    BrokerDynamicConfig config = new BrokerDynamicConfig(null);
+    BrokerDynamicConfig config = BrokerDynamicConfig.builder().build();
     Mockito.when(brokerViewOfBrokerConfig.getDynamicConfig()).thenReturn(config);
 
     Response response = resource.getBrokerDynamicConfig();
@@ -63,7 +63,7 @@ public class BrokerDynamicConfigResourceTest
   @Test
   public void testSetBrokerDynamicConfig()
   {
-    BrokerDynamicConfig config = new BrokerDynamicConfig(null);
+    BrokerDynamicConfig config = BrokerDynamicConfig.builder().build();
 
     Response response = resource.setBrokerDynamicConfig(config);
 
--- a/server/src/test/java/org/apache/druid/client/BrokerViewOfBrokerConfigTest.java
+++ b/server/src/test/java/org/apache/druid/client/BrokerViewOfBrokerConfigTest.java
@@ -38,7 +38,7 @@ public class BrokerViewOfBrokerConfigTest
   @Before
   public void setUp() throws Exception
   {
-    config = new BrokerDynamicConfig(null);
+    config = BrokerDynamicConfig.builder().build();
     coordinatorClient = Mockito.mock(CoordinatorClient.class);
     Mockito.when(coordinatorClient.getBrokerDynamicConfig()).thenReturn(Futures.immediateFuture(config));
     target = new BrokerViewOfBrokerConfig(coordinatorClient);
--- a/server/src/test/java/org/apache/druid/client/broker/BrokerClientImplTest.java
+++ b/server/src/test/java/org/apache/druid/client/broker/BrokerClientImplTest.java
@@ -190,7 +190,7 @@ public class BrokerClientImplTest
   @Test
   public void testUpdateBrokerDynamicConfig() throws Exception
   {
-    final BrokerDynamicConfig config = new BrokerDynamicConfig(null);
+    final BrokerDynamicConfig config = BrokerDynamicConfig.builder().build();
 
     serviceClient.expectAndRespond(
         new RequestBuilder(HttpMethod.POST, "/druid-internal/v1/config/broker")
--- a/server/src/test/java/org/apache/druid/server/http/CoordinatorBrokerConfigsResourceTest.java
+++ b/server/src/test/java/org/apache/druid/server/http/CoordinatorBrokerConfigsResourceTest.java
@@ -51,7 +51,7 @@ public class CoordinatorBrokerConfigsResourceTest
   @Test
   public void testGetBrokerDynamicConfig()
   {
-    BrokerDynamicConfig config = new BrokerDynamicConfig(null);
+    BrokerDynamicConfig config = BrokerDynamicConfig.builder().build();
     AtomicReference<BrokerDynamicConfig> currentConfig = new AtomicReference<>(config);
 
     EasyMock.expect(
@@ -79,7 +79,7 @@ public class CoordinatorBrokerConfigsResourceTest
   @Test
   public void testGetBrokerDynamicConfigHistory()
   {
-    BrokerDynamicConfig config = new BrokerDynamicConfig(null);
+    BrokerDynamicConfig config = BrokerDynamicConfig.builder().build();
     AtomicReference<BrokerDynamicConfig> currentConfig = new AtomicReference<>(config);
 
     EasyMock.expect(
@@ -115,7 +115,7 @@ public class CoordinatorBrokerConfigsResourceTest
   @Test
   public void testGetBrokerDynamicConfigHistoryWithNullIntervalAndCount()
   {
-    BrokerDynamicConfig config = new BrokerDynamicConfig(null);
+    BrokerDynamicConfig config = BrokerDynamicConfig.builder().build();
     AtomicReference<BrokerDynamicConfig> currentConfig = new AtomicReference<>(config);
 
     EasyMock.expect(
@@ -151,7 +151,7 @@ public class CoordinatorBrokerConfigsResourceTest
   @Test
   public void testSetBrokerDynamicConfig()
   {
-    BrokerDynamicConfig config = new BrokerDynamicConfig(null);
+    BrokerDynamicConfig config = BrokerDynamicConfig.builder().build();
     AtomicReference<BrokerDynamicConfig> currentConfig = new AtomicReference<>(config);
     HttpServletRequest request = EasyMock.createNiceMock(HttpServletRequest.class);
 


server/src/main/java/org/apache/druid/server/coordinator/CoordinatorDynamicConfig.java
```diff
@@ -172,6 +181,7 @@ public CoordinatorDynamicConfig(
     this.validDebugDimensions = validateDebugDimensions(debugDimensions);
     this.turboLoadingNodes = Configs.valueOrDefault(turboLoadingNodes, Set.of());
     this.cloneServers = Configs.valueOrDefault(cloneServers, Map.of());
+    this.queryBlocklist = queryBlocklist != null ? ImmutableList.copyOf(queryBlocklist) : Defaults.QUERY_BLOCKLIST;
```
jtuglu1: Is there a reason we are creating an immutable copy of this? Maybe let's use `Configs.valueOrDefault` to follow convention here with the rest of the containers being serde'd.

server/src/main/java/org/apache/druid/server/coordinator/CoordinatorDynamicConfig.java
```diff
@@ -454,6 +478,7 @@ public static int getDefaultBalancerComputeThreads()
 
   private static class Defaults
   {
+    static final List<QueryBlocklistRule> QUERY_BLOCKLIST = ImmutableList.of();
```
jtuglu1: nit: `List.of()`

server/src/main/java/org/apache/druid/server/QueryBlocklistRule.java
```diff
@@ -0,0 +1,176 @@
...
+import org.apache.druid.query.Query;
+
+import javax.annotation.Nullable;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * A rule for matching queries against blocklist criteria. A query matches this rule if ALL
+ * specified criteria match (AND logic). Null or empty criteria match everything.
```
jtuglu1, line 36: ~~I think we should negate this. Queries matching all criteria in the `QueryBlocklistRule` are blocked, and those that do not match are not blocked.~~

Ah – NVM, this is an AND. Ignore

server/src/main/java/org/apache/druid/server/QueryBlocklistRule.java
```diff
@@ -0,0 +1,176 @@
...
+      @JsonProperty("queryTypes") @Nullable Set<String> queryTypes,
+      @JsonProperty("contextMatches") @Nullable Map<String, String> contextMatches
+  )
+  {
+    Preconditions.checkArgument(
+        !Strings.isNullOrEmpty(ruleName),
+        "ruleName must not be null or empty"
+    );
+
+    // At least one criterion must be specified to prevent accidentally blocking all queries
```
jtuglu1, line 65: ~~This seems backwards to me. In my mind, passing all nulls should result in the default behavior. In other words, I need to "opt-in" to the blocking behavior.~~

~~IMO, we should make it easier for users to do the "default" thing which is allowing the query through. A rule matching a query <===> that query is blocked. Right now, it's the other way around.~~

NVM.

server/src/main/java/org/apache/druid/server/QueryLifecycle.java
```diff
@@ -310,6 +317,35 @@ private void preAuthorized(
...
+    if (config == null) {
+      return; // Config not loaded yet, allow query (best effort)
+    }
+
+    for (QueryBlocklistRule rule : config.getQueryBlocklist()) {
+      if (rule.matches(this.baseQuery)) {
+        throw DruidException.forPersona(DruidException.Persona.USER)
+                            .ofCategory(DruidException.Category.FORBIDDEN)
+                            .build(
+                                "Query blocked by rule[%s]",
```
jtuglu1: Does this log the query ID? It'd be nice to co-locate the query ID and the rule name in a log so users/operators don't need to grep that separately.

server/src/main/java/org/apache/druid/server/QueryBlocklistRule.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /**
+   * Returns true if the query matches ALL specified criteria (AND logic).
+   * Null or empty criteria match everything.
+   *
+   * @param query the query to check
+   * @return true if the query matches this rule, false otherwise
+   */
+  public boolean matches(Query<?> query)
```
jtuglu1, line 116: It might be worth caching this result for all 3 types. JVM might be smart and branch predict this every time but since this is called per-query on constant fields, it might be worth explicitly doing this to save ≥ 2 comparisons per query.

server/src/main/java/org/apache/druid/server/QueryBlocklistRule.java
jtuglu1: Ah – NVM, this is an AND. Ignore

server/src/main/java/org/apache/druid/server/QueryBlocklistRule.java
jtuglu1: I think this is moot given the above, ignore.

jtuglu1:
Another thing I forgot to mention is docs + UI, namely:

Docs:
1. Add the field to the coordinator dynamic config description, explaining what it does.
2. Separate documentation section for the rulesets. What the rule format looks like (description of each field, its default, etc.). The AND logic between the fields' predicates.

UI:
For completeness since coordinator config already in UI, do we want to add this as a field? Should be pretty simple to do with an AI tool.

sql/src/test/java/org/apache/druid/sql/calcite/util/QueryFrameworkUtils.java
```diff
@@ -102,7 +102,8 @@ public static QueryLifecycleFactory createMockQueryLifecycleFactory(
         new AuthConfig(),
         NoopPolicyEnforcer.instance(),
         authorizerMapper,
-        Suppliers.ofInstance(new DefaultQueryConfig(ImmutableMap.of()))
+        Suppliers.ofInstance(new DefaultQueryConfig(ImmutableMap.of())),
```
jtuglu1: nit: `ImmutableMap.of() -> Map.of()`

server/src/test/java/org/apache/druid/server/QueryResourceTest.java
```diff
@@ -1789,7 +1801,8 @@ public <T> QueryRunner<T> getQueryRunnerForSegments(Query<T> query, Iterable<Seg
             new AuthConfig(),
             NoopPolicyEnforcer.instance(),
             AuthTestUtils.TEST_AUTHORIZER_MAPPER,
-            Suppliers.ofInstance(new DefaultQueryConfig(ImmutableMap.of()))
+            Suppliers.ofInstance(new DefaultQueryConfig(ImmutableMap.of())),
```
jtuglu1: same as below

server/src/test/java/org/apache/druid/server/QueryResourceTest.java
```diff
@@ -1400,7 +1411,8 @@ public Access authorize(AuthenticationResult authenticationResult, Resource reso
             new AuthConfig(),
             NoopPolicyEnforcer.instance(),
             authMapper,
-            Suppliers.ofInstance(new DefaultQueryConfig(ImmutableMap.of()))
+            Suppliers.ofInstance(new DefaultQueryConfig(ImmutableMap.of())),
```
jtuglu1: same

server/src/test/java/org/apache/druid/server/QueryBlocklistRuleTest.java
```diff
@@ -0,0 +1,179 @@
...
+    Assert.assertFalse(rule.matches(onlyContextMatches));
+  }
+
+  @Test
+  public void testWildcardBehavior_nullQueryTypes()
+  {
+    QueryBlocklistRule rule = new QueryBlocklistRule(
+        "block-datasource-all-types",
+        ImmutableSet.of("blocked_ds"),
+        null,  // null means match all query types
```
jtuglu1, line 162: It might be worth specifying in the docs that `null` values in the other 2 parameters will default to match all (`*`) behavior when evaluating.

server/src/main/java/org/apache/druid/server/QueryBlocklistRule.java
```diff
@@ -0,0 +1,175 @@
...
+    if (hasQueryTypeCriteria) {
+      if (!queryTypes.contains(query.getType())) {
+        return false;
+      }
+    }
+
+    if (hasContextCriteria) {
+      for (Map.Entry<String, String> entry : contextMatches.entrySet()) {
+        Object contextValue = query.getContext().get(entry.getKey());
+        if (!entry.getValue().equals(String.valueOf(contextValue))) {
```
jtuglu1: Unless we're restricting the possible value set, this String.valueOf() will NPE if the `contextValue` is `null`. While this isn't a typical context value, it's still valid context. I think we should be a bit more graceful and allow the query or 403 depending on how we want to handle the `null` in contextValue(in other fields, we do as match-all for example).

server/src/test/java/org/apache/druid/server/QueryBlocklistRuleTest.java
mshahid6: https://github.com/apache/druid/pull/19011/changes/BASE..ab49ff09e6954ca00d9485650a55ce1b395dcfb6#diff-23b572e1c4ec981b4ca8ebdac28e9bb94dc71bb29c8342a68bd483c39f362ffeR857 
i can add to this?

server/src/test/java/org/apache/druid/segment/metadata/SegmentMetadataCacheTestBase.java
```diff
@@ -292,7 +292,8 @@ public QueryLifecycleFactory getQueryLifecycleFactory(QuerySegmentWalker walker)
         new AuthConfig(),
         NoopPolicyEnforcer.instance(),
         AuthTestUtils.TEST_AUTHORIZER_MAPPER,
-        Suppliers.ofInstance(new DefaultQueryConfig(ImmutableMap.of()))
+        Suppliers.ofInstance(new DefaultQueryConfig(ImmutableMap.of())),
```
jtuglu1: one more 😅 ImmutableMap -> Map.of()

server/src/main/java/org/apache/druid/server/QueryLifecycle.java
```diff
@@ -134,6 +138,7 @@ public QueryLifecycle(
     this.defaultQueryConfig = defaultQueryConfig;
     this.authConfig = authConfig;
     this.policyEnforcer = policyEnforcer;
+    this.brokerViewOfCoordinatorConfig = brokerViewOfCoordinatorConfig;
```
jtuglu1: nit: maybe add small comment saying this can only be `null` if not running on the broker

server/src/test/java/org/apache/druid/server/QueryResourceTest.java
```diff
@@ -1102,7 +1110,8 @@ public Access authorize(AuthenticationResult authenticationResult, Resource reso
...
```
jtuglu1: Can we replace all these too? `Map.of()`

jtuglu1:
@gianm any final thoughts here? Otherwise, I'm planning to merge this as-is. We can address any subsequent comments in a follow-up.

mshahid6:
@maytasm oh interesting, just curious, what are the advantages of doing it that way?

jtuglu1:
> LGTM. One thing Im wondering is if we should make a new API (on the coordinator) for this blocklist rules and store it separately from the Coordinator Dynamic configs. For example, in Basic Security, you can POST(/druid-ext/basic-security/authentication/db/{authenticatorName}/users/{userName}) on the Coordinator to create a new User. Basically we would have a new set of APIs for GET, POST, etc for blocklist resource. I think this is what @gianm is thinking of in [#19011 (review)](https://github.com/apache/druid/pull/19011#pullrequestreview-3786192719). Although reusing the Coordinator Dynamic configs gives us for 'free' all the existing framework for syncing configs between Broker and Coordinator. Let's wait for @gianm

I'm not sure this is needed right now. 1) IMO we don't have a need programmatically (yet) for this that we cannot solve with a get/put of dynamic config API. 2) you can make the same argument for a good portion of the overlord dynamic config as well, yet that also is a single, dynamic config API. I think it keeps things simple for now, and if we want to expand it later, we can. You get all the built-in updating/propagation/auditing support for something that (hopefully) shouldn't be updated that often. However, if we were to extend this to datasource aliasing (which I don't think would fit in here), then I can imagine a higher write tput necessitating a separate API. The litmus test, IMO, for whether to extract something from dynamic config should be: how frequently is it read/written to (by clients), how "large" is the config (does it scale linearly with some sort of cluster resource – datasource, user, node, etc.), and how long do these configs tend to stay set to a specific value (e.g. is it truly "dynamic").

maytasm:
> @maytasm oh interesting, just curious, what are the advantages of doing it that way?

I agree on all of @jtuglu1 points above. I don't have strong opinion but just feel a little weird for this config to be in Coordinator Dynamic config as it has nothing to do with Coordination. All the other configs in Coordinator Dynamic config are related to Coordinator Duties like moving segments, loading segments, killing segments, replicating segments, etc.

server/src/main/java/org/apache/druid/server/coordinator/CoordinatorDynamicConfig.java
```diff
@@ -351,6 +360,17 @@ public Set<String> getTurboLoadingNodes()
     return turboLoadingNodes;
   }
 
+  /**
+   * List of query blocklist rules for dynamically blocking queries on brokers.
+   *
+   * @return List of query blocklist rules
+   */
+  @JsonProperty
+  public List<QueryBlocklistRule> getQueryBlocklist()
```
gianm: Please make this `@JsonInclude(NON_EMPTY)` so it is skipped from the serialized form of config if not present. Keeps things cleaner.

server/src/main/java/org/apache/druid/server/QueryLifecycle.java
```diff
@@ -121,6 +124,7 @@ public QueryLifecycle(
       final DefaultQueryConfig defaultQueryConfig,
       final AuthConfig authConfig,
       final PolicyEnforcer policyEnforcer,
+      @Nullable final BrokerViewOfCoordinatorConfig brokerViewOfCoordinatorConfig,
```
gianm: Better to keep things targeted and only pass in the `List<QueryBlocklistRule>`. It makes usage and testing easier.

server/src/main/java/org/apache/druid/server/QueryLifecycle.java
```diff
@@ -159,6 +164,8 @@ public <T> QueryResponse<T> runSimple(
   {
     initialize(query);
 
+    checkQueryBlocklist();
```
gianm: `runSimple` is not always used, which means the blocklist won't always be checked if the call is put here. Some usages of `QueryLifecycle` call the methods individually.

IMO best to put it in `doAuthorize`. This method is always called-- it's the only place the state can transition to `AUTHORIZED`.

It would also work to put it in `execute`.

docs/configuration/index.md
```diff
@@ -838,6 +838,73 @@ The following table shows the dynamic configuration properties for the Coordinat
...
+|`queryBlocklist`| List of rules to block queries based on datasource, query type, and/or query context parameters. Each rule defines criteria that are combined with AND logic. Blocked queries return an HTTP 403 error. See [Query blocklist rules](#query-blocklist-rules) for details.|none|
+
+##### Query blocklist rules
+
+Query blocklist rules allow you to block specific queries based on datasource, query type, and/or query context parameters. This feature is useful for preventing expensive or problematic queries from impacting cluster performance.
```
gianm: The documentation should mention that blocking is best-effort, and may not be applied in certain cases (such as if the Broker has recently started up, can't contact the Coordinator, etc).

docs/configuration/index.md
```diff
@@ -838,6 +838,73 @@ The following table shows the dynamic configuration properties for the Coordinat
...
+
+Each rule in the `queryBlocklist` array is a JSON object with the following properties:
+
+|Property|Description|Required|Default|
+|--------|-----------|--------|-------|
+|`ruleName`|Unique name identifying this blocklist rule. Used in error messages when queries are blocked.|Yes|N/A|
+|`dataSources`|List of datasource names to match. A query matches if it references any datasource in this list.|No|Matches all datasources|
+|`queryTypes`|List of query types to match (e.g., `scan`, `timeseries`, `groupBy`, `topN`). A query matches if its type is in this list.|No|Matches all query types|
```
gianm: I don't see a test for this `queryTypes` feature. Please add one.

server/src/main/java/org/apache/druid/server/http/CoordinatorBrokerConfigsResource.java
```diff
@@ -0,0 +1,259 @@
...
+    this.auditManager = auditManager;
+    this.currentConfig = configManager.watch(
+        BrokerDynamicConfig.CONFIG_KEY,
+        BrokerDynamicConfig.class,
+        new BrokerDynamicConfig(null)
+    );
+    this.clientFactory = clientFactory;
+    this.jsonMapper = jsonMapper;
+    this.druidNodeDiscovery = druidNodeDiscoveryProvider;
+    this.exec = Execs.singleThreaded("CoordinatorBrokerConfigPush-%d");
```
maytasm: where is this used? Should the pushConfigToBrokers be done using this exec so that it doesn't block calling thread?

server/src/main/java/org/apache/druid/server/http/CoordinatorBrokerConfigsResource.java
```diff
@@ -0,0 +1,259 @@
...
+  @Consumes(MediaType.APPLICATION_JSON)
+  public Response setBrokerDynamicConfig(
+      BrokerDynamicConfig newConfig,
+      @Context HttpServletRequest req
+  )
+  {
+    try {
+      final SetResult setResult = configManager.set(
+          BrokerDynamicConfig.CONFIG_KEY,
+          newConfig,
```
maytasm, line 94: The current Coordinator dynamic config supports giving partial config (only updating the keys that is given in the payload and other keys are kept as-is unchanged). Can we do the same here too? i.e. building the newConfig from currentConfig + updating the keys/values from the request

server/src/main/java/org/apache/druid/server/http/CoordinatorBrokerConfigsResource.java
```diff
@@ -0,0 +1,259 @@
...
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.Collection;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
+
+@Path("/druid/coordinator/v1/broker/config")
+@ResourceFilters(ConfigResourceFilter.class)
+public class CoordinatorBrokerConfigsResource
```
maytasm, line 48: There's a lot of repeated/common code here and CoordinatorDynamicConfigSyncer and a few repeated/common code here and CoordinatorConfigManager. Can we refactor the common code into a base class (i.e. BaseDynamicConfigSyncer) and then have CoordinatorDynamicConfigSyncer/BrokerDynamicConfigSyncer extends from the base class?

server/src/main/java/org/apache/druid/client/BrokerViewOfBrokerConfig.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public class BrokerViewOfBrokerConfig
+{
+  private static final Logger log = new Logger(BrokerViewOfBrokerConfig.class);
+  private final AtomicReference<BrokerDynamicConfig> config;
+
+  @Inject
+  public BrokerViewOfBrokerConfig(JacksonConfigManager configManager)
+  {
+    this.config = configManager.watch(
```
maytasm: This should poll from the Coordinator. Brokers do not talk to the metadata store. This would be similar to BrokerViewOfCoordinatorConfig. Can we see if we can refactor common code into a base class and have BrokerViewOfCoordinatorConfig/BrokerViewOfBrokerConfig extends from the base class.

mshahid6: would only BrokerViewOfCoordinatorConfig implement HistoricalFilter and have the queryable server logic?

maytasm: Since those configs are already in Coordinator Dynamic configs, we can leave it in BrokerViewOfCoordinatorConfig. So BrokerViewOfCoordinatorConfig can extends BaseBrokerViewConfig and extends HistoricalFilter while BrokerViewOfBrokerConfig can just extends BaseBrokerViewConfig

server/src/main/java/org/apache/druid/server/broker/BrokerDynamicConfig.java
```diff
@@ -0,0 +1,139 @@
...
+     * the default value if not configured.
+     */
+    public BrokerDynamicConfig build(@Nullable BrokerDynamicConfig defaults)
+    {
+      return new BrokerDynamicConfig(
+        valueOrDefault(queryBlocklist, defaults != null ? defaults.getQueryBlocklist() : null)
+      );
+    }
+
+    private static <T> T valueOrDefault(@Nullable T value, @NotNull T defaultValue)
```
jtuglu1: nit: let's use `Configs.valueOrDefault`

server/src/main/java/org/apache/druid/server/http/BaseDynamicConfigSyncer.java
```diff
@@ -0,0 +1,275 @@
...
+  @VisibleForTesting
+  public void broadcastConfigToBrokers()
+  {
+    invalidateInSyncBrokersIfNeeded();
+    final Stopwatch stopwatch = Stopwatch.createStarted();
+    for (DiscoveryDruidNode broker : getKnownBrokers()) {
+      pushConfigToBrokerNode(broker);
+    }
+    emitStat(
+        Stats.Configuration.TOTAL_SYNC_TIME,
```
jtuglu1, line 131: maybe now that we are syncing 2 types of configs to brokers, we should make the metric distinguish which config type is being pushed? I'm leaning more towards a dimension rather than a separate metric.

server/src/main/java/org/apache/druid/server/http/BaseDynamicConfigSyncer.java
```diff
@@ -0,0 +1,275 @@
...
+  {
+    return Set.copyOf(inSyncBrokers);
+  }
+
+  /**
+   * Schedules a periodic sync with brokers when the coordinator becomes the leader.
+   */
+  public void onLeaderStart()
+  {
+    log.info("Starting %s syncing to brokers on leader node.", getConfigTypeName());
```
jtuglu1: nit: `Starting syncing config [%s] to brokers on leader node`

server/src/main/java/org/apache/druid/server/http/CoordinatorDynamicConfigSyncer.java
```diff
@@ -86,179 +47,31 @@ public CoordinatorDynamicConfigSyncer(
...
-  private void emitStat(CoordinatorStat stat, RowKey rowKey, long value)
+  @Override
+  protected String getConfigTypeName()
   {
-    ServiceMetricEvent.Builder eventBuilder = new ServiceMetricEvent.Builder();
-    rowKey.getValues().forEach(
-        (dim, dimValue) -> eventBuilder.setDimension(dim.reportedName(), dimValue)
-    );
-    emitter.emit(eventBuilder.setMetric(stat.getMetricName(), value));
+    return "coordinator dynamic configuration";
```
jtuglu1: maybe make this a single phrase? `coordinator` or `broker` would show up better in logs/metric dimensions.

server/src/test/java/org/apache/druid/server/BrokerDynamicConfigResourceTest.java
```diff
@@ -0,0 +1,97 @@
...
+    BrokerDynamicConfig config = new BrokerDynamicConfig(null);
+
+    Response response = resource.setBrokerDynamicConfig(config);
+
+    Assert.assertEquals(200, response.getStatus());
+    Mockito.verify(brokerViewOfBrokerConfig, Mockito.times(1)).setDynamicConfig(config);
+  }
+
+  @Test
+  public void testGetDynamicConfig()
```
jtuglu1: let's maybe rename these get/set to have coordinator dynamic in them (distinguish from broker dynamic).

server/src/main/java/org/apache/druid/server/http/BrokerDynamicConfigSyncer.java
```diff
@@ -0,0 +1,83 @@
...
+  @Override
+  protected boolean pushConfigToBroker(BrokerClient brokerClient, BrokerDynamicConfig config) throws Exception
+  {
+    return brokerClient.updateBrokerDynamicConfig(config).get();
+  }
+
+  @Override
+  protected String getConfigTypeName()
+  {
+    return "broker dynamic configuration";
```
jtuglu1: same thing here about type name

server/src/main/java/org/apache/druid/client/BaseBrokerViewOfConfig.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Base class for broker's view of dynamic configuration fetched from the Coordinator.
+ * Subclasses must implement:
+ * {@link #fetchConfigFromClient()} to fetch configuration from their specific client</li>
+ * {@link #getConfigTypeName()} for logging purposes</li>
+ *
+ * @param <T> the type of dynamic configuration (e.g., CoordinatorDynamicConfig, BrokerDynamicConfig)
+ */
+public abstract class BaseBrokerViewOfConfig<T>
```
jtuglu1: could we make this template parameter's name more match what it is? e.g. `DynamicConfig`

docs/configuration/index.md
```diff
@@ -1862,6 +1862,89 @@ queries in order to avoid running as a default priority of 0.
 |--------|-----------|-------|
 |`druid.broker.internal.query.config.context`|A string formatted `key:value` map of a query context to add to internally generated broker queries.|null|
 
+#### Dynamic configuration
```
jtuglu1, line 1865: let's move this section to `docs/api-reference/dynamic-configuration-api.md`.

server/src/main/java/org/apache/druid/server/http/BrokerDynamicConfigSyncer.java
```diff
@@ -0,0 +1,83 @@
...
+        clientFactory,
+        jsonMapper,
+        druidNodeDiscoveryProvider,
+        emitter,
+        Execs.scheduledSingleThreaded("BrokerDynamicConfigSyncer-%d")
+    );
+    this.currentConfig = configManager.watch(
+        BrokerDynamicConfig.CONFIG_KEY,
+        BrokerDynamicConfig.class,
+        new BrokerDynamicConfig(null)
```
maytasm: nit: use Builder

maytasm (approved):
LGTM. Thanks!!

abhishekrb19:
I think something from this PR is causing one of the embedded tests to consistently fail in master and on all recent PRs:

```
[Execute: .github/scripts/run_unit-tests -Dtest=!QTest,'I*,A*,U*' -Dmaven.test.failure.ignore=true](https://github.com/apache/druid/actions/runs/22832980666/job/66354459774#annotation:6:8087)
Process completed with exit code 1.
```

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.5.4:test (default-test) on project druid-embedded-tests: 
Error:  
Error:  See /home/runner/work/druid/druid/embedded-tests/target/surefire-reports for the individual test results.
Error:  See dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
Error:  The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
Error:  Command was /bin/sh -c cd '/home/runner/work/druid/druid/embedded-tests' && '/opt/hostedtoolcache/Java_Zulu_jdk/21.0.10-7/x64/bin/java' '--add-exports=java.base/jdk.internal.ref=ALL-UNNAMED' '--add-exports=java.base/jdk.internal.misc=ALL-UNNAMED' '--add-opens=java.base/java.nio=ALL-UNNAMED' '--add-opens=java.base/sun.nio.ch=ALL-UNNAMED' '--add-opens=java.base/jdk.internal.ref=ALL-UNNAMED' '--add-opens=java.base/java.io=ALL-UNNAMED' '--add-opens=java.base/java.lang=ALL-UNNAMED' '--add-opens=jdk.management/com.sun.management.internal=ALL-UNNAMED' '--add-opens=java.base/java.util=ALL-UNNAMED' '-Djava.security.manager=allow' '-Xmx2048m' '-XX:MaxDirectMemorySize=2500m' '-XX:+ExitOnOutOfMemoryError' '-XX:+HeapDumpOnOutOfMemoryError' '-Duser.language=en' '-Duser.GroupByQueryRunnerTest.javacountry=US' '-Dfile.encoding=UTF-8' '-Duser.timezone=UTC' '-Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager' '-Daws.region=us-east-1' '-Ddruid.test.stupidPool.poison=true' '-XX:OnOutOfMemoryError=/home/runner/work/druid/druid/dev/chmod-heap-dumps.sh' '-XX:HeapDumpPath=/home/runner/work/druid/druid/target' '-Ddruid.indexing.doubleStorage=double' '-javaagent:/home/runner/work/druid/druid/jfr-profiler.jar' '-Djfr.profiler.http.username=druid-ci' '-Djfr.profiler.http.***' '-Djfr.profiler.tags.project=druid' '-Djfr.profiler.tags.jvm_version=21.0.10' '-Djfr.profiler.tags.run_id=22832980666' '-Djfr.profiler.tags.run_number=19767' '-Djfr.profiler.tags.run_attempt=2' '-Djfr.profiler.tags.key=test-jdk21-[I*,A*,U*]' '-Djfr.profiler.tags.event_ref=pull_request-19011' '-Djfr.profiler.tags.run_url=https://github.com/apache/druid/actions/runs/22832980666' '-jar' '/home/runner/work/druid/druid/embedded-tests/target/surefire/surefirebooter-20260309201804026_2049.jar' '/home/runner/work/druid/druid/embedded-tests/target/surefire' '2026-03-09T20-00-18_955-jvmRun1' 'surefire-20260309201804026_2047tmp' 'surefire_23-20260309201804026_2048tmp'
Error:  Error occurred in starting fork, check output in log
Error:  Process Exit Code: 1
Error:  org.apache.maven.surefire.booter.SurefireBooterForkException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
Error:  Command was /bin/sh -c cd '/home/runner/work/druid/druid/embedded-tests' && '/opt/hostedtoolcache/Java_Zulu_jdk/21.0.10-7/x64/bin/java' '--add-exports=java.base/jdk.internal.ref=ALL-UNNAMED' '--add-exports=java.base/jdk.internal.misc=ALL-UNNAMED' '--add-opens=java.base/java.nio=ALL-UNNAMED' '--add-opens=java.base/sun.nio.ch=ALL-UNNAMED' '--add-opens=java.base/jdk.internal.ref=ALL-UNNAMED' '--add-opens=java.base/java.io=ALL-UNNAMED' '--add-opens=java.base/java.lang=ALL-UNNAMED' '--add-opens=jdk.management/com.sun.management.internal=ALL-UNNAMED' '--add-opens=java.base/java.util=ALL-UNNAMED' '-Djava.security.manager=allow' '-Xmx2048m' '-XX:MaxDirectMemorySize=2500m' '-XX:+ExitOnOutOfMemoryError' '-XX:+HeapDumpOnOutOfMemoryError' '-Duser.language=en' '-Duser.GroupByQueryRunnerTest.javacountry=US' '-Dfile.encoding=UTF-8' '-Duser.timezone=UTC' '-Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager' '-Daws.region=us-east-1' '-Ddruid.test.stupidPool.poison=true' '-XX:OnOutOfMemoryError=/home/runner/work/druid/druid/dev/chmod-heap-dumps.sh' '-XX:HeapDumpPath=/home/runner/work/druid/druid/target' '-Ddruid.indexing.doubleStorage=double' '-javaagent:/home/runner/work/druid/druid/jfr-profiler.jar' '-Djfr.profiler.http.username=druid-ci' '-Djfr.profiler.http.***' '-Djfr.profiler.tags.project=druid' '-Djfr.profiler.tags.jvm_version=21.0.10' '-Djfr.profiler.tags.run_id=22832980666' '-Djfr.profiler.tags.run_number=19767' '-Djfr.profiler.tags.run_attempt=2' '-Djfr.profiler.tags.key=test-jdk21-[I*,A*,U*]' '-Djfr.profiler.tags.event_ref=pull_request-19011' '-Djfr.profiler.tags.run_url=https://github.com/apache/druid/actions/runs/22832980666' '-jar' '/home/runner/work/druid/druid/embedded-tests/target/surefire/surefirebooter-20260309201804026_2049.jar' '/home/runner/work/druid/druid/embedded-tests/target/surefire' '2026-03-09T20-00-18_955-jvmRun1' 'surefire-20260309201804026_2047tmp' 'surefire_23-20260309201804026_2048tmp'
Error:  Error occurred in starting fork, check output in log
Error:  Process Exit Code: 1
Error:  	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:643)
```

 https://github.com/apache/druid/actions/runs/22832980666/job/66354459774

abhishekrb19:
Just downloaded the test-report artifacts from GitHub and I see `org.apache.druid.testing.embedded.docker.IngestionBackwardCompatibilityDockerTest` was the last embedded test and the test JVM crashes.

Since the test [uses](https://github.com/apache/druid/blob/master/embedded-tests/src/test/java/org/apache/druid/testing/embedded/docker/IngestionBackwardCompatibilityDockerTest.java#L42) an old Coordinator from 31.0.0, would some of the changes here be compatible with that? I’m wondering if the test might actually be uncovering a legitimate backwards compatibility issue with an old Coordinator + new Brokers.

cc: @mshahid6 @jtuglu1

jtuglu1:
> IngestionBackwardCompatibilityDockerTest.java

Yes, I saw that too – looking. My guess is brokers are calling coordinator config endpoint on boot (which doesn't exist on old coordinators).

server/src/main/java/org/apache/druid/client/coordinator/CoordinatorClientImpl.java
```diff
@@ -253,6 +254,22 @@ public ListenableFuture<CoordinatorDynamicConfig> getCoordinatorDynamicConfig()
...
 
+  @Override
+  public ListenableFuture<BrokerDynamicConfig> getBrokerDynamicConfig()
+  {
+    return FutureUtils.transform(
+        client.asyncRequest(
+            new RequestBuilder(HttpMethod.GET, "/druid/coordinator/v1/broker/config"),
+            new BytesFullResponseHandler()
+        ),
+        holder -> JacksonUtils.readValue(
```
abhishekrb19, line 265: Looks like this change is causing `org.apache.druid.testing.embedded.docker.IngestionBackwardCompatibilityDockerTest` to fail consistently (which seems like it would be a legit issue during rolling upgrades since the Coordinators are typically upgraded in the end).


I haven’t looked closely at the surrounding code yet, but either this code or the caller likely needs to account for the new `/druid/coordinator/v1/broker/config` endpoint not being available on Coordinators during rolling upgrades in a backward compatible manner.

kfaraz: Yes, that must be the reason.

Probably just catching the not found exception and continuing with the Broker start up should suffice.

jtuglu1: Yeah – I have a test branch that I'll raise in a bit to fix this

jtuglu1: Will summarize the changes there

abhishekrb19: Thanks for the [fix](https://github.com/apache/druid/pull/19120), @jtuglu1! (And to @kfaraz for adding the backward compatibility embedded test 🙂)

PR #19014, opened 2026-02-12
feat: Kubernetes executionconfig API should return effective config

GWphua:
### Description
#18591 introduced the capability to change the max number of K8s tasks created by Overlord, and we can track the values using `/druid/indexer/v1/k8s/taskrunner/executionconfig` API.

#### Execution Config Returns Static Config by Default
Currently, calling GET on the `/druid/indexer/v1/k8s/taskrunner/executionconfig` API will return an empty payload if we did not submit any dynamic configs via POST on the same API. Users who depend on the API to retrieve the currently effective K8s task runner information will not see the information before the first dynamic config change is submitted. 

Changed the API to return the effective config, so the users will always see what is currently being used.

#### Release note

Prior to this change, GET `/druid/indexer/v1/k8s/taskrunner/executionconfig` will return an empty value when the dynamic config has not been updated via POST `/druid/indexer/v1/k8s/taskrunner/executionconfig`. This PR changes this behaviour to always reflect the dynamic config that will be used by the task runner to create K8s jobs.

<hr>

This PR has:

- [x] been self-reviewed.


FrankChen021:
the behaviour of this API changes. please also update the document

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/execution/KubernetesTaskExecutionConfigResource.java
```diff
@@ -84,12 +86,9 @@ public Response setExecutionConfig(
       @Context final HttpServletRequest req
   )
   {
-    KubernetesTaskRunnerDynamicConfig currentConfig = getDynamicConfig();
-    KubernetesTaskRunnerDynamicConfig mergedConfig = dynamicConfig;
+    KubernetesTaskRunnerDynamicConfig currentConfig = getCurrentConfiguration();
+    KubernetesTaskRunnerDynamicConfig mergedConfig = currentConfig.merge(dynamicConfig);
```
FrankChen021: P2 POST persists static fallback values as dynamic overrides

Merging the request into getCurrentConfiguration() means the value saved to metadata is the effective config, not just the dynamic override. For example, with no dynamic config and a static druid.indexer.runner.capacity of 10, a POST that only changes podTemplateSelectStrategy will persist capacity=10 as dynamic config. If the operator later changes the static capacity to 20 and restarts, the persisted dynamic capacity still wins, so the task runner unexpectedly stays capped at 10. The POST path should merge against the persisted dynamic config, or otherwise avoid writing static fallback values into dynamic config.

FrankChen021 (commented):

---

docs/development/extensions-core/k8s-jobs.md
```diff
@@ -83,7 +83,9 @@ To use these APIs, ensure you have read and write permissions for the CONFIG res
 
 #### Get dynamic configuration
 
-Retrieves the current dynamic execution config for the Kubernetes task runner. 
+> Prior to Druid 37.0.0, this API will return an empty value when the dynamic config has not been updated via the POST method below. This has since changed to always reflect the dynamic config that will be used by the task runner to create K8s jobs.
```
FrankChen021:
```suggestion
> Prior to Druid 38.0.0, this API will return an empty value when the dynamic config has not been updated via the POST method below. This has since changed to always reflect the dynamic config that will be used by the task runner to create K8s jobs.
```

FrankChen021 (approved):

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/execution/DefaultKubernetesTaskRunnerDynamicConfig.java
```diff
@@ -35,14 +35,17 @@ public class DefaultKubernetesTaskRunnerDynamicConfig implements KubernetesTaskR
...
       @JsonProperty("podTemplateSelectStrategy") PodTemplateSelectStrategy podTemplateSelectStrategy,
+      @Nullable
       @JsonProperty("capacity") Integer capacity
   )
   {
     this.podTemplateSelectStrategy = podTemplateSelectStrategy;
     this.capacity = capacity;
   }
 
+  @Nullable
```
FrankChen021, line 48: [P1] Duplicate @Nullable annotations prevent compilation

The merge result places one `@Nullable` before `@Override` and another after `@JsonProperty` on both getters. `javax.annotation.Nullable` is not a repeatable annotation, so javac rejects this class with `Nullable is not a repeatable annotation type` before the module can build or test. Keep only one annotation on each getter (the second duplicate is at line 57).

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/execution/KubernetesTaskExecutionConfigResource.java
```diff
@@ -57,16 +57,18 @@ public class KubernetesTaskExecutionConfigResource
...
   private final AuditManager auditManager;
-  private AtomicReference<KubernetesTaskRunnerDynamicConfig> dynamicConfigRef = null;
+  private final KubernetesTaskRunnerEffectiveConfig effectiveConfig;
 
   @Inject
   public KubernetesTaskExecutionConfigResource(
       final JacksonConfigManager configManager,
-      final AuditManager auditManager
+      final AuditManager auditManager,
+      final KubernetesTaskRunnerEffectiveConfig effectiveConfig
```
FrankChen021, line 66: [P2] Avoid requiring single-cluster config for multik8s

Injecting `KubernetesTaskRunnerEffectiveConfig` makes this resource resolve the module's `KubernetesTaskRunnerStaticConfig`, whose `namespace` field is `@NotNull`. The `multik8s` path instead reads `MultipleKubernetesTaskRunnerConfig` and explicitly allows top-level `druid.indexer.runner.namespace` to be absent when cluster entries provide `taskNamespace` (for example with the custom-template adapter). On that supported configuration, provisioning this resource fails validation before GET/POST can run. Keep the resource dependency runner-agnostic or provide the active multik8s configuration.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

---

Issue #19021, opened 2026-02-13
Allow filtering of metrics in the LoggingEmitter


abhishekrb19:
### Motivation

The LoggingEmitter currently emits _all_ metrics it receives without filtering. On busy clusters with high query throughput, this can result in a very large volume of metric logs - on some of our clusters, we've seen ~3M metric logs in 15 minutes.

### Proposed changes
- Add a file to capture the default metrics for the LoggingEmitter, similar to the Prometheus and Dropwizard emitters (e.g., `defaultMetricDimensions.json`).
- This file can contain the list of allowed metrics.
- In the future, we can extend this file to annotate additional user-defined dimensions, if needed.
- In the LoggingEmitter `emit(Event event)` method, check the allowlist metrics map and decide whether to include the event or not.

### Rationale

Typically, the LoggingEmitter is used in conjunction with a composite emitter that handles low- to medium-cardinality metrics, such as Prometheus, Dropwizard, or Kafka forwarding to a downstream system. Filtering out metrics that are already covered by these emitters would help reduce logging costs while continuing to rely on the appropriate emitter for low- to medium-cardinality metrics.

### Operational impact

None. By default, all metrics will continue to be emitted for backwards compatibility with a runtime property. Operators can opt-in to enable filtering with the property.

jtuglu1:
SGTM – we also have this issue. I've seen 10M+ in 15mins 🙃. 

Edit: I think there's also chance we should look at removing some metrics that IMO could be turned into (debug) logs.

jtuglu1:
As mentioned [here](https://github.com/apache/druid/pull/18966#discussion_r2761898386), I think there's also opportunity to just completely remove unused/unuseful metrics and replace them with (debug) logs.

PR #19026, opened 2026-02-17
dev: Unified metric filtering logic across emitters

nozjkoitop:
Extends #19030.

### Description

This PR expands metric-name-based filtering beyond `LoggingEmitter` and introduces shared filtering support that can be reused across multiple emitters.

### What changed

- Added shared emitter metric-filtering config:
  - `shouldFilterMetrics`
  - `metricSpecPath`
- Added common filtering support in `processing`:
  - `AbstractFilteringEmitter`
  - shared allowlist loading/parsing utilities
- Updated `LoggingEmitter`,  `HttpPostEmitter`,  `ParametrizedUriEmitter` and  `PrometheusEmitter` to use the shared filtering mechanism

### Behavior

- filtering is optional and disabled by default
- when filtering is enabled, metric events are emitted only if their metric name is present in the configured allowlist
- when no custom spec path is configured, the emitter uses its bundled default metric spec
- when a custom spec path is configured, that file is used as the allowlist source
- non-metric event handling remains emitter-specific
  
#### Release note
Added shared metric allowlist filtering support for multiple emitters. `LoggingEmitter`, `HttpPostEmitter`, `ParametrizedUriEmitter`, and `PrometheusEmitter` can optionally restrict emitted metric events to an allowlisted set of metric names using the shared filtering configuration.

This PR has:

- [X] been self-reviewed.


abhishekrb19:
Thanks @nozjkoitop for the contribution! Cross-linking the comment here on a similar implementation - https://github.com/apache/druid/pull/19030#pullrequestreview-3822957889

nozjkoitop:
@abhishekrb19 quick ping, could you take a look at the latest updates here when you have a moment? I incorporated the schema + test coverage deltas discussed on #19030 (along with your comment from the review), so we can move this feature forward

abhishekrb19:
@nozjkoitop I will find time later this week to review these changes alongside https://github.com/apache/druid/pull/19030. Thanks!

jtuglu1:
Could we make this functionality generic to all types of emitters?

nozjkoitop:
yea, lemme try it out
perhaps i can add a shared util to the emitters

nozjkoitop:
added couple of emitters as a concept, @jtuglu1 wdyt?

jtuglu1:
cc @maytasm

abhishekrb19:
> Could we make this functionality generic to all types of emitters?


Agreed, this will be broadly useful as a core concept for all emitters. Can we do this as a separate change though and scope this PR down to only add the functionality for the LoggingEmitter as mentioned in https://github.com/apache/druid/issues/19021?

Existing emitters like prometheus-emitter, dropwizard-emitter, opentsdb-emitter, etc. already have a notion of an allow list, so we’ll need to be careful not to break compatibility with them. IMO it will be better to come up with a common interface and enable it for all the emitters in a backwards-compatible manner to existing emitters in a separate change (or alternatively define the interface in this patch, but only integrate it with the LoggingEmitter, but I'd prefer that we do this separately too). Please let me know what you think @nozjkoitop @jtuglu1.

processing/src/main/java/org/apache/druid/java/util/emitter/core/MetricAllowlistParser.java
```diff
@@ -0,0 +1,29 @@
...
+package org.apache.druid.java.util.emitter.core;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import java.util.Set;
+
+public interface MetricAllowlistParser
+{
+  Set<String> parse(JsonNode metricConfig, String source);
+}
```
abhishekrb19, line 42: Please add javadocs for common interfaces.

jtuglu1:
> Existing emitters like prometheus-emitter, dropwizard-emitter, opentsdb-emitter, etc. already have a notion of an allow list, so we’ll need to be careful not to break compatibility with them.

Yes, and, ideally this can be extended so any new/existing emitter can "plugin" to the existing mechanism.

> or alternatively define the interface in this patch, but only integrate it with the LoggingEmitter

I'd prefer this if possible, since we'd just be deferring the unavoidable unification of all these emitters' filtering mechanisms to down-the-line and I don't want to keep adding more risk that these various different emitter filtering mechanisms become too disparate and a "clean" interface is hard to do.

If possible, I'd like to define it in this PR (and let users of other emitters "opt-in" as necessary via changes down-the-line).

processing/src/main/java/org/apache/druid/java/util/emitter/core/GlobalEmitterConfig.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Optional;
+
+public class GlobalEmitterConfig
+{
+  @JsonProperty
+  boolean filterMetrics;
+
```
abhishekrb19, line 40: Please add javadocs

processing/src/main/java/org/apache/druid/java/util/emitter/core/GlobalEmitterConfig.java
```diff
@@ -0,0 +1,43 @@
...
+  @JsonProperty
+  String metricAllowlistPath;
+
```
abhishekrb19, line 60: These properties can be private-scoped

processing/src/main/java/org/apache/druid/java/util/emitter/core/MetricAllowlistLoader.java
```diff
@@ -0,0 +1,68 @@
...
+      ObjectMapper mapper,
+      String allowlistPath,
+      MetricAllowlistParser parser
+  )
+  {
+    try (final InputStream is = openAllowlistFile(allowlistPath)) {
+      return parser.parse(mapper.readTree(is), allowlistPath);
+    }
+    catch (IOException e) {
+      throw new ISE(e, "Failed to parse metric allowlist file [%s]", allowlistPath);
```
abhishekrb19: Please use DruidException - style guide ref: https://github.com/apache/druid/blob/master/dev/style-conventions.md#message-formatting-for-logs-and-exceptions

processing/src/main/java/org/apache/druid/java/util/emitter/core/MetricAllowlistLoader.java
```diff
@@ -0,0 +1,68 @@
...
+    try {
+      return new FileInputStream(allowlistPath);
+    }
+    catch (FileNotFoundException e) {
+      final InputStream classpathInputStream = MetricAllowlistLoader.class.getClassLoader().getResourceAsStream(allowlistPath);
+      if (classpathInputStream != null) {
+        return classpathInputStream;
+      }
+      throw new IAE(e, "Metric allowlist file [%s] not found", allowlistPath);
+    }
```
abhishekrb19: DruidExceptions here as well similar to https://github.com/apache/druid/pull/19030/:
 https://github.com/apache/druid/pull/19030/changes#diff-e6a24d753a6939a848cd7909791a6ceaf358cb5976e7b3ed084912204778cb63R137-R145

processing/src/main/java/org/apache/druid/java/util/emitter/core/MetricAllowlistParsers.java
```diff
@@ -0,0 +1,40 @@
...
+
+public final class MetricAllowlistParsers
+{
+
+  public static Set<String> parseMetricNameObject(JsonNode metricConfig, String source)
+  {
+    if (!metricConfig.isObject()) {
+      throw new ISE("Metric allowlist file [%s] must be a JSON object of metric names", source);
+    }
+    final ImmutableSet.Builder<String> metricNames = ImmutableSet.builder();
```
abhishekrb19, line 38: Does this work with the different defaultMetrics.json formats that emitters support (they look close enough but have slight differences). Could you also update some of the existing emitters like prometheus-emitter to exercise these code paths?

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
```diff
@@ -145,6 +167,9 @@ public void emit(Event event)
   private void emitMetric(ServiceMetricEvent metricEvent)
   {
     String name = metricEvent.getMetric();
+    if (shouldFilterOutMetric(name)) {
+      return;
+    }
```
abhishekrb19: Hmm, instead of doing this for every single emitter, what do you think about adding an abstract class that implements `MetricFilteringEmitter implements Emitter` and overrides `emit()` ? This class can have awareness about global configs instantiated from the specific sub-class Emitters like PrometheusEmitter, LoggingEmitter, etc and delegate the filtering to it if the config is enabled or just fall back to current behavior? I will think about this more. @jtuglu1 - do you have suggestions here?

abhishekrb19 (commented):
@nozjkoitop, thanks for the changes! I left some high-level comments on the common config that would apply to all emitters.

Since the scope of this PR has expanded and we’ve been porting some of the logging-emitter changes from https://github.com/apache/druid/pull/19030, I’ll go ahead and merge 19030 to keep the drift minimal and make things easier to review. Then we can repurpose this PR to extend the filtering functionality broadly across all emitter types, as @jtuglu1 suggested. Please let me know what you think.

maytasm:
Since @abhishekrb19 merged https://github.com/apache/druid/pull/19030, this PR will have to be rebased. There are also duplicate configs between the two PRs too. @nozjkoitop please comment here when the PR is ready for review. Thanks!

processing/src/main/java/org/apache/druid/java/util/emitter/core/MetricAllowlistParsers.java
nozjkoitop: It does work with the json objects, as long as the metric names are the field names 
In case of other structure there is an abstract getMetricAllowlistParser where custom parser can be provided

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/PrometheusEmitter.java
nozjkoitop: added an AbstractFilteringEmitter with shouldFilterMetrics and allowedMetricNames

nozjkoitop:
@maytasm latest master is merged, please take a look when you have some time

nozjkoitop:
Hi @abhishekrb19, following up on this PR. It has been rebased and updated based on the latest review comments, so I’d appreciate another look when you get a chance.

processing/src/main/java/org/apache/druid/java/util/emitter/core/Emitters.java
```diff
@@ -145,9 +157,9 @@ static Map<String, Object> makeLoggingMap(Properties props)
           "shouldFilterMetrics", Boolean.parseBoolean(props.getProperty("org.apache.druid.java.util.emitter.logging.shouldFilterMetrics"))
       );
     }
-    if (props.containsKey("org.apache.druid.java.util.emitter.logging.allowedMetricsPath")) {
+    if (props.containsKey("org.apache.druid.java.util.emitter.logging.metricSpecPath")) {
```
FrankChen021, line 160: P2 Keep the old logging allowlist property as an alias

This drops support for org.apache.druid.java.util.emitter.logging.allowedMetricsPath, which was the documented property before this PR. Druid's DefaultObjectMapper ignores unknown properties, so existing configs with shouldFilterMetrics=true and the old path will silently ignore the custom allowlist and fall back to defaultMetrics.json, filtering out metrics operators expected to keep. Please accept the old property as a deprecated alias or migrate it explicitly.

processing/src/main/java/org/apache/druid/java/util/emitter/core/GlobalEmitterConfig.java
```diff
@@ -0,0 +1,70 @@
...
+   */
+  @JsonProperty
+  private String metricSpecPath;
+
+  public boolean isShouldFilterMetrics()
+  {
+    return shouldFilterMetrics;
+  }
+
+  public Optional<String> getMetricSpecPath()
```
FrankChen021, line 56: P2 Treat empty metricSpecPath as absent

The docs and class comment say null or empty metricSpecPath should use the bundled default, but this getter returns Optional.of(""). HTTP, parametrized URI, and Prometheus filtering then call loadAllowlistFromFile(""), where validation throws during emitter construction instead of loading the default. Normalize empty strings to Optional.empty() here, or before calling loadAllowedMetricNames, and add coverage for an empty configured path.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 3 |
| P2 | 2 |
| P3 | 0 |
| Total | 5 |

Reviewed 30 of 30 changed files.

Static review only; no build or test commands were run.



---

processing/src/main/java/org/apache/druid/java/util/emitter/core/MetricAllowlistLoader.java
```diff
@@ -0,0 +1,104 @@
...
+
+  public static Set<String> loadAllowlistFromClasspath(
+      final ObjectMapper mapper,
+      final String resourcePath,
+      final MetricAllowlistParser parser
+  )
+  {
+    validateAllowlistPath(resourcePath);
+
+    final InputStream classpathInputStream = MetricAllowlistLoader.class.getClassLoader().getResourceAsStream(resourcePath);
```
FrankChen021, line 69: [P1] Load extension resources with the extension classloader

`MetricAllowlistLoader` is part of the core `processing` jar, so `MetricAllowlistLoader.class.getClassLoader()` is the Druid/core classloader. The default Prometheus spec, however, is packaged only in the separately loaded `prometheus-emitter` extension jar. With the normal external-extension classloader setup, enabling filtering without a custom path makes this lookup return null and throws `NOT_FOUND` during `PrometheusEmitter` construction, so the Prometheus emitter cannot start. Please load this resource through the Prometheus/context classloader or pass the resource-owning classloader into the shared loader.

extensions-contrib/prometheus-emitter/src/main/java/org/apache/druid/emitter/prometheus/Metrics.java
```diff
@@ -66,7 +66,7 @@ public DimensionsAndCollector getByName(String name, String service)
   public Metrics(PrometheusEmitterConfig config)
   {
     String namespace = config.getNamespace();
-    String path = config.getDimensionMapPath();
+    String path = config.getMetricSpecPath().orElse(config.getDimensionMapPath());
```
FrankChen021, line 69: [P1] Keep the filter allowlist separate from the Prometheus mapping

`metricSpecPath` is the shared filter-only allowlist path (the documented shape permits values such as `[]`), while `dimensionMapPath` contains Prometheus `Metric` definitions. Preferring `metricSpecPath` here unconditionally makes a custom allowlist override a valid dimension map; `readConfig` then tries to deserialize entries like `"query/time": []` as `Metric` objects and the emitter construction fails (or registers no collectors for object values). This also changes behavior when filtering is false, even though the shared config says the path is only used for filtering. Keep `Metrics` on `dimensionMapPath`/its default and use `metricSpecPath` only for the allowlist.

processing/src/main/resources/defaultMetrics.json
```diff
@@ -94,56 +111,34 @@
...
-    "query/cache/total/hitRate": [],
-    "query/cache/total/hits": [],
-    "query/cache/total/misses": [],
-    "query/cache/total/numEntries": [],
-    "query/cache/total/put/error": [],
-    "query/cache/total/put/ok": [],
-    "query/cache/total/put/oversized": [],
-    "query/cache/total/sizeBytes": [],
-    "query/cache/total/timeouts": [],
+    "query/cache/memcached/delta": [],
```
FrankChen021, line 118: [P1] Preserve active metrics in the default allowlist

When filtering is enabled without a custom spec, the new shared emitters use this file. This change removes the existing `query/cache/delta/*` and `query/cache/total/*` entries, all `kafka/consumer/*` entries, and `service/heartbeat`, even though `CacheMonitor`, `KafkaConsumerMonitor`, and `ServiceStatusMonitor` still emit those names (and they were present in the previous default). Those events will now be silently dropped by the logging, HTTP, and parametrized emitters for users relying on the default allowlist. Please retain the active previously allowed metrics or otherwise preserve this compatibility.

processing/src/main/java/org/apache/druid/java/util/emitter/core/GlobalEmitterConfig.java
```diff
@@ -0,0 +1,70 @@
...
+  {
+    return Optional.ofNullable(metricSpecPath);
```
FrankChen021, line 58: [P2] Treat an empty metric path as absent

The config Javadocs and the logging documentation promise that a null or empty `metricSpecPath` selects the bundled default. `Optional.ofNullable("")` is present, so `AbstractFilteringEmitter.loadAllowedMetricNames` sends the empty string to `loadAllowlistFromFile`, whose validation rejects it. Consequently, enabling filtering with `metricSpecPath: ""` fails construction for the HTTP, parametrized, and Prometheus emitters (only LoggingEmitter currently normalizes it). Normalize empty values in the shared getter/helper.

processing/src/main/java/org/apache/druid/java/util/emitter/core/Emitters.java
FrankChen021, line 160: [P2] Preserve the legacy logging path alias

This replaces the existing `org.apache.druid.java.util.emitter.logging.allowedMetricsPath` property with the new name without translating the old key. A deployment using the previously supported `druid.emitter.logging.allowedMetricsPath` (or the legacy `org.apache...logging.allowedMetricsPath`) will still have filtering enabled but its custom file will be omitted, causing a silent fallback to the bundled allowlist; the old `LoggingEmitterConfig` getter is removed as well. Please accept the old property as a deprecated alias and map it to `metricSpecPath`.

PR #19028, opened 2026-02-18
Cost-based autoscaler: allow scaling action only when metrics are available

Fly-Style:
Default processing rate was a crutch that allowed autoscaler to start scaling actions earlier than the first metrics became available. In fact, it is confusing parameter and actually a crutch, which was removed in that patch.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
@@ -151,6 +151,11 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
   public int computeTaskCountForScaleAction()
   {
     lastKnownMetrics = collectMetrics();
+    if (lastKnownMetrics == null) {
+      log.debug("Metrics not available for supervisorId [%s], skipping scaling action", supervisorId);
+      return -1;
+    }
+
     final int optimalTaskCount = computeOptimalTaskCount(lastKnownMetrics);
     final int currentTaskCount = lastKnownMetrics.getCurrentTaskCount();
 
--- a/indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerTest.java
@@ -469,7 +469,7 @@ public class CostBasedAutoScalerTest
   }
 
   @Test
-  public void testCollectMetricsReturnsNullWhenMovingAvgRateNegative()
+  public void testScalingActionSkippedWhenMovingAverageRateUnavailable()
   {
     SupervisorSpec spec = Mockito.mock(SupervisorSpec.class);
     SeekableStreamSupervisor supervisor = Mockito.mock(SeekableStreamSupervisor.class);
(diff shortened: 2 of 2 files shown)


indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerTest.java
```diff
@@ -454,29 +468,40 @@ public void testComputeTaskCountForRolloverAndConfigProperties()
...
   }
 
-  private CostMetrics createMetrics(
-      double avgPartitionLag,
-      int currentTaskCount,
-      int partitionCount,
-      double pollIdleRatio
-  )
+  @Test
+  public void testCollectMetricsReturnsNullWhenMovingAvgRateNegative()
```
kfaraz: Instead of this, the test should verify that scaling action is skipped when moving average rate is not available. The test should care less about the internal implementation details of the target and more about its overall behaviour.

kfaraz (commented):
Minor suggestion to improve test behaviour.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerTest.java
Fly-Style: This test was added mostly to satisfy jacoco test coverage :)

kfaraz: 😛  yay, jacoco!

Let's take the opportunity to make it a useful test then. 🙂

Fly-Style: Done!

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -150,6 +151,11 @@ public int computeTaskCountForRollover()
   public int computeTaskCountForScaleAction()
   {
     lastKnownMetrics = collectMetrics();
+    if (lastKnownMetrics == null) {
```
kfaraz, line 154: good catch!

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -448,7 +455,8 @@ static double extractMovingAverage(Map<String, Map<String, Object>> taskStats)
     return count > 0 ? sum / count : -1;
   }
 
-  private CostMetrics collectMetrics()
+  @Nullable
+  CostMetrics collectMetrics()
```
kfaraz: I don't think this needs to be visible anymore.

```suggestion
  private CostMetrics collectMetrics()
```

Fly-Style: @kfaraz unfortunately, not.

kfaraz:
Merged the PR since the failure in the docker tests is unrelated and has been seen on other PRs.
Attempt to fix the docker tests here #18849

PR #19030, opened 2026-02-18
add defaultMetrics for logging emitter

sarangv:
Fixes [19021](https://github.com/apache/druid/issues/19021).

### Description

`LoggingEmitter` can restrict which metrics are written to the log. When this behavior is turned on, only `ServiceMetricEvent` metrics that appear in a configured permitted set are emitted; alerts and other non-metric events are never dropped. When the feature is off, the emitter behaves as before (all events logged).

#### Runtime properties

- **`druid.emitter.logging.shouldFilterMetrics`** (default `false`): If `true`, logging is limited to metrics named in the permitted set. If `false`, every event is logged.
- **`druid.emitter.logging.allowedMetricsPath`** (optional): File path for the permitted-metrics definition. Omitted or empty → the in-JAR resource `defaultMetrics.json` is used. If a path is given but the file does not exist, a warning is logged and the in-JAR resource is used instead (startup continues; operators can spot the warning).

#### Permitted-metrics file shape

The file must be a **JSON object** whose keys are metric names; values are placeholders (e.g. `[]`), e.g. `{"query/time": [], "query/bytes": []}`. This shape matches other emitter default configs (e.g. Prometheus) and leaves room for per-metric metadata later. On invalid content, operators see: *"Allowed metrics file must be a JSON object with metric names as keys"*.

#### Rationale

- **DruidException**: Errors use `DruidException.forPersona(Persona.OPERATOR)` with `NOT_FOUND` (default resource absent) and `RUNTIME_FAILURE` (parse/read failure), matching the project’s guidance for operator-facing failures.
- **Missing custom file**: Rather than failing startup, the code logs a WARN and uses the in-JAR default so a wrong path does not prevent the process from starting; the log still exposes the mistake.
- **Object format**: A key-value object was chosen over a JSON array so the schema can later carry extra fields per metric without a breaking change.
- **Emitters.java**: The two new keys are written into the legacy logging map only when the matching property exists, preserving Java defaults and avoiding odd empty-string behavior.
- **No set when off**: With filtering disabled, `allowedMetrics` stays `null`; no extra collection is created. Contains checks stay O(1).

#### Release note

Operators can limit which metrics the logging emitter writes by setting `druid.emitter.logging.shouldFilterMetrics=true` and, if desired, `druid.emitter.logging.allowedMetricsPath` to a JSON object file (keys = metric names). A missing custom file results in a warning and use of the bundled `defaultMetrics.json`. Alerts and other non-metric events are always logged.

<hr>

- `LoggingEmitter` – load permitted set (classpath or file; use default if file missing), apply in `emit()`, DruidException on bad/missing config
- `LoggingEmitterConfig` – `shouldFilterMetrics`, `allowedMetricsPath` plus Javadoc
- `Emitters` – map the new keys from legacy properties only when set
- `processing/src/main/resources/defaultMetrics.json` – in-JAR permitted set (object form)
- `processing/src/test/resources/defaultMetrics.json` – test resource
- `LoggingEmitterTest` – seven cases: feature off, default resource, custom file, missing file uses default, non-metrics pass through, empty set, path ignored when off; `@After` tearDown for close
- `LoggingEmitterConfigTest` – new config fields covered for defaults and legacy mapping

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -2019,7 +2019,7 @@ log4j config to route these logs to different sources based on the feed of the e
 |--------|-----------|--------|
 |`druid.emitter.logging.loggerClass`|The class used for logging.|`org.apache.druid.java.util.emitter.core.LoggingEmitter`|
 |`druid.emitter.logging.logLevel`|Choices: debug, info, warn, error. The log level at which message are logged.|info|
-|`druid.emitter.logging.shouldFilterMetrics`|When true, only metrics listed in the allowlist are emitted; non-metric events (e.g. alerts) are always emitted. When false, all events are logged (backward-compatible).|false|
+|`druid.emitter.logging.shouldFilterMetrics`|When true, only metrics listed in the allow list are emitted; non-metric events (e.g. alerts) are always emitted. When false, all events are logged (backward-compatible).|false|
 |`druid.emitter.logging.allowedMetricsPath`|Path to a JSON file whose keys are the allowed metric names. Only used when `shouldFilterMetrics` is true. If null or empty, the bundled classpath resource `defaultMetrics.json` is used. If a path is set but the file is missing, a warning is logged and the emitter falls back to the default classpath resource.|null|
 
 #### HTTP emitter module


docs/configuration/index.md
```diff
@@ -2019,6 +2019,8 @@ log4j config to route these logs to different sources based on the feed of the e
 |--------|-----------|--------|
 |`druid.emitter.logging.loggerClass`|The class used for logging.|`org.apache.druid.java.util.emitter.core.LoggingEmitter`|
 |`druid.emitter.logging.logLevel`|Choices: debug, info, warn, error. The log level at which message are logged.|info|
+|`druid.emitter.logging.shouldFilterMetrics`|When true, only metrics listed in the allowlist are emitted; non-metric events (e.g. alerts) are always emitted. When false, all events are logged (backward-compatible).|false|
+|`druid.emitter.logging.allowedMetricsPath`|Path to a JSON file whose keys are the allowed metric names. Only used when `shouldFilterMetrics` is true. If null or empty, the bundled classpath resource `defaultMetrics.json` is used. If a path is set but the file is missing, a warning is logged and the emitter falls back to the default classpath resource.|null|
```
abhishekrb19, line 2023: >If a path is set but the file is missing, a warning is logged and the emitter falls back to the default classpath resource.

IMO, in this case, we should fail fast by throwing an exception instead of falling back in the event of a misconfiguration, since an operator would be intentionally opting in to enable metric filtering and provide a custom path

nozjkoitop:
Sure thing! I updated #19026 to match the deltas noted here (more test coverage and extensible default-metrics schema). @sarangv could you take a quick look at the updated #19026 and leave any review notes there? I’d prefer we keep review in a single PR to avoid duplicating effort

docs/configuration/index.md
```diff
@@ -2019,6 +2019,8 @@ log4j config to route these logs to different sources based on the feed of the e
 |--------|-----------|--------|
 |`druid.emitter.logging.loggerClass`|The class used for logging.|`org.apache.druid.java.util.emitter.core.LoggingEmitter`|
 |`druid.emitter.logging.logLevel`|Choices: debug, info, warn, error. The log level at which message are logged.|info|
+|`druid.emitter.logging.shouldFilterMetrics`|When true, only metrics listed in the allowlist are emitted; non-metric events (e.g. alerts) are always emitted. When false, all events are logged (backward-compatible).|false|
```
abhishekrb19: To fix the spell check failure:
```suggestion
|`druid.emitter.logging.shouldFilterMetrics`|When true, only metrics listed in the allow list are emitted; non-metric events (e.g. alerts) are always emitted. When false, all events are logged (backward-compatible).|false|
```

PR #19034, opened 2026-02-19
Fix recurring bug "Inconsistency between stored metadata" during auto-scaling

kfaraz:
### Description

During aggressive auto-scaling, the tasks frequently fail with the error "Inconsistency between stored metadata and target state" causing ingestion lag. This is typically a self-healing issue as the supervisor re-launches the failed tasks with updated offsets, but it is still operational overhead and often causes ingestion lag.

```java
java.util.concurrent.ExecutionException: org.apache.druid.java.util.common.ISE:
  Failed to publish segments because of
[Inconsistency between stored metadata state[KafkaDataSourceMetadata{}] and target state[KafkaDataSourceMetadata{}].
```

The root cause behind this failure seems to be the following race condition:

- Scaling event is triggered.
- `changeTaskCount()` is called
- `checkTaskDuration()` tries to checkpoint the actively reading tasks and moves them to pending completion
- `checkTaskDuration()` also updates the `partitionOffsets` with the latest result of the checkpointing
- ⚠️ `clearAllocationInfo()` clears `partitionOffsets`
- New task group B is created and is assigned a partition P1 which an old task group (still pending completion) was also reading from.
- ⚠️ (race) Task group B is initialized with offsets present in the metadata store. But this does not reflect the latest checkpoint since task group A is yet to publish.
- Task group A publishes offsets and updates the metadata store.
- ❌ Task group B tries to publish and fails since the committed offsets have now diverged.

The bug does not occur if task group A is able to finish publishing the offsets before task group B has been created.

### Changes

- Do not clear `partitionOffsets` before auto-scaling so that subsequent tasks know where the previous tasks had left off.
- Simplify the condition in `IndexerSQLMetadataStorageCoordinator`
- Add some comments and javadocs
- Update error message to be more user-friendly
- Fixed `KafkaSupervisorTest` to not use reflection and updated test to verify clearing of `partitionOffsets`

### Note

This bug may still occur if Overlord leadership changes right before the scaling event.
But there is currently no way to handle that since `partitionOffsets` is an in-memory data structure and is not meant to be persisted.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/test/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinatorTest.java
+++ b/server/src/test/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinatorTest.java
@@ -948,7 +948,7 @@ public class IndexerSQLMetadataStorageCoordinatorTest extends IndexerSqlMetadata
             "Stored metadata state[ObjectMetadata{theObject={foo=baz}}] has already been updated by other tasks"
             + " and has diverged from the expected start metadata state[ObjectMetadata{theObject=null}]."
             + " This task will be replaced by the supervisor with a new task using updated start offsets."
-            + " Reset the supervisor if the issue persists."
+            + " Try resetting the supervisor if the issue persists."
         ),
         result2
     );
@@ -1098,7 +1098,7 @@ public class IndexerSQLMetadataStorageCoordinatorTest extends IndexerSqlMetadata
             "Stored metadata state[ObjectMetadata{theObject={foo=baz}}] has already been updated by other tasks"
             + " and has diverged from the expected start metadata state[ObjectMetadata{theObject={foo=qux}}]."
             + " This task will be replaced by the supervisor with a new task using updated start offsets."
-            + " Reset the supervisor if the issue persists."
+            + " Try resetting the supervisor if the issue persists."
         ),
         result2
     );


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -621,18 +621,30 @@ private void changeTaskCountInIOConfig(int desiredActiveTaskCount)
...
    * tasks are publishing and auto-scaler does not repeatedly attempt a scale down until these tasks
    * complete. If this is cleared, the next {@link #discoverTasks()} might add these tasks to
    * {@link #activelyReadingTaskGroups}.
+   * <p>
+   * Also does not clear {@link #partitionOffsets} so that the new tasks remember
+   * where the previous tasks had left off.
+   * <p>
+   * Since both of these are in-memory structures, a change in Overlord leadership
+   * might cause duplicate scaling actions and/or intermittent task failures due
+   * to {@code "Inconsistency between stored metadata and target"}.
```
Fly-Style: It already has a new exception description :)

Fly-Style (approved):
Looks good to me!

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -2319,12 +2321,15 @@ protected SegmentPublishResult updateDataSourceMetadataInTransaction(
...
+    } else {
+      // startMetadata is older than committed metadata
+      // The task trying to publish is probably a replica trying to commit offsets already published by another task.
+      // OR the metadata has been updated manually
       return SegmentPublishResult.fail(
-          "Inconsistency between stored metadata state[%s] and target state[%s]. Try resetting the supervisor.",
+          "Stored metadata state[%s] has already been updated by other tasks and"
+          + " has diverged from the expected start metadata state[%s]."
+          + " This task will be replaced by the supervisor with a new task using updated start offsets."
+          + " Reset the supervisor if the issue persists.",
```
capistrant: nit: I wonder if we are better off suggesting the idea of resetting versus directing to do it. The new message is much better than the old though since you offer some reasoning and mention resetting only if it persists rather than the old message that just says to try resetting it 😜. same apples to all the error messages in this file and is really just semantics. Trying to reduce the Druid liability (not that there is such a thing in this sense) if an operator gets mad cuz they reset without knowing what it does and blames the error message

kfaraz: Thanks for calling this out. Yeah, I guess it's safer to continue with the suggestive tone for the time being. 😛

kfaraz:
Thanks a lot for the reviews, @Fly-Style , @capistrant !

kfaraz:
Merging off as failure is unrelated.

PR #19040, opened 2026-02-20
Extend server priorities to realtime indexing tasks for query isolation

### LINKED ISSUE #19018: Extend server priorities to realtime indexing tasks for query isolation
### Motivation

We have high throughput and high QPM in some environments and we use query prioritization strategy to prioritize some query workloads over others. However, the query prioritization and  [routing](https://druid.apache.org/docs/latest/configuration/#query-routing) strategy is currently only configurable on the Brokers/Historical servers. 

Peons launch with a default server priority of 0, meaning all peon servers are eligible to serve all queries, regardless of the strategy defined on the Brokers. As a result, workload isolation doesn't apply here since real-time servers aren't segmented by priority in the same way as Historicals and Brokers.

### Proposed changes

To enable priority-aware replica allocation for Peon servers, we can introduce a new supervisor spec property, `serverPriorityToReplica`. This allows the existing Broker-side prioritization strategy to extend naturally to Peons without additional architectural changes.

For example, `serverPriorityToReplica: {"1": 2, "0": 1}` will launch 2 CliPeon servers with priority 1 and 1 CliPeon server with priority 0.

In this setup:

- Tasks with `serverPriority = 1` would have higher fault tolerance and availability
- Tasks with `serverPriority = 0` would have a single replica and lower fault tolerance

This will work with the following Broker configurations and provide isolation:
```properties
druid.broker.select.tier=custom
druid.broker.select.tier.custom.priorities=[0]
```

or a different isolated broker that will route queries to Peons with priority 1:

```properties
druid.broker.select.tier=custom
druid.broker.select.tier.custom.priorities=[1]
```

### Operational impact

None. By default, this property when unspecified will be null. Operators using the existing `replicas` configuration will continue to work as before, with all replicas assigned to the default server priority 0.

abhishekrb19:
Fixes https://github.com/apache/druid/issues/19018

Currently all task replicas by default answer all realtime queries regardless of the priority or [routing](https://druid.apache.org/docs/latest/configuration/#query-routing) strategy applied on the Broker.
For mixed query workloads, some queries may have higher priority than others, so isolation on the replicas helps prevent degradation for critical workloads. We've seen bad queries take down some task replicas causing noisy neighbor problems. 

This extends Druid’s query prioritization and [routing](https://druid.apache.org/docs/latest/configuration/#query-routing) strategies for Peon servers by letting operators configure how many replicas per server priority to configure for the tasks, similar to how it works for Historicals and Brokers.  This is done by exposing `serverPriorityToReplics` in the supervisor's `ioConfig` that operators can optionally configure with the number of replicas allocated per server priority for realtime indexing tasks.

For example, some replicas can be configured to handle queries of all priorities, while others may only respond to specific priority ranges. This would isolate certain Peon replicas for certain priorities and others for more exploratory / dashboarding usecases.

#### Approach:

To support this, a new `serverPriorityToReplicas` property is added to the supervisor’s `ioConfig`. 
The `SeekableStreamSupervisor` assigns priorities to `SeekableStreamIndexTask`s  as they are created for a group. Similarly they're removed from internal bookkeeping when the tasks terminate. The `ForkingTaskRunner` then passes the appropriate server priority when initializing the Peon server. In the absence of this configuration, Peons continue to run with the default priority 0.

`serverPriorityToReplicas` is optional and is compatible with the existing `replicas` property if specified.  

### Release note

Added  `serverPriorityToReplicas` parameter to the streaming supervisor specs (kafka/kinesis/rabbit). This allows operators to distribute task replicas across different server priorities for realtime indexing tasks. Similar to historical tiering, this enables query isolation for mixed workload scenarios on the Peons, allowing some task replicas to handle queries of specific priorities.

This PR has:

- [x] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -145,6 +162,13 @@ public Integer getReplicas()
     return replicas;
   }
 
+  @Nullable
+  @JsonProperty
+  public Map<Integer, Integer> getserverPriorityToReplicas()
```
aho135:
```suggestion
  public Map<Integer, Integer> getServerPriorityToReplicas()
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -119,6 +121,21 @@ public SeekableStreamSupervisorIOConfig(
     }
 
     this.idleConfig = idleConfig;
+    this.serverPriorityToReplicas = serverPriorityToReplicas;
+    if (this.serverPriorityToReplicas != null) {
+      final int replicaCount = this.serverPriorityToReplicas.values().stream().mapToInt(Integer::intValue).sum();
+      if (replicas != null && replicas != replicaCount) {
+        throw InvalidInput.exception(
```
aho135, line 142: What do you think about dropping this requirement? Seems like additional configuration overhead since we set it explicitly below anyway

abhishekrb19: This would be triggered only if someone unintentionally sets conflicting values that cause ambiguity. I wanted to keep the behavior straightforward for end users rather than document any precedence rules in such cases. I've updated the error message accordingly, please let me know if that makes sense.

docs/ingestion/supervisor.md
```diff
@@ -56,7 +56,7 @@ For configuration properties specific to Kafka and Kinesis, see [Kafka I/O confi
...
 |`autoScalerConfig`|Object|Defines auto scaling behavior for ingestion tasks. See [Task autoscaler](#task-autoscaler) for more information.|No|null|
 |`taskCount`|Integer|The maximum number of reading tasks in a replica set. Multiply `taskCount` and replicas to measure the maximum number of reading tasks. The total number of tasks, reading and publishing, is higher than the maximum number of reading tasks. See [Capacity planning](../ingestion/supervisor.md#capacity-planning) for more details. When `taskCount` is greater than the number of Kafka partitions or Kinesis shards, the actual number of reading tasks is less than the `taskCount` value.|No|1|
-|`replicas`|Integer|The number of replica sets, where 1 is a single set of tasks (no replication). Druid always assigns replicate tasks to different workers to provide resiliency against process failure.|No|1|
+|`replicas`|Integer|The number of replica sets, where 1 is a single set of tasks (no replication). Druid always assigns replicate tasks to different workers to provide resiliency against process failure. See `serverPriorityToReplicas` to control server priorities for task replicas.|No|1|
```
jtuglu1: typo: replicate -> replica

docs/ingestion/supervisor.md
```diff
@@ -65,6 +65,7 @@ For configuration properties specific to Kafka and Kinesis, see [Kafka I/O confi
...
 |`stopTaskCount`|Integer|Limits the number of ingestion tasks Druid can cycle at any given time. If not set, Druid can cycle all tasks at the same time. If set to a value less than `taskCount`, your cluster needs fewer available slots to run the supervisor. You can save costs by scaling down your ingestion tier, but this can lead to slower cycle times and lag. See [`stopTaskCount`](#stoptaskcount) for more information.|No|`taskCount` value|
+|`serverPriorityToReplicas`|Object (`Map<Integer, Integer>`)|Map of server priority to the number of replicas per priority. When set, each task replica is assigned a server priority that corresponds to `druid.server.priority` on the Peon process to enable query isolation for mixed workloads by [query routing strategies](../configuration/index.md#query-routing)|No|null|
```
jtuglu1: Let's give an example here or somewhere else to instruct the user what this might look like. Also helpful for AIs who could be auto-generating these specs.

extensions-contrib/rabbit-stream-indexing-service/src/main/java/org/apache/druid/indexing/rabbitstream/RabbitStreamIndexTask.java
```diff
@@ -67,7 +67,8 @@ public RabbitStreamIndexTask(
         tuningConfig,
         ioConfig,
         context,
-        getFormattedGroupId(Configs.valueOrDefault(supervisorId, dataSchema.getDataSource()), TYPE)
+        getFormattedGroupId(Configs.valueOrDefault(supervisorId, dataSchema.getDataSource()), TYPE),
```
jtuglu1, line 71: Can we add this to the other streaming supervisors too? Ideally we keep these all in parity, especially if it's being added to `SeekableStreamSupervisorIOConfig.java`

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -119,6 +121,22 @@ public SeekableStreamSupervisorIOConfig(
     }
 
     this.idleConfig = idleConfig;
+    this.serverPriorityToReplicas = serverPriorityToReplicas;
+    if (this.serverPriorityToReplicas != null) {
+      final int replicaCount = this.serverPriorityToReplicas.values().stream().mapToInt(Integer::intValue).sum();
```
jtuglu1: can we also validate they must be ≥ 0.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -119,6 +121,22 @@ public SeekableStreamSupervisorIOConfig(
...
+      if (replicas != null && replicas != replicaCount) {
```
jtuglu1: should also mention IMO in the docs how this will scale proportionately with auto-scaling of task count. i.e 3 tasks with 3 replicas on (1:2,10:1) priority would turn into   15 replicas (10 with priority 1, 5 with priority 1) with a base task count of 5.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -145,6 +163,13 @@ public Integer getReplicas()
     return replicas;
   }
 
+  @Nullable
```
jtuglu1, line 180: not sure if we want to leave it nullable or just exclude from the spec if non-existent

jtuglu1 (commented):
Did a first pass – general approach LGTM.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
abhishekrb19: Yes, I’ve added a brief note with an example to clarify this

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
abhishekrb19: I think we can leave this out without `@JsonIgnore` to be consistent with other properties in ioConfig, auto scaler configs, etc

extensions-contrib/rabbit-stream-indexing-service/src/main/java/org/apache/druid/indexing/rabbitstream/RabbitStreamIndexTask.java
abhishekrb19: Agreed. I had left this out to keep the scope minimal. Since the core changes in the base supervisor and task are already in place, we can pass the server priority through instead of null in a few places, and it should work out of the box, similar to Kafka. I can add a comment for those two types and follow up with tests for these changes separately. Would that work?

abhishekrb19: Nvm, I just pushed an update with the wiring for the Kinesis and Rabbit supervisors as well :)

docs/ingestion/supervisor.md
```diff
@@ -65,6 +65,7 @@ For configuration properties specific to Kafka and Kinesis, see [Kafka I/O confi
...
+|`serverPriorityToReplicas`|Object (`Map<Integer, Integer>`)|Map of server priorities to the number of replicas per priority. When set, each task replica is assigned a server priority that corresponds to `druid.server.priority` on the Peon process to enable query isolation for mixed workloads using [query routing strategies](../configuration/index.md#query-routing). If not configured, the `replicas` setting applies and all task replicas are assigned a default priority of 0.<br/><br/>For example, setting `serverPriorityToReplicas` to `{"1": 2, "0": 1}` creates 2 task replicas with `druid.server.priority=1` and 1 task replica with `druid.server.priority=0` per task group. This configuration scales proportionally with `taskCount`. For example, if `taskCount` is set to 5, this results in 15 total task replicas - 10 replicas with priority 1 and 5 replicas with priority 0.|No|null|
```
aho135:
```suggestion
|`serverPriorityToReplicas`|Object (`Map<Integer, Integer>`)|Map of server priorities to the number of replicas per priority. When set, each task replica is assigned a server priority that corresponds to `druid.server.priority` on the Peon process to enable query isolation for mixed workloads using [query routing strategies](../configuration/index.md#query-routing). If not configured, the `replicas` setting applies and all task replicas are assigned a default priority of 0.<br/><br/>For example, setting `serverPriorityToReplicas` to `{"1": 2, "0": 1}` creates 2 task replicas with `druid.server.priority=1` and 1 task replica with `druid.server.priority=0` per task group. This configuration scales proportionally with `taskCount`. For example, if `taskCount` is set to 5, this results in 15 total tasks - 10 tasks with priority 1 and 5 tasks with priority 0.|No|null|
```

I thought using the word `replicas` here was a bit confusing since we're referring to task counts here

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -119,6 +121,28 @@ public SeekableStreamSupervisorIOConfig(
     }
 
     this.idleConfig = idleConfig;
+    this.serverPriorityToReplicas = serverPriorityToReplicas;
+    if (this.serverPriorityToReplicas != null) {
+      final int serverPriorityReplicas = this.serverPriorityToReplicas.values().stream().mapToInt(Integer::intValue).sum();
```
aho135: Something like this would also pass validation
```
Map.of(
  1, 2,
  0, -1
)
```
Maybe make this more strict such that any value has to be > 0

abhishekrb19: Yep, good point. I’ve made the validation check stricter to use >= 0 since the existing `replicas` can be set to 0

docs/ingestion/supervisor.md
```diff
@@ -65,6 +65,7 @@ For configuration properties specific to Kafka and Kinesis, see [Kafka I/O confi
...
+|`serverPriorityToReplicas`|Object (`Map<Integer, Integer>`)|Map of server priorities to the number of replicas per priority. When set, each task replica is assigned a server priority that corresponds to `druid.server.priority` on the Peon process to enable query isolation for mixed workloads using [query routing strategies](../configuration/index.md#query-routing). If not configured, the `replicas` setting applies and all task replicas are assigned a default priority of 0.<br/><br/>For example, setting `serverPriorityToReplicas` to `{"1": 2, "0": 1}` creates 2 task replicas with `druid.server.priority=1` and 1 task replica with `druid.server.priority=0` per task group. This configuration scales proportionally with `taskCount`. For example, if `taskCount` is set to 5, this results in 15 total tasks - 10 tasks with priority 1 and 5 tasks with priority 0.|No|null|
```
aho135:
```suggestion
|`serverPriorityToReplicas`|Object (`Map<Integer, Integer>`)|Map of server priorities to the number of replicas per priority. When set, each task replica is assigned a server priority that corresponds to `druid.server.priority` on the Peon process to enable query isolation for mixed workloads using [query routing strategies](../configuration/index.md#query-routing). If not configured, the `replicas` setting applies and all task replicas are assigned a default priority of 0.<br/><br/>For example, setting `replicas` to 3 and `serverPriorityToReplicas` to `{"1": 2, "0": 1}` creates 2 task replicas with `druid.server.priority=1` and 1 task replica with `druid.server.priority=0` per task group. This configuration scales proportionally with `taskCount`. For example, if `taskCount` is set to 5, this results in 15 total tasks - 10 tasks with priority 1 and 5 tasks with priority 0.|No|null|
```

This helps make it more clear that the sum of Replicas has to equal configured `replicas`

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
```diff
@@ -5244,6 +5272,112 @@ public void test_autoScaler_doesNotRepeatScaleDownActions_ifTasksAreStillPublish
...
+  public void testComputeUnassignedServerPriorities_whenMultipleReplicasPerPriorityIsSet()
+  {
+    final Map<String, Object> consumerProperties = KafkaConsumerConfigs.getConsumerProperties();
+    consumerProperties.put("bootstrap.servers", kafkaHost);
+
+    final KafkaSupervisorIOConfig kafkaSupervisorIOConfig = new KafkaSupervisorIOConfig(
+        topic,
+        null,
+        INPUT_FORMAT,
+        null,
```
aho135, line 5383: Did you want to enforce `replicas = sum of Replicas configured in serverPriorityToReplicas`?

abhishekrb19: Yeah, it's verified in this test `SeekableStreamSupervisorIOConfigTest#testMismatchBetweenReplicasAndServerPriorityReplicasThrowsException()` 

(In terms of structure, I added most tests to the seekable stream package for code coverage, since it serves as the base implementation and added a subset of those tests to the specific supervisor types)

aho135 (approved):

abhishekrb19:
Thanks for the reviews @aho135 and @jtuglu1! Jesse, please feel free to share any feedback even after the PR is merged and I can address it in a follow-up.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -276,6 +277,25 @@ Set<String> taskIds()
...
+      taskIdToServerPriority.remove(taskId);
+    }
+
+    /**
+     * Removes a task from {@link #tasks} using the provided iterator and also removes the corresponding entry
+     * in {@link #taskIdToServerPriority}. Useful when removing tasks during iteration.
+     */
+    void removeTask(Iterator<Entry<String, TaskData>> taskDataIterator, String taskId)
+    {
+      taskDataIterator.remove();
```
jtuglu1, line 298: Do we not need to remove taskId from `tasks`? If not, can we document why not?

jtuglu1 (approved):
LGTM

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
abhishekrb19: This is already an iterator from `tasks.keySet()`, so it does remove the entry from `tasks`. I’ve updated the javadoc to clarify that and added a test as well

PR #19042, opened 2026-02-21
introduce SegmentPruner interface to abstract segment pruning operations

clintropolis:
changes:
* add new `SegmentPruner` interface as a replacement for `DimFilterUtils.filterShards` static method
* `SegmentPruner` interface has a single 'prune' method that takes an `Iterable<T>` and `Function` to extract a `DataSegment` from the iterable objects, returning a `Collection<T>` of the filtered objects
* `DimFilterUtils.filterShards` has been removed, its logic has been migrated to `FilterSegmentPruner` implementation of the new `SegmentPruner` interface
* `TableInputSpec` now has a `SegmentPruner` field instead of separate `filter` and `filterFields` (these fields still remain for compat purposes)
* transition `DartTableInputSpecSlicer`, `IndexerTableInputSpecSlicer`, and `CachingClusteredClient` to use `SegmentPruner` (the former 2 via the `TableInputSpec`)

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/querykit/BaseLeafStageProcessor.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/querykit/BaseLeafStageProcessor.java
@@ -181,10 +181,7 @@ public abstract class BaseLeafStageProcessor extends BasicStageProcessor
   @Override
   public SegmentPruner getPruner(InputSpec inputSpec, int inputNumber)
   {
-    if (query.context().isSecondaryPartitionPruningEnabled()) {
-      return executionVertexSupplier.get().getSegmentPruner();
-    }
-    return null;
+    return executionVertexSupplier.get().getSegmentPruner();
   }
 
   private ProcessorManager<Object, Long> createBaseLeafProcessorManagerWithHandoff(
--- a/processing/src/main/java/org/apache/druid/query/planning/ExecutionVertex.java
+++ b/processing/src/main/java/org/apache/druid/query/planning/ExecutionVertex.java
@@ -217,15 +217,19 @@ public class ExecutionVertex
   @Nullable
   public SegmentPruner getSegmentPruner()
   {
-    final Set<String> baseFields = new HashSet<>();
+    if (!topQuery.context().isSecondaryPartitionPruningEnabled()) {
+      return null;
+    }
     if (topQuery.getFilter() == null) {
       return null;
     }
+    final Set<String> baseFields = new HashSet<>();
     for (final String field : topQuery.getFilter().getRequiredColumns()) {
       if (isBaseColumn(field)) {
         baseFields.add(field);
       }
     }
+
     return new FilterSegmentPruner(
         topQuery.getFilter(),
         baseFields


multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/BasicStageProcessor.java
```diff
@@ -54,4 +56,11 @@ public ExtraInfoHolder makeExtraInfoHolder(@Nullable Object extra)
 
     return NilExtraInfoHolder.instance();
   }
+
+  @Nullable
+  @Override
+  public SegmentPruner getPruner(InputSpec inputSpec, int inputNumber)
```
gianm: IMO this would be better as a default impl on `StageProcessor`.

multi-stage-query/src/main/java/org/apache/druid/msq/exec/StageProcessor.java
```diff
@@ -77,4 +79,7 @@ public interface StageProcessor<R, ExtraInfoType>
    */
   @SuppressWarnings("rawtypes")
   ExtraInfoHolder makeExtraInfoHolder(@Nullable ExtraInfoType extra);
+
+  @Nullable
+  SegmentPruner getPruner(InputSpec inputSpec, int inputNumber);
```
gianm: Javadocs please.

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/processor/SegmentGeneratorStageProcessor.java
```diff
@@ -271,6 +273,13 @@ public ExtraInfoHolder makeExtraInfoHolder(final List<SegmentIdWithShardSpec> se
     return new SegmentGeneratorExtraInfoHolder(segmentIdsWithShardSpecs);
   }
 
+  @Nullable
+  @Override
+  public SegmentPruner getPruner(InputSpec inputSpec, int inputNumber)
```
gianm: Shouldn't be needed once the default impl is moved up to the interface.

multi-stage-query/src/main/java/org/apache/druid/msq/kernel/StageDefinition.java
```diff
@@ -231,6 +232,11 @@ public boolean doesSortDuringShuffle()
     }
   }
 
+  public SegmentPruner getSegmentPruner(int inputNumber)
```
gianm, line 240: Javadoc please.

multi-stage-query/src/main/java/org/apache/druid/msq/querykit/results/ExportResultsStageProcessor.java
```diff
@@ -201,6 +203,13 @@ public ExtraInfoHolder makeExtraInfoHolder(@Nullable Object extra)
     return NilExtraInfoHolder.instance();
   }
 
+  @Nullable
+  @Override
+  public SegmentPruner getPruner(InputSpec inputSpec, int inputNumber)
```
gianm: Should be removable once the default impl moves up.

multi-stage-query/src/main/java/org/apache/druid/msq/exec/std/BasicStageProcessor.java
clintropolis: done

gianm (approved):

PR #19043, opened 2026-02-22
MSQ: Controller result batching, consolidated lifecycle.

gianm:
Main changes:

1) QueryListener is changed to deal in RowsAndColumns rather than
   individual rows.

2) Logic for converting Frames to SQL-formatted rows is moved into
   TaskReportQueryListener, out of QueryResultsReader.

3) DartQueryMaker's streaming mode (fullReport = false) is adjusted
   to use SequenceQueryListener + FrameChannelSequence.

4) Controller register/deregister and execution lifecycle related logic
   is consolidated and moved into ControllerHolder#runAsync, to simplify
   DartQueryMaker.

### DIFF
--- a/processing/src/main/java/org/apache/druid/frame/channel/BlockingQueueFrameChannel.java
+++ b/processing/src/main/java/org/apache/druid/frame/channel/BlockingQueueFrameChannel.java
@@ -173,6 +173,8 @@ public class BlockingQueueFrameChannel
           // If this happens, it's a bug, potentially due to incorrectly using this class with multiple writers.
           throw new ISE("Could not write error to channel");
         }
+
+        notifyReader();
       }
     }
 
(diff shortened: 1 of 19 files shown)


processing/src/main/java/org/apache/druid/frame/channel/BlockingQueueFrameChannel.java
```diff
@@ -173,6 +173,8 @@ public void fail(@Nullable Throwable cause)
           // If this happens, it's a bug, potentially due to incorrectly using this class with multiple writers.
           throw new ISE("Could not write error to channel");
         }
+
+        notifyReader();
```
clintropolis, line 177: hmm, the javadoc on `fail` says that `close` must be called after `fail`, `close` also calls `notifyReader()`, does it matter? I guess not since it clears the future so it would be a no-op or whatever

gianm: I suppose it probably doesn't matter.

gianm:
Something seems wrong with the Docker tests:

```
Error:  org.apache.druid.testing.embedded.k8s.KubernetesTaskRunnerCachingModeDockerTest  Time elapsed: 181.621 s  <<< ERROR!
java.lang.RuntimeException: Failed to execute helm command
Caused by: java.lang.RuntimeException: Helm command failed: repo add datainfra https://charts.datainfra.io/
```

I don't think I can do anything about it in this PR, so will merge this anyway.

PR #19044, opened 2026-02-22
SegementCacheManager: acquireCachedSegment with SegmentId.

gianm:
Prior to this patch, acquireCachedSegment required the entire DataSegment object. This patch loosens the requirement to just the SegmentId.

This is useful under MSQ, where caching is fully on-demand (no load rules) and the full DataSegment object is not necessarily available in the local timeline, even if a segment is cached locally. See the change in RegularLoadableSegment.


clintropolis (approved):

gianm:
The failure was in Docker tests due to [19047](https://github.com/apache/druid/pull/19047), merging anyway.

PR #19045, opened 2026-02-23
Final cost-based autoscaler knobs adjusments

Fly-Style:
This patch makes autoscaling during high lag even smoother, by reducing logarihmic mutliplier used a knob for that.
Old multiplier recommended to scale up too much tasks.
The result might be seen in the plot under the hood.

<details>

Plot conditions were taken from real scenario: ~5M lag, 100k processing rate, 48 partitions (10 and 24 partitions are present here for differential analysis purposes).

image

</details>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
@@ -33,7 +33,7 @@ public class WeightedCostFunction
    * Multiplier for a lag amplification factor; it was carefully chosen
    * during extensive testing as the most balanced multiplier for high-lag recovery.
    */
-  private static final double LAG_AMPLIFICATION_MULTIPLIER = 0.05;
+  static final double LAG_AMPLIFICATION_MULTIPLIER = 0.05;
 
   /**
    * Computes cost for a given task count using compute time metrics.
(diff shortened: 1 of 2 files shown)


kfaraz:
Thanks for the PR, @Fly-Style !

Could you please share details on why the value 0.05 is better than 0.2?
I wonder if the same result cannot be achieved by setting the value of `lagWeight` to a large value, say > 1.

In general, for a new supervisor, what would be the steps to reach the optimal values for `lagWeight` and `idleWeight`?

Fly-Style:
@kfaraz I updated PR description.

Regarding the:

> I wonder if the same result cannot be achieved by setting the value of lagWeight to a large value, say > 1.

Technically, yes, but such a decision does sound logical, at least to my taste. The core idea is behind the juggling of the _normalized_ cost fuction weights. Well, maybe my mind is not so agile 😅, and that may make sense.

kfaraz (approved):
Approved.

@Fly-Style , could you please share some thoughts on:
> In general, for a new supervisor, what would be the steps to reach the optimal values for lagWeight and idleWeight?

It would be very helpful to allow users to experiment with and tune the new autoscaler.

Fly-Style:
@kfaraz sure.
Optimal values for typical smooth workload `lagWeight` / `idleWeight` are in the range of `[0.2/0.8]` to ~`[0.65/0.35]`.
The default values of 0.25 for `lagWeight` and 0.75 for `idleWeight` behaved well both during theoretical analysis and practical long-running tests.

PR #19048, opened 2026-02-24
always use StorageMonitor, also make StorageMonitor work for MSQ tasks

clintropolis:
Makes `StorageMonitor`, added in #18742, be always on. The low overhead seems worth the utility. Also added support for MSQ tasks to work correctly with this monitor, emitting extra `taskId` and `groupId` dimensions.

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/IndexerWorkerContext.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/IndexerWorkerContext.java
@@ -67,7 +67,6 @@ import org.apache.druid.storage.StorageConnectorProvider;
 
 import javax.annotation.Nullable;
 import java.io.File;
-import java.util.Map;
 
 public class IndexerWorkerContext implements WorkerContext
 {


clintropolis:
i think i'd rather do as a follow-up since none of the vsf stuff is documented yet and just get this as part of that. I think its mature enough to start documenting in the next release

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/IndexerWorkerContext.java
```diff
@@ -152,10 +158,12 @@ public static IndexerWorkerContext createProductionInstance(
...
   {
     final IndexIO indexIO = injector.getInstance(IndexIO.class);
-    final SegmentManager segmentManager = new SegmentManager(
+    final SegmentCacheManager cacheManager =
         injector.getInstance(SegmentCacheManagerFactory.class)
-                .manufacturate(new File(toolbox.getIndexingTmpDir(), "segment-fetch"), true)
-    );
+                .manufacturate(new File(toolbox.getIndexingTmpDir(), "segment-fetch"), true);
+    final SegmentManager segmentManager = new SegmentManager(cacheManager);
+    final StorageMonitor storageMonitor = new StorageMonitor(cacheManager, Map.of("taskId", task.getId(), "groupId", task.getGroupId()));
```
gianm: Best to pass in `task::getMetricBuilder` (a supplier of `ServiceMetricEvent.Builder`). It includes `taskId` and `groupId`, also some other stuff like `dataSource`.

clintropolis: ah nice, done :+1:

PR #19050, opened 2026-02-24
deprecate segment lock gran & NumberedOverwriteShardSpec

cecemei:
Deprecate segment lock gran & NumberedOverwriteShardSpec. New features should use TIME_CHUNK lock .

<hr>

This PR has:

- [x] been self-reviewed.


gianm:
Let's do deprecations in separate PRs. I'm certain I like the idea of deprecating segment locking: it never made it out of experimental status, and we have new ways of solving the same problems that it was meant to solve.

I'm not as certain that I like deprecating native compaction right now, although, maybe? Per this comment I do think we should be moving towards MSQ compaction in general: https://github.com/apache/druid/pull/18996#issuecomment-3895961637 so maybe now is a good time to deprecate native. Another possibility is that as a preparatory step we should make MSQ the default (if not already), nudge people towards it in the docs and web console (if not already), etc. I am not sure of the current state of those things.

Couple other things about deprecations:

- Docs need an update as well as code. Anywhere a deprecated feature is mentioned, there should be an infobox (if entire section) or small note (if passing mention) saying that the feature is deprecated.
- PRs that deprecate stuff should be tagged "Release Notes" so we can add deprecations to the notes.

gianm:
In terms of doc updates, there's also:

- references to "segment locking" in `docs/design/storage.md`, `docs/ingestion/tasks.md`
- reference to `druid.indexer.tasklock.forceTimeChunkLock` in `docs/configuration/index.md`

These should be changed from describing the feature as "experimental" to "deprecated".

PR #19051, opened 2026-02-25
feat: Add Supervisor UI component to view timeline of rules applied by a cascading reindexing supervisor

capistrant:
### Description

tied to the effort in https://github.com/apache/druid/issues/19092 to productionalize this new feature

#### New Supervisor API (cascading reindexing supervisors only)

```
GET /druid/indexer/v1/supervisor/{supervisorId}/reindexingTimeline
```

This generates a timeline of search intervals for reindexing with the effective sets of rules that are used to create the underlying inline compaction configs for the different intervals. This is the business end of the console UI described below

#### Console visualization of reindexing timeline (cascading reindexing supervisors only)

For cascading compaction supervisors you can get a timeline view of how rules from the rule provider(s) will apply to your datasource. You can view how skip offsets (if they are configured) will change the timeline too. This visualization is a great help for operators in understanding in how their underlying rules will be applied across the datasource history. I will add some screen shots for reference

image

image

image

image

#### Refactor `CompactionSupervisorTest` Embedded Tests

Split out the cascading template tests to their own class from the inline compaction supervisor tests. Sharing one file was getting convoluted. They share a base class for what makes sense.

#### Try it Yourself

If you build this branch locally and start up a cluster. post the following spec example to supervisors api and you can view a prospective timeline plus play with the config if you want to see how it can change. note that you do not have to pre-ingest the wikipedia data! this supervisor will be suspended and has no reliance on underlying data (caveat is that if you try to do skipOffsetFromLatest and click the dynamic button to calculate, you would need to have the data ingested)

<details>
  <summary>Click to expand code</summary>

  ```JSON
{
  "type": "autocompact",
  "spec": {
    "type": "reindexCascade",
    "dataSource": "wikipedia",
    "taskPriority": 25,
    "inputSegmentSizeBytes": 100000000000000,
    "ruleProvider": {
      "type": "inline",
      "deletionRules": [
        {
          "id": "delete_japan",
          "description": null,
          "olderThan": "P90D",
          "deleteWhere": {
            "type": "equals",
            "column": "countryName",
            "matchValueType": "STRING",
            "matchValue": "Japan"
          },
          "virtualColumns": null
        }
      ],
      "partitioningRules": [],
      "indexSpecRules": [],
      "dataSchemaRules": [
        {
          "id": "dataSchema-1",
          "description": null,
          "olderThan": "P30D",
          "dimensionsSpec": {
            "dimensions": null
          },
          "metricsSpec": [],
          "queryGranularity": "HOUR",
          "rollup": true,
          "projections": []
        }
      ]
    },
    "taskContext": {
      "useConcurrentLocks": true,
      "maxNumTasks": 2
    },
    "skipOffsetFromLatest": null,
    "skipOffsetFromNow": null,
    "defaultSegmentGranularity": "DAY",
    "defaultPartitionsSpec": {
      "type": "range",
      "targetRowsPerSegment": 5,
      "maxRowsPerSegment": null,
      "partitionDimensions": [
        "city_and_country",
        "isRobot"
      ],
      "assumeGrouped": false
    },
    "defaultPartitioningVirtualColumns": [
      {
        "type": "expression",
        "name": "city_and_country",
        "expression": "concat(cityName, ':', countryName)",
        "outputType": "STRING"
      }
    ],
    "tuningConfig": null
  },
  "suspended": true
}
  ```

</details>

#### Release note

<hr>

 * `ReindexingTimelineView`
 * `SupervisorResource`
 * `web-console/src/components/reindexing-timeline/reindexing-timeline.scss`

<hr>

This PR has:

- [ ] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/compact/ReindexingTimelineView.java
```diff
@@ -0,0 +1,497 @@
...
+  }
+
+  /**
+   * Information about skip offsets and whether they were applied.
+   * Exactly one of {@code applied} or {@code notApplied} must be non-null.
+   */
+  public static class SkipOffsetInfo
+  {
+    private final AppliedSkipOffset applied;
+    private final NotAppliedSkipOffset notApplied;
```
kfaraz: Nit: These two classes feel a little unnecessary. We could have just have the actual fields in the class `SkipOffsetInfo` directly: `type`, `period`, a boolean for `isApplied`, a nullable `effectiveEndTime` mutually exclusive with a nullable `errorMessage`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -351,6 +357,80 @@ public Response getAllTaskStats(
...
+                           .build();
+          }
+
+          SupervisorSpec spec = specOptional.get();
+          if (!(spec instanceof CompactionSupervisorSpec)) {
+            return Response.status(Response.Status.BAD_REQUEST)
+                           .entity(ImmutableMap.of(
+                               "error",
+                               StringUtils.format(
+                                   "[%s] is not a compaction supervisor (type: %s)",
```
kfaraz: I think we can omit reporting back the actual type of the supervisor found.
```suggestion
                                   "Supervisor[%s] is not a compaction supervisor",
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -351,6 +357,80 @@ public Response getAllTaskStats(
...
+      @PathParam("id") final String id,
+      @QueryParam("referenceTime") @Nullable final String referenceTimeStr
+  )
+  {
+    return asLeaderWithSupervisorManager(
+        manager -> {
+          Optional<SupervisorSpec> specOptional = manager.getSupervisorSpec(id);
+          if (!specOptional.isPresent()) {
+            return Response.status(Response.Status.NOT_FOUND)
+                           .entity(ImmutableMap.of("error", StringUtils.format("[%s] does not exist", id)))
```
kfaraz:
```suggestion
                           .entity(ImmutableMap.of("error", StringUtils.format("Supervisor[%s] does not exist", id)))
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -351,6 +357,80 @@ public Response getAllTaskStats(
...
+          DateTime referenceTime;
+          if (referenceTimeStr != null) {
+            try {
+              referenceTime = DateTimes.of(referenceTimeStr);
+            }
+            catch (IllegalArgumentException e) {
+              return Response.status(Response.Status.BAD_REQUEST)
+                             .entity(ImmutableMap.of(
+                                 "error",
+                                 StringUtils.format("Invalid referenceTime format: %s", referenceTimeStr)
```
kfaraz:
```suggestion
                                 StringUtils.format("Reference time[%s] is in invalid format. Use ISO 8601 standard format.")
```

By the way, you could also try doing the following to make the various conditions in this method cleaner.
```java
return ServletResourceUtils.buildErrorResponseFrom(
    InvalidInput.exception(...)
);
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -351,6 +357,80 @@ public Response getAllTaskStats(
...
+          }
+
+          CompactionJobTemplate template = compactionSpec.getTemplate();
+          if (!(template instanceof CascadingReindexingTemplate)) {
+            return Response.status(Response.Status.BAD_REQUEST)
+                           .entity(ImmutableMap.of(
+                               "error",
+                               StringUtils.format(
+                                   "Reindexing timeline is only available for cascading reindexing supervisors. " +
+                                   "Supervisor [%s] uses template type: %s",
```
kfaraz:
```suggestion
                                   "Reindexing timeline is not available for supervisor[%s] as it does not use a cascading template",
```

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -1507,6 +1516,323 @@ public void test_generateAlignedSearchIntervals_failsWhenOlderRuleHasFinerGranul
...
+        new EqualityFilter("device", ColumnType.STRING, "mobile", null),
+        null
+    );
+
+    // Tuning and IO rules
+    ReindexingTuningConfigRule tuning7d = new ReindexingTuningConfigRule(
+        "tuning-7d",
+        null,
+        Period.days(7),
+        new UserCompactionTaskQueryTuningConfig(
```
kfaraz: Didn't we add a builder for this class in the previous PR?

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -1507,6 +1516,323 @@ public void test_generateAlignedSearchIntervals_failsWhenOlderRuleHasFinerGranul
...
+    for (ReindexingTimelineView.IntervalConfig intervalConfig : timeline.getIntervals()) {
+      if (intervalConfig.getInterval().getEnd().isAfter(expectedEffectiveEndTime)) {
+        Assertions.assertEquals(0, intervalConfig.getRuleCount());
+      }
+    }
+  }
+
+  /**
+   * Test validation error when granularity timeline is invalid
+   */
```
kfaraz: Not really needed since the test name captures this already.

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -1507,6 +1516,323 @@ public void test_generateAlignedSearchIntervals_failsWhenOlderRuleHasFinerGranul
...
+      if (intervalConfig.getInterval().getEnd().isAfter(referenceTime.minusDays(10))) {
+        Assertions.assertTrue(hasTuningRule, "Recent intervals should have tuning rules");
+        Assertions.assertTrue(hasIORule, "Recent intervals should have IO rules");
+      }
+    }
+  }
+
+  /**
+   * Test that skipOffsetFromNow correctly skips intervals and populates skipOffset.applied
+   */
```
kfaraz: Not really needed as the test name already captures this.

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
```diff
@@ -1507,6 +1516,323 @@ public void test_generateAlignedSearchIntervals_failsWhenOlderRuleHasFinerGranul
...
+  @Test
+  public void test_getReindexingTimelineView_validationError_invalidGranularityTimeline()
+  {
+    DateTime referenceTime = DateTimes.of("2025-01-29T16:15:00Z");
+
+    // Create rules that violate the granularity constraint:
+    // Older data (P90D) has DAY granularity, newer data (P30D) has HOUR granularity
+    // This means as we move from past to present, granularity gets finer (valid)
+    // But then if we add MONTH for recent data, it becomes coarser (invalid)
+    ReindexingRuleProvider provider = InlineReindexingRuleProvider.builder()
```
kfaraz: Style tip: Move the `.builder()` to the next line to make the whole thing feel less crowded.
```suggestion
    ReindexingRuleProvider provider = InlineReindexingRuleProvider
        .builder()
```

kfaraz (commented):
Pretty cool!
Left some minor suggestions.

Will try this out locally.
Would also be nice to get a review from @vogievetsky on the web-console stuff.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
```diff
@@ -217,6 +243,118 @@ private static boolean intervalEndsAfter(Interval interval, DateTime boundary)
...
 
+  /**
+   * Generates a timeline view showing the search intervals and their associated reindexing
+   * configurations. This is useful for operators to understand how rules are applied across
+   * different time periods and to preview the effects of rule changes before they are applied.
+   *
+   * @param referenceTime the reference time to use for computing rule periods (typically DateTime.now())
+   * @return a view of the reindexing timeline with intervals and their configs
+   */
+  public ReindexingTimelineView getReindexingTimelineView(DateTime referenceTime)
```
kfaraz, line 389: I feel like this method should not live here as it kind of pollutes the template definition.

Either have the timeline be built using a static method like `ReindexingTimelineView.fromTemplate(template, referenceTime)` or have a separate builder class altogether.

kfaraz:
@capistrant , I took the liberty to slightly update the title for clarity. Hope it seems less ambiguous now.

kfaraz:
Was trying out the embedded test in this PR to check out the timeline.

@capistrant , in the snapshot below, what does the "4 rules applied" mean?
The widget below it shows only 2 items - a segment granularity and a deletion rule.

Maybe we should include an info popup for each of the chips.
Also, the chip says `"1" deletion rule". When can there be more than 1 for a single interval?

Screenshot 2026-02-25 at 12 50 41 PM

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
capistrant: hmm. I wonder about how we'd manage to re-use a lot of the underlying private implementation details that are quite convoluted (unfortunately). Like creating the base timeline from segment gran rules and what not is all shared between the timeline creation and the task creation. I agree that it feels a little weird here, but I didn't know how else to approach it without either duping a lot of internal logic or exposing it to both somehow. will keep thinking though

capistrant:
> Was trying out the embedded test in this PR to check out the timeline.
> 
> @capistrant , in the snapshot below, what does the "4 rules applied" mean? The widget below it shows only 2 items - a segment granularity and a deletion rule.
> 
> Maybe we should include an info popup for each of the chips. Also, the chip says `"1" deletion rule". When can there be more than 1 for a single interval?
> 
> Screenshot 2026-02-25 at 12 50 41 PM

4 rules applied means that the config created for the interval is made up of 4 different reindexing rules. In this case based on the test I assume you are running, it is
* segmentGran rule
* deletion rule
* ioConfig rule
* tuningConfig rule

You could confirm by clicking the view raw rules button.

I think this is a UX issue. Hand up, I don't have any experience curating a good UX :). If I had to guess what UX decision is best here, it is to remove the `N rules applied` chip. if someone really wants to see the explicit underlying rules they can click view raw rules.

As for deletion rules, N deletion rules can apply to any given interval. They accumulate (unlike other rule types). Every deletion rule that applies a search interval will be combined together.
* deleteWhere "countryName=france" olderThan P7D"
* deleteWhere "countryName=italy" olderThan P14D"

Any interval where both of those apply would have 2 deletion rules.

The chips are intended to give very quick reference to what is in the config for the given search interval. They are not strictly tied to rules even. In my pictures you see metrics and dims in separate chips, but both belong to the same rule. So maybe the deletion rules chip should actually be like 2 deletion filters or 2 deletion clauses. I didn't know how valuable chips would be for tuningConfig and ioConfig so those are not surfaced, likely leading to your initial confusion. You'd have to view the config payload or the raw rules to see them.

indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
capistrant: no, it is coming in https://github.com/apache/druid/pull/18996 though

capistrant:
> > Was trying out the embedded test in this PR to check out the timeline.
> > @capistrant , in the snapshot below, what does the "4 rules applied" mean? The widget below it shows only 2 items - a segment granularity and a deletion rule.
> > Maybe we should include an info popup for each of the chips. Also, the chip says `"1" deletion rule". When can there be more than 1 for a single interval?
> > Screenshot 2026-02-25 at 12 50 41 PM
> 
> 4 rules applied means that the config created for the interval is made up of 4 different reindexing rules. In this case based on the test I assume you are running, it is
> 
> * segmentGran rule
> * deletion rule
> * ioConfig rule
> * tuningConfig rule
> 
> You could confirm by clicking the view raw rules button.
> 
> I think this is a UX issue. Hand up, I don't have any experience curating a good UX :). If I had to guess what UX decision is best here, it is to remove the `N rules applied` chip. if someone really wants to see the explicit underlying rules they can click view raw rules.
> 
> As for deletion rules, N deletion rules can apply to any given interval. They accumulate (unlike other rule types). Every deletion rule that applies a search interval will be combined together.
> 
> * deleteWhere "countryName=france" olderThan P7D"
> * deleteWhere "countryName=italy" olderThan P14D"
> 
> Any interval where both of those apply would have 2 deletion rules.
> 
> The chips are intended to give very quick reference to what is in the config for the given search interval. They are not strictly tied to rules even. In my pictures you see metrics and dims in separate chips, but both belong to the same rule. So maybe the deletion rules chip should actually be like 2 deletion filters or 2 deletion clauses. I didn't know how valuable chips would be for tuningConfig and ioConfig so those are not surfaced, likely leading to your initial confusion. You'd have to view the config payload or the raw rules to see them.

I made some minor UX updates that I think help clear up some of your initial confusion. I updated the description images to reflect (2nd and 3rd image)

1. no more `N rules applied` chip. Instead updated the raw rules button to say how many rules there are applied.
2. removed the word "rule" from the deletion chip. I don't want folks to be confused into thinking that executive summary of the interval are 1:1 with rules.
3. In that same summary row of chips I added a label "Effective Configuration" which has a tooltip to tell the operator that it is a summary and the full config can be referenced via the button(s) below.
4.  Added a partitioning type chip as that may be interesting.

kfaraz:
@capistrant , is this PR ready for another round of review or should we hold off until the refactor PR #19106  is reviewed and merged?

capistrant:
> @capistrant , is this PR ready for another round of review or should we hold off until the refactor PR #19106  is reviewed and merged?

Can probably hold off for now until the rule refactor is done

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
```diff
@@ -262,6 +288,121 @@ private static boolean intervalEndsAfter(Interval interval, DateTime boundary)
...
+        continue;
+      }
+
+      InlineSchemaDataSourceCompactionConfig.Builder builder = createBaseBuilder();
+      ReindexingConfigBuilder configBuilder = new ReindexingConfigBuilder(
+          ruleProvider,
+          searchInterval,
+          referenceTime,
+          searchIntervals,
+          null
```
FrankChen021: P2 Timeline preview drops template tuningConfig

getReindexingTimelineView builds each interval config with a ReindexingConfigBuilder whose baseTuningConfig is null, while createCompactionJobs passes the template's tuningConfig into the same builder. Any cascading supervisor with static tuning settings such as maxRowsInMemory, splitHintSpec, maxNumConcurrentSubTasks, indexSpecForIntermediatePersists, or retry/chat settings will have those settings in the actual compaction jobs but not in the new reindexingTimeline API response or downloaded UI config. This makes the preview/API incompatible with the real effective config. Pass tuningConfig here as createCompactionJobs does, and add a test that a non-partitionsSpec tuning field survives in the timeline view.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
```diff
@@ -312,6 +338,121 @@ private static boolean intervalEndsAfter(Interval interval, DateTime boundary)
...
+      );
+    }
+
+    // Build configs for each interval
+    List<ReindexingTimelineView.IntervalConfig> intervalConfigs = new ArrayList<>();
+    for (IntervalPartitioningInfo intervalInfo : searchIntervals) {
+      Interval searchInterval = intervalInfo.getInterval();
+
+      // Check if interval extends past skip offset
+      if (intervalEndsAfter(searchInterval, effectiveEndTime)) {
```
FrankChen021: [P2] Timeline marks partially compacted intervals as skipped

createCompactionJobs now truncates an interval that crosses the skip-offset boundary and still creates jobs for the older aligned subrange, but getReindexingTimelineView still returns ruleCount 0/config null for any interval whose end is after effectiveEndTime. For skipOffsetFromNow, and for the console's skipOffsetFromLatest preview after querying maxTime, the UI will show an interval as fully skipped even though the supervisor will compact part of it. Please mirror the truncation/splitting behavior in the timeline view so the preview matches actual job generation.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

web-console/src/components/reindexing-timeline/reindexing-timeline.tsx
```diff
@@ -0,0 +1,686 @@
...
+  const { intervals, skipOffset, referenceTime, validationError } = timelineData;
+
+  const handleQueryMaxTime = async () => {
+    setQueryingMaxTime(true);
+    try {
+      const query = {
+        queryType: 'timeBoundary',
+        dataSource: timelineData.dataSource,
+      };
+      const resp = await Api.instance.post('/druid/v2', query);
```
FrankChen021, line 157: [P2] Compute skipOffsetFromLatest from the supervisor timeline

The `Query latest timestamp` path bases the preview boundary on a broker timeBoundary `maxTime`, then subtracts the period in the browser. The actual cascading supervisor path applies `skipOffsetFromLatest` to the segment timeline end and then truncates to an interval granularity-aligned boundary, so row maxTime can be earlier than the segment end and straddling intervals are handled differently. For normal DAY/HOUR segments this can show intervals as skipped or active at the wrong boundary. Please derive the preview from segment metadata/server-side timeline logic, or mirror the same segment-end and granularity-alignment rules.

web-console/src/dialogs/supervisor-table-action-dialog/supervisor-table-action-dialog.tsx
```diff
@@ -67,6 +69,16 @@ export const SupervisorTableActionDialog = React.memo(function SupervisorTableAc
       active: activeTab === 'history',
       onClick: () => setActiveTab('history'),
     },
+    ...(isCompactionSupervisor
```
FrankChen021, line 72: [P3] Hide the timeline tab for non-cascading compaction supervisors

This adds the Timeline tab for every supervisor ID that starts with `autocompact__`, but the new endpoint returns 400 unless the compaction supervisor template is `CascadingReindexingTemplate`. Existing inline-config compaction supervisors also use `autocompact__` IDs, so standard compaction supervisors now expose a tab that only shows an error. Gate the tab on the loaded supervisor spec/template type, or render a deliberate unavailable state instead of adding the tab solely from the ID prefix.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 1 |
| Total | 2 |

---

web-console/src/components/reindexing-timeline/reindexing-timeline.tsx
```diff
@@ -0,0 +1,686 @@
...
+              </Tag>
+            )}
+            {getPartitionType(config.tuningConfig) && (
+              <Tag minimal intent={Intent.PRIMARY} icon={IconNames.SPLIT_COLUMNS}>
+                Partitioning: {getPartitionType(config.tuningConfig)}
+              </Tag>
+            )}
+            {metricsCount > 0 && (
+              <Tag minimal intent={Intent.SUCCESS} icon={IconNames.CALCULATOR}>
+                {metricsCount} metric{metricsCount !== 1 ? 's' : ''}
```
vogievetsky, line 514: this is what the `pluralIfNeeded` utility function is for.

web-console/src/components/reindexing-timeline/reindexing-timeline.tsx
```diff
@@ -0,0 +1,686 @@
...
+        <div className="detail-content">
+          <Tooltip
+            content="A high-level summary of the merged rule configuration for this interval — use View Full Configuration for complete details."
+            position="bottom"
+          >
+            <div className="config-summary-label">Effective Configuration</div>
+          </Tooltip>
+          <div className="config-badges">
+            {config.granularitySpec?.segmentGranularity && (
+              <Tag minimal intent={Intent.PRIMARY} icon={IconNames.CALENDAR}>
```
vogievetsky, line 498: not shy about using `intent={Intent.PRIMARY}`!

web-console/src/components/reindexing-timeline/reindexing-timeline.tsx
```diff
@@ -0,0 +1,686 @@
...
+            onClick={onClose}
+            aria-label="Close interval details"
+          />
+        </div>
+
+        <div className="detail-content">
+          <Tooltip
+            content="A high-level summary of the merged rule configuration for this interval — use View Full Configuration for complete details."
+            position="bottom"
+          >
```
vogievetsky, line 493: Web console does not really use `<Tooltip>` it is more extreme! Search for `data-label`

web-console/src/components/reindexing-timeline/reindexing-timeline.tsx
```diff
@@ -0,0 +1,686 @@
...
+              <Tag minimal intent={Intent.WARNING} icon={IconNames.TRASH}>
+                {deletionRuleCount} deletion clause{deletionRuleCount !== 1 ? 's' : ''}
+              </Tag>
+            )}
+          </div>
+
+          <div className="full-config-section">
+            <Button
+              icon={IconNames.CODE}
+              text="View Full Configuration"
```
vogievetsky, line 537: Web console has the convention of doing sentence case for buttons and internal labels in general so this should be `"View full configuration"`

web-console/src/components/reindexing-timeline/reindexing-timeline.tsx
```diff
@@ -0,0 +1,686 @@
...
+                Partitioning: {getPartitionType(config.tuningConfig)}
+              </Tag>
+            )}
+            {metricsCount > 0 && (
+              <Tag minimal intent={Intent.SUCCESS} icon={IconNames.CALCULATOR}>
+                {metricsCount} metric{metricsCount !== 1 ? 's' : ''}
+              </Tag>
+            )}
+            {dimensionsCount > 0 && (
+              <Tag minimal intent={Intent.SUCCESS} icon={IconNames.TAG}>
```
vogievetsky, line 518: I feel like the intents are being used here just to add color coding to the `Tag` s but that is not how intents should be used. A better approach would be some colored text with custom colors (not the intent colors)

web-console/src/components/reindexing-timeline/reindexing-timeline.tsx
```diff
@@ -0,0 +1,686 @@
...
+              text="Download"
+              icon={IconNames.DOWNLOAD}
+              intent={Intent.PRIMARY}
+              onClick={() =>
+                handleDownloadJson(
+                  config,
+                  `reindexing-config-${interval.interval.replace(/\//g, '-')}.json`,
+                )
+              }
+            />
```
vogievetsky, line 583: neither of the buttons should be primary. There should be a primary `Close` button instead.

web-console/src/components/reindexing-timeline/reindexing-timeline.tsx
```diff
@@ -0,0 +1,686 @@
...
+            content={`${interval.appliedRules.length} rules for ${interval.interval}`}
+            position="bottom"
+          >
+            <span>Rules for {formatInterval(interval.interval)}</span>
+          </Tooltip>
+        }
+        className="reindexing-config-dialog"
+        canOutsideClickClose
+      >
+        <div className="bp5-dialog-body">
```
vogievetsky, line 602: doing `className="bp5-dialog-body"` is an anti-pattern prefer `className={Classes.DIALOG_BODY}` instead

web-console/src/components/reindexing-timeline/reindexing-timeline.tsx
```diff
@@ -0,0 +1,686 @@
...
+}
+
+function ConfigJsonViewer({ config, isRulesList }: ConfigJsonViewerProps) {
+  let jsonValue: string;
+
+  if (isRulesList && Array.isArray(config)) {
+    // Group rules by type for better readability
+    const rulesByType: Record<string, ReindexingRule[]> = {};
+    config.forEach((rule: ReindexingRule) => {
+      const type = getRuleTypeName(rule);
```
vogievetsky, line 643: I don't like that we are creating fake JSONs with custom names in them. There is a general rule of thumb in the console (that I am just realizing) that all JSON shown to the user should correspond to something in the Druid API. Here I suggest we redesign this dialog, maybe adding several JSON fields.

capistrant:
https://github.com/apache/druid/pull/19511 is going to handle the API changes and @vogievetsky is going to tackle the UI

PR #19052, opened 2026-02-25
Perf: Remove redundant map lookups in CustomTierSelectorStrategy comparator

abhishekrb19:
Avoid repeated `containsKey()` and `get()` calls by retrieving each map value once and using null checks instead. This reduces redundant lookups and simplifies the comparator logic.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/client/selector/CustomTierSelectorStrategy.java
+++ b/server/src/main/java/org/apache/druid/client/selector/CustomTierSelectorStrategy.java
@@ -48,18 +48,23 @@ public class CustomTierSelectorStrategy extends AbstractTierSelectorStrategy
     }
 
     // Tiers with priorities explicitly specified in the custom priority list config always have higher priority than
-    // those not and those not specified fall back to use highest priority strategy among themselves
+    // those not and those not specified fall back to use the highest priority strategy among themselves
     this.comparator = (p1, p2) -> {
-      if (lookup.containsKey(p1) && lookup.containsKey(p2)) {
-        return Integer.compare(lookup.get(p1), lookup.get(p2));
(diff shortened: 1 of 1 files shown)


GWphua (approved):
LGTM, regardless of the answer to the question below. 

How much of a performance boost is this, given that it's looking up a hashmap?

abhishekrb19:
Thanks for taking a look, @GWphua!

>How much of a performance boost is this, given that it's looking up a hashmap?

I didn’t benchmark this change. It should remove about 2–3 unnecessary map lookups per query.

abhishekrb19:
I got curious about the performance improvement and ran a quick benchmark. With this lookup access pattern, we see a good ~18–20% speed up on average - [CustomTierSelectorComparatorBenchmark.java](https://github.com/user-attachments/files/25587699/CustomTierSelectorComparatorBenchmark.java). I did a quick spot check in other areas of the codebase and there seems to be a [few](https://github.com/apache/druid/blob/master/extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisor.java#L450-L455) [other](https://github.com/apache/druid/blob/master/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java#L3154-L3155) occurrences where we can simply do `map.get(key)` instead of `map.containsKey(key)` + `map.get(key)`.

```
Benchmark                                                Mode  Cnt    Score   Error  Units
CustomTierSelectorComparatorBenchmark.newImplementation  avgt   10  456.969 ± 5.175  ns/op
CustomTierSelectorComparatorBenchmark.oldImplementation  avgt   10  571.750 ± 4.194  ns/op
```

PR #19056, opened 2026-02-25
fix: Restarted the watch after stream reset #19055

### LINKED ISSUE #19055: Broker retains stale Historical IP after pod recreation; service discovery watch not recovering from stream reset
## Description

We are encountering an issue where the [Apache Druid](chatgpt://generic-entity?number=0) Broker fails to refresh its service discovery state when Historical pods are recreated with new IP addresses in Kubernetes.

This was initially observed during a cluster upgrade, where there may have been transient connectivity issues with the Kubernetes API server. However, the problem persisted even after the cluster stabilized.

The Broker continues attempting to connect to a stale Historical IP address (which may already be reassigned to a non-Druid pod). The issue does not self-heal and only resolves after manually restarting the Broker pod.

## Environment

- Deployment: Kubernetes
- Discovery: K8s-based node discovery
- Module: extensions-core/kubernetes-extensions
- Components affected: Broker

## Symptoms

- Historical pod is terminated.
- A new Historical pod is created with a new IP.
- Broker continues attempting to connect to the old IP.
- Issue persists indefinitely until Broker pod is restarted manually.

Broker logs repeatedly show:

```
org.apache.druid.server.coordination.ChangeRequestHttpSyncer - Sync failed for server[https://xx:8091/] while [Handling response with code[0], description[null]]. Failed [1] times in the last [0] seconds. (org.jboss.netty.channel.ChannelException: Faulty channel in resource pool)
```

Additionally, errors from the K8s discovery layer:
```
ERROR [org.apache.druid.k8s.discovery.K8sDruidNodeDiscoveryProvider$NodeRoleWatcherpeon] 
org.apache.druid.k8s.discovery.K8sDruidNodeDiscoveryProvider$NodeRoleWatcher - 
Error while watching role[PEON]
java.lang.RuntimeException: IO Exception during hasNext method.
    at io.kubernetes.client.util.Watch.hasNext(Watch.java:183)
    at org.apache.druid.k8s.discovery.DefaultK8sApiClient$2.hasNext(DefaultK8sApiClient.java:132)
    at org.apache.druid.k8s.discovery.K8sDruidNodeDiscoveryProvider$NodeRoleWatcher.keepWatching(K8sDruidNodeDiscoveryProvider.java:266)
    ...
Caused by: okhttp3.internal.http2.StreamResetException: stream was reset: CANCEL
    at okhttp3.internal.http2.Http2Stream$FramingSource.read(Http2Stream.kt:355)
    at okhttp3.internal.connection.Exchange$ResponseBodySource.read(Exchange.kt:281)
    at okio.RealBufferedSource.exhausted(RealBufferedSource.kt:200)
    at io.kubernetes.client.util.Watch.hasNext(Watch.java:181)
```

## Root Cause

In extensions-core/kubernetes-extensions, the K8sDruidNodeDiscoveryProvider$NodeRoleWatcher encounters:

```
okhttp3.internal.http2.StreamResetException: stream was reset: CANCEL
```

## Expected Behavior

- On watch stream reset or IO exception, the watcher should automatically re-establish the watch.
- A fresh watch should be established
- Node inventory should be reconciled.
- Terminated Historical nodes should be removed from Broker view.

vivek807:
Fixes #19055.

### Description
Restarted the watch after stream reset 

#### Fixed the bug #19055

This PR has:

- [x] been self-reviewed.


extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/K8sDruidNodeDiscoveryProvider.java
```diff
@@ -252,45 +253,43 @@ private void watch()
...
+            // This should be updated after the action has been dealt with successfully
+            nextResourceVersion = item.object.getResourceVersion();
           }
 
+          LOGGER.trace("Watch closed normally for role[%s]", this.nodeRole);
+          return;
+        }
+        catch (StreamResetException ex) {
+          LOGGER.debug("Watch stream terminated normally for role[%s], restarting", this.nodeRole);
+          return;
```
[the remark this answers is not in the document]
FrankChen021: I think this comment is fair. The `K8SDruidNodeDiscoveryProvider` should not be aware of underlying channel implementation(like okhttp3) and exception like `StreamResetException` in okhttp3. 
can you move this catch into the `DefaultK8sApiClient` and re-throw a new named exception like `ChannelResetException`

extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/K8sDruidNodeDiscoveryProvider.java
```diff
@@ -252,45 +253,43 @@ private void watch()
...
         }
```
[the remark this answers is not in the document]
FrankChen021: true

FrankChen021:
@vivek807 Thanks for reporting the issue and submitting this PR for fix. can you address above comments?

vivek807:
> @vivek807 Thanks for reporting the issue and submitting this PR for fix. can you address above comments?

updated, please recheck.

extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/K8sDruidNodeDiscoveryProvider.java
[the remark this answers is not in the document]
vivek807: Added test

extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/K8sDruidNodeDiscoveryProvider.java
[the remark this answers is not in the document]
vivek807: updated, thanks

extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/DefaultK8sApiClient.java
```diff
@@ -163,9 +164,11 @@ public boolean hasNext() throws SocketTimeoutException
           catch (RuntimeException ex) {
             if (ex.getCause() instanceof SocketTimeoutException) {
               throw (SocketTimeoutException) ex.getCause();
-            } else {
-              throw ex;
             }
+            if (ex.getCause() instanceof StreamResetException) {
```
capistrant: this is still relying on an okhttp internal. Maybe we can just catch any wrapped IOException and re-throw? then catch IOException in the discovery provider after we catch the socket timeout. and log/return to force a re-list? One fear is that is too heavy handed though as full list is expensive, especially in clusters with lots of pods.

capistrant (changes requested):
Still think we need to push further in removing the okhttp internals from this PR. My initial thought is that we just need to handle all IOExceptions with a full re-list. We can keep the one off handling of the known ok IOException that a simple retry of the watch from same resource version will work for. But for all others force a re-list? My biggest fear is that this is an overreaction.

extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/DefaultK8sApiClient.java
vivek807: I’m specifically trying to handle the okhttp StreamResetException, so it seems we’ll need to retain this dependency. Any better idea?

gianm: If there is a need to look for okhttp-specific exceptions, and more generic logic cannot be used, you could check for them by class name. i.e. `if ex.getCause().getClass().getName().equals("foo.bar.StreamResetException")`.

vivek807: Yes, I think that’s doable. However, I’m concerned whether retaining the dependency and relying on a class-based check could cause issues. Other Kubernetes components are also using this library, as evidenced by the specific exception we’re encountering.

vivek807: Hi @gianm,

Let me know if that's ok, else I will add class name check as suggested.

extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/DefaultK8sApiClient.java
```diff
@@ -33,6 +33,7 @@
 import io.kubernetes.client.openapi.models.V1PodList;
 import io.kubernetes.client.util.PatchUtils;
 import io.kubernetes.client.util.Watch;
+import okhttp3.internal.http2.StreamResetException;
```
FrankChen021: P2 Avoid depending on OkHttp internal exception type

The fix imports okhttp3.internal.http2.StreamResetException, which is an internal OkHttp class rather than a supported API. This makes the extension compile- and behavior-sensitive to OkHttp internals, so a Kubernetes client or OkHttp upgrade can break this code even if public APIs remain compatible. Since StreamResetException is an IOException, the watch layer can avoid this dependency by treating non-timeout IOExceptions as full-watch restarts, while preserving the existing SocketTimeoutException retry behavior.

vivek807: Thanks @FrankChen021. I have fixed the dependency and moved the exception to non-interrupt IOException

FrankChen021 (commented):
Follow-up handled: the OkHttp internal `StreamResetException` dependency was removed, and the watch handling now treats non-interrupt `IOException`s as restartable channel resets while preserving timeout handling. No further inline reply needed from me.

---

---

extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/DefaultK8sApiClient.java
vivek807: Fixed the dependency and moved the exception to non-interrupt IOException

vivek807:
@FrankChen021 @gianm @capistrant  Please let me know if there’s anything I can do to help move this PR forward.

FrankChen021 (commented):
I reviewed the update and full PR for correctness, lifecycle, retry, cancellation, cleanup, and concurrency risks; no new issues were found.

---

AdheipSingh:
@capistrant - Can you please check this.

extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/K8sDruidNodeDiscoveryProvider.java
```diff
@@ -254,50 +254,48 @@ private void watch()
...
+            // This should be updated after the action has been dealt with successfully
+            nextResourceVersion = item.object.getResourceVersion();
           }
 
+          LOGGER.trace("Watch closed normally for role[%s]", this.nodeRole);
+          return;
+        }
+        catch (ChannelResetException ex) {
+          LOGGER.debug("Watch stream terminated normally for role[%s], restarting", this.nodeRole);
+          return;
```
FrankChen021, line 298: [P2] Persistent resets can cause a tight relist loop

Returning from this handler immediately re-enters the outer `watch()` loop, which performs `listPods()` and opens another watch without using `watcherErrorRetryWaitMS`. If the API server or an intermediary persistently resets every watch stream, this becomes an unbounded list/watch loop that can hammer the Kubernetes API and consume the watcher thread. Apply the configured retry backoff (or otherwise rate-limit full resyncs) before retrying this failure path.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

I reviewed the full current diff because the prior reviewed baseline SHA was unavailable.

---

extensions-core/kubernetes-extensions/src/main/java/org/apache/druid/k8s/discovery/DefaultK8sApiClient.java
```diff
@@ -263,8 +264,14 @@
...
-                // Kubernetes API documentation for details.
-                LOGGER.debug("BOOKMARK event fired, no nothing, only update resourceVersion");
+                // Kubernetes API documentation for details. Build a response carrying just the
+                //updated resourceVersion; otherwise obj would still hold the previous event (or
+                // null on the first event), corrupting the caller's resourceVersion tracking.
+                LOGGER.debug("BOOKMARK event fired, only updating resourceVersion");
+                obj = new Watch.Response<>(
+                        WatchResult.BOOKMARK,
+                        new DiscoveryDruidNodeAndResourceVersion(
+                                item.object.getMetadata().getResourceVersion(), null));
```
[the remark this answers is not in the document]
vivek807: BOOKMARK events always has resource version

FrankChen021 (commented):
I reviewed the incremental update and the full current diff for correctness, edge cases, concurrency, lifecycle, retry, cleanup, compatibility, and data-loss risks; no issues found.

---

FrankChen021 (approved):

AdheipSingh:
I have manually tested this PR on k8s version 1.35.

AdheipSingh (approved):

PR #19058, opened 2026-02-26
Refactor: Ensure that CompactionCandidate.currentStatus is non-null

kfaraz:
### Description

Inspired by #18968 , this PR tries to ensure that the `currentStatus` inside a `CompactionCandidate` is always non-null,
with a relatively smaller changeset.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/test/java/org/apache/druid/server/compaction/CompactionStatusTest.java
+++ b/server/src/test/java/org/apache/druid/server/compaction/CompactionStatusTest.java
@@ -574,7 +574,7 @@ public class CompactionStatusTest
     List<DataSegment> segments = List.of(
         DataSegment.builder(WIKI_SEGMENT).indexingStateFingerprint(oldFingerprint).build()
     );
-    CompactionStatus status = CompactionStatus.compute(segments, newConfig, fingerprintMapperr);
+    CompactionStatus status = CompactionStatus.compute(segments, newConfig, fingerprintMapper);
     Assert.assertFalse(status.isComplete());
     Assert.assertTrue(status.getReason().startsWith("'transformSpec' mismatch"));
   }


server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -46,13 +47,30 @@ public class CompactionCandidate
...
 
-  public static CompactionCandidate from(
+  public static Interval getCompactionInterval(
       List<DataSegment> segments,
       @Nullable Granularity targetSegmentGranularity
   )
+  {
+    final Set<Interval> segmentIntervals =
+        segments.stream().map(DataSegment::getInterval).collect(Collectors.toSet());
+    final Interval umbrellaInterval = JodaUtils.umbrellaInterval(segmentIntervals);
```
cecemei, line 57: wondering if umbrellaInterval would be different from compactionInterval? seems like `DataSourceCompactibleSegmentIterator.populateQueue` would create `DataSegment` with updated interval and shard spec, if `config.getSegmentGranularity` is set.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -68,7 +86,7 @@ public static CompactionCandidate from(
         umbrellaInterval,
         compactionInterval,
         segmentIntervals.size(),
```
cecemei, line 88: nit: could use CompactionStatistics to wrap numIntervals, totalBytes and numSegments.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
kfaraz: Yes, they can be different. Umbrella interval is the interval spanning all the segments (i.e. start of earliest segment to end of latest segment).

Compaction interval is nothing but the umbrella interval expanded to align with the desired segment granularity (if specified).

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
kfaraz: Sure, will include in a follow up.

kfaraz:
Thanks for the review, @cecemei !

PR #19059, opened 2026-02-26
add minor compaction for Overlord based compaction supervisors using the MSQ engine

cecemei:
Description

  This PR introduces minor compaction support in msq engine to Apache Druid, enabling compaction operations on only uncompacted segments within an interval while upgrading already-compacted segments for consistency.

  Key Changes

  1. CompactionMode Enum Extension
  - Added ALL_SEGMENTS mode (full compaction of all segments)
  - Added UNCOMPACTED_SEGMENTS_ONLY mode (compacts only new segments, upgrades existing ones)

  2. MinorCompactionInputSpec
  A new CompactionInputSpec implementation specifically for minor compaction:
  - Specifies uncompacted segments to compact within an interval
  - Required (non-nullable) interval and uncompactedSegments fields
  - Type: "uncompacted"
  - Replaces nullable field handling in CompactionIntervalSpec, making the intent of minor compaction explicit

  3. MarkSegmentToUpgradeAction
  A new task action that updates segment metadata without rewriting data, modifying partition numbers and shard specifications of already-compacted segments.

  4. ShardSpec Interface Enhancements
  Extended with mutation methods including withPartitionNum(), withCorePartitions(), and isNumChunkSupported() checks.

  6. Policy-Level Control
  Introduced minUncompactedBytesPercentForFullCompaction threshold in MostFragmentedIntervalFirstPolicy determining when incremental versus full compaction applies based on uncompacted-to-total segment byte ratios.

  7. Builder Pattern Implementation
  Replaced verbose constructors in UserCompactionTaskQueryTuningConfig with a builder pattern for improved readability.

  Release Notes

  This feature enables compacting newly ingested segments while upgrading already compacted data, with configuration through percentage thresholds in compaction policies. Support is currently limited to Overlord based compaction supervisors using the MSQ engine.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/multi-stage-query/src/test/java/org/apache/druid/msq/exec/MSQCompactionTaskRunTest.java
+++ b/multi-stage-query/src/test/java/org/apache/druid/msq/exec/MSQCompactionTaskRunTest.java
@@ -592,7 +592,13 @@ public class MSQCompactionTaskRunTest extends CompactionTaskRunBase
 
     final Pair<TaskStatus, DataSegmentsWithSchemas> resultPair1 = runTask(compactionTask1);
     verifyTaskSuccessRowsAndSchemaMatch(resultPair1, TOTAL_TEST_ROWS);
-    verifyCompactedSegment(List.copyOf(resultPair1.rhs.getSegments()), segmentGranularity, DEFAULT_QUERY_GRAN, false);
+    verifyCompactedSegment(
+        compactionTask1.getCompactionRunner(),
+        List.copyOf(resultPair1.rhs.getSegments()),
+        segmentGranularity,
+        DEFAULT_QUERY_GRAN,
+        false
+    );
     Assert.assertEquals(1, resultPair1.rhs.getSegments().size());
     final DataSegment compactedSegment1 = Iterables.getOnlyElement(resultPair1.rhs.getSegments());
 


server/src/main/java/org/apache/druid/server/compaction/CompactionMode.java
```diff
@@ -27,5 +27,6 @@ public enum CompactionMode
   /**
    * Indicates that all existing segments of the interval will be picked for compaction.
    */
-  FULL_COMPACTION;
+  FULL_COMPACTION,
+  INCREMENTAL_COMPACTION;
```
kfaraz: Should we just rename these mode types to `ALL_SEGMENTS`, `UNCOMPACTED_SEGMENTS_ONLY`?
It will clarify the functionality and intent better.

Also "incremental" can be ambiguous. (As @clintropolis had pointed out offline, all auto-compaction is incremental in nature, since it happens one-interval-at-a-time).

capistrant: please provide a similar comment generalizing what INCREMENTAL_COMPACTION is, like is done for full

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -124,7 +147,12 @@ public boolean equals(Object o)
     MostFragmentedIntervalFirstPolicy policy = (MostFragmentedIntervalFirstPolicy) o;
     return minUncompactedCount == policy.minUncompactedCount
            && Objects.equals(minUncompactedBytes, policy.minUncompactedBytes)
-           && Objects.equals(maxAverageUncompactedBytesPerSegment, policy.maxAverageUncompactedBytesPerSegment);
+           && Objects.equals(maxAverageUncompactedBytesPerSegment, policy.maxAverageUncompactedBytesPerSegment)
+           // Use Double.compare instead of == to handle NaN correctly and keep equals() consistent with hashCode() (especially for +0.0 vs -0.0).
+           && Double.compare(
+        incrementalCompactionUncompactedRatioThreshold,
+        policy.incrementalCompactionUncompactedRatioThreshold
+    ) == 0;
```
capistrant: is this indentation right? it looks kind of odd, at least in my browser

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -151,6 +151,13 @@ public CompactionStatistics getUncompactedStats()
            ? null : currentStatus.getUncompactedStats();
   }
 
+  @Nullable
+  public List<DataSegment> getUncompactedSegments()
+  {
+    return (currentStatus == null || currentStatus.getUncompactedSegments() == null)
+           ? null : currentStatus.getUncompactedSegments();
```
capistrant: nit
```suggestion
    return currentStatus == null ? null : currentStatus.getUncompactedSegments();
```

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
```diff
@@ -282,18 +283,20 @@ private boolean startJobIfPendingAndReady(
     }
 
     // Check if the job is already running, completed or skipped
```
capistrant: nit: this comment might need a refresh to reflect how the status handling has changed versus the former code. I was confused about why PENDING and COMPLETE resulted in a throw until I reviewed deriveCompactionStatus

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
capistrant: should we update the javadoc to call out how this method now offers special handling for incremental compaction that includes upgrading already compacted segments?

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentUpgradeAction.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.druid.indexing.common.task.Task;
+import org.apache.druid.java.util.common.IAE;
+import org.apache.druid.metadata.ReplaceTaskLock;
+import org.apache.druid.timeline.DataSegment;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class SegmentUpgradeAction implements TaskAction<Integer>
```
capistrant: I think this warrants a javadoc

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/MarkSegmentToUpgradeActionTest.java
capistrant, line 1: since this is a net new file, can we go straight to junit5?

processing/src/main/java/org/apache/druid/timeline/partition/LinearShardSpec.java
```diff
@@ -55,6 +55,18 @@ public int getNumCorePartitions()
     return 0;
   }
 
+  @Override
+  public ShardSpec withPartitionNum(int partitionNum1)
```
capistrant: repeated through the shard specs we have this naming pattern with `1` suffix. is that intentional? I don't follow the significance of it if so

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -1852,9 +1853,18 @@ protected Set<DataSegment> insertSegments(
   }
 
   /**
-   * Creates new versions of segments appended while a "REPLACE" task was in progress.
+   * Creates upgraded versions of segments that were appended while a REPLACE task was in progress.
```
capistrant: Is this wording out of date now that there are potential for upgrade segments that are appended during the replace as well as upgrade segments that are pass through for incremental compaction?

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
cecemei: there's no logic change here, just made the code more defensive.  updated the comment here. ihmo, `CompactionStatus` is not the best class to use here.

indexing-service/src/test/java/org/apache/druid/indexing/common/actions/MarkSegmentToUpgradeActionTest.java
cecemei: updated

processing/src/main/java/org/apache/druid/timeline/partition/LinearShardSpec.java
cecemei: this is from earlier implementation, reverted the 1 suffix

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
cecemei: true, updated the wording

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
cecemei: updated

server/src/main/java/org/apache/druid/server/compaction/CompactionMode.java
cecemei: done

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
cecemei: yea it looks odds because intellij styles method call wrapping takes precedence over binary expression continuation, fixed it

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
cecemei: actually i guess there's some logic change, the policy check has been moved to `CompactionConfigBasedJobTemplate` since creating the job needs to know the compaction mode (which is decided by the policy), but still the `computeCompactionStatus` only returns running/skipped/pending state.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -202,6 +235,125 @@ public void test_ingestDayGranularity_andCompactToMonthGranularity_andCompactToY
...
+                                                   .stream()
+                                                   .map(StringDimensionSchema::new)
+                                                   .collect(Collectors.toUnmodifiableList())))
+            .withTaskContext(Map.of("useConcurrentLocks", true))
+            .withIoConfig(new UserCompactionTaskIOConfig(true))
+            .withTuningConfig(UserCompactionTaskQueryTuningConfig.builder().partitionsSpec(partitionsSpec).build())
+            .build();
+
+    runCompactionWithSpec(dayGranularityConfig);
+    Thread.sleep(2_000L);
```
kfaraz: Please try to avoid using sleeps in embedded tests. Use the `LatchableEmitter` and watch for applicable events instead.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -314,15 +466,19 @@ public void test_cascadingCompactionTemplate_multiplePeriodsApplyDifferentCompac
...
     if (compactionEngine == CompactionEngine.NATIVE) {
-      ruleProvider = ruleProvider.ioConfigRules(
-          List.of(new ReindexingIOConfigRule("dropExisting", null, Period.days(7), new UserCompactionTaskIOConfig(true)))
-      );
+      ruleProvider = ruleProvider.ioConfigRules(List.of(new ReindexingIOConfigRule(
+          "dropExisting",
+          null,
+          Period.days(7),
+          new UserCompactionTaskIOConfig(true)
+      )));
```
kfaraz: Please avoid these unnecessary reformats, they bloat up the PR.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -547,12 +664,8 @@ public void test_compactionWithTransformFilteringAllRows_createsTombstones(
...
     String sql = StringUtils.format("SELECT COUNT(*) as cnt FROM \"%s\"", dataSource);
-    String result = cluster.callApi().onAnyBroker(b -> b.submitSqlQuery(new ClientSqlQuery(sql, null, false, false, false, null, null)));
-    List<Map<String, Object>> rows = JacksonUtils.readValue(
-        new DefaultObjectMapper(),
-        result.getBytes(StandardCharsets.UTF_8),
-        new TypeReference<>() {}
-    );
+    ClientSqlQuery clientSqlQuery = new ClientSqlQuery(sql, null, false, false, false, null, null);
+    List<Map<String, Object>> rows = parse(cluster.callApi().onAnyBroker(b -> b.submitSqlQuery(clientSqlQuery)));
     return ((Number) rows.get(0).get("cnt")).intValue();
```
kfaraz: The contents of this method can be replaced with:

```java
return Integer.parseInt(cluster.runSql("SELECT COUNT(*) FROM %s", dataSource));
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -129,19 +153,28 @@ public EmbeddedDruidCluster createCluster()
   }
 
 
-  private void configureCompaction(CompactionEngine compactionEngine)
+  private void configureCompaction(CompactionEngine compactionEngine, @Nullable CompactionCandidateSearchPolicy policy)
   {
     final UpdateResponse updateResponse = cluster.callApi().onLeaderOverlord(
-        o -> o.updateClusterCompactionConfig(new ClusterCompactionConfig(1.0, 100, null, true, compactionEngine, true))
+        o -> o.updateClusterCompactionConfig(new ClusterCompactionConfig(
```
kfaraz: Reformat is not needed here.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentUpgradeAction.java
```diff
@@ -0,0 +1,103 @@
...
+ * This action is used during compaction to track which segments are being replaced.
+ * It validates that all segments to be upgraded are covered by
+ * {@link ReplaceTaskLock}s before inserting them into the upgrade segments table.
+ * <p>
+ * The action will fail if any of the upgrade segments do not have a corresponding
+ * replace lock, ensuring that only properly locked segments can be marked for upgrade.
+ *
+ * @return the number of segments successfully inserted into the upgrade segments table
+ */
+public class SegmentUpgradeAction implements TaskAction<Integer>
```
kfaraz: This task action doesn't actually upgrade the segments, it only marks them to be upgraded.

```suggestion
public class MarkSegmentToUpgradeAction implements TaskAction<Integer>
```

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/TaskAction.java
```diff
@@ -37,6 +37,7 @@
     @JsonSubTypes.Type(name = "segmentTransactionalInsert", value = SegmentTransactionalInsertAction.class),
     @JsonSubTypes.Type(name = "segmentTransactionalAppend", value = SegmentTransactionalAppendAction.class),
     @JsonSubTypes.Type(name = "segmentTransactionalReplace", value = SegmentTransactionalReplaceAction.class),
+    @JsonSubTypes.Type(name = "segmentUpgrade", value = SegmentUpgradeAction.class),
```
kfaraz:
```suggestion
    @JsonSubTypes.Type(name = "markSegmentsToUpgrade", value = SegmentUpgradeAction.class),
```

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentUpgradeAction.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @Override
+  public Integer perform(Task task, TaskActionToolbox toolbox)
+  {
+    final String datasource = task.getDataSource();
+    final Map<DataSegment, ReplaceTaskLock> segmentToReplaceLock
+        = TaskLocks.findReplaceLocksCoveringSegments(datasource, toolbox.getTaskLockbox(), Set.copyOf(upgradeSegments));
+
+    if (segmentToReplaceLock.size() < upgradeSegments.size()) {
+      throw new IAE(
```
kfaraz: Use `InvalidInput.exception` or some other DruidException instead.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionIntervalSpec.java
```diff
@@ -42,18 +44,52 @@ public class CompactionIntervalSpec implements CompactionInputSpec
...
 
+  public CompactionIntervalSpec(Interval interval, String sha256OfSortedSegmentIds)
+  {
+    this(interval, null, sha256OfSortedSegmentIds);
+  }
+
   @JsonCreator
   public CompactionIntervalSpec(
       @JsonProperty("interval") Interval interval,
+      @JsonProperty("uncompactedSegments") @Nullable
```
kfaraz: Instead of having to deal with the nullables, please leave this class unchanged.
Add a new class which implements `CompactionInputSpec` and contains both an `Interval` and a List of segments. Using a separate dedicated class helps clarify the intent of minor compaction better.

server/src/main/java/org/apache/druid/indexing/overlord/IndexerMetadataStorageCoordinator.java
```diff
@@ -387,6 +387,29 @@ SegmentPublishResult commitAppendSegmentsAndMetadata(
...
+   * In incremental compaction scenarios:
+   * <ul>
+   * <li>Some segments may already meet compaction criteria (e.g., already compacted, correct partitioning)
+   * and don't need to be rewritten.</li>
+   * <li>These segments are tracked in the upgrade_segments table via this method, mapped to their
+   * corresponding REPLACE lock.</li>
+   * <li>When the compaction task finishes via {@link #commitReplaceSegments}, these segments are
+   * upgraded directly to the new version to maintain version consistency across the interval.</li>
+   * <li>After the task completes, entries are cleaned up via {@link #deleteUpgradeSegmentsForTask}.</li>
+   * </ul>
```
kfaraz: This method need not be aware of whether this is being done for compaction or some other purpose.
These details may be captured in `CompactionTask` or some other class which actually triggers the incremental compaction.

```suggestion
   * Inserts entries into the upgrade_segments table for segments that need to be upgraded
   * to a new version when a REPLACE task finishes using {@link #commitReplaceSegments}.
```

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -151,6 +151,12 @@ public CompactionStatistics getUncompactedStats()
            ? null : currentStatus.getUncompactedStats();
   }
 
+  @Nullable
+  public List<DataSegment> getUncompactedSegments()
+  {
+    return currentStatus == null ? null : currentStatus.getUncompactedSegments();
```
kfaraz: `currentStatus` cannot be null anymore.
I think we can skip adding this new method altogether and just do `candidate.getCurrentStatus().getUncompactedSegments()` unless it is used in a large number of places and we really do need a shorthand.

server/src/main/java/org/apache/druid/server/compaction/CompactionMode.java
```diff
@@ -27,5 +27,9 @@ public enum CompactionMode
   /**
    * Indicates that all existing segments of the interval will be picked for compaction.
    */
-  FULL_COMPACTION;
+  FULL_COMPACTION,
+  /**
+   * Indicates that only uncompacted segments of the interval will be picked for compaction.
+   */
+  INCREMENTAL_COMPACTION;
```
kfaraz: I think we should use the enum values as `ALL_SEGMENTS` and `UNCOMPACTED_SEGMENTS_ONLY`,
as that clarifies the intent better.

(Other names could be `MINOR` vs `MAJOR` as that is an accepted term for compaction in databases, but I don't think that's really needed since the mode concept is never really exposed to the user (except as a configuration in the search policy).)

The term "incremental" can be ambiguous as all auto-compaction is incremental, in the sense that it
always works one-interval-at-a-time.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -69,11 +72,20 @@ public MostFragmentedIntervalFirstPolicy(
         "'minUncompactedCount'[%s] must be greater than 0",
         maxAverageUncompactedBytesPerSegment
     );
+    InvalidInput.conditionalException(
+        incrementalCompactionUncompactedRatioThreshold == null
```
kfaraz: Nit: Rather than doing the null check inside the `conditionalException`, put it in an `if` and put the `conditionException` check inside the if block.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -47,13 +47,16 @@ public class MostFragmentedIntervalFirstPolicy extends BaseCandidateSearchPolicy
...
   private final HumanReadableBytes maxAverageUncompactedBytesPerSegment;
+  private final double incrementalCompactionUncompactedRatioThreshold;
 
   @JsonCreator
   public MostFragmentedIntervalFirstPolicy(
       @JsonProperty("minUncompactedCount") @Nullable Integer minUncompactedCount,
       @JsonProperty("minUncompactedBytes") @Nullable HumanReadableBytes minUncompactedBytes,
       @JsonProperty("maxAverageUncompactedBytesPerSegment") @Nullable
       HumanReadableBytes maxAverageUncompactedBytesPerSegment,
+      @JsonProperty("incrementalCompactionUncompactedRatioThreshold") @Nullable
```
kfaraz: Maybe consider a more declarative name for this field like `minUncompactedBytesPercentForFullCompaction`, so that it aligns with existing field names. Use percentages as they are easier to visualize.

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
```diff
@@ -106,6 +118,17 @@ public HumanReadableBytes getMaxAverageUncompactedBytesPerSegment()
     return maxAverageUncompactedBytesPerSegment;
   }
 
+  /**
+   * Threshold ratio of uncompacted bytes to compacted bytes below which
```
kfaraz: It is the ratio between uncompacted bytes and total segment bytes in the interval.

server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
```diff
@@ -463,14 +470,30 @@ private static ClientCompactionTaskQuery compactSegments(
     context.put("priority", compactionTaskPriority);
 
     final String taskId = IdUtils.newTaskId(TASK_ID_PREFIX, ClientCompactionTaskQuery.TYPE, dataSource, null);
+    final ClientCompactionIntervalSpec clientCompactionIntervalSpec;
+    switch (compactionMode) {
+      case FULL_COMPACTION:
+        clientCompactionIntervalSpec = new ClientCompactionIntervalSpec(entry.getCompactionInterval(), null, null);
+        break;
+      case INCREMENTAL_COMPACTION:
```
kfaraz: As discussed earlier in the old PR, we should avoid supporting incremental compaction in the Coordinator-based `CompactSegments` duty to drive adoption of the Overlord-based compaction supervisors.
This will allow us to deprecate and eventually remove Coordinator-based compaction duty.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionConfigBasedJobTemplate.java
```diff
@@ -91,12 +92,30 @@ public List<CompactionJob> createCompactionJobs(
     // Create a job for each CompactionCandidate
     while (segmentIterator.hasNext()) {
       final CompactionCandidate candidate = segmentIterator.next();
+      final CompactionCandidateSearchPolicy.Eligibility eligibility =
+          params.getClusterCompactionConfig().getCompactionPolicy().checkEligibilityForCompaction(candidate, null);
```
kfaraz: We shouldn't pass null for the task status.
Please do the following instead:
1. Write a functional interface `CompactionTaskStatusProvider` with a single method:
```java
Optional<CompactionTaskStatus> getLatestTaskStatus(CompactionCandidate candidate);
```
2. Pass a lambda for this interface into the `CompactionJobParams`. The lambda would delegate the actual retrieval to the `CompactionStatusTracker`.
3. Use the functional interface in the `CompactionConfigBasedJobTemplate` to get the optional task status.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionConfigBasedJobTemplate.java
```diff
@@ -91,12 +92,30 @@ public List<CompactionJob> createCompactionJobs(
...
+        continue;
+      }
+      final CompactionCandidate finalCandidate;
+      switch (eligibility.getMode()) {
+        case FULL_COMPACTION:
+          finalCandidate = candidate;
+          break;
+        case INCREMENTAL_COMPACTION:
+          finalCandidate = CompactionCandidate.from(candidate.getUncompactedSegments(), null)
+                                              .withCurrentStatus(candidate.getCurrentStatus());
```
kfaraz: Why is this needed?

kfaraz (commented):
Leaving a partial review for the supporting changes.
Will review the core changes in `CompactionTask` and `IndexerSQLMetadataStorageCoordinator` shortly.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentUpgradeAction.java
cecemei: done

server/src/main/java/org/apache/druid/indexing/overlord/IndexerMetadataStorageCoordinator.java
cecemei: done

server/src/main/java/org/apache/druid/server/compaction/CompactionMode.java
cecemei: done

server/src/main/java/org/apache/druid/server/compaction/CompactionMode.java
cecemei: done

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
cecemei: done

server/src/main/java/org/apache/druid/server/compaction/MostFragmentedIntervalFirstPolicy.java
cecemei: done

server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
cecemei: this method is used by both coordinator and overlord, coordinator would always use ALL_SEGMENTS compaction mode in line 266.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionIntervalSpec.java
cecemei: updated

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/MarkSegmentToUpgradeAction.java
```diff
@@ -0,0 +1,103 @@
...
+    {
+    };
+  }
+
+  @Override
+  public Integer perform(Task task, TaskActionToolbox toolbox)
+  {
+    final String datasource = task.getDataSource();
+    final Map<DataSegment, ReplaceTaskLock> segmentToReplaceLock
+        = TaskLocks.findReplaceLocksCoveringSegments(datasource, toolbox.getTaskLockbox(), Set.copyOf(upgradeSegments));
```
kfaraz: Since we use a set here, I wonder if the constructor of this class should also accept a `Set` instead of `List`.
Otherwise, it is possible for users to pass in duplicate segments and then they would an exception `Not all segments are hold by replace lock`, which might be confusing.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/MarkSegmentToUpgradeAction.java
```diff
@@ -0,0 +1,103 @@
...
+
+    if (segmentToReplaceLock.size() < upgradeSegments.size()) {
+      throw InvalidInput.exception(
+          "Not all segments are hold by a replace lock, only [%d] segments out of total segments[%d] are hold by repalce lock",
```
kfaraz:
```suggestion
          "Segments to upgrade must be covered by a REPLACE lock. Only [%d] out of [%d] segments are covered.",
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/MinorCompactionInputSpec.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.druid.timeline.DataSegment;
+import org.joda.time.Interval;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Specifies uncompacted segments to compact within an interval.
+ * Used for minor compaction to compact only uncompacted segments while leaving compacted segments untouched.
```
kfaraz:
```suggestion
 * Used for MSQ-based minor compaction to compact only uncompacted segments while upgrading compacted segments (i.e. no change to physical segment files).
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/UncompactedInputSpec.java
```diff
@@ -0,0 +1,132 @@
...
+
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Specifies uncompacted segments to compact within an interval.
+ * Used for minor compaction to compact only uncompacted segments while leaving compacted segments untouched.
+ */
+public class UncompactedInputSpec implements CompactionInputSpec
```
kfaraz: May be call it `MinorCompactionInputSpec`?

indexing-service/src/main/java/org/apache/druid/indexing/common/task/UncompactedInputSpec.java
```diff
@@ -0,0 +1,132 @@
...
+  private final List<SegmentDescriptor> uncompactedSegments;
+
+  @JsonCreator
+  public UncompactedInputSpec(
+      @JsonProperty("interval") Interval interval,
+      @JsonProperty("uncompactedSegments") List<SegmentDescriptor> uncompactedSegments
+  )
+  {
+    if (interval == null) {
+      throw new IAE("Uncompacted interval must not be null");
```
kfaraz: Please use `InvalidInput` exceptions instead.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/UncompactedInputSpec.java
```diff
@@ -0,0 +1,132 @@
...
+    // Validate that all segments are within the interval
+    List<SegmentDescriptor> segmentsNotInInterval =
+        uncompactedSegments.stream().filter(s -> !interval.contains(s.getInterval())).collect(Collectors.toList());
+    if (!segmentsNotInInterval.isEmpty()) {
+      throw new IAE(
+          "All uncompacted segments must be within interval[%s], got segments outside interval: %s",
+          interval,
+          segmentsNotInInterval
+      );
+    }
```
kfaraz: Nit: move this to a separate private method

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionSupervisorSpec.java
```diff
@@ -96,12 +97,12 @@ public CompactionSupervisor createSupervisor()
   /**
    * @return {@link CompactionJobTemplate} used to create jobs for the supervisor.
    */
-  public CompactionJobTemplate getTemplate()
+  public CompactionJobTemplate getTemplate(CompactionStatusTracker statusTracker)
```
kfaraz: Please don't make this change. This breaks the clean template APIs already in place.
Please follow the steps outlined here instead:
https://github.com/apache/druid/pull/19059#discussion_r2864067544

server/src/main/java/org/apache/druid/client/indexing/ClientCompactionIntervalSpec.java
```diff
@@ -20,43 +20,67 @@
...
+    }
+    this.uncompactedSegments = uncompactedSegments;
     this.sha256OfSortedSegmentIds = sha256OfSortedSegmentIds;
   }
 
   @JsonProperty
   public String getType()
   {
-    return TYPE;
+    return (uncompactedSegments == null) ? TYPE_ALL_SEGMENTS : TYPE_UNCOMPACTED_SEGMENTS_ONLY;
```
kfaraz, line 83: This seems unclean. Please add a separate class instead.

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -259,6 +259,12 @@ public boolean isRunning()
     return started.get();
   }
 
+  @Override
+  public CompactionStatusTracker getCompactionStatusTracker()
```
kfaraz: this shouldn't be required

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionConfigBasedJobTemplate.java
```diff
@@ -91,12 +96,35 @@ public List<CompactionJob> createCompactionJobs(
...
+      switch (eligibility.getMode()) {
+        case ALL_SEGMENTS:
+          finalCandidate = candidate;
+          break;
+        case UNCOMPACTED_SEGMENTS_ONLY:
+          finalCandidate = CompactionCandidate.from(
+              candidate.getUncompactedSegments(),
+              null,
+              candidate.getCurrentStatus()
+          );
```
kfaraz, line 112: Why is this change needed? Is the candidate being changed in any way?

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
```diff
@@ -110,7 +113,8 @@ public CascadingReindexingTemplate(
       @JsonProperty("taskContext") @Nullable Map<String, Object> taskContext,
       @JsonProperty("skipOffsetFromLatest") @Nullable Period skipOffsetFromLatest,
       @JsonProperty("skipOffsetFromNow") @Nullable Period skipOffsetFromNow,
-      @JsonProperty("defaultSegmentGranularity") Granularity defaultSegmentGranularity
+      @JsonProperty("defaultSegmentGranularity") Granularity defaultSegmentGranularity,
+      @JacksonInject CompactionStatusTracker statusTracker
```
kfaraz: The status tracker should not be passed here. It would be passed via `CompactionJobParams` in a manner similar to the `IndexingStateFingerprintMapper` interface. Please refer to the other comments regarding the `CompactionStatusTracker`.

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -2097,17 +2119,27 @@ private SegmentMetadata getSegmentMetadataFromSchemaMappingOrUpgradeMetadata(
...
+        transaction -> insertIntoUpgradeSegmentsTableDoWork(transaction, segmentToReplaceLock)
+    );
+  }
+
   /**
    * Inserts entries into the upgrade_segments table in batches of size
    * {@link #MAX_NUM_SEGMENTS_TO_ANNOUNCE_AT_ONCE}.
    */
-  private void insertIntoUpgradeSegmentsTable(
+  private int insertIntoUpgradeSegmentsTableDoWork(
```
kfaraz: Please do not rename this method if the intention is only to distinguish it from the new `insertIntoUpgradeSegmentsTable(map)` method. It is okay for two methods to have the same name if they serve a similar purpose but have different args.

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -2186,12 +2222,13 @@ private Map<String, String> getAppendSegmentsCommittedDuringTask(
     );
 
     ResultIterator<Pair<String, String>> resultIterator = transaction.getHandle()
```
kfaraz: Nit: Move the `.getHandle()` to the next line to make the reformatting changes smaller and the code cleaner (less indented).

processing/src/main/java/org/apache/druid/timeline/partition/ShardSpec.java
```diff
@@ -56,6 +57,15 @@
 })
 public interface ShardSpec
 {
+  /**
+   * Returns whether {@link #createChunk} returns a {@link NumberedPartitionChunk} instance.
+   * This is necessary for supporting {@link PartitionHolder#isComplete()} if updating to a new corePartitions spec.
+   */
+  default boolean isNumChunkSupported()
```
kfaraz:
```suggestion
  default boolean canCreateNumberedPartitionChunk()
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -572,17 +584,37 @@ static Map<Interval, DataSchema> createDataSchemasForIntervals(
       return Collections.emptyMap();
     }
 
+    if (segmentProvider.minorCompaction) {
+      Iterable<DataSegment> segmentsNotCompletelyWithinin =
+          Iterables.filter(timelineSegments, s -> !segmentProvider.interval.contains(s.getInterval()));
+      if (segmentsNotCompletelyWithinin.iterator().hasNext()) {
+        throw new ISE(
```
kfaraz: Please use `DruidException`.

Is this case possible? Isn't this validation already done in the new `CompactionInputSpec` class?

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -572,17 +584,37 @@ static Map<Interval, DataSchema> createDataSchemasForIntervals(
...
-                          .add(dataSegment);
+        if (segmentProvider.segmentsToUpgradePredicate.test(dataSegment)) {
+          upgradeSegments.add(dataSegment);
+        } else {
+          intervalToSegments.computeIfAbsent(dataSegment.getInterval(), k -> new ArrayList<>())
+                            .add(dataSegment);
+        }
+      }
+      if (!upgradeSegments.isEmpty()) {
+        toolbox.getTaskActionClient().submit(new MarkSegmentToUpgradeAction(segmentProvider.dataSource, upgradeSegments));
```
kfaraz: A log line here would be helpful.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -1222,11 +1274,26 @@ static class SegmentProvider
     private final CompactionInputSpec inputSpec;
     private final Interval interval;
 
+    private final boolean minorCompaction;
+    private final Predicate<DataSegment> segmentsToUpgradePredicate;
+    private final Predicate<DataSegment> segmentsToCompactPredicate;
```
kfaraz: Nit: Instead of predicates, please add private methods `shouldUpgradeSegment(DataSegment)` and `shouldCompactSegment(DataSegment)` to the `SegmentProvider` class.
Use of predicates does not really seem warranted here.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -627,18 +659,35 @@ static Map<Interval, DataSchema> createDataSchemasForIntervals(
             projections,
             needMultiValuedColumns
         );
-        intervalDataSchemaMap.put(interval, dataSchema);
+        inputSchemas.put(
+            segmentProvider.minorCompaction
+            ? new MultipleSpecificSegmentSpec(segmentsToCompact.stream()
+                                                               .map(DataSegment::toDescriptor)
+                                                               .collect(Collectors.toList()))
+            : new MultipleIntervalSegmentSpec(List.of(interval)), dataSchema);
```
kfaraz: Compute the value for the key `QuerySegmentSpec` in a separate line rather than computing it inside the `put`. Current code is difficult to follow.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -627,18 +659,35 @@ static Map<Interval, DataSchema> createDataSchemasForIntervals(
...
       }
-      return intervalDataSchemaMap;
+      return inputSchemas;
     } else {
       // given segment granularity
+      List<DataSegment> upgradeSegments = Lists.newArrayList(Iterables.filter(
+          timelineSegments,
+          segmentProvider.segmentsToUpgradePredicate
```
```suggestion
          segmentProvider::shouldUpgradeSegment
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -627,18 +659,35 @@ static Map<Interval, DataSchema> createDataSchemasForIntervals(
...
+            ? new MultipleSpecificSegmentSpec(segmentsToCompact.stream()
+                                                               .map(DataSegment::toDescriptor)
+                                                               .collect(Collectors.toList()))
+            : new MultipleIntervalSegmentSpec(List.of(interval)), dataSchema);
       }
-      return intervalDataSchemaMap;
+      return inputSchemas;
     } else {
       // given segment granularity
+      List<DataSegment> upgradeSegments = Lists.newArrayList(Iterables.filter(
```
```suggestion
      List<DataSegment> segmentsToUpgrade = Lists.newArrayList(Iterables.filter(
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -627,18 +659,35 @@ static Map<Interval, DataSchema> createDataSchemasForIntervals(
...
+          timelineSegments,
+          segmentProvider.segmentsToUpgradePredicate
+      ));
+      if (!upgradeSegments.isEmpty()) {
+        toolbox.getTaskActionClient().submit(new MarkSegmentToUpgradeAction(segmentProvider.dataSource, upgradeSegments));
```
kfaraz: Please add a log line here.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -648,7 +697,11 @@ static Map<Interval, DataSchema> createDataSchemasForIntervals(
           projections,
           needMultiValuedColumns
       );
-      return Collections.singletonMap(segmentProvider.interval, dataSchema);
+      return Map.of(segmentProvider.minorCompaction
+                    ? new MultipleSpecificSegmentSpec(StreamSupport.stream(segmentsToCompact.spliterator(), false)
+                                                                   .map(DataSegment::toDescriptor)
+                                                                   .collect(Collectors.toList()))
+                    : new MultipleIntervalSegmentSpec(List.of(segmentProvider.interval)), dataSchema);
```
kfaraz: Current code is difficult to follow. Please use `if-else` and a separate variable to clean it up.

```suggestion
      final QuerySegmentSpec segmentSpec;
      if (segmentProvider.isMinorCompaction) {
          segmentSpec = new MultipleSpecificSegmentSpec(
                  StreamSupport.stream(segmentsToCompact.spliterator(), false)
                                            .map(DataSegment::toDescriptor)
                                            .collect(Collectors.toList())
            );
      } else {
            segmentSpec = new MultipleIntervalSegmentSpec(List.of(segmentProvider.interval));
      }
      
      return Map.of(segmentSpec, dataSchema);
```

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NativeCompactionRunner.java
```diff
@@ -37,9 +37,11 @@
 import org.apache.druid.indexing.input.DruidInputSource;
 import org.apache.druid.java.util.common.IAE;
 import org.apache.druid.java.util.common.Intervals;
+import org.apache.druid.java.util.common.JodaUtils;
```
kfaraz, line 40: The validation in `NativeCompactionRunner.validateCompactionTask()` should return a `failure` if the `CompactionInputSpec` is of the new type `Uncompacted` (aka minor compaction).

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -1959,9 +1968,21 @@ private Set<DataSegmentPlus> createNewIdsOfAppendSegmentsAfterReplace(
               oldSegmentMetadata.getIndexingStateFingerprint()
           )
       );
+      segmentsToInsert.add(dataSegment);
     }
 
-    return upgradedSegments;
+    // update corePartitions in shard spec
+    return Pair.of(upgradedSegments, segmentsToInsert.stream().map(segment -> {
```
kfaraz: Please break up this statement to separate out the return and the computation.

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -1959,9 +1968,21 @@ private Set<DataSegmentPlus> createNewIdsOfAppendSegmentsAfterReplace(
...
+      Integer partitionNum = intervalToCurrentPartitionNum.get(segment.getInterval());
+      if (!segment.isTombstone()
+          && !numChunkNotSupported.contains(segment.getInterval())
+          && partitionNum != null
+          && partitionNum + 1 != segment.getShardSpec().getNumCorePartitions()) {
```
kfaraz: Why this condition?

Is it possible that existing compacted segments in an interval were range partitioned and had say 10 core partitions. And the segments created by the minor compaction are also range partitioned and have say 5 core partitions. In this case, none of the segments have core partition = 16, so will their shard spec remain unchanged?

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -1852,9 +1853,14 @@ protected Set<DataSegment> insertSegments(
...
-   * Creates new versions of segments appended while a "REPLACE" task was in progress.
+   * Retrieves segments from the upgrade segments table and creates upgraded versions with new intervals,
+   * versions, and partition numbers. Combines upgraded segments with replace segments and updates shard
+   * specs with correct core partition counts.
+   *
+   * @return pair of (upgraded segments for metadata tracking, segments to insert into segment table)
+   * @throws DruidException if a replace interval partially overlaps a segment being upgraded
    */
-  private Set<DataSegmentPlus> createNewIdsOfAppendSegmentsAfterReplace(
+  private Pair<Set<DataSegmentPlus>, Set<DataSegment>> createNewSegmentsAfterReplace(
```
kfaraz: This method should just return a `Set<DataSegmentPlus>`. Each `DataSegmentPlus` already contains a `DataSegment` inside it.
The `DataSegmentPlus.getUpgradedFromSegmentId()` should be used to distinguish if the segment is an upgraded one or not.

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -1925,15 +1933,16 @@ private Set<DataSegmentPlus> createNewIdsOfAppendSegmentsAfterReplace(
...
           newInterval,
           (i, value) -> value == null ? 0 : value + 1
       );
-      final int numCorePartitions = intervalToNumCorePartitions.get(newInterval);
-      ShardSpec shardSpec = new NumberedShardSpec(partitionNum, numCorePartitions);
-
+      final ShardSpec shardSpec = oldSegment.getShardSpec().withPartitionNum(partitionNum);
       // Create upgraded segment with the correct interval, version and shard spec
       String lockVersion = upgradeSegmentToLockVersion.get(oldSegment.getId().toString());
       DataSegment dataSegment = DataSegment.builder(oldSegment)
```
kfaraz, line 1952: Use the final shardSpec to build this `DataSegment`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -202,6 +229,133 @@ public void test_ingestDayGranularity_andCompactToMonthGranularity_andCompactToY
...
+    runCompactionWithSpec(dayGranularityConfig);
+    waitForAllCompactionTasksToFinish();
+
+    pauseCompaction(dayGranularityConfig);
+    Assertions.assertEquals(0, getNumSegmentsWith(Granularities.HOUR));
+    Assertions.assertEquals(1, getNumSegmentsWith(Granularities.DAY));
+
+    verifyCompactedSegmentsHaveFingerprints(dayGranularityConfig);
+
+    // Set up another topic and supervisor
```
kfaraz: Why do we need a separate topic and supervisor?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -202,6 +229,133 @@ public void test_ingestDayGranularity_andCompactToMonthGranularity_andCompactToY
...
+    waitForAllCompactionTasksToFinish();
+
+    // wait for new segments have been updated to the cache
+    overlord.latchableEmitter().waitForEvent(
+        event -> event.hasMetricName("segment/metadataCache/used/count")
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource)
+                      .hasValueMatching(Matchers.greaterThan(totalUsed)));
+
+    // performed incremental compaction: 1 previously compacted segment + 1 incrementally compacted segment
+    Assertions.assertEquals(2, getNumSegmentsWith(Granularities.DAY));
```
kfaraz, line 314: Please also add assertions to verify the results of some queries on the final compacted data.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -202,6 +229,133 @@ public void test_ingestDayGranularity_andCompactToMonthGranularity_andCompactToY
...
+    final KafkaSupervisorSpec supervisor1 = kafkaSupervisorSpecBuilder.withId(topic1).build(dataSource, topic1);
+    cluster.callApi().postSupervisor(supervisor1);
+
+    final int totalRowCount = publish1kRecords(topic1, true) + publish1kRecords(topic1, false);
+    waitUntilPublishedRecordsAreIngested(totalRowCount);
+
+    // Before compaction
+    Assertions.assertEquals(4, getNumSegmentsWith(Granularities.HOUR));
+
+    PartitionsSpec partitionsSpec = new DimensionRangePartitionsSpec(null, 5000, List.of("page"), false);
```
kfaraz: Please add a test case for dynamic partitioning too.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
cecemei: no need for a separate topic, removed

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
cecemei: added parameterized test

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
cecemei: updated

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/MarkSegmentToUpgradeAction.java
cecemei: updated

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
cecemei: not sure i understand, the builder is already using the final shardSpec.

kfaraz: Yes, but the number of core partitions is updated later in the code flow, right?
That would effectively change the shard spec.
Even though the number of core partitions is not used to determine segment ID, it would still be cleaner to update the inner `DataSegment` object (or maybe construct it lazily) so that it uses the correct shard spec.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: timelineSegments is not user input, but the segments overlapping with the interval. the interval can overlap with some segments, in this case we cant simiply upgrade the overlapping segment or include it in the query. this is an edge case we discussed before.

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: done

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: done

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: done

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: done

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: done

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
cecemei: done

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionConfigBasedJobTemplate.java
cecemei: yes the candidate is now reconstructed with only uncompactedsegments

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
cecemei: updated

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
cecemei: updated

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
cecemei: updated

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
cecemei: done

processing/src/main/java/org/apache/druid/timeline/partition/ShardSpec.java
cecemei: done

server/src/main/java/org/apache/druid/client/indexing/ClientCompactionIntervalSpec.java
cecemei: whats the point of `ClientCompactionIntervalSpec` if every class needs to be mapped to CompactionIntervalSpec class?

indexing-service/src/main/java/org/apache/druid/indexing/common/task/NativeCompactionRunner.java
cecemei: done

indexing-service/src/main/java/org/apache/druid/indexing/common/task/CompactionTask.java
```diff
@@ -588,33 +586,37 @@ static Map<QuerySegmentSpec, DataSchema> createInputDataSchemas(
...
+          segmentsToUpgrade.add(dataSegment);
         } else {
           intervalToSegments.computeIfAbsent(dataSegment.getInterval(), k -> new ArrayList<>())
                             .add(dataSegment);
         }
       }
-      if (!upgradeSegments.isEmpty()) {
-        toolbox.getTaskActionClient().submit(new MarkSegmentToUpgradeAction(segmentProvider.dataSource, upgradeSegments));
+      if (!segmentsToUpgrade.isEmpty()) {
+        log.info("Marking [%d]segments to upgrade", segmentsToUpgrade.size());
```
kfaraz: I think we should include some (if not all) of the segment IDs in the log too. It might be useful in debugging.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -206,6 +231,129 @@ public void test_ingestDayGranularity_andCompactToMonthGranularity_andCompactToY
...
+    // published another 1k
+    final int appendedRowCount = publish1kRecords(topic1, true);
+    indexer.latchableEmitter().flush();
+    waitUntilPublishedRecordsAreIngested(appendedRowCount);
+
+    // Tear down both topics and supervisors
+    kafkaServer.deleteTopic(topic1);
+    cluster.callApi().postSupervisor(supervisor1.createSuspendedSpec());
+
+    long totalUsed = overlord.latchableEmitter().getMetricValues(
```
kfaraz, line 294: This need not give the correct value of total used segments currently present in the cache.
This metric is emitted after every cache sync, and as such the total aggregate would be higher than the actual number of segments.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -206,6 +231,129 @@ public void test_ingestDayGranularity_andCompactToMonthGranularity_andCompactToY
...
+    Assertions.assertEquals(0, getNumSegmentsWith(Granularities.HOUR));
+    // 1 compacted segment + 2 appended segment
+    Assertions.assertEquals(3, getNumSegmentsWith(Granularities.DAY));
+    Assertions.assertEquals("3000", cluster.runSql("SELECT COUNT(*) FROM %s", dataSource));
+
+    runCompactionWithSpec(dayGranularityConfig);
+    waitForAllCompactionTasksToFinish();
+
+    // wait for new segments have been updated to the cache
+    overlord.latchableEmitter().waitForEvent(
```
kfaraz, line 308: If the intent is to wait for the __next__ cache sync, do the following instead:

```java
overlord.latchableEmitter().waitForNextEvent(
    event -> event.hasMetricName("segment/metadataCache/sync/time")
)
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -373,7 +521,7 @@ public void test_cascadingReindexing_withVirtualColumnOnNestedData_filtersCorrec
     cluster.callApi().runTask(task.withId(IdUtils.getRandomId()), overlord);
     cluster.callApi().waitForAllSegmentsToBeAvailable(dataSource, coordinator, broker);
 
-    Assertions.assertEquals(4, getTotalRowCount());
```
kfaraz, line 376: I think it was cleaner to use the `getTotalRowCount()` method.
Just update the contents of that method to use `cluster.runSql("SELECT COUNT(*)")`.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -626,12 +735,8 @@ private void verifyNoRowsWithNestedValue(String nestedColumn, String field, Stri
...
         value
     );
-    String result = cluster.callApi().onAnyBroker(b -> b.submitSqlQuery(new ClientSqlQuery(sql, null, false, false, false, null, null)));
-    List<Map<String, Object>> rows = JacksonUtils.readValue(
-        new DefaultObjectMapper(),
-        result.getBytes(StandardCharsets.UTF_8),
-        new TypeReference<>() {}
-    );
+    ClientSqlQuery clientSqlQuery = new ClientSqlQuery(sql, null, false, false, false, null, null);
+    List<Map<String, Object>> rows = parse(cluster.callApi().onAnyBroker(b -> b.submitSqlQuery(clientSqlQuery)));
```
kfaraz: `parse` method should not be needed. Ideally, we should be simply using `cluster.runSql()` for all of these cases.

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -2169,13 +2222,13 @@ private List<DataSegmentPlus> retrieveSegmentsById(
   }
 
   /**
-   * Finds the append segments that were covered by the given task REPLACE locks.
-   * These append segments must now be upgraded to the same version as the segments
+   * Finds segments were covered by the given task REPLACE locks.
```
kfaraz:
```suggestion
   * Finds segments that were covered by the given task REPLACE locks.
```

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -2169,13 +2222,13 @@ private List<DataSegmentPlus> retrieveSegmentsById(
...
+   * These segments must now be upgraded to the same version as the segments
    * being committed by this replace task.
    *
    * @return Map from append Segment ID to REPLACE lock version
    */
-  private Map<String, String> getAppendSegmentsCommittedDuringTask(
+  private Map<String, String> getSegmentsCoveredByTaskLock(
```
kfaraz:
```suggestion
  private Map<String, String> getSegmentsCoveredByReplaceTaskLock(
```

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -1867,36 +1874,40 @@ private Set<DataSegmentPlus> createNewIdsOfAppendSegmentsAfterReplace(
...
                                                 .map(ReplaceTaskLock::getSupervisorTaskId)
                                                 .findFirst().orElse(null);
     final Map<String, String> upgradeSegmentToLockVersion
-        = getAppendSegmentsCommittedDuringTask(transaction, taskId);
+        = getSegmentsCoveredByTaskLock(transaction, taskId);
 
     final List<DataSegmentPlus> segmentsToUpgrade
         = retrieveSegmentsById(dataSource, transaction, upgradeSegmentToLockVersion.keySet());
 
     if (segmentsToUpgrade.isEmpty()) {
```
kfaraz, line 1903: A short comment might help here:

```suggestion
    // If there is nothing to upgrade, return the replaceSegments unchanged
    if (segmentsToUpgrade.isEmpty()) {
```

server/src/main/java/org/apache/druid/metadata/IndexerSQLMetadataStorageCoordinator.java
```diff
@@ -1959,9 +1972,35 @@ private Set<DataSegmentPlus> createNewIdsOfAppendSegmentsAfterReplace(
...
       );
+      segmentsToInsert.add(dataSegment);
     }
 
-    return upgradedSegments;
+    return segmentsToInsert.stream().map(segment -> {
+      // update corePartitions in shard spec
+      Integer partitionNum = intervalToCurrentPartitionNum.get(segment.getInterval());
+      if (!segment.isTombstone() && !numChunkNotSupported.contains(segment.getInterval()) && partitionNum != null) {
+        return segment.withShardSpec(segment.getShardSpec().withCorePartitions(partitionNum + 1));
```
kfaraz, line 1983: This doesn't seem correct. The number of core partitions should be the total number of segments i.e. replaceSegments + upgradedSegments.

cecemei: `intervalToCurrentPartitionNum` keeps track of the highest partition num we've seen in replaceSegments, in each upgrade segment, we increment the partitionNum in the map and use it as the partitionNum in the upgrade segment, so it's effectively `replaceSegments + upgradedSegments`.

kfaraz: I see, thanks for the clarification! I must have missed that in the diff.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
cecemei: the last value (not aggregate) should be number of used segments right?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
cecemei: wouldn't that be the same to wait for the published segments to show up in the sync?

kfaraz (approved):
LGTM 🚀 🚀

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
kfaraz: Yes, but there can be race conditions. There might be some delay before the cache is synced with the metadata store. It is better to wait for the actual event you want.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
kfaraz: Yes, this check itself is correct. The problem is the value of `totalUsed` that this check depends on.

PR #19060, opened 2026-02-26
DartControllerModule: Move ControllerThreadPool inside ActualModule.

gianm:
It was at the incorrect level of indentation, so it was provided always, even when Dart was not enabled.


kfaraz:
Thanks for the fix, @gianm !

PR #19062, opened 2026-02-26
Configurable Broker tier selection for Realtime servers

abhishekrb19:
Currently, the Broker uses the same tier selection and balancer strategies configured by `druid.broker.select.tier` and `druid.broker.balancer.type` for both Historicals and realtime servers. 

This patch allows operators to optionally override that behavior for realtime servers via the `druid.broker.realtime.select.tier`  and `druid.broker.realtime.balancer.type` properties. This is useful in environments where Historicals and Realtime servers are configured differently.

- If the new properties `druid.broker.realtime.select.tier` and `druid.broker.realtime.balancer.type` are not configured (default behavior), realtime servers will continue to use the default strategy or the one configured via `druid.broker.select.tier` and `druid.broker.select.balancer.type`, for backward compatibility.
- All the balancer and tier selection strategies are applicable to the realtime ones as well

Approach:
- Add a separate module `BrokerRealtimeSelectorModule` that's added to `CliBroker`
- This module is essentially named providers for realtime servers' - `TierSelectorStrategy` and `ServerSelectorStrategy`
- By default, if this isn't configured, the default ones configured from `druid.broker.balancer.type` and `druid.broker.select.tier` are applied for the realtime servers (backward compatible behavior)

While at it, I also added `toString()` and related helpers for `TierSelectorStrategy` implementations to improve debuggability and test verification.

I can document this configuration separately, since the relevant sections in the [existing docs](https://druid.apache.org/docs/latest/operations/mixed-workloads/#service-tiering) for `druid.broker.select.tier` could also use an update.

Release note:

Added `druid.broker.realtime.select.tier` and `druid.broker.realtime.balancer.type` on the Brokers to optionally override the Broker’s tier selection and balancer strategies for realtime servers. If these properties are unset (by default), realtime servers continue to use the existing `druid.broker.select` and `druid.broker.balancer` configurations that applies to both historical and realtime servers.
 

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/services/src/test/java/org/apache/druid/cli/CliBrokerTest.java
+++ b/services/src/test/java/org/apache/druid/cli/CliBrokerTest.java
@@ -29,6 +29,8 @@ import org.apache.druid.client.selector.ConnectionCountServerSelectorStrategy;
 import org.apache.druid.client.selector.CustomTierSelectorStrategy;
 import org.apache.druid.client.selector.HighestPriorityTierSelectorStrategy;
 import org.apache.druid.client.selector.LowestPriorityTierSelectorStrategy;
+import org.apache.druid.client.selector.PreferredTierSelectorStrategy;
+import org.apache.druid.client.selector.PreferredTierSelectorStrategyConfig;
 import org.apache.druid.client.selector.RandomServerSelectorStrategy;
 import org.apache.druid.client.selector.ServerSelectorStrategy;
 import org.apache.druid.client.selector.TierSelectorStrategy;
@@ -174,7 +176,7 @@ public class CliBrokerTest
 
 
   @Test
-  public void testServerSelectorStrategyFallbackToGlobal()
+  public void testServerSelectorStrategyFallbackToDefault()
   {
     final Properties properties = new Properties();
     properties.setProperty("druid.broker.select.tier", "custom");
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/druid/client/selector/PreferredTierSelectorStrategy.java
```diff
@@ -111,4 +114,17 @@ public <T> List<QueryableDruidServer> pick(
...
+  public PreferredTierSelectorStrategyConfig getConfig()
+  {
+    return config;
+  }
+
+  @Override
+  public String toString()
+  {
+    return "PreferredTierSelectorStrategy{" +
+           ", config=" + config +
```
aho135:
```suggestion
           "config=" + config +
```

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -81,15 +84,19 @@ public class BrokerServerView implements TimelineServerView
   public BrokerServerView(
       final QueryableDruidServer.Maker directDruidClientFactory,
       final FilteredServerInventoryView baseView,
-      final TierSelectorStrategy tierSelectorStrategy,
+      final TierSelectorStrategy historicalTierSelectorStrategy,
+      @Named(REALTIME_SELECTOR) final TierSelectorStrategy realtimeTierSelectorStrategy,
```
aho135: Can we also have `@Named` for historicalTierSelectorStrategy? This makes things more explicit in the test code

abhishekrb19: Hmm agreed, that would make it clearer for the historical one as well. However, it would require introducing a separate provider to bind the `@Named` variant and move the existing bindings from `CliBroker`, since the same class cannot be bound multiple times in Guice. I punted on that for now in favor of a brief comment here indicating where these bindings are injected from. Hopefully that clarifies it?

aho135: Ahh I see, thanks for clarifying. Makes sense, and the comments help!

aho135 (approved):

services/src/test/java/org/apache/druid/cli/CliBrokerTest.java
```diff
@@ -0,0 +1,216 @@
...
+    properties.setProperty("druid.broker.select.tier", "lowestPriority");
+
+    final Injector injector = makeBrokerInjector(properties);
+    Assert.assertTrue(injector.getInstance(TierSelectorStrategy.class) instanceof LowestPriorityTierSelectorStrategy);
+    Assert.assertTrue(
+        injector.getInstance(Key.get(TierSelectorStrategy.class, Names.named(BrokerServerView.REALTIME_SELECTOR)))
+        instanceof LowestPriorityTierSelectorStrategy
+    );
+  }
+
```
aho135, line 94: Could we have one additional test case for the Preferred strategy just to validate the bindings? Thanks!

abhishekrb19: Thanks for calling that out! Added tests for `PreferredTierSelectorStrategy` as well.

PR #19063, opened 2026-02-27
fix: Planning for CASE-filtered SUM with ELSE 0.

### LINKED ISSUE #17768: Null query result when using SQL CASE expression
### Affected Version

Druid 32.0.0

### Description

After upgrading to version 32, where the `druid.generic.useDefaultValueForNull=true` feature is deprecated, we’re getting incorrect results when executing SQL queries using `CASE`. 
The query is supposed to return a result based on a given condition, but when a type value doesn’t match any condition, it returns NULL instead of the expected ELSE value or COALESCE result.

```
SELECT
        t.id,
        SUM(CASE WHEN type = 'X'
            THEN NVL(amount, 0)
            ELSE 0
        END) AS total_amount
    FROM
        example t
    GROUP BY t.id
```

We followed all the steps in the [migration guide](https://druid.apache.org/docs/latest/release-info/migr-ansi-sql-null/), and no null values were inserted during ingestion. All raw data contains non-null values.

Image

gianm:
Fixes #17768.

Previously, when `extendedFilteredSumRewrite: true` (the default), the aggregator `SUM(CASE WHEN COND THEN COL1 ELSE 0 END)` would be planned into `SUM(COL1) FILTER(WHERE COND)`. This creates a discrepancy in the situation where the underlying relation is nonempty but COND does not match any rows: SQL standard behavior is for the aggregator to return zero, but we returned NULL.

This patch fixes it by wrapping such aggregators in `COALESCE(_, CASE WHEN COUNT(*) > 0 THEN 0 END)`.


FrankChen021:
- processing/src/main/java/org/apache/druid/query/aggregation/FilteredAggregatorFactory.java:287 - [P1] Avoid empty IntervalDimFilter for elseValue no-match segments. When an interval filter has no overlap with the segment and elseValue is non-null, this no longer returns SuppressedAggregatorFactory and falls through to construct a new IntervalDimFilter with effectiveFilterIntervals empty. IntervalDimFilter rejects empty interval lists, so queries planned as $SUM_FILTER_ELSE over a time predicate outside a segment can fail during per-segment optimization instead of returning the intended else/null result. Handle the all-excluded case with a false filter or another aggregator that can emit elseValue without creating an empty IntervalDimFilter.

gianm:
> Findings that could not be attached inline:
>     * processing/src/main/java/org/apache/druid/query/aggregation/FilteredAggregatorFactory.java:287 - [P1] Avoid empty IntervalDimFilter for elseValue no-match segments.

Fixed this in the latest commit, and added tests to `FilteredAggregatorFactoryTest` and `filtered_sum.iq`.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

- processing/src/main/java/org/apache/druid/segment/projections/Projections.java:332 - [P2] Preserve elseValue when remapping filtered projection aggregators. This projection rewrite unwraps a FilteredAggregatorFactory, remaps its filter, then recreates it with the two-argument constructor, which drops the new elseValue. A $SUM_FILTER_ELSE query that is satisfied from an aggregate projection can therefore lose the PR's zero-on-nonempty-no-match behavior and return NULL again. Recreate the filtered factory with filteredQueryAgg.getElseValue() (and the correct name if needed) when building remappedFilteredAgg.

---

gianm:
> processing/src/main/java/org/apache/druid/segment/projections/Projections.java:332 - [P2] Preserve elseValue when remapping filtered projection aggregators. This projection rewrite unwraps a FilteredAggregatorFactory, remaps its filter, then recreates it with the two-argument constructor, which drops the new elseValue. A $SUM_FILTER_ELSE query that is satisfied from an aggregate projection can therefore lose the PR's zero-on-nonempty-no-match behavior and return NULL again. Recreate the filtered factory with filteredQueryAgg.getElseValue() (and the correct name if needed) when building remappedFilteredAgg.

Fair enough, fixed it.

FrankChen021 (commented):

---

FrankChen021 (commented):

---

FrankChen021:
Found two related correctness issues while re-reviewing, both stemming from `elseValue` being typed as `Long` regardless of the aggregator's actual value type:

- `sql/src/main/java/org/apache/druid/sql/calcite/aggregation/builtin/SumFilterElseSqlAggregator.java:112`

```java
// The producing rule only emits this aggregate when elseValue is an integer literal of value zero,
// so Long is a safe target type.
final Number elseValue = ((RexLiteral) elseRex).getValueAs(Long.class);
```

  This is only safe if the *value type* of the SUM is also integral. But a few lines later, `valueType` (`Calcites.getColumnTypeForRelDataType(aggregateCall.getType())`) can resolve to FLOAT/DOUBLE, and `sumFactory` is built as a `DoubleSumAggregatorFactory`/`FloatSumAggregatorFactory` accordingly. `elseValue` stays a boxed `Long(0)` even though the delegate aggregator is a double/float sum.

  Repro: `SELECT SUM(CASE WHEN dim1 = 'nonexistent' THEN m1 ELSE 0 END) FROM foo` where `m1` is DOUBLE, over a nonempty relation where the filter matches zero rows. The result comes back as a boxed `Long` instead of `Double`.

- `processing/src/main/java/org/apache/druid/query/aggregation/FilteredAggregator.java:62` (and the same pattern in `FilteredBufferAggregator.get()` / `FilteredVectorAggregator.get()`):

```java
public Object get()
{
  if (elseValue != null && hasUnmatchedRow && delegate.isNull()) {
    return elseValue;
  } else {
    return delegate.get();
  }
}
```

  `get()` returns the raw `elseValue` `Number` unmodified, whereas `getFloat()`/`getLong()`/`getDouble()` correctly coerce via `.floatValue()`/`.longValue()`/`.doubleValue()`. So `get()` is the odd one out — it can hand back a boxed type that doesn't match what the column's declared type promises. This is reachable independent of issue 1 too: `elseValue` is `@JsonProperty`-deserialized on `FilteredAggregatorFactory` with no validation against the delegate's result type, so a hand-written native query combining a `doubleSum`/`floatSum` delegate with an integer `elseValue` hits the same mismatch.

Suggest coercing `elseValue` to `valueType` at construction in `SumFilterElseSqlAggregator` (fixes the root cause for the current SQL-only producer), and/or coercing in each `get()` to close the gap for any future/native caller.

PR #19067, opened 2026-02-27
add SQL parameter values in RequestLogLine so they can be logged/emitted

clintropolis:
Currently SQL dynamic parameter values are not logged or emitted. This PR adds them to `RequestLogLine` so they can be logged and emitted with all the other stuff. This is useful for reconstructing the query when troubleshooting errors, since otherwise we only have the SQL with a bunch of `?`.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/log/DefaultRequestLogEvent.java
+++ b/server/src/main/java/org/apache/druid/server/log/DefaultRequestLogEvent.java
@@ -19,6 +19,7 @@
 
 package org.apache.druid.server.log;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/druid/server/RequestLogLine.java
```diff
@@ -76,7 +81,19 @@ public static RequestLogLine forSql(
       QueryStats queryStats
   )
   {
-    return new RequestLogLine(null, sql, sqlQueryContext, timestamp, remoteAddr, queryStats);
+    return forSql(sql, null, sqlQueryContext, timestamp, remoteAddr, queryStats);
+  }
+
+  public static RequestLogLine forSql(
+      String sql,
+      List<ClientSqlParameter> parameters,
```
abhishekrb19, line 89: nit: add `@Nullable` since we explicitly pass null in the above call

sql/src/main/java/org/apache/druid/sql/SqlQueryPlus.java
```diff
@@ -134,6 +136,20 @@ public List<TypedValue> parameters()
     return parameters;
   }
 
+  /**
+   * Convert parameters list to serde friendly {@link SqlParameter}
```
abhishekrb19, line 140:
```suggestion
   * Convert parameters list to serde friendly {@link ClientSqlParameter}
```

PR #19069, opened 2026-02-28
Configurable brokerService for console queries

aho135:
### Description
When using the manual routing strategy in combination with `druid.broker.segment.watchedTiers` we are losing console visibility for Historicals. This can happen with the combination of configurations

Router
```
druid.router.sql.enable: true
druid.router.strategies: [{"type":"manual","defaultManualBrokerService":"druid/broker1"}]
druid.router.tierToBrokerMap: {"_default_tier": "druid/broker1", "broker2": "druid/broker2"}
```
broker1
```
druid.broker.segment.watchedTiers=["t1","t2"]
```
broker2
```
druid.broker.segment.watchedTiers=["t1","t2","t3","t4"]
```

Since the Router query goes to druid/broker1 which has limited segment visibility, the console will incorrectly report Historicals in t3/t4 as having no data. This patch allows us to configure Router console queries to go to druid/broker2 instead so that the console can have full visibility over Historicals.

#### Release note
Allows operators to configure brokerService for Router console queries. This can be accomplished by setting something like 
```
consoleBrokerService: 'druid/broker2',
```
in web-console/console-config.js

<hr>

 * `druid-query.ts`
 * `entry.tsx`

<hr>

This PR has:

- [X] been self-reviewed.

### DIFF
--- a/web-console/src/druid-models/web-console-config/web-console-config.tsx
+++ b/web-console/src/druid-models/web-console-config/web-console-config.tsx
@@ -43,8 +43,8 @@ export const WEB_CONSOLE_CONFIG_FIELDS: Field<WebConsoleConfig>[] = [
       <>
         Sets brokerService in the query context for console system queries
         (e.g.,&quot;druid/broker-all-tiers&quot;). Useful when configuring
-        druid.broker.segment.watchedTiers if the default Broker server doesn&apos;t have visibility
-        into some Historical tiers. Does not affect user-initiated queries.
+        druid.broker.segment.watchedTiers if the default Broker server does not have visibility into
+        some Historical tiers. Does not affect user-initiated queries.
       </>
     ),
   },


abhishek-chouhan:
There is also the case where broker1 watches ["t1","t2"] and broker2 watches ["t3","t4"]. In this case, to get a complete view both need to be queried. Would this case work as well?

aho135:
> There is also the case where broker1 watches ["t1","t2"] and broker2 watches ["t3","t4"]. In this case, to get a complete view both need to be queried. Would this case work as well?

Thanks for the review @abhishek-chouhan! This case is not covered. This would be more invasive as it requires merging BrokerServerView from broker1 and broker2. In this scenario I think it would be simpler to spin up an additional broker (e.g. druid/broker-admin) which watches all tiers and configure consoleBrokerService to point there.

web-console/src/entry.tsx
```diff
@@ -70,6 +70,10 @@ interface ConsoleConfig {
 
   // Allow for namespacing the local storage in case multiple clusters share a URL due to proxying
   localStorageNamespace?: string;
+
+  // Broker service name to use for all console SQL queries (for tier isolation)
+  // This injects "brokerService" into the query context for routing via the manual strategy
+  consoleBrokerService?: string;
```
abhishekrb19: What do you think about exposing this config via `web-console-config.tsx` as well? It has this dialog on the console, which seems cool:

Image

I think `web-console-config.tsx` is session-based that stores config overrides in the browser, while the `ConsoleConfig` from `console-config.js`  is persistent via the build.  Not super familiar with the web-console code here, so correct me if I'm missing something here.

Can `console-config.js` be overridden per environment, given that this file is baked into the build as part of the source code?

web-console/src/utils/druid-query.ts
```diff
@@ -322,13 +322,35 @@ export class DruidError extends Error {
...
+// Broker service to use for all console queries (for tier isolation)
+let consoleBrokerService: string | undefined;
+
+export function setConsoleBrokerService(brokerService: string | undefined): void {
+  consoleBrokerService = brokerService;
+}
+
+export function getConsoleBrokerService(): string | undefined {
+  return consoleBrokerService;
+}
```
abhishekrb19: Is this used anywhere?

web-console/src/utils/druid-query.ts
```diff
@@ -341,7 +363,18 @@ export async function queryDruidSql<T = any>(
...
+      ? {
+          ...sqlQueryPayload,
+          context: {
+            ...sqlQueryPayload.context,
+            brokerService: consoleBrokerService,
+          },
+        }
+      : sqlQueryPayload;
+
+    sqlResultResp = await Api.instance.post('/druid/v2/sql', payload, { signal });
```
abhishekrb19, line 362: I see that https://github.com/apache/druid/blob/master/web-console/src/views/explore-view/explore-view.tsx#L64
 directly calls the `/druid/v2/sql` endpoint without invoking `queryDruidSql`. Should we update it to use this function instead, so the context is appended for the Explore view as well?

abhishek-chouhan:
> Thanks for the review @abhishek-chouhan! This case is not covered. This would be more invasive as it requires merging BrokerServerView from broker1 and broker2. In this scenario I think it would be simpler to spin up an additional broker (e.g. druid/broker-admin) which watches all tiers and configure consoleBrokerService to point there.

I do think we need the merging functionality to support the watchedTiers functionality in its entirety. While the workaround with a separate broker works, its too heavy of a workaround just to make the console work with watchedTiers. 
This change is still valuable, we can take up the merging brokerServerView change as a followup enhancement.

web-console/src/entry.tsx
abhishekrb19, line 64: Since `brokerService` is part of the query context, would setting `baseQueryContext` or `defaultQueryContext` in `console-config.js` work out of the box?

web-console/src/entry.tsx
aho135: Thanks for the review @abhishekrb19!

Yeah the `console-config.js` is persistent via the build, so overriding per environment is not possible. For our own use case this would still be fine, since it would fall back to the default druid/broker if the configured brokerService doesn't exist. Let me take a look at `web-console-config.tsx`

web-console/src/utils/druid-query.ts
aho135: Yeah this was added just for consistency, but I can remove it since it's unused

web-console/src/utils/druid-query.ts
aho135: Since the Explore tab is more for user-issued queries I thought to leave it out for now. I wanted to limit the scope of this to just Router console queries

web-console/src/entry.tsx
aho135: These settings only apply to queries run from the `Query` tab and not the console queries

web-console/src/entry.tsx
aho135: Image

I like this idea of adding it to the web-console-config. This uses the browser's local storage and persists through restarts

web-console/src/utils/druid-query.ts
abhishekrb19: Ah I see, thanks for clarifying. I got tripped up by the name `consoleBrokerService` - I assumed all queries originating from the console would use this, but it sounds like only the console-generated system queries would use it.

What do you think about renaming this to something less ambiguous, like `consoleSystemQueryBrokerService`? Alternatively, we could leave the name as-is if we expect the scope to expand in the future and instead clarify it in the help text and [documentation](https://druid.apache.org/docs/latest/configuration/#query-configuration) along with the limitation that @abhishek-chouhan mentioned with watch tiers (this can be done separately in a follow up PR)

web-console/src/utils/druid-query.ts
```diff
@@ -322,13 +322,31 @@ export class DruidError extends Error {
...
+      ? {
+          ...runeQuery,
+          context: {
+            ...runeQuery.context,
+            brokerService: consoleBrokerService,
+          },
+        }
+      : runeQuery;
+
+    runeResultResp = await Api.instance.post('/druid/v2', query, { signal });
```
abhishekrb19: This endpoint would be used by native queries/specs submitted by the sampler flow. Should we remove this?

web-console/src/utils/druid-query.ts
aho135: Thanks @abhishekrb19 this makes sense. Let me update the name to make clear that it is only for system queries

web-console/src/druid-models/web-console-config/web-console-config.tsx
```diff
@@ -34,4 +35,16 @@ export const WEB_CONSOLE_CONFIG_FIELDS: Field<WebConsoleConfig>[] = [
...
   },
+  {
+    name: 'consoleSystemQueryBrokerService',
+    type: 'string',
+    placeholder: 'druid/broker',
+    info: (
+      <>
+        Sets brokerService in the query context for console system queries
+        (e.g.,&quot;druid/broker-all-tiers&quot;). Useful when configuring
+        druid.broker.segment.watchedTiers. Does not affect user-initiated queries.
```
abhishekrb19:
```suggestion
        Sets brokerService in the query context for console system queries
        (e.g.,&quot;druid/broker-all-tiers&quot;). Useful when configuring
        druid.broker.segment.watchedTiers if the default Broker server doesn't have visibility into some Historical tiers. Does not affect user-initiated queries.
```

PR #19070, opened 2026-03-01
SegmentManager: Close references on failure of getSegmentsBundle.

gianm:
Similar to the approach in ServerManager#getOrLoadBundleSegments, we use a safetyNet to close references if bundle creation fails.


clintropolis (approved):
oops, yea i think this got lost in #18710, thanks for fixing

Issue #7384, opened 2019-03-29
Prohibit @author tags in Javadocs


leventov:
Once this is done, the corresponding note should be removed from CONTRIBUTING.md to reduce the amount of noise.

lml2468:
#7682

PR #19071, opened 2026-03-02
Upgrade Kubernetes clients and related dependencies in `kubernetes-overlord-extensions` and `druid-kubernetes-extensions`

capistrant:
## Summary

This PR bumps Kubernetes-related dependencies across both `druid-kubernetes-extensions` and `druid-kubernetes-overlord-extensions`, centralizes version management in the parent pom, and adds a new `WebClientOptions` pass-through feature for the vert.x HTTP client.

### Dependency version bumps

Managed in the parent `pom.xml`:

| Dependency | Old Version | New Version | Reason |
|---|---|---|---|
| `io.fabric8:kubernetes-client` | 7.5.2 | 7.6.0 | Exposes `additionalConfig(WebClientOptions)` hook ([#7384](https://github.com/fabric8io/kubernetes-client/pull/7384)) |
| `io.kubernetes:client-java` | 19.0.0 | 25.0.0-legacy | Major upgrade; `-legacy` variant avoids `validateJsonElement` strict validation regression in 20+ |
| `com.squareup.okhttp3:okhttp` | 4.12.0 | 5.3.2 | Aligned across both k8s extensions; version now centralized in parent pom |
| `org.jetbrains.kotlin:kotlin-stdlib` | 1.9.25 | 2.2.21 | Required by okhttp 5.x |
| `commons-codec:commons-codec` | 1.17.1 | 1.20.0 | Enforcer upper-bound from client-java 25.x transitives |
| `org.apache.commons:commons-lang3` | 3.19.0 | 3.20.0 | Enforcer upper-bound from client-java 25.x transitives |
| `com.google.code.gson:gson` | 2.12.0 | 2.13.2 | Enforcer upper-bound from client-java 25.x transitives |
| `com.amazonaws:aws-java-sdk` | 1.12.784 | 1.12.793 | Enforcer upper-bound from client-java 25.x transitives |

### `druid-kubernetes-extensions` changes

- **Removed local version management**: `kubernetes.client.version` property and `dependencyManagement` block removed; versions now flow from the parent pom
- **client-java API migration**: `DefaultK8sApiClient` updated for the 25.x-legacy positional-parameter API signatures. Pod patch operations now use `PatchUtils.patch()` with `V1Patch.PATCH_FORMAT_JSON_PATCH` to set the correct `Content-Type` header (fixes 415 Unsupported Media Type)
- **Dependency exclusions**:
  - `client-java-proto` + `protobuf-java`: proto brings in protobuf-java 4.x which Druid does not yet support project-wide; the extension uses JSON wire format only
  - `adal4j`: Azure AD auth library (and its transitive nimbus OIDC/JWT stack) not used by this extension
  - `jakarta.ws.rs-api:4.0`: conflicts with the `javax.ws.rs` 1.x used elsewhere in Druid
- **Added `okhttp` as explicit dependency due to being flagged by dep analyzer** (previously undeclared transitive)

### `druid-kubernetes-overlord-extensions` changes

- **`WebClientOptions` pass-through** (new feature): `DruidKubernetesVertxHttpClientFactory` now extends `VertxHttpClientFactory` and overrides `additionalConfig(WebClientOptions)` to apply a generic `Map<String, Object>` of operator-configured properties. This is powered by the new hook in fabric8 7.6.0.
- `DruidKubernetesVertxHttpClientConfig`: added `webClientOptions` map field
- `KubernetesOverlordModule`: injects `@Json ObjectMapper` for the `WebClientOptions` map-to-POJO deserialization
- Added `vertx-uri-template` dependency (required at runtime by fabric8 7.6.0)
- Removed hardcoded okhttp version (now managed by parent)
- New tests: `DruidKubernetesVertxHttpClientConfigTest`, `DruidKubernetesVertxHttpClientFactoryTest`

### Other

- `embedded-tests/pom.xml`: removed hardcoded `client-java-api` version (now managed by parent)
- `licenses.yaml`: updated versions for all changed dependencies, added new entries for `jakarta.annotation-api`, `swagger-annotations` (protobuf-extensions), `okhttp`/`okio` (overlord-extensions)
- Documentation added to `docs/development/extensions-core/k8s-jobs.md`

## Motivation

1. **Connection keep-alive control**: Druid clusters using the vert.x HTTP client for `kubernetes-overlord-extensions` can experience connection errors when an intermediate network component (AWS ALB, Kong, service mesh, etc.) closes idle connections before the client does. The `WebClientOptions` pass-through lets operators set client-side keep-alive timeouts lower than the server-side timeout.

2. **Dependency hygiene**: The `io.kubernetes:client-java` and `okhttp` versions were managed locally in the kubernetes-extensions module, diverging from the rest of the project. Centralizing in the parent pom and upgrading to current versions improves consistency and resolves transitive dependency conflicts.

## Usage of new vert.x http connection pool configuration overrides

Add properties to your Overlord runtime configuration using the prefix `druid.indexer.runner.k8sAndWorker.http.vertx.webClientOptions.<property>`:

```properties
druid.indexer.runner.k8sAndWorker.http.vertx.webClientOptions.keepAliveTimeout=30
druid.indexer.runner.k8sAndWorker.http.vertx.webClientOptions.idleTimeout=30
```

Any property with a public setter on [`WebClientOptions`](https://vertx.io/docs/apidocs/io/vertx/ext/web/client/WebClientOptions.html) or its parent classes is supported. Invalid property names will cause the Overlord to fail on startup with a clear error message.

#### Release note

Bumped `fabric8/kubernetes-client` from 7.5.2 to 7.6.0 and `io.kubernetes/client-java` from 19.0.0 to 25.0.0-legacy across both Kubernetes extensions. Upgraded `okhttp` from 4.x to 5.3.2 (centralized in parent pom). Several transitive dependency versions bumped to satisfy enforcer upper bounds: `commons-codec` 1.20.0, `commons-lang3` 3.20.0, `gson` 2.13.2, `kotlin-stdlib` 2.2.21, `aws-java-sdk` 1.12.793.

Added a new `WebClientOptions` pass-through for the vert.x HTTP client in `kubernetes-overlord-extensions`. Operators can now configure any property on the underlying Vert.x `WebClientOptions` object (connection pool size, keep-alive timeouts, idle timeouts, etc.) via Druid runtime properties. This is particularly useful for environments with intermediate load balancers that close idle connections. Most Druid deployments will not need this configuration.

<hr>

 * `DefaultK8sApiClient` — migrated to client-java 25.x-legacy API + `PatchUtils`
 * `DruidKubernetesVertxHttpClientFactory` — now extends `VertxHttpClientFactory`, added `WebClientOptions` pass-through
 * `DruidKubernetesVertxHttpClientConfig` — added `webClientOptions` map
 * `KubernetesOverlordModule` — `@Json ObjectMapper` injection for vert.x factory

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/kubernetes-overlord-extensions/pom.xml
+++ b/extensions-core/kubernetes-overlord-extensions/pom.xml
@@ -150,11 +150,6 @@
     </dependency>
     <dependency>
       <groupId>io.vertx</groupId>
-      <artifactId>vertx-uri-template</artifactId>
-      <version>${vertx.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>io.vertx</groupId>
       <artifactId>vertx-web-client</artifactId>
       <version>${vertx.version}</version>
     </dependency>


extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/httpclient/vertx/DruidKubernetesVertxHttpClientFactory.java
```diff
@@ -19,31 +19,51 @@
...
  */
-public class DruidKubernetesVertxHttpClientFactory implements DruidKubernetesHttpClientFactory
+public class DruidKubernetesVertxHttpClientFactory extends VertxHttpClientFactory implements DruidKubernetesHttpClientFactory
 {
   public static final String TYPE_NAME = "vertx";
-  private final Vertx vertx;
 
-  public DruidKubernetesVertxHttpClientFactory(final DruidKubernetesVertxHttpClientConfig httpClientConfig)
+  private static final Logger LOG = new Logger(DruidKubernetesVertxHttpClientFactory.class);
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
gianm: I suppose it's fine not to use the server's `ObjectMapper` here, but it would still be cleaner to, I think. It avoids having this static lying around when it isn't going to be used after server startup.

extensions-core/kubernetes-overlord-extensions/pom.xml
```diff
@@ -139,12 +142,17 @@
     <dependency>
       <groupId>com.squareup.okhttp3</groupId>
       <artifactId>okhttp</artifactId>
-      <version>4.12.0</version>
+      <version>5.3.2</version>
```
gianm: Anything to worry about with this major upgrade?

The same `okhttp-4.12.0` is also used in `druid-kubernetes-extensions`. Possible to keep these consistent and update that one too?

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/common/httpclient/vertx/DruidKubernetesVertxHttpClientFactory.java
capistrant: thx for flagging

extensions-core/kubernetes-overlord-extensions/pom.xml
capistrant: for this extension in particular, no. It is a minimum req for the the upgraded fabric8 library. We just pull the dependency explicitly to configure some dispatcher stuff under the hood.

re multiple okhttp versions, I think I'm going to pivot this PR to be a bit wider to keep this in sync. in the druid-kubernetes-extensions, we have fallen way behind on the native k8s client version. If we update that we can support okhttp 5 across both. the upgrade seems straightforward and testing locally has been positive

capistrant: eh maybe I spoke too soon on that io.kubernetes update being simple... that might force a bunch of other major dep updates as part of the dependency enforcer. I will maybe go back to the drawing board on this

capistrant: ok, after a song and dance it seems like getting aligned on okhttp 5.3.2 is doable by bumping the kubernetes client for both k8s extensions. It sure would be nice if we had a single client that we used for k8s instead of both io.kubernetes + fabric8. Maybe if all goes well with this fabric8 bump we can shed the okhttp and native jdk http clients plus use fabric8 in both k8s related extensions. Would probably save a lot of annoyances in the future.

gianm: It's possibly an artifact of the two different extensions being developed separately. Agree it would be good to standardize on a single k8s client.

PR #19072, opened 2026-03-03
fix issue with nested column processing/storage of empty fields

clintropolis:
### Description

changes:
* fix bug in `NestedPathFinder.toNormalizedJsonPath` creating incorrect paths when given an empty field name
* fix `NestedPathFinder.parseJsonPath` to correctly detect illegal empty paths consisting of consecutive . characters
* added `NestedPathFinder.parseBadJsonPath` to read nested column fields dictionaries and detect and fixup illegal path expressions, using a newly added `FieldsFixupIndexed` to swap the bad values with good values
* `NestedDataColumnSupplier` on column read attempts to detect bad paths written by the bugged version of `NestedPathFinder.toNormalizedJsonPath`
* added 'pathParserVersion' field to nested column part serde so that newly written nested columns after the bug fix can skip checking for the bug


processing/src/main/java/org/apache/druid/segment/serde/NestedCommonFormatColumnPartSerde.java
```diff
@@ -87,7 +88,8 @@ public static NestedCommonFormatColumnPartSerde createDeserializer(
       @JsonProperty("enforceLogicalType") boolean enforceLogicalType,
       @JsonProperty("byteOrder") ByteOrder byteOrder,
       @JsonProperty("bitmapSerdeFactory") BitmapSerdeFactory bitmapSerdeFactory,
-      @JsonProperty("columnFormatSpec") @Nullable FormatSpec columnFormatSpec
+      @JsonProperty("columnFormatSpec") @Nullable FormatSpec columnFormatSpec,
+      @JsonProperty("pathParserVersion") @Nullable Byte pathParserVersion
```
gianm, line 93: I don't see a `@JsonProperty` getter for this. Is it going to make it into the serialized form?

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSupplier.java
```diff
@@ -182,6 +194,38 @@ public static NestedDataColumnSupplier read(
     }
   }
 
+
+  @VisibleForTesting
+  static Supplier<? extends Indexed<ByteBuffer>> getAndFixFieldsSupplier(
```
gianm, line 214: Should have some javadoc explaining what "fix" means. Having a short description and then linking to a GitHub issue or PR for more details is a useful way to do it.

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSupplier.java
```diff
@@ -268,4 +312,87 @@ public <T> T as(Class<T> clazz)
     }
     return null;
   }
+
+  @VisibleForTesting
+  public static class FieldsFixupIndexed implements Indexed<ByteBuffer>
```
gianm, line 345: Similar to the static creator method, this should have some javadoc explaining what "fix" means. Having a short description and then linking to a GitHub issue or PR for more details is a useful way to do it.

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSupplier.java
```diff
@@ -268,4 +312,87 @@ public <T> T as(Class<T> clazz)
...
+          pos++;
+          return delegateIterator.next();
+        }
+      };
+    }
+
+    @Override
+    public boolean isSorted()
+    {
+      return delegate.isSorted();
```
gianm: Is this always going to be true (will the fixed-up keys sort the same way as the bad keys)? Also, does it matter (does anything care if this indexed is sorted)?

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSupplier.java
```diff
@@ -268,4 +312,87 @@ public <T> T as(Class<T> clazz)
...
+
+    @Override
+    public int indexOf(@Nullable ByteBuffer value)
+    {
+      for (Int2ObjectMap.Entry<ByteBuffer> entry : fixup.int2ObjectEntrySet()) {
+        if (entry.getValue().equals(value)) {
+          return entry.getIntKey();
+        }
+      }
+      return delegate.indexOf(value);
```
gianm, line 380: Should return this only if the key is not present in `fixup`. Otherwise it will return a nonnegative index for the bad keys.

processing/src/main/java/org/apache/druid/segment/nested/NestedPathFinder.java
```diff
@@ -173,7 +177,105 @@ public static List<NestedPathPart> parseJsonPath(@Nullable String path)
...
+        badFormatJsonPath(path, "unterminated '['");
+      }
+      parts.add(new NestedPathField(path.substring(partMark)));
+    }
+
+    return parts;
+  }
+
+  /**
+   * split a JSONPath path into a series of extractors to find things in stuff. This method is mostly a duplicate of
```
gianm: Surely something can make more sense then "find things in stuff".

processing/src/main/java/org/apache/druid/segment/nested/NestedPathFinder.java
```diff
@@ -173,7 +177,105 @@ public static List<NestedPathPart> parseJsonPath(@Nullable String path)
...
+   * {@link #parseJsonPath(String)} fixing up any bugs encountered from bad paths like '$..a' or '$.[0].a' from
+   * previously bugged versions of {@link #toNormalizedJsonPath(List)} and should be kept in sync.
```
gianm: Can these be made to share a common parse function that is parameterized with some boolean like `allowBadPaths`? If not, at least add a comment in `parseJsonPath` too, so someone modifying that can remember to update this one too.

clintropolis: done, called parameter `allowFixEmptyFieldsBadPaths` since it is pretty specific on the bad paths it allows

processing/src/main/java/org/apache/druid/segment/nested/NestedPathFinder.java
clintropolis: updated javadocs to be a bit clearer

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSupplier.java
clintropolis: Do you mean like if a caller is searching for the bad keys? I don't think that should be possible for most callers since when calling this method we generally run it through `NestedPathFinder.toNormalizedJsonPath` before searching because usually at the point we call this the path expressions have been converted to `List<NestedPathPart>` during SQL planning.

Also, i believe it should be harmless for `indexOf` to report the positions of the bad paths if using the bad path expressions.

Added explanation to javadocs

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSupplier.java
clintropolis: nope, good catch, removed this since we can't guarantee it and added note in javadocs about it not mattering

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSupplier.java
clintropolis: added javadocs

processing/src/main/java/org/apache/druid/segment/serde/NestedCommonFormatColumnPartSerde.java
clintropolis: good catch, fixed

clintropolis: added a serde test too

processing/src/main/java/org/apache/druid/segment/nested/NestedDataColumnSupplier.java
gianm: Yes, I meant if someone is searching for the bad keys. In that case the behavior here doesn't adhere to the Indexed interface. If we're going to deviate from it, that can be fine given it's only used in this specific place, but any intentional deviation should be javadoc'd.

PR #19068, opened 2026-02-28
Fix flaky QueryVirtualStorageTest

maytasm:
Fix flaky QueryVirtualStorageTest

### Description

- QueryVirtualStorageTest: Still seeing timeout with 20seconds, bumping up to 30 seconds 
```
Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 34.09 s <<< FAILURE! -- in org.apache.druid.testing.embedded.query.QueryVirtualStorageTest
Error:  org.apache.druid.testing.embedded.query.QueryVirtualStorageTest -- Time elapsed: 34.09 s <<< ERROR!
org.apache.druid.java.util.common.ISE: Timed out waiting for event after [20,000]ms
	at org.apache.druid.server.metrics.LatchableEmitter.waitForEvent(LatchableEmitter.java:130)
	at org.apache.druid.server.metrics.LatchableEmitter.waitForEvent(LatchableEmitter.java:175)
	at org.apache.druid.testing.embedded.EmbeddedClusterApis.waitForTaskToFinish(EmbeddedClusterApis.java:242)
	at org.apache.druid.testing.embedded.EmbeddedClusterApis.waitForTaskToSucceed(EmbeddedClusterApis.java:214)
	at org.apache.druid.testing.embedded.msq.EmbeddedMSQApis.runTaskSqlAndGetReport(EmbeddedMSQApis.java:143)
	at org.apache.druid.testing.embedded.query.QueryVirtualStorageTest.loadWikiData(QueryVirtualStorageTest.java:423)
	at org.apache.druid.testing.embedded.query.QueryVirtualStorageTest.loadData(QueryVirtualStorageTest.java:143)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

- QueryVirtualStorageTest: handle null loadWait in DART report assertion

ChannelCounters.Snapshot.getLoadWait() is Nullable — it returns null when all values are zero, which can happens when
segment load wait time is sub-millisecond (the internal nanosecond value is converted to milliseconds via
TimeUnit.NANOSECONDS.toMillis, truncating tiny waits to zero). On fast machines (i.e. with a lot of cores) or when the executor has free threads, the wait is effectively zero, causing a NullPointerException when the test accessed getLoadWait()[0] unconditionally.

Updated the assertion to tolerate the null case, since zero wait time is a legitimate outcome:
```
Assertions.assertTrue(
[...]

PR #19073, opened 2026-03-03
ChannelCounters: Store segment load/wait time in nanos.

gianm:
This patch changes the loadTime and loadWait such that they are accumulated in nanoseconds and snapshotted in milliseconds. Snapshots are in the same units they were before (milliseconds), but by using nanoseconds for accumulation, we avoid accumulating rounding errors.

This fixes some flakiness in QueryVirtualStorageTest, which asserts that the loadWait is > 0. If the test runs quickly enough that the individual waits are always sub-ms, this would end up zero due to rounding errors.


jtuglu1:
Potentially fixes https://github.com/apache/druid/pull/19068/changes

gianm:
> Potentially fixes https://github.com/apache/druid/pull/19068/changes

Ah, oops, I didn't notice that PR. Yes, this PR is meaning to fix the same thing.

PR #19074, opened 2026-03-03
Use time-based retention for MSQ query result file cleanup

cecemei:
### Description
Adds a configurable retention duration (default: 6 hours) to `DurableStorageCleanerConfig` and updates the cleaner to retain query result files based on task creation time rather than checking known task IDs.

### Release Note
  The durable storage cleaner now supports configurable time-based retention for MSQ query results. Previously, query results were retained for all known tasks list, which was unreliable for completed tasks. With this change, query results are retained for a configurable time period based on the task creation time.

  The new configuration property `druid.msq.intermediate.storage.cleaner.durationToRetain` controls the retention period for query results. The default retention period is 6 hours.

<hr>

This PR has:

- [x] been self-reviewed.


multi-stage-query/src/main/java/org/apache/druid/msq/indexing/cleaner/DurableStorageCleaner.java
```diff
@@ -102,9 +102,14 @@ public void run() throws Exception
...
         if (runningTaskIds.contains(nextDirName)) {
           // do nothing
-        } else if (DurableStorageUtils.QUERY_RESULTS_DIR.equals(nextDirName)
-                   && DurableStorageUtils.isQueryResultFileActive(currentFile, knownTaskIds)) {
-          // query results should not be cleaned even if the task has finished running
+        } else if (
+            DurableStorageUtils.QUERY_RESULTS_DIR.equals(nextDirName)
+            && DurableStorageUtils.isQueryResultFileActive(
+                currentFile,
+                taskId -> Optional.fromNullable(taskStorage.getTaskInfo(taskId)).transform(TaskInfo::getCreatedTime),
```
gianm: This is going to be expensive; each usage of `taskStorage` is a call to the metadata store. It's better to batch these, such as by using one call to get all of the task IDs that have completed within the retention period. Then you can delete anything that doesn't appear in that list.

processing/src/main/java/org/apache/druid/frame/util/DurableStorageUtils.java
```diff
@@ -255,6 +257,8 @@ public static boolean isQueryResultFileActive(String path, Set<String> knownTask
     if (!DurableStorageUtils.QUERY_RESULTS_DIR.equals(elements.get(0))) {
       return false;
     }
-    return knownTasks.contains(elements.get(1));
+    Optional<DateTime> taskCreatedAt = taskCreationTimeFn.apply(elements.get(1));
```
gianm: I think this isn't going to work, because the directory is computed using `DurableStorageUtils.getControllerDirectory`, and isn't exactly the same as the controller task ID. Please double check this and update the tests so they confirm this part works.

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/cleaner/DurableStorageCleanerConfig.java
```diff
@@ -39,6 +40,9 @@ public class DurableStorageCleanerConfig
   @Min(1)
   public long delaySeconds = 86400L;
 
+  @JsonProperty
+  public long durationToRetain = TimeUnit.HOURS.toMillis(6);
```
gianm: Please add a config for this property and change its type to `Duration` so it can be specified using ISO8601 duration strings, to make it consistent with other duration parameters. Please also document it.

multi-stage-query/src/test/java/org/apache/druid/msq/indexing/DurableStorageCleanerTest.java
```diff
@@ -78,24 +88,52 @@ public void testRun() throws Exception
...
-    EasyMock.replay(TASK_MASTER, TASK_RUNNER, TASK_RUNNER_WORK_ITEM, STORAGE_CONNECTOR);
+    EasyMock.replay(TASK_STORAGE, TASK_MASTER, TASK_RUNNER, TASK_RUNNER_WORK_ITEM, STORAGE_CONNECTOR);
 
     durableStorageCleaner.run();
 
     Assert.assertEquals(Sets.newHashSet(STRAY_DIR), capturedArguments.getValue());
   }
 
+  @Test
+  public void testRunClearsStaleTask() throws Exception
```
gianm: Please also include a test for a task not found

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/cleaner/DurableStorageCleaner.java
gianm: Btw, ideally that call to get all those task IDs should only look at type `query_controller`. It will generally be a much smaller and more manageable list than "all tasks".

processing/src/main/java/org/apache/druid/frame/util/DurableStorageUtils.java
cecemei: true, updated this

multi-stage-query/src/test/java/org/apache/druid/msq/indexing/DurableStorageCleanerTest.java
cecemei: with the new change, stale is the as as not found, updated test name

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/cleaner/DurableStorageCleaner.java
cecemei: updated to only fetch completed task within retention period, we don't have filter by type in `CompleteTaskLookup`, it wont be too much work to add such filter. i just assume that even without this filter we wont have too much tasks finished within the last 6 hours.

processing/src/main/java/org/apache/druid/frame/util/DurableStorageUtils.java
```diff
@@ -52,6 +52,23 @@ private static String getQueryResultsControllerDirectory(final String controller
...
+    if (elements.size() < 2) {
+      return null;
+    }
+    if (!DurableStorageUtils.QUERY_RESULTS_DIR.equals(elements.get(0))) {
+      return null;
+    }
+    if (!elements.get(1).startsWith("controller_")) {
+      return null;
+    }
+    return elements.get(1).substring(11);
```
gianm: Better to clarify what `11` means, such as by using a constant like `private static final int CONTROLLER_PREFIX_LENGTH = "controller_".length()`

multi-stage-query/src/test/java/org/apache/druid/msq/indexing/DurableStorageCleanerTest.java
```diff
@@ -127,10 +172,109 @@ public void testGetSchedule()
...
+        null,
+        null
+    );
 
     DutySchedule schedule = durableStorageCleaner.getSchedule();
     Assert.assertEquals(cleanerConfig.delaySeconds * 1000, schedule.getPeriodMillis());
     Assert.assertEquals(cleanerConfig.delaySeconds * 1000, schedule.getInitialDelayMillis());
   }
+
+  public static class TestTask implements Task
```
gianm: Could this be a NoopTask?

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/cleaner/DurableStorageCleaner.java
```diff
@@ -52,17 +55,20 @@ public class DurableStorageCleaner implements OverlordDuty
...
   private final StorageConnector storageConnector;
   private final Provider<TaskMaster> taskMasterProvider;
+  private final TaskStorage taskStorage;
 
   @Inject
   public DurableStorageCleaner(
       final DurableStorageCleanerConfig config,
       final @MultiStageQuery StorageConnectorProvider storageConnectorProvider,
-      @JacksonInject final Provider<TaskMaster> taskMasterProvider
+      @JacksonInject final Provider<TaskMaster> taskMasterProvider,
```
gianm: I don't think the `@JacksonInject` is doing anything here, since this is not read using Jackson. I see it was already there, but since you're editing the constructor anyway, see if things still work if you remove it.

PR #18935, opened 2026-01-21
Update milestone on PR close and ensure they are visible for the originally desired milestone

kgyrtkirk:
this PR enables to maintain the milestones - if a PR is closed on master:
* it assign the PR to the next release milestone based on the pom.xml
* if there was a different milestone on the PR create an issue about backporting that PR for that is created

also adds logic to:
* if the backport related issue is labeled with `approved` a backport pr is created
   * the workflow which creates the issue could not trigger other workflows; also someone might want to choose a different way

written by claude with a decent amount of convincing how not to do it...

PR #19081, opened 2026-03-04
Fix add-milestone GHA job

abhishekrb19:
A GitHub action to auto-add milestones were added in https://github.com/apache/druid/pull/18935

As @kgyrtkirk noted in this [comment](https://github.com/apache/druid/pull/18935#issuecomment-3780550138), this check is failing with a 403 error for [all](https://github.com/apache/druid/pull/19052) PRs coming from forks once they are merged into master:
```json
{"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/issues/issues#update-an-issue","status":"403"}
``` 

Given the 403, it does indicate a permissions issue likely with the GitHub token used for this. Trying `pull_request_target` instead of `pull_request` similar to the existing [labeler](https://github.com/apache/druid/blob/master/.github/workflows/labeler.yml#L26) action which uses the same. Since this specific add-milestone check doesn't export any credentials, I believe using `pull_request_target` should be okay. Hopefully, with the elevated base token, this should work. 🤞

### DIFF
--- a/.github/workflows/pr-merged.yml
+++ b/.github/workflows/pr-merged.yml
@@ -20,7 +20,7 @@
 name: "PR Merged"
 
 on:
-  pull_request:
+  pull_request_target:
     types: [closed]
 
 jobs:


abhishekrb19:
Thanks for the review, @capistrant! 

The milestone has been successfully added now. We'll still need to manually add the 37.0.0 milestone for existing PRs where it wasn’t added bc of the 403, but this should work going forward for new PRs.

PR #19084, opened 2026-03-04
Add AGENTS.md

abhishekrb19:
Add `AGENTS.md` that AI tools like Claude and Cursor can leverage. The general recommendation from these tools is to keep the  markdown files lightweight but collaborative, so most folks using these tools in the repo for development or otherwise can use it right out of the box.

Other Apache repos like [Spark](https://github.com/apache/spark/blob/master/AGENTS.md)  and [Superset](https://github.com/apache/superset/blob/master/AGENTS.md) follow a similar structure as well.

Use https://gist.github.com/gianm/031a578a75b4dc44f3e45661d86bcca9 as a baseline with some minor tweaks.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,4 +18,68 @@
   -->
 
 
-This file provides context and guidelines for AI coding assistants working with the Apache Druid codebase.
+# Apache Druid
+
+Real-time analytics database. Java, Maven, multi-module project.
+
+## Key Modules
+
+- `processing/`: Core query processing, aggregation, data structures
+- `server/`: Common server functionality, HTTP services
+- `sql/`: SQL planning (Calcite), SQL-to-native translation
+- `indexing-service/`: Task framework, ingestion coordination
+- `multi-stage-query/`: MSQ engine
+- `extensions-core/`: Core extensions (S3, Kafka, Kinesis, Parquet, DataSketches, etc.)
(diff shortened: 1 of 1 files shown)


gianm:
This is the one I use: https://gist.github.com/gianm/031a578a75b4dc44f3e45661d86bcca9

It focuses on the module structure, code style, & Maven commands for running tests, because those were the things I had seen agents having the most trouble with.

capistrant:
I like Gian's as a reference place to start! In addition to that, I think some thing brief about how to build and test the web-console + generate snapshots would also be useful as that is a bit of a pain point for folks who may once a year have to make a UI change and would be nice if an agent could easily handle.

abhishekrb19:
> This is the one I use: https://gist.github.com/gianm/031a578a75b4dc44f3e45661d86bcca9
> 
> It focuses on the module structure, code style, & Maven commands for running tests, because those were the things I had seen agents having the most trouble with.

Thanks, this seems useful. I've updated it to use this as the base template with a few tweaks:
- Based on @capistrant's suggestion, I added a brief section for web-console development referring to `web-console/README.md`. If there are additional specific details, I think it'd be better to include them in the web-console's README.
- Included a small Documentation section after noticing agents updating existing table formatting, which caused unnecessary diffs / spelling errors with not backquoting properties.

PR #19089, opened 2026-03-05
dev: create convention commit syntax

jtuglu1:
### Description

Something I've seen many large open source projects adopt:

1. [Conventional commit syntax](https://www.conventionalcommits.org/en/v1.0.0/) (easier to auto-generate release notes, bisect for issues, etc. as it requires more context to be stored in the commits themselves, and not the PR/issue).

#### Release note

Enforce conventional commit format. This will also help making debugging issues, etc. as in addition to GH tags (something not persisted in the git lineage) breaking changes/bug fixes/feature information is recorded in the git tree.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -13,12 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: PR Checks
+name: "PR Commit Checks"
 
 on:
-  # pull_request_target is required so that the workflow can post comments on
-  # PRs from forks (which don't have write permissions otherwise).
-  pull_request_target:
+  pull_request:
     types: [opened, edited, synchronize, reopened]
 
 concurrency:
@@ -27,8 +25,7 @@ concurrency:
 
 jobs:
   conventional-commits:
-    permissions:
-      pull-requests: write
+    if: github.repository == 'apache/druid' && github.event.pull_request.base.ref == 'master'
     name: Verify PR title conforms to Conventional Commits
     runs-on: ubuntu-latest
     steps:


.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,101 @@
...
+  # modify labels on the PR.
+  pull_request_target:
+    types: [opened, edited, synchronize, reopened]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  labeler:
```
gianm: What's the `labeler` job for? We have a labeler job already in `.github/workflows/labeler.yml`, it seems redundant to have two.

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,101 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          fail_on_error: true
+  commitlint:
+    permissions:
+      pull-requests: write
+    name: Verify PR title / description conforms to semantic-release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v4
```
gianm: I've also seen https://github.com/amannn/action-semantic-pull-request (`uses: amannn/action-semantic-pull-request@v5`) used to do this. Being a dedicated github action, it seems preferable since usage is simpler.

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,101 @@
...
+      - run: npm install @commitlint/config-conventional
+      - run: >
+          echo 'module.exports = {
+            "rules": {
+              "body-max-line-length": [0, "always", Infinity],
+              "footer-max-line-length": [0, "always", Infinity],
+              "body-leading-blank": [0, "always"]
+            }
+          }' > .commitlintrc.js
+      - run: npx commitlint --extends @commitlint/config-conventional --verbose <<< $COMMIT_MSG
```
gianm: Docs (https://github.com/srvaroa/labeler) suggest the builtin list is the following. I think we should customize the list. In addition to these, it's also useful to have `backport`, `release` (release-related changes like version bumps), `revert`, and `dev` (changes geared towards developers; this PR would be categorized as `dev` if we adopted it).

If we're allowing bike shedding, I prefer `minor` to `chore` for small non-user-visible changes like correcting comments. `chore` has some negative sound to it, whereas `minor` to me sounds more neutral.

```
[
  'build',
  'chore',
  'ci',
  'docs',
  'feat',
  'fix',
  'perf',
  'refactor',
  'revert',
  'style',
  'test'
]
```

benchmarks/pom.xml
```diff
@@ -17,88 +17,94 @@
...
+
+  <properties>
+    <jmh.version>1.21</jmh.version>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <uberjar.name>benchmarks</uberjar.name>
+  </properties>
+
   <dependencies>
     <dependency>
-      <groupId>org.jdbi</groupId>
```
gianm, line 35: I don't think we need to enforce that the dependencies are sorted. I don't believe it adds much, and the diff is gigantic, which always makes me think twice about something.

Enforced ordering of the top-level pom sections (parent, artifactId, packaging, name, properties, dependencies, etc) does seem more useful and less likely to generate a gigantic diff, if you'd want to explore that.

jtuglu1:
@gianm updated. I think we should also make the default repo settings on PRs to do a merge and take the PR description (e.g. not just force a *-list of commit messages (which often aren't descriptive at all).

benchmarks/pom.xml
jtuglu1: removing that change from this PR.

jtuglu1:
@gianm thoughts here?

jtuglu1:
@gianm any thoughts here?

gianm:
> @gianm updated. I think we should also make the default repo settings on PRs to do a merge and take the PR description (e.g. not just force a *-list of commit messages (which often aren't descriptive at all).

By "do a merge" you mean "squash and merge"?  I think that's already the default and IMO is best.

I'm not sure how I feel about the default description. The problem with concatenating the commits is it can include a bunch of junk. The problem with using the PR description is they tend to be long, IMO too long to be good commit messages. I generally try to edit the commit message when I'm merging a PR. I think there's no way around manual edits being a good idea. Anyway, I'm sort of ambivalent about the default messages, as a result.

jtuglu1:
> > @gianm updated. I think we should also make the default repo settings on PRs to do a merge and take the PR description (e.g. not just force a *-list of commit messages (which often aren't descriptive at all).
> 
> By "do a merge" you mean "squash and merge"? I think that's already the default and IMO is best.
> 
> I'm not sure how I feel about the default description. The problem with concatenating the commits is it can include a bunch of junk. The problem with using the PR description is they tend to be long, IMO too long to be good commit messages. I generally try to edit the commit message when I'm merging a PR. I think there's no way around manual edits being a good idea. Anyway, I'm sort of ambivalent about the default messages, as a result.

Ah – I wasn't referring to the merge method itself, but rather the default commit message provided by GH when merging (I notice it's a *-bulleted list of commit messages, which aren't that descriptive). I was thinking the commit message could be by default the PR description (and people can trim as needed). Just a minor thing, not too important.

gianm:
Btw, this PR should probably be `dev:`? It doesn't seem like a minor thing, as it will affect the dev workflow for everyone. To me `minor:` is for stuff like correcting spelling errors. And here we go with second-guessing the tags before they're even added 😄

jtuglu1:
> Btw, this PR should probably be `dev:`? It doesn't seem like a minor thing, as it will affect the dev workflow for everyone. To me `minor:` is for stuff like correcting spelling errors. And here we go with second-guessing the tags before they're even added 😄

Done. I think that difference is largely immaterial; what **does** matter is that we are now recording context inside the git tree, which makes it WAY easier to bisect for issues without digging through PR links. It also helps that GH release notes structure can be auto-generated from the conventional commit syntax (meaning you don't need to rely on users specifying Release Notes tag and instead do something like (fix!: blah) to know that it's a breaking change).

I could add a general guide on how/what to label things in the workflow – that might be helpful.

PR #19091, opened 2026-03-05
Do not kill a task if offsets are inconsistent but publish from another group is pending

kfaraz:
### Description

Follow up to #19034 , which addressed the issue of using the correct starting sequences after a scaling event occurs.

Discovered another race condition which causes task failures in pairs.

### Change summary

- Add method `SeekableStreamSupervisor.isAnotherTaskGroupPublishingToPartitions()`
- Use this method to check if a task needs to wait before publishing its own offests
- Update `CostBasedAutoScalerIntegrationTest`
- Update `SegmentTransactionalAppendAction` and `SegmentTransactionalInsertAction` to return a retryable error response only if there is a pending publish that conflicts with the current action
- Fix bug in `SeekableStreamSupervisorIOConfig`
- Fix bug in `CostBasedAutoScaler` to avoid spurious scale downs
- Validate metrics in `CostBasedAutoScaler` before proceeding with scaling action

### Race condition

<details>
<summary>Example error</summary>

```
First failure:
Killing task[A] as its checkpoints[...] are not consistent with group checkpoints[...] or latest offsets in DB[...].

Second failure:
Stored metadata state[...] has already been updated by other tasks and has diverged from the start metadata state[...].
```

</details>

<details>
<summary>Typical scenario</summary>

This typically plays out as follows:
- Offsets in DB are currently at E0.
- Scaling event occurs
- New task group B is created and is assigned a partition P1 which an old task group A (still pending completion) was also reading from.
- Task group A is requested to checkpoint and start publishing.
- Task group A returns the final end offset as E1.
- Task group B is assigned starting offset as E1 (fixed in #19034 and now works as expected)
- Next invocation of `runInternal()` calls `verifyAndMergeCheckpoints()`
- Task group B is found to have offsets E1 which are inconsistent with DB (since A is yet to publish E1 and DB is still at E0).
- Task group B is killed
- New task group C is launched with starting offsets E0
- Task group A finishes publishing and updates offsets in metadata store from E0 to E1
- Task group C also fails while publishing since offsets have already advanced from what C knew (i.e. E0)

There is already a check to not kill task group B if there are some `pendingCompletionTaskGroups` but only for the same group B. When scaling happens, partitions are reassigned and tasks from a different group A may also end up updating the offsets that B is reading from.

</details>

### Fix

While performing `verifyAndMergeCheckpoints`, consider a task eligible for checkpoint verification only if no other task group (irrespective of groupId) is waiting to publish to any of the partitions from which a task is reading.

This still feels like a temporary fix.

A cleaner, more long-term fix would be to rethink the way `TaskGroup` is handled inside `SeekableStreamSupervisor`, so that it lends itself to scaling and partition reassignments better.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalAppendAction.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalAppendAction.java
@@ -220,29 +220,11 @@ public class SegmentTransactionalAppendAction implements TaskAction<SegmentPubli
 
     IndexTaskUtils.emitSegmentPublishMetrics(retVal, task, toolbox);
 
-    if (shouldFailImmediately(retVal, task, toolbox)) {
+    if (toolbox.shouldFailSegmentPublishImmediately(retVal, task, supervisorId, startMetadata)) {
       return SegmentPublishResult.fail(retVal.getErrorMsg());
     } else {
       return retVal;
     }
-  }
-
-  /**
-   * Checks if the current publish action should be failed without allowing any
-   * more retries. A failed publish action should be retried only if there is
-   * another task waiting to publish offsets for an overlapping set of partitions.
-   */
-  private boolean shouldFailImmediately(SegmentPublishResult result, Task task, TaskActionToolbox toolbox)
-  {
-    if (result.isSuccess() || !result.isRetryable() || startMetadata == null) {
-      return false;
-    }
-
-    return !toolbox.getSupervisorManager().isAnotherTaskGroupPublishingToPartitions(
-        supervisorId,
-        task.getId(),
-        startMetadata
-    );
   }
 
   @Override
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertAction.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertAction.java
@@ -260,7 +260,7 @@ public class SegmentTransactionalInsertAction implements TaskAction<SegmentPubli
 
     IndexTaskUtils.emitSegmentPublishMetrics(retVal, task, toolbox);
 
-    if (shouldFailImmediately(retVal, task, toolbox)) {
+    if (toolbox.shouldFailSegmentPublishImmediately(retVal, task, supervisorId, startMetadata)) {
       return SegmentPublishResult.fail(retVal.getErrorMsg());
     } else {
       return retVal;
@@ -326,24 +326,6 @@ public class SegmentTransactionalInsertAction implements TaskAction<SegmentPubli
       }
     }));
     return segmentsMap;
-  }
-
-  /**
-   * Checks if the current publish action should be failed without allowing any
-   * more retries. A failed publish action should be retried only if there is
-   * another task waiting to publish offsets for an overlapping set of partitions.
-   */
-  private boolean shouldFailImmediately(SegmentPublishResult result, Task task, TaskActionToolbox toolbox)
-  {
-    if (result.isSuccess() || !result.isRetryable() || startMetadata == null) {
-      return false;
-    }
-
-    return !toolbox.getSupervisorManager().isAnotherTaskGroupPublishingToPartitions(
-        supervisorId,
-        task.getId(),
-        startMetadata
-    );
   }
 
   @Override
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
@@ -89,7 +89,6 @@ public abstract class SeekableStreamSupervisorIOConfig
     // Could be null
     this.autoScalerConfig = autoScalerConfig;
     this.autoScalerEnabled = autoScalerConfig != null && autoScalerConfig.getEnableTaskAutoScaler();
-    // if autoscaler is enabled, then taskCount will be ignored here and initial taskCount will equal to taskCountStart/taskCountMin
     if (autoScalerEnabled) {
       // Priority: taskCountStart > taskCount > taskCountMin
       this.taskCount = Configs.valueOrDefault(
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
@@ -553,7 +553,7 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
       log.makeAlert(
           "Supervisor[%s] has high partition lag[%.0f] but processing rate is zero. Check if the tasks are stuck.",
           supervisorId, metrics.getAvgPartitionLag()
-      );
+      ).emit();
       return Either.error("Lag is high but processing rate is zero");
     } else {
       return Either.value(true);
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/TaskActionToolbox.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/actions/TaskActionToolbox.java
@@ -23,8 +23,11 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Optional;
 import com.google.inject.Inject;
 import org.apache.druid.guice.annotations.Json;
+import org.apache.druid.indexing.common.task.Task;
+import org.apache.druid.indexing.overlord.DataSourceMetadata;
 import org.apache.druid.indexing.overlord.GlobalTaskLockbox;
 import org.apache.druid.indexing.overlord.IndexerMetadataStorageCoordinator;
+import org.apache.druid.indexing.overlord.SegmentPublishResult;
 import org.apache.druid.indexing.overlord.TaskRunner;
 import org.apache.druid.indexing.overlord.TaskRunnerFactory;
 import org.apache.druid.indexing.overlord.TaskStorage;
@@ -134,5 +137,28 @@ public class TaskActionToolbox
   public boolean canBatchSegmentAllocation()
   {
     return segmentAllocationQueue != null && segmentAllocationQueue.isEnabled();
+  }
+
+  /**
+   * Checks if the given publish action should be failed without allowing any
+   * more retries. A failed publish action should be retried only if there is
+   * another task waiting to publish offsets for an overlapping set of partitions.
+   */
+  public boolean shouldFailSegmentPublishImmediately(
+      SegmentPublishResult result,
+      Task task,
+      String supervisorId,
+      DataSourceMetadata startMetadata
+  )
+  {
+    if (result.isSuccess() || !result.isRetryable() || startMetadata == null) {
+      return false;
+    }
+
+    return !getSupervisorManager().isAnotherTaskGroupPublishingToPartitions(
+        supervisorId,
+        task.getId(),
+        startMetadata
+    );
   }
 }


server/src/main/java/org/apache/druid/segment/realtime/appenderator/TransactionalSegmentPublisher.java
```diff
@@ -33,8 +33,12 @@
...
 public abstract class TransactionalSegmentPublisher
 {
-  private static final int QUIET_RETRIES = 3;
-  private static final int MAX_RETRIES = 5;
+  private static final int QUIET_RETRIES = 5;
+
+  /**
+   * Approximately 10 minutes of retrying using {@link RetryUtils#nextRetrySleepMillis(int)}.
+   */
+  private static final int MAX_RETRIES = 13;
```
capistrant, line 41: Any concern about something that will never succeed no matter what take 10 min to fail here? Not sure what we could do about it anyways though. If this is needed to stabilize auto scaling then I guess it is worth it

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -2669,18 +2677,50 @@ private void verifyAndMergeCheckpoints(
...
+   * older taskGroups to finish publishing before it can publish its own offsets.
+   */
+  private boolean isAnotherTaskGroupPublishingToPartitionsOf(TaskGroup taskGroup)
+  {
+    final Set<PartitionIdType> partitionsPendingPublishFromOtherGroups =
+        pendingCompletionTaskGroups
+            .values()
+            .stream()
+            .flatMap(Collection::stream)
+            .filter(group -> !group.equals(taskGroup))
```
capistrant: TaskGroup doesn't have an override for the equals method. can we compare by group id or add an equals override? Or am I missing the point here where the reference equality is fine

kfaraz: Thanks for calling this out!

Yeah, I didn't override the equals/hashCode on purpose since each `TaskGroup` object is supposed to be distinct. I will add a comment to that effect.

Also the equality filter in this method is just a safe-side measure since the `taskGroup` passed into this method will always be one of the `activelyReadingTaskGroups` and it will be compared against `pendingCompletionTaskGroups`. The two sets will always be distinct except when the target `taskGroup` is moved from `activeReadingTaskGroups` to `pendingCompletionTaskGroups` while this method is in progress (which should not happen either since the notices are all handled by the single-threaded `exec`).

Let me know if that makes sense or if I should update the logic or add more comments.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/TransactionalSegmentPublisher.java
kfaraz: Yeah, there may still be cases that may not succeed even if after the 10 min window.
Handling that would require some kind of queueing mechanism for publishing tasks on the supervisor side.

I intend to explore that angle soon and make the `TaskGroup` more auto-scaler friendly as well. The `TaskGroup`/`groupId` concept is currently tightly tied to the assumption of a fixed task count.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
gianm: If you are intentionally using reference equality, it's best to use `==` to make the intent clear. This will also make it robust to someone possibly adding `equals` to `TaskGroup` in the future for some reason.

jtuglu1:
In the example above, how are we avoiding the case where we don't actually want to allow B to continue b/c it's using invalid offsets? For example, what if A fails to publish?

server/src/main/java/org/apache/druid/segment/realtime/appenderator/TransactionalSegmentPublisher.java
gianm: What's the rationale behind extending the retries? If it's useful, go ahead— 10 minutes isn't so bad— but I worry about retrying things that aren't possibly going to succeed.

kfaraz:
> In the example above, how are we avoiding the case where we don't actually want to allow B to continue b/c it's using invalid offsets? For example, what if A fails to publish?

@jtuglu1 , that case is already handled in the existing code.

After A publishes to fail, the code flow moves into the `else` block and we check if the checkpoints of B are consistent with the persisted offsets or not. Since they wouldn't be consistent, B will be killed off and a new task group C will be created.

As long as A is pending publish though, we have to assume that it is going to succeed and continue with B as usual.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -2626,8 +2634,8 @@ private void verifyAndMergeCheckpoints(
           taskGroup.checkpointSequences.clear();
           taskGroup.checkpointSequences.putAll(latestCheckpoints);
         } else {
-          log.debug(
-              "Adding task[%s] to kill list, checkpoints[%s], latestoffsets from DB [%s]",
+          log.warn(
+              "Adding task[%s] to kill list as its checkpoints[%s] are not consistent with latestoffsets from DB[%s]",
```
jtuglu1: nit: can we keep this `..checkpoints[%s]` so its more predictable to grep in logs?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -2641,8 +2649,8 @@ private void verifyAndMergeCheckpoints(
                                 .equals(taskGroup.checkpointSequences.firstEntry().getValue()))
             || taskCheckpoints.tailMap(taskGroup.checkpointSequences.firstKey()).size()
                != taskGroup.checkpointSequences.size()) {
-          log.debug(
-              "Adding task[%s] to kill list, checkpoints[%s], taskgroup checkpoints [%s]",
+          log.warn(
+              "Adding task[%s] to kill list as its checkpoints[%s] are not consistent with taskgroup checkpoints [%s]",
```
jtuglu1: nit: checkpoints[%s].

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -2669,18 +2677,50 @@ private void verifyAndMergeCheckpoints(
...
+
+    if (partitionsPendingPublishFromOtherGroups.isEmpty()) {
+      return false;
+    }
+
+    for (PartitionIdType partitionId : taskGroup.startingSequences.keySet()) {
+      if (partitionsPendingPublishFromOtherGroups.contains(partitionId)) {
+        log.info(
+            "taskGroup[%s] with sequenceName[%s] needs to wait before publishing"
+            + " as another taskGroup is currently publishing to partition[%s].",
```
jtuglu1: Can we log the other taskGroup

server/src/main/java/org/apache/druid/segment/realtime/appenderator/TransactionalSegmentPublisher.java
kfaraz: I am concerned about the case where we decide not to kill off B since A is currently pending publish,
and when it is finally time for B to publish, A still hasn't finished publishing.

In that case, it would make sense for B to retry for a while.
The current retry count of 5 amounts to only about 1 minute.

I do want to improve upon the retry algorithm such that we throw a retryable exception only if there is another task group pending publish for the partitions that the current task action is trying to update (in fact, let me try including that in this PR since we already have the new method `isAnotherGroupPublishing`).

jtuglu1: I wonder how many of these cases will actually be solved by extending retries. From what I've seen, 90%+ of the cases where B fails to commit are because task A also outright failed (not because it is slow to commit). In this case, we are prolonging the inevitable.

kfaraz: Yeah, that's a fair point.

I have updated `SegmentTransactionalAppendAction` and `SegmentTransactionalInsertAction` to return a retryable response only if there is a conflicting publish queued up. Otherwise, we will just fail the publish action and not retry anymore.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
kfaraz: Updated.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
kfaraz: I had to update the equality check since method signature has also changed. I am now simply checking the set of taskIds in the `TaskGroup` against the target `taskId`.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertAction.java
```diff
@@ -323,6 +328,19 @@ private static Map<Interval, List<DataSegment>> groupSegmentsByIntervalAndSort(S
     return segmentsMap;
   }
 
+  /**
+   * A failed publish action should be retried only if there is another task
+   * waiting to publish offsets for an overlapping set of partitions.
+   */
+  private boolean shouldNotRetryFailure(SegmentPublishResult result, Task task, TaskActionToolbox toolbox)
```
gianm: It would be clearer to call this `shouldFailImmediately`.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertAction.java
```diff
@@ -323,6 +328,19 @@ private static Map<Interval, List<DataSegment>> groupSegmentsByIntervalAndSort(S
...
+  {
+    if (result.isSuccess() || !result.isRetryable() || startMetadata != null) {
```
gianm: Was this meant to be `startMetadata == null`?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -452,7 +453,12 @@ static double extractMovingAverage(Map<String, Map<String, Object>> taskStats)
       }
     }
 
-    return count > 0 ? sum / count : -1;
+    if (count > 0) {
+      return sum / count;
+    } else {
+      return -1;
+    }
+    // return count > 0 ? sum / count : -1;
```
gianm: Please don't include commented-out code.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaIndexAutoScalingTest.java
```diff
@@ -0,0 +1,160 @@
...
+import org.joda.time.Duration;
+import org.joda.time.Period;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Work in progress.
```
gianm: I assume you intend to finish this before merging the PR.

kfaraz: Yes, still ironing out some kinks in the test.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertAction.java
kfaraz: Ah, yes, thanks for catching this!

abhishekrb19:
@kfaraz, thank you for fixing these ingestion errors: `Inconsistency between stored metadata state [KafkaDataSourceMetadata]`. We’ve seen these recurring from time to time in our production clusters as well, causing transient lag bursts, etc. A few questions:

1. Is this fix and https://github.com/apache/druid/pull/19034
 only applicable to auto-scaling, or are they broadly applicable when tasks roll over with task replicas enabled? We don’t have auto-scalers set up but still see these errors, so I’m wondering if these fixes would actually help us.

2. We hardcoded `MAX_RETRIES = 15` in `TransactionalSegmentPublisher` to account for slower publish times and occasional spurious task failures as a result of these metadata inconsistencies. It has been running in production for some time now. Some of the tables ingest a large volume of complex JSON objects, so we’d noticed slower handoff times and increasing this seemed to help to an extent. 

Is it worth considering making this property configurable rather than hardcoding it to 13 in this patch? Alternatively, maybe it could be derived dynamically as a function of `completionTimeout`?

kfaraz:
#### [1]

> are they broadly applicable when tasks roll over with task replicas enabled?

Yes, @abhishekrb19 , the `"Inconsistency between stored metadata"` error can sometimes occur when two replicas try to publish at around the same time. The first replica is able to publish segments successfully, while the other one would fail to publish. However, this is not a retryable failure and the slow replica just ends up marking the publish as success.

https://github.com/apache/druid/blob/8307d8a26c6b7286a8a57b85940a7f29d3d43248/server/src/main/java/org/apache/druid/segment/realtime/appenderator/BaseAppenderatorDriver.java#L670-L673

I assume you would be seeing more errors of the format below:
```
"The new start metadata state[%s] is ahead of the last committed end state[%s]. Try resetting the supervisor."
```
These are retryable errors which typically occur when publish is slow.
Please let me know if that is not the case.

#### [2]

> Is it worth considering making this property configurable rather than hardcoding it to 13 in this patch? Alternatively, maybe it could be derived dynamically as a function of completionTimeout?

I am not really a fan of the current retry mechanism , since it requires the task to blindly retry without knowing how long it should wait. Exponential backoff retries make more sense when the issue is transient (and likely self-resolving). Slow publish, on the other hand, is somewhat predictable/repeatable behaviour and we should probably be able to incorporate that in the retry timeout.

I don't want to make the number of retries configurable just yet. Instead, tying it to `completionTimeout` probably makes more sense. On top of that, I think the task could submit an "async" task action which kind of works like a long poll. The supervisor could keep a queue of pending publish actions, and as soon as this one finishes, we send back a success response. The `completionTimeout` (or some function of it) would then simply be used as the request timeout for this async action. Thus, no retries needed at all.

I want to merge the current patch and verify if the `isAnotherTaskGroupPublishing()` method is effective in identifying if there are prior pending publish actions. If it works as expected, I think we can build upon it further to get the queue semantics on the supervisor.

@gianm , @jtuglu1 , what are your thoughts?

kfaraz:
@gianm , @jtuglu1 , @capistrant , I have updated this PR with fixes for a couple of bugs that I discovered:

1. `CostBasedAutoScaler` does not scale tasks if processing rate is zero, irrespective of lag. When both lag and processing rate are low, it would make sense to scale down.
2. `CostBasedAutoScaler` scales down spuriously even when all candidates task counts have same cost. In such cases, leaving the task count unchanged is desirable.
3. `SeekableStreamSupervisorIOConfig` did not honor the correct priority order of `taskCountStart` > `taskCount` > `taskCountMin`.
    - This meant that even after `ioConfig` was updated and supervisor was persisted, supervisor spec didn't actually updated.
    - We didn't encounter this typically since the supervisor spec was being served from Overlord memory.
    - But the supervisor history didn't work since it is served from DB.
    - Overlord leadership change would cause the `taskCount` to revert back to 1.

gianm:
> * `CostBasedAutoScaler` scales down tasks if processing rate is zero, even if lag is high. (I guess this shouldn't happen in practice since if lag is high, tasks must be busy with some processing. Except when there is a bug and task threads are stuck somewhere.)

What'd you have it do in this situation? I suppose if processing rate is zero but lag is high, we shouldn't scale at all (either up or down). The combination of those two metrics suggests something is broken that scaling isn't going to fix.

> * `SeekableStreamSupervisorIOConfig` did not honor the correct priority order of `taskCountStart` > `taskCount` > `taskCountMin`

Huh, I thought we already changed that in #18745. Could you please also update the comment above the code block that currently says: `// if autoscaler is enabled, then taskCount will be ignored`.

Please also update `docs/ingestion/supervisor.md`. It says some things that are no longer accurate, such as saying in the docs for `taskCountMin` and `taskCountStart` that "When you enable the autoscaler, Druid ignores the value of taskCount".

> I want to merge the current patch and verify if the `isAnotherTaskGroupPublishing()` method is effective in identifying if there are prior pending publish actions. If it works as expected, I think we can build upon it further to get the queue semantics on the supervisor.
> 
> @gianm , @jtuglu1 , what are your thoughts?

I'm ok with going ahead with this patch given that more work is coming in this area, and that hopefully the `retryable` flag is going to prevent us from retrying things that will never succeed. I also believe that it would be better to move away from retries completely. The supervisor should be able to do more to help tasks sequence their publishes. That would also fix the long-standing issue that can happen with short-lived tasks: if task A is publishing, and its replacement task B runs for a short period of time and then starts publishing, it can finish before A and fail with a mismatched offsets error.

kfaraz:
> Huh, I thought we already changed that in https://github.com/apache/druid/pull/18745. Could you please also update the comment above the code block that currently says: // if autoscaler is enabled, then taskCount will be ignored.

Yeah, apparently, we missed a case in `SeekableStreamSupervisorIOConfig`.

> I also believe that it would be better to move away from retries completely. The supervisor should be able to do more to help tasks sequence their publishes.

Yes, agreed.

> I suppose if processing rate is zero but lag is high, we shouldn't scale at all (either up or down). The combination of those two metrics suggests something is broken that scaling isn't going to fix.

Yeah, that sounds reasonable. Let me add that check and maybe raise an alert for the same.

>> CostBasedAutoScaler scales down tasks if processing rate is zero, even if lag is high. (I guess this shouldn't happen in practice since if lag is high, tasks must be busy with some processing. Except when there is a bug and task threads are stuck somewhere.)

> What'd you have it do in this situation?

Oh, this was an old comment. I later rephrased it. The actual problem was 2-fold: (1) auto-scaler had a bug where it would do spurious scale downs even when cost was some for all task counts, and (2) auto-scaler did not take any action if processing rate was zero. Instead, it should have skipped scaling when processing rate was zero _but_ lag was high.

I will update the check accordingly.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -161,29 +172,26 @@ public int computeTaskCountForRollover()
   public int computeTaskCountForScaleAction()
   {
     lastKnownMetrics = collectMetrics();
-    if (lastKnownMetrics == null) {
-      log.debug("Metrics not available for supervisorId [%s], skipping scaling action", supervisorId);
-      return -1;
-    }
 
     final int optimalTaskCount = computeOptimalTaskCount(lastKnownMetrics);
     final int currentTaskCount = lastKnownMetrics.getCurrentTaskCount();
```
capistrant: NPE seems possible here now that there is no null check above. Looks like `collectMetrics()` is nullable only when supervisor is suspended, so perhaps unlikely/impossible we hit NPE here if scaler doesn't run in suspended state? Either way, `computeOptimalTaskCount` code path handles the `null` metrics state, so should we add a guard here too?

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertAction.java
```diff
@@ -323,6 +328,24 @@ private static Map<Interval, List<DataSegment>> groupSegmentsByIntervalAndSort(S
     return segmentsMap;
   }
 
+  /**
+   * Checks if the current publish action should be failed without allowing any
+   * more retries. A failed publish action should be retried only if there is
+   * another task waiting to publish offsets for an overlapping set of partitions.
+   */
+  private boolean shouldFailImmediately(SegmentPublishResult result, Task task, TaskActionToolbox toolbox)
```
capistrant: applies to both impls of `shouldFailImmediately` 

I am confused, if `isAnotherTaskGroupPublishingToPartitions` returns true doesn't that mean we **do not** want to fail immediately? But the way I read this is that this case returns `true`, fail immediately

kfaraz: ah, good catch!

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
kfaraz: Yeah, fair point, let me add a check here.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalAppendAction.java
```diff
@@ -219,7 +219,30 @@ public SegmentPublishResult perform(Task task, TaskActionToolbox toolbox)
...
+  }
+
+  /**
+   * Checks if the current publish action should be failed without allowing any
+   * more retries. A failed publish action should be retried only if there is
+   * another task waiting to publish offsets for an overlapping set of partitions.
+   */
+  private boolean shouldFailImmediately(SegmentPublishResult result, Task task, TaskActionToolbox toolbox)
+  {
+    if (result.isSuccess() || !result.isRetryable() || startMetadata == null) {
```
jtuglu1: I suppose this might regress the case where task B reads before task A commits the write, then task B proceeds to write and fails (because older task A publish completed), failing, then checking:
```
if there is another task waiting to publish offsets for an overlapping set of partitions.
```
which of course there isn't since task A completed then proceeds to fail(even though on retries it would succeed)?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -427,6 +430,48 @@ public boolean registerUpgradedPendingSegmentOnSupervisor(
...
+  public boolean isAnotherTaskGroupPublishingToPartitions(
+      String supervisorId,
+      String taskId,
+      DataSourceMetadata startMetadata
+  )
+  {
+    try {
+      InvalidInput.conditionalException(supervisorId != null, "'supervisorId' cannot be null");
+      if (!(startMetadata instanceof SeekableStreamDataSourceMetadata<?, ?>)) {
+        throw InvalidInput.exception("Start metadata[%s] is not streaming metadata", supervisorId);
```
jtuglu1: not sure when/how this could occur, but might want to log the type here.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -427,6 +430,48 @@ public boolean registerUpgradedPendingSegmentOnSupervisor(
...
+      }
+
+      Pair<Supervisor, SupervisorSpec> supervisor = supervisors.get(supervisorId);
+      if (supervisor == null) {
+        throw NotFound.exception("Could not find supervisor[%s]", supervisorId);
+      }
+      if (!(supervisor.lhs instanceof SeekableStreamSupervisor<?, ?, ?>)) {
+        throw InvalidInput.exception("Supervisor[%s] is not a streaming supervisor", supervisorId);
```
jtuglu1: same about the type

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -62,10 +63,17 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
...
+  public static final String INVALID_METRICS_COUNT = "task/autoScaler/costBased/invalidMetrics";
 
   static final int MAX_INCREASE_IN_PARTITIONS_PER_TASK = 2;
   static final int MAX_DECREASE_IN_PARTITIONS_PER_TASK = MAX_INCREASE_IN_PARTITIONS_PER_TASK * 2;
 
+  /**
+   * If average partition lag crosses this value and the processing rate is
+   * still zero, scaling actions are skipped and an alert is raised.
+   */
+  static final int MAX_IDLENESS_PARTITION_LAG = 10_000;
```
jtuglu1, line 75: I'm not too crazy about hardcoding this value. Could we expose this as a config? The reason I ask it's totally feasible for a topic to never reach above 10k event lag (it could only see a few 100 event lag for critical topics).

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -161,29 +172,26 @@ public int computeTaskCountForRollover()
...
                && optimalTaskCount < currentTaskCount
                && optimalTaskCount > 0) {
       taskCount = optimalTaskCount;
       lastScaleActionTimeMillis = DateTimes.nowUtc().getMillis();
-      log.info("New task count [%d] on supervisor [%s], scaling down", taskCount, supervisorId);
+      log.info("Updating taskCount for supervisor[%s] from [%d] to [%d] (scale down).", supervisorId, currentTaskCount, taskCount);
     } else {
-      log.info("No scaling required for supervisor [%s]", supervisorId);
+      taskCount = -1;
+      log.debug("No scaling required for supervisor [%s]", supervisorId);
```
jtuglu1: `supervisor [%s] -> supervisor[%s]`

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -2626,8 +2640,8 @@ private void verifyAndMergeCheckpoints(
           taskGroup.checkpointSequences.clear();
           taskGroup.checkpointSequences.putAll(latestCheckpoints);
         } else {
-          log.debug(
-              "Adding task[%s] to kill list, checkpoints[%s], latestoffsets from DB [%s]",
+          log.warn(
+              "Adding task[%s] to kill list as its checkpoints[%s] are not consistent with checkpoints[%s] in metadata store",
```
jtuglu1, line 2644: I wonder if it's useful to create a checkpoint diff function for logs so users don't need to manually compare the 2 checkpoints?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -2911,7 +2964,13 @@ protected boolean supportsPartitionExpiration()
 
   public int getPartitionCount()
   {
-    return recordSupplier.getPartitionIds(ioConfig.getStream()).size();
+    recordSupplierLock.lock();
```
jtuglu1, line 2967: nice: this fixes a race I've seen with auto-scaler racing with recordsupplier on reads.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
jtuglu1, line 3496: do we not need to set `hasTaskRolloverStarted.set(true);` here as well for completeness? For example, what if a user manually triggers this handoff from the UI – should we not check to scale then?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
kfaraz: Yeah, it will be useful to have. We can try to add it in a follow up.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
kfaraz: > The reason I ask it's totally feasible for a topic to never reach above 10k event lag

Yeah, that was kind of the intention here.
If the lag remains below this value, autoscaling will work as usual.
But if the lag exceeds this value AND processing rate is zero, that indicates something is wrong with the tasks. In such a case, scaling will be skipped and an alert will be raised.

Let me know if you still feel that we need this as a config in this PR.

(I was initially using the config value `CostBasedAutoScalerConfig.highLagThreshold` for this, but it is also being used for another completely different purpose. So I decided not to overload the meaning of that config.)

jtuglu1: > But if the lag exceeds this value AND processing rate is zero, that indicates something is wrong with the tasks. 

I guess my point is we have topics where exceeding 10k is probably too late to detect something is up (we've already broken an SLO). We can leave it for now to avoid config bloat, but I don't really like to hardcode this stuff. IMO, when we start to add more tweakable configs/magic numbers to the solution it points at a larger underlying issue.

kfaraz: Yeah, fair point, I will add it as a part of config cleanup in the `CostBasedAutoScaler`.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalAppendAction.java
kfaraz: Do you mean the case where task B was launched with incorrect start offsets?

Under normal conditions, start offsets of task B are the same as end offsets of task A (even if they have not been committed yet). This was addressed in the previous PR #19034 .

Then if task B tries to commit before A, it will be asked to retry.
If task B tries to commit after A, it will succeed since the metadata store has now been updated with end offsets of A.

Please let me know if you were referring to some other case.

jtuglu1: > Under normal conditions, start offsets of task B are the same as end offsets of task A (even if they have not been committed yet). This was addressed in the previous PR https://github.com/apache/druid/pull/19034 .

Missed this – then I think this is covered.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
kfaraz: Yeah, I guess we could do that too. Since we are stopping tasks anyway, we might as well consider scaling the taskCount.

kfaraz: Updated. I am now setting `hasTaskRolloverStarted` when `stopTasksEarly` is set to `true`.
Not setting it when `group.getHandoffEarly()` is true since that might be limited to only some task groups.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
kfaraz: updated.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
kfaraz: updated.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
kfaraz: done.

kfaraz:
Thanks for the reviews, @capistrant , @jtuglu1 !
I have addressed the latest comments.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertAction.java
jtuglu1: One thing that might be nice is to move `SegmentTransactionalInsertAction::shouldFailImmediately` and `SegmentTransactionalAppendAction::shouldFailImmediately` to a shared common method so we don't need to maintain 2 copies of the same logic.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -90,8 +91,11 @@ public SeekableStreamSupervisorIOConfig(
     this.autoScalerEnabled = autoScalerConfig != null && autoScalerConfig.getEnableTaskAutoScaler();
     // if autoscaler is enabled, then taskCount will be ignored here and initial taskCount will equal to taskCountStart/taskCountMin
```
gianm: This comment is no longer accurate, please fix it.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -501,6 +510,56 @@ CostMetrics collectMetrics()
...
+    if (metrics == null) {
+      return Either.error("No metrics collected");
+    } else if (metrics.getAvgProcessingRate() < 0 || metrics.getPollIdleRatio() < 0) {
+      return Either.error("Task metrics not available");
+    } else if (metrics.getCurrentTaskCount() <= 0 || metrics.getPartitionCount() <= 0) {
+      return Either.error("Supervisor metrics not available");
+    } else if (metrics.getAvgPartitionLag() < 0) {
+      return Either.error("Lag metrics not available");
+    } else if (metrics.getAvgProcessingRate() < 1 && metrics.getAvgPartitionLag() > MAX_IDLENESS_PARTITION_LAG) {
+      log.makeAlert(
```
gianm, line 553: After calling `makeAlert` you get an `AlertBuilder` that needs to have `emit()` called on it. As written, it will not actually emit.

kfaraz: Fixed.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
kfaraz: Removed.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalInsertAction.java
kfaraz: Updated. Moved to `TaskActionToolbox`.

PR #19094, opened 2026-03-05
Add `strict` and `pooled` Broker tier selector strategies

abhishekrb19:
### Summary

Adds two new Broker `TierSelectorStrategy` implementations to provide finer control over how Brokers select Historical and Realtime servers for query execution. 

- **strict** – Only selects servers whose priorities match the configured list.  
  Unlike other existing strategies, there is no fallback to servers with other priorities if the configured priorities are unavailable. This also addresses a current limitation with watched tiers: when multiple tiers are configured, Brokers can still retain visibility into the state of the cluster, while enforcing query isolation at the time of server selection rather than filtering servers at the time of building the Broker's server view.

- **pooled** – Pools servers across the configured priorities and selects among them, allowing queries to utilize multiple priority tiers for improved availability. This is particularly useful for querying realtime servers where the number of task replicas per tier may be limited for cost reasons.

Both strategies require the configured set of priorities to be non-empty. Similar to queries routed to tiers that are not part of the watched tiers, these strategies may result in queries returning no data if the configured tiers are unavailable.

### Release note

Operators can now configure two new Broker `TierSelectorStrategy` implementations:

- strict - Only selects servers whose priorities match the configured list.  Example configuration: `druid.broker.select.tier=strict` and `druid.broker.select.tier.strict.priorities=[1]`. 
- pooled - Pools servers across the configured priorities and selects among them, allowing queries to utilize multiple priority tiers for improved availability. Example configuration: `druid.broker.select.tier=pooled` and `druid.broker.select.tier.pooled.priorities=[2,1]`. 
These strategies can also be configured for Realtime servers using `druid.broker.realtime.select.tier`.

This PR has:

- [x] been self-reviewed.
- [ ] added documentation for new or modified features or behaviors.  (Will add docs in a follow up along with previous realtime tier configuration)

### DIFF
--- a/examples/conf/druid/auto/broker/runtime.properties
+++ b/examples/conf/druid/auto/broker/runtime.properties
@@ -39,3 +39,7 @@ druid.processing.tmpDir=var/druid/processing
 # Query cache disabled -- push down caching and merging instead
 druid.broker.cache.useCache=false
 druid.broker.cache.populateCache=false
+
+# Poll segments from the Coordinator quickly (default is 60s) so sys.segments updates more quickly
+# This config should generally be raised to reduce load once you are running at scale
+druid.sql.planner.metadataSegmentPollPeriod=5000
--- a/examples/conf/druid/auto/coordinator-overlord/runtime.properties
+++ b/examples/conf/druid/auto/coordinator-overlord/runtime.properties
@@ -22,6 +22,9 @@ druid.plaintextPort=8081
 
 druid.coordinator.startDelay=PT10S
 druid.coordinator.period=PT5S
+
+# Poll segments from the metadata store quickly (default is 60s) so newly ingeseted segments load more quickly
+# This config should generally be raised to reduce load once you are running at scale
 druid.manager.segments.pollDuration=PT5S
 
 # Run the overlord service in the coordinator process
--- a/examples/conf/druid/cluster/master/coordinator-overlord/runtime.properties
+++ b/examples/conf/druid/cluster/master/coordinator-overlord/runtime.properties
@@ -22,6 +22,9 @@ druid.plaintextPort=8081
 
 druid.coordinator.startDelay=PT10S
 druid.coordinator.period=PT5S
+
+# Poll segments from the metadata store quickly (default is 60s) so newly ingeseted segments load more quickly
+# This config should generally be raised to reduce load once you are running at scale
 druid.manager.segments.pollDuration=PT5S
 
 # Run the overlord service in the coordinator process
--- a/examples/conf/druid/cluster/query/broker/runtime.properties
+++ b/examples/conf/druid/cluster/query/broker/runtime.properties
@@ -35,3 +35,7 @@ druid.processing.tmpDir=var/druid/processing
 # Query cache disabled -- push down caching and merging instead
 druid.broker.cache.useCache=false
 druid.broker.cache.populateCache=false
+
+# Poll segments from the Coordinator quickly (default is 60s) so sys.segments updates more quickly
+# This config should generally be raised to reduce load once you are running at scale
+druid.sql.planner.metadataSegmentPollPeriod=5000
--- a/examples/conf/druid/single-server/large/broker/runtime.properties
+++ b/examples/conf/druid/single-server/large/broker/runtime.properties
@@ -35,3 +35,7 @@ druid.processing.tmpDir=var/druid/processing
 # Query cache disabled -- push down caching and merging instead
 druid.broker.cache.useCache=false
 druid.broker.cache.populateCache=false
+
+# Poll segments from the Coordinator quickly (default is 60s) so sys.segments updates more quickly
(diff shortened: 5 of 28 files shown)


server/src/main/java/org/apache/druid/client/selector/PooledTierSelectorStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    final List<QueryableDruidServer> selectedServers = serverSelectorStrategy.pick(query, candidateServerPool, segment, numServersToPick);
+    log.debug("Selected servers[%s] for query[%s] from given servers[%s] and candidateServerPool[%s]", selectedServers, query, prioritizedServers, candidateServerPool);
+    return selectedServers;
+  }
+
+  /**
+   * @return the natural order of priorities since priority order doesn't matter for this strategy as the configured set of
+   * priorities in the pool are treated equally and delegated ot {@link #serverSelectorStrategy}.
```
aho135:
```suggestion
   * priorities in the pool are treated equally and delegated to {@link #serverSelectorStrategy}.
```

server/src/main/java/org/apache/druid/client/selector/PooledTierSelectorStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+  public static final String TYPE = "pooled";
+
+  private final ServerSelectorStrategy serverSelectorStrategy;
+  private final PooledTierSelectorStrategyConfig config;
+  private final ServiceEmitter emitter;
+  private final Set<Integer> configuredPriorities;
+
+  @JsonCreator
+  public PooledTierSelectorStrategy(
+      @JacksonInject final ServerSelectorStrategy serverSelectorStrategy,
```
aho135, line 69: nit: We could make ServerSelectorStrategy in AbstractTierSelectorStrategy `protected`

server/src/main/java/org/apache/druid/client/selector/StrictTierSelectorStrategy.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A {@link TierSelectorStrategy} that only considers servers whose priorities
+ * are explicitly listed in {@link StrictTierSelectorStrategyConfig#getPriorities()}.
+ * <p>
+ * Unlike other strategies like {@link CustomTierSelectorStrategy} that falls back to servers with different priorities,
```
aho135:
```suggestion
 * Unlike other strategies like {@link CustomTierSelectorStrategy} that fall back to servers with different priorities,
```

server/src/main/java/org/apache/druid/client/selector/StrictTierSelectorStrategy.java
```diff
@@ -0,0 +1,183 @@
...
+      } else {
+        log.debug(
+            "Server priority[%d] not in the configured list of priorities[%s] so ignore servers[%s] for query[%s]",
+            priority, config.getPriorities(), servers, query
+        );
+      }
+    }
+
+    if (candidatePrioritizedServers.isEmpty()) {
+      if (query == null || query instanceof SegmentMetadataQuery) {
```
aho135, line 136: Would we want to include dataSourceMetadata queries here as well?

Maybe a small helper function in Query: isMetadataQuery()?

server/src/test/java/org/apache/druid/client/selector/TierSelectorStrategyTest.java
```diff
@@ -568,4 +594,326 @@ public void testPreferredTierSelectorStrategyDefaultPriority()
         preferredTierHighPriority, preferredTierLowPriority, nonPreferredTierHighestPriority
     );
   }
+
```
aho135, line 597: Maybe one test case where we have partial priority match. e.g. Servers with priority [-1, 0, 2] but StrictTierSelectorStrategy configured with [0,1]

server/src/main/java/org/apache/druid/client/selector/StrictTierSelectorStrategy.java
abhishekrb19: I think the legacy `dataSourceMetadata` queries can be treated the same as other query types. Unlike system-generated queries such as `segmentMetadata`, these would be user-initiated and won't likely cause alert/log spam

server/src/test/java/org/apache/druid/client/selector/TierSelectorStrategyTest.java
abhishekrb19: Thanks, good call! Added a case for these mixed priorities.

abhishekrb19:
Thanks for the review @aho135! The Docker test failures `test_runIndexTask_andKillData ` and `test_runIndexParallelTask_andCompactData` are unrelated to this patch (the first one is a usual suspect)

Issue #19092, opened 2026-03-05
CascadingReindexingTemplate Production Readiness


capistrant:
#18939 added a new template that can be used with Compaction supervisors. It is experimental in nature, and will be experimental in Druid 37, but I'd like to scope out work needed to get it as close to production ready as possible for Druid 37 so that we can pull the experimental tag in Druid 38

# Todo

- [x] drop native compaction support
- [x] refactor rules concept
- [x] documentation
- [ ] console UI
    - [ ] phase 1 - visualization of existing supervisors timeline
    - [ ] phase 2 - builder helper for building a timeline
- [ ] catalog backed rule provider

### Drop Native Compaction Support.

This was recently completed in #19078. Rationale here is that MSQ engine is the desired future of compaction supervisors. Supporting two engines is extra overhead, can lead to confusion, and slows development of more powerful features.

### Refactor Rules Concept

This warrants a sub issue perhaps. But with #19061 adding CLUSTERED BY support on expressions, it has exposed some gaps in the initial rule structure that is coupled closely with compaction state. The original proposal for #18939 was even more coupled than the version that got merged. In the merged version we have `ReindexingDataSchemaRule` which is more general and swallows up compaction state concepts for dims, metrics, query gran, etc.

We are thinking we need to go further with rule refactoring. One thought is to create a rule that encompasses partitioning ideas like segment gran, partitions spec, etc. The tuning config rule stuff that is pretty static in nearly all cases would be moved outside of the rules and into the top level config for the template. This would be shared by all generated tasks.

### Documentation

As the design has been in flux, we have not committed to any druid docs yet. Once the above two sections are finalized a new set of docs about this template + integration with the existing automatic compaction docs is needed. This will be an experimental feature in d37. The docs should make it easy to see the value of this template + guide an operator to onboard to it and evaluate it

### UI in the router

https://github.com/apache/druid/pull/19051 is open for adding visualization of the timeline of search intervals and compaction configs generated by a ruleset for an existing supervisor

I also think it would be cool to add a supervisor builder UI where you can visualize the timeline as you create/edit a supervisor spec instead of only when it already exists.

### More robust rule provider

Right now the only rule provider is `inline` which works and is nice for many smaller use cases, but it becomes hard to reason about quickly. A rule provider in core that sources rules from something like druid-catalog feels like a more production ready long term approach

PR #19106, opened 2026-03-06
refactor: CascadingReindexingTemplate Refactoring

capistrant:
### Description

relates to #19092 

After taking the initial implementation of the CascadingReindexingTemplate and associated RuleProvider concepts for a spin, some refactoring opportunities have been identified. I hope that these will make more conceptual sense to operators trying to configure this as well as make things generally easier to reason about.

#### Rule Type Refactoring

`ReindexingSegmentGranularityRule` and `ReindexingTuningConfigRule` have been deleted.

##### ReindexingPartitioningRule

This new rule type consolidates partitioning details into a single rule. This rule takes the place of `ReindexingSegmentGranularityRule` in that it has special consideration for creating the base timeline of reindexing search intervals and that all search intervals created align to natural Segment Granularity boundaries being applied.

The components of this rule are:
* [required] Segment Granularity (Limited to the same strict set of granularities that the original rule type was)
* [required] Partition Spec
* [optional] virtual columns
  * This is an integration with #19061 so that the new range partitioning with expressions is supported by reindexing

 ##### ReindexingIndexSpecRule

A new rule type that is scoped to the `IndexSpec` that goes in a task tuning config.

#### `CascadingReindexingTemplate` new top level fields

I have considered combining the below 2 + defaultSegmentGranularity into a defaultPartitioningSomething top level field but did not do this as of now.

##### `defaultPartitionsSpec`

Required partitionsspec that will be applied in cases where a synthetic `ReindexingPartitioningRule` needs to be generated to complete the reindexing timeline

##### `defaultPartitioningVirtualColumns`

Optional default partitioning virtual column expressions that would be used if your default partitionspec utilized the new functionality in #19061 

##### `tuningConfig`

`CascadingReindexingTemplate` now has an optional top level `tuningConfig` field where an operator can provide a `UserCompactionTaskQueryTuningConfig` that will apply to all tasks generated by the supervisor. Note that any partitions spec here is useless and will be blown away by either the defaultPartitionsSpec or the applicable ReindexingPartitioningRule's partition spec. IndexSpec however can take a default here that will apply in any case where there is not an IndexSpecRule that applies to the task being generated

#### Release note

N/A for now but we will have a larger release note I can share encompassing all of the new reindexing rule provider stuff for druid 37.

-->

<hr>

 * `ReindexingConfigBuilder`
 * `ReindexingPartitioningRule`
 * `ReindexingIndexSpecRule`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
@@ -139,6 +139,13 @@ public class CascadingReindexingTemplate implements CompactionJobTemplate, DataS
 
     this.defaultPartitioningVirtualColumns = defaultPartitioningVirtualColumns;
 
+    if (tuningConfig != null && tuningConfig.getPartitionsSpec() != null) {
+      throw InvalidInput.exception(
+          "Cannot set 'partitionsSpec' inside 'tuningConfig' for a cascading reindexing supervisor. "
+          + "Partitioning is controlled by 'defaultPartitionsSpec' and partitioning rules. "
+          + "Any 'partitionsSpec' in 'tuningConfig' would be ignored."
+      );
+    }
     this.tuningConfig = tuningConfig;
 
     if (skipOffsetFromNow != null && skipOffsetFromLatest != null) {
--- a/indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
+++ b/indexing-service/src/test/java/org/apache/druid/indexing/compact/CascadingReindexingTemplateTest.java
@@ -45,6 +45,7 @@ import org.apache.druid.server.compaction.ReindexingRule;
 import org.apache.druid.server.compaction.ReindexingRuleProvider;
 import org.apache.druid.server.coordinator.DataSourceCompactionConfig;
 import org.apache.druid.server.coordinator.InlineSchemaDataSourceCompactionConfig;
+import org.apache.druid.server.coordinator.UserCompactionTaskQueryTuningConfig;
 import org.apache.druid.testing.InitializedNullHandlingTest;
 import org.apache.druid.timeline.DataSegment;
 import org.apache.druid.timeline.SegmentTimeline;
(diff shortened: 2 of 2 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
```diff
@@ -123,11 +134,24 @@ public CascadingReindexingTemplate(
...
       throw InvalidInput.exception("Cannot set both skipOffsetFromNow and skipOffsetFromLatest");
     }
     this.skipOffsetFromNow = skipOffsetFromNow;
     this.skipOffsetFromLatest = skipOffsetFromLatest;
+
+    this.defaultPartitioningRule = ReindexingPartitioningRule.syntheticRule(
+        defaultSegmentGranularity,
+        defaultPartitionsSpec,
+        defaultPartitioningVirtualColumns
+    );
```
clintropolis, line 161: should this check if it is set on tuningConfig? or forbid defining it on the tuningConfig?

indexing-service/src/main/java/org/apache/druid/indexing/compact/ReindexingConfigBuilder.java
```diff
@@ -128,64 +138,126 @@ BuildResult applyToWithDetails(InlineSchemaDataSourceCompactionConfig.Builder bu
...
+  @Nullable
+  private static VirtualColumns mergeVirtualColumns(
+      @Nullable VirtualColumns partitioningVCs,
+      List<VirtualColumn> deletionVCs
+  )
+  {
+    List<VirtualColumn> allVCs = new ArrayList<>(deletionVCs);
+
+    if (partitioningVCs != null && !partitioningVCs.isEmpty()) {
+      // Check for name collisions
```
clintropolis: `VirtualColumns.create` does a check like this too, though i guess you lose your more specific exception messaging here. I suppose you could catch the `IAE` thrown by that and make the nicer message that way, or leave a comment about why validation is being done externally here

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
capistrant: by `it` do you mean partitionsSpec? I had planned on documenting it as being blown away by whatever your 

1. applicable rule defines
2. default top level partitions spec is if no rule applies

alternatives to the above:

* force a tuningConfig to be defined top level with at least a partitions spec and use that as the default
* instead of silently blowing away the static tuningConfig partitions spec, throw an error at submit time saying you can't define it there if someone tries

indexing-service/src/main/java/org/apache/druid/indexing/compact/ReindexingConfigBuilder.java
capistrant: huh, I hadn't realized that `create` could handle this for me. I like delegating to that and enriching the message if thrown

indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
clintropolis: >by it do you mean partitionsSpec?

yes

> I had planned on documenting it as being blown away ...

it is probably fine if we document that whatever you set there is ignored. Maybe perhaps a log.warn or something if it is set on the tuningconfig but a `defaultPartitionsSpec` nor rules are set?

capistrant: after reflecting, I think failing with clear error on trying to use unsupported field in the static tuningConfig is ok and right way to head off operator confusion

PR #19107, opened 2026-03-07
feat: add OpenLineage request logger extension

mshahid6:
### Description
Added `extensions-contrib/openlineage-emitter` as a contrib extension that uses the `RequestLogger` to transform and send lineage information to any [OpenLineage](https://openlineage.io)-compatible API.
For SQL queries, the SQL text is parsed with the Calcite parser to extract input datasources (FROM clauses, JOINs, CTEs) and output datasources (INSERT INTO). For native queries, table names are read from `DataSource.getTableNames()`. Native sub-queries spawned by a SQL execution are deduplicated against the SQL-level event.
Each event includes standard OpenLineage facets (`processing_engine`, `jobType`, `sql`,`errorMessage`) and custom Druid facets (`druid_query_context` with user identity and query metadata, `druid_query_statistics` with duration and bytes).

Transport is configurable: `CONSOLE` (default) logs JSON to the Druid log; `HTTP` POSTs to an OpenLineage endpoint such as [Marquez](https://marquezproject.ai). Can be combined with other loggers via the `composing` provider.

Screenshot 2026-05-13 at 3 21 30 PM

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/docs/development/extensions-contrib/openlineage-emitter.md
+++ b/docs/development/extensions-contrib/openlineage-emitter.md
@@ -1,0 +1,95 @@
+---
+id: openlineage-emitter
+title: "OpenLineage Emitter"
+---
+
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+
+To use this Apache Druid extension, [include](../../configuration/extensions.md#loading-extensions) `openlineage-emitter` in the extensions load list.
+
+## Introduction
+
+This extension emits [OpenLineage](https://openlineage.io) `RunEvent`s for each completed Druid query, enabling data lineage tracking with any OpenLineage-compatible backend such as [Marquez](https://marquezproject.ai).
+
+For SQL queries, the SQL text is parsed to extract input datasources (FROM clauses, JOINs, CTEs) and the output datasource (INSERT INTO). For native queries, table names are resolved from the datasource tree. Native sub-queries spawned by a SQL execution are deduplicated against the SQL-level event.
+
+:::note
+SQL table extraction relies on `calcite-core` being on the classpath, which is the case on Broker nodes. Native query lineage is available on all nodes.
+:::
+
+## Configuration
+
+All configuration parameters are under `druid.request.logging`.
+
+| Property | Description | Required | Default |
+|---|---|---|---|
+| `druid.request.logging.type` | Set to `openlineage` to enable this extension. | yes | — |
+| `druid.request.logging.namespace` | Namespace used for OpenLineage job and dataset URIs. Typically the Broker URL. | no | `druid://localhost` |
+| `druid.request.logging.transportType` | Where to send events. `CONSOLE` logs JSON to the Druid log; `HTTP` POSTs to an OpenLineage API endpoint. | no | `CONSOLE` |
+| `druid.request.logging.transportUrl` | OpenLineage API endpoint URL. Required when `transportType=HTTP`. | no | — |
+
+### Examples
+
+**Console (development)**
+
+```properties
+druid.request.logging.type=openlineage
+druid.request.logging.namespace=druid://broker.prod:8082
+```
+
+**HTTP (production)**
+
+```properties
+druid.request.logging.type=openlineage
+druid.request.logging.namespace=druid://broker.prod:8082
+druid.request.logging.transportType=HTTP
+druid.request.logging.transportUrl=http://marquez:5000/api/v1/lineage
+```
+
+**Combined with another logger using the `composing` provider**
+
+```properties
+druid.request.logging.type=composing
+druid.request.logging.loggerProviders[0].type=slf4j
+druid.request.logging.loggerProviders[1].type=openlineage
+druid.request.logging.loggerProviders[1].namespace=druid://broker.prod:8082
+druid.request.logging.loggerProviders[1].transportType=HTTP
+druid.request.logging.loggerProviders[1].transportUrl=http://marquez:5000/api/v1/lineage
+```
+
+## Event structure
+
+Each emitted event follows the [OpenLineage spec](https://openlineage.io/spec/2-0-2/OpenLineage.json) and includes the following facets.
+
+### Run facets
+
+| Facet | Description |
+|---|---|
+| `processing_engine` | Engine name (`druid`). Standard OpenLineage facet. |
+| `druid_query_context` | Query metadata: `identity` (authenticated user), `remoteAddress`, `queryType`, and `nativeQueryIds` (for SQL queries). |
+| `druid_query_statistics` | Execution stats: `durationMs`, `bytes`, `planningTimeMs`, `statusCode`. |
+| `errorMessage` | Exception message for failed queries. Standard OpenLineage facet. |
+
+### Job facets
+
+| Facet | Description |
+|---|---|
+| `jobType` | `processingType=BATCH`, `integration=DRUID`, `jobType=QUERY`. Standard OpenLineage facet. |
+| `sql` | Raw SQL text. Present on SQL queries only. Standard OpenLineage facet. |
--- a/docs/development/extensions-contrib/openlineage-emitter.md
+++ b/docs/development/extensions-contrib/openlineage-emitter.md
@@ -41,9 +41,10 @@ All configuration parameters are under `druid.request.logging`.
 | Property | Description | Required | Default |
 |---|---|---|---|
 | `druid.request.logging.type` | Set to `openlineage` to enable this extension. | yes | — |
-| `druid.request.logging.namespace` | Namespace used for OpenLineage job and dataset URIs. Typically the Broker URL. | no | `druid://localhost` |
+| `druid.request.logging.namespace` | Namespace used for OpenLineage job and dataset URIs. Typically the Broker URL. | no | `druid://<hostname>` |
 | `druid.request.logging.transportType` | Where to send events. `CONSOLE` logs JSON to the Druid log; `HTTP` POSTs to an OpenLineage API endpoint. | no | `CONSOLE` |
 | `druid.request.logging.transportUrl` | OpenLineage API endpoint URL. Required when `transportType=HTTP`. | no | — |
+| `druid.request.logging.excludedNativeQueryTypes` | Native query types to exclude from lineage emission. Internal broker queries like segment metadata lookups produce noisy, low-value events. | no | `["segmentMetadata", "dataSourceMetadata", "timeBoundary"]` |
 
 ### Examples
 
@@ -67,11 +68,7 @@ druid.request.logging.transportUrl=http://marquez:5000/api/v1/lineage
 
 ```properties
 druid.request.logging.type=composing
-druid.request.logging.loggerProviders[0].type=slf4j
-druid.request.logging.loggerProviders[1].type=openlineage
-druid.request.logging.loggerProviders[1].namespace=druid://broker.prod:8082
-druid.request.logging.loggerProviders[1].transportType=HTTP
-druid.request.logging.loggerProviders[1].transportUrl=http://marquez:5000/api/v1/lineage
+druid.request.logging.loggerProviders=[{"type":"slf4j"},{"type":"openlineage","namespace":"druid://broker.prod:8082","transportType":"HTTP","transportUrl":"http://marquez:5000/api/v1/lineage"}]
 ```
 
 ## Event structure
--- a/docs/development/extensions-contrib/openlineage-emitter.md
+++ b/docs/development/extensions-contrib/openlineage-emitter.md
@@ -28,7 +28,7 @@ To use this Apache Druid extension, [include](../../configuration/extensions.md#
 
 This extension emits [OpenLineage](https://openlineage.io) `RunEvent`s for each completed Druid query, enabling data lineage tracking with any OpenLineage-compatible backend such as [Marquez](https://marquezproject.ai).
 
-For SQL queries, the SQL text is parsed to extract input datasources (FROM clauses, JOINs, CTEs) and the output datasource (INSERT INTO). For native queries, table names are resolved from the datasource tree. Native sub-queries spawned by a SQL execution are deduplicated against the SQL-level event.
+For MSQ DML statements (`INSERT INTO` / `REPLACE INTO`), the output datasource is extracted from the SQL text and emitted as an output dataset. Input extraction from SQL is not performed — reliably resolving `FROM` / `JOIN` tables at the logger layer would duplicate planner work. For native queries, input table names are resolved from the datasource tree and emitted as input datasets. Native sub-queries spawned by a SQL execution carry a `sqlQueryId` in their context facet for correlation with the parent SQL event.
 
 :::note
 SQL table extraction relies on `calcite-core` being on the classpath, which is the case on Broker nodes. Native query lineage is available on all nodes.
@@ -48,9 +48,9 @@ All configuration parameters are under `druid.request.logging`.
 | `druid.request.logging.emitQueueCapacity` | Maximum number of events buffered in the async HTTP emit queue. Events are dropped (with a warning) when the queue is full. Only applies when `transportType=HTTP`. | no | `1000` |
 | `druid.request.logging.emitThreadCount` | Number of background threads used to POST events to the HTTP endpoint. Only applies when `transportType=HTTP`. | no | `1` |
 | `druid.request.logging.trustStorePath` | Path to the TrustStore file for HTTPS transport. Only applies when `transportType=HTTP`. | no | — |
-| `druid.request.logging.trustStorePassword` | Password for the TrustStore. Only applies when `transportType=HTTP`. | no | — |
+| `druid.request.logging.trustStorePassword` | Password for the TrustStore. Accepts a plain string or a [PasswordProvider](../../operations/password-provider.md) (e.g. an environment variable). Only applies when `transportType=HTTP`. | no | — |
 | `druid.request.logging.keyStorePath` | Path to the KeyStore file for mutual TLS. Only applies when `transportType=HTTP`. | no | — |
-| `druid.request.logging.keyStorePassword` | Password for the KeyStore. Only applies when `transportType=HTTP`. | no | — |
+| `druid.request.logging.keyStorePassword` | Password for the KeyStore. Accepts a plain string or a [PasswordProvider](../../operations/password-provider.md). Only applies when `transportType=HTTP`. | no | — |
 
 ### Examples
 
@@ -86,7 +86,7 @@ Each emitted event follows the [OpenLineage spec](https://openlineage.io/spec/2-
 | Facet | Description |
 |---|---|
 | `processing_engine` | Engine name (`druid`). Standard OpenLineage facet. |
-| `druid_query_context` | Query metadata: `identity` (authenticated user), `remoteAddress`, `queryType`, and `nativeQueryIds` (for SQL queries). |
+| `druid_query_context` | Query metadata: `identity` (authenticated user), `remoteAddress`, `queryType`, and `sqlQueryId` (on native sub-queries of SQL, for correlation with the parent SQL event). |
 | `druid_query_statistics` | Execution stats: `durationMs`, `bytes`, `planningTimeMs`, `statusCode`. |
 | `errorMessage` | Exception message for failed queries. Standard OpenLineage facet. |
 
--- a/docs/development/extensions-contrib/openlineage-emitter.md
+++ b/docs/development/extensions-contrib/openlineage-emitter.md
@@ -45,6 +45,12 @@ All configuration parameters are under `druid.request.logging`.
 | `druid.request.logging.transportType` | Where to send events. `CONSOLE` logs JSON to the Druid log; `HTTP` POSTs to an OpenLineage API endpoint. | no | `CONSOLE` |
 | `druid.request.logging.transportUrl` | OpenLineage API endpoint URL. Required when `transportType=HTTP`. | no | — |
 | `druid.request.logging.excludedNativeQueryTypes` | Native query types to exclude from lineage emission. Internal broker queries like segment metadata lookups produce noisy, low-value events. | no | `["segmentMetadata", "dataSourceMetadata", "timeBoundary"]` |
+| `druid.request.logging.emitQueueCapacity` | Maximum number of events buffered in the async HTTP emit queue. Events are dropped (with a warning) when the queue is full. Only applies when `transportType=HTTP`. | no | `1000` |
+| `druid.request.logging.emitThreadCount` | Number of background threads used to POST events to the HTTP endpoint. Only applies when `transportType=HTTP`. | no | `1` |
+| `druid.request.logging.trustStorePath` | Path to the TrustStore file for HTTPS transport. Only applies when `transportType=HTTP`. | no | — |
+| `druid.request.logging.trustStorePassword` | Password for the TrustStore. Only applies when `transportType=HTTP`. | no | — |
+| `druid.request.logging.keyStorePath` | Path to the KeyStore file for mutual TLS. Only applies when `transportType=HTTP`. | no | — |
+| `druid.request.logging.keyStorePassword` | Password for the KeyStore. Only applies when `transportType=HTTP`. | no | — |
 
 ### Examples
 
--- a/docs/development/extensions-contrib/openlineage-emitter.md
+++ b/docs/development/extensions-contrib/openlineage-emitter.md
@@ -31,7 +31,7 @@ This extension emits [OpenLineage](https://openlineage.io) `RunEvent`s for each 
 For MSQ DML statements (`INSERT INTO` / `REPLACE INTO`), the output datasource is extracted from the SQL text and emitted as an output dataset. Input extraction from SQL is not performed — reliably resolving `FROM` / `JOIN` tables at the logger layer would duplicate planner work. For native queries, input table names are resolved from the datasource tree and emitted as input datasets. Native sub-queries spawned by a SQL execution carry a `sqlQueryId` in their context facet for correlation with the parent SQL event.
 
 :::note
-SQL table extraction relies on `calcite-core` being on the classpath, which is the case on Broker nodes. Native query lineage is available on all nodes.
+MSQ DML output extraction relies on `calcite-core` being on the classpath, which is the case on Broker nodes. Native query lineage is available on all nodes.
 :::
 
 ## Configuration
--- a/extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
+++ b/extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
@@ -64,8 +64,10 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * OpenLineage RunEvents for completed Druid queries.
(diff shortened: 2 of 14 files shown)


docs/development/extensions-contrib/openlineage-emitter.md
```diff
@@ -0,0 +1,95 @@
...
+druid.request.logging.namespace=druid://broker.prod:8082
+druid.request.logging.transportType=HTTP
+druid.request.logging.transportUrl=http://marquez:5000/api/v1/lineage
+```
+
+**Combined with another logger using the `composing` provider**
+
+```properties
+druid.request.logging.type=composing
+druid.request.logging.loggerProviders[0].type=slf4j
```
maytasm: Are these correct? Don't you have to set druid.request.logging.loggerProviders to a List with a map for each Logger?

docs/development/extensions-contrib/openlineage-emitter.md
```diff
@@ -0,0 +1,95 @@
...
+:::
+
+## Configuration
+
+All configuration parameters are under `druid.request.logging`.
+
+| Property | Description | Required | Default |
+|---|---|---|---|
+| `druid.request.logging.type` | Set to `openlineage` to enable this extension. | yes | — |
+| `druid.request.logging.namespace` | Namespace used for OpenLineage job and dataset URIs. Typically the Broker URL. | no | `druid://localhost` |
```
maytasm: Can this default to the hostname?

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,498 @@
...
+      .withQuoting(Quoting.DOUBLE_QUOTE);
+
+  private static final int EMIT_QUEUE_CAPACITY = 1000;
+  private static final int EMIT_THREAD_COUNT = 1;
+
+  /**
+   * Internal broker query types that are automatically issued for schema discovery and metadata
+   * caching. These are not user-initiated and produce noisy, low-value lineage events.
+   */
+  private static final Set<String> INTERNAL_QUERY_TYPES = Set.of(
```
maytasm: Maybe make this configurable?

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,498 @@
...
+      OpenLineageRequestLoggerProvider.TransportType transportType,
+      @Nullable String transportUrl
+  )
+  {
+    this.jsonMapper = jsonMapper;
+    this.namespace = namespace;
+    this.transportType = transportType;
+    this.transportUrl = transportUrl;
+    if (transportType == OpenLineageRequestLoggerProvider.TransportType.HTTP) {
+      this.httpClient = HttpClientBuilder.create().build();
```
maytasm: We may want to be able to override this client in case we require some specific http client config (keystore, truststore, etc)

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,498 @@
...
+
+    ObjectNode facets = jsonMapper.createObjectNode();
+
+    ObjectNode jobTypeFacet = createFacet(JOB_TYPE_FACET_SCHEMA_URL);
+    jobTypeFacet.put("processingType", "BATCH");
+    jobTypeFacet.put("integration", "DRUID");
+    jobTypeFacet.put("jobType", "QUERY");
+    facets.set("jobType", jobTypeFacet);
+
+    if (sql != null) {
```
maytasm, line 408: Does it make sense to include native query if this is not sql?

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,498 @@
...
+  }
+
+  private ArrayNode buildDatasets(List<String> tableNames)
+  {
+    ArrayNode array = jsonMapper.createArrayNode();
+    for (String name : tableNames) {
+      ObjectNode node = jsonMapper.createObjectNode();
+      node.put("namespace", namespace);
+      node.put("name", name);
+      node.set("facets", jsonMapper.createObjectNode());
```
maytasm, line 444: Can we create a Github issue to track work to include the fields?

maytasm (changes requested):

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
mshahid6: do you mean if we should only emit lineage if sql?

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
mshahid6: https://github.com/apache/druid/issues/19314

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLoggerProvider.java
```diff
@@ -0,0 +1,81 @@
...
+  @Nullable
+  @JsonProperty
+  private String transportUrl;
+
+  @JsonProperty
+  @NotNull
+  private Set<String> excludedNativeQueryTypes = Set.of(
+      "segmentMetadata",
+      "dataSourceMetadata",
+      "timeBoundary"
```
jtuglu1, line 82: Why excluding timeBoundary?

docs/development/extensions-contrib/openlineage-emitter.md
```diff
@@ -0,0 +1,92 @@
...
+| `processing_engine` | Engine name (`druid`). Standard OpenLineage facet. |
+| `druid_query_context` | Query metadata: `identity` (authenticated user), `remoteAddress`, `queryType`, and `nativeQueryIds` (for SQL queries). |
+| `druid_query_statistics` | Execution stats: `durationMs`, `bytes`, `planningTimeMs`, `statusCode`. |
+| `errorMessage` | Exception message for failed queries. Standard OpenLineage facet. |
+
+### Job facets
+
+| Facet | Description |
+|---|---|
+| `jobType` | `processingType=BATCH`, `integration=DRUID`, `jobType=QUERY`. Standard OpenLineage facet. |
```
jtuglu1, line 97: processingType should be QUERY?

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,503 @@
...
+  // Standard Calcite parser (not Druid's custom parser): sufficient for table name extraction.
+  // Druid-specific syntax (REPLACE, EXTERN, etc.) falls through to the SqlParseException handler.
+  private static final SqlParser.Config SQL_PARSER_CONFIG = SqlParser
+      .config()
+      .withCaseSensitive(true)
+      .withUnquotedCasing(Casing.UNCHANGED)
+      .withQuotedCasing(Casing.UNCHANGED)
+      .withQuoting(Quoting.DOUBLE_QUOTE);
+
+  private static final int EMIT_QUEUE_CAPACITY = 1000;
```
jtuglu1: we should probably make these configurable

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,503 @@
...
+  )
+  {
+    this.jsonMapper = jsonMapper;
+    this.namespace = namespace;
+    this.transportType = transportType;
+    this.transportUrl = transportUrl;
+    this.excludedNativeQueryTypes = excludedNativeQueryTypes;
+    if (transportType == OpenLineageRequestLoggerProvider.TransportType.HTTP) {
+      this.httpClient = httpClient != null ? httpClient : HttpClientBuilder.create().build();
+      // Bounded queue: if the queue is full, silently drop the event rather than blocking
```
jtuglu1: let's not silently drop events without a warning – is there a callback, etc. we can use with an atomic counter or similar? Perhaps overload ThreadPoolExecutor.DiscardPolicy() to increment an atomic, and then we can periodically poll that on emit() or similar.

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,503 @@
...
+  }
+
+  protected void emit(ObjectNode event)
+  {
+    try {
+      String json = jsonMapper.writeValueAsString(event);
+      if (transportType == OpenLineageRequestLoggerProvider.TransportType.HTTP) {
+        emitExecutor.submit(() -> emitHttp(json));
+      } else {
+        log.info("OpenLineage event: %s", json);
```
jtuglu1: nit:debug

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,503 @@
...
+    }
+  }
+
+  private String extractSqlQueryId(RequestLogLine requestLogLine)
+  {
+    Object id = requestLogLine.getSqlQueryContext().get(BaseQuery.SQL_QUERY_ID);
+    if (id != null) {
+      return id.toString();
+    }
+    return UUID.randomUUID().toString();
```
jtuglu1: Unless this uniqueness is absolutely required by the spec I feel like we should make this some default fixed value so we can know if there are queries somehow making it through without an ID.

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
jtuglu1: We should be emitting lineage for both types. `logNativeQuery` handles the native path.

One thing to call out is to check whether issuing an MSQ/Dart query triggers a lineage emission (I suspect it will, but just to be safe).

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLoggerProvider.java
mshahid6: wasn't sure if it'd be useful for lineage and might be noisy. it's configurable but can remove it from defaults if you think it should be tracked

docs/development/extensions-contrib/openlineage-emitter.md
mshahid6: https://openlineage.io/docs/spec/facets/job-facets/job-type `processingType` which can be `STREAMING` or `BATCH`

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
mshahid6: msq-dart query emits a lineage event

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,551 @@
...
+    if (from instanceof SqlIdentifier) {
+      String name = String.join(".", ((SqlIdentifier) from).names);
+      if (!excludeNames.contains(name)) {
+        tables.add(name);
+      }
+    } else if (from instanceof SqlJoin) {
+      SqlJoin join = (SqlJoin) from;
+      collectFromClause(join.getLeft(), tables, excludeNames);
+      collectFromClause(join.getRight(), tables, excludeNames);
+    } else if (from instanceof SqlSelect) {
```
FrankChen021: P2 SQL lineage misses common nested inputs

The SQL walker only recurses into SqlSelect.getFrom() and only handles SqlBasicCall when it is AS, so inputs referenced by subqueries in WHERE/HAVING/SELECT clauses are omitted, and set operations such as SELECT ... FROM a UNION ALL SELECT ... FROM b can emit no inputs at all. This makes the emitted OpenLineage event materially incomplete for valid Druid SQL; consider walking the full SqlNode tree with the relevant CTE alias exclusions instead of only FROM/JOIN nodes.

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,551 @@
...
+      log.error(e, "Failed to serialize OpenLineage event");
+    }
+  }
+
+  private void emitHttp(String json)
+  {
+    HttpPost post = new HttpPost(transportUrl);
+    post.setEntity(new StringEntity(json, ContentType.APPLICATION_JSON));
+    try {
+      org.apache.http.HttpResponse response = httpClient.execute(post);
```
FrankChen021: P2 HTTP transport silently drops rejected events

emitHttp treats any HTTP response as success: it executes the POST and consumes the entity without checking the status code. If Marquez/OpenLineage returns 400/401/429/500 due to schema, auth, rate limiting, or backend errors, the async logger discards the event with no warning, causing silent lineage loss. Check for non-2xx responses and log enough detail to diagnose delivery failures.

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLoggerProvider.java
```diff
@@ -0,0 +1,94 @@
...
+  @NotNull
+  private TransportType transportType = TransportType.CONSOLE;
+
+  @Nullable
+  @JsonProperty
+  private String transportUrl;
+
+  @JsonProperty
+  @NotNull
+  private Set<String> excludedNativeQueryTypes = Set.of(
```
FrankChen021, line 79: P3 Default exclusions differ from docs

The provider default excludes only segmentMetadata and dataSourceMetadata, but the new docs and test constant list timeBoundary as a default exclusion too. Users who rely on the documented default will still emit low-value timeBoundary native query events unless they override the property explicitly. Add timeBoundary here or update the docs/tests to match the actual default.

mshahid6: discussion for reference (https://github.com/apache/druid/pull/19107#discussion_r3091137636), can re-add timeBoundary

FrankChen021: Thanks, yes, re-adding `timeBoundary` to the provider default would resolve the mismatch with the docs/tests and keep the documented default behavior accurate.

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,654 @@
...
+ */
+
+package org.apache.druid.extensions.openlineage;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.apache.calcite.avatica.util.Casing;
+import org.apache.calcite.avatica.util.Quoting;
+import org.apache.calcite.sql.SqlBasicCall;
```
maytasm: Would these Calcite imports cause issue when loading this extension on nodes without Calcite on the classpath?

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,654 @@
...
+            try {
+              SqlNode selectNode = SqlParser.create(selectMatcher.group(1), SQL_PARSER_CONFIG).parseQuery();
+              inputs = extractInputs(selectNode);
+            }
+            catch (SqlParseException ignored) {
+              // EXTERN or other non-standard sources — inputs left empty
+            }
+          }
+        }
+        Matcher m = REPLACE_INTO_PATTERN.matcher(sql);
```
maytasm: This matcher will run even when we already have a match from INSERT_INTO_PATTERN. This is unnecessary since INSERT INTO and REPLACE INTO are mutually exclusive patterns

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,654 @@
...
+    this.transportType = transportType;
+    this.transportUrl = transportUrl;
+    this.excludedNativeQueryTypes = excludedNativeQueryTypes;
+    if (transportType == OpenLineageRequestLoggerProvider.TransportType.HTTP && transportUrl == null) {
+      throw new IllegalStateException(
+          "druid.request.logging.transportUrl must be set when transportType=HTTP"
+      );
+    }
+    if (transportType == OpenLineageRequestLoggerProvider.TransportType.HTTP) {
+      this.httpClient = httpClient != null ? httpClient : HttpClientBuilder.create().build();
```
maytasm, line 148: Generally best practice to be able to set and have reasonable defaults for socket, connection, Connection Request timeouts

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLoggerProvider.java
```diff
@@ -0,0 +1,95 @@
...
+      "segmentMetadata",
+      "dataSourceMetadata",
+      "timeBoundary"
+  );
+
+  @JsonProperty
+  private int emitQueueCapacity = OpenLineageRequestLogger.DEFAULT_EMIT_QUEUE_CAPACITY;
```
maytasm, line 86: This seems to be configurable but don't see it in the docs

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLoggerProvider.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @JsonProperty
+  private int emitThreadCount = OpenLineageRequestLogger.DEFAULT_EMIT_THREAD_COUNT;
```
maytasm, line 89: This seems to be configurable but don't see it in the docs

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,654 @@
...
+    facets.set("processing_engine", engineFacet);
+
+    ObjectNode contextFacet = createFacet(null);
+    contextFacet.put("queryType", queryType);
+    contextFacet.put("remoteAddress", requestLogLine.getRemoteAddr());
+    Object identity = stats.get("identity");
+    if (identity != null) {
+      contextFacet.put("identity", identity.toString());
+    }
+    Object nativeQueryIds = requestLogLine.getSqlQueryContext().get("nativeQueryIds");
```
maytasm: getSqlQueryContext is Nullable

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,654 @@
...
+      log.error(e, "Failed to POST OpenLineage event to [%s]", transportUrl);
+    }
+    finally {
+      post.releaseConnection();
+    }
+  }
+
+  private String extractSqlQueryId(RequestLogLine requestLogLine)
+  {
+    Object id = requestLogLine.getSqlQueryContext().get(BaseQuery.SQL_QUERY_ID);
```
maytasm: getSqlQueryContext is Nullable

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
maytasm, line 148: Might also want to make properties like trustStorePath, trustStorePassword, keyStorePath, keyStorePassword configurable. Something similar to InfluxdbEmitterConfig

maytasm (changes requested):

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,654 @@
...
+      .withQuoting(Quoting.DOUBLE_QUOTE);
+
+  // Matches: INSERT INTO <table> ... (fallback when EXTERN prevents standard Calcite parsing)
+  // Handles both quoted ("table") and unquoted (table) identifiers.
+  private static final Pattern INSERT_INTO_PATTERN =
+      Pattern.compile("(?i)^\\s*INSERT\\s+INTO\\s+\"?([^\"\\s]+)\"?");
+  // Matches: REPLACE INTO <table> OVERWRITE ... (Druid-specific, not parseable by standard Calcite)
+  private static final Pattern REPLACE_INTO_PATTERN =
+      Pattern.compile("(?i)^\\s*REPLACE\\s+INTO\\s+\"?([^\"\\s]+)\"?");
+  // Extracts the SELECT subquery from a REPLACE INTO statement for input lineage
```
maytasm: nit: the commend and variable name here indicates the pattern only applies to REPLACE INTO, however, this applies to both INSERT INTO and REPLACE INTO

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,654 @@
...
+      SqlJoin join = (SqlJoin) node;
+      collectFromPosition(join.getLeft(), tables, excludeNames);
+      collectFromPosition(join.getRight(), tables, excludeNames);
+    } else if (node instanceof SqlSelect) {
+      // Subquery in FROM: recurse into the full select.
+      collectTableNames(node, tables, excludeNames);
+    } else if (node instanceof SqlBasicCall && node.getKind() == SqlKind.AS) {
+      // Alias: the table reference is the first operand.
+      collectFromPosition(((SqlBasicCall) node).operand(0), tables, excludeNames);
+    } else if (node instanceof SqlCall) {
```
FrankChen021: P2 UNNEST column operands are reported as input datasets

The generic SqlCall branch in collectFromPosition recurses into every operand and then treats any SqlIdentifier as a table reference. For Druid SQL such as SELECT a FROM druid.arrays, UNNEST(arrayString) AS u(a), the UNNEST operand arrayString is a column, not a datasource, but this walker will add it to the OpenLineage inputs alongside druid.arrays. This produces incorrect lineage for supported UNNEST queries; table-function or UNNEST calls should avoid collecting column operands as datasets, or only recurse into operands that are known relation-producing subqueries.

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,626 @@
...
+    this.transportType = transportType;
+    this.transportUrl = transportUrl;
+    this.excludedNativeQueryTypes = excludedNativeQueryTypes;
+    if (transportType == OpenLineageRequestLoggerProvider.TransportType.HTTP && transportUrl == null) {
+      throw new IllegalStateException(
+          "druid.request.logging.transportUrl must be set when transportType=HTTP"
+      );
+    }
+    if (transportType == OpenLineageRequestLoggerProvider.TransportType.HTTP) {
+      this.httpClient = httpClient;
```
FrankChen021: [P2] Direct HTTP logger construction leaves client null

The public 5-argument constructor still delegates here with a null httpClient. After this update, HTTP mode stores that null instead of creating a default client, so direct construction with HTTP and a transportUrl enqueues events whose worker hits a NullPointerException at httpClient.execute and never posts lineage. Please keep default client creation when httpClient is null, or make the client-injection path non-public and route all HTTP construction through the provider.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,425 @@
...
+   *
+   * <p>For SQL queries, the broker converts SQL to native queries before execution.
+   * Each native sub-query calls {@link #logNativeQuery}, which extracts inputs from
+   * {@code query.getDataSource().getTableNames()}. The native events carry the parent
+   * SQL query ID in their context for correlation.
+   */
+  @Override
+  public void logSqlQuery(RequestLogLine requestLogLine) throws IOException
+  {
+    // No-op: lineage is emitted from native sub-queries.
```
FrankChen021: [P1] SQL no-op drops MSQ lineage events

Making `logSqlQuery` a no-op only preserves SQL lineage for native-engine SQL queries that pass through `QueryLifecycle.logNativeQuery`. MSQ-backed SQL requests, including INSERT/REPLACE, go through `MSQTaskQueryMaker` to submit an `MSQControllerTask`; the request-log callback for the SQL request is emitted by `SqlExecutionReporter` via `logSqlQuery`, with no corresponding native request-log event on this path. With this method empty, those SQL executions emit no OpenLineage event at all, losing the ingestion/output lineage this extension is meant to report. Please either keep SQL-level emission for non-native SQL engines or add an MSQ path that emits equivalent lineage before making `logSqlQuery` a no-op.

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
```diff
@@ -0,0 +1,449 @@
...
+      "https://openlineage.io/spec/facets/2-0-2/JobTypeJobFacet.json";
+  static final int DEFAULT_EMIT_QUEUE_CAPACITY = 1000;
+  static final int DEFAULT_EMIT_THREAD_COUNT = 1;
+  private static final int DISCARD_WARNING_INTERVAL = 1000;
+
+  static final String UNKNOWN_QUERY_ID = "unknown-query-id";
+
+  // Matches the output table of MSQ DML: INSERT INTO "foo" or REPLACE INTO foo
+  private static final Pattern MSQ_DML_OUTPUT_PATTERN = Pattern.compile(
+      "(?si)^\\s*(?:INSERT|REPLACE)\\s+INTO\\s+\"?([^\"\\s(,]+)\"?"
```
FrankChen021: [P2] MSQ output regex reports the wrong destination for valid INSERT targets

The new SQL DML path extracts the output dataset from raw SQL with this regex, but Druid accepts targets that do not map 1:1 to that text. For example, `INSERT INTO druid.dst ...` is valid and is normalized by the ingest planner to datasource `dst`, while this emits output dataset `druid.dst`; `INSERT INTO EXTERN(...) AS CSV ...` is also valid and this emits a bogus Druid dataset named `EXTERN`. This produces incorrect OpenLineage outputs for supported MSQ statements. Use the parsed ingest target or the planner's resolved destination instead of matching the SQL string.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

extensions-contrib/openlineage-emitter/src/main/java/org/apache/druid/extensions/openlineage/OpenLineageRequestLogger.java
mshahid6: updated

FrankChen021 (commented):
Follow-up handled: the MSQ output extraction now uses `DruidSqlParser`, normalizes schema/catalog-prefixed datasource targets, skips `EXTERN` exports, and includes regression coverage for the cases raised.

---

FrankChen021 (commented):

---

jtuglu1 (approved):
LGTM once CI is green

PR #19109, opened 2026-03-07
remove support for hadoop ingestion

clintropolis:
### Description
This PR drops support for the Apache Hadoop based ingestion tasks, which run Druid ingestion processes using Hadoop YARN. It was officially deprecated in Druid 34 in #18286 where it was scheduled for deletion in Druid 37, based on discussion in this dev list thread https://lists.apache.org/thread/5jyl10tm4glvzlfps74o3q4r3rb0h1x9 and elsewhere over the last several years. Note that Druid will still work with a Hadoop ecosystem, the `druid-hdfs-storage` extension is not part of this removal, so we still support readiing from and writing to HDFS, we still support reading Hadoopy formats like `druid-orc-extensions` and others, and the `druid-kerberos` and `druid-ranger-security` extensions still function.

Hadoop ingestion served us well, we wouldn't be where we are without it, but I think it is time to say goodbye 🫡.

There are a number of reasons for doing this - the biggest pain point over the years being Java version support and dependencies. But it is also because Hadoop ingestion is not even close to as well maintained or tested as the rest of Druid, and no one puts much effort into making new features work with it. The primary reason is that codepaths between Hadoop ingestion and the 'native' Druid ingestion types diverged several years ago with the switch from `InputRowParser` to `InputSource`/`InputFormat` starting in #8823.

As for the intended replacements, the Kubernetes based task runner of `kubernetes-overlord-extensions` is a superior alternative for having task auto scaling capabilities and has matured since becoming a core extension. Additionally we believe that SQL based ingestion with the multi-stage query engine is the future of batch ingestion, and so is the focus of most new development.

I am aware that Hadoop 3.5 is currently nearing release, which would solve our Java 17+ problems that are the next most immediate problems we are facing, but still think this is the right move to make at this time. That said, if anyone in the community wishes to collect the pieces we are deleting here and migrate them into a 'contrib' extension, then I think we would be happy to accept adding it back in that form, which as a contrib extension it would much better reflect the state it has actually been in for the last several years.

#### Legacy integration tests
Additionally, this also removes the legacy integration-tests, since the hadoop ingestion tests are now the sole remaining legacy ITs!

#### Release note
Support for Apache Hadoop-based ingestion was removed from Apache Druid 37.0.0. Please use
[SQL-based ingestion](https://druid.apache.org/docs/latest/multi-stage-query/) or [native batch](https://druid.apache.org/docs/latest/ingestion/native-batch) instead.

[...]

PR #19111, opened 2026-03-09
add thrift inputformat implementation

clintropolis:
### Description
Adds a thrift `InputFormat` implementation, which I believe is the last data format extension which has not implemented `InputFormat`. This PR coupled with #19109, will allow us to remove all operator facing `InputRowParser`/`ParseSpec` implementations.

I had claude do most of the work here since its perfect for this kind of thing, with a few manual adjustments at the end.

### DIFF
--- a/extensions-contrib/thrift-extensions/src/main/java/org/apache/druid/data/input/thrift/ThriftReader.java
+++ b/extensions-contrib/thrift-extensions/src/main/java/org/apache/druid/data/input/thrift/ThriftReader.java
@@ -124,7 +124,7 @@ public class ThriftReader extends IntermediateRowParsingReader<byte[]>
     }
   }
 
-  @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


extensions-contrib/thrift-extensions/src/main/java/org/apache/druid/data/input/thrift/ThriftInputFormat.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+public class ThriftInputFormat extends NestedInputFormat
+{
+  private final String jarPath;
+  private final String thriftClassName;
+
+  @JsonCreator
+  public ThriftInputFormat(
+      @JsonProperty("flattenSpec") @Nullable JSONPathSpec flattenSpec,
+      @JsonProperty("thriftJar") @Nullable String jarPath,
```
kfaraz: Nit: Best to align the field/argument name with the json property name.

PR #19134, opened 2026-03-11
gha should not ignore flaky failure

cecemei:
gha should not ignore flaky failure, see [example run](https://github.com/apache/druid/actions/runs/22970395735/job/66685238114) converts flaky failures to actual failures and fails the run. For parameterized tests, <flakyFailure> is not reliable. It can hide real failures because the reporter treats intermittent or partial failures as “flaky” for the entire method.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/.github/workflows/worker.yml
+++ b/.github/workflows/worker.yml
@@ -89,6 +89,27 @@ jobs:
       - name: 'Execute: ${{ inputs.script }}'
         run: ${{ inputs.script }}
         timeout-minutes: 60
+      - name: Summarize flaky tests
+        if: always()
+        run: |
+          # Find all flaky failures
+          flaky_count=$(find . -name "TEST-*.xml" -exec grep -c "<flakyFailure" {} \; | awk '{sum+=$1} END {print sum}')
+          
+          # Write count to GitHub Actions summary
+          echo "### Flaky Tests: $flaky_count found" >> $GITHUB_STEP_SUMMARY
+          
+          # If there are any, list them as bullets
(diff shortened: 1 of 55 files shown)


kgyrtkirk:
this looks odd to me....I think it would be better to set `surefire.rerunFailingTestsCount` to `0`  and hammer out the issues....

.github/workflows/worker.yml
```diff
@@ -90,6 +90,25 @@ jobs:
...
+          
+          if [ "$flaky_count" -gt 0 ]; then
+            echo "#### Flaky Test Details"
+            grep -R "<flakyFailure" . | sed -E 's#.*/TEST-(.*)\.xml.*message="([^"]*)".*#- \1: \2#'
+          fi
+
+      - name: Normalize flaky failures
+        if: ${{ hashFiles('**/TEST-*.xml') != '' }}
+        run: |
+          find . -name "TEST-*.xml" -exec sed -i 's/flakyFailure/failure/g' {} +
```
FrankChen021, line 110: P1 Flaky errors are still ignored by CI

The normalization only rewrites flakyFailure elements, but Maven Surefire also emits flakyError for tests that error on an earlier attempt and pass on rerun. Those XML entries will remain flaky-only, so mikepenz/action-junit-report can still treat them as retries/flakes instead of failing the job. This leaves a common class of intermittent test failures hidden. The same handling should include flakyError -> error, and the summary/count should include both tag types.

PR #19138, opened 2026-03-11
perf: Interval tree for managing segment metadata in memory

pirvtech:
Segment metadata stored in memory of the Historicals, is used when looking up segments that match an interval for query and segment loading purposes. Currently this is a serial scan that goes through all segments metadata in ascending start time order to find the right matching segments. 

This changes introduces an Interval Tree as a more efficient way to store segment metadata in memory, to speed up searches for segments, that can potentially cut down search times from O(n) to O(logn).

Core changes to file processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
Interval tree implementation in file processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
Documentation comments have been included in important files and sections of code. Unit tests added.

This has been reviewed internally and is in a production Druid cluster.


jtuglu1:
Hi – thanks. Do you have benchmarks for this?

pirvtech:
> Hi – thanks. Do you have benchmarks for this?

I will add the benchmarks.

processing/src/main/java/org/apache/druid/java/util/common/guava/Comparators.java
```diff
@@ -119,6 +119,34 @@ public int compare(Interval lhs, Interval rhs)
...
+    public int compare(Interval lhs, Interval rhs)
+    {
+      if (lhs.getChronology().equals(rhs.getChronology())) {
+        return Long.compare(lhs.getStartMillis(), rhs.getStartMillis());
+      }
+      return dateTimeComp.compare(lhs.getStart(), rhs.getStart());
+    }
+  };
+
+  private static final Comparator<Interval> INTERVAL_BY_END = new Comparator<>()
```
kfaraz: Why do we need these if we already have `Comparators.INTERVAL_BY_START_THEN_END` and `Comparators.INTERVAL_BY_END_THEN_START`?

processing/src/main/java/org/apache/druid/java/util/common/guava/Comparators.java
```diff
@@ -119,6 +119,34 @@ public int compare(Interval lhs, Interval rhs)
     }
   };
 
+  private static final Comparator<Interval> INTERVAL_BY_START = new Comparator<>()
```
kfaraz, line 122: Can't we just reuse the existing comparators `INTERVAL_BY_START_THEN_END` and `INTERVAL_BY_END_THEN_START`?

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
```diff
@@ -74,28 +75,77 @@
...
+    public boolean isEnabled(Capability capability)
+    {
+      return capabilities.contains(capability);
+    }
+  }
+
+  private static IntervalTreeMatchMode intervalTreeMatchMode = IntervalTreeMatchMode.NONE;
+
+  static {
+    String mode = System.getProperty("experimental.timeline.intervalTreeMatchMode");
```
kfaraz: - Please don't use system properties here directly. Instead pass a flag into the constructor of `VersionedIntervalTimeline` indicating whether the improved datastructures should be used or not.
- Property name should be more like `druid.segment.timeline.fastIntervalSearch` with possible values `true` and `false`. (I don't think we need 3 modes).
- The class creating the timeline should pass in the correct value for the flag.

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
```diff
@@ -74,28 +75,77 @@
...
+      try {
+        intervalTreeMatchMode = IntervalTreeMatchMode.valueOf(mode);
+      }
+      catch (IllegalArgumentException e) {
+        logger.warn(e, "Unrecognized interval tree match mode specified [%s]", mode);
+      }
+    }
+  }
+
+  {
```
kfaraz, line 119: Please move this field initialization inside the constructor.

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
```diff
@@ -74,28 +75,77 @@
...
   // Set this to true if the client needs to skip tombstones upon lookup (like the broker)
   private final boolean skipObjectsWithNoData;
 
+  // 0 Legacy functionality
+  // 1 Use IntervalTree only for basic segment timeline
+  // 2 Use IntervalTree for querying segments as well
+  private enum IntervalTreeMatchMode
+  {
+    NONE,
+    ENTRIES_ONLY(Capability.ENTRIES),
```
kfaraz: Is there a specific case when we would want to support improved search on entries only and not queries?
I think just 2 modes should suffice: ALL or NONE

kfaraz (commented):
Thanks for putting this together, @pirvtech!
I have often felt the need for such a datastructure too, even outside the VIT (`VersionedIntervalTimeline`).

__Thoughts on perf__
- I doubt if a single query is really going to benefit from this change since doing a `contains()` or an `overlaps()` check on say 25k intervals (which is a fairly large number of intervals for a typical Druid cluster) would not be very compute intensive.
  - You can think of 25k intervals as roughly 3 years worth of HOUR-granularity data.
- But in high concurrency, this would still be beneficial since the VIT does all of its computations inside a giant lock. The shorter we hold the lock, the better.
- Either way, we should add benchmarks as @jtuglu1 suggested for queries as well as VIT itself.

__Notes on the implementation__

I have left some inline comments.

Along with that, the approach would be much cleaner if you do something like this instead:
- Add an `IntervalNavigableMap<T>` interface which extends `NavigableMap<Interval, T>`. This interface should have the following new methods:
  - `entriesContaining(Interval interval)`
  - `entriesOverlapping(Interval interval)`
  - `entriesMatching(Interval interval)` (Is this really needed?)
  - Alternatively, instead of methods that return a sub map, you could have methods that return a matching entry set.
- Instead of `HashMap` and `NavigableMap`, the VIT class should use this new interface.
- Add a class `IntervalTreeMap<T> implements IntervalNavigableMap<T> extends TreeMap<Interval, T>` and provides default implementations for the new methods.
   - For example, for the `entriesContaining()` method, we return the whole map.
- Add a new class `FastSearchIntervalMap<T>` which performs the optimised search.
- Based on value of the `fastSearch` flag passed to constructor of VIT, choose the implementation of the map.
- This would ensure that there are minimal changes to `VIT` and we can easily swap out different map implementations.

pirvtech:
@kfaraz thanks for the comments. We were facing a scenario where the number of segments were in the order of millions, due to the amount of parallelization we have in ingestion tasks, the time spread of data being ingested, and relative time offset when a time interval gets compacted. This change helped our historicals be able to load and serve segments faster. I had made another change to make the scanning and loading of local on-disk segments during historical startup multi-threaded from a single threa that it was, but looks like someone beat us to the punch in submission :). I will review your code feedback and provide my responses.

jtuglu1:
@kfaraz I think this PR will have much more impact than just Historical. See https://github.com/apache/druid/pull/19278. Broker timeline operations become terribly slow under high # of intervals. We can use the interval tree here to speed up those operations drastically and speed up queries under heavy segment load/remove callback load.

jtuglu1:
@pirvtech – are you still working on this patch? We'd love to get this in and experimenting on Broker-side to fix some issues we've run into with large #s of intervals in the VIT. Happy to pick this up otherwise.

kfaraz:
Thanks for the clarification, @jtuglu1 !
Could you share an estimate of the typical number of intervals that causes the slowness? Were you able to capture flamegraphs?

Regardless, I agree that this feature would be generally useful. We just need to clean up the changes a bit.

jtuglu1:
> Thanks for the clarification, @jtuglu1 ! Could you share an estimate of the typical number of intervals that causes the slowness? Were you able to capture flamegraphs?
> 
> Regardless, I agree that this feature would be generally useful. We just need to clean up the changes a bit.

I can get a flamegraph, but see the description of https://github.com/apache/druid/pull/19278 for explanation. Basically for a full year of MINUTE granularity you get ~500k intervals.

pirvtech:
> @pirvtech – are you still working on this patch? We'd love to get this in and experimenting on Broker-side to fix some issues we've run into with large #s of intervals in the VIT. Happy to pick this up otherwise.

@jtuglu1 yes was planning to send out updates to review comments this week. Let me know if you have any feedback as well.

jtuglu1:
> > @pirvtech – are you still working on this patch? We'd love to get this in and experimenting on Broker-side to fix some issues we've run into with large #s of intervals in the VIT. Happy to pick this up otherwise.
> 
> @jtuglu1 yes was planning to send out updates to review comments this week. Let me know if you have any feedback as well.

Great, thanks! Feel free to use my benchmarks in https://github.com/apache/druid/pull/19278. I think it'd be great to extend on those to summarize/benchmark the common operations with a variable size of intervals to ensure we're seeing the expected throughput/latency.

processing/src/main/java/org/apache/druid/java/util/common/guava/Comparators.java
pirvtech: In IntervalTree we need a separate start and end comparators as we need to compare start and end times of intervals independently in different situations. Unfortunately INTERVAL_BY_START_THEN_END for example, compares end time when start times match and that leads to wrong results. I did initially try to use the existing comparators.

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
pirvtech: Agree. The mode was more for easing in this feature into the production deployment with ability to fall back in case things didn't work out as planned. A single binary feature flag maybe sufficient and coming from a configuration like you suggested. I will make the change.

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
pirvtech: Agree

jtuglu1:
@pirvtech any updates here? I'm happy to get this merged if you're busy

pirvtech:
@jtuglu1 thanks for the patience, submitting updates today.

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
pirvtech: @kfaraz please see if the modifications in the new commit are the way you see this parameter should be configured. I wanted to get a feedback on this, before making the modifications to pass in the parameter for all the classes that use the timeline (directly or through layers of indirection), including test classes and there are quite a few of them.

pirvtech:
Please see commend above on pushed config related changes @kfaraz

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
```diff
@@ -747,21 +795,38 @@ private List<TimelineObjectHolder<VersionType, ObjectType>> lookup(Interval inte
...
-      if ((!skipObjectsWithNoData || val.partitionHolder.hasData()) && timelineInterval.overlaps(interval)) {
-        retVal.add(
-            new TimelineObjectHolder<>(
-                timelineInterval,
-                val.getTrueInterval(),
-                val.getVersion(),
-                PartitionHolder.copyWithOnlyVisibleChunks(val.getPartitionHolder())
-            )
-        );
+    if (fastIntervalSearch) {
```
FrankChen021, line 806: P2 Fast timeline path is not covered by existing VIT tests

The new fastIntervalSearch branch changes lookup/findChunk/remove/overshadowing behavior, but the existing VersionedIntervalTimeline test factory still constructs only the default false path. IntervalTree unit tests do not cover VIT's adjusted interval splitting and overshadowing semantics, so this can regress query segment selection when the new config is enabled. Please parameterize the existing VIT tests to run with fastIntervalSearch=true as well.

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
```diff
@@ -0,0 +1,797 @@
...
+    return entry != null ? entry.getKey() : null;
+  }
+
+  @Override
+  public Map.Entry<Interval, T> ceilingEntry(Interval key)
+  {
+    Node<T> cnode = null;
+    Node<T> node = root;
+    while (node != null) {
+      if (node.getKey().equals(key)) {
```
FrankChen021: P3 ceilingEntry violates comparator equality

ceilingEntry checks Interval.equals instead of compareInterval == 0. The tree treats keys as equal by start/end comparator in put/get, so an Interval with the same instants but a different chronology can be stored as the same key but ceilingKey can skip it and return the next higher key or null. This breaks NavigableMap semantics for the public IntervalTree API.

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
pirvtech: Yes, was waiting for feedback on whether this configuration approach is acceptable before making changes at all the places. Sounds like it is, will propagate the changes.

FrankChen021: Thanks. Yes, that is the remaining concern: please run the existing VersionedIntervalTimeline coverage against fastIntervalSearch=true as well, since the IntervalTree tests don't exercise VIT's interval splitting/overshadowing behavior.

pirvtech: Parameterized the tests to exercise fastIntervalSearch true as well

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
pirvtech: Good point, fixed.

pirvtech:
Addressed comments from @FrankChen021 in new commit

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

---

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
```diff
@@ -0,0 +1,796 @@
...
+  {
+
+    if (node == null) {
+      return;
+    }
+
+    // Process in-order
+
+    // Search left
+    if ((node.left != null) && condition.apply(node.left.range)) {
```
FrankChen021: [P3] Generic matching can skip valid predicate matches

findMatching/forEachMatching accept an arbitrary Predicate<Interval>, but subtree traversal prunes by applying that same predicate to the subtree aggregate range. That is only valid for range-monotonic predicates such as contains/overlaps. For example, a predicate looking for one exact child interval can return false for the wider subtree range, causing traversal to skip the child entirely. Either keep the public API limited to supported range predicates or split the subtree-pruning predicate from the per-node match predicate.

pirvtech: Yes it is meant specifically to be used for predicates that can work with range which is the main optimization feature in the IntervalTree. Exact match predicate for example will not be able to work with it and should traverse the tree instead. I can change the method name to disambiguate it from sounding generic.

pirvtech: Actually added ability to specify a separate predicate for subtree pruning.

FrankChen021: Thanks, the separate pruning predicate helps callers that know both checks. The original concern still applies to the existing public `findMatching` and two-argument `forEachMatching`, though: both still pass the same arbitrary `condition` as the subtree pruning predicate, so a caller using an exact-match or otherwise non-range-monotonic predicate can still get silently incomplete results. Could we either make those APIs private/rename them to state the range-pruning contract, or have the public generic-looking form traverse without pruning?

pirvtech: I added new methods for full traversal with new naming and clear documentation for all the search methods, including the caveats. I would still prefer to keep the current naming for methods that do the pruning as it is an important performance optimization which I would like people to look at first to see if they can use it.

FrankChen021 (commented):
Reviewed 10 of 10 changed files. The follow-up adds full-traversal APIs and documents the range-pruning contract; current PR call sites use range-monotonic predicates, so no inline reply is needed.

---

pirvtech:
Merged latest master into this branch and resolved conflicts.

pirvtech:
What are the next steps?

FrankChen021 (commented):

---

pirvtech:
Any other comments @FrankChen021 @kfaraz @jtuglu1 or is it ready to merge.

pirvtech:
Any updates from reviewers if it is ready to merge?

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
```diff
@@ -0,0 +1,858 @@
...
+ * not contain a matching candidate interval.
+ * <p>
+ * To optimize the balancing cost w.r.t the operation time, the tree is not balanced on every modification operation.
+ * Rather, a configurable imbalance tolerance from the theoretical ideal height of log2(N) is allowed, breaching which
+ * triggers the rebalance.
+ * <p>
+ * Not thread safe.
+ * <p>
+ */
+public class IntervalTree<T> extends AbstractMap<Interval, T> implements NavigableMap<Interval, T>
```
kfaraz: Nit: Rename since this class is meant to be used as a navigable map inspired by Interval Trees and not a `Tree`.
```suggestion
public class IntervalTreeMap<T> extends AbstractMap<Interval, T> implements NavigableMap<Interval, T>
```

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
```diff
@@ -0,0 +1,858 @@
...
+      if (rnode == node.right) {
+        setRightNode(node, rnode.left);
+        rnode.left = null;
+      }
+      recomputeState(node);
+      return rnode;
+    }
+  }
+
+  /*
```
kfaraz: Please remove the commented code.

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
```diff
@@ -0,0 +1,858 @@
...
+  }
+
+  class EntrySet extends AbstractSet<Map.Entry<Interval, T>>
+  {
+
+    // Currently this returns a distinct collection when iterating
+    @Override
+    public Iterator<Map.Entry<Interval, T>> iterator()
+    {
+      //return inOrderTraverse();
```
kfaraz: Please remove commented out code.

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
```diff
@@ -0,0 +1,858 @@
...
+            current = current.parent;
+          } while ((current != null) && (current.right == prev));
+        }
+        return entry;
+      }
+    }
+
+  }
+
+  private void checkRebalance()
```
kfaraz, line 808: A small javadoc would help here.

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
```diff
@@ -0,0 +1,858 @@
...
+    if (node.right != right) {
+      node.right = right;
+      if (right != null) {
+        right.parent = node;
+      }
+    }
+  }
+
+  @VisibleForTesting
+  public String print()
```
kfaraz: Rename this to `toTreeString` (or use `toString` itself), add a javadoc, remove the `@VisibleForTesting` annotation.

processing/src/main/java/org/apache/druid/java/util/common/guava/Comparators.java
kfaraz: Thanks for the clarification!

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
```diff
@@ -0,0 +1,858 @@
...
+
+  @VisibleForTesting
+  Node<T> root;
+  int size;
+
+  // Deviation allowed from ideal height for the maximum height on either side of the tree, expressed as a
+  // percentage of ideal height
+  int imbalanceTolerance = 50;
+
+  EntrySet entrySet = new EntrySet();
```
kfaraz: Please make all of these private. Add getters if needed.

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
```diff
@@ -0,0 +1,858 @@
...
+
+  public IntervalTree(Comparator<Interval> startComparator, Comparator<Interval> endComparator)
```
kfaraz: Do we really need the start and end comparators in the constructor? Wouldn't they always be `INTERVAL_BY_START` and `INTERVAL_BY_END` respectively?

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
```diff
@@ -74,37 +75,51 @@
...
   // IncompletePartitionsTimeline also includes completePartitionsTimeline
   // adjusted interval -> timelineEntry
   @VisibleForTesting
-  final NavigableMap<Interval, TimelineEntry> incompletePartitionsTimeline = new TreeMap<>(
-      Comparators.intervalsByStartThenEnd()
-  );
+  final NavigableMap<Interval, TimelineEntry> incompletePartitionsTimeline;
   // true interval -> version -> timelineEntry
   private final Map<Interval, TreeMap<VersionType, TimelineEntry>> allTimelineEntries = new HashMap<>();
+  private final IntervalTree<TreeMap<VersionType, TimelineEntry>> allTimeIntervals = new IntervalTree<>(Comparators.intervalsByStart(), Comparators.intervalsByEnd());
```
kfaraz: This map should be initialized only if `fastIntervalSearch` is true.

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
```diff
@@ -74,37 +75,51 @@
...
   {
     this(versionComparator, false);
   }
 
   public VersionedIntervalTimeline(Comparator<? super VersionType> versionComparator, boolean skipObjectsWithNoData)
+  {
+    this(versionComparator, skipObjectsWithNoData, false);
+  }
+
+  public VersionedIntervalTimeline(Comparator<? super VersionType> versionComparator, boolean skipObjectsWithNoData, boolean fastIntervalSearch)
```
kfaraz, line 118: Please add a javadoc with a 1-line description for `fastIntervalSearch`.

processing/src/test/java/org/apache/druid/timeline/IntervalTreeTest.java
```diff
@@ -0,0 +1,548 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.Set;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+public class IntervalTreeTest
```
kfaraz: Please use JUnit5 for the test.

processing/src/test/java/org/apache/druid/timeline/IntervalTreeMapTest.java
```diff
@@ -0,0 +1,548 @@
...
+      }
+    }
+    System.out.println("Seq find time " + (System.currentTimeMillis() - start));
+    start = System.currentTimeMillis();
+    for (int i = 0; i < total; i++) {
+      Pair<Interval, String> pair = expectedData.get(i);
+      Interval interval = pair.lhs;
+      tree.findEncompassing(interval);
+    }
+    System.out.println("Tree find time " + (System.currentTimeMillis() - start));
```
kfaraz: Please remove `System.out` statements from the test.

server/src/main/java/org/apache/druid/segment/indexing/TimelineConfig.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.apache.druid.segment.indexing;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.annotation.Nullable;
+
+public class TimelineConfig
```
kfaraz: Instead of a new class, we should probably add the new config `fastIntervalSearch` in class `SegmentMetadataQueryConfig` which is bound as `druid.query.segmentMetadata`.

server/src/main/java/org/apache/druid/server/SegmentManager.java
```diff
@@ -489,15 +499,20 @@ public void shutdown()
...
-        new VersionedIntervalTimeline<>(Ordering.natural());
+    private final VersionedIntervalTimeline<String, DataSegment> timeline;
 
     private final ConcurrentHashMap<SegmentId, ReferenceCountedIndexedTableProvider> tablesLookup = new ConcurrentHashMap<>();
     private long totalSegmentSize;
     private long numSegments;
     private long rowCount;
     private final SegmentRowCountDistribution segmentRowCountDistribution = new SegmentRowCountDistribution();
 
+    @Inject
```
kfaraz: Annotation is not needed here since this is a static class and is never bound by Guice.

processing/src/main/java/org/apache/druid/timeline/VersionedIntervalTimeline.java
```diff
@@ -289,13 +310,40 @@ public PartitionChunk<ObjectType> findChunk(Interval interval, VersionType versi
...
+        }
+      }
+
+      // If an exact interval match is not found search for an encapsulating interval
+
+      // If tree search is enabled use it else revert to checking all intervals
+      if (fastIntervalSearch) {
+        Map<Interval, TreeMap<VersionType, TimelineEntry>> possibleMatches = allTimeIntervals.findEncompassing(interval);
+        for (Entry<Interval, TreeMap<VersionType, TimelineEntry>> entry : possibleMatches.entrySet()) {
+          Interval eninterval = entry.getKey();
```
kfaraz:
```suggestion
          Interval candidateInterval = entry.getKey();
```

kfaraz (commented):
Thanks for updating the patch and for your patience, @pirvtech !

Overall, the changes look good. I have left some minor suggestions. A couple of files remain for review, will finish those soon as well.

Please also update the PR description to include a __Release note__ section which describes the new config.

pirvtech:
> Thanks for updating the patch and for your patience, @pirvtech !
> 
> Overall, the changes look good. I have left some minor suggestions. A couple of files remain for review, will finish those soon as well.
> 
> Please also update the PR description to include a **Release note** section which describes the new config.

@kfaraz will look into your comments and address them.

server/src/main/java/org/apache/druid/segment/indexing/TimelineConfig.java
pirvtech: I considered this earlier but the interval tree is not only used during query processing but also in other cases when loading segments from disk.

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
pirvtech: I think it is better to keep it configurable where it could be used with other implementations of the start and end comparators in other parts of the project potentially in the future.

server/src/main/java/org/apache/druid/segment/indexing/TimelineConfig.java
kfaraz: Yeah, I suppose that's fair as we might want to use this on non-queryable nodes as well.
Please add a short javadoc though, and maybe rename to `SegmentTimelineConfig`.

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
kfaraz: Even in future cases, I can't think of any other comparators on start or end that would make sense while searching for an interval, since using any other comparator would not offer any real benefit.
For now, you can just hard-code the start and end comparators to the respective constants, as it makes this class easier to use.
If really needed in the future, we can add these arguments back.

kfaraz:
@pirvtech , the test failures seem genuine.
Also, there are 2 open comments.

Please try to get these resolved soon if this patch needs to be included in Druid 38.

pirvtech:
@kfaraz let me look at these

processing/src/test/java/org/apache/druid/timeline/IntervalTreeTest.java
pirvtech: Looks like other tests in the project are using JUnit4. What needs to be done to move to JUnit5 (new dependency, annotations, imports?).

pirvtech:
> @pirvtech , the test failures seem genuine. Also, there are 2 open comments.
> 
> Please try to get these resolved soon if this patch needs to be included in Druid 38.

Error seems to be about missing @Inject annotation on the TimelineConfig constructor. Not sure why guice is looking for the @Inject annotation on the single arg constructor or requiring a default constructor. The class is registered using JsonConfigProvider with the guice binder, and the JsonConfigProvider should be able to use the single arg connector as it has the @JsonCreator annotation on it. This seems to be ok for KafkaEmitterConfig.

pirvtech:
> > @pirvtech , the test failures seem genuine. Also, there are 2 open comments.
> > Please try to get these resolved soon if this patch needs to be included in Druid 38.
> 
> Error seems to be about missing @Inject annotation on the TimelineConfig constructor. Not sure why guice is looking for the @Inject annotation on the single arg constructor or requiring a default constructor. The class is registered using JsonConfigProvider with the guice binder, and the JsonConfigProvider should be able to use the single arg connector as it has the @JsonCreator annotation on it. This seems to be ok for KafkaEmitterConfig.

Figured why this is happening and added a default constructor

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
```diff
@@ -826,8 +828,11 @@ private void setRightNode(Node<T> node, Node<T> right)
...
 
-  @VisibleForTesting
-  public String print()
+  /**
+   * @return The contents of the tree as a string
+   */
+  @Override
+  public @Nullable String toString()
   {
     return (root != null) ? root.print(1) : null;
```
kfaraz: `toString()` should ideally not be null for a non-null object.
Return an empty string instead.

server/src/main/java/org/apache/druid/segment/indexing/SegmentTimelineConfig.java
```diff
@@ -24,13 +24,27 @@
...
+public class SegmentTimelineConfig
 {
+  /**
+   * Use an indexing mechanism based on an interval tree for organizing segments in memory that leads to faster search
+   */
   @JsonProperty
   private final boolean fastIntervalSearch;
 
+  // Default constructor in cases where the class is not being constructed from JSON properties (such as direct
+  // injection from Guice without the JsonConfigProvider)
```
kfaraz: Where is it being injected by Guice directly?
We should not support this mode.

server/src/main/java/org/apache/druid/segment/indexing/SegmentTimelineConfig.java
```diff
@@ -24,13 +24,27 @@
 
 import javax.annotation.Nullable;
 
-public class TimelineConfig
+/**
+ * Configuration settings related to segment timeline management
+ */
+public class SegmentTimelineConfig
 {
+  /**
```
kfaraz: Move this javadoc to the getter of this field instead.

processing/src/test/java/org/apache/druid/timeline/IntervalTreeTest.java
kfaraz: The dependency should already be added. You can just start by using the JUnit5 `@Test` annotation and `Assertions` instead of `Assert`.

processing/src/main/java/org/apache/druid/timeline/IntervalTree.java
kfaraz: @pirvtech , please fix this as well while you address the other open items.

server/src/main/java/org/apache/druid/segment/indexing/SegmentTimelineConfig.java
pirvtech: Found a test failure in one of the earlier runs
Screenshot 2026-07-20 at 10 23 21 AM


https://github.com/apache/druid/actions/runs/28553582931/job/86809994534

pirvtech: @kfaraz any comments?

kfaraz: @pirvtech , for that, you can fix up the test instead. In the `makeInjectorWithProperties`, either include the `StorageNodeModule` or bind the `SegmentTimelineConfig` directly, the same way other classes like `Validator`, `Properties`, etc are being bound.

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
```diff
@@ -0,0 +1,859 @@
...
+    this.startComparator = startComparator;
+    this.endComparator = endComparator;
+  }
+
+  public int getImbalanceTolerance()
+  {
+    return imbalanceTolerance;
+  }
+
+  public void setImbalanceTolerance(int imbalanceTolerance)
```
kfaraz, line 118: This is an internal implementation detail and need not be exposed.

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
```diff
@@ -0,0 +1,859 @@
...
+  // The comparator for comparing interval end times
+  private final Comparator<Interval> endComparator;
+
+  @VisibleForTesting
+  private Node<T> root;
+  private int size;
+
+  // Deviation allowed from ideal height for the maximum height on either side of the tree, expressed as a
+  // percentage of ideal height
+  private int imbalanceTolerance = 50;
```
kfaraz: This field should be final and accepted as an arg in the constructor. There should also be a zero-arg constructor which uses the default value for this field and the comparators.

kfaraz (changes requested):
@pirvtech , there is 1 open blocking comment:
https://github.com/apache/druid/pull/19138#discussion_r3673609335

and a couple other comments too which will be nice to address here.

Since we wanted to include this patch in Druid 38, please let me know if you would be able to address the comment and the CI issue soon.

FrankChen021 (commented):
I reviewed the update and full PR for correctness, interval boundaries, ordering, mutation, concurrency, and integration risks; no new issues were found.

---

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
pirvtech: This is configurable post construction as well. I have added constructors as requested and also documentation for the property.

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
pirvtech: This is a tunable parameter that has an effect on tree modification performance and would be useful for users of this class to be able to set depending on their circumstance. Added javadoc for it.

server/src/main/java/org/apache/druid/segment/indexing/SegmentTimelineConfig.java
pirvtech: Modified the test and removed the default constructor in SegmentTimelineConfig class

pirvtech:
> @pirvtech , there is 1 open blocking comment: [#19138 (comment)](https://github.com/apache/druid/pull/19138#discussion_r3673609335)
> 
> and a couple other comments too which will be nice to address here.
> 
> Since we wanted to include this patch in Druid 38, please let me know if you would be able to address the comment and the CI issue soon.

Addressed the comments @kfaraz

pirvtech:
@kfaraz 

> > @pirvtech , there is 1 open blocking comment: [#19138 (comment)](https://github.com/apache/druid/pull/19138#discussion_r3673609335)
> > and a couple other comments too which will be nice to address here.
> > Since we wanted to include this patch in Druid 38, please let me know if you would be able to address the comment and the CI issue soon.
> 
> Addressed the comments @kfaraz

@kfaraz I am not seeing the CI issue, could you send me a link to the test(s) that may be failing

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
```diff
@@ -0,0 +1,875 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.timeline;
+
+import com.google.common.base.Predicate;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.VisibleForTesting;
```
kfaraz: to avoid an additional dependency and prevent static CI failures.
```suggestion
import com.google.common.annotations.VisibleForTesting;
```

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
```diff
@@ -0,0 +1,875 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.timeline;
+
+import com.google.common.base.Predicate;
+import org.jetbrains.annotations.NotNull;
```
kfaraz: to simplify dependencies:
```suggestion
import javax.validation.constraints.NotNull;
```

kfaraz:
> @kfaraz I am not seeing the CI issue, could you send me a link to the test(s) that may be failing

@pirvtech , I have run the workflows, you should be able to see the failures now.
I have left comments to help with the static CI dependency check failure.

The test failures seem unrelated. For that, you could try merging the latest master into this branch and see if that helps.

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
```diff
@@ -0,0 +1,875 @@
...
+  // The compartor for comparing the interval start timnes
+  private final Comparator<Interval> startComparator;
+  // The comparator for comparing interval end times
+  private final Comparator<Interval> endComparator;
+
+  @VisibleForTesting
+  private Node<T> root;
+  private int size;
+
+  private int imbalanceTolerance;
```
kfaraz, line 81:
```suggestion
  private final int imbalanceTolerance;
```

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
```diff
@@ -0,0 +1,875 @@
...
+   * Returns the allowed tolerance between the right and left branches of the tree before triggering a rebalance.
+   * The tolerance is expressed as a percentage deviation from ideal tree height.
+   * @return The imbalance tolerance
+   */
+  public int getImbalanceTolerance()
+  {
+    return imbalanceTolerance;
+  }
+
+  public void setImbalanceTolerance(int imbalanceTolerance)
```
kfaraz, line 118: Don't use a setter.

FrankChen021 (commented):

Validation: `git diff --check` passed. A narrow Maven compile could not start because the sandbox denied writes to `~/.m2`.

---

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
pirvtech: It can be modified after construction to change the threshold.

pirvtech: Explained behavior in javadoc

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
pirvtech: It can be modified post construction, please see comment below.

kfaraz: When would you need to modify it after construction?

processing/src/main/java/org/apache/druid/timeline/IntervalTreeMap.java
kfaraz: Thanks for the details!
I am just unclear on why we would need to set this after construction.
It is currently not being called anywhere except in tests.

pirvtech:
I think configurable tolerance will be a useful feature to have, even though it isnt being used
currently, such as if different tolerance were desired when tree size is
large vs small.

kfaraz:
> I think configurable tolerance will be a useful feature to have, even though it isnt being used
currently, such as if different tolerance were desired when tree size is
large vs small.

@pirvtech , fair enough, but please don't add methods or features which are currently not being used.
We can always add it back if needed in the future. Same case for passing the comparators in the constructor of the `IntervalTreeMap`. Doing this makes it easy to understand why a certain a method or field exists in the first place. There is a lot of unused features in Druid and they only lead to confusion and clutter.

processing/src/test/java/org/apache/druid/timeline/IntervalTreeMapTest.java
```diff
@@ -0,0 +1,551 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.timeline;
+
+import org.apache.commons.collections.CollectionUtils;
```
kfaraz: This dependency can be avoided, it is currently causing the static check to fail.

processing/src/test/java/org/apache/druid/timeline/IntervalTreeMapTest.java
```diff
@@ -0,0 +1,551 @@
...
+        }
+    );
+  }
+
+  @Test
+  public void testValues()
+  {
+    IntervalTreeMap<String> tree = setupTree(baseData);
+    Collection<String> values = tree.values();
+    Collection<String> bvalues = baseData.stream().map(entry -> entry.rhs).collect(Collectors.toList());
```
kfaraz: Convert these 2 collections to sets and then do an `Assertions.assertEquals` to avoid the dependency on `CollectionUtils` for now.

kfaraz:
@pirvtech , to help with the test failures, you might need to do something similar to `SqlModuleTest` in `MsqCompactionTaskRunTest` as well.

pirvtech:
@

> @pirvtech , to help with the test failures, you might need to do something similar to `SqlModuleTest` in `MsqCompactionTaskRunTest` as well.

Fixed the test

kfaraz (approved):

PR #19179, opened 2026-03-19
feat: adds a dry run API for overlord compaction

cecemei:
### Description

  This PR **adds dry run API for overlord compaction** , providing detailed visibility into what compaction jobs would be submitted without actually running them.

  #### Created CompactionStatusDetailedStats for dry run mode

  - New class to collect detailed compaction statistics during dry run mode in Overlord
  - Organizes candidates by compaction status (COMPLETE, RUNNING, PENDING, SKIPPED) into table format
  - Includes comprehensive metrics: segment counts, bytes, rows, and uncompacted statistics
  - JSON serializable for API responses

  #### Added dryRun API endpoint

  - Added `dryRunWithConfig()` method to `CompactionScheduler` interface
  - Added `/dryRun` POST endpoint to `OverlordCompactionResource` that accepts a `ClusterCompactionConfig`
  - Returns `CompactionStatusDetailedStats` showing what would be compacted without actually submitting jobs
  - Deprecated `simulateRunWithConfigUpdate()` in favor of the new dry run API
  - Helps operators preview compaction behavior and validate configuration changes before applying them

  #### Updated CompactionStatusTracker for dry run support

  - Added optional `CompactionStatusDetailedStats` field to track detailed statistics during dry run
  - Added `recordPendingTask()` method to record tasks that could be allocated without submission
  - Updated javadocs to clarify dry run behavior

  #### Release note

  Added a dry run API (`/druid/indexer/v1/compaction/dryRun`) to the Overlord that shows what compaction jobs would be submitted for a given cluster compaction configuration without actually submitting them. This helps operators preview compaction behavior and validate configuration changes. The API returns detailed statistics including segment counts, bytes, and rows for candidates in different compaction states (pending, skipped, complete, running).
 
<hr>

This PR has:

- [x] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordCompactionResource.java
```diff
@@ -368,6 +372,16 @@ public Response simulateRunWithConfigUpdate(
     ).build();
   }
 
+  @POST
+  @Path("/dryRun")
```
kfaraz, line 376: The `/simulate` API is already meant to do a dry run. Please don't add another API for the same purpose.
We should fix the existing simulate flow instead.
It is okay to break backward compatibility with the `/simulate` API since it is not used anywhere, being experimental and not documented.

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -370,45 +378,56 @@ private synchronized void scheduledRun()
...
-        taskActionClientFactory,
-        taskLockbox,
-        overlordClient,
-        brokerClient,
-        objectMapper,
-        indexingStateStorage,
-        indexingStateCache
-    );
-    latestJobQueue.set(queue);
+    synchronized (lock) {
```
kfaraz: Please don't add another lock, this method is already `synchronized`. We shouldn't need to update the prod flow methods anyway.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionConfigBasedJobTemplate.java
```diff
@@ -99,19 +104,12 @@ public List<CompactionJob> createCompactionJobs(
                 .getCompactionPolicy()
                 .checkEligibilityForCompaction(candidate, params.getLatestTaskStatus(candidate));
       if (!eligibility.isEligible()) {
+        params.collectCompactionStatus(candidate, eligibility.getReason());
         continue;
       }
-      final CompactionCandidate finalCandidate;
       switch (eligibility.getMode()) {
```
kfaraz, line 111: We don't need this switch anymore.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -266,17 +309,40 @@ public void test_minorCompactionWithMSQ(PartitionsSpec partitionsSpec) throws Ex
             .withIoConfig(new UserCompactionTaskIOConfig(true))
             .withTuningConfig(UserCompactionTaskQueryTuningConfig.builder().partitionsSpec(partitionsSpec).build())
             .build();
-
-    runCompactionWithSpec(dayGranularityConfig);
+    String supervisorId = enableSupervisor(dayGranularityConfig);
+
+    Map<CompactionStatus.State, Table> result1 = dryRun(CompactionEngine.MSQ, policy).getCompactionStates();
+    // Expect dry run to return 1 compaction job with 2 segments
+    System.out.println("--- result1: " + result1);
```
kfaraz: Note: please remove this before committing.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/compact/CompactionSupervisorTest.java
```diff
@@ -784,12 +872,12 @@ private void verifyCompactedSegmentsHaveFingerprints(DataSourceCompactionConfig
...
 
-  private void runCompactionWithSpec(DataSourceCompactionConfig config)
+  private String enableSupervisor(DataSourceCompactionConfig config)
   {
-    cluster.callApi().postSupervisor(new CompactionSupervisorSpec(config, false, null));
+    return cluster.callApi().postSupervisor(new CompactionSupervisorSpec(config, false, null));
   }
 
-  private void pauseCompaction(DataSourceCompactionConfig config)
+  private void terminateSupervisor(String supervisorId)
```
kfaraz: Method name and arg have changed but the method contents haven't.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobParams.java
```diff
@@ -24,7 +24,9 @@
 import org.apache.druid.server.compaction.CompactionSnapshotBuilder;
 import org.apache.druid.server.compaction.CompactionTaskStatus;
 import org.apache.druid.server.coordinator.ClusterCompactionConfig;
+import org.apache.druid.server.coordinator.DataSourceCompactionConfig;
 import org.apache.druid.timeline.SegmentTimeline;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
kfaraz, line 29: Use `javax.annotation.Nullable` instead.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -187,7 +192,13 @@ public CompactionStatus getCurrentStatus()
    */
   public CompactionCandidate withCurrentStatus(CompactionStatus status)
   {
-    return new CompactionCandidate(segments, umbrellaInterval, compactionInterval, numIntervals, status);
+    return new CompactionCandidate(
+        segments,
+        umbrellaInterval,
+        compactionInterval,
+        Math.toIntExact(compactionStatistics.getNumIntervals()),
```
kfaraz, line 199: Instead of changing long to int here, update `CompactionStatistics.numIntervals` field to be an `int` instead. It should have been an int in the first place.

server/src/main/java/org/apache/druid/server/compaction/CompactionCandidate.java
```diff
@@ -42,9 +42,8 @@ public class CompactionCandidate
   private final Interval umbrellaInterval;
   private final Interval compactionInterval;
   private final String dataSource;
-  private final long totalBytes;
-  private final int numIntervals;
 
+  private final CompactionStatistics compactionStatistics;
```
kfaraz, line 46: Super Nit: Maybe rename this field to `stats` to align with the getter `getStats()`. Also, `stats` just seems nicer 🙂

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionScheduler.java
```diff
@@ -83,7 +84,15 @@ public interface CompactionScheduler
...
    */
+  @Deprecated
   CompactionSimulateResult simulateRunWithConfigUpdate(ClusterCompactionConfig updateRequest);
 
+  /**
+   * Dry run with the given config to view compaction status without submitting jobs.
+   *
+   * @return Detailed compaction statistics
+   */
+  CompactionStatusDetailedStats dryRunWithConfig(ClusterCompactionConfig config);
```
kfaraz, line 97: Please do not add this new method, use the existing `simulateRunWithConfigUpdate` method.

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -370,45 +378,56 @@ private synchronized void scheduledRun()
   }
 
   /**
-   * Creates and launches eligible compaction jobs.
+   * Creates a new compaction job queue and enqueues eligible jobs.
+   * In dry run mode, no jobs are launched and no metrics are emitted but detailed stats are collected.
    */
-  private synchronized void resetCompactionJobQueue()
+  private synchronized void resetCompactionJobQueue(
```
kfaraz, line 379: Don't update this method. Instead, construct a separate `CompactionJobQueue`, if needed in the simulate method. See https://github.com/apache/druid/pull/19179#pullrequestreview-3973379513 for details.

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -111,10 +113,16 @@ public class OverlordCompactionScheduler implements CompactionScheduler
...
-   * Compaction job queue built in the last invocation of {@link #resetCompactionJobQueue()}.
+   * Compaction job queue built in the last invocation of {@link #resetCompactionJobQueue}.
    */
   private final AtomicReference<CompactionJobQueue> latestJobQueue;
 
+  /**
+   * Lock used to synchronize access when resetting the compaction job queue.
+   * Ensures thread-safe updates to the queue and related state during queue resets.
+   */
+  private final Object lock = new Object();
```
kfaraz: This shouldn't be needed.

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -459,6 +478,9 @@ private void onTaskFinished(String taskId, TaskStatus taskStatus)
 
   private void updateCompactionSnapshots(CompactionJobQueue queue)
   {
+    if (queue.isDryRun()) {
```
kfaraz, line 474: This shouldn't be needed.

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -503,6 +526,16 @@ public CompactionSimulateResult simulateRunWithConfigUpdate(ClusterCompactionCon
     }
   }
 
+  @Override
+  public CompactionStatusDetailedStats dryRunWithConfig(ClusterCompactionConfig config)
```
kfaraz, line 523: Please remove this method. 
Instead, see https://github.com/apache/druid/pull/19179#pullrequestreview-3973379513

(None of the prod flow methods like `resetCompactionJobQueue` etc should be called in the simulate flow. The actual `CompactionStatusTracker` should be used only in a read-only capacity, same as in `CompactionRunSimulator`.)

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusDetailedStats.java
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.druid.error.DruidException;
+
+import javax.annotation.Nullable;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Collects detailed compaction statistics in table format during dry run mode in overlord.
+ */
+public class CompactionStatusDetailedStats
```
kfaraz, line 34: Instead of adding a new result class, add a builder for the existing `CompactionSimulateResult`.

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobQueue.java
```diff
@@ -107,6 +107,7 @@ public CompactionJobQueue(
       DataSourcesSnapshot dataSourcesSnapshot,
       ClusterCompactionConfig clusterCompactionConfig,
       CompactionStatusTracker statusTracker,
+      boolean dryRun,
```
kfaraz, line 110: We shouldn't need to make any change in `CompactionJobQueue`, except for those which allow it to be extended by `SimulateCompactionJobQueue`.

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -539,9 +571,9 @@ private DataSourcesSnapshot getDatasourceSnapshot()
     return segmentManager.getRecentDataSourcesSnapshot();
   }
 
-  private void scheduleOnExecutor(Runnable runnable, long delayMillis)
+  private ScheduledFuture<?> scheduleOnExecutor(Runnable runnable, long delayMillis)
```
kfaraz, line 574: Where is the future used?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordCompactionResource.java
```diff
@@ -354,6 +354,10 @@ public Response getDatasourceCompactionConfigHistory(
     }
   }
 
+  /**
+   * @deprecated Use {@link #dryRun(ClusterCompactionConfig)} instead
+   */
+  @Deprecated
```
kfaraz, line 360: Let's not deprecate this API.

kfaraz (changes requested):
The changes to the `MostFragmentedIntervalFirstPolicy` in the PR look good.

But the simulation part needs to be fixed as described below.
Given the simulation fix would be large and might need more time for reviews, it would make sense to create a separate PR for the simulation changes so that the policy changes can be merged sooner.

### Suggested approach

Simulation flow should aim to fix the current `OverlordCompactionScheduler.simulate` method and the underlying API instead of adding a new set of API/methods.
The changes made to production (i.e. non simulation) flow should be minimal (e.g. we shouldn't need to touch `CompactionJobQueue`, `CompactionStatusTracker` or even much of `OverlordCompactionScheduler`).

Overall simulation flow would be contained in `OverlordCompactionScheduler.simulate`, something like:

```java
@Override
  public CompactionSimulateResult simulateRunWithConfigUpdate(ClusterCompactionConfig updateRequest)
  {
    if (!isRunning()) {
       return new CompactionSimulateResult(Map.of());
    }
    
    // Construct simulation specific dependencies which perform read
    // from prod classes but never write to them
    // See CompactionRunSimulator for reference
    final ReadOnlyOverlordClient overlordClient = ...;
    final SimulateCompactionStatusTracker wrappedStatusTracker = ...;
    
    // This class extends CompactionJobQueue but does not perform any write ops
    // like submitting a task, updating locks, updating indexing storage/cache, etc.
    final SimulateCompactionJobQueue jobQueue = new SimulateCompactionJobQueue(...);
    
    final CompactionSimulateResult.Builder resultBuilder = new Builder();
    
    jobQueue.getSkipped().forEach(resultBuilder::addSkipped); // jobs maybe skipped either by datasource iterator or by policy
    jobQueue.getCompleted().forEach(resultBuilder::addComplete); // fully compacted
    jobQueue.getQueued().forEach(resultBuilder::addQueued); // for queued jobs, we should also capture the mode i.e. full or minor
    
    return resultBuilder.build();
  }
```

indexing-service/src/main/java/org/apache/druid/indexing/compact/CompactionJobParams.java
```diff
@@ -35,6 +37,7 @@ public class CompactionJobParams
   private final DateTime scheduleStartTime;
   private final TimelineProvider timelineProvider;
   private final CompactionTaskStatusProvider compactionTaskStatusProvider;
+  private final CompactionStatusCollector compactionStatusCollector;
```
kfaraz, line 40: We shouldn't need this. It seems very similar in role to the snapshot builder.
We can keep track of the skipped/completed jobs in the snapshot builder itself.

cecemei:
> e a separate PR for the simulation changes so that the policy changes can be merged sooner.

hey split up the policy change here: https://github.com/apache/druid/pull/19205

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusDetailedStats.java
```diff
@@ -0,0 +1,131 @@
...
+    Preconditions.checkNotNull(candidate.getStats(), "compaction stats");
+    Preconditions.checkNotNull(candidate.getUncompactedStats(), "uncompacted stats");
+    stateToTable.get(CompactionStatus.State.RUNNING).addRow(
+        candidate.getDataSource(),
+        candidate.getCompactionInterval(),
+        candidate.numSegments(),
+        candidate.getStats().getTotalBytes(),
+        candidate.getStats().getTotalRows(),
+        candidate.getUncompactedStats().getNumSegments(),
+        candidate.getUncompactedStats().getTotalRows(),
```
FrankChen021, line 114: [P2] RUNNING rows omit uncompactedBytes

recordSubmittedTask declares tables with 10 columns including uncompactedBytes, but RUNNING rows only add 9 values here: uncompactedRows is written into the uncompactedBytes position, reason is shifted into uncompactedRows, and mode is shifted into reasonToCompactOrSkip. Clients reading dryRun RUNNING results by column name will get corrupted values; add getUncompactedStats().getTotalBytes() before totalRows and update the test expectation.

FrankChen021 (commented):
- indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java:404 - [P2] Dry run replaces the live compaction queue. The dry-run path calls resetCompactionJobQueue(true, ...) and still stores the newly created dry-run queue in latestJobQueue. That same shared queue is used by task-finish callbacks to requeue/launch pending real compaction jobs, so a POST /compaction/dryRun can discard the live queue and leave subsequent completions operating on a dry-run queue until the next scheduled run. The dry run should not replace scheduler state, or it should restore the previous queue/status collector before returning.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 3 |
| P3 | 0 |
| Total | 3 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 3 |
| P3 | 0 |
| Total | 3 |

Reviewed 15 of 15 changed files.

---

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -503,6 +519,22 @@ public CompactionSimulateResult simulateRunWithConfigUpdate(ClusterCompactionCon
     }
   }
 
+  @Override
+  public CompactionStatusDetailedStats dryRunWithConfig(ClusterCompactionConfig config)
+  {
+    CompactionStatusDetailedStats detailedStats = new CompactionStatusDetailedStats();
+    if (isRunning() && isEnabled()) {
+      try {
+        scheduleOnExecutor(() -> resetCompactionJobQueue(true, config, detailedStats), 0L).get();
```
FrankChen021, line 528: [P2] Dry run replaces the live compaction queue

This invokes the same resetCompactionJobQueue path used by the live scheduler, which clears latestJobQueue and then installs a dry-run queue. After this future completes, task-finish callbacks and later queue updates operate on that dry-run queue, so real pending compaction jobs can stop launching and snapshots can remain stale until the next scheduled reset. Build the dry-run state independently, or save and restore the live queue and tracker state in a finally block.

server/src/main/java/org/apache/druid/server/compaction/CompactionStatusDetailedStats.java
FrankChen021, line 114: [P2] Running-stat rows omit uncompacted bytes

The table declares ten columns in the order uncompactedSegments, uncompactedBytes, uncompactedRows, reasonToCompactOrSkip, and mode, but this row supplies only nine values and omits getUncompactedStats().getTotalBytes(). The remaining values shift into the wrong columns and mode is missing from the row, so the new running-task response is malformed. Add uncompacted bytes between the segment count and row count.

indexing-service/src/main/java/org/apache/druid/indexing/compact/OverlordCompactionScheduler.java
```diff
@@ -539,9 +571,9 @@ private DataSourcesSnapshot getDatasourceSnapshot()
     return segmentManager.getRecentDataSourcesSnapshot();
   }
 
-  private void scheduleOnExecutor(Runnable runnable, long delayMillis)
+  private ScheduledFuture<?> scheduleOnExecutor(Runnable runnable, long delayMillis)
   {
-    executor.schedule(
+    return executor.schedule(
```
FrankChen021, line 576: [P2] Dry-run executor failures are swallowed

The newly returned future is not enough to propagate failures because the scheduled runnable still catches every Throwable and only logs it. Consequently, dryRunWithConfig(...).get() completes normally even when resetCompactionJobQueue fails, and the endpoint can return HTTP 200 with empty or partial statistics instead of the intended error response. Use a propagation path for this synchronous dry-run call, or rethrow the task failure through the future while preserving the existing logging behavior for asynchronous scheduling.

PR #19235, opened 2026-03-30
feat: Early query cancellation based on per-segment sampling

mshahid6:
### Description

Added early query cancellation on historicals by sampling segment completion times and extrapolating whether the query will exceed its timeout. After a configurable number of segments complete, elapsed time is measured and used to project total query time. If the projection exceeds the remaining timeout, all pending segment futures are cancelled immediately.

New query context parameter is `perSegmentSamplingWindow` (number of segments to complete before extrapolating) with a default value of 0 i.e. disabled.

Can be set per-query via query context, or as a system default via druid.query.default.context.perSegmentSamplingWindow.

**Example**

```
  {
    "context": {
      "timeout": 30000,
      "perSegmentSamplingWindow": 5
    }
  }
```

After 5 segments complete, if wall-clock extrapolation suggests the query will exceed 30 seconds, it fails fast instead of waiting for the full timeout.

 * `ChainedExecutionQueryRunner`
 * `QueryContexts/QueryContext`
 * `ChainedExecutionQueryRunnerExtrapolationTest`

**Limitations**

- GroupByMergingQueryRunner is not covered yet (follow-up)
- Extrapolation assumes roughly uniform segment cost; skewed first segments could cause false positives or misses

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
+++ b/processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
@@ -28,6 +28,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 import org.apache.druid.common.guava.GuavaUtils;
 import org.apache.druid.java.util.common.ISE;
 import org.apache.druid.java.util.common.StringUtils;
+import org.apache.druid.java.util.common.concurrent.Execs;
 import org.apache.druid.java.util.common.guava.BaseSequence;
 import org.apache.druid.java.util.common.guava.MergeIterable;
 import org.apache.druid.java.util.common.guava.Sequence;
@@ -40,6 +41,8 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * A QueryRunner that combines a list of other QueryRunners and executes them in parallel on an executor.
@@ -85,6 +88,13 @@ public class ChainedExecutionQueryRunner<T> implements QueryRunner<T>
     final QueryContext context = query.context();
     final boolean usePerSegmentTimeout = context.usePerSegmentTimeout();
     final long perSegmentTimeout = context.getPerSegmentTimeout();
+    final int samplingWindow = context.getPerSegmentSamplingWindow();
+    final long queryStartNanos = System.nanoTime();
+
+    // Shared state for extrapolation — only allocated when sampling is enabled
+    final AtomicInteger completedSegments = samplingWindow > 0 ? new AtomicInteger(0) : null;
+    final AtomicBoolean extrapolationCancelled = samplingWindow > 0 ? new AtomicBoolean(false) : null;
+
     return new BaseSequence<>(
         new BaseSequence.IteratorMaker<>()
         {
@@ -120,6 +130,10 @@ public class ChainedExecutionQueryRunner<T> implements QueryRunner<T>
                                   throw new ISE("Got a null list of results");
                                 }
 
+                                if (completedSegments != null) {
+                                  completedSegments.incrementAndGet();
+                                }
+
                                 return retVal;
                               }
                               catch (QueryInterruptedException e) {
@@ -153,8 +167,32 @@ public class ChainedExecutionQueryRunner<T> implements QueryRunner<T>
                     )
                 );
 
+            final int totalSegments = futures.size();
             ListenableFuture<List<Iterable<T>>> future = Futures.allAsList(futures);
             queryWatcher.registerQueryFuture(query, future);
+            
+            if (completedSegments != null && totalSegments >= samplingWindow && context.hasTimeout()) {
+              for (ListenableFuture<?> f : futures) {
+                f.addListener(
+                    () -> {
+                      if (extrapolationCancelled.get()) {
+                        return;
+                      }
+                      int completed = completedSegments.get();
+                      if (completed >= samplingWindow) {
+                        long elapsedMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - queryStartNanos);
+                        long extrapolatedMs = elapsedMs * totalSegments / completed;
+                        long remainingMs = context.getTimeout() - elapsedMs;
+                        if (extrapolatedMs > context.getTimeout() && remainingMs < extrapolatedMs - elapsedMs) {
+                          extrapolationCancelled.set(true);
+                          GuavaUtils.cancelAll(true, future, futures);
+                        }
+                      }
+                    },
+                    Execs.directExecutor()
+                );
+              }
+            }
 
             try {
               return new MergeIterable<>(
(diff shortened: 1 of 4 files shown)


processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
```diff
@@ -153,8 +167,32 @@ public Iterable<T> call()
...
+                        long elapsedMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - queryStartNanos);
+                        long extrapolatedMs = elapsedMs * totalSegments / completed;
+                        long remainingMs = context.getTimeout() - elapsedMs;
+                        if (extrapolatedMs > context.getTimeout() && remainingMs < extrapolatedMs - elapsedMs) {
+                          extrapolationCancelled.set(true);
+                          GuavaUtils.cancelAll(true, future, futures);
+                        }
+                      }
+                    },
+                    Execs.directExecutor()
```
jtuglu1, line 192: I have some concerns over noisy-neighbor/variability that might cause thing to fire more often than not. Another concern is since this is operating on the servicing Jetty thread, this might cause interrupts to occur on the thread blocking on .get():

I think we should be careful about scheduling async tasks on the main executor aside from the primary timeout. The main thread's job is to wait for completion of all futures, but if there are other competing tasks it needs to service, I want to make sure that:

a) Under contention we cannot possibly get InterruptedException (to go service a callback) and bail the processing on a valid query.
b) We don't delay processing of the future group because we are busy servicing a callback for one processing future.

Do we have a way of validating this won't happen?

jtuglu1 (commented):
Few comments, I would like to actually test this under load to understand whether we will actually benefit from something like this. I've noticed lots of skew/delay in processing handling if the thread pools are overloaded.

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
```diff
@@ -153,8 +167,32 @@ public Iterable<T> call()
...
+              for (ListenableFuture<?> f : futures) {
+                f.addListener(
+                    () -> {
+                      if (extrapolationCancelled.get()) {
+                        return;
+                      }
+                      int completed = completedSegments.get();
+                      if (completed >= samplingWindow) {
+                        long elapsedMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - queryStartNanos);
+                        long extrapolatedMs = elapsedMs * totalSegments / completed;
```
jtuglu1, line 184: I'm also worried about this calculation is not accounting for parallelism correctly and being "overkill." It's approximating of the time taken to complete a query as effectively avg segment time * total segments, which is not usually the case(it is assuming a purely serial execution). We have 100+ threads processing segments, and queries can sometimes query 100-1000s of segments per historical.

jtuglu1: For example, you could have a query that queries 300 segments each taking 1100ms with a total query timeout of 300s, we would preemptively kill this query even though assuming moderate contention it would complete in ~84s.

jtuglu1: We should instead incorporate the parallelism of the threadpool into the calculation to avoid killing ok queries. For example, something like (sum of extrapolated segment times) / (thread pool parallelism).

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
```diff
@@ -85,6 +88,13 @@ public Sequence<T> run(final QueryPlus<T> queryPlus, final ResponseContext respo
     final QueryContext context = query.context();
     final boolean usePerSegmentTimeout = context.usePerSegmentTimeout();
     final long perSegmentTimeout = context.getPerSegmentTimeout();
+    final int samplingWindow = context.getPerSegmentSamplingWindow();
+    final long queryStartNanos = System.nanoTime();
```
FrankChen021, line 92: [P2] Start extrapolation clock when the lazy sequence is consumed

ChainedExecutionQueryRunner.run returns a lazy BaseSequence, and the segment futures are not submitted until IteratorMaker.make() runs during sequence consumption. Starting queryStartNanos here means any delay between creating the sequence and consuming it is counted as segment execution time, unlike the existing future.get(context.getTimeout(), ...) timeout which starts inside make(). With perSegmentSamplingWindow enabled, a wrapper or delayed consumer can cause the first completed segments to extrapolate from idle time and cancel a query that has not actually exceeded its processing timeout; move the start timestamp into make() near the future submission/listener setup.

maytasm: +1 Move queryStartNanos capture inside make(), just before the futures are submitted.

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
maytasm, line 184: elapsedMs * totalSegments can overflow (although unlikely). can rewrite to `long extrapolatedMs = elapsedMs / completed * totalSegments;`

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
```diff
@@ -153,8 +167,32 @@ public Iterable<T> call()
...
+                    () -> {
+                      if (extrapolationCancelled.get()) {
+                        return;
+                      }
+                      int completed = completedSegments.get();
+                      if (completed >= samplingWindow) {
+                        long elapsedMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - queryStartNanos);
+                        long extrapolatedMs = elapsedMs * totalSegments / completed;
+                        long remainingMs = context.getTimeout() - elapsedMs;
+                        if (extrapolatedMs > context.getTimeout() && remainingMs < extrapolatedMs - elapsedMs) {
```
maytasm, line 186: Aren't these two conditions equivalent? Why do we need both?

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
maytasm: I think the formula does account for parallelism. The parallelism is already baked into elapsedMs / completed since elapsedMs is average wall-clock time per completion not avg per segment time.
Example: 100 threads, 1000 segments, each segment taking 300ms:                                                                                                                                                                                                                                                                                                               
  - First 100 segments run in parallel, complete at ~300ms wall-clock                                                                                                                                                                                                                                                                                                       
  - elapsedMs=300, completed=100, totalSegments=1000
  - Formula: 300 * 1000 / 100 = 3,000ms                                                                                                                                                                                                                                                                                                                                     
  - Actual: ceil(1000/100) * 300ms = 3,000ms

maytasm: Actually this calculation is an issue too. The wall-clock extrapolation conflates queue wait time with processing time. Under thread pool contention, segments sit in the queue, elapsed wall-clock grows, but completed stays low — the ratio elapsedMs / completed explodes.

maytasm: track actual per-segment processing time (inside the callable, before/after the input.run() call) rather than wall-clock time from query start. If we do this, then as @jtuglu1 mentioned above, we should also account for threadPoolSize (aka the parallelism). So maybe something like estimated_remaining = avg_segment_time * ceil(remainingSegments / threadPoolSize)

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
maytasm: The listeners use Execs.directExecutor(), which means they run on the processing pool thread that completes the future, not on the Jetty/main thread. The main thread is parked inside future.get() waiting for the allAsList future to resolve.

maytasm: cancel interrupts the task runner threads (processing pool), not threads waiting on get(). The Jetty thread receives a CancellationException through the normal Future.get() contract — the same mechanism that already fires for regular query cancellation.

maytasm (changes requested):

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
jtuglu1: > I think the formula does account for parallelism. The parallelism is already baked into elapsedMs / completed since elapsedMs is average wall-clock time per completion not avg per segment time.
Example: 100 threads, 1000 segments, each segment taking 300ms:

I disagree. I think this is an example of a "good" case. Look at something like samplingWindow=5, pool size of 100. You need to take parallelism into account (at least, your sample size should be ≥ parallelism).

processing/src/main/java/org/apache/druid/query/ChainedExecutionQueryRunner.java
FrankChen021: Thanks, agreed. The current PR head still captures `queryStartNanos` in `run()`, so moving it into `make()` immediately before the futures are submitted/listeners are registered should address this by measuring only the actual sequence consumption/execution window.

PR #19236, opened 2026-03-30
feat: support user identity forwarding to tasks via TaskAuthContext

### LINKED ISSUE #18957: User authentication propagation to ingestion tasks and credential vending
### Description

This would allow a user to POST a query/task and allow the task to use the identity of the user to access some external state (e.g. Iceberg table). Currently, the Iceberg plugin assumes once the catalog loads the table, the subtasks will have access to the data (via static warehouse credentials). This is brittle from a security standpoint as it relies on Druid having perpetual access to external Iceberg data. The ideal solution would provide a way for Druid tasks to "vend" credentials for accessing external resources using on behalf of a user identity to ingest data.

My initial thought is:
```java
@ExtensionPoint
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
public interface TaskAuthContext
{
  /**
   * Returns the authenticated identity (e.g., username, email, service account name).
   * This value is safe to serialize and log.
   *
   * @return the identity string
   */
  String getIdentity();

  /**
   * Returns sensitive credentials (e.g., OAuth tokens, API keys).
   * This method MUST be redacted during serialization via {@link TaskAuthContextRedactionMixIn}.
   *
   * <p>The map keys are credential identifiers (e.g., "token", "oauth2-token") and values
   * are the actual credential strings. The specific keys expected depend on the consumer
   * (e.g., Iceberg REST Catalog may expect "token").
   *
   * @return map of credential name to credential value, or null if no credentials
   */
  @Nullable
  Map<String, String> getCredentials();

  /**
   * Returns non-sensitive metadata about the auth context (e.g., token expiry time, scopes, roles).
   * This value is safe to serialize and log.
   *
   * @return map of metadata, or null if no metadata
   */
  @Nullable
  default Map<String, Object> getMetadata()
  {
    return null;
  }
}
```

then to create the TaskAuthContext:

```java
@ExtensionPoint
public interface TaskAuthContextProvider
{
  /**
   * Extract auth context from the authentication result for the given task.
   *
   * @param authenticationResult the authentication result from the Authenticator,
   *                             containing identity and context map with credentials
   * @param task                 the task being submitted, can be used to make decisions
   *                             based on task type, datasource, etc.
   * @return TaskAuthContext to inject into the task, or null to skip injection
   */
  @Nullable
  TaskAuthContext createTaskAuthContext(AuthenticationResult authenticationResult, Task task);
}
```

and finally, to inject the task auth context into the task spec in `OverlordResource`:

```java
// Inject auth context if provider is configured
if (taskAuthContextProvider != null) {
  final AuthenticationResult authenticationResult = AuthorizationUtils.authenticationResultFromRequest(req);
  if (authenticationResult != null) {
    final TaskAuthContext taskAuthContext = taskAuthContextProvider.createTaskAuthContext(
        authenticationResult,
        task
    );
    if (taskAuthContext != null) {
      task.setTaskAuthContext(taskAuthContext);
    }
  }
}
```

My proposed flow would be:
- User submits MSQ query/task
- User identity is propagated into task via a mixin (to hide credentials)
- Iceberg catalog (or any other input source for that matter) uses user credentials to vend warehouse credentials to read Iceberg data from object store
- These warehouse credentials are then passed to sub-tasks who are handed the warehouse input source.
- The user credentials are not persisted to DB and are present only via the serialized task payload sent from Overlord -> Peon. This limits the restartability of tasks (due to lack of dynamic credential refresh), but still should work for most use-cases.
- This paves way for things like dynamic credential refresh (if needed).

### Motivation

Support credential vending using Druid Authorizer logic.

jtuglu1:
Closes https://github.com/apache/druid/issues/18957.

### Description

Create a mechanism to propagate user identity context from authenticator to tasks. This allows tasks to safely operate with a sort of per-task identity, allowing tasks to access information dynamically that would be hard to setup in the current state of things. This allows for auth mechanisms like catalog credential vending (through Iceberg Catalog, etc.) and allows Druid to operate more like a standalone engine that can integrate with other things within a larger data ecosystem.

#### Constraints

This currently does NOT support task restarts.

#### Release note

Create a mechanism to propagate user identity context from authenticator to tasks to support things like credential vending, etc.

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/common/task/batch/parallel/ParallelIndexSupervisorTask.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/common/task/batch/parallel/ParallelIndexSupervisorTask.java
@@ -34,6 +34,7 @@ import org.apache.datasketches.memory.Memory;
 import org.apache.druid.common.guava.FutureUtils;
 import org.apache.druid.data.input.InputFormat;
 import org.apache.druid.data.input.InputSource;
+import org.apache.druid.data.input.TaskAuthContextAware;
 import org.apache.druid.indexer.IngestionState;
 import org.apache.druid.indexer.TaskState;
 import org.apache.druid.indexer.TaskStatus;
@@ -539,6 +540,11 @@ public class ParallelIndexSupervisorTask extends AbstractBatchIndexTask
 
       initializeSubTaskCleaner();
       this.toolbox = toolbox;
+
+      // Propagate auth context to input source for credential vending (e.g., Iceberg REST Catalog)
+      if (baseInputSource instanceof TaskAuthContextAware) {
+        ((TaskAuthContextAware) baseInputSource).setTaskAuthContext(getTaskAuthContext());
+      }
 
       if (isParallelMode()) {
         // emit metric for parallel batch ingestion mode:
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordResource.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordResource.java
@@ -30,6 +30,7 @@ import com.google.inject.Inject;
 import com.sun.jersey.spi.container.ResourceFilters;
 import org.apache.druid.audit.AuditEntry;
 import org.apache.druid.audit.AuditManager;
+import org.apache.druid.auth.TaskAuthContext;
 import org.apache.druid.client.indexing.ClientTaskQuery;
 import org.apache.druid.common.config.ConfigManager.SetResult;
 import org.apache.druid.common.config.Configs;
@@ -43,6 +44,7 @@ import org.apache.druid.indexer.TaskStatusPlus;
 import org.apache.druid.indexing.common.actions.TaskActionClient;
 import org.apache.druid.indexing.common.actions.TaskActionHolder;
 import org.apache.druid.indexing.common.task.Task;
+import org.apache.druid.indexing.common.task.TaskAuthContextProvider;
 import org.apache.druid.indexing.overlord.DruidOverlord;
 import org.apache.druid.indexing.overlord.IndexerMetadataStorageAdapter;
 import org.apache.druid.indexing.overlord.TaskMaster;
@@ -67,6 +69,7 @@ import org.apache.druid.server.http.security.DatasourceResourceFilter;
 import org.apache.druid.server.http.security.StateResourceFilter;
 import org.apache.druid.server.security.Action;
 import org.apache.druid.server.security.AuthConfig;
+import org.apache.druid.server.security.AuthenticationResult;
 import org.apache.druid.server.security.AuthorizationResult;
 import org.apache.druid.server.security.AuthorizationUtils;
 import org.apache.druid.server.security.AuthorizerMapper;
@@ -122,6 +125,13 @@ public class OverlordResource
 
   private final AuthConfig authConfig;
 
+  /**
+   * Optional provider for injecting auth context into tasks.
+   * If not bound, no auth context will be injected.
+   */
+  @Nullable
+  private TaskAuthContextProvider taskAuthContextProvider;
+
   private static final List<String> API_TASK_STATES = ImmutableList.of("pending", "waiting", "running", "complete");
   private static final Set<String> AUDITED_TASK_TYPES
       = ImmutableSet.of("index", "index_parallel", "compact", "kill");
@@ -150,6 +160,16 @@ public class OverlordResource
     this.authorizerMapper = authorizerMapper;
     this.workerTaskRunnerQueryAdapter = workerTaskRunnerQueryAdapter;
     this.authConfig = authConfig;
+  }
+
+  /**
+   * Optional setter for TaskAuthContextProvider. If bound in Guice, this will be called
+   * to enable auth context injection into tasks.
+   */
+  @Inject(optional = true)
+  public void setTaskAuthContextProvider(@Nullable TaskAuthContextProvider provider)
+  {
+    this.taskAuthContextProvider = provider;
   }
 
   /**
@@ -183,6 +203,27 @@ public class OverlordResource
     );
     if (!authResult.allowAccessWithNoRestriction()) {
       throw new ForbiddenException(authResult.getErrorMessage());
+    }
+
+    // Inject auth context if provider is configured
+    if (taskAuthContextProvider != null) {
+      final AuthenticationResult authenticationResult = AuthorizationUtils.authenticationResultFromRequest(req);
+      if (authenticationResult != null) {
+        final TaskAuthContext taskAuthContext = taskAuthContextProvider.createTaskAuthContext(
+            authenticationResult,
+            task
+        );
+        if (taskAuthContext != null) {
+          task.setTaskAuthContext(taskAuthContext);
+          if (task.getTaskAuthContext() == null) {
+            log.warn(
+                "Auth context was provided but task type [%s] does not support it (dropped for task [%s])",
+                task.getType(),
+                task.getId()
+            );
+          }
+        }
+      }
     }
 
     return asLeaderWith(
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/sql/MSQTaskQueryMaker.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/sql/MSQTaskQueryMaker.java
@@ -81,6 +81,7 @@ import org.apache.druid.sql.destination.IngestDestination;
 import org.apache.druid.sql.destination.TableDestination;
 import org.apache.druid.sql.hook.DruidHook;
 import org.apache.druid.sql.http.ResultFormat;
+import org.apache.druid.utils.CollectionUtils;
 import org.joda.time.Interval;
 
 import javax.annotation.Nullable;
@@ -171,7 +172,13 @@ public class MSQTaskQueryMaker implements QueryMaker
         taskContext
     );
 
-    FutureUtils.getUnchecked(overlordClient.runTask(taskId, controllerTask), true);
+    // Propagate auth context headers to Overlord for consumption
+    if (plannerContext.getAuthenticationResult() != null && plannerContext.getAuthenticationResult().getContext() != null) {
+      final Map<String, String> extraHeaders = CollectionUtils.mapValues(plannerContext.getAuthenticationResult().getContext(), String::valueOf);
+      FutureUtils.getUnchecked(overlordClient.runTask(taskId, controllerTask, extraHeaders), true);
+    } else {
+      FutureUtils.getUnchecked(overlordClient.runTask(taskId, controllerTask), true);
+    }
     return QueryResponse.withEmptyContext(Sequences.simple(Collections.singletonList(new Object[]{taskId})));
   }
 
--- a/processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
+++ b/processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.auth;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import org.apache.druid.guice.annotations.ExtensionPoint;
+
+import javax.annotation.Nullable;
+import java.util.Map;
+
+/**
+ * Holds authentication context that can be passed to tasks for accessing external services
+ * that require user credentials (e.g., Iceberg REST Catalog with OAuth).
+ *
+ * <p>Implementations must ensure credentials are redacted during serialization by applying
+ * {@link TaskAuthContextRedactionMixIn} to the ObjectMapper used for persistence/logging.
+ * This follows the same pattern as {@link org.apache.druid.metadata.PasswordProvider} and
+ * {@link org.apache.druid.metadata.PasswordProviderRedactionMixIn}.
+ *
+ * <p>The auth context is injected into tasks at submission time by {@code TaskAuthContextProvider}
+ * and is available during task execution. It is NOT persisted to metadata storage - only the
+ * non-sensitive fields (identity, metadata) are serialized.
+ *
+ * <p><b>Credential lifetime:</b> Vended credentials (OAuth tokens, STS session tokens) have
+ * limited lifetimes, typically 1-12 hours. There is currently no credential refresh mechanism,
+ * so long-running tasks may fail when credentials expire. Task restarts are also not supported
+ * since the original credentials are not preserved.
+ *
+ * @see TaskAuthContextRedactionMixIn
+ */
+@ExtensionPoint
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public interface TaskAuthContext
+{
+  /**
+   * Returns the authenticated identity (e.g., username, email, service account name).
+   * This value is safe to serialize and log.
+   *
+   * @return the identity string
+   */
+  String getIdentity();
+
+  /**
+   * Returns sensitive credentials (e.g., OAuth tokens, API keys).
+   * This method MUST be redacted during serialization via {@link TaskAuthContextRedactionMixIn}.
+   *
+   * <p>The map keys are credential identifiers (e.g., "token", "oauth2-token") and values
+   * are the actual credential strings. The specific keys expected depend on the consumer
+   * (e.g., Iceberg REST Catalog may expect "token").
+   *
+   * @return map of credential name to credential value, or null if no credentials
+   */
+  @Nullable
+  Map<String, String> getCredentials();
+
+  /**
+   * Returns non-sensitive metadata about the auth context (e.g., token expiry time, scopes, roles).
+   * This value is safe to serialize and log.
+   *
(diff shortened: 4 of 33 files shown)


cecemei:
@jtuglu1 whats the status for this PR? should we move this to 38?

jtuglu1:
> @jtuglu1 whats the status for this PR? should we move this to 38?

This one I wanted to get into v37 – I was hoping to get this merged today.

processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
```diff
@@ -0,0 +1,84 @@
...
+   * Returns the authenticated identity (e.g., username, email, service account name).
+   * This value is safe to serialize and log.
+   *
+   * @return the identity string
+   */
+  String getIdentity();
+
+  /**
+   * Returns sensitive credentials (e.g., OAuth tokens, API keys).
+   * This method MUST be redacted during serialization via {@link TaskAuthContextRedactionMixIn}.
```
gianm, line 62: This is over-eager, isn't it? The credentials must be serialized in some cases. When we send a task spec from Overlord to the server that will actually run the task, the credentials must be included. In some cases the nonredacted task file will need to end up on disk, so it can be run. Perhaps this should say "MUST be redacted when written to log files, the metadata store, or returned from user-facing APIs".

processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
```diff
@@ -0,0 +1,84 @@
...
+ * This follows the same pattern as {@link org.apache.druid.metadata.PasswordProvider} and
+ * {@link org.apache.druid.metadata.PasswordProviderRedactionMixIn}.
+ *
+ * <p>The auth context is injected into tasks at submission time by {@code TaskAuthContextProvider}
+ * and is available during task execution. It is NOT persisted to metadata storage - only the
+ * non-sensitive fields (identity, metadata) are serialized.
+ *
+ * <p><b>Credential lifetime:</b> Vended credentials (OAuth tokens, STS session tokens) have
+ * limited lifetimes, typically 1-12 hours. There is currently no credential refresh mechanism,
+ * so long-running tasks may fail when credentials expire. Task restarts are also not supported
```
gianm, line 43: Any thoughts on how to handle long-running tasks? I expect it will be a problem. It's not uncommon for tasks to take hours.

processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
```diff
@@ -0,0 +1,84 @@
...
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public interface TaskAuthContext
+{
+  /**
+   * Returns the authenticated identity (e.g., username, email, service account name).
+   * This value is safe to serialize and log.
+   *
+   * @return the identity string
+   */
+  String getIdentity();
```
gianm, line 58: The auth context is a `@JsonProperty` so I believe that means people can set it explicitly when they submit tasks. Does anything bad happen if someone sets a context and sets the identity to someone else? What do you think about clearing the user-provided auth context, if any, in `OverlordResource`?

multi-stage-query/src/main/java/org/apache/druid/msq/sql/MSQTaskQueryMaker.java
```diff
@@ -171,7 +172,13 @@ public QueryResponse<Object[]> runQuery(final DruidQuery druidQuery)
         taskContext
     );
 
-    FutureUtils.getUnchecked(overlordClient.runTask(taskId, controllerTask), true);
+    // Propagate auth context headers to Overlord for consumption
+    if (plannerContext.getAuthenticationResult() != null && plannerContext.getAuthenticationResult().getContext() != null) {
```
gianm, line 176: What is the purpose of this? Authentication context is meant to be an arbitrary extension-defined in-memory map. It may not in general take well to being stuffed into a header. It may also include sensitive information that shouldn't be sent in a header.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordResource.java
```diff
@@ -185,6 +205,27 @@ public Response taskPost(
       throw new ForbiddenException(authResult.getErrorMessage());
     }
 
+    // Inject auth context if provider is configured
+    if (taskAuthContextProvider != null) {
+      final AuthenticationResult authenticationResult = AuthorizationUtils.authenticationResultFromRequest(req);
```
gianm, line 210: How will this work in the SQL DML path, where the user submits a task to `/druid/v2/sql/task/` and the Broker then submits the task using its own credentials? The current design is that the Broker authenticates the user, authorizes the DML operation, and then submits it to the Overlord using a service account (not the user's own credentials).

IMO it would make more sense in this case for the Broker to get the vended credentials and pass them along to the Overlord.

processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
```diff
@@ -0,0 +1,84 @@
...
+ * <p><b>Credential lifetime:</b> Vended credentials (OAuth tokens, STS session tokens) have
+ * limited lifetimes, typically 1-12 hours. There is currently no credential refresh mechanism,
+ * so long-running tasks may fail when credentials expire. Task restarts are also not supported
+ * since the original credentials are not preserved.
+ *
+ * @see TaskAuthContextRedactionMixIn
+ */
+@ExtensionPoint
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public interface TaskAuthContext
```
gianm, line 50: I didn't see an implementation of `TaskAuthContext` or `TaskAuthContextProvider` in this PR. Are they meant to be added later? What would they look like? I was wondering in particular what `taskAuthContextProvider.createTaskAuthContext` would do exactly with the `authenticationResult` that is passed in.

gianm (commented):
I reviewed the core changes only, focusing on trying to understand the security properties of the changes.

I wonder what your thoughts are on an alternate design that keeps the vended credentials inside the input source itself:

- Add a new method to `InputSource` like `scopeForUser(AuthenticationResult authResult)`. Default implementation is `return this`
- Whenever a task is submitted, call `scopeForUser` on all of its input sources at whichever service initially accepts the task (either Broker [for SQL DML] or Overlord [for anything else]).
- The `IcebergInputSource` would implement `scopeForUser` to fetch the vended credentials and transform itself into an input source that bakes in the vended credentials. It would use a `PasswordProvider` so it is redactable.

The idea would be to avoid the need for a new credential vending system in core, putting most of the changes inside the Iceberg extension instead.

processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
jtuglu1: Yeah the comment here is a bit aggressive, but that's currently what is already done. The field is not redacted when being submitted from Overlord to MM/Peon/Indexer. It is only redacted currently on logging, persistence (to disk or db).

processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
jtuglu1: I was going to add this in a separate change. Using a task auth context, inputSource could conceivably just use that with some implementable revend() method. The only issue I would want to avoid is having the subtasks refresh the credentials (and not the driver task), since this might cause thundering herd effect where many subtasks all attempt to refresh at once. Spark does a better job of this by letting executor handle credential refresh on the driver, then propagate that to the executors.

processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
jtuglu1: > Does anything bad happen if someone sets a context and sets the identity to someone else?

If we want to scope the auth context to Druid-settable only, then yes we should. IMO, I think that's a safer option but I could see people wanting to use this apart from Authorization result.

multi-stage-query/src/main/java/org/apache/druid/msq/sql/MSQTaskQueryMaker.java
jtuglu1: We need a way to propagate AuthenticationResult from the broker to the overlord. I could make an overridable serialization method/class for AuthenticationResult to ensure we only pass-through what's needed (and let the user specify that), but propagating the headers here was the least intrusive approach for that change. We still need to pass the authentication result somehow.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordResource.java
jtuglu1: > How will this work in the SQL DML path, where the user submits a task to /druid/v2/sql/task/ and the Broker then submits the task using its own credentials?

That's the thing. In our implementation, the broker passes through the user auth context and not its own credentials, so there's no use of the Broker credentials in the task payload (only used for validating the request came from the Brokers).

processing/src/main/java/org/apache/druid/auth/TaskAuthContext.java
jtuglu1: They are meant for users to implement. I have internal versions of these classes which take our internal identity tokens and propagate them through to the Iceberg input source, to be then used for vending S3 credentials to read data from Iceberg.

jtuglu1:
> I reviewed the core changes only, focusing on trying to understand the security properties of the changes.
> 
> I wonder what your thoughts are on an alternate design that keeps the vended credentials inside the input source itself:
> 
> * Add a new method to `InputSource` like `scopeForUser(AuthenticationResult authResult)`. Default implementation is `return this`
> * Whenever a task is submitted, call `scopeForUser` on all of its input sources at whichever service initially accepts the task (either Broker [for SQL DML] or Overlord [for anything else]).
> * The `IcebergInputSource` would implement `scopeForUser` to fetch the vended credentials and transform itself into an input source that bakes in the vended credentials. It would use a `PasswordProvider` so it is redactable.
> 
> The idea would be to avoid the need for a new credential vending system in core, putting most of the changes inside the Iceberg extension instead.

Are you proposing pushing the vending of credentials using an identity to the broker/overlord prior to task submission? I'd ideally like to propagate the auth context to the task and have it vend the credentials at runtime, not at submit time.

Currently, the way this works is:

- We extract user identity
- We pass to an Iceberg catalog client which uses this identity to then vend S3 credentials
- These S3 credentials are then attached to the subtasks that spawn from the driver task

jtuglu1:
@gianm any further thoughts on the above?

gianm:
> Are you proposing pushing the vending of credentials using an identity to the broker/overlord prior to task submission? I'd ideally like to propagate the auth context to the task and have it vend the credentials at runtime, not at submit time.

Yes, that's what I was proposing. There are a couple of reasons. First, because doing it at submit-time needs fewer changes to core. Second, because one of the proposed changes to core— the propagation of the user's own credentials between Druid services— is not something we've done before, and I feel this requires extra care from a security perspective. I was hoping to avoid the need for it completely.

Maybe a hybrid approach would work? We could introduce `scopeForUser` in core and run it at submit time. In your custom extension, rather than applying vended credentials at scope/submit time, you could use `scopeForUser` to embed the user's own credentials in the input source. We could add a `PasswordProvider` field to `IcebergInputSource` to support that. Then you could use them at runtime in the task to acquire vended credentials.

jtuglu1:
> Maybe a hybrid approach would work? We could introduce `scopeForUser` in core and run it at submit time. In your custom extension, rather than applying vended credentials at scope/submit time, you could use `scopeForUser` to embed the user's own credentials in the input source. We could add a `PasswordProvider` field to `IcebergInputSource` to support that. Then you could use them at runtime in the task to acquire vended credentials.

I think this makes sense – my only concern is, for example, being able to reliably "touch-up" these inputSource fields in arbitrary specs (both on overlord and on broker) to apply scopeForUser(). For example, you can have an MSQ query that queries multiple iceberg tables and joins against a Druid table.

In general, having the identity at the task-level also opens up ability for Druid to do auth with other resources that tasks might read from (e.g. Kafka), so generally want to lean in direction of having identity/credentials exposable at the task-level (not just seeing the end credentials for whatever is needed).

jtuglu1:
> > Maybe a hybrid approach would work? We could introduce `scopeForUser` in core and run it at submit time. In your custom extension, rather than applying vended credentials at scope/submit time, you could use `scopeForUser` to embed the user's own credentials in the input source. We could add a `PasswordProvider` field to `IcebergInputSource` to support that. Then you could use them at runtime in the task to acquire vended credentials.
> 
> I think this makes sense – my only concern is, for example, being able to reliably "touch-up" these inputSource fields in arbitrary specs (both on overlord and on broker) to apply scopeForUser(). For example, you can have an MSQ query that queries multiple iceberg tables and joins against a Druid table.
> 
> In general, having the identity at the task-level also opens up ability for Druid to do auth with other resources that tasks might read from (e.g. Kafka), so generally want to lean in direction of having identity/credentials exposable at the task-level (not just seeing the end credentials for whatever is needed).



> > Maybe a hybrid approach would work? We could introduce `scopeForUser` in core and run it at submit time. In your custom extension, rather than applying vended credentials at scope/submit time, you could use `scopeForUser` to embed the user's own credentials in the input source. We could add a `PasswordProvider` field to `IcebergInputSource` to support that. Then you could use them at runtime in the task to acquire vended credentials.
> 
> I think this makes sense – my only concern is, for example, being able to reliably "touch-up" these inputSource fields in arbitrary specs (both on overlord and on broker) to apply scopeForUser(). For example, you can have an MSQ query that queries multiple iceberg tables and joins against a Druid table.
> 
> In general, having the identity at the task-level also opens up ability for Druid to do auth with other resources that tasks might read from (e.g. Kafka), so generally want to lean in direction of having identity/credentials exposable at the task-level (not just seeing the end credentials for whatever is needed).



> > Maybe a hybrid approach would work? We could introduce `scopeForUser` in core and run it at submit time. In your custom extension, rather than applying vended credentials at scope/submit time, you could use `scopeForUser` to embed the user's own credentials in the input source. We could add a `PasswordProvider` field to `IcebergInputSource` to support that. Then you could use them at runtime in the task to acquire vended credentials.
> 
> I think this makes sense – my only concern is, for example, being able to reliably "touch-up" these inputSource fields in arbitrary specs (both on overlord and on broker) to apply scopeForUser(). For example, you can have an MSQ query that queries multiple iceberg tables and joins against a Druid table.
> 
> In general, having the identity at the task-level also opens up ability for Druid to do auth with other resources that tasks might read from (e.g. Kafka), so generally want to lean in direction of having identity/credentials exposable at the task-level (not just seeing the end credentials for whatever is needed).

Another thing to add here is I'd like to (if possible) avoid putting the burden on the caller for "injecting" this user identity. IMO, if we can always do it at the overlord level (and put the burden on the task type implementor to actually do something with a valid, provided identity), that should be more maintainable. Especially in a world with multiple input sources and IMO code related to input sources (e.g. Kafka, Iceberg, Delta) should not be involved on anything but a supervisor thread and the task process itself. I think the (one?) exception to this might be with Dart on broker-only (but I'm not sure if accepts reading an external inputSource on broker-only w/o creating a task).


This pattern is seen in other master/worker ETL systems like Spark, Ray, Trino, etc.

jtuglu1:
@gianm @clintropolis any more thoughts here?

cecemei:
@jtuglu1 looks like there's a merge conflict that needs to be resolved.

jtuglu1:
@cecemei I'm unsure this will get into v37 as we have not settled on design. I've removed this from v37 milestone.

multi-stage-query/src/main/java/org/apache/druid/msq/sql/MSQTaskQueryMaker.java
```diff
@@ -171,7 +172,13 @@ public QueryResponse<Object[]> runQuery(final DruidQuery druidQuery)
...
+      final Map<String, String> extraHeaders = CollectionUtils.mapValues(plannerContext.getAuthenticationResult().getContext(), String::valueOf);
```
FrankChen021, line 177: [P2] Do not forward authenticator context as raw headers

AuthenticationResult.getContext() is authenticator-defined metadata, not a map of HTTP headers. Forwarding every entry directly into the Broker-to-Overlord request can leak arbitrary auth context to the internal RPC path and can also break MSQ task submission when a provider stores non-header-safe keys or values. Please propagate only an explicit allowlisted header/token shape or use a typed inter-service auth-context carrier.

jtuglu1:
> We could add a PasswordProvider field to IcebergInputSource to support that. Then you could use them at runtime in the task to acquire vended credentials.

@gianm one other risk here is that this runs the risk of credential injection. For example, it does not guarantee that the identity of the user POSTing the task is the same as that which is embedded in the field in the spec. By making this a public, settable field by the user, an authorized user (who can WRITE to that datasource) may be able to submit a spec containing the credentials of another user (who has access to that input source). There is no mechanism to check whether the identity of the user submitting the task is the same as the identity accessing the resources (this could be enforced client-side but IMO Druid should be able to propagate and handle setting of this field). Users should not need to implement any of this logic client-side IMO. All they should need to worry about is translating a Druid identity into something their input source can understand/auth against.

Whether we want to support the odd case where a user might want to support calling with an inputSource-specific password provider is a valid consideration here, but I think we need to have some mechanism in-place to ensure that we do the natural thing which users expect by default (take my identity and pass it through to the task, using my identity to auth against any external resources).

indexing-service/src/main/java/org/apache/druid/indexing/common/task/batch/parallel/ParallelIndexSupervisorTask.java
```diff
@@ -540,6 +541,11 @@ public TaskStatus runTask(TaskToolbox toolbox) throws Exception
       initializeSubTaskCleaner();
       this.toolbox = toolbox;
 
+      // Propagate auth context to input source for credential vending (e.g., Iceberg REST Catalog)
+      if (baseInputSource instanceof TaskAuthContextAware) {
```
FrankChen021, line 545: [P1] Propagate auth context to ordinary index tasks

This is the only ingestion-framework hook added for TaskAuthContextAware, but IcebergInputSource is also usable by ordinary index tasks. IndexTask obtains the source at IndexTask.java:462 and AbstractBatchIndexTask.inputSourceReader calls it without setTaskAuthContext, so a normal index task never forwards its context to RestIcebergCatalog; catalog authentication and credential vending then fail. Add the propagation to the common batch path or every input-source execution path.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Reviewed 33 of 33 changed files in the full merge-base diff. The full diff was audited because this was a new review.

---

PR #19247, opened 2026-04-01
feat: Add additional context about the authorization request that can be passed to Authorizer

maytasm:
Add additional context about the authorization request that can be passed to Authorizer

### Description
Currently, the `Authorizer#authorize` interface only accepts `AuthenticationResult`, `Resource`, and `Action`, which provides no context about where an authorization request originated. As a result, authorization requests for the same resource from different code paths look identical. For example, an authorization request for a resource initiated by a user query is indistinguishable from one initiated by a `SystemSchema` request triggered by the UI.

This becomes a problem when analyzing authorization denials. If a user wants to investigate denials for their datasource, those denials are aggregated with denials from `SystemSchema`/`InformationSchema`, making it difficult to distinguish genuine unauthorized access attempts from routine internal checks. When a user refreshes the UI, it iterates over all datasources and issues an internal authorization check for each one; this happens in many places, such as the Supervisors tab, Query tab, and others. As a result, metrics emitted by the `Authorizer` (such as authorization outcomes) become noisy, since denials from internal UI-driven checks are mixed together with denials from actual user queries.

This change adds an optional `Map<String, Object>` `context` parameter to `Authorizer#authorize`. Callers such as `InformationSchema` and `SystemSchema` populate this context map with a `callerPath` key to indicate the source of the authorization request. This context can then be included in the metrics emitted by `Authorizer#authorize`, allowing users to filter out internal authorization checks from user-initiated ones.

 * `server/src/main/java/org/apache/druid/server/security/Authorizer.java`
 * `server/src/main/java/org/apache/druid/server/security/AuthorizationUtils.java`
 * `sql/src/main/java/org/apache/druid/sql/calcite/schema/InformationSchema.java`
 * `sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java`

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
+++ b/sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
@@ -121,6 +121,14 @@ public class SystemSchema extends AbstractSchema
           segment.getDataSource())
       );
 
+  /**
+   * Context map passed to {@link AuthorizationUtils} methods to indicate that authorization
+   * is being performed from the SystemSchema.
+   */
+  static final String AUTHORIZATION_CONTEXT_CALLER_PATH_VALUE = SystemSchema.class.getSimpleName();
+  private static final Map<String, Object> SYSTEM_SCHEMA_AUTHORIZATION_CONTEXT =
+      ImmutableMap.of(AuthorizationUtils.AUTHORIZATION_CONTEXT_CALLER_PATH_CONTEXT_KEY, AUTHORIZATION_CONTEXT_CALLER_PATH_VALUE);
+
   private static final long REPLICATION_FACTOR_UNKNOWN = -1L;
 
   /**
@@ -486,7 +494,8 @@ public class SystemSchema extends AbstractSchema
               authenticationResult,
               () -> it,
               SEGMENT_STATUS_IN_CLUSTER_RA_GENERATOR,
-              authorizerMapper
+              authorizerMapper,
+              SYSTEM_SCHEMA_AUTHORIZATION_CONTEXT
           );
       return authorizedSegments.iterator();
     }
@@ -511,7 +520,8 @@ public class SystemSchema extends AbstractSchema
               authenticationResult,
               () -> availableSegmentEntries,
               raGenerator,
-              authorizerMapper
+              authorizerMapper,
+              SYSTEM_SCHEMA_AUTHORIZATION_CONTEXT
           );
 
       return authorizedSegments.iterator();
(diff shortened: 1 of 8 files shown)


jtuglu1 (approved):
Given we have done review internally, LGTM once PR title is fixed and tests are passing.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -121,6 +121,14 @@ public class SystemSchema extends AbstractSchema
           segment.getDataSource())
       );
 
+  /**
+   * Context map passed to {@link AuthorizationUtils} methods to indicate that authorization
+   * is being performed from the SystemSchema.
+   */
+  static final String AUTHORIZATION_CONTEXT_CALLER_PATH_VALUE = SystemSchema.class.getSimpleName();
+  private static final Map<String, Object> SYSTEM_SCHEMA_AUTHORIZATION_CONTEXT =
```
FrankChen021, line 129: [P2] Pass SystemSchema context to server state check

SYSTEM_SCHEMA_AUTHORIZATION_CONTEXT is threaded through the SystemSchema datasource filters, but checkStateReadAccessForServers still calls the three-argument authorizeAllResourceActions, so querying sys.servers performs its STATE READ authorization with an empty context. That leaves this SystemSchema-originated denial indistinguishable from ordinary state reads, undermining the new callerPath contract for one of the UI/system-schema paths; pass SYSTEM_SCHEMA_AUTHORIZATION_CONTEXT into that helper call too.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemSchema.java
```diff
@@ -121,6 +121,14 @@ public class SystemSchema extends AbstractSchema
...
+      ImmutableMap.of(AuthorizationUtils.AUTHORIZATION_CONTEXT_CALLER_PATH_CONTEXT_KEY, AUTHORIZATION_CONTEXT_CALLER_PATH_VALUE);
```
FrankChen021, line 130: [P2] Pass context through server state checks

checkStateReadAccessForServers still calls the four-argument authorizeAllResourceActions overload without SYSTEM_SCHEMA_AUTHORIZATION_CONTEXT. The servers and server_segments scans use that helper, so their STATE READ authorization omits the new callerPath even though row filtering passes it. Thread the context through this helper and its callers so custom authorizers receive consistent context.

PR #19252, opened 2026-04-01
docs: Recommend Overlord-based auto-compaction and mark useIncrementalCache production ready

cecemei:
## Description
  - **Rename `CoordinatorRunStats` to `DruidRunStats`**
  - Some changes in automatic compaction documentation and default configuration settings.

### Automatic Compaction
  - **Overlord-based compaction supervisors are now the recommended and default approach** for automatic compaction. This method provides better reactivity, MSQ task engine support, and easier management through the supervisor framework. Coordinator-based auto-compaction remains available as an alternative.
  - Automatic compaction tasks are now prefixed with `auto` instead of `coordinator-issued`.

  ### Incremental cache 
  - **Incremental segment metadata cache (`useIncrementalCache`) is no longer experimental and defaults to `ifSynced`**. 
 

<hr>

This PR has:

- [x] been self-reviewed.


capistrant:
The default for useSupervisors should be true in the cluster compaction config if we are recommending it going forward. that way all new deploys will get the recommended config

cecemei:
> useSupervisors

updated default to true, PTAL!

docs/configuration/index.md
```diff
@@ -800,14 +800,14 @@ These Coordinator static configurations can be defined in the `coordinator/runti
...
-|`druid.manager.rules.pollDuration`|The duration between polls the Coordinator does for updates to the set of active rules. Generally defines the amount of lag time it can take for the Coordinator to notice rules.|`PT1M`|
-|`druid.manager.rules.defaultRule`|The default rule for the cluster|`_default`|
-|`druid.manager.rules.alertThreshold`|The duration after a failed poll upon which an alert should be emitted.|`PT10M`|
+|Property|Description| Default    |
```
clintropolis: what's up with all the unrelated formatting changes?

server/src/main/java/org/apache/druid/server/coordinator/stats/DruidRunStats.java
```diff
@@ -35,44 +35,44 @@
  * runtime of a single coordinator duty.
  */
 @ThreadSafe
-public class CoordinatorRunStats
+public class DruidRunStats
```
clintropolis: I can't help but wonder if there is a better name for this, maybe `DutyRunStats` or something that means 'thing to collect stuff to emit later from regularly occurring internal chores'? I guess 'duty' isn't quite right because that is basically only used to refer to periodic coordinator tasks, not supervisor stuff. The javadoc still only mentions coordinator run/duties, which should be fixed.

Also, it is kind of weird having the name `DruidRunStats` but the thing it has is still called `CoordinatorStat`, it seems like that naming should be changed to reflect the change here.

Stepping back, what exactly is the motivation for renaming, i guess that compaction uses and runs as a supervisor now so it isn't really specific to the coordinator? While this is still used quite heavily by all of the things the coordinator does, it seems reasonable to give it a more generic name of some sort, I was just wondering if this one is a bit too generic, but maybe is fine too as long as the javadoc clarifies its purpose?

Some addtional thoughts: I believe some of us would like to eventually merge the coordinator and overlord into a single service. Since they both now basically need a heavy segment timeline and so have similar footprint requirements, there aren't a lot of compelling reasons to keep them separate anymore. In my mind 'coordinator' would be the remaining service, with all of the overlords functionality merged into it (though this hasn't really been discussed so maybe other people have other opinions), so if that were true then this would basically become something only used by the coordinator again heh. There is a lot of work to do for something like this, so it is not really a short term goal afaik and needs more official discussion at some point, just adding it here for additional stuff to think about.

docs/data-management/automatic-compaction.md
```diff
@@ -83,9 +80,135 @@ maximize performance and minimize disk usage of the `compact` tasks launched by
...
+:::
+
+The recommended way to run automatic compaction is using compaction supervisors on the Overlord. Compaction supervisors provide the following benefits:
+
+* Can use the supervisor framework to get information about the auto-compaction, such as status or state
+* More easily suspend or resume compaction for a datasource
+* Can use either the native compaction engine or the [MSQ task engine](#use-msq-for-auto-compaction)
+* More reactive and submits tasks as soon as a compaction slot is available
+* Tracked compaction task status to avoid re-compacting an interval repeatedly
+* Uses new Indexing State Fingerprinting mechanisms to store less data per segment in metadata storage
```
clintropolis: i know this isn't new, but by default we still store compaction state afaict (`ClusterCompactionConfig.storeCompactionStatePerSegment` defaults to true), so this should be reworded to be like 'can be configured to store only fingerprints' or whatever

docs/data-management/cascading-reindexing.md
```diff
@@ -36,15 +36,6 @@ For example, you might want to:
...
 
-## Prerequisites
-
-Before using cascading reindexing, ensure your cluster meets the following requirements:
-
-- **Compaction supervisors enabled**: Set `useSupervisors` to `true` in the [compaction dynamic config](../api-reference/automatic-compaction-api.md#update-cluster-level-compaction-config).
-- **MSQ compaction engine**: Set `engine` to `msq` in the compaction dynamic config or in the supervisor spec.
-- **Incremental segment metadata caching**: Set `druid.manager.segments.useIncrementalCache` to `always` or `ifSynced` in your Overlord and Coordinator runtime properties. See [Segment metadata caching](../configuration/index.md#metadata-retrieval).
-- **At least two compaction task slots**: The MSQ task engine requires at least two tasks (one controller, one worker).
-
```
clintropolis, line 47: i think we need to leave part of this, no? the default engine is still native, so people still need to set engine to msq, and you need 2 compaction slots since its using msq engine

server/src/main/java/org/apache/druid/server/coordinator/stats/DruidRunStats.java
cecemei: right it'd still be under coordinator, same as some compaction config.

this is probably not a good idea to change right here, so i'm going to revert this change, and delay it to druid 38

docs/data-management/cascading-reindexing.md
cecemei: added the MSQ related change back, originally i was thinking making MSQ the default engine but caused some circular dependency in test, next release ...

docs/configuration/index.md
cecemei: intellij, anyway the formatting change has been reverted

docs/data-management/automatic-compaction.md
cecemei: updated wording and also mentioned minor compaction is supported here

server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
```diff
@@ -94,7 +94,7 @@ public class CompactSegments implements CoordinatorCustomDuty
 
   private static final Logger LOG = new Logger(CompactSegments.class);
 
-  private static final String TASK_ID_PREFIX = "coordinator-issued";
+  private static final String TASK_ID_PREFIX = "auto";
```
clintropolis, line 97: should we also update `KillUnusedSegments`/`UnusedSegmentsKiller` prefix to match to be more consistent?

Also might be nice to mention these cosmetic changes in release notes

cecemei: updated release notes, i can open another pr to update KillUnusedSegments/UnusedSegmentsKiller as a follow up.

clintropolis (approved):

PR #19253, opened 2026-04-01
feat: Server view improvements for historical detail (decommissioning, turbo segment loading, cloning)

capistrant:
### Description

Broken out from #19250 due to that PR getting too large and difficult to review. This PR focuses on just the services view where we add information in the "Detail" column that will help operators know about any special state that their historical services are in

image

In the image you can see there are new details for:
* DECOMMISSIONING
* TURBO SEGMENT LOADING
* cloning

The creating of this PR flagged a bug in `CloneHistoricals.java`. When going from N to 0 clone mappings; when N > 0, you end up not resetting the clone status manager and that means the api is inaccurately reporting clones when there are no mappings anymore. This PR fixes that issue. Note that the bug appears to be a reporting issue, the coordinator is not still treating the removed mappings as clones.

#### Release note

Enhance the services view in the web console by adding information to the "Detail" column for Historical services who are:
* cloning from another historical
* in turbo loading mode
* in decommissioning mode

<hr>

 * `CloneHistoricals`

<hr>

This PR has:

- [x] been self-reviewed.


web-console/src/views/services-view/services-view.tsx
```diff
@@ -366,6 +483,40 @@ ORDER BY
...
+        if (capabilities.hasCoordinatorAccess() && visibleColumns.shown('Detail')) {
+          auxiliaryQueries.push(async (servicesWithAuxiliaryInfo, signal) => {
+            const [cloneStatusResp, configResp] = await Promise.all([
+              getApiArrayFromKey<CloneStatusInfo & { targetServer: string }>(
+                '/druid/coordinator/v1/config/cloneStatus',
+                'cloneStatus',
+                signal,
+              ).catch(() => [] as (CloneStatusInfo & { targetServer: string })[]),
+              Api.instance
+                .get('/druid/coordinator/v1/config', { signal })
```
gianm: I think this should be `Api.instance.get<CoordinatorDynamicConfig>` in order to pick up the types for the Coordinator config.

web-console/src/views/services-view/services-view.tsx
```diff
@@ -220,6 +243,100 @@ function aggregateLoadQueueInfos(loadQueueInfos: LoadQueueInfo[]): LoadQueueInfo
...
+    case 'overlord':
+      return <>{is_leader === 1 ? 'Leader' : ''}</>;
+
+    case 'historical': {
+      const loadQueueInfo = loadQueueInfoContext[service];
+      const cloneInfo = cloneStatusContext[service];
+
+      const parts: string[] = [];
+      if (serverModeInfo.decommissioningNodes.has(service)) {
+        parts.push('DECOMMISSIONING');
```
gianm: Seems needlessly screamy. How about "Decommissioning"? It would better match the casing of the existing things that go into this line.

web-console/src/views/services-view/services-view.tsx
```diff
@@ -220,6 +243,100 @@ function aggregateLoadQueueInfos(loadQueueInfos: LoadQueueInfo[]): LoadQueueInfo
...
+      }
+      if (serverModeInfo.turboLoadingNodes.has(service)) {
+        parts.push('TURBO SEGMENT LOADING');
```
gianm: Similar comment here, how about "Turbo loading"? (also suggesting cutting the word "segment" for length.)

web-console/src/views/services-view/services-view.tsx
```diff
@@ -220,6 +243,100 @@ function aggregateLoadQueueInfos(loadQueueInfos: LoadQueueInfo[]): LoadQueueInfo
...
+    default:
+      return null;
+  }
+}
+
+interface AggregatedDetailCellProps {
+  subRows: { _original: ServiceResultRow }[];
+}
+
+function AggregatedDetailCell({ subRows }: AggregatedDetailCellProps) {
```
gianm, line 327: This could include aggregated info on decommissioning and turbo loading servers (e.g. count them and do "X decommissioning" and/or "Y turbo loading" if any are in those states).

web-console/src/views/services-view/services-view.tsx
```diff
@@ -366,6 +483,40 @@ ORDER BY
...
+            const [cloneStatusResp, configResp] = await Promise.all([
+              getApiArrayFromKey<CloneStatusInfo & { targetServer: string }>(
+                '/druid/coordinator/v1/config/cloneStatus',
+                'cloneStatus',
+                signal,
+              ).catch(() => [] as (CloneStatusInfo & { targetServer: string })[]),
+              Api.instance
+                .get('/druid/coordinator/v1/config', { signal })
+                .then(r => r.data)
+                .catch(() => null),
```
gianm: When the load queue infos fail to fetch, there's a toast that says there was an error fetching them. Consider that here too. (and also for a failure on `/druid/coordinator/v1/config/cloneStatus`)

web-console/src/views/services-view/services-view.tsx
```diff
@@ -366,6 +483,40 @@ ORDER BY
           });
         }
 
+        if (capabilities.hasCoordinatorAccess() && visibleColumns.shown('Detail')) {
+          auxiliaryQueries.push(async (servicesWithAuxiliaryInfo, signal) => {
+            const [cloneStatusResp, configResp] = await Promise.all([
+              getApiArrayFromKey<CloneStatusInfo & { targetServer: string }>(
```
gianm: why not add `targetServer` to `CloneStatusInfo`?

web-console/src/views/services-view/services-view.tsx
capistrant: got rid of all caps here and below

web-console/src/views/services-view/services-view.tsx
capistrant: ok added aggregation support for turbo and decom as well as cloning. I split cloning agg into 3 groups. 

cloning, cloned, and clone error (right now clone error is just pinned on having an invalid source server defined)

web-console/src/views/services-view/services-view.tsx
capistrant: good point

capistrant:
I'm convinced that I'm cursed with the CI failure. or something is sticky somehow. I do not see any connection whatsoever to my code and the constant failures. This is a test class that I tried to fix in a recent commit but I guess maybe did not succeed

web-console/src/views/services-view/services-view.tsx
```diff
@@ -366,6 +524,54 @@ ORDER BY
...
+              ).catch(() => {
+                AppToaster.show({
+                  icon: IconNames.ERROR,
+                  intent: Intent.DANGER,
+                  message: 'There was an error getting the clone status map',
+                });
+                return [] as CloneStatusInfo[];
+              }),
+              Api.instance
+                .get<CoordinatorDynamicConfig>('/druid/coordinator/v1/config', { signal })
```
FrankChen021, line 543: [P2] Avoid requiring config permission for services view

The services view is available when hasCoordinatorAccess() is true, which only proves the user can reach coordinator state endpoints. This new auxiliary request calls /druid/coordinator/v1/config, which is protected by ConfigResourceFilter, so users with STATE read access but without CONFIG read access will now get a failure toast on every Services refresh/open even though the table itself can load. The decommissioning/turbo details should be fetched through a state-authorized endpoint or the 403 should be treated as unavailable without surfacing an error.

gianm (approved):
LGTM. Good luck with CI.

With regard to this comment: https://github.com/apache/druid/pull/19253#discussion_r3141492431 it should be quite rare for someone to have STATE permission but not CONFIG. I think it's ok to not worry about it, and have a toast pop up in that case.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

- processing/src/main/java/org/apache/druid/segment/file/SegmentFileBuilderV10.java:541 - P2 Validate delegated writer byte count on close. The delegated addWithChannel path only rejects over-writes; close() accepts the temp file even when bytesWritten is less than the declared size. This differs from the non-delegated writer, which throws if the caller closes after writing fewer bytes than promised. Nested serializer writes can now silently produce truncated internal files and only fail later during segment read/deserialization. Add the same bytesWritten == size check before moving the delegate to completedDelegates.

---

capistrant:
ok I ended up missing places in the tests where I had to remove the cloneservers mock. I plan to revisit the lookups manager test though, seems sad it was using globally shared state.

PR #18033, opened 2025-05-26
Add QOS filtering of overlord requests

cryptoe:
Adds QOS filtering for the overlord so that health check threads are not blocked.

PR #19271, opened 2026-04-07
minor: Add QOS filtering for coordinator endpoints

cryptoe:
Add QoS filtering to the Coordinator to protect health check endpoints under load
                                                                                                                                                                                                                                                                                                                                                                                                                                           
The Coordinator had no request throttling, meaning a burst of slow API requests (segment load status, datasource queries, compaction configs, etc.) could exhaust the Jetty thread pool and make /status/health and /status/ready unresponsive. This is the same problem the Overlord solved in #18033 for its action endpoints.


## Release notes

Add QoS filtering for Coordinator endpoints
* Added configurable QoS filtering to the Coordinator to prevent a burst of slow API requests from exhausting the thread pool and blocking health checks . 
* New config is added: `druid.coordinator.server.maxConcurrentRequests` `(default: max(1, max(serverHttpNumThreads - 4, serverHttpNumThreads * 0.8)), set <= 0 to disable)`. 
* Leader-status endpoints are excluded from the cap so load balancers/monitors stay unaffected.


FrankChen021:

FrankChen021 (commented):

---

services/src/main/java/org/apache/druid/cli/CliCoordinator.java
```diff
@@ -229,6 +231,19 @@ public void configure(Binder binder)
...
 
+            // QoS filtering to prevent heavy coordinator API requests from starving health check endpoints.
+            // Set druid.coordinator.server.maxConcurrentRequests=-1 to disable.
+            final int serverHttpNumThreads = ServerConfig.getNumThreadsFromProperties(properties);
+            final int maxConcurrentRequests = properties.containsKey("druid.coordinator.server.maxConcurrentRequests")
+                    ? Integer.parseInt(properties.getProperty("druid.coordinator.server.maxConcurrentRequests"))
+                    : ServerConfig.getDefaultMaxConcurrentRequests(serverHttpNumThreads);
+            if (maxConcurrentRequests > 0) {
+              log.info("Coordinator QoS filtering enabled. Max concurrent requests: [%d]", maxConcurrentRequests);
+              JettyBindings.addQosFilter(binder, new String[]{"/druid/coordinator/v1/*", "/druid-internal/*"}, maxConcurrentRequests);
```
capistrant: There are a bunch of light service status endpoints under `/druid/coordinator/v1/*`. found here: https://druid.apache.org/docs/latest/api-reference/service-status-api ... do we really want those blanketed under Qos? I think that `isLeader` could specifically be sensitive to Qos restrictions since monitors/loadBalancers could rely on it. also plain `leader` maybe

docs/configuration/index.md
```diff
@@ -746,6 +746,7 @@ These Coordinator static configurations can be defined in the `coordinator/runti
 |Property|Description|Default|
 |--------|-----------|-------|
 |`druid.coordinator.period`|The run period for the Coordinator. The Coordinator operates by maintaining the current state of the world in memory and periodically looking at the set of "used" segments and segments being served to make decisions about whether any changes need to be made to the data topology. This property sets the delay between each of these runs.|`PT60S`|
+|`druid.coordinator.server.maxConcurrentRequests`|Maximum number of concurrent requests to coordinator API endpoints (`/druid/coordinator/v1/*`, `/druid-internal/*`) that the Coordinator will process simultaneously. This prevents thread exhaustion while preserving access to health check endpoints. Set to `-1` to disable quality of service filtering entirely. If not specified, defaults to `max(1, max(serverHttpNumThreads - 4, serverHttpNumThreads * 0.8))`.|`max(1, max(serverHttpNumThreads - 4, serverHttpNumThreads * 0.8))`|
```
capistrant: looks like the code actually gates Qos on `> 0`. Not that it is a big deal but either the doc or the code should reflect this. I think it would be the doc that just says set to `<= 0` to disable?

services/src/main/java/org/apache/druid/cli/CliCoordinator.java
capistrant: could consider writing a CliCoordinatorTest and testing that this binding is added when desired

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1347,6 +1460,11 @@ public void tryInit()
       try {
         recordSupplier = setupRecordSupplier();
 
+        // Initialize bounded partitions BEFORE first run
+        if (ioConfig.isBounded()) {
```
FrankChen021, line 1541: [P1] Bounded-mode calls do not compile

This introduces calls to `ioConfig.isBounded()` and `ioConfig.getBoundedStreamConfig()` plus references to `BoundedStreamConfig`, but `SeekableStreamSupervisorIOConfig` does not define those methods and no `BoundedStreamConfig` type exists in the worktree. The indexing-service module will fail to compile as soon as this file is built; add the config/type and accessors, or remove these bounded-mode paths.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 3 |
| P2 | 0 |
| P3 | 0 |
| Total | 3 |

Found compile blockers in the incremental `SeekableStreamSupervisor` changes. The coordinator QoS wiring, docs, shared defaults, and Overlord refactor were also inspected, but the PR cannot be validated while these compile errors remain.

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -3156,11 +3311,64 @@ private boolean updatePartitionDataFromStream()
   /**
```
FrankChen021, line 3388: [P1] Public hook changes break protected overrides

This changes `getLatestSequencesFromStream` from protected to public; similar public changes were made to `updatePartitionLagFromStream`, `getOffsetsFromMetadataStorage`, and `createDataSourceMetaDataForReset`. Existing overrides in Kafka, Kinesis, and test supervisors are still protected, which Java rejects as reducing visibility. Keep these hooks protected or update every override to public.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4692,6 +5180,25 @@ protected abstract List<SeekableStreamIndexTask<PartitionIdType, SequenceOffsetT
...
   ) throws JsonProcessingException;
 
+  /**
+   * Converts a string representation of a partition ID to the typed partition ID.
+   * Used for deserializing bounded stream config where partition keys come as strings.
+   *
+   * @param partitionIdString string representation of partition ID
+   * @return typed partition ID
+   */
+  protected abstract PartitionIdType createPartitionIdFromString(String partitionIdString);
```
FrankChen021, line 5267: [P1] New abstract hooks are not implemented

The base class now declares `createPartitionIdFromString`, `createSequenceOffsetFromObject`, `isOffsetAtOrBeyond`, and `isEndOffsetExclusive` as abstract, but the concrete Kafka/Kinesis supervisors and the seekable-stream test supervisors do not implement them. Those classes will no longer compile unless these hooks get default implementations or all subclasses are updated.

services/src/main/java/org/apache/druid/cli/CliCoordinator.java
cryptoe: Done both. Added a PathExclusionQOS filter which felt better than adding a complicated regex matching.

cryptoe:
@capistrant Thank you for taking a look. Addressed your feedback. Please take a look

docs/configuration/index.md
```diff
@@ -739,7 +739,7 @@ These Coordinator static configurations can be defined in the `coordinator/runti
...
 |`druid.coordinator.kill.maxSegments`|The number of unused segments to kill per kill task. This number must be greater than 0. This only applies when `druid.coordinator.kill.on=true`.|100|
 |`druid.coordinator.kill.maxInterval`|The largest interval, as an [ISO 8601 duration](https://en.wikipedia.org/wiki/ISO_8601#Durations), of segments to delete per kill task. Set to zero, e.g. `PT0S`, for unlimited. This only applies when `druid.coordinator.kill.on=true`.|`P30D`|
-
+|`druid.coordinator.server.maxConcurrentRequests`| Maximum number of concurrent requests to coordinator API endpoints (`/druid/coordinator/v1/*`, `/druid-internal/*`) that the Coordinator will process simultaneously. This prevents thread exhaustion while preserving access to health check endpoints. Set to any value <=0 like `-1` to disable quality of service filtering entirely. If not specified, defaults to `max(1, max(serverHttpNumThreads - 4, serverHttpNumThreads * 0.8))`.|`max(1, max(serverHttpNumThreads - 4, serverHttpNumThreads * 0.8))`|
```
capistrant: suggested re-word to more explicitly state config purpose and discourage operators from lowering value below default (this would be an exceptional circumstance IMO... more harm can be done than good in almost all cases if the value is lowered below default)

```suggestion
|`druid.coordinator.server.maxConcurrentRequests`|Maximum number of requests to non-exempt Coordinator API paths processed concurrently. Requests beyond this are queued briefly and rejected with HTTP 503 if no slot frees. Leadership endpoints (`/leader`, `/isLeader`) and non-matching paths such as `/status/*` are never throttled. This bounds Jetty thread-pool exhaustion from heavy or excessive calls (for example large `/loadstatus`, `/metadata/segments`, or `/intervals` scans). Note: this path also carries recurring internal control-plane traffic (segment handoff checks, historical bootstrap, lookup and rule sync), so setting this too low can throttle that traffic and degrade the cluster; lowering the default is discouraged. Set to any value `<= 0` (for example `-1`) to disable QoS filtering entirely.|`max(1, max(serverHttpNumThreads - 4, serverHttpNumThreads * 0.8))`|
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1456,12 +1456,12 @@ public int registerNewVersionOfPendingSegment(
...
-          taskAllocatorId,
-          supervisorId,
-          pendingSegmentRecord.getId(),
-          pendingSegmentRecord.getUpgradedFromSegmentId()
+              "Could not find any task matching taskAllocatorId[%s] in supervisor[%s] for upgraded pending segment[%s]"
+                      + " (upgradedFrom[%s]); it will not be re-announced until handoff.",
+              taskAllocatorId,
+              supervisorId,
+              pendingSegmentRecord.getId(),
+              pendingSegmentRecord.getUpgradedFromSegmentId()
```
capistrant, line 1464: unrelated reformatting. if it is fixing legit formatting issue I guess ok to leave, but it is unclear why this file would be being touched for this PR

capistrant (commented):
Thanks for adding leader exclusions. I think this is probably fine. I'm a little bit worried about the large surface area of endpoints under QoS since many of those are called by internal services and are important for cluster operation. I think this is fine for when operator uses default config, but I worry about operator thinking dropping the config makes things more safe and in reality makes it easier to degrade cluster health. I offered a suggestion for docs to try and discourage operators from tinkering beyond enable/disable of QoS for coordinator to mitigate this fear. let me know what you think

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
cryptoe: Its one of those I started changing this file but as I made progress, realized I donot need to touch this file. 
The formatting felt off and hence did a reformat of this piece only.

capistrant (approved):
looks like spellcheck will need to be addressed and can you also add release note section to the description to help D39 release note writing. Thanks!

PR #19286, opened 2026-04-09
feat: support scaling direction-aware cooldown for task auto-scalers

jtuglu1:
### Description

Adds support for configuring different cooldowns for scaling direction. While both scaling actions do cause temporary disruption to ingestion, scaling down can cause more disruption than scaling up due to having less resources than when you started to recover from lag. Therefore, to allow for aggressive scale up while having a more conservative scale-down approach, this adds configuration for cool down period for both directions. Cool down for a specific scaling direction is evaluated as follows:

1. `minScaleUpDelay`/`minScaleDownDelay`
2. `minTriggerScaleActionFrequencyMillis` (marked as deprecated)
3. The default of `600000`.

This also does the following:

1. Cleans up the core scaling logic to be more clear/readable
2. Adds a custom `ScaleActionSupplier` which makes the contract clear for auto-scalers who want to implement it.
5. Fixes a few notification bugs, namely:
  - Emitting/logging metrics with skip reasons irrespective of whether we are actually planning to scale or not (e.g. auto-scaler function returns value != -1, but that value is equal to the current task count). Now, we emit a required task count metric with a specific skip reason.

#### Release note

Adds support for configuring different cooldowns for scaling direction for streaming task auto-scalers.

<hr>

This PR has:

- [ ] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -523,25 +523,51 @@ public void handle()
...
+                  supervisorId,
+                  dataSource,
+                  desiredTaskCount,
+                  getActiveTaskGroupsCount()
+              );
+
+              if (desiredTaskCount > 0) {
+                emitter.emit(event.setDimension(
+                                      AUTOSCALER_SKIP_REASON_DIMENSION,
+                                      "minTriggerScaleActionFrequencyMillis not elapsed yet"
```
jtuglu1: opted not to modify the reason here

kfaraz:
@jtuglu1 , cost-based auto scaler already uses a `minScaleDownDelay` config for similar purposes. I feel we should simply expose that at the `AutoScalerConfig` instead of two new min trigger fields. (I never liked the name `minTriggerScaleActionFrequencyMillis` as it's verbose and still ambiguous.)

The functionality of the `minScaleDownDelay` may be modified as required by this PR.

cc: @Fly-Style

jtuglu1:
> @jtuglu1 , cost-based auto scaler already uses a `minScaleDownDelay` config for similar purposes. I feel we should simply expose that at the `AutoScalerConfig` instead of two new min trigger fields. (I never liked the name `minTriggerScaleActionFrequencyMillis` as it's verbose and still ambiguous.)
> 
> The functionality of the `minScaleDownDelay` may be modified as required by this PR.
> 
> cc: @Fly-Style

Sure, I always want to reduce config bloat. However, I'd prefer to settle on a conventional naming scheme if possible (e.g. `minScaleUpDelay`) to keep kicking the ambiguous can that is `minTriggerScaleActionFrequencyMillis` down the road. This way, we can deprecate that config and instead just treat it as an alias for both if unspecified in future releases.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -561,16 +562,18 @@ private Either<String, Boolean> validateMetricsForScaling(CostMetrics metrics)
   }
 
   /**
-   * Determines if a scale action is currently allowed based on the elapsed time
-   * since the last scale action and the configured minimum scale-down delay.
+   * Determines if a scale action in the given direction is currently allowed based on the elapsed time
+   * since the last scale action and the configured delay for that direction.
    */
-  private boolean isScaleActionAllowed()
+  private boolean isScaleActionAllowed(boolean isScaleUp)
```
maytasm: nit: having isScaleUp as a boolean parameter for this method is a little weird. Maybe refactoring to clear methods like isScaleUpActionAllowed and isScaleDownActionAllowed?

maytasm (approved):

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -499,11 +500,34 @@ public void handle()
...
+          }
+
           ServiceMetricEvent.Builder event = ServiceMetricEvent.builder()
                                                                .setDimension(DruidMetrics.SUPERVISOR_ID, supervisorId)
                                                                .setDimension(DruidMetrics.DATASOURCE, dataSource)
                                                                .setDimension(DruidMetrics.STREAM, getIoConfig().getStream());
+
+          // 1) This should already be handled by the auto-scaler implementation, but make sure we catch/record these for auditability
+          if (desiredTaskCount == currentTaskCount) {
+            log.info(
```
Fly-Style: It definitely deserves `warn` or even `error` log level; it detects an issue with the autoscaler implementation.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Looks like a perfect place to apply `ScaleDirection` enum

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScaler.java
```diff
@@ -83,23 +82,24 @@ public LagBasedAutoScaler(
...
-      }
-      catch (Exception ex) {
-        log.warn(ex, "Exception while computing desired task count for supervisor[%s]", spec.getId());
-      }
-      finally {
-        LOCK.unlock();
-      }
-      return desiredTaskCount;
-    };
+    LOCK.lock();
```
Fly-Style, line 88: I wonder if we really need a lock here, since the `STORE` into shared memory is eliminated and  `lagMetricsQueue` is captured into list in the callsite.

Fly-Style (commented):
Thank you for your contribution! The patch has nice direction, some polishing and we will be glad to merge it.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScaler.java
jtuglu1: I'm not planning to make any logical changes to the core lag auto-scaler logic in this PR (it's big enough as-is). I think a follow-up can be done to decide whether this lock is needed or not.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
jtuglu1, line 611: Looking at this again, I don't think this is the proper place for this logic TBH. IMO, an auto-scaler implementation should not be deciding this (the supervisor should and it already does in the main loop) since this scale up/down logic is shared across scalers. Will see what refactoring looks like.

I'm open to differing opinions, but generally an auto-scaler should take in some input and output the desired optimal task count. It shouldn't worry about skip reason (due to cooldown not elapsed, or min/max being hit, etc.) as these are all shared logical component between all scalers.

Fly-Style: > IMO, an auto-scaler implementation should not be deciding this (the supervisor should and it already does in the main loop) since this scale up/down logic is shared across scalers. 

I like SOLID, especially first letter, which is applicable here, I tend to agree with the general direction.
But, we should clearly define the _responsibility limits_ of this entity, because e.g, when I started doing CBA, I clearly saw that the supervisor delegates a **decision making** to the autoscaler. We may reduce the responsibility just to raw task count calculation, and move decision making logic to other entity.  That would be clean, IMO.

jtuglu1: > We may reduce the responsibility just to raw task count calculation, and move decision making logic to other entity. That would be clean, IMO.

Yeah, my feeling is CBA/Lag can be seen as functions which just return an optimal (or no) task count. The supervisor can handle what to do with this recommendation and emitting metrics on what it chose to do.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -62,6 +61,7 @@ public class CostBasedAutoScalerConfig implements AutoScalerConfig
   private final double idleWeight;
   private final boolean useTaskCountBoundaries;
   private final int highLagThreshold;
+  @Nullable private final Duration minScaleUpDelay;
```
jtuglu1: this should not be nullable

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -513,45 +537,59 @@ public void handle()
...
+          final ScaleDirection scaleDirection;
+          final long cooldownMillis;
+
+          if (desiredTaskCount > currentTaskCount) {
+            scaleDirection = ScaleDirection.SCALE_UP;
+            cooldownMillis = autoScalerConfig.getMinScaleUpDelay().getMillis();
+          } else { // desiredTaskCount < currentTaskCount
+            scaleDirection = ScaleDirection.SCALE_DOWN;
+            cooldownMillis = autoScalerConfig.getMinScaleDownDelay().getMillis();
+          }
```
Fly-Style, line 564: Unfortunately, the invariant with the fallback to `minTriggerScaleActionFrequencyMillis` is not present.

jtuglu1: These scale up/down fields default to `minTriggerScaleActionFrequencyMillis` if not already set, which itself defaults to 60s. It is there.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
jtuglu1, line 261: Looking more closely at this code – why are we bounding the output results of this result by the min/max task count? Shouldn't we bounding the results by the theoretical maximum (e.g. the partition count)?

Then, the calling code can validate whether min/max are breached (and emit/log a specific error indicating the scaler wanted to breach the min/max thresholds, alerting the operator)?

My issue with this is that we are removing observability into the case where the min/max task count is indeed misconfigured relative to what the autoscaler thinks is the optimal task count. Would it not make more sense to compute the valid task counts relative to the stream? Then use the min/max config bounds to flag these cases **outside** of this method?

IMO, this function should not care about the configured min/max task counts of the supervisor, and instead simply look at the stream data (lag, partitions, etc.) to derive the optimal task count. Whether that task count falls in the bounds of the min/max config can be enforced later, but we lose signal into what the original desired task count value was if we eagerly cap this value in `computeValidTaskCounts`.

Fly-Style: @jtuglu1, the config always was a source of truth, implying there is no sense in making a computation for a case that never happens. 

> My issue with this is that we are removing observability into the case where the min/max task count is indeed misconfigured relative to what the autoscaler thinks is the optimal task count. 

That's fair, but there was no intention to make autoscaler smarter than it is.

jtuglu1: Sure, we can leave for now I guess. I do think incorporating the partition count into the calculation would be useful for identifying misconfigured max task counts (providing a better estimate of what the true max should be).

Fly-Style: @jtuglu1 good target for a subclass of CBA with that ability?

jtuglu1: For context, before your [change](https://github.com/apache/druid/commit/5f37b7b5b74c98b341588b95daae21b1f38ca01d), the [line](https://github.com/apache/druid/blob/master/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScaler.java#L278) the `LagBasedAutoscaler` used to emit the required task count (even if this was above max task count config value configured in the spec). IMO, this is a regression and we should not be "clamping" required task counts to the maximum task count before emitting (e.g. follow prior behavior). The point of this metric (and the logs) are to allow users to understand the scaler required X but was clamped at the maximum task count (which they can already derive from the spec). They can then use this metric X to figure out what to set the maximum value to next.

jtuglu1: I'm shrinking the diff for now to get my changes in, then I'll follow-up with another PR that extracts the min/max logic and applies the other changes I was making in this PR.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -90,14 +91,15 @@ public CostBasedAutoScalerConfig(
                                    : DEFAULT_SCALE_ACTION_PERIOD_MILLIS;
     this.minTriggerScaleActionFrequencyMillis = Configs.valueOrDefault(
         minTriggerScaleActionFrequencyMillis,
-        DEFAULT_MIN_TRIGGER_SCALE_ACTION_FREQUENCY_MILLIS
+        DEFAULT_SCALE_ACTION_PERIOD_MILLIS / 2
```
Fly-Style: I realized that it should be equal to `DEFAULT_SCALE_ACTION_PERIOD_MILLIS`. That was my mistake.
In fact, here we're trying to scale twice as often as we are trying to compute new `taskCount`.

jtuglu1: Ok, the above change isn't changing anything from existing behavior (just rewriting to remove the `DEFAULT_MIN_TRIGGER_SCALE_ACTION_FREQUENCY_MILLIS` dependency). I can change that I guess

jtuglu1: See scaleUpDelay, I've already set that as `DEFAULT_SCALE_ACTION_PERIOD_MILLIS` and as `minTriggerScaleActionFrequencyMillis` is deprecated/unused this shouldn't be hit anyways.

Fly-Style (approved):
Really like the improvement, clear differentiation between scale-up and scale-down, while preserving old behaviour as a fallback. LGTM!

PR #16504, opened 2024-05-28
Upgrade calcite to 1.37.0

kgyrtkirk:
* contains #16503
* some minor api changes pair/entry
* some unnecessary aggregation was removed from a set of queries in `CalciteSubqueryTest`
* `AliasedOperatorConversion` was detecting `CHAR_LENGTH` as not a function ; I've removed the check
  * the field it was using doesn't look maintained that much
  * the `kind` is passed for the created `SqlFunction` so I don't think this check is actually needed
* some decoupled test cases become broken - will be fixed later
* some aggregate related changes
* `CalciteQueryTest#testExactCountDistinctWithFilter` is now executable

I recommend to turn on whitespace ignore while taking a look at the changes

PR #14510, opened 2023-06-30
Calcite upgrade 1.35

somu-imply:
Work in progress to update from Calcite 1.21 to Calcite 1.35. This currently runs using the rc-3 of Calcite 1.35 at https://dist.apache.org/repos/dist/dev/calcite/apache-calcite-1.35.0-rc3 or using a fork of Calcite built locally. To run this do the following

On a local calcite fork:
Clone the repo:
```
git clone https://github.com/gianm/calcite.git 
```
and build branch changes-for-druid-update
```
./gradlew clean && ./gradlew build publishToMavenLocal  -x test
```

On the artifact in https://dist.apache.org/repos/dist/dev/calcite/apache-calcite-1.35.0-rc3/, download the tarball, ensure you have Java 8 and gradle 7.6.1/7.6.2 and do the following
```
gradle build
gradle build publishToMavenLocal  -x test
```
This will publish a local version 1.35.0-SNAPSHOT.
You'll need to do a full commandline build to regenerate the SQL parser:
```
mvn clean compile -T8
```

**Update:** 1.35 is now available under maven central https://central.sonatype.com/artifact/org.apache.calcite/calcite-core/1.35.0. The pom.xml has been updated to pick up the calcite jars from maven central

This PR has:

- [ ] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] a release note entry in the PR description.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [ ] been tested in a test Druid cluster.

PR #19298, opened 2026-04-11
fix: INTERVAL n WEEK resolves to seven days instead of one hour

### LINKED ISSUE #18665: INTERVAL 1 WEEK in Druid is wrong interval
The `INTERVAL 1 WEEK` syntax in SQL is resolving to `PT1H`.

### Affected Version

Found in Apache Druid 33.0.0

### Description

Most easily detected by running this SQL query in Unified Console Workbench:
```
SELECT MILLIS_TO_TIMESTAMP(0) + INTERVAL 1 HOUR, 
  MILLIS_TO_TIMESTAMP(0) + INTERVAL 1 DAY, 
  MILLIS_TO_TIMESTAMP(0) + INTERVAL 7 DAY, 
  MILLIS_TO_TIMESTAMP(0) + INTERVAL 14 DAY, 
  MILLIS_TO_TIMESTAMP(0) + INTERVAL 1 WEEK,
  MILLIS_TO_TIMESTAMP(0) + INTERVAL 2 WEEK
```
Expected Results:
```
EXPR$0,EXPR$1,EXPR$2,EXPR$3,EXPR$4,EXPR$5
1970-01-01T01:00:00.000Z,1970-01-02T00:00:00.000Z,1970-01-08T00:00:00.000Z,1970-01-15T00:00:00.000Z,1970-01-08T00:00:00.000Z,1970-01-15T00:00:00.000Z
```
Actual Results:
```
EXPR$0,EXPR$1,EXPR$2,EXPR$3,EXPR$4,EXPR$5
1970-01-01T01:00:00.000Z,1970-01-02T00:00:00.000Z,1970-01-08T00:00:00.000Z,1970-01-15T00:00:00.000Z,1970-01-01T01:00:00.000Z,1970-01-01T02:00:00.000Z
```

barry3406:
Fixes #18665.

### Description

`INTERVAL '1' WEEK` (and the unquoted `INTERVAL 1 WEEK`) was resolving to one hour instead of seven days. The repro from #18665 makes the symptom obvious:

```sql
SELECT
  MILLIS_TO_TIMESTAMP(0) + INTERVAL 1 WEEK,
  MILLIS_TO_TIMESTAMP(0) + INTERVAL 2 WEEK
```

was returning `1970-01-01T01:00:00`, `1970-01-01T02:00:00` (one and two hours) instead of the expected `1970-01-08`, `1970-01-15`.

Per @gianm's review, `INTERVAL n QUARTER` is broken in Calcite 1.37.0 in the same way, so this PR also covers QUARTER.

#### Root cause

The bugs live in Calcite 1.37.0:

  - `SqlIntervalQualifier.evaluateIntervalLiteralAsWeek` packages a WEEK literal into a year-month shaped int array (`[sign, 0, week]`), even though `SqlIntervalQualifier.typeName()` reports a WEEK qualifier as `INTERVAL_DAY`. The downstream `SqlParserUtil.intervalToMillis` loop walks `ret[1..]` as `[day, hour, minute, second, ms]`, so the week count lands in the hour slot. `INTERVAL '1' WEEK` therefore becomes `1 * 3_600_000 ms = 1 hour`.
  - `SqlIntervalQualifier.evaluateIntervalLiteralAsQuarter` packages a QUARTER literal into the same year-month shaped array (`[sign, 0, quarter]`) without multiplying by three. `SqlParserUtil.intervalToMonths` reads `12 * 0 + 1 * quarter = quarter` months instead of `3 * quarter` months. `INTERVAL '1' QUARTER` therefore becomes one month.

Both bugs were fixed upstream as [CALCITE-6581](https://issues.apache.org/jira/browse/CALCITE-6581) in Calcite 1.38. Druid is still on Calcite 1.37, so we work around them inside the planner instead of upgrading Calcite in this PR.

#### Fixed the bug ...

Added `SqlIntervalWeekRewriteShuttle`, an `SqlShuttle` that walks the parsed `SqlNode` tree before validation and rewrites the affected parsed forms into equivalent intervals in a unit Calcite 1.37 handles correctly:

  - `INTERVAL '1' WEEK` parses to a `SqlIntervalLiteral` with a WEEK qualifier. The shuttle replaces it with a `SqlIntervalLiteral` whose value has been multiplied by seven and whose qualifier is now DAY.
  - `INTERVAL 1 WEEK` parses to `SqlStdOperatorTable.INTERVAL.createCall(pos, n, qualifier)`. The shuttle rewrites the call so the numeric operand becomes `MULTIPLY(n, 7)` and the qualifier becomes DAY.
  - `INTERVAL '1' QUARTER` and `INTERVAL 1 QUARTER` are rewritten the same way, using MONTH and a factor of three.

`WEEK(SUNDAY)..WEEK(SATURDAY)` and `ISOWEEK` qualifiers carry a non-null `timeFrameName` and are handled differently in Calcite, so the shuttle leaves them alone and only touches the bare `WEEK` / `QUARTER` forms that are broken in 1.37.

The shuttle is invoked in `DruidPlanner.validate` right before `rewriteParameters`, alongside the existing `SqlParameterizerShuttle` pass. It runs once per validate call and is a no-op for any tree that does not contain a WEEK or QUARTER interval.

This shuttle is a temporary workaround and should be removed when Druid upgrades to Calcite 1.38+.

#### Tests

  - `SqlIntervalWeekRewriteShuttleTest` exercises the shuttle directly against constructed `SqlIntervalLiteral` and `SqlCall` nodes for WEEK, QUARTER, DAY and MONTH (positive, negative, multi-unit), and against parsed queries that use both quoted and unquoted week and quarter intervals.
  - `CalciteQueryTest#testIntervalWeekResolution` is the end-to-end regression test from #18665. Without the fix it produces an interval narrowing on `2000-01-01T01:00:00.001` / `2000-01-01T02:00:00.001` (the bug); with the fix it correctly narrows to `2000-01-08` and `2000-01-15`.
  - `CalciteQueryTest#testIntervalQuarterResolution` is the companion regression for QUARTER: `INTERVAL '1' QUARTER` narrows to `2000-04-01` (three months later) and `INTERVAL 2 QUARTER` to `2000-07-01`.

I verified the bug is reachable by temporarily disabling the shuttle and watching the regression tests fail with exactly the actual-vs-expected from the issue. Re-enabling the shuttle makes them pass.

I also re-ran the surrounding test classes to make sure the rewrite did not regress any existing INTERVAL behavior:

  - `CalciteQueryTest` (441 tests)
  - `CalciteSelectQueryTest` (67 tests)
  - `CalciteJoinQueryTest` (594 tests)
  - `CalciteSubqueryTest` (54 tests)
  - `CalciteParameterQueryTest` (20 tests)
  - `CalciteInsertDmlTest` / `CalciteReplaceDmlTest` (107 tests)
  - `ExpressionsTest` (54 tests)
  - `DruidSqlParserUtilsTest` (57 tests)
  - `SqlQuidemTest` (10 tests)

All green. `mvn -pl sql -am verify -DskipTests=true` (checkstyle, forbiddenapis, animal-sniffer) and `mvn -pl sql -P rat verify -DskipTests=true` (license headers) also pass.

#### Release note

Fixed an issue where `INTERVAL n WEEK` in SQL queries resolved to `n` hours instead of `n * 7` days, and `INTERVAL n QUARTER` resolved to `n` months instead of `n * 3` months. `MILLIS_TO_TIMESTAMP(0) + INTERVAL '1' WEEK` now correctly returns `1970-01-08`, and `INTERVAL '1' QUARTER` adds three months, matching the SQL standard and other databases.

<hr>

 * `SqlIntervalWeekRewriteShuttle` (new, covers WEEK and QUARTER)
 * `DruidPlanner` (calls the shuttle in `validate`)
 * `SqlIntervalWeekRewriteShuttleTest` (new)
 * `CalciteQueryTest` (regression tests `testIntervalWeekResolution` and `testIntervalQuarterResolution`)

<hr>

This PR has:

- [x] been self-reviewed.
- [x] added unit tests or modified existing tests to cover new code paths.

### DIFF
--- a/sql/src/main/java/org/apache/druid/sql/calcite/planner/SqlIntervalWeekRewriteShuttle.java
+++ b/sql/src/main/java/org/apache/druid/sql/calcite/planner/SqlIntervalWeekRewriteShuttle.java
@@ -1,0 +1,146 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.sql.calcite.planner;
+
+import org.apache.calcite.avatica.util.TimeUnit;
+import org.apache.calcite.avatica.util.TimeUnitRange;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlIntervalLiteral;
+import org.apache.calcite.sql.SqlIntervalQualifier;
+import org.apache.calcite.sql.SqlLiteral;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.util.SqlShuttle;
+
+import java.math.BigDecimal;
+
+/**
+ * Workaround for a Calcite 1.37.0 bug in which {@code INTERVAL n WEEK} is
+ * incorrectly converted to {@code n} hours instead of {@code n * 7} days.
+ *
+ * <p>The bug is in {@code SqlIntervalQualifier.evaluateIntervalLiteralAsWeek},
+ * which packages the WEEK value into a year-month shaped int array
+ * ({@code [sign, 0, week]}) even though {@code SqlIntervalQualifier.typeName()}
+ * reports WEEK intervals as {@link org.apache.calcite.sql.type.SqlTypeName#INTERVAL_DAY}.
+ * Downstream, {@code SqlParserUtil.intervalToMillis} interprets that array as
+ * {@code [sign, day, hour, minute, second, ms]}, so the week value lands in
+ * the hour slot. The bug was fixed upstream in Calcite 1.38.0
+ * (<a href="https://issues.apache.org/jira/browse/CALCITE-6391">CALCITE-6391</a>),
+ * but Druid is still on Calcite 1.37.0, so we work around it by rewriting any
+ * {@code WEEK} interval in the parsed {@link SqlNode} tree to an equivalent
+ * {@code DAY} interval whose value has been multiplied by seven.
+ *
+ * <p>Two parsed forms reach this shuttle:
+ * <ul>
+ *   <li>Quoted: {@code INTERVAL '1' WEEK} parses to a {@link SqlIntervalLiteral}
+ *       with a WEEK qualifier. We replace it with a {@link SqlIntervalLiteral}
+ *       containing the original numeric value multiplied by seven and a
+ *       {@link TimeUnit#DAY} qualifier.</li>
+ *   <li>Unquoted: {@code INTERVAL 1 WEEK} parses to
+ *       {@code SqlStdOperatorTable.INTERVAL.createCall(pos, n, qualifier)}.
+ *       We rewrite the call so the numeric operand becomes
+ *       {@code MULTIPLY(n, 7)} and the qualifier becomes {@code DAY}.</li>
+ * </ul>
+ *
+ * <p>This rewrite preserves the semantics of {@code TimeUnitRange.WEEK} while
+ * avoiding the buggy code path inside Calcite. Druid still treats WEEK as a
+ * day-time (millisecond) interval in arithmetic, which matches the expectation
+ * documented in the issue: {@code INTERVAL '1' WEEK} should add seven days.
+ */
+public class SqlIntervalWeekRewriteShuttle extends SqlShuttle
+{
+  private static final BigDecimal SEVEN = BigDecimal.valueOf(7);
+
+  @Override
+  public SqlNode visit(SqlLiteral literal)
+  {
+    if (literal instanceof SqlIntervalLiteral) {
+      final SqlIntervalLiteral intervalLiteral = (SqlIntervalLiteral) literal;
+      final SqlIntervalLiteral.IntervalValue value =
+          (SqlIntervalLiteral.IntervalValue) intervalLiteral.getValue();
+      if (value != null && isPlainWeek(value.getIntervalQualifier())) {
+        final String multiplied = multiplyByWeek(value.getIntervalLiteral());
+        if (multiplied != null) {
+          return SqlLiteral.createInterval(
+              value.getSign(),
+              multiplied,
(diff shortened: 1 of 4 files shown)


gianm:
Did you run into some problem while upgrading Calcite? It would be better to fix the bug that way, rather than through a workaround.

Also, the bug was fixed in https://issues.apache.org/jira/browse/CALCITE-6581. I know because I wrote the patch 🙂. At the time I figured we would get the fix the next time we upgraded Calcite.

barry3406:
Good point — I didn't attempt the Calcite upgrade because it felt too big for a bug-fix PR (1.37 → 1.38 tends to pull in a lot of changes). This shuttle is removable once the upgrade lands.

Also thanks for the correction, the actual fix is CALCITE-6581 not CALCITE-6391 — I'll update the references. If you'd rather just bump Calcite instead I'm fine closing this.

gianm:
Certainly I'd prefer upgrading Calcite if it's straightforward. That's why I was wondering if you tried and ran into any issues. It can a be a lot of work (see #14510, #16504) but I was hoping it wouldn't be too bad this time.

If upgrading Calcite isn't straightforward, we could merge your fix here temporarily, then remove it once we do upgrade Calcite. Btw, if we are going to do the temporary fix, please include `QUARTER` too since that is also broken in Calcite 1.37.0.

barry3406:
Added QUARTER support in 09b6250 — mirrors the WEEK rewrite (`n MONTH * 3`) with the same test coverage (quoted/unquoted, positive/negative, multi-unit, parser end-to-end, plus a `testIntervalQuarterResolution` regression in `CalciteQueryTest` that proves `INTERVAL 2 QUARTER` narrows to `2000-07-01`). Also corrected the CALCITE-6581 reference throughout the code and comments. Haven't attempted the 1.38 bump — happy to keep this as the temp workaround per your suggestion, and equally happy to drop it when someone gets to the upgrade.

gianm:
I raised a PR for the Calcite update here: #19370. That should also fix this issue. The PR adds a test `sql/src/test/quidem/org.apache.druid.quidem.SqlQuidemTest/interval_arithmetic.iq` verifying it.

FrankChen021:

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

sql/src/main/java/org/apache/druid/sql/calcite/planner/SqlIntervalWeekRewriteShuttle.java
```diff
@@ -0,0 +1,201 @@
...
+          (SqlIntervalLiteral.IntervalValue) intervalLiteral.getValue();
+      if (value != null) {
+        final SqlIntervalQualifier qualifier = value.getIntervalQualifier();
+        if (isPlainWeek(qualifier)) {
+          final String multiplied = multiplyLiteral(value.getIntervalLiteral(), SEVEN);
+          if (multiplied != null) {
+            return SqlLiteral.createInterval(
+                value.getSign(),
+                multiplied,
+                new SqlIntervalQualifier(TimeUnit.DAY, null, qualifier.getParserPosition()),
```
FrankChen021, line 107: [P2] Preserve interval leading precision

Rewriting WEEK/QUARTER literals to DAY/MONTH qualifiers drops any explicit leading precision from the original qualifier before Calcite validation runs. For example, INTERVAL '100' WEEK(2) should be rejected because the WEEK leading field has three digits with precision 2, but after this rewrite it becomes an unspecified-precision DAY literal and can validate successfully. The same precision loss happens for QUARTER and for the unquoted INTERVAL call path. Preserve or explicitly validate the original qualifier precision before replacing it.

PR #19301, opened 2026-04-13
feat: Framing for FIRST_VALUE, LAST_VALUE.

gianm:
This patch implements framing support for the FIRST_VALUE, LAST_VALUE window functions and updates all relevant tests. New tests are also added for the processors themselves.


processing/src/main/java/org/apache/druid/query/operator/window/value/WindowValueProcessorBase.java
```diff
@@ -108,4 +112,59 @@ public List<String> getOutputColumnNames()
...
+        final int frameLower = Math.max(0, i + lower);
+        final int frameUpper = Math.min(numRows - 1, i + upper);
+        if (frameLower <= frameUpper) {
+          results[i] = accessor.getObject(first ? frameLower : frameUpper);
+        }
+      }
+      return;
+    }
+
+    final WindowFrame.Groups groupsFrame = frame.unwrap(WindowFrame.Groups.class);
```
FrankChen021, line 149: [P1] Bounded RANGE frames still execute with GROUPS semantics

The new framed path treats every non-ROWS frame as `WindowFrame.Groups`, and `Windowing` maps Calcite's `isRows == false` windows into that representation. That is not correct for bounded SQL RANGE frames, which must use value-distance on the ORDER BY expression. Queries like `FIRST_VALUE(x) OVER (ORDER BY t RANGE BETWEEN 1 PRECEDING AND CURRENT ROW)` will still return wrong answers even though this PR now enables framed FIRST_VALUE/LAST_VALUE.

gianm: Processing won't get this far due to validation. There's a test for it in `CalciteQueryTest#testUnSupportedRangeBounds`, expecting an error message like:

> Order By with RANGE clause currently supports only UNBOUNDED or CURRENT ROW.

I added a comment here about that.

FrankChen021 (approved):

FrankChen021 (commented):
Reviewed 79 of 79 changed files. The original bounded RANGE concern is fixed on the SQL path because validation rejects offset-based non-ROWS frames before the GROUPS-based implementation. No additional high-confidence PR-caused issues found.

---

clintropolis (approved):

PR #19303, opened 2026-04-13
fix: Fix diskNormalized cost balancer strategy

jtuglu1:
### Description

We've seen issues with the `cost` segment balancing strategy not working correctly in production, where it leaves historicals in a tier in a bimodal distribution: ~50% in ~100% utilization and ~50% in 60% utilization. This is due to the cost-based balancer not having any visibility into disk util on each historical. The diskNormalized strategy is available, but is marked as not "production-ready," I believe due to the following reason:

The “diskNormalized” balancer works by extending from the “cost” balancer and overriding the computeCost function in the following way:

```
double cost = super.computeCost(proposalSegment, server, includeCurrentServer);
if (cost == Double.POSITIVE_INFINITY) {
  return cost;
}
int nSegments = 1;
if (server.getServer().getNumSegments() > 0) {
  nSegments = server.getServer().getNumSegments();
}
double normalizedCost = cost / nSegments;
double usageRatio = (double) server.getSizeUsed() / (double) server.getServer().getMaxSize();
return normalizedCost * usageRatio;
```

This logic is a bit strange. Consider the case where historicals are all homogeneous, so server.getServer().getMaxSize() can be treated as a constant. In that case, the returned value simplifies to:

```
normalizedCost * usageRatio
= normalizedCost * usedSize / maxSize
= normalizedCost * usedSize / CONST
= cost / numSegments * usedSize / CONST
= cost * (usedSize / numSegments) / CONST
= cost * avgSegmentSize / CONST
```

So, in practice, it’s just adjusting cost by a factor of avgSegmentSize. This could lead to imbalances that make themselves worse over time, especially if newly created segments tend to have smaller sizes than older ones (which is common with append-based ingestion methods).

This fixes the diskNormalized strategy just weight the `CostBalancerStrategy` with the utilization of the server, defaulting to a 5% lenience threshold (TODO: make this adjustable via config?) to avoid ping-ponging segments between 2 historicals due to utilization differences.

#### Release note

Fix diskNormalized cost balancer strategy

<hr>

This PR has:

- [ ] been self-reviewed.


server/src/main/java/org/apache/druid/server/coordinator/balancer/DiskNormalizedCostBalancerStrategyConfig.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.apache.druid.server.coordinator.balancer;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class DiskNormalizedCostBalancerStrategyConfig
+{
+  @JsonProperty
+  private double diskUtilThresholdTolerance = DiskNormalizedCostBalancerStrategy.DEFAULT_DISK_UTIL_THRESHOLD_TOLERANCE;
```
maytasm: Can we add validation for bounds/range check?

services/src/main/java/org/apache/druid/cli/CliCoordinator.java
```diff
@@ -197,6 +198,7 @@ public void configure(Binder binder)
             JsonConfigProvider.bind(binder, "druid.coordinator.period", CoordinatorPeriodConfig.class);
             JsonConfigProvider.bind(binder, "druid.coordinator.loadqueuepeon.http", HttpLoadQueuePeonConfig.class);
             JsonConfigProvider.bind(binder, "druid.coordinator.balancer", BalancerStrategyFactory.class);
+            JsonConfigProvider.bind(binder, "druid.coordinator.balancer.diskNormalized", DiskNormalizedCostBalancerStrategyConfig.class);
```
maytasm, line 201: can you add docs for this new config?

maytasm (approved):
nit: diskUtilThresholdTolerance is a weird name. It doesn't indicate that a segment is already on the server. Maybe like a ...stickiness... or ...movingCost... is better?

PR #12802, opened 2022-07-19
Peons should not report SysMonitor stats since MiddleManager reports them.

tejaswini-imply:
### Description

`Sysmonitor` stats (mem, fs, disk, net, cpu, swap, sys, tcp) are reported by all Druid processes, including _Peons_ that are ephemeral in nature. Since _Peons_ always run on the same host as the _MiddleManager_ that spawned them and is unlikely to change, the SyMonitor metrics emitted by _Peon_ are merely duplicates. This is often not a problem except when machines are super-beefy. Imagine a 64-core machine and 32 workers running on this machine. now you will have each _Peon_ reporting metrics for each core. that's an increase of (32 * 64)x in the number of metrics. This leads to a metric explosion.

This PR updates `MetricsModule` to check node role running while registering `SysMonitor` and not to load any existing `SysMonitor$Stats`.

<hr>

##### Key changed/added classes in this PR
 * `MetricsModule`
 * `SysMonitor`

<hr>

This PR has:
- [x] been self-reviewed.
   - [ ] using the [concurrency checklist](https://github.com/apache/druid/blob/master/dev/code-review/concurrency.md) (Remove this item if the PR doesn't have any relation to concurrency.)
- [ ] added documentation for new or modified features or behaviors.
- [ ] added Javadocs for most classes and all non-trivial methods. Linked related entities via Javadoc links.
- [ ] added or updated version, license, or notice information in [licenses.yaml](https://github.com/apache/druid/blob/master/dev/license.md)
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [x] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for [code coverage](https://github.com/apache/druid/blob/master/dev/code-review/code-coverage.md) is met.
- [ ] added integration tests.
- [x] been tested in a test Druid cluster.

PR #19305, opened 2026-04-14
fix: Emit sys metrics from k8s peons.

gianm:
Currently the SysMonitor and OshiSysMonitor are both skipped for peons, on the logic that they are redundant to the MiddleManager (see #12802). However, with the Kubernetes runner, we launch peons directly and we want them to emit sys metrics.

To make it possible to differentiate these cases, this patch adds a property "druid.peon.managed" that informs the MetricsModule that the peon is managed by another process and should not emit metrics.


capistrant (commented):
Nice idea. The MetricsModuleTest fail is legit as the change breaks the validation since the new property is not set there, to get the noop monitor. Other test fails look unrelated

capistrant (approved):

PR #19308, opened 2026-04-14
minor: configure spotless

kgyrtkirk:
Spotless could do a little bit more than checkstyle which is only pointing out where the problems are - it also  provides a way to fix issues.

This changeset configures the importorder and some other simpler rules.
I wanted to also configure the formatter - but that changes indentation in a lot of files; haven't figured out a way to avoid most changes which seem to be just noise.

Even the import order changes a lot of files; there are a lot of files where a blank line between `javax` and `java` was missing.

The real meaningfull changes could be viewed with:
```
git diff `git merge-base HEAD apache/master` HEAD -w  --ignore-blank-lines
```


FrankChen021:

pom.xml
```diff
@@ -1762,6 +1762,52 @@
...
+                <version>3.4.0</version>
+                <configuration>
+                    <skip>${spotless.skip}</skip>
+                    <java>
+                        <excludes>
+                            <exclude>**/extendedset/intset/**</exclude>
+                        </excludes>
+                        <cleanthat>
+                            <version>2.25</version>
+                            <sourceJdk>${maven.compiler.source}</sourceJdk>
```
FrankChen021, line 1777: [P2] Use the Java 17 release for CleanThat

The new Spotless/CleanThat configuration uses `${maven.compiler.source}`, which resolves from the Apache parent to `1.8`, while this repo compiles with `${maven.compiler.release}` set to Java 17 and already contains Java 17 syntax such as text blocks and records. Since Spotless runs during `validate` across modules, CleanThat can parse/rewrite valid project sources as Java 8 or fail on them, breaking normal builds. Point `sourceJdk` at `${maven.compiler.release}` or `${java.version}` instead.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

Issue #18439, opened 2025-08-26
Kafka 4.x upgrade for Queue Semantics support in  Kafka Ingestion


Shekharrajak:
Key Issues:
  1. Static Partition Management: Supervisor manually assigns partitions to tasks
  2. No Dynamic Rebalancing: Requires supervisor intervention for scaling
  3. Complex Coordination: Custom offset management and task coordination

The Shared Group API (KIP-932) Kafka handles partition assignment automatically and consumer group coordination can be simplified, hence upgrading to Kafka Share Consumer API can have significant performance improvements.

gianm:
Are there any backwards compatibility issues with older Kafka brokers? If not then upgrading sounds good to me.

Shekharrajak:
Actually there is different API `KafkaShareConsumer`  to leverage the Kafka Queue Semantics: https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka

Shekharrajak:
Spend sometime in finding out gaps and missing pieces in share group Kafka implementation but now since Kafka Queues released with GA, I think we should start desiging druid Kafka ingestion to add features of Kafka Share group. I will share the draft design doc.

Shekharrajak:
To make sure there is no data loss : 

```
Phase 1 — POLL
  records = consumer.poll(pollTimeout)
  Broker: acquisition lock set on each record (TTL = lock.duration.ms)
  Local:  recordsInFlight = new LinkedHashMap<>()  // (topic,partition,offset) -> record
  RENEW thread: started (see Section 10)

Phase 2 — PARSE
  for each record r:
    try:
      row = inputRowParser.parse(r)
      pendingRows.add(row)
      pendingAccept.add(r)
    catch ParseException:
      parseExceptionCount++
      if parseExceptionCount > maxParseExceptions:
        // Too many bad records — operator must fix schema. Fail the task.
        // RELEASE all records: broker redelivers after lock expiry.
        // (Do not REJECT on task failure — preserve redelivery for non-parse records.)
        releaseAllAndFail()
      pendingReject.add(r)   // REJECT: permanent schema error, no value in retrying

Phase 3 — INGEST
  appenderator.add(pendingRows)   // write to JVM heap buffer

Phase 4 — PERSIST  (local disk — crash-safe boundary 1)
  if rowCount >= maxRowsInMemory OR elapsed >= intermediatePersistPeriod:
    try:
[...]

PR #19311, opened 2026-04-14
feat: [Phase 1] Queue Semantics support in Kafka Ingestion

Shekharrajak:
Phase 1 https://github.com/apache/druid/issues/18439.

### Description

  - Adds index_kafka_share_group task and kafka_share_group IO config.
  - Wraps KafkaShareConsumer with explicit per-record ACK support.
  - Publishes Druid segments first, then ACKs Kafka records.
  - Supports concurrent tasks with the same groupId, including more tasks than Kafka partitions.
  - Adds docs, unit tests, and embedded ITs for JSON submit, restart, crash recovery, partition rebalancing, and concurrent task ingestion.
 

 Next Phases

  1. Add KafkaShareGroupSupervisor for managed task lifecycle, scaling, restart, and UI/API parity with Kafka supervisors.
  2. Add deduplication/EOS plan: record identity tracking, published-batch metadata, and duplicate suppression across retries.
  3. Add production hardening: lock-duration tuning guidance, metrics/alerts, better task autoscaling behavior, and more chaos/failure tests.

#### Release note

Druid now supports experimental Kafka share-group ingestion via index_kafka_share_group, enabling queue-style Kafka consumption with broker-managed delivery state and at-least-once ingestion semantics.

  Read semantics:

  - Kafka broker owns delivery state and offsets.
  - Druid does not manage start/end offsets.
  - Records are delivered with share-group acquisition locks.
  - Multiple tasks can read from the same partition concurrently.
  - Per-partition ordering is not guaranteed.
  - Unacknowledged records are redelivered after lock expiry.

  Write semantics:

  - Druid writes through append segment publishing.
  - Segments are atomically published to Druid metadata.
  - Kafka records are ACKed only after successful segment publish.
  - If failure happens after publish but before ACK commit, Kafka redelivers and Druid may contain duplicates.
  - No dedup cache yet, so this is not EOS.

This PR has:

- [x] been self-reviewed.

# Demo 

https://youtu.be/K_O1MH-AaE8


docs/ingestion/kafka-share-group-ingestion.md
```diff
@@ -0,0 +1,260 @@
...
+  -Pskip-static-checks -Dmaven.test.skip=true -T1C -q
+
+# Download and extract Druid release
+cd ~/Downloads
+curl -O https://dlcdn.apache.org/druid/31.0.0/apache-druid-31.0.0-bin.tar.gz
+tar -xzf apache-druid-31.0.0-bin.tar.gz
+cd apache-druid-31.0.0
+
+# Replace the kafka extension with our build
+rm extensions/druid-kafka-indexing-service/*.jar
```
Shekharrajak, line 196: We can build from the source and use the tar

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/simulate/EmbeddedShareGroupIngestionTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+    cluster.callApi().submitTask(task);
+
+    indexer.latchableEmitter().waitForEventAggregate(
+        event -> event.hasMetricName("ingest/events/processed")
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource),
+        agg -> agg.hasSumAtLeast(numRecords)
+    );
+
+    Assertions.assertEquals(
```
Shekharrajak, line 158: Making sure we read all the records.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
```diff
@@ -0,0 +1,363 @@
...
+          batchOffsets.computeIfAbsent(record.getPartitionId(), k -> new ArrayList<>())
+                      .add(record.getSequenceNumber());
+
+          if (record.getData() == null) {
+            continue;
+          }
+
+          for (ByteEntity entity : record.getData()) {
+            try {
+              final InputRow row = parseRow(entity, inputFormat, inputRowSchema, toolbox);
```
FrankChen021: [P1] Share-group ingestion drops rows when one Kafka record expands to multiple rows

This runner processes each Kafka message as though it produces exactly one InputRow. `parseRow()` only returns `iterator.next()`, so later rows from the same message are silently discarded, and zero-row messages can pass null into `driver.add`. Because the code still ACKs the Kafka record after publish, those skipped rows are lost permanently rather than being retried.

Shekharrajak: Thanks for checking. Yes, in phase 2 implementation I am trying to make sure no data loss. 
Also trying couple of benchmark when Kafka Queue Semantics actually benefit us.

FrankChen021: Thanks for confirming. I still think this needs to be resolved or explicitly scoped out before this PR lands: the current task/docs already state the no-data-loss invariant, but `parseRow()` still consumes only the first row from the reader and the batch ACK then accepts the whole Kafka record. Any input format that expands one Kafka record into multiple rows would still silently lose the later rows after publish/commit. Could we either iterate and add every row before ACKing, or enforce/document a one-row-per-record limitation for this phase so the invariant is accurate?

Shekharrajak: Sure, I will keep in mind. @FrankChen021 after these changes should we go ahead and phase wise rollout the kafka share group implementation in Druid ? 
I will address this  review comment and also the previous one

> Create a new NonSharedQueueKafkaIndexTask, which is exactly current implementation
create a new SharedQueueKafkaIndexTask, which is the one your want to implement
the KafkaIndexTask holds either instance of NonSharedQueueKafkaIndexTask or SharedQueueKafkaIndexTask based on the "type": "kafka_share_group" defined in the ioConfig, and delegate all public methods to these instance

I find that after implementing the supervisor and dedup cache (in my local branch) it has become a huge code changes that would be hard to review.

FrankChen021: Phasing the implementation sounds reasonable to me, especially if supervisor and dedup-cache work makes the PR too large to review as one change. But I would still treat this specific issue as a merge blocker unless this phase either fixes it or explicitly scopes it out in code/docs: the current runner still ACKs the whole Kafka record after publish while `parseRow()` only consumes the first row from the input reader. That means the no-data-loss invariant is still inaccurate for any input format that can produce multiple rows from one Kafka record. For this phase, I think the minimum viable resolution is either to iterate and add every row before ACKing, or to reject/document a one-row-per-record limitation so users cannot assume the stronger invariant.

Shekharrajak: Sure, I am working on it and need full round of local validation. I will share the code changes soon.

Shekharrajak:
Spending time on  Integeration testing embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedShareGroupIngestionTest.java

Shekharrajak:
For smaller blast radius and clean separation - parking the suggestion https://github.com/apache/druid/issues/18439#issuecomment-4294451459 , Also KafkaShareGroupSupervisor (planned) will be structurally different from KafkaSupervisor — no partition-to-task assignment, no offset bookkeeping, seek/rest .

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedShareGroupIngestionTest.java
```diff
@@ -0,0 +1,446 @@
...
+           .addServer(overlord)
+           .addServer(indexer)
+           .addServer(historical)
+           .addServer(broker);
+
+    return cluster;
+  }
+
+  @Test
+  public void test_shareGroupIngestion_basicEndToEnd() throws InterruptedException
```
Shekharrajak, line 102: Subscribe to topic with KafkaShareConsumer (groupId), poll → parse (CSV) → persist → publish → acknowledge(ACCEPT) → commitSync → broker offset persistence. SQL COUNT(*) == 10.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedShareGroupIngestionTest.java
```diff
@@ -0,0 +1,446 @@
...
+        new JsonInputFormat(null, Collections.emptyMap(), null, null, true),
+        defaultTuningConfig()
+    );
+
+    Thread.sleep(SHARE_CONSUMER_READY_DELAY_MS);
+
+    final int numRecords = 5;
+    final int rowsPerRecord = 4;
+    kafkaServer.produceRecordsToTopic(
+        generateMultiObjectJsonRecords(topic, numRecords, rowsPerRecord, DateTimes.of("2025-07-01"))
```
Shekharrajak, line 149: StreamChunkReader multi-row path  One Kafka record carrying 4 whitespace-separated JSON objects produces 4 rows. SQL count = numRecords × rowsPerRecord (5 × 4 = 20). Verifies no row is dropped before ACK.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
Shekharrajak: Updated for multi row : https://github.com/apache/druid/pull/19311/changes#r3207054221

Shekharrajak:
To make sure no data loss we are ack only after the publish completed https://github.com/apache/druid/pull/19311/changes#diff-487176908fb14d87486db6f683c9ddb34465f3a1572f5c01b11d7b9c6a482289R381 

Although this will be slow and no concurrency but I think for the intial phase this is fine and we can improve the performance step by step. 

WDYT  @gianm @a2l007 @clintropolis @FrankChen021  ?

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
FrankChen021: Thanks, the current version now uses `StreamChunkReader` and iterates every returned `InputRow` before acknowledging the batch, so the multi-row and zero-row cases from this thread look addressed.

Shekharrajak:
CI checks are failing due to one flaky test unrelated to this PR : https://github.com/apache/druid/issues/19435

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTask.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Indexing task that consumes from a Kafka topic with share-group semantics
+ * (KIP-932). The broker tracks delivery state; the task acknowledges records
+ * after the segments containing them are published. Phase 1: single-threaded,
+ * no supervisor, no dedup cache.
+ */
+public class ShareGroupIndexTask extends AbstractTask implements PendingSegmentAllocatingTask
```
FrankChen021, line 54: [P2] Share-group tasks inherit priority 0

Because ShareGroupIndexTask extends AbstractTask directly and does not override getPriority(), tasks submitted without an explicit context priority use Tasks.DEFAULT_TASK_PRIORITY (0). Existing Kafka realtime tasks use Tasks.DEFAULT_REALTIME_TASK_PRIORITY (75), and TaskLockbox uses task.getPriority() when acquiring or competing for append locks, so these ingestion tasks can lose lock contention or be ordered differently from equivalent Kafka ingestion. Add the same realtime priority override used by SeekableStreamIndexTask.

extensions-core/kafka-indexing-service/pom.xml
```diff
@@ -252,6 +257,30 @@
...
+          kafka_2.13:4.x (KRaft) or Testcontainers. Share-group code paths are
+          covered by ShareGroupIndexTaskRunnerTest, KafkaShareGroupRecordSupplierTest
+          (mock-based) and embedded-tests/EmbeddedShareGroupIngestionTest
+          (Testcontainers). Tracked as a follow-up of apache/druid#19311.
+        -->
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <excludes>
+            <exclude>**/KafkaIndexTaskTest.java</exclude>
```
FrankChen021: [P2] Kafka compatibility tests are skipped globally

This Surefire exclusion removes the main KafkaIndexTask, KafkaRecordSupplier, KafkaSamplerSpec, and KafkaSupervisor test classes from every normal test run. Since this PR also upgrades kafka-clients to 4.2.0, those are the regression tests most likely to catch compatibility breaks in existing index_kafka ingestion; the new share-group mock and embedded tests do not cover that existing API surface. Please migrate or narrowly isolate the TestBroker-dependent cases instead of excluding the whole classes.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

Reviewed 22 of 22 changed files. Findings focus on task scheduling semantics and Kafka compatibility test coverage.

---

extensions-core/kafka-indexing-service/pom.xml
Shekharrajak: Yes, planning to fix (upgrade to 4.2 version in all places) as part of this https://github.com/apache/druid/issues/19322 - I feel that we need bunch of changes, API compatibility issues in running tests , kafka docker image for IT tests and in source code - so want to keep it in separate PR.

This PR mainly focus on Share group implementation.

Shekharrajak:
Struggling with CI check related to license : 

```

Unsupported license: MPL 1.1
For:org.javassist javassist in: druid-kerberos
org.locationtech.jts jts-core in: parquet-extensions with: Eclipse Public License 2.0 ignoring Eclipse Distribution License - v 1.0
Error:  Command execution failed.
org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
    at org.apache.commons.exec.DefaultExecutor.executeInternal (DefaultExecutor.java:394)
    at org.apache.commons.exec.DefaultExecutor.execute (DefaultExecutor.java:292)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:903)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:863)
    at org.codehaus.mojo.exec.ExecMojo.execute (ExecMojo.java:454)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:328)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute (DefaultMojosExecutionStrategy.java:39)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:105)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:73)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:53)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:118)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:261)
[INFO] simple-client-sslcontext ........................... SUCCESS [  5.195 s]
[INFO] druid-basic-security ............................... SUCCESS [  9.326 s]
[INFO] druid-google-extensions ............................ SUCCESS [  7.697 s]
[INFO] druid-catalog ...................................... SUCCESS [  9.519 s]
[INFO] druid-testing-tools ................................ SUCCESS [  8.122 s]
[INFO] druid-testcontainers ............................... SUCCESS [  2.149 s]
[INFO] druid-influx-extensions ............................ SUCCESS [  4.579 s]
[INFO] druid-cassandra-storage ............................ SUCCESS [  3.987 s]
[INFO] dropwizard-emitter ................................. SUCCESS [  6.528 s]
[INFO] druid-cloudfiles-extensions ........................ SUCCESS [  5.857 s]
[INFO] druid-consul-extensions ............................ SUCCESS [  6.595 s]
[INFO] graphite-emitter ................................... SUCCESS [  6.085 s]
[INFO] druid-distinctcount ................................ SUCCESS [  4.219 s]
[INFO] druid-exact-count-bitmap ........................... SUCCESS [  7.275 s]
[INFO] statsd-emitter ..................................... SUCCESS [  6.282 s]
[INFO] druid-time-min-max ................................. SUCCESS [  5.725 s]
[INFO] druid-virtual-columns .............................. SUCCESS [  4.011 s]
[INFO] druid-thrift-extensions ............................ SUCCESS [  4.915 s]
[INFO] ambari-metrics-emitter ............................. SUCCESS [  6.306 s]
[INFO] sqlserver-metadata-storage ......................... SUCCESS [  6.266 s]
[INFO] kafka-emitter ...................................... SUCCESS [  6.186 s]
[INFO] druid-redis-cache .................................. SUCCESS [  7.346 s]
[INFO] druid-opentsdb-emitter ............................. SUCCESS [  5.524 s]
[INFO] druid-momentsketch ................................. SUCCESS [  6.063 s]
[INFO] druid-moving-average-query ......................... SUCCESS [  7.560 s]
[INFO] tdigestsketch ...................................... SUCCESS [  7.308 s]
[INFO] druid-ddsketch ..................................... SUCCESS [  7.400 s]
[INFO] druid-influxdb-emitter ............................. SUCCESS [  4.017 s]
[INFO] gce-extensions ..................................... SUCCESS [  6.621 s]
[INFO] aliyun-oss-extensions .............................. SUCCESS [  7.980 s]
[INFO] prometheus-emitter ................................. SUCCESS [  5.877 s]
[INFO] opentelemetry-emitter .............................. SUCCESS [  8.167 s]
[INFO] grpc-query ......................................... SUCCESS [ 14.755 s]
[INFO] druid-iceberg-extensions ........................... SUCCESS [  9.575 s]
[INFO] druid-deltalake-extensions ......................... SUCCESS [  6.163 s]
[INFO] druid-spectator-histogram .......................... SUCCESS [  8.500 s]
[INFO] druid-rabbit-indexing-service ...................... SUCCESS [  7.905 s]
[INFO] druid-ranger-security .............................. SUCCESS [  9.363 s]
[INFO] distribution ....................................... FAILURE [05:29 min]
[INFO] druid-quidem-ut .................................... SKIPPED
[INFO] druid-embedded-tests ............................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  22:30 min
[INFO] Finished at: 2026-05-09T05:59:40Z
[INFO] ------------------------------------------------------------------------
Error:  Failed to execute goal org.codehaus.mojo:exec-maven-plugin:3.5.1:exec (check-licenses) on project distribution: Command execution failed. Process exited with an error: 1 (Exit value: 1) -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
Error:  
Error:  After correcting the problems, you can resume the build with the command
Error:    mvn <args> -rf :distribution
Error: Process completed with exit code 1.

```

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTask.java
Shekharrajak: Thanks for sharing, this is important for task scheduling order in TaskQueue. Updating..

Shekharrajak: Updated https://github.com/apache/druid/pull/19311/commits/ede507254b92a60cef4fd174d10a5caf64dcca92

FrankChen021 (commented):
Reviewed 3 of 27 changed files. Follow-up review handled: the Kafka extension POM no longer has the global Surefire exclusions, and ShareGroupIndexTask now uses DEFAULT_REALTIME_TASK_PRIORITY with tests covering the default and context override cases.

---

docs/ingestion/kafka-share-group-ingestion.md
```diff
@@ -0,0 +1,337 @@
...
+```
+
+### Step 2: Create topic and produce messages
+
+```bash
+cd kafka_2.13-4.2.0
+
+bin/kafka-topics.sh --create --topic druid-share-test --partitions 4 --bootstrap-server localhost:9092
+
+bin/kafka-console-producer.sh --topic druid-share-test --bootstrap-server localhost:9092
```
FrankChen021, line 229: [P2] Demo produces records before the share group can read them

The demo creates and produces the sample messages before submitting druid-demo-share-group, but share-group starting position is broker-controlled and this PR strips auto.offset.reset; the tests explicitly set share.auto.offset.reset=earliest for this reason. With the broker default of latest, users following the demo will submit the task after the sample records already exist and ingest zero rows. Move producing after task startup or add a kafka-configs/AdminClient step that sets share.auto.offset.reset=earliest for the demo group before producing.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaShareGroupRecordSupplier.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+  }
+
+  private static KafkaShareConsumer<byte[], byte[]> createShareConsumer(
+      Map<String, Object> consumerProperties,
+      ObjectMapper sortingMapper,
+      String groupId
+  )
+  {
+    final Map<String, Object> sanitized = ShareGroupConsumerProperties.sanitize(consumerProperties);
```
FrankChen021, line 250: [P2] Unsupported share-consumer configs can still enter through dynamic config

sanitize() runs before KafkaRecordSupplier.addConsumerPropertiesFromConfig, but that helper expands druid.dynamic.config.provider afterward and copies all returned entries directly into the Properties. If a dynamic provider supplies a forbidden key such as enable.auto.commit, auto.offset.reset, or group.instance.id, KafkaShareConsumer still receives it and the task can fail at startup despite the documented stripping behavior. Sanitize after dynamic expansion as well, or expand into a map first and then strip unsupported keys once.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

---

docs/ingestion/kafka-share-group-ingestion.md
Shekharrajak: Updated. Thanks!

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaShareGroupRecordSupplier.java
Shekharrajak: updated.

FrankChen021 (commented):
Follow-up handled. Reviewed 21 of 21 changed files. The demo now configures `share.auto.offset.reset=earliest` before producing sample records, and share-group consumer properties are sanitized again after dynamic config expansion. I do not have further inline replies on the two follow-up threads.

---

Shekharrajak:
CI Check error -  unrelated timeouts  :

```
  embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.class:282 | Timed out waiting for event after [20,000]ms 20.391s
  embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.class:337 | Timed out waiting for event after [20,000]ms 20.015s
  embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.class:166 | Timed out waiting for event after [20,000]ms 20.019s
  ℹ️ Posting with conclusion 'failure' to https://github.com/apache/druid/pull/19311 (sha: 2d094d5fbe6e11b6d22b1df8bf54c8cf7b7d03d2)
🚀 Publish results
  ℹ️ - Test Report - test-jdk25-[I*,A*,U*] - 18604 tests run, 17847 passed, 754 skipped, 3 failed.
     🧪 - embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.class | Timed out waiting for event after [20,000]ms
     🧪 - embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.class | Timed out waiting for event after [20,000]ms
     🧪 - embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.class | Timed out waiting for event after [20,000]ms

```

Raised ticket https://github.com/apache/druid/issues/19463

Shekharrajak:
Another flaky test: https://github.com/apache/druid/issues/19464

Shekharrajak:
fixing the CI check for coverage improvements. Looks like it is bit tricky to unit test`extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java` run method

Shekharrajak:
CI checks are looking fine. 

Almost all IT tests are written here embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedShareGroupIngestionTest.java 

Let me see if we can have more IT tests so that it is easy to understand the how to use it [ we can have more test case in followup PRs as well the main logical test cases are already added.]

cc @FrankChen021

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
```diff
@@ -0,0 +1,487 @@
...
+      StreamAppenderatorDriver driver,
+      Appenderator appenderator,
+      AcknowledgingRecordSupplier<KafkaTopicPartition, Long, KafkaRecordEntity> recordSupplier,
+      StreamChunkReader chunkReader,
+      ShareGroupIndexTaskIOConfig ioConfig,
+      KafkaIndexTaskTuningConfig tuningConfig,
+      TaskToolbox toolbox
+  ) throws Exception
+  {
+    recordSupplier.subscribe(Collections.singleton(ioConfig.getTopic()));
```
FrankChen021: [P2] Close the appenderator when setup fails

After this refactor, appenderator.closeNow() only runs in the try/finally that starts after recordSupplier.subscribe(...) and driver.startJob(...). If supplier creation, subscribe, or startJob throws, run() only calls driver.close(), and BaseAppenderatorDriver.close explicitly does not close the underlying Appenderator, so failed startup can leak realtime appenderator resources. Keep the appenderator close guard in run() or include setup in runLoop's try/finally.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexTaskJsonSubmitIT.java
```diff
@@ -0,0 +1,244 @@
...
+    cluster.callApi().onLeaderOverlord(o -> o.runTask(task.getId(), taskJson));
+
+    Thread.sleep(SHARE_CONSUMER_READY_DELAY_MS);
+
+    final org.apache.druid.client.indexing.TaskPayloadResponse payloadResponse =
+        cluster.callApi().onLeaderOverlord(o -> o.taskPayload(task.getId()));
+    Assertions.assertEquals(task.getId(), payloadResponse.getTask());
+
+    final JsonNode payloadJson = mapper.valueToTree(payloadResponse.getPayload());
+    Assertions.assertEquals("share_group_index", payloadJson.path("type").asText());
```
FrankChen021: [P2] Fix JSON-submit payload expectations

This new IT serializes a real ShareGroupIndexTask and reads the Overlord payload, but the task is registered as index_kafka_share_group and serializes ioConfig at the top level, not under spec. The round-trip test therefore fails on the type assertion, and still fails on the topic/group assertions after that is fixed; the malformed-json test below also uses the wrong type/shape, so it does not exercise the intended missing-field path.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexTaskJsonSubmitIT.java
```diff
@@ -0,0 +1,244 @@
...
+    kafkaServer.publishRecordsToTopic(topic, generateRecords(numRecords));
+
+    indexer.latchableEmitter().waitForEventAggregate(
+        event -> event.hasMetricName("ingest/events/processed")
+                      .hasDimension(DruidMetrics.DATASOURCE, dataSource),
+        agg -> agg.hasSumAtLeast(numRecords)
+    );
+    cluster.callApi().waitForAllSegmentsToBeAvailable(dataSource, coordinator, broker);
+
+    final long deadlineMs = System.currentTimeMillis() + 30_000L;
```
FrankChen021: [P3] Fail when the row-count wait expires

If verifySqlQuery never observes the expected count, this loop exits when the deadline expires and the test continues to cancel the task without failing. A broken JSON-submit ingestion/query path can pass after 30 seconds. Track the last assertion and fail after the deadline, or reuse the existing assertRowCountEventually pattern.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 1 |
| Total | 3 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 1 |
| Total | 3 |

---

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupPartitionRebalancingIT.java
```diff
@@ -0,0 +1,259 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Verifies that the Kafka broker automatically rebalances new partitions to share-group consumers
+ * without any Druid-side intervention (no supervisor, no task restart).
```
Shekharrajak, line 53: without any Druid-side intervention (no supervisor, no task restart) - Main advantages of using share group - no stop the world case.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexerCrashIT.java
```diff
@@ -0,0 +1,272 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Verifies at-least-once semantics: uncommitted records are redelivered after Indexer crash.
+ */
```
Shekharrajak, line 53: Validation of no data loss in crash situation.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexTaskJsonSubmitIT.java
Shekharrajak: https://github.com/apache/druid/pull/19311/changes/872d6f3b93031491ed35811765e0291bcbf98bd6#diff-c24dbac31b9f7e4062cd05cd6260d05c23c1c49b1c15ab8df3f9813efd806cb6

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexTaskJsonSubmitIT.java
Shekharrajak: https://github.com/apache/druid/pull/19311/changes/872d6f3b93031491ed35811765e0291bcbf98bd6#diff-c24dbac31b9f7e4062cd05cd6260d05c23c1c49b1c15ab8df3f9813efd806cb6

FrankChen021 (commented):
Reviewed all 26 changed files in this follow-up. The JSON-submit payload expectation and row-count timeout threads are addressed by the current changes, and I did not find any new high-confidence PR-caused issues.

---

Shekharrajak:
fixing one unit test : 

```
mbedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/ShareGroupIndexTaskJsonSubmitIT.class:96 | org.apache.druid.rpc.HttpResponseException: Server error [400 Bad Request]; body: {"error":"Please make sure to load all the necessary extensions and jars with type 'ShareGroupIndexTask'. Could not resolve type id 'ShareGroupIndexTask' as a subtype of `org.apache.druid.indexing.common.task.Task` known type ids = [archive, compact, index, index_hadoop, index_kafka, index_kafka_share_group, index_parallel, index_sub, kill, move, noop, partial_dimension_cardinality, partial_dimension_distribution, partial_index_generate, partial_index_generic_merge, partial_range_index_generate, query_controller, query_worker, restore, single_phase_sub_task]"} 0.153s
  ℹ️ Posting with conclusion 'failure' to https://github.com/apache/druid/pull/19311 (sha: ad1f11ac54ba82b18a9bee9c1d37e4a853b2e36b)
🚀 Publish results
  ℹ️ - Test Report - test-jdk25-[S*] - 38737 tests run, 38719 passed, 13 skipped, 5 failed.
     🧪 - embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/ShareGroupPartitionRebalancingIT.class | Expected at least [20] rows but got [10] ==> expected: <true> but was: <false>
     🧪 - embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/ShareGroupPartitionRebalancingIT.class | Expected at least [30] rows but got [10] ==> expected: <true> but was: <false>
     🧪 - embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/ShareGroupIndexerCrashIT.class | Timed out waiting for event after [60,000]ms
     🧪 - embedded-tests/target/test-classes/org/apache/druid/testing/embedded/indexing/ShareGroupIndexTaskJsonSubmitIT.class | org.apache.druid.rpc.HttpResponseException: Server error [400 Bad Request]; body: {"error":"Please make sure to load all the necessary extensions and jars with type 'ShareGroupIndexTask'. Could not resolve type id 'ShareGroupIndexTask' as a subtype of `org.apache.druid.indexing.common.task.Task` known type ids = [archive, compact, index, index_hadoop, index_kafka, index_kafka_share_group, index_parallel, index_sub, kill, move, noop, partial_dimension_cardinality, partial_dimension_distribution, partial_index_generate, partial_index_generic_merge, partial_range_index_generate, query_controller, query_worker, restore, single_phase_sub_task]"}

```

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Reviewed 26 of 26 changed files.

---

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexerCrashIT.java
```diff
@@ -0,0 +1,267 @@
...
+
+    indexer.stop();
+    indexer.start();
+
+    // Task 2 (same group): broker must redeliver all numRecords because none were acked.
+    final String taskId2 = submitTask(topic, GROUP_ID + "-noack");
+    Thread.sleep(SHARE_CONSUMER_READY_DELAY_MS);
+
+    // Give Task 2 time to consume redelivered records before cancel.
+    Thread.sleep(15_000L);
```
FrankChen021: [P2] Crash tests no longer prove redelivery

Replacing the post-restart waitForEventAggregate checks with fixed sleeps means these crash tests no longer assert that Task 2 consumed any redelivered records. Task 1 is allowed to run for 3 seconds before indexer.stop(), and this runner publishes and acks every batch, so it can already make the final SQL row count pass; in the first test the new batch can also satisfy the count. A regression that fails to redeliver unacknowledged records could now slip through. Please wait on a post-restart ingest signal or otherwise assert rows were processed by the restarted task before cancelling it.

Shekharrajak: Updated https://github.com/apache/druid/pull/19311/changes/94cbff5b73dfb8bf88101cd26ed4fe431b41f6ea

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/common/AcknowledgeTypeTest.java
```diff
@@ -42,4 +49,69 @@ public void testValueOfRoundTripsForEachConstant()
...
+          }
+
+          @Override
+          public void close()
+          {
+          }
+        };
+
+    stub.wakeup();
+    Assert.assertFalse(stub.acquisitionLockTimeoutMs().isPresent());
```
Shekharrajak, line 115: Just to have test coverage CI check pass.

FrankChen021 (commented):
Handled the `ShareGroupIndexerCrashIT` follow-up. The latest update replaces fixed post-restart sleeps with processed-event baselines/waits before cancelling Task 2, so I don't have a further inline reply.

Reviewed 1 of 26 changed files.

---

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
```diff
@@ -0,0 +1,497 @@
...
+        for (OrderedPartitionableRecord<KafkaTopicPartition, Long, KafkaRecordEntity> record : records) {
+          batchOffsets.computeIfAbsent(record.getPartitionId(), k -> new ArrayList<>())
+                      .add(record.getSequenceNumber());
+
+          final List<InputRow> rows = chunkReader.parse(record.getData(), false);
+
+          for (InputRow row : rows) {
+            final AppenderatorDriverAddResult addResult = driver.add(
+                row,
+                SEQUENCE_NAME,
```
FrankChen021: [P1] Use per-task sequence names for segment allocation

Every `ShareGroupIndexTaskRunner` uses the same hard-coded sequence name when adding rows. Druid segment allocation is idempotent for the same `sequenceName`, and this call also skips lineage checks, so concurrent share-group tasks ingesting the same datasource interval can be handed the same pending segment id. If one task publishes that segment first, another task can treat the id as already published and then acknowledge its Kafka records even though its rows were not registered. Make the allocation sequence unique per task or otherwise isolate concurrent producers before supporting multiple tasks in one share group.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Reviewed 26 of 26 changed files.

---

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
Shekharrajak: Thanks for checking .

FIx : https://github.com/apache/druid/pull/19311/changes/bf4b2ec9a66b2c67566f9a76e44e4be6758a20b9
tests: https://github.com/apache/druid/pull/19311/changes/cd7873b72143961e30e8749d02e4bf02a15aabde

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunnerTest.java
```diff
@@ -271,6 +272,61 @@ public void testRunLoop_segmentAllocationFails_throwsAndExits() throws Exception
...
+    final ShareGroupIndexTask taskA = buildTask("share_group_task_A");
+    final ShareGroupIndexTask taskB = buildTask("share_group_task_B");
+    final java.util.Set<String> seqNamesUsedByA = runOnceAndCaptureSequenceNames(taskA);
+    final java.util.Set<String> seqNamesUsedByB = runOnceAndCaptureSequenceNames(taskB);
+    Assert.assertFalse("Task A should call driver.add at least once", seqNamesUsedByA.isEmpty());
+    Assert.assertFalse("Task B should call driver.add at least once", seqNamesUsedByB.isEmpty());
+    final java.util.Set<String> shared = new java.util.HashSet<>(seqNamesUsedByA);
+    shared.retainAll(seqNamesUsedByB);
+    Assert.assertTrue(
+        "Concurrent share-group tasks must not share segment sequence names; shared=" + shared,
```
Shekharrajak, line 287: validating that seq name is different for both task.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
```diff
@@ -121,6 +121,7 @@ public class ShareGroupIndexTaskRunner
     this.task = task;
     this.toolbox = toolbox;
     this.configMapper = configMapper;
+    this.sequenceName = task.getId();
```
Shekharrajak, line 124: task ID is unique so using it for sequence name @FrankChen021

Shekharrajak:
One CI check failing - looks like network error. Please help in re-trigger

FrankChen021 (commented):
Follow-up handled. The sequence-name issue is addressed in the current code: `ShareGroupIndexTaskRunner` now uses `task.getId()` for the appenderator sequence name in both add and publish paths, and the regression test covers concurrent tasks using distinct sequence names. I did not find any new high-confidence PR-caused issues.

Reviewed 27 of 27 changed files.

---

FrankChen021 (commented):
I have reviewed the code for correctness, edge cases, concurrency, lifecycle/cleanup/retry risks, API compatibility, and missing-test risks; no issues found.

Reviewed 27 of 27 changed files.

---

Shekharrajak:
Hi @FrankChen021 , We have added IT , unit tests : embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedShareGroupIngestionTest.java - please let me know if we can get this PR in and Druid will be among first who have share group implementation available and then focus on enhancements.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/common/AcknowledgingRecordSupplier.java
```diff
@@ -0,0 +1,99 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.indexing.seekablestream.common;
+
+import org.apache.druid.data.input.impl.ByteEntity;
+
+import javax.validation.constraints.NotNull;
```
FrankChen021: [P1] Use Jakarta validation imports

**Finding:** This new interface imports `javax.validation.constraints.NotNull`, but the current indexing-service declares only `jakarta.validation-api` and the existing `RecordSupplier` uses the Jakarta package. The same invalid import is present in `KafkaShareGroupRecordSupplier.java`; neither source can compile because the javax validation package is unavailable.

**Suggestion:** Change both new `NotNull` imports to the repository's `jakarta.validation.constraints.NotNull` package.

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaShareGroupRecordSupplierTest.java
```diff
@@ -0,0 +1,421 @@
...
+import org.apache.druid.indexing.seekablestream.common.AcknowledgeType;
+import org.apache.druid.indexing.seekablestream.common.OrderedPartitionableRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.KafkaShareConsumer;
+import org.apache.kafka.common.KafkaException;
+import org.apache.kafka.common.TopicIdPartition;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.Uuid;
+import org.junit.After;
```
FrankChen021: [P1] Keep share-group tests on the configured JUnit platform

**Finding:** This and the other new share-group unit tests import JUnit 4 (`org.junit.*`), but the current `kafka-indexing-service` and `indexing-service` POMs declare only JUnit Jupiter; the current dependency update also removed `junit:junit` and the Vintage engine. The six new unit-test sources therefore cannot compile or be discovered in the module's current test setup.

**Suggestion:** Migrate all six new unit tests to JUnit Jupiter, or explicitly restore JUnit 4 plus the Vintage engine for these modules.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
```diff
@@ -0,0 +1,498 @@
...
+    log.info("Starting share group ingestion loop for topic[%s], group[%s].",
+             ioConfig.getTopic(), ioConfig.getGroupId());
+
+    long totalRowsIngested = 0;
+    boolean lockTimeoutLogged = false;
+    boolean appenderatorClosedNormally = false;
+
+    try {
+      recordSupplier.subscribe(Collections.singleton(ioConfig.getTopic()));
+      driver.startJob(segmentId -> true);
```
FrankChen021: [P2] Reacquire locks for restored segments

**Finding:** `StreamAppenderatorDriver.startJob` calls its lock helper for every restored non-pushed segment, but this lambda returns true without acquiring or validating any lock. On a same-worker retry with persisted `persist/` state, the subsequent `SegmentTransactionalAppendAction` lock-coverage check can reject those restored segments, leaving their records unacknowledged and making recovery repeatedly fail.

**Suggestion:** Pass a lock helper that reacquires the configured segment/time-chunk lock, as `SeekableStreamIndexTaskRunner` does, and cover the restored-segment retry path.

FrankChen021 (commented):
## 🟡 Changes recommended

This updated-head review found two build/test compatibility blockers and one recovery-path issue. The new acknowledging supplier sources still import the unavailable javax validation package; the new unit tests still use JUnit 4 after the current module dependencies became JUnit 5-only. In addition, restored appenderator segments are reported as locked without actually reacquiring their task locks, which can strand persisted rows after a same-worker retry.

Reviewed 27 of 27 changed files. I started with the incremental diff from c289f31d25cd360fff4b8c78ce4685c05e6a04dc, isolated the current PR paths from the merge artifact, then rechecked the full current merge-base diff and surrounding Kafka share-consumer, appenderator, lock, task lifecycle, and test configuration code.

Validation: targeted static repository searches, read-only source/API inspection, and `git diff --check e3f06eac3ec902937f72caa31871cf97d5a1df12..48a40e04634ac5bd3e8b2f22ac2a3454c68197f9` (passed). No builds, tests, dependency installs, formatters, fetches, cleanup, or worktree mutations were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 1 |
| P3 | 0 |
| Total | 3 |



---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/common/AcknowledgingRecordSupplier.java
Shekharrajak: Thanks for checking https://github.com/apache/druid/pull/19311/commits/fd1b9af8e1da39f24eefac9a1d3a8e0d24fb2c8d

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaShareGroupRecordSupplierTest.java
FrankChen021: P1 Use JUnit 5 APIs for the new unit tests

**Finding:** This new test source imports org.junit.* (including the import at this line), and the same JUnit 4 API is used by ShareGroupConsumerPropertiesTest, ShareGroupIndexTaskIOConfigTest, ShareGroupIndexTaskRunnerTest, ShareGroupIndexTaskTest, and indexing-service's AcknowledgeTypeTest. The kafka-indexing-service and indexing-service POMs provide only JUnit Jupiter dependencies and no JUnit 4 or Vintage artifact, so normal test compilation cannot resolve these imports and the added tests cannot run.

**Suggestion:** Migrate all six new unit-test classes to the repository's JUnit 5 APIs, lifecycle annotations, and assertions, or add an explicitly supported JUnit 4/Vintage dependency if that compatibility is intentional.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
FrankChen021: P2 Reacquire locks for restored appenderator segments

**Finding:** On a restart or retry with persisted appenderator state, StreamAppenderatorDriver.startJob invokes the supplied lock helper for every restored non-dropped segment. This unconditional true reports each segment as locked without acquiring the task's current append/time-chunk lock, so the task can resume or publish rows under a lock it does not own; when the previous lock is gone, the restored segment can be rejected or stranded instead of being safely resumed.

**Suggestion:** Pass a lock helper that acquires and verifies the same lock granularity and type used by SegmentAllocateAction, matching the existing SeekableStreamIndexTaskRunner startup path, and fail restoration when acquisition fails.

FrankChen021 (commented):
## 🟡 Changes recommended

This is a current-head review for the reply-needed follow-up. The original javax.validation finding is resolved: both new acknowledging-supplier sources now use jakarta.validation.constraints.NotNull. Two current-head issues remain: the six new unit-test classes use JUnit 4 while their modules declare only JUnit Jupiter, and the appenderator restore callback reports success without reacquiring task locks.

Reviewed 27 of 27 changed files:
- docs/ingestion/kafka-share-group-ingestion.md
- embedded-tests/pom.xml
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedShareGroupIngestionTest.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupConcurrentTasksIT.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexTaskJsonSubmitIT.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexerCrashIT.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupKafkaResource.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupPartitionRebalancingIT.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupProbeTest.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupTaskRestartIT.java
- extensions-core/kafka-indexing-service/pom.xml
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskModule.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaShareGroupRecordSupplier.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupConsumerProperties.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTask.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskIOConfig.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaShareGroupRecordSupplierTest.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/ShareGroupConsumerPropertiesTest.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskIOConfigTest.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunnerTest.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskTest.java
- indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/StreamChunkReader.java
- indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/common/AcknowledgeType.java
- indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/common/AcknowledgingRecordSupplier.java
- indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/common/AcknowledgeTypeTest.java
- website/.spelling

Validation: targeted read-only repository searches and git diff --check e3f06eac3ec902937f72caa31871cf97d5a1df12..fd1b9af8e1da39f24eefac9a1d3a8e0d24fb2c8d (passed). No builds, tests, dependency installs, formatters, fetches, cleanup, or worktree mutations were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |



---

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
Shekharrajak: Updated https://github.com/apache/druid/pull/19311/commits/8341dc6f352eea13572598e196800c425f8d94b8

FrankChen021 (commented):
## 🟡 Changes recommended

Follow-up assessment: the author’s latest commit replaces the unconditional restore callback with lock acquisition and revocation checks, so the supplied restore-lock finding is resolved. The current head still has two P2 configuration/lock-contract failures and one P3 documentation mismatch; the P2 issues can prevent valid task configurations from restoring or publishing batches, so changes are recommended before merge.

Reviewed exactly 27 changed files in the prepared PR worktree and compared both the full merge-base diff and the incremental diff since the previous reviewed SHA. Coverage:

- docs/ingestion/kafka-share-group-ingestion.md
- embedded-tests/pom.xml
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/EmbeddedShareGroupIngestionTest.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupConcurrentTasksIT.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexTaskJsonSubmitIT.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupIndexerCrashIT.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupKafkaResource.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupPartitionRebalancingIT.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupProbeTest.java
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/ShareGroupTaskRestartIT.java
- extensions-core/kafka-indexing-service/pom.xml
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskModule.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaShareGroupRecordSupplier.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupConsumerProperties.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTask.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskIOConfig.java
- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaShareGroupRecordSupplierTest.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/ShareGroupConsumerPropertiesTest.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskIOConfigTest.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunnerTest.java
- extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskTest.java
- indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/StreamChunkReader.java
- indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/common/AcknowledgeType.java
- indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/common/AcknowledgingRecordSupplier.java
- indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/common/AcknowledgeTypeTest.java
- website/.spelling

Validation run: `git diff --check e3f06eac3ec902937f72caa31871cf97d5a1df12...HEAD` and `git diff --check fd1b9af8e1da39f24eefac9a1d3a8e0d24fb2c8d...HEAD`; both passed. No builds, tests, dependency installs, formatters, or broad validations were run, per the prepared-context review scope.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 1 |
| Total | 3 |



---

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
```diff
@@ -0,0 +1,535 @@
...
+              Set<DataSegment> segmentsToPush,
+              @Nullable Object commitMetadata,
+              SegmentSchemaMapping segmentSchemaMapping
+          ) throws IOException
+          {
+            if (segmentsToPush.isEmpty()) {
+              return SegmentPublishResult.ok(segmentsToPush);
+            }
+            return toolbox.getTaskActionClient().submit(
+                SegmentTransactionalAppendAction.forSegments(segmentsToPush, segmentSchemaMapping)
```
FrankChen021, line 403: [P2] Select the publisher for the configured lock type

**Finding:** ShareGroupIndexTask only defaults useConcurrentLocks with putIfAbsent, so a task may explicitly set it false or choose useSharedLock/taskLockType and acquire EXCLUSIVE or SHARED locks. This publisher always submits SegmentTransactionalAppendAction, whose perform method rejects every task lock that is not APPEND, so the first non-empty batch fails during publish and its records are never acknowledged for that otherwise accepted task context.

**Suggestion:** Either reject or normalize non-APPEND lock contexts for this task, or select the publish action from the actual lock type as the existing batch-task path does; add coverage for useConcurrentLocks=false.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/ShareGroupIndexTaskRunner.java
```diff
@@ -0,0 +1,535 @@
...
+    return TaskStatus.success(task.getId());
+  }
+
+  private boolean acquireLockForRestoredSegment(SegmentIdWithShardSpec segmentId)
+  {
+    try {
+      if (lockGranularity == LockGranularity.SEGMENT) {
+        return toolbox.getTaskActionClient().submit(
+            new SegmentLockAcquireAction(
+                lockType,
```
FrankChen021, line 475: [P2] Avoid APPEND locks with segment granularity

**Finding:** When forceTimeChunkLock is false, this task still defaults useConcurrentLocks to true, making lockType APPEND. The segment restore branch passes that type to SegmentLockAcquireAction even though Druid permits APPEND only with time-chunk locking and TaskLockbox rejects an APPEND segment request when conflicting locks exist; SegmentAllocateAction receives the same unsupported pairing. A segment-granularity task can therefore fail to restore a persisted segment, whereas the existing SeekableStreamIndexTaskRunner uses EXCLUSIVE for this branch.

**Suggestion:** Use a supported lock-type/granularity pairing for segment mode and keep allocation, restoration, and publishing consistent, or explicitly reject forceTimeChunkLock=false; cover the configuration with a lockbox-backed test.

docs/ingestion/kafka-share-group-ingestion.md
```diff
@@ -0,0 +1,349 @@
...
+
+## IO configuration
+
+| Property | Type | Required | Default | Description |
+|----------|------|----------|---------|-------------|
+| `topic` | String | Yes | -- | Kafka topic to consume from. |
+| `groupId` | String | Yes | -- | Share group identifier. Multiple tasks with the same `groupId` share the workload. |
+| `consumerProperties` | Map | Yes | -- | Kafka consumer properties. Must include `bootstrap.servers`. See [Consumer property restrictions](#consumer-property-restrictions). |
+| `inputFormat` | Object | Yes | -- | Input format for parsing records (json, csv, avro, etc.). |
+| `pollTimeout` | Long | No | 2000 | Poll timeout in milliseconds. |
```
FrankChen021, line 95: [P3] Document the actual pollTimeout default

**Finding:** The configuration table says an omitted pollTimeout defaults to 2000 ms, but ShareGroupIndexTaskIOConfig uses KafkaSupervisorIOConfig.DEFAULT_POLL_TIMEOUT_MILLIS, which is 100 ms. Users relying on this documentation will poll twenty times more frequently than the implementation indicates in the docs.

**Suggestion:** Make the documentation match the implementation, or change the new configuration default and its test to the intended 2000 ms value.

PR #19317, opened 2026-04-15
feat: consolidate S3 client creation and enable ARN role for MSQ export

cecemei:
### Description

 Consolidates the creation of `ServerSideEncryptingAmazonS3` into a single static builder method to reduce code duplication and improve maintainability.

### Release Note
- Export query result to S3 now supports role ARN, e.x.
```
INSERT INTO
EXTERN(
  s3(bucket => 'cecemei-test2', prefix => 'export', assumeRoleArn => 'arn:aws:iam::00000:role/cecemei-test-20260520'))
AS CSV
SELECT ...
```
 
<hr>

This PR has:

- [x] been self-reviewed.


FrankChen021:

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/ServerSideEncryptingAmazonS3.java
```diff
@@ -274,6 +301,161 @@ public void upload(String bucket, String key, File file, @Nullable Grant aclGran
...
+    AssumeRoleRequest.Builder assumeRoleRequestBuilder =
+        AssumeRoleRequest.builder().roleArn(assumeRoleArn).roleSessionName(roleSessionName).durationSeconds(3600);
+    if (assumeRoleExternalId != null) {
+      assumeRoleRequestBuilder.externalId(assumeRoleExternalId);
+    }
+
+    return StsAssumeRoleCredentialsProvider.builder()
+                                           .stsClient(stsBuilder.build())
+                                           .refreshRequest(assumeRoleRequestBuilder.build())
+                                           .asyncCredentialUpdateEnabled(true)
```
FrankChen021, line 466: [P2] Avoid leaking assume-role refresh resources

When an MSQ export specifies assumeRoleArn, S3ExportStorageProvider.createStorageConnector builds a fresh ServerSideEncryptingAmazonS3 for each connector creation. That path creates an StsAssumeRoleCredentialsProvider with asyncCredentialUpdateEnabled(true) plus a new STS client, but neither the provider, STS client, nor S3 clients are lifecycle-managed or closed after the export connector is done. Since exports create connectors for the empty-location check, worker writes, and manifest writing, each ARN export can leave background refresh/client resources behind. Reuse a lifecycle-managed role-specific client/provider or make the connector own and close the resources it creates.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/output/S3ExportStorageProvider.java
```diff
@@ -93,7 +124,22 @@ public StorageConnector createStorageConnector(File taskTempDir)
         s3ExportConfig.getChunkSize(),
         s3ExportConfig.getMaxRetry()
     );
-    return new S3StorageConnector(s3OutputConfig, s3, s3UploadManager);
+    if (Strings.isNullOrEmpty(assumeRoleArn)) {
+      return new S3StorageConnector(s3OutputConfig, s3Client, s3UploadManager);
+    }
+    return new S3StorageConnector(
+        s3OutputConfig,
+        ServerSideEncryptingAmazonS3.builder(
```
gianm: Frank's automated review makes a good point about this here: https://github.com/apache/druid/pull/19317/files#r3161338202

Each of these `ServerSideEncryptingAmazonS3` is going to leak some resources: thread pools, connection pools, and the like. This was a pre-existing problem with `S3InputSource` when a `S3InputDataConfig` is provided (via the `properties` field). I think because we have an `S3InputSource` per file (due to splitting), we'd potentially even in `master` be creating a lot of `ServerSideEncryptingAmazonS3` when `properties` is set.

This PR isn't making it a ton worse, but it also isn't making it much better. I suppose it would be OK to merge it given that it's not a ton worse, but, could you please memoize the client in the constructor of `S3ExportStorageProvider` (similar to what `S3InputSource` does). That will at least allow it to be shared across calls to `createStorageConnector`.

Could you please also put some comments about the problem in `ServerSideEncryptingAmazonS3#builder`. Ultimately fixing it would I think involve making `ServerSideEncryptingAmazonS3` closeable, and arranging for `close` to actually be called, which would be a larger change.

cecemei: > Frank's automated review makes a good point about this here: https://github.com/apache/druid/pull/19317/files#r3161338202
> 
> Each of these `ServerSideEncryptingAmazonS3` is going to leak some resources: thread pools, connection pools, and the like. This was a pre-existing problem with `S3InputSource` when a `S3InputDataConfig` is provided (via the `properties` field). I think because we have an `S3InputSource` per file (due to splitting), we'd potentially even in `master` be creating a lot of `ServerSideEncryptingAmazonS3` when `properties` is set.
> 
> This PR isn't making it a ton worse, but it also isn't making it much better. I suppose it would be OK to merge it given that it's not a ton worse, but, could you please memoize the client in the constructor of `S3ExportStorageProvider` (similar to what `S3InputSource` does). That will at least allow it to be shared across calls to `createStorageConnector`.
> 
> Could you please also put some comments about the problem in `ServerSideEncryptingAmazonS3#builder`. Ultimately fixing it would I think involve making `ServerSideEncryptingAmazonS3` closeable, and arranging for `close` to actually be called, which would be a larger change.

Updated to use `Suppliers.memoize` and added javadoc for `ServerSideEncryptingAmazonS3.Builder.build()`. PTAL!

docs/tutorials/tutorial-extern.md
```diff
@@ -167,7 +167,7 @@ Druid supports Amazon S3 or Google Cloud Storage (GCS) as cloud storage destinat
     ```sql
     INSERT INTO
     EXTERN(
-      s3(bucket => 'your_bucket', prefix => 'prefix/to/files'))
+      s3(bucket => 'your_bucket', prefix => 'prefix/to/files', assumeRoleArn => 'arn:aws:iam::some-role'))
```
gianm, line 170: Please also update `docs/multi-stage-query/reference.md`. The table of `s3` options should include `assumeRoleArn` and `assumeRoleExternalId`.

gianm (commented):
Main code looks good to me, just some docs are missing.

FrankChen021 (commented):

---

gianm (approved):

PR #19302, opened 2026-04-13
test: convert tests in druid-processing to junit5 pt. 3

clintropolis:
next batch 

previous:
* #19289
* #19299

PR #19299, opened 2026-04-11
test: convert tests in druid-processing to junit5 pt. 2

clintropolis:
convert more tests in `druid-processing` to junit5, part 2 of n

PR #19289, opened 2026-04-10
test: convert tests in druid-processing to junit5 pt. 1

clintropolis:
There are still a bunch left to do, but I plan split this across multiple PRs so they don't get too big.

Issue #13948, opened 2023-03-19
JUnit4 to JUnit5 migration


UngureanuSerg:
Hi team,

A proposal to moving on after the adoption of java8+ is to start the migration from junit4 to jnuit5.

- JUnit 4 never advanced beyond Java 7, missing out on a lot of features from Java 8. JUnit 5 makes good use of the Java 8 features.
- A single jar library contains the entire framework. We need to import the whole library, even when we only require a particular feature. In JUnit 5, we get more granularity and can import only what's necessary.
- Only one test runner can execute tests at a time in JUnit 4 (e.g. @RunWith(Parameterized.class)
 or provide druid extension with the adoption of JUnit Platform). JUnit 5 allows multiple runners to work simultaneously.

As a strategy for the approach I would suggest to migrate one module at a time(it will be easier for the prs) and create tasks for the rest of the modules. As a starting point we can still depend on JUnit Vintage for backward compatibility with JUnit 4 but gradually moving to JUnit5.

I think all the point mentioned describes why it would be beneficial for apache druid as well as I would like to pick this issue if the rest of the community is interested.

cryptoe:
@UngureanuSerg Thanks for starting this. Yes lets start the migration. Pretty excited  🚀 .
IMHO `JUnit Vintage` should be our last option. If we are able to move to the JUnit 5 using `Unit Platform` dependencies that would be event cooler.

gianm:
> @UngureanuSerg Thanks for starting this. Yes lets start the migration. Pretty excited 🚀 .
> IMHO `JUnit Vintage` should be our last option. If we are able to move to the JUnit 5 using `Unit Platform` dependencies that would be event cooler.

I don't know the details, but if "JUnit Vintage" allows us to do a migration piece-by-piece instead of updating every test case at once, that sounds like a good thing. Updating every test case at once could potentially be a prohibitive amount of work for a single contributor.

UngureanuSerg:
Hey Gian I didn't know that junit-vintage give this kind of capabilities but I will check it for sure

dampcake:
In other projects I use junit-vintage to mix both JUnit5 and JUnit4 tests in the same suites. This allows for new tests or tests where you want to take advantage of the features to be converted while leaving all the existing tests on Junit4. Pretty much a best of both worlds scenario.

kgyrtkirk:
there was a related PR to this: #16106 ; 

[...]

PR #19327, opened 2026-04-15
test: convert tests in druid-processing to junit5 pt. 5

clintropolis:
next batch

previous:
* #19289
* #19299
* #19302
* #19313

### DIFF
--- a/processing/src/test/java/org/apache/druid/collections/RangeIntSetTest.java
+++ b/processing/src/test/java/org/apache/druid/collections/RangeIntSetTest.java
@@ -21,8 +21,8 @@ package org.apache.druid.collections;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import org.junit.Assert;
-import org.junit.Test;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 
 import java.util.Collections;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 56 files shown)


jtuglu1:
Curious – why not just do all-at-once?

clintropolis:
> Curious – why not just do all-at-once?

idk, I was thinking it would minimize things like #19326 if migration was broken into smaller chunks, but no idea if that is actually true or not. Also, the github UI was pretty sad the last time i looked at a huge PR, so was thinking if i kept changes to ~1k lines people might actually glance over things before approving to double check stuff :shrug:

I can do the rest of processing in one shot after this PR if people think that is better, I don't feel too strongly about it

capistrant:
> > Curious – why not just do all-at-once?
> 
> idk, I was thinking it would minimize things like #19326 if migration was broken into smaller chunks, but no idea if that is actually true or not. Also, the github UI was pretty sad the last time i looked at a huge PR, so was thinking if i kept changes to ~1k lines people might actually glance over things before approving to double check stuff 🤷
> 
> I can do the rest of processing in one shot after this PR if people think that is better, I don't feel too strongly about it

It is kind of nice to have it split up in the cases of files with more change than just the change to using `Assertions` everywhere. I have kind of been quick scanning those simple files and looking a little more deeply at ones that do more than that. In a big PR it would be tempting to just smash approve without looking or after looking at half :P

capistrant (approved):

FrankChen021:
Part of #13948 — JUnit4 to JUnit5 migration.

PR #19328, opened 2026-04-15
feat: Deprioritize turbo loading historicals for broker query routing

capistrant:
### Description

Arguably could be `fix` commit type I guess. But I think `feat` is better since it is a behavior change that comes from learnings running this code in Druid clusters and may have been more of a gap in the initial design spec.

When a historical is in turbo mode it is going to be eagerly loading segments on more threads than normal runtime. This can tie up CPU and cause contention for queries. This contention is called out in our docs so it was initially acknowledged as being a reality of using turbo loading.

Upon reflection, I am proposing that the Broker takes turbo loading nodes and artificially de-prioritizes them in its historical selector structure. This will allow queries to target non-turbo nodes whenever such hosts contain the targeted segments for a query. This impl sends turbo nodes to dead last priority across all tiers. This means that if you have 1 replica on hot and one on a slower cold. if that hot is in turbo mode you will query cold. An argument could be made that this is wrong (called out below in design decisions)

##### Sending turbo nodes to dead last priority

You could argue that you may only want a turbo node to be the least priority server within its own tier or existing priority level. That way if you have cold storage type historicals that are super low priority you'd still query the turbo hot replica over that one, thinking that it may still be better faster.

The reason I did not go this direction is a feel that it was unnecessary complexity. Turbo mode is designed to be transient and I feel it is tolerable to have the simpler solution of dead last priority for some (hopefully) short period of time where your queries may hit slower nodes if that is where replicas live. If the turbo nodes truly are pegging CPU it is hopefully better to do it my proposed way anyways.

#### Release note

Turbo loading historical nodes are now deprioritized by brokers when routing queries. If there are non-turbo nodes serving replicas a broker needs to query, the query will be dispatched there instaed.

<hr>

 * `BrokerViewOfCoordinatorConfig`

<hr>

This PR has:

- [x] been self-reviewed.


adarshsanjeev (approved):
I think the changes here shouldn't cause any issues.
I believe the intention initially was to have turboLoading be experimentally used only with cloning, and a change like this would be necessary if that assumption is not being followed.

capistrant:
> I think the changes here shouldn't cause any issues. I believe the intention initially was to have turboLoading be experimentally used only with cloning, and a change like this would be necessary if that assumption is not being followed.

ya, I think the case has been that when operators are doing maintenance, or adding nodes without being clones they see the big load queue and turn on turbo. which I think is reasonable, just maybe not foreseen by the initial turbo/clone design pairing. ty for review!

PR #19341, opened 2026-04-16
fix: Avoid early cleanup of Dart workers.

gianm:
In #19233 controllers were updated to be canceled using a combination of stop + direct thread interrupt. This patch makes worker cancellation work the same way.

The change is necessary to fix a problem where the worker context could be closed before the worker finishes running, since the worker context is closed using a listener on the run future. To ensure proper sequencing, the run future must not be directly cancelable, since that would cause the listener to fire too early.


Fly-Style (approved):
Good finding!

PR #19343, opened 2026-04-17
fix: Console handling for tables without intervals.

gianm:
Since #19264, the "intervals" field is omitted from table specs when it is ETERNITY (i.e. no time filter). This caused a crash in the web console, which is fixed by this patch.


capistrant (approved):
ty!

PR #19350, opened 2026-04-18
minor: add getServiceDims() getter to ServiceMetricEvent

abhishekrb19:
## Summary
  Adds a `getServiceDims()` getter method to `ServiceMetricEvent` to provide direct access to the immutable service dimensions map. This can be used by custom extensions along with the existing getters in the class.

  ## Motivation
  Previously, callers could access individual service dimension values via `getService()` and `getHost()`, but had no way to retrieve the complete service dimensions map. This getter provides direct access to the immutable map of all service
  dimensions by custom extensions.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
+++ b/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
@@ -93,7 +93,7 @@ public class ServiceMetricEvent implements Event
     return serviceDims.get(HOST);
   }
 
-  public ImmutableMap<String, String> getServiceDims()
+  public Map<String, String> getServiceDims()
   {
     return serviceDims;
   }
--- a/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
+++ b/processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
@@ -93,6 +93,11 @@ public class ServiceMetricEvent implements Event
     return serviceDims.get(HOST);
   }
 
+  public ImmutableMap<String, String> getServiceDims()
+  {
+    return serviceDims;
+  }
+
   public Map<String, Object> getUserDims()
   {
     return userDims;
--- a/processing/src/test/java/org/apache/druid/java/util/emitter/service/ServiceMetricEventTest.java
+++ b/processing/src/test/java/org/apache/druid/java/util/emitter/service/ServiceMetricEventTest.java
@@ -76,6 +76,9 @@ public class ServiceMetricEventTest
         builderEvent.toMap()
     );
 
+    Assert.assertEquals("test", builderEvent.getServiceDims().get("service"));
+    Assert.assertEquals("localhost", builderEvent.getServiceDims().get("host"));
+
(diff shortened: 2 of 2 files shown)


processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
```diff
@@ -93,6 +93,11 @@ public String getHost()
     return serviceDims.get(HOST);
   }
 
+  public ImmutableMap<String, String> getServiceDims()
```
utafrali: The return type `ImmutableMap<String, String>` leaks the concrete implementation, which is inconsistent with `getUserDims()` in the same class — that method returns `Map<String, Object>` even though its backing field is also an `ImmutableMap`. For a public API, prefer returning the interface type:

```java
public Map<String, String> getServiceDims()
{
  return serviceDims;
}
```

Callers who need the immutability guarantee can check `instanceof ImmutableMap` or the Javadoc can document it. Returning the concrete type here makes it harder to swap the implementation later.

processing/src/test/java/org/apache/druid/java/util/emitter/service/ServiceMetricEventTest.java
```diff
@@ -76,6 +76,9 @@ public void testBuilder()
         builderEvent.toMap()
     );
 
+    Assert.assertEquals("test", builderEvent.getServiceDims().get("service"));
```
utafrali, line 79: The string literals `"service"` and `"host"` duplicate the key constants already defined in the class (used inside `getService()` and `getHost()`). If those constants (`SERVICE`, `HOST`) are accessible from the test, use them directly to avoid silent drift if the keys ever change:

```java
Assert.assertEquals("test", builderEvent.getServiceDims().get(ServiceMetricEvent.SERVICE));
Assert.assertEquals("localhost", builderEvent.getServiceDims().get(ServiceMetricEvent.HOST));
```

If the constants are private, at minimum add a comment explaining why literals are used here.

utafrali (commented):
The getter is functionally correct and well-tested, but returning `ImmutableMap<String, String>` is inconsistent with the existing `getUserDims()` which returns the `Map` interface — both fields are backed by `ImmutableMap`, yet the API surface is different. The return type should be `Map<String, String>` to keep the class API coherent.

processing/src/main/java/org/apache/druid/java/util/emitter/service/ServiceMetricEvent.java
abhishekrb19: Yes, that's a good point - thanks for calling that out!

processing/src/test/java/org/apache/druid/java/util/emitter/service/ServiceMetricEventTest.java
abhishekrb19: “service” and “host” are [documented](https://druid.apache.org/docs/latest/operations/metrics/) metric dimensions. Hence, they cannot change due to backward compatibility guarantees, and this test and other existing ones that assert on the literals would catch that unlikely drift.

aho135 (approved):

PR #19353, opened 2026-04-18
perf: vectorize topN native engine

jtuglu1:
### Description

Vectorizes the TopN native query engine. More work should be done to make TopN spilling like GroupBy, so it can use a more efficient, fixed-size, off-heap buffer to back the memory for the aggregation. For now, I'm using a heap-backed grouper similar to HashVectorGrouper in the GroupBy engine. I'm also not sure about the state of things in Dart/MSQe. Speed ups of roughly ~20-50% across the board.

#### Benchmarks

Run on my Apple M3 Pro, 12 physical cores, 18GB mem

```
Benchmark                                  (indexType)  (numSegments)  (queryGranularity)  (rowsPerSegment)        (schemaAndQuery)  (threshold)  (vectorize)  Mode  Cnt        Score        Error  Units
TopNBenchmark.queryMultiQueryableIndex             N/A              1                 all            750000                 basic.A           10        false  avgt    5    69122.980 ±   4462.770  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                 all            750000                 basic.A           10        force  avgt    5    54046.641 ±   4367.487  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                 all            750000       basic.numericSort           10        false  avgt    5   103399.888 ±  44324.233  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                 all            750000       basic.numericSort           10        force  avgt    5    87322.720 ±  23678.746  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                 all            750000  basic.alphanumericSort           10        false  avgt    5    75924.740 ±  14120.944  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                 all            750000  basic.alphanumericSort           10        force  avgt    5    66410.182 ±  11979.505  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                hour            750000                 basic.A           10        false  avgt    5   133947.144 ±  50106.898  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                hour            750000                 basic.A           10        force  avgt    5    96217.492 ±   8871.746  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                hour            750000       basic.numericSort           10        false  avgt    5   557642.770 ± 314507.572  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                hour            750000       basic.numericSort           10        force  avgt    5   389809.531 ±  33972.861  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                hour            750000  basic.alphanumericSort           10        false  avgt    5   435965.653 ± 137802.112  us/op
TopNBenchmark.queryMultiQueryableIndex             N/A              1                hour            750000  basic.alphanumericSort           10        force  avgt    5   386555.540 ±   6790.380  us/op
TopNBenchmark.querySingleIncrementalIndex       onheap            N/A                 all            750000                 basic.A           10        false  avgt    5  2082680.574 ± 656078.762  us/op
TopNBenchmark.querySingleIncrementalIndex       onheap            N/A                 all            750000       basic.numericSort           10        false  avgt    5   165785.912 ±  22669.944  us/op
TopNBenchmark.querySingleIncrementalIndex       onheap            N/A                 all            750000  basic.alphanumericSort           10        false  avgt    5   155416.663 ±  77528.414  us/op
TopNBenchmark.querySingleIncrementalIndex       onheap            N/A                hour            750000                 basic.A           10        false  avgt    5  2102892.546 ± 634413.985  us/op
TopNBenchmark.querySingleIncrementalIndex       onheap            N/A                hour            750000       basic.numericSort           10        false  avgt    5   302456.169 ±  88204.908  us/op
TopNBenchmark.querySingleIncrementalIndex       onheap            N/A                hour            750000  basic.alphanumericSort           10        false  avgt    5   347785.978 ± 247907.083  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                 all            750000                 basic.A           10        false  avgt    5    78131.986 ±  12065.964  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                 all            750000                 basic.A           10        force  avgt    5    56115.802 ±   4901.144  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                 all            750000       basic.numericSort           10        false  avgt    5    44441.229 ±  17818.964  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                 all            750000       basic.numericSort           10        force  avgt    5    29320.550 ±   2044.562  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                 all            750000  basic.alphanumericSort           10        false  avgt    5    40142.959 ±  20420.132  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                 all            750000  basic.alphanumericSort           10        force  avgt    5    32129.217 ±   2168.716  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                hour            750000                 basic.A           10        false  avgt    5    86430.593 ±   8643.281  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                hour            750000                 basic.A           10        force  avgt    5    78887.059 ±  31017.995  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                hour            750000       basic.numericSort           10        false  avgt    5   132638.263 ±  11458.490  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                hour            750000       basic.numericSort           10        force  avgt    5   132583.761 ±  24221.660  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                hour            750000  basic.alphanumericSort           10        false  avgt    5   180629.884 ±  59802.573  us/op
TopNBenchmark.querySingleQueryableIndex            N/A              1                hour            750000  basic.alphanumericSort           10        force  avgt    5   181480.345 ±   6770.964  us/op
```

#### Release note

Vectorize TopN native query engine.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/HeapVectorGrouper.java
+++ b/processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/HeapVectorGrouper.java
@@ -1,0 +1,218 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.query.groupby.epinephelinae;
+
+import it.unimi.dsi.fastutil.Hash;
+import it.unimi.dsi.fastutil.objects.Object2IntMap;
+import it.unimi.dsi.fastutil.objects.Object2IntOpenCustomHashMap;
+import org.apache.datasketches.memory.Memory;
+import org.apache.druid.java.util.common.ISE;
+import org.apache.druid.java.util.common.parsers.CloseableIterator;
+import org.apache.druid.query.aggregation.AggregatorAdapters;
+import org.apache.druid.query.groupby.epinephelinae.collection.MemoryPointer;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.Arrays;
+import java.util.Iterator;
+
+/**
+ * On-heap {@link VectorGrouper} that grows aggregator state on demand, up to maximum limit of 2GB.
+ *
+ * Vectorized analogue of {@link org.apache.druid.query.topn.BaseTopNAlgorithm}'s
+ * {@code runWithCardinalityUnknown} path: used when dimension cardinality is unknown (numeric columns,
+ * non-dict-encoded string virtual columns) or when a dict-encoded string column's cardinality exceeds
+ * the processing buffer. Memory footprint is on-heap and grows with the distinct-key count — matching
+ * the non-vectorized path's memory profile for the same queries.
+ */
+public class HeapVectorGrouper implements VectorGrouper
+{
+  private static final Hash.Strategy<byte[]> BYTE_ARRAY_HASH_STRATEGY = new Hash.Strategy<byte[]>()
+  {
+    @Override
+    public int hashCode(byte[] o)
+    {
+      return Arrays.hashCode(o);
+    }
+
+    @Override
+    public boolean equals(byte[] a, byte[] b)
+    {
+      return Arrays.equals(a, b);
+    }
+  };
+
+  private static final int MIN_INITIAL_STATE_BUFFER_SIZE = 4096;
+
+  private final AggregatorAdapters aggregators;
+  private final int keySize;
+  private final int aggStateSize;
+  private final Object2IntOpenCustomHashMap<byte[]> keyToOffset;
+
+  private boolean initialized;
+  private ByteBuffer aggStateBuffer;
+  private int aggStateEnd;
+
+  private int[] vAggregationPositions;
+  private int[] vAggregationRows;
+  private byte[] keyScratch;
+
+  public HeapVectorGrouper(final AggregatorAdapters aggregators, final int keySize)
+  {
+    this.aggregators = aggregators;
+    this.keySize = keySize;
+    this.aggStateSize = aggregators.spaceNeeded();
+    this.keyToOffset = new Object2IntOpenCustomHashMap<>(BYTE_ARRAY_HASH_STRATEGY);
+    this.keyToOffset.defaultReturnValue(-1);
+  }
+
+  @Override
+  public void initVectorized(final int maxVectorSize)
+  {
+    if (initialized) {
+      if (vAggregationPositions.length != maxVectorSize) {
+        throw new ISE(
+            "initVectorized called with different maxVectorSize (existing=%d, new=%d)",
+            vAggregationPositions.length,
+            maxVectorSize
+        );
+      }
+      return;
+    }
+    this.aggStateBuffer = ByteBuffer.allocate(MIN_INITIAL_STATE_BUFFER_SIZE);
+    this.vAggregationPositions = new int[maxVectorSize];
+    this.vAggregationRows = new int[maxVectorSize];
+    this.keyScratch = new byte[keySize];
+    this.aggStateEnd = 0;
+    this.initialized = true;
+  }
+
+  /**
+   * Contract: keys for rows [startRow, endRow) must be packed contiguously at {@code keySpace[0 ..
+   * numRows * keySize)}; {@code startRow}/{@code endRow} are source-vector indices used to look up aggregator
+   * input values.
+   */
+  @Override
+  public AggregateResult aggregateVector(final Memory keySpace, final int startRow, final int endRow)
+  {
+    final int numRows = endRow - startRow;
+
+    for (int i = 0; i < numRows; i++) {
+      keySpace.getByteArray((long) i * keySize, keyScratch, 0, keySize);
+      int offset = keyToOffset.getInt(keyScratch);
+      if (offset == -1) {
+        if ((long) aggStateEnd + aggStateSize > aggStateBuffer.capacity()) {
+          growBuffer((long) aggStateEnd + aggStateSize);
+        }
+        offset = aggStateEnd;
+        final byte[] keyCopy = Arrays.copyOf(keyScratch, keySize);
+        keyToOffset.put(keyCopy, offset);
+        aggregators.init(aggStateBuffer, offset);
+        aggStateEnd += aggStateSize;
+      }
+      vAggregationPositions[i] = offset;
+    }
+
+    aggregators.aggregateVector(
+        aggStateBuffer,
+        numRows,
+        vAggregationPositions,
+        Groupers.writeAggregationRows(vAggregationRows, startRow, endRow)
+    );
+
+    return AggregateResult.ok();
+  }
+
+  private void growBuffer(final long neededCapacity)
+  {
+    if (neededCapacity > Integer.MAX_VALUE) {
+      throw new ISE("Aggregator state exceeds 2 GB; cardinality too high for HeapVectorGrouper");
+    }
+    int newCapacity = aggStateBuffer.capacity();
+    while (newCapacity < neededCapacity) {
+      final long doubled = (long) newCapacity * 2;
+      newCapacity = doubled > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) doubled;
+    }
+
+    final ByteBuffer oldBuffer = aggStateBuffer;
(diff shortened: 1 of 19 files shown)


processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/HeapVectorGrouper.java
```diff
@@ -0,0 +1,213 @@
...
+        vAggregationPositions,
+        Groupers.writeAggregationRows(vAggregationRows, startRow, endRow)
+    );
+
+    return AggregateResult.ok();
+  }
+
+  private void growBuffer(final long neededCapacity)
+  {
+    if (neededCapacity > Integer.MAX_VALUE) {
```
jtuglu1, line 145: probably want to make this limit configurable

processing/src/main/java/org/apache/druid/query/topn/vector/VectorTopNEngine.java
```diff
@@ -0,0 +1,367 @@
...
+      return false;
+    }
+
+    for (final AggregatorFactory agg : query.getAggregatorSpecs()) {
+      if (!agg.canVectorize(effectiveInspector)) {
+        return false;
+      }
+    }
+
+    return true;
```
FrankChen021, line 200: [P1] `canVectorize` admits unsupported object/COMPLEX dimensions.

`VectorTopNEngine.canVectorize` only filters out decorated specs, arrays, and multi-value columns, then returns true for any remaining type whose output type matches the column capabilities. But `TopNVectorColumnProcessorFactory.makeObjectProcessor` only handles STRING object selectors and throws for every other object/COMPLEX type. That means a query using a `DefaultDimensionSpec` over a nested/COMPLEX column can be marked vectorizable here and then fail at runtime in `makeObjectProcessor` instead of falling back to the row path. `canVectorize` should reject non-STRING object/COMPLEX dimensions up front so capability checks match actual factory support.

processing/src/main/java/org/apache/druid/query/topn/TopNQueryEngine.java
```diff
@@ -96,13 +98,34 @@ public Sequence<Result<TopNResultValue>> query(
...
+      if (shouldVectorize) {
+        final ResourceHolder<ByteBuffer> bufHolder = bufferPool.take();
+        try {
+          final Closer resourceCloser = Closer.create();
+          resourceCloser.register(bufHolder);
+          resourceCloser.register(cursorHolder);
+          return Sequences.filter(
+              VectorTopNEngine.process(query, timeBoundaryInspector, cursorHolder, bufHolder.get()),
+              Predicates.notNull()
+          ).withBaggage(resourceCloser);
```
FrankChen021, line 117: [P2] Vectorized TopN bypasses existing query metrics reporting.

The new early return into `VectorTopNEngine.process` skips the row-path bookkeeping that reports TopN metrics today. In the non-vector path this method records `queryMetrics.cursor(...)`, then `getMapFn` records `dimensionCardinality(...)` and algorithm selection, and `TopNMapFn` records selector and pass-size metrics. None of that runs when `shouldVectorize` is true, so enabling vectorization changes emitted TopN metrics and removes operational visibility into algorithm choice and cardinality. If that loss is intended it should be wired back explicitly; otherwise this is a regression.

jtuglu1: Something we should discuss is what algorithm/cursor metrics we want to expose in the vectorized path?

jtuglu1:
@gianm any thoughts here?

gianm:
I'm a bit backlogged on reviews but hope to have a look soon.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/HeapVectorGrouper.java
gianm: Agreed, this should be configurable. It will help with testing, since we can configure a low limit to test limit-hitting behavior without having to allocate 2 GB in the tests. It will also be useful in production since 2 GB per processing thread can add up really quickly.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/HeapVectorGrouper.java
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * On-heap {@link VectorGrouper} that grows aggregator state on demand, up to maximum limit of 2GB.
+ *
+ * Vectorized analogue of {@link org.apache.druid.query.topn.BaseTopNAlgorithm}'s
+ * {@code runWithCardinalityUnknown} path: used when dimension cardinality is unknown (numeric columns,
+ * non-dict-encoded string virtual columns) or when a dict-encoded string column's cardinality exceeds
+ * the processing buffer. Memory footprint is on-heap and grows with the distinct-key count — matching
+ * the non-vectorized path's memory profile for the same queries.
+ */
+public class HeapVectorGrouper implements VectorGrouper
```
gianm, line 45: This seems kind of similar to `HashVectorGrouper`, used by `groupBy`. The difference I can see is that this one uses heap buffers rather than the processing buffer. It'd be nice to share code and it'd be nice to use the processing buffer here if possible.

Other than that, I'm also concerned that there are some aggregators which tend to take up much more space for the `BufferAggregator` / `VectorAggregator` form than for the `Aggregator` form for small numbers of aggregated data points (certain sketches, especially theta sketches but also sometimes hll). With `groupBy` we spill these and then the spill files tend to be very small. The problem is described in #19357 and #19439.

So, overflowing into the heap when the processing buffer is full, but continuing to use `VectorAggregator` in-heap, can cause large amounts of heap to be used. I wonder if this would work well:

- Start `HashVectorGrouper` (the one from `groupBy`) in the processing buffer
- If aggregation completes with `HashVectorGrouper`, emit that
- If aggregation stops due to the processing buffer being full, "spill" to heap and then continue aggregating in the processing buffer.
- If the total in-heap "spill" memory exceeds some threshold then fail the query.
- Once done, merge what's in the buffer along with any "spills" into the final result.

processing/src/main/java/org/apache/druid/query/topn/TopNQueryEngine.java
```diff
@@ -96,13 +98,34 @@ public Sequence<Result<TopNResultValue>> query(
       if (cursorHolder.isPreAggregated()) {
         query = query.withAggregatorSpecs(Preconditions.checkNotNull(cursorHolder.getAggregatorsForPreAggregated()));
       }
+
+      final TimeBoundaryInspector timeBoundaryInspector = segment.as(TimeBoundaryInspector.class);
+
+      final boolean canVectorize = cursorHolder.canVectorize()
+                                   && VectorTopNEngine.canVectorize(query, cursorFactory);
+      final boolean shouldVectorize = query.context().getVectorize().shouldVectorize(canVectorize);
```
gianm, line 106: I'd like to have a `vectorizeTopN` that is separate from `vectorize`, and *both* must be set in order to vectorize topN. The idea is that if someone is having issues with vectorized topNs, they can set `vectorizeTopN: false` across the board rather than have to micro-manage the `vectorize` setting per-query. Which isn't generally practical anyway.

processing/src/main/java/org/apache/druid/query/topn/vector/VectorTopNEngine.java
```diff
@@ -0,0 +1,373 @@
...
+   */
+  public static boolean canVectorize(final TopNQuery query, final ColumnInspector inspector)
+  {
+    final DimensionSpec dimensionSpec = query.getDimensionSpec();
+
+    if (!dimensionSpec.canVectorize()) {
+      return false;
+    }
+
+    // Decorated specs (e.g. extraction functions that are not one-to-one) change value semantics in ways that
```
gianm, line 160: This comment doesn't sound right: `ExtractionDimensionSpec` does not in fact decorate. (It returns `false` from `mustDecorate`.) There may be some other reason why decoration isn't supported?

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/HeapVectorGrouper.java
```diff
@@ -0,0 +1,218 @@
...
+        Groupers.writeAggregationRows(vAggregationRows, startRow, endRow)
+    );
+
+    return AggregateResult.ok();
+  }
+
+  private void growBuffer(final long neededCapacity)
+  {
+    if (neededCapacity > Integer.MAX_VALUE) {
+      throw new ISE("Aggregator state exceeds 2 GB; cardinality too high for HeapVectorGrouper");
```
gianm, line 146: This should be a `DruidException` with `Persona.USER` and `Category.RUNTIME_FAILURE`. It should include some advice about how to fix it, such as setting `vectorizeTopN: false` in query context.

gianm (commented):
I left some line comments about the general approach. Beyond that, for some additional test coverage, try parameterizing `TopNQueryRunnerFactory` on `vectorize: {force, false}`. You could model the test on `GroupByQueryRunnerTest`; the way we do it there is that tests that can't vectorize are marked `cannotVectorize()`.

PR #19359, opened 2026-04-21
fix: package logging emitter resource and rename to avoid classpath issues

sarangv:
Fixes #[19030](https://github.com/apache/druid/pull/19030) (follow-up).

### Description

Follow-up fix for PR #19030 (`add defaultMetrics for logging emitter`), which introduced metric filtering for the `LoggingEmitter`. Two issues were discovered during deployment testing in a real Druid cluster:

#### Fixed resource packaging in processing/pom.xml

The `processing/pom.xml` already had a `<resources>` block that only included the Sigar native library directory. When Maven sees any explicit `<resources>` section, it stops automatically including the default `src/main/resources` directory. This meant `defaultMetrics.json` was never packaged into the processing JAR, causing a `DruidException` (NOT_FOUND) at runtime when `LoggingEmitter` attempted to load it via `getResourceAsStream()`.

The fix adds `<resource><directory>src/main/resources</directory></resource>` explicitly alongside the existing Sigar entry.

#### Renamed defaultMetrics.json to loggingEmitterAllowedMetrics.json

Both the `processing` module and the `extensions-contrib/prometheus-emitter` module ship a classpath resource called `defaultMetrics.json`, but they have completely different schemas:

- **Prometheus emitter**: objects with `dimensions`, `type`, `conversionFactor`, `help` fields
- **Logging emitter**: flat object with metric name keys and empty `[]` values

Depending on classloader ordering at runtime, `LoggingEmitter.class.getClassLoader().getResourceAsStream("defaultMetrics.json")` could resolve to the Prometheus emitter's file instead, causing a silent misconfiguration or parse failure. Renaming to `loggingEmitterAllowedMetrics.json` gives the logging emitter a unique resource name with no collision risk.

#### Release note

Operators using the logging emitter metric filtering feature introduced in #19030 should note that the bundled default allowlist resource has been renamed from `defaultMetrics.json` to `loggingEmitterAllowedMetrics.json`. No configuration changes are needed -- the emitter automatically loads the correct resource. The `druid.emitter.logging.allowedMetricsPath` property continues to work as documented for custom allowlist files.

---

- `LoggingEmitter` -- updated resource constant and Javadoc to reference `loggingEmitterAllowedMetrics.json`
- `LoggingEmitterConfig` -- updated Javadoc to reference `loggingEmitterAllowedMetrics.json`
- `processing/pom.xml` -- added explicit `src/main/resources` to `<resources>` block
- `processing/src/main/resources/loggingEmitterAllowedMetrics.json` -- renamed from `defaultMetrics.json`
- `processing/src/test/resources/loggingEmitterAllowedMetrics.json` -- renamed from `defaultMetrics.json`

---

This PR has:

- [x] been self-reviewed.


abhishekrb19 (approved):
Thanks for the fix, @sarangv.
I’ve tagged the PR with the 37.0.0 milestone so we can include this in the upcoming release.
cc: @cecemei

PR #19369, opened 2026-04-23
refactor: clean up min/max task count limits in streaming task autoscaler

jtuglu1:
### Description

This makes auto-scaling min/max handling done in a single place (since it is common logic shared by all auto-scalers) in the supervisor dynamic task call action. This also cleans up the task auto scaler contract of returning `-1` *only* when the scaler failed to produce a desired task count (lack of metrics, etc.). Otherwise, it is expected to return a desired task count that the supervisor will then audit/clamp/skip.

#### Release note

Clean up min/max task count limits in streaming task autoscaler

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
@@ -545,8 +545,7 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
                                                                .setDimension(DruidMetrics.DATASOURCE, dataSource)
                                                                .setDimension(DruidMetrics.STREAM, getIoConfig().getStream());
 
-          // Auto-scaler contract: a negative return is pathological (scaler couldn't compute a
-          // useful answer, e.g. insufficient metrics).
+          // Negative return is the scaler's error signal (see SupervisorTaskAutoScaler).
           if (desiredTaskCount < 0) {
             log.warn(
                 "Auto-scaler returned pathological taskCount[%d] for supervisor[%s] for dataSource[%s]; skipping scale.",
@@ -559,12 +558,6 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
             return;
           }
 
-          // Clamp the scaler's preferred count to the supervisor's configured bounds. The scaler
-          // is responsible only for computing what it wants; bounds enforcement and the
-          // corresponding skip-reason emissions live here. Partition count is a hard ceiling — we
-          // cannot usefully run more tasks than partitions for stream ingestion — so it caps both
-          // configured min and max (protects against misconfigurations where an operator-set min
-          // exceeds partitionCount, which would otherwise force us to allocate idle tasks).
           final int partitionCount = getPartitionCount();
           final int rawMin = autoScalerConfig.getTaskCountMin();
           final int rawMax = autoScalerConfig.getTaskCountMax();
@@ -572,9 +565,6 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
           final int taskCountMax = partitionCount > 0 ? Math.min(rawMax, partitionCount) : rawMax;
           final int clampedTaskCount = Math.min(taskCountMax, Math.max(taskCountMin, desiredTaskCount));
 
-          // When the clamped value equals the current count there is nothing to do. Emit a specific
-          // skip reason reflecting *why* we stopped: at a configured bound if clamping was the cause,
-          // otherwise the scaler simply prefers the status quo.
           if (clampedTaskCount == currentTaskCount) {
             final String skipReason;
             if (desiredTaskCount > taskCountMax) {
@@ -594,7 +584,6 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
                 taskCountMin,
                 taskCountMax
             );
-            // Emit the *unclamped* desired count so operators see what the scaler actually wants.
             emitter.emit(event.setDimension(AUTOSCALER_SKIP_REASON_DIMENSION, skipReason)
                              .setMetric(AUTOSCALER_REQUIRED_TASKS_METRIC, desiredTaskCount));
             return;
@@ -626,8 +615,7 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
           // under-resourced while it recovers from lag induced by the scale event, so callers may
           // configure a longer cooldown for scale-down than for scale-up. Both directions are measured against the same
           // last-scale timestamp so that a rapid up/down oscillation is still subject to the appropriate cooldown,
-          // regardless of which direction triggered last. Direction is computed against the
-          // *clamped* value — that is the scale we would actually apply.
+          // regardless of which direction triggered last.
           final ScaleDirection scaleDirection;
           final long cooldownMillis;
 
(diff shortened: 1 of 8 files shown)


FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -166,61 +166,47 @@ public int computeTaskCountForRollover()
...
     final int optimalTaskCount = computeOptimalTaskCount(lastKnownMetrics);
-    int currentTaskCount = supervisor.getIoConfig().getTaskCount();
-
-    // Take the current task count but clamp it to the configured boundaries if it is outside the boundaries.
-    // There might be a configuration instance with a handwritten taskCount that is outside the boundaries.
-    final boolean isTaskCountOutOfBounds = currentTaskCount < config.getTaskCountMin()
-                                     || currentTaskCount > config.getTaskCountMax();
-    if (isTaskCountOutOfBounds) {
-      currentTaskCount = Math.min(config.getTaskCountMax(), Math.max(config.getTaskCountMin(), currentTaskCount));
+    if (optimalTaskCount <= 0) {
```
FrankChen021, line 172: [P2] Cost scaler can leave out-of-bound counts untouched

This now treats every non-positive optimal count as a pathological scaler failure, but `computeOptimalTaskCount` still returns `-1` for the normal case where the current count remains cheapest. Because the old out-of-bounds clamp was removed from this method, a cost-based supervisor whose current `taskCount` is outside `[taskCountMin, taskCountMax]` can return `-1` before the supervisor gets any boundary target to clamp, leaving the configured min/max violated and emitting a misleading failure metric. Restore an explicit out-of-bounds boundary return before cost optimization, or separate the normal no-op signal from invalid-metrics failures.

Fly-Style (approved):
LGTM, but address please the message from @FrankChen021 .
Thanks for the enhancement in streaming supervisor autoscaler area!

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

- indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java:111 - [P2] Preserve legacy task-count boundary configs. Existing persisted supervisor specs that set `useTaskCountBoundaries: true` now deserialize with `useTaskCountBoundariesOnScaleUp` defaulting to false and `useTaskCountBoundariesOnScaleDown` defaulting to true, because the legacy property is accepted but ignored. After an upgrade, the same cost-based autoscaler can therefore evaluate the full scale-up candidate set and jump directly toward `taskCountMax`, where previously the legacy boundary limited the step size. Please map the legacy value to the new scale-up/scale-down flags when those flags are absent, or otherwise reject/migrate the old property instead of silently changing behavior.

---

FrankChen021 (commented):

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/LagBasedAutoScaler.java
```diff
@@ -237,74 +212,31 @@ int computeDesiredTaskCount(List<Long> lags)
...
-                             "Already at min task count"
-                         )
-                         .setMetric(SeekableStreamSupervisor.AUTOSCALER_REQUIRED_TASKS_METRIC, taskCount));
-        return -1;
-      } else {
-        desiredActiveTaskCount = Math.max(taskCount, actualTaskCountMin);
-      }
-      return desiredActiveTaskCount;
+      // scale-in: step down from current (the supervisor will clamp to taskCountMin)
+      return currentActiveTaskCount - lagBasedAutoScalerConfig.getScaleInStep();
```
gianm: I think this needs `max(1, ...)` to avoid returning a zero or negative task count when `scaleInStep >= currentActiveTaskCount`. The supervisor will clamp but not if the returned number hits the "Auto-scaler returned pathological taskCount" check.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -503,29 +503,56 @@ public void handle()
...
-
           ServiceMetricEvent.Builder event = ServiceMetricEvent.builder()
                                                                .setDimension(DruidMetrics.SUPERVISOR_ID, supervisorId)
                                                                .setDimension(DruidMetrics.DATASOURCE, dataSource)
                                                                .setDimension(DruidMetrics.STREAM, getIoConfig().getStream());
 
-          // 1) This should already be handled by the auto-scaler implementation, but make sure we catch/record these for auditability
-          if (desiredTaskCount == currentTaskCount) {
+          // Negative return is the scaler's error signal (see SupervisorTaskAutoScaler).
+          if (desiredTaskCount < 0) {
```
gianm: Why not `<= 0`? Zero tasks seems pathological to me too.

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/autoscaler/SupervisorTaskAutoScaler.java
```diff
@@ -19,17 +19,31 @@
...
 
+/**
+ * Task-count auto-scaler driven by a streaming supervisor.
+ * <p>
+ * Scaler return-value contract for {@link #computeTaskCountForRollover()} and any
+ * implementation-specific scale-action method:
+ * <ul>
+ *   <li>{@code -1} — error case: metrics unavailable or an answer cannot be computed. The
+ *       supervisor will skip scaling and emit a failure metric.</li>
+ *   <li>Otherwise — the scaler's preferred task count, <i>unclamped</i> by configured min/max
```
gianm: IMO this should require that the task count be at least 1.

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/autoscaler/SupervisorTaskAutoScaler.java
```diff
@@ -19,17 +19,31 @@
 
 package org.apache.druid.indexing.overlord.supervisor.autoscaler;
 
+/**
+ * Task-count auto-scaler driven by a streaming supervisor.
+ * <p>
+ * Scaler return-value contract for {@link #computeTaskCountForRollover()} and any
+ * implementation-specific scale-action method:
+ * <ul>
+ *   <li>{@code -1} — error case: metrics unavailable or an answer cannot be computed. The
```
gianm: nit: would be nice to define this as a `static int CANNOT_COMPUTE = -1` on the interface, so implementations can `return CANNOT_COMPUTE;`.

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/autoscaler/SupervisorTaskAutoScaler.java
jtuglu1: What about for idling supervisors? We technically don't support auto-scale to IDLE state currently, so this contract can be changed later I suppose.

gianm: I figure that "go idle" is something that the main supervisor decides, not the autoscaler. I suppose it could be moved to the autoscaler, in which case allowing it to scale to zero tasks would make sense. OTOH it being the supervisor's responsibility is kind of nice, because it means each autoscaler doesn't have to implement idleness detection on its own.

gianm (approved):

PR #19370, opened 2026-04-23
feat: Update to Calcite 1.41.0.

gianm:
Main changes:

1) Refreshed Parser.jj using dev/upgrade-calcite-parser.

2) Added DruidTypeCoercion to ensure array comparison works properly. We need to be able to coerce BOOLEAN arrays to numeric arrays.

3) Update various toDruidQueryForExplaining implementations to use unique field names for their row signatures. This was a pre-existing bug that was triggered by new Calcite behavior.

4) Update DruidJoinRule to add null gates when hoisting literals over full or left outer joins. This was a pre-existing bug that was triggered by new Calcite behavior.

5) Update FilterDecomposeConcatRule to handle SEARCH.

6) Add InlineValuesSubQueryRule, which is needed to avoid unplannable plans for certain queries mixing UNNEST and SELECT-FROM-VALUES.

7) Add DruidStripUnionArmCastRule, which is needed to remove new CASTs that would otherwise prevent table-level unions from planning.

8) Remove workarounds and negative tests for [CALCITE-6435](https://issues.apache.org/jira/browse/CALCITE-6435), [CALCITE-6715](https://issues.apache.org/jira/browse/CALCITE-6715), and [CALCITE-6473](https://issues.apache.org/jira/browse/CALCITE-6473).

9) Add test to verify WEEK and QUARTER math work properly ([CALCITE-6581](https://issues.apache.org/jira/browse/CALCITE-6581)).


gianm:
This will fix #18665 as well.

sql/src/main/java/org/apache/druid/sql/calcite/rule/DruidStripUnionArmCastRule.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  @Override
+  public void onMatch(final RelOptRuleCall call)
+  {
+    final Union union = call.rel(0);
+    final List<RelNode> newInputs = new ArrayList<>(union.getInputs().size());
+    boolean anyChanged = false;
+    for (final RelNode input : union.getInputs()) {
+      final RelNode arm = input.stripped();
```
FrankChen021, line 85: P1 Stripping union-arm casts can change native output types

stripArmCasts removes numeric casts from UNION ALL arms whenever the resulting union row type still equals the original row type. That planner-level equality does not guarantee the native Druid query still emits the same column type: removing a BIGINT-to-FLOAT or DECIMAL-to-DOUBLE cast lets the scan arm expose its underlying native type, and the union datasource can then mix arm output types or choose a different runtime type than the SQL row type promised. This can change query results or fail during native union execution. Keep the casts in a project above each arm, or prove the native union path preserves the SQL cast semantics before stripping them.

gianm: This is true, although this relative looseness with table-level `UNION ALL` types happened prior to this PR too. This rule is about restoring the pre-Calcite-1.4.1 behavior, which is required for `UNION ALL` to work in the face of small schema differences. Generally it isn't a problem, since either an aggregator, expression, or result coercion would be applied to the data coming out of a `UNION ALL`, which would harmonize the types.

FrankChen021: Thanks, that makes sense. Given this is restoring the prior table-level UNION ALL behavior and the native union path already only checks column-name compatibility, I agree this should not block the PR. The original type-coercion risk is still worth keeping in mind for direct UNION ALL results without a downstream expression, aggregation, or result coercion, but I’m okay treating it as existing behavior rather than a new regression here.

abhishekrb19:
@gianm I see CI is green, but is this patch pending review and/or testing? Should we try to get the Calcite upgrade in for 39.0.0 (it's probably too late for 38.0.0)?

gianm:
> @gianm I see CI is green, but is this patch pending review and/or testing? Should we try to get the Calcite upgrade in for 39.0.0 (it's probably too late for 38.0.0)?

It's just pending review. And yeah I think targeting 39.0.0 would be good.

clintropolis (approved):

PR #19372, opened 2026-04-24
feat: Bounded Stream Supervisor

aho135:
### Description

Introduces a new property to the Stream Supervisor spec IOConfig called boundedStreamConfig which allows operators to specify start and end offset ranges for short-lived supervised ingestion. This property modifies the main Supervisor run loop to only ingest from and monitor partitions specified in the boundedStreamConfig. After the offset range has been consumed the Supervisor will transition into a terminal state (COMPLETED). The motivation for this PR came out of https://github.com/apache/druid/pull/19191 which submits backfill tasks that are unsupervised. Once this change is merged, 19191 can be enhanced to use the boundedStreamConfig so that the backfill tasks are supervised. 

#### Release note
Adds a property called boundedStreamConfig to the SeekableStreamSupervisorIOConfig which allows operators to spin up a Supervisor that consumes only a specified offset range.

<hr>

 * `SeekableStreamSupervisorIOConfig`
 * `BoundedStreamConfig`
 * `SeekableStreamSupervisor`

<hr>

This PR has:

- [X] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/BoundedStreamConfig.java
```diff
@@ -0,0 +1,90 @@
...
+  @JsonCreator
+  public BoundedStreamConfig(
+      @JsonProperty("startSequenceNumbers") Map<?, ?> startSequenceNumbers,
+      @JsonProperty("endSequenceNumbers") Map<?, ?> endSequenceNumbers
+  )
+  {
+    this.startSequenceNumbers = Preconditions.checkNotNull(startSequenceNumbers, "startSequenceNumbers");
+    this.endSequenceNumbers = Preconditions.checkNotNull(endSequenceNumbers, "endSequenceNumbers");
+
+    // Validation
```
abhishekrb19: As a guard rail, I think it'll be good to have stricter checks for each partition so there's no unintended behavior with incorrect ranges specified:
`startSequenceNumbers < endSequenceNumbers`

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/BoundedStreamConfig.java
```diff
@@ -0,0 +1,90 @@
...
+package org.apache.druid.indexing.seekablestream.supervisor;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Preconditions;
+
+import java.util.Map;
+
+/**
+ * Configuration for bounded (one-time) stream processing with explicit start/end offsets.
```
abhishekrb19, line 29: could you document that the start & end offsets are inclusive?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/BoundedStreamConfig.java
```diff
@@ -0,0 +1,90 @@
...
+  private final Map<?, ?> endSequenceNumbers;    // Partition -> End Offset
+
+  @JsonCreator
+  public BoundedStreamConfig(
+      @JsonProperty("startSequenceNumbers") Map<?, ?> startSequenceNumbers,
+      @JsonProperty("endSequenceNumbers") Map<?, ?> endSequenceNumbers
+  )
+  {
+    this.startSequenceNumbers = Preconditions.checkNotNull(startSequenceNumbers, "startSequenceNumbers");
+    this.endSequenceNumbers = Preconditions.checkNotNull(endSequenceNumbers, "endSequenceNumbers");
```
abhishekrb19: nit: we could throw `DruidException`s here with the admin persona for these validation errors here

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorStateManager.java
```diff
@@ -64,7 +64,8 @@ public enum BasicState implements State
     RUNNING(true, false),
     IDLE(true, false),
     SUSPENDED(true, false),
-    STOPPING(true, false);
+    STOPPING(true, false),
+    COMPLETED(true, false);
```
abhishekrb19, line 68: Just wondering, if it's possible to have a completed but unhealthy state? Or would that be `UNHEALTHY_SUPERVISOR`?

abhishekrb19 (commented):
Thanks @aho135 - took a quick glance and the approach looks good to me overall!
I’m still going through some of the main files and just checkpointing my review. Do you think it would be possible to add a simple embedded test with the new config for some end-to-end coverage?

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorIOConfig.java
```diff
@@ -78,7 +79,8 @@ public KafkaSupervisorIOConfig(
       @JsonProperty("idleConfig") IdleConfig idleConfig,
       @JsonProperty("stopTaskCount") Integer stopTaskCount,
       @Nullable @JsonProperty("emitTimeLagMetrics") Boolean emitTimeLagMetrics,
-      @Nullable @JsonProperty("serverPriorityToReplicas") Map<Integer, Integer> serverPriorityToReplicas
+      @Nullable @JsonProperty("serverPriorityToReplicas") Map<Integer, Integer> serverPriorityToReplicas,
+      @Nullable @JsonProperty("boundedStreamConfig") BoundedStreamConfig boundedStreamConfig
```
abhishekrb19, line 83: How do the bounds interplay with the supervisor’s behavior with other properties like `taskDuration`, `completionTimeout`, etc.?

I suppose if the bounds are too wide, tasks would still roll over and all existing properties would be honored as-is. The only additional behavioral change is when this optional config is specified and the end sequence numbers are reached, the supervisor eventually moves to a terminal `COMPLETED` state?

abhishekrb19: It would also be good to update some user-facing documentation in `supervisor.md` for this functionality

extensions-core/kinesis-indexing-service/src/main/java/org/apache/druid/indexing/kinesis/supervisor/KinesisSupervisor.java
```diff
@@ -381,6 +381,28 @@ protected boolean isShardExpirationMarker(String seqNum)
...
+    return current.compareTo(target) >= 0;
+  }
+
+  @Override
+  protected String createPartitionIdFromString(String partitionIdString)
+  {
+    // Kinesis uses String as partition ID, so just return the string as-is
+    return partitionIdString;
+  }
+
```
FrankChen021, line 400: P1 Kinesis bounded completion compares offsets lexicographically

isOffsetAtOrBeyond uses current.compareTo(target), which compares Kinesis sequence numbers lexicographically. Kinesis sequence numbers are decimal strings whose numeric order is not generally the same as lexicographic order, especially across different digit lengths. A bounded supervisor can mark a shard complete before the numeric end offset is reached, or keep recreating tasks after it passed the end. Compare Kinesis offsets using their numeric value or the existing Kinesis sequence-number ordering semantics.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4255,6 +4418,23 @@ private OrderedSequenceNumber<SequenceOffsetType> getOffsetFromStorageForPartiti
       }
```
FrankChen021, line 4517: P1 Bounded starts can be ignored when metadata exists

getOffsetFromStorageForPartition only falls back to boundedStreamConfig.startSequenceNumbers when no metadata/checkpoint offset exists. If a supervisor is reset or reconfigured with a requested bounded start while metadata storage still has an older offset, the stored metadata wins and the task starts from the stale position instead of the user-supplied bounded start. That can skip the requested backfill range or process a different interval than configured. Bounded mode should either clear/namespace old metadata for the run or explicitly prefer the configured start when initializing the bounded task group.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/BoundedStreamConfig.java
aho135: Start offsets are inclusive, but end offsets depend on the implementation for isEndOffsetExclusive. For kinesis it is inclusive, but for Kafka and Rabbit it is exclusive

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorStateManager.java
aho135: Yeah if tasks are failing the Supervisor would enter an `UNHEALTHY` state. It would only transition to `COMPLETED` after the specified offset range was successfully processed

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorIOConfig.java
aho135: Yup, the idea is for `taskDuration` and `completionTimeout` to work as they currently do. If a task exceeds the configured `taskDuration` but the offset range hasn't been fully processed yet then `isOffsetAtOrBeyond` will return false and the Supervisor will spin up a new task starting from the checkpointed metadata

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
aho135: Thanks for the review @FrankChen021!

> Bounded mode should either clear/namespace old metadata for the run

I'm a bit wary of automatic cleanup of metadata. I'm thinking through the scenario where a cluster operator has a running Supervisor. They want to re-ingest some older data so they resubmit the exact same spec (forgetting to update `id`) so the Bounded Supervisor succeeds but the previously committed offset gets lost.

> explicitly prefer the configured start when initializing the bounded task group.

This falls into the same issue as above where if the operator forgets to set the `id` to something different than the running Supervisor then the previous committed offset is lost forever

I'm leaning towards adding validation that if metadata already exists for the `id` then just throw an exception and suggest the operator to resubmit with a different `id` or reset the Supervisor. Curious to hear your thoughts on this approach. Thanks again

FrankChen021: That validation approach sounds right to me. The important part is preventing a bounded supervisor from accidentally mixing explicit bounded start offsets with existing committed metadata for the same id; failing fast with guidance to use a different id or reset the supervisor would avoid the silent stale-offset behavior without deleting or overwriting a running supervisor's offsets.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4347,6 +4478,179 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+
+  /**
+   * Handle bounded processing completion by shutting down the supervisor.
+   * At this point, all task groups are already empty (verified by isBoundedWorkComplete),
+   * so we just need to mark the supervisor as completed.
+   */
+  private void handleBoundedCompletion()
+  {
+    log.info("Bounded processing complete for supervisor[%s]. Marking as COMPLETED.", supervisorId);
+    stateManager.maybeSetState(SupervisorStateManager.BasicState.COMPLETED);
```
abhishekrb19, line 4748: Should this call `stop()` with this COMPLETED state so things get unregistered and the executor is removed?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -259,6 +262,42 @@ public void validateSpecUpdateTo(SupervisorSpec proposedSpec) throws DruidExcept
     if (!this.getSource().equals(other.getSource())) {
       throw InvalidInput.exception(ILLEGAL_INPUT_SOURCE_UPDATE_ERROR_MESSAGE, this.getSource(), other.getSource());
     }
+
+    // Validate bounded stream configuration
+    validateBoundedStreamConfig(other);
```
abhishekrb19: Looks like this is only invoked on an update. Should we validate this during creation time as well?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -259,6 +262,42 @@ public void validateSpecUpdateTo(SupervisorSpec proposedSpec) throws DruidExcept
...
+    if (ioConfig.isBounded()) {
+      // Validate partition consistency
+      BoundedStreamConfig boundedConfig = ioConfig.getBoundedStreamConfig();
+      if (!boundedConfig.getStartSequenceNumbers().keySet().equals(boundedConfig.getEndSequenceNumbers().keySet())) {
+        throw InvalidInput.exception(
+            "Bounded stream config has mismatched partitions. Start: %s, End: %s",
+            boundedConfig.getStartSequenceNumbers().keySet(),
+            boundedConfig.getEndSequenceNumbers().keySet()
+        );
+      }
```
abhishekrb19: I wonder if some of these checks during create & update flows can be shared. Perhaps add a static helper inside `BoundedStreamConfig` that can be validated by both create & `validateSpecUpdateTo`?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -259,6 +262,42 @@ public void validateSpecUpdateTo(SupervisorSpec proposedSpec) throws DruidExcept
...
+
+      // Warn if useConcurrentLocks is not enabled
+      Map<String, Object> context = spec.getContext();
+      if (context == null || !Boolean.TRUE.equals(context.get("useConcurrentLocks"))) {
+        log.warn(
+            "Bounded stream processing without 'useConcurrentLocks=true' may fail " +
+            "if other supervisors are running or segments already exist for these intervals. " +
+            "Consider setting useConcurrentLocks=true in the supervisor context."
+        );
```
abhishekrb19: Is this requirement needed in the general case when an operator is submitting a supervisor with these bounds? I’d expect `useConcurrentLocks` to be required only when a backfill supervisor is submitted in parallel with resetting the currently running supervisor. If that’s the case, would it make sense to enforce this at the reset and backfill API layer instead?

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/BoundedStreamConfigTest.java
```diff
@@ -0,0 +1,183 @@
...
+  @Test
+  public void testConstructorWithNullEndSequenceNumbers()
+  {
+    Map<String, Long> startOffsets = new HashMap<>();
+    startOffsets.put("0", 100L);
+
+    NullPointerException ex = Assert.assertThrows(
+        NullPointerException.class,
+        () -> new BoundedStreamConfig(startOffsets, null)
+    );
```
abhishekrb19: Seems strange to assert on NPEs. Should we have defensive code that throws DruidExceptions instead of NPEs wtih preconditions?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4347,6 +4478,179 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+   *
+   * So completion is: no active tasks, no pending tasks, and createNewTasks() chose not to create any.
+   * This is indicated by empty task groups after createNewTasks() has run.
+   *
+   * We do NOT separately check metadata storage here because:
+   * 1. Metadata may contain stale offsets from previous supervisor runs
+   * 2. createNewTasks() already does the offset checking logic
+   * 3. If tasks were killed/failed and work is incomplete, createNewTasks() will recreate them
+   *
+   * @return true if all bounded work is complete, false otherwise
```
abhishekrb19: Hmm, indirectly inferring bounded work completion from the absence of tasks seems error-prone since these things can race and function independently in some ways.
For example, if all tasks failed or if tasks were queued but haven’t started yet as mentioned in point 3, would the supervisor transition to a completed state prematurely?

I think a more foolproof way to check completion would be to directly check metadata storage to track offsets for work completion and have the design reflect that with appropriate guardrails. I believe this also relates to the discussion here: https://github.com/apache/druid/pull/19372/changes#r3141594283

aho135: Ahh, let me update the javadoc to be more clear. The implementation does take that scenario into account. Checking active/pending task group is just a cheap check to avoid hitting the metadata store every run. But if both are empty it will check the offsets in the metadata store as the final validation that it is complete

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
aho135: Yeah, this is just logged as a warning and is not a strict requirement. In the reset and backfill API that would be a strict enforcement

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
aho135: One workflow I was testing out was to submit a bounded Supervisor and have it run to completion. Then I adjusted the start/end offsets and re-submitted the spec. Then I did a hard reset to clear the metadata so it could ingest the new offset range. For this kind of workflow we would need the executor to continue running even after the initial completion

aho135: The latest [commit](https://github.com/apache/druid/pull/19372/changes/4bed6580f659959873a01b1ea926bb8c8c96fa26#diff-ee5ec6696f9263bed19eefcb5f24e25b4141d15aaa4f13853e8790227e52258b) handles this workflow

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
aho135: @FrankChen021 Thinking through this one a bit more, the validation approach is a bit tricky. It's not straightforward to tell if the existing metadata is from the bounded Supervisor itself (in which case starting from the metadata would be the correct behavior) or if it's from a previous Supervisor. 

One approach we can take is that if the metadata offset falls within the configured start/end offsets then use that, otherwise fall back to startOffset. This does run the risk of partial ingestion of the specified range though.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/BoundedStreamConfig.java
aho135: This validation is a bit difficult to do within BoundedStreamConfig because of the generic typing. But I do see there is already validation for this in Task IOConfig. [kafka](https://github.com/apache/druid/blob/master/extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaIndexTaskIOConfig.java#L91-L97)

aho135: The validation there isn't strict enough though because startOffset can equal endOffset. In that scenario the Supervisor spins up a task that consumes nothing and then shuts down. But since no data was consumed there is no metadata update so it gets stuck in a loop where it keeps spinning up tasks. I added additional validation to handle this scenario in this [commit](https://github.com/apache/druid/pull/19372/changes/c9181f025713915977dc7bdc237069d160189791)

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
FrankChen021: I agree the source of the metadata is the hard part here, but I would avoid using "metadata is within the configured start/end range" as the deciding rule. If the stored offset is inside `[start, end)`, starting there can still silently skip the prefix `[start, storedOffset)`, which is the same class of surprise as the original issue. I think the safer behavior is still to fail fast when bounded mode finds existing metadata for the configured bounded partitions unless there is an explicit signal that this is a resume of the same bounded run.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
aho135: I went ahead and removed this, along with the logging since these were largely the same validation checks that already run within BoundedStreamConfig

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
aho135: Thanks @FrankChen021! I took a stab at this in the most recent [commit](https://github.com/apache/druid/pull/19372/changes/5965ac4de5f1e344906ddbe1a94afce29b85b04e)
Please let me know your thoughts when you get the chance

FrankChen021: Thanks, this looks like the right direction to me. Persisting the bounded config in the datasource metadata and rejecting existing metadata whose bounded config is missing or different addresses the silent stale-offset case, while still allowing a supervisor to resume metadata from the same bounded run.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4298,6 +4497,163 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+      return false;
+    }
+
+    // Check if start >= end for all partitions (empty range)
+    // If so, there's no work to do - treat as already complete
+    boolean allPartitionsEmptyRange = true;
+    for (PartitionIdType partition : partitionsInGroup) {
+      SequenceOffsetType start = startOffsets.get(partition);
+      SequenceOffsetType end = endOffsets.get(partition);
+      if (!isOffsetAtOrBeyond(start, end)) {
```
FrankChen021: [P1] Kinesis bounded ranges with start == end are skipped

The new empty-range check treats start >= end as completed for all bounded supervisors before creating any task. That is valid for Kafka's exclusive end offsets, but Kinesis declares bounded end offsets as inclusive and its task runner returns isEndOffsetExclusive() == false, so a Kinesis bounded ingestion for a single record where startSequenceNumbers equals endSequenceNumbers is marked COMPLETED without reading that record. This should be provider-aware, for example only treating equality as empty when the end offset is exclusive, while still rejecting/handling start > end appropriately.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

FrankChen021 (commented):

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4347,6 +4546,175 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+        );
+        return false; // This partition hasn't reached its end
+      }
+    }
+
+    log.info(
+        "All partitions in taskGroup[%d] have reached their end offsets",
+        groupId
+    );
+    return true; // All partitions have reached their end offsets
```
FrankChen021, line 4681: [P1] Validate bounded metadata before completing from old offsets

A new bounded supervisor can silently complete using offsets from a previous bounded run with a different boundedStreamConfig. The mismatch check only runs in getOffsetFromStorageForPartition(), but isBoundedWorkComplete() calls hasTaskGroupReachedBoundedEnd() first, which reads raw metadata offsets and returns true when currentOffset >= the new end offset. For example, after a run [0,100] commits offset 100, posting [50,75] for the same datasource reaches COMPLETED without running tasks or throwing the documented metadata mismatch error. Validate the stored bounded config before treating metadata offsets as completion, or avoid using prior offsets when the bounded config differs.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
aho135: Thanks @FrankChen021 I did leave that as a comment in the [embedded test](https://github.com/apache/druid/pull/19372/changes#diff-0a11abc4cce2bf91968fdffd9cb4cbde67428de4b4f7541101a7b2ef7f34b35cR185-R189)

I suppose it would be better to be explicit about this mismatch here. Will make that update

FrankChen021: Thanks, the current update addresses my concern: `hasTaskGroupReachedBoundedEnd()` now refuses to treat stored offsets as completion evidence when the metadata bounded config differs, so task creation can reach the existing mismatch error path. The added stale-offset regression test covers the case I was worried about.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/BoundedStreamConfigTest.java
```diff
@@ -0,0 +1,329 @@
...
+
+    Map<String, Long> start2 = new HashMap<>();
+    start2.put("0", 0L);
+    Map<String, Long> end2 = new HashMap<>();
+    end2.put("0", 100L);
+
+    BoundedStreamConfig config1 = new BoundedStreamConfig(start1, end1);
+    BoundedStreamConfig config2 = new BoundedStreamConfig(start2, end2);
+
+    Assert.assertEquals(config1.hashCode(), config2.hashCode());
```
abhishekrb19: Could replace it with:
```
EqualsVerifier.forClass(BoundedStreamConfig.class).withNonnullFields("startSequenceNumbers", "endSequenceNumbers").usingGetClass().verify();
```

(Perhaps `EqualsVerifier` tests would replace some of the test permutations for equals/hashcode here)

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4347,6 +4546,182 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+        ));
+  }
+
+  /**
+   * For bounded supervisors, we determine completion by checking if new tasks would be created.
+   * In createNewTasks(), bounded mode checks hasTaskGroupReachedBoundedEnd() before creating tasks.
+   * If that returns true (offsets reached), no new tasks are created.
+   * So completion is: no active tasks, no pending tasks, and createNewTasks() chose not to create any.
+   *
+   * @return true if all bounded work is complete, false otherwise
```
abhishekrb19, line 4712: Thanks for clarifying the docs
nit: for easier navigation between these javadoc functions:
```suggestion
  /**
   * For bounded supervisors, we determine completion by checking if new tasks would be created.
   * In {@link #createNewTasks()}, bounded mode checks {@link #hasTaskGroupReachedBoundedEnd(int)} before creating tasks.
   * If that returns true (offsets reached), no new tasks are created.
   * So completion is: no active tasks, no pending tasks, and {@link #createNewTasks()} chose not to create any.
   *
   * @return true if all bounded work is complete, false otherwise
```

extensions-core/kinesis-indexing-service/src/main/java/org/apache/druid/indexing/kinesis/supervisor/KinesisSupervisor.java
```diff
@@ -381,12 +387,54 @@ protected boolean isShardExpirationMarker(String seqNum)
...
+    catch (NumberFormatException e) {
+      throw new IAE(
+          StringUtils.format(
+              "Invalid Kinesis sequence number. Expected numeric string but got current=[%s], target=[%s]",
+              current,
+              target
+          ),
+          e
+      );
+    }
```
abhishekrb19: Kinesis has some existing utilities for this which also accounts for some special handling of sequence numbers. Could probably be simplified to:
```
return KinesisSequenceNumber.of(current).compareTo(KinesisSequenceNumber.of(target)) >= 0;
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
```diff
@@ -0,0 +1,342 @@
...
+  private final KafkaResource kafkaServer = new KafkaResource();
+
+  @Override
+  protected StreamIngestResource<?> getStreamIngestResource()
+  {
+    return kafkaServer;
+  }
+
+  @Test
+  public void test_boundedSupervisor_ingestsDataAndCompletes()
```
abhishekrb19, line 62: Nice, thanks for these tests!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
```diff
@@ -0,0 +1,342 @@
...
+    Assertions.assertFalse(status2.isHealthy(), "Supervisor should be unhealthy after detecting metadata mismatch");
+    Assertions.assertEquals("UNHEALTHY_SUPERVISOR", status2.getState(), "Supervisor state should be UNHEALTHY_SUPERVISOR");
+  }
+
+  /**
+   * Regression test: a new bounded run whose endOffset is less than the offset committed by a prior
+   * run must not silently reach COMPLETED. Before the fix, hasTaskGroupReachedBoundedEnd() compared
+   * the stale committed offset against the new endOffset (e.g. committed=100 >= newEnd=50) and
+   * returned true, bypassing task creation and the documented mismatch error entirely.
+   */
```
abhishekrb19, line 246: Regression test or the javadoc can be entirely removed as these are new tests from this PR/self reviews itself:
```suggestion
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
```diff
@@ -0,0 +1,342 @@
...
+    publish1kRecords(topic, false);
+
+    // Run 1: ingest up to offset 100 on each partition and complete.
+    Map<String, Long> startOffsets1 = new HashMap<>();
+    startOffsets1.put("0", 0L);
+    startOffsets1.put("1", 0L);
+
+    Map<String, Long> endOffsets1 = new HashMap<>();
+    endOffsets1.put("0", 100L);
+    endOffsets1.put("1", 100L);
```
abhishekrb19: Just for completeness, maybe have unequal start / end offset marker for partitions 0 & 1 (in this test or the previous ones)?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
```diff
@@ -0,0 +1,342 @@
...
+    cluster.callApi().postSupervisor(supervisor2);
+
+    // Wait for the supervisor to process and detect the metadata mismatch
+    // The exception we're testing for is thrown and logged, and causes the supervisor to become unhealthy
+    waitForSupervisorToBeUnhealthy(supervisor2.getId());
+
+    // Verify the supervisor is unhealthy
+    final SupervisorStatus status2 = cluster.callApi().getSupervisorStatus(supervisor2.getId());
+    Assertions.assertFalse(status2.isHealthy(), "Supervisor should be unhealthy after detecting metadata mismatch");
+    Assertions.assertEquals("UNHEALTHY_SUPERVISOR", status2.getState(), "Supervisor state should be UNHEALTHY_SUPERVISOR");
```
abhishekrb19, line 240: Perhaps also validate that the records ingested by this bounded supervisor is 0 and/or the tasks spun up by this supervisor actually failed?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4347,6 +4546,182 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+    BoundedStreamConfig boundedConfig = ioConfig.getBoundedStreamConfig();
+    Map<PartitionIdType, SequenceOffsetType> startOffsets =
+        convertBoundedConfigMap(boundedConfig.getStartSequenceNumbers());
+    Map<PartitionIdType, SequenceOffsetType> endOffsets =
+        convertBoundedConfigMap(boundedConfig.getEndSequenceNumbers());
+
+    Set<PartitionIdType> partitionsInGroup = partitionGroups.get(groupId);
+    if (partitionsInGroup == null || partitionsInGroup.isEmpty()) {
+      return false;
+    }
```
abhishekrb19: Have an early return before the convert* methods are called?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4347,6 +4546,182 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+   * @param groupId The task group ID to check
+   * @return true if all partitions in the group have reached their end offsets, false otherwise
+   */
+  private boolean hasTaskGroupReachedBoundedEnd(int groupId)
+  {
+    BoundedStreamConfig boundedConfig = ioConfig.getBoundedStreamConfig();
+    Map<PartitionIdType, SequenceOffsetType> startOffsets =
+        convertBoundedConfigMap(boundedConfig.getStartSequenceNumbers());
+    Map<PartitionIdType, SequenceOffsetType> endOffsets =
+        convertBoundedConfigMap(boundedConfig.getEndSequenceNumbers());
```
abhishekrb19, line 4597: Actually, do you think it's worth resolving this once instead of computing it every time its invoked from the internal loop (given that it doesn't change)?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4347,6 +4546,182 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+      return false;
+    }
+
+    // Check if all partitions have empty ranges
+    // For exclusive end offsets (Kafka): start >= end means empty
+    // For inclusive end offsets (Kinesis): only start > end means empty (start == end is one record)
+    boolean allPartitionsEmptyRange = true;
+    for (PartitionIdType partition : partitionsInGroup) {
+      SequenceOffsetType start = startOffsets.get(partition);
+      SequenceOffsetType end = endOffsets.get(partition);
```
abhishekrb19, line 4607: nit: could we make variables final here & in surrounding code where applicable so the intent is clear and nothing accidentally mutates these offsets (except `allPartitionsEmptyRange`)?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4347,6 +4546,182 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+          isEndOffsetExclusive() ? ">=" : ">",
+          startOffsets,
+          endOffsets
+      );
+      return true;
+    }
+
+    // Offsets from a prior bounded run with a different config must not count as completion evidence.
+    // Returning false lets createNewTasks() run, which calls getOffsetFromStorageForPartition()
+    // and throws the documented mismatch error.
```
abhishekrb19: Not sure what this means without enough context: `// and throws the documented mismatch error.`

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4263,18 +4412,51 @@ private Map<PartitionIdType, OrderedSequenceNumber<SequenceOffsetType>> generate
...
+        // If configs don't match (or metadata has no config), throw exception
+        if (!currentBoundedConfig.equals(metadataBoundedConfig)) {
+          throw DruidException.forPersona(DruidException.Persona.ADMIN)
+                              .ofCategory(DruidException.Category.INVALID_INPUT)
+                              .build(
+                                  "Bounded supervisor detected existing metadata from a different run. "
+                                  + "Metadata bounded config [%s] does not match current config [%s]. "
+                                  + "To start a new bounded ingestion, either: "
+                                  + "(1) use the supervisor reset API to clear existing metadata, or "
+                                  + "(2) use a different supervisor ID / datasource.",
```
abhishekrb19: I think just a different supervisor ID would suffice for this backfill requirement? A different datasource would already be supported

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
```diff
@@ -0,0 +1,342 @@
...
+    endOffsets2.put("1", 50L);
+
+    BoundedStreamConfig boundedConfig2 = new BoundedStreamConfig(startOffsets2, endOffsets2);
+    final KafkaSupervisorSpec supervisor2 = createBoundedKafkaSupervisor(kafkaServer, topic, boundedConfig2);
+
+    cluster.callApi().postSupervisor(supervisor2);
+    waitForSupervisorToBeUnhealthy(supervisor2.getId());
+
+    final SupervisorStatus status2 = cluster.callApi().getSupervisorStatus(supervisor2.getId());
+    Assertions.assertFalse(status2.isHealthy(), "Supervisor should be unhealthy after detecting metadata mismatch");
```
abhishekrb19, line 291: Do you think validating a third supervisor with a different ID but the same bounds would be worth adding that succeeds? (either in this test or the following one)

docs/ingestion/supervisor.md
```diff
@@ -65,6 +65,7 @@ For configuration properties specific to Kafka and Kinesis, see [Kafka I/O confi
...
 |`stopTaskCount`|Integer|Limits the number of ingestion tasks Druid can cycle at any given time. If not set, Druid can cycle all tasks at the same time. If set to a value less than `taskCount`, your cluster needs fewer available slots to run the supervisor. You can save costs by scaling down your ingestion tier, but this can lead to slower cycle times and lag. See [`stopTaskCount`](#stoptaskcount) for more information.|No|`taskCount` value|
+|`boundedStreamConfig`|Object|Configures the supervisor for bounded (one-time) ingestion with explicit start and end offsets. When set, the supervisor creates tasks that read from `startSequenceNumbers` to `endSequenceNumbers`, then automatically terminates when all data is ingested. The bounded configuration is stored with datasource metadata; if a supervisor is restarted or a new supervisor is created with different offsets for the same datasource, it will fail. To retry with different offsets, use the supervisor reset API to clear metadata or use a different datasource. Useful for backfills and historical reprocessing. See [Bounded stream configuration](#bounded-stream-configuration) for details.|No|null|
```
abhishekrb19:
```suggestion
|`boundedStreamConfig`|Object|Configures the supervisor for bounded (one-time) ingestion with explicit start and end offsets. When set, the supervisor creates tasks that read from `startSequenceNumbers` to `endSequenceNumbers`, then automatically terminates when all data is ingested. The bounded configuration is stored with datasource metadata; if a supervisor is restarted or a new supervisor is created with different offsets for the same datasource, it will fail. To retry with different offsets, use the supervisor reset API to clear metadata or use a different supervisor ID. Useful for backfills and historical reprocessing. See [Bounded stream configuration](#bounded-stream-configuration) for details.|No|null|
```

docs/ingestion/supervisor.md
```diff
@@ -251,6 +252,57 @@ Before you set `stopTaskCount`, note the following:
...
+
+When configured, the supervisor:
+1. Creates tasks that start reading from `startSequenceNumbers`
+2. Tasks automatically stop when they reach `endSequenceNumbers`
+3. Supervisor does not create replacement tasks after tasks complete
+4. Supervisor transitions to `COMPLETED` state and terminates when all tasks finish
+
+**Metadata consistency:** The bounded configuration is stored in datasource metadata along with checkpointed offsets. If you restart the supervisor or create a new supervisor with a different `boundedStreamConfig` for the same datasource, the supervisor will fail with an error. To start a new bounded ingestion with different offsets, either:
+- Use the [supervisor reset API](../api-reference/supervisor-api.md#reset-a-supervisor) to clear existing metadata
+- Use a different datasource name
```
abhishekrb19: Ditto: I think using a different datasource name would already be supported by the multi-supervisor feature (I think we meant different supervisor ID here?)

abhishekrb19 (commented):
Thanks @aho135 for addressing reviews. Some additional comments, but I think it's pretty close

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
aho135: In the case of metadata mismatch the tasks are never actually spun up. An exception is thrown before tasks are ever created.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
```diff
@@ -0,0 +1,340 @@
...
+      }
+      catch (Exception e) {
+        // Supervisor might not be found immediately, retry
+        attempt++;
+        try {
+          Thread.sleep(1000);
+        }
+        catch (InterruptedException ie) {
+          Thread.currentThread().interrupt();
+          throw new RuntimeException("Interrupted while waiting", ie);
```
abhishekrb19: I think we should be able to latch on this metric instead of retry-with-sleeps:

```java
    overlord.latchableEmitter().waitForEvent(
        event -> event.hasMetricName("supervisor/count")
                      .hasDimension(DruidMetrics.SUPERVISOR_ID, supervisorId)
                      .hasDimension("state", "COMPLETED")
    );
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
```diff
@@ -0,0 +1,340 @@
...
+        Thread.sleep(1000);
+        attempt++;
+      }
+      catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new RuntimeException("Interrupted while waiting for supervisor to become unhealthy", e);
+      }
+      catch (Exception e) {
+        // Supervisor might not be found immediately, retry
+        attempt++;
```
abhishekrb19: Same comment - what do you think about enabling the `SupervisorStatsMonitor` to the base class's monitors? I think it’d be cleaner and more resilient to latch on the `supervisor/count` metric emitted by it:

```java
        .addCommonProperty("druid.monitoring.emissionPeriod", "PT1s")
        .addCommonProperty(
            "druid.monitoring.monitors",
            "[\"org.apache.druid.server.metrics.SupervisorStatsMonitor\"]"
        )
```

abhishekrb19 (approved):
👍

extensions-core/kinesis-indexing-service/src/test/java/org/apache/druid/indexing/kinesis/supervisor/KinesisSupervisorTest.java
```diff
@@ -4725,6 +4730,201 @@ public void test_doesTaskMatchSupervisor()
...
+  @Test
+  public void testIsOffsetAtOrBeyond_invalidSequenceNumber()
+  {
+    supervisor = getTestableSupervisor(1, 1, true, "PT1H", null, null);
+
+    Exception e = Assert.assertThrows(
+        IllegalArgumentException.class,
+        () -> supervisor.isOffsetAtOrBeyond("not-a-number", "12345")
+    );
+    Assert.assertTrue(e.getMessage().contains("Invalid Kinesis sequence number"));
```
FrankChen021: [P1] Fix invalid sequence assertion

This new assertion expects the exception message to contain `Invalid Kinesis sequence number`, but `KinesisSupervisor.isOffsetAtOrBeyond` delegates to `KinesisSequenceNumber.of`, which constructs a `BigInteger` and throws `NumberFormatException` with a different message for `not-a-number`. As written, this test should fail; either wrap invalid sequence numbers with the expected message in the production method or relax the assertion.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4377,6 +4577,179 @@ && checkSourceMetadataMatch(dataSourceMetadata)) {
...
+    // For inclusive end offsets (Kinesis): only start > end means empty (start == end is one record)
+    boolean allPartitionsEmptyRange = true;
+    for (PartitionIdType partition : partitionsInGroup) {
+      SequenceOffsetType start = startOffsets.get(partition);
+      SequenceOffsetType end = endOffsets.get(partition);
+
+      boolean isEmpty;
+      if (isEndOffsetExclusive()) {
+        // Exclusive: empty if start >= end
+        isEmpty = isOffsetAtOrBeyond(start, end);
```
FrankChen021: [P2] Reject reversed bounded ranges instead of completing

For exclusive streams, this treats `start > end` the same as the valid empty case `start == end`, so a config like `start=500/end=100` makes every partition look complete and the supervisor can move to COMPLETED without creating a task. `KafkaIndexTaskIOConfig` would reject `end < start` if a task were created, but this completion check bypasses that validation. Reversed ranges should fail validation instead of being reported as successful bounded ingestion.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 2 |
| P3 | 0 |
| Total | 3 |

Reviewed 47 of 47 changed files.

- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaDataSourceMetadata.java:120 - [P2] Check bounded config before multi-topic metadata fallback. Adding `boundedStreamConfig` to metadata equality makes line 121 reject different bounded configs initially, but the multi-topic fallback then compares only sequence numbers and can return true for metadata from a different bounded run. In multi-topic bounded ingestion, that lets transactional metadata matching ignore the bounded config mismatch and can allow stale tasks to publish or overwrite metadata. Add an explicit `boundedStreamConfig` equality check before entering the fallback.

---

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
aho135: Thanks for the suggestion. That is much cleaner, I went ahead and made that update

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
aho135: Yeah good point, I think that would be more optimal. But since these are relatively cheap anyway I'll defer this change for now

aho135:
Thanks for all the reviews on this one @abhishekrb19 @FrankChen021 I'll go ahead and merge this one to checkpoint the work. Let me know if you spot anything else, and I'll include it in a follow up PR along with the resetOffsetAndBackfill change

PR #19373, opened 2026-04-24
feat: Config for MM / Indexer to always start enabled.

gianm:
PR #19368 updated the enabled/disabled state of an MM or Indexer to persist across restarts. In some cases, such as disabling a server before doing an in-place restart, it is more convenient to automatically re-enable on restart.

This patch adds a config parameter to restore the old behavior: `druid.worker.startAlwaysEnabled`.

**Release note:** MiddleManagers and Indexers now persist their enabled or disabled state across restarts. If you prefer the old behavior, where the server re-enables itself after a restart, set `druid.worker.startAlwaysEnabled = true`.


capistrant (approved):
pending the checkstyle fixup, this looks good

PR #19375, opened 2026-04-24
feat: rework SegmentFileBuilderV10 to organize internal containers by projection to be friendly to partial loading/eviction

clintropolis:
### Description
 Rework `SegmentFileBuilderV10` so V10 containers are scoped by a new declared file group concept instead of size alone. Groups provide a layout hint: a new container rolls when the declared group changes, giving readers a clean 1-to-N mapping between groups and containers that will support per-group partial loading in the future. Projections are the primary caller today, but the mechanism is deliberately generic (shared data across columns, internal metadata, etc.). 

changes:
* adds new method `SegmentFileBuilder.startFileGroup(@Nullable String)` which is a hint to the file builder about how to group the files a caller is adding
* `SegmentFileBuilderV10` replaced the internal `FileSmoosher` with a purpose-built container writer that rolls on group changes or when a file won't fit, organizing containers by the new file group concept
* `IndexMergerV10`/`IndexMergerBase` use the new file group method to organize v10 segments by projection
* removed stuff from `FileSmoosher` that was only needed by v10 builder delegating to it
* `AggregateProjectionSpec` validates that the projection cannot be named `__base`

### DIFF
--- a/processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
+++ b/processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
@@ -39,6 +39,7 @@ import org.apache.druid.segment.VirtualColumns;
 import org.apache.druid.segment.column.ColumnHolder;
 import org.apache.druid.segment.column.ValueType;
 import org.apache.druid.segment.projections.AggregateProjectionSchema;
+import org.apache.druid.segment.projections.Projections;
 import org.apache.druid.utils.CollectionUtils;
 import org.joda.time.DateTimeZone;
 
@@ -101,6 +102,9 @@ public class AggregateProjectionSpec
   {
     if (name == null || name.isEmpty()) {
       throw InvalidInput.exception("projection name cannot be null or empty");
+    }
+    if (Projections.BASE_TABLE_PROJECTION_NAME.equals(name)) {
+      throw InvalidInput.exception("projection cannot use reserved name[%s]", Projections.BASE_TABLE_PROJECTION_NAME);
     }
     this.name = name;
     if (CollectionUtils.isNullOrEmpty(groupingColumns) && (aggregators == null || aggregators.length == 0)) {
--- a/processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
+++ b/processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
@@ -100,13 +100,7 @@ public class AggregateProjectionSpec
       @JsonProperty("aggregators") @Nullable AggregatorFactory[] aggregators
   )
   {
-    if (name == null || name.isEmpty()) {
-      throw InvalidInput.exception("projection name cannot be null or empty");
-    }
-    if (Projections.BASE_TABLE_PROJECTION_NAME.equals(name)) {
-      throw InvalidInput.exception("projection cannot use reserved name[%s]", Projections.BASE_TABLE_PROJECTION_NAME);
-    }
-    this.name = name;
+    this.name = Projections.validateProjectionName(name);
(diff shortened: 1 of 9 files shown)


FrankChen021:
- processing/src/main/java/org/apache/druid/segment/file/SegmentFileBuilderV10.java:718 - P1 Delegated nested writes lose their file group. Nested addWithChannel calls are delegated while writerCurrentlyInUse is true, but mergeDelegatedFiles later replays those files with add(name, file) using whatever currentFileGroup is active at merge time. The original group active when the nested write was requested is not retained. If a nested file is created under one projection/container group and merged after the outer writer advances or clears the group, that nested file can be appended to the wrong container group, breaking the projection-to-container layout that partial loading depends on. Store the file group with each delegated temp file and restore/use it when merging.

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
```diff
@@ -102,6 +103,9 @@ public AggregateProjectionSpec(
     if (name == null || name.isEmpty()) {
       throw InvalidInput.exception("projection name cannot be null or empty");
     }
+    if (Projections.BASE_TABLE_PROJECTION_NAME.equals(name)) {
```
gianm: Might as well reserve all names starting with `__`, just in case we need another one later.

gianm (commented):
Seems ok to me although I'm wondering about Frank's comment.

processing/src/main/java/org/apache/druid/data/input/impl/AggregateProjectionSpec.java
clintropolis: done, moved to a method on `Projections`

clintropolis:
>Seems ok to me although I'm wondering about Frank's comment.

Its like kind of real, but also maybe unlikely to ever happen in practice given standard uses of nested writers. The usage is like a column serializer is open and writing and needs to open a nested writer to write some of its parts to separate files which needs some temp files to do that get merged back in later. The problem described would occur only i think if someone closed the outer thingy but not the nested, then set a new group, then close the nested ones (which i'm not entirely sure how would occur with our current index merger paths). Nothing does this, because its pretty strange?

I went ahead and fixed the problem. Well, with the same limitations as the builder currently has here, that if the files aren't added to a group contiguously then they get stored in different containers, but it does at least ensure that containers have at most 1 group. So, if someone manages to do ^, it will put them into separate container associated with the group at the time of delegation instead of the current active group. 🤷

FrankChen021 (approved):

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -57,6 +57,17 @@ public class Projections
 
   private static final ConcurrentHashMap<byte[], Boolean> PERIOD_GRAN_CACHE = new ConcurrentHashMap<>();
 
+  public static String validateProjectionName(@Nullable String name)
+  {
+    if (name == null || name.isEmpty()) {
+      throw InvalidInput.exception("projection name cannot be null or empty");
+    }
+    if (name.startsWith("__")) {
+      throw InvalidInput.exception("projection cannot use reserved name[%s]", BASE_TABLE_PROJECTION_NAME);
```
capistrant, line 66: implies `name` is `__base` but don't we only know that name is prefixed with `__` at this point? could be misleading in the messaging. does it make more sense to say `projection[name] cannot use reserved prefix[__]`

gianm (approved):

capistrant (commented):
this already got merged while I was looking through it but I will post the one pending comment I had so far. not a big deal I don't think

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
clintropolis: oops, yea, will fix this up in another PR

PR #19376, opened 2026-04-24
feat: Include caller thread name in LimitedPoolIndexMerger threads.

gianm:
In Indexers (CliIndexer), a thread pool LimitedPoolIndexMerger is used to control the concurrency of IndexMerger operations and avoid excessive memory usage. This patch updates the thread names of this pool to reflect the calling thread, which will include the task ID.


FrankChen021 (approved):

PR #19377, opened 2026-04-25
refactor: remove zookeeper segment announcement and discovery

clintropolis:
### Description
Druid has carried both ZooKeeper- and HTTP-based paths for segment announcement and inventory for several releases. The ZK paths have been `@Deprecated` for years, off by default (druid.serverview.type=http is the default), and segment-publishing path, the ZK-based inventory view, and the now-orphaned `ZkCoordinator` / `DataSegmentServerAnnouncer` / `CuratorDataSegmentServerAnnouncer`.

After this change, ZooKeeper's responsibilities in Druid are limited to:                                                                
  1. Coordinator and Overlord leader election
  2. Service (node) announcement and discovery
  3. Overlord / MiddleManager task management (out of scope for this PR; but will look to remove this too in a follow-up)

`BatchDataSegmentAnnouncer` seems a lot nicer now since it only has to track change history for the lister resource,  to something better in a follow-up, but skipped for now.

#### ZooKeeper-based segment announcement and discovery removed

The ZooKeeper-based segment announcement and inventory view, which have been deprecated and off by default for several releases, have been removed. The HTTP-based path (the default for `druid.serverview.type=http`) is now the only supported option.

If your configuration sets `druid.serverview.type` to anything other than `http`, startup will now fail with a clear error message. Remove the property (or set it to `http`, which is the default) to proceed.

The following configuration properties are no longer recognized and should be removed from `common.runtime.properties`:

- `druid.announcer.segmentsPerNode`
- `druid.announcer.maxBytesPerNode`
- `druid.announcer.skipLoadSpec`
- `druid.announcer.skipDimensionsAndMetrics`
- `druid.announcer.skipSegmentAnnouncementOnZk`
- `druid.zk.paths.announcementsPath`
- `druid.zk.paths.liveSegmentsPath`
- `druid.zk.paths.propertiesPath`
- `druid.zk.paths.connectorPath`

ZooKeeper is still used for leader election, service (node) announcement and discovery, and Overlord-to-MiddleManager task management.


FrankChen021 (approved):

server/src/test/java/org/apache/druid/client/BrokerServerViewTest.java
```diff
@@ -1,871 +0,0 @@
...
-
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.Executor;
-import java.util.stream.Collectors;
-
-public class BrokerServerViewTest extends CuratorTestBase
```
gianm, line 75: There's coverage of `BrokerServerView` in this test that we should preserve, such as testing for handling of tiers, etc. It should be reworked to use some other base view.

server/src/test/java/org/apache/druid/client/CoordinatorServerViewTest.java
```diff
@@ -1,502 +0,0 @@
...
-import org.junit.runners.Parameterized;
-
-import java.util.Arrays;
-import java.util.List;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.Executor;
-import java.util.concurrent.ExecutorService;
-
-@RunWith(Parameterized.class)
-public class CoordinatorServerViewTest extends CuratorTestBase
```
gianm, line 62: Like `BrokerServerView`, the tests here should be reworked to use some other base view.

server/src/test/java/org/apache/druid/server/coordination/coordination/BatchDataSegmentAnnouncerTest.java
```diff
@@ -19,628 +19,106 @@
...
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicInteger;
 
-/**
- *
- */
 public class BatchDataSegmentAnnouncerTest
```
gianm, line 53: Some of the removed tests in this file look like they are testing functionality that still exists, so they should be restored. Including:

- testAnnounceSegmentWithSameSegmentConcurrently
- testAnnounceSegmentsWithSameSegmentConcurrently
- testSchemaAnnounce

gianm (commented):
👍  on the removal. I think a few too many tests got removed, as mentioned in line comments. Beyond that there's a couple of references that could also be removed:

- in `docs/configuration/logging.md`, the example log file mentions `BatchServerInventoryView` which doesn't exist anymore
- in `MainTest` there is a setting of `druid.serverview.type = batch` in one of the tests. It's expecting an exception to be thrown. I believe it's now passing for a different reason than intended.

server/src/test/java/org/apache/druid/client/BrokerServerViewTest.java
clintropolis: oops, restored, thanks for catching

FrankChen021 (approved):

FrankChen021 (approved):

gianm (approved):

docs/design/zookeeper.md
```diff
@@ -29,43 +29,45 @@ Apache Druid uses [Apache ZooKeeper](http://zookeeper.apache.org/) (ZK) for mana
...
 ```
 
-Which signifies that they exist. They will also subsequently create a permanent znode at
+## Service announcement and discovery
+
+All Druid services publish their presence by creating an ephemeral znode at
 
 ```
-${druid.zk.paths.liveSegmentsPath}/${druid.host}
+${druid.zk.paths.announcementsPath}/${druid.host}
```
gianm: Is `druid.zk.paths.announcementsPath` really still used? It looked like this PR removed it.

server/src/main/java/org/apache/druid/server/initialization/ZkPathsConfig.java
```diff
@@ -27,42 +27,20 @@ public class ZkPathsConfig
   @JsonProperty
   private String base = "druid";
   @JsonProperty
-  private String propertiesPath;
-  @JsonProperty
   private String announcementsPath;
```
gianm: It looks like this isn't used anymore

gianm (commented):
I know I just approved this but a few other things to consider removing are inline. Beyond that, check `docs/api-reference/service-status-api.md`, it references "Zookeeper segment discovery" which doesn't happen anymore.

docs/design/zookeeper.md
clintropolis: oops, yea it can go too

gianm (approved):

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

- server/src/main/java/org/apache/druid/server/coordinator/rules/PartialLoadRule.java:94 - [P1] Partial load rules still enqueue full segment loads. `RunRules` calls `rule.run(...)` after `appliesTo`, but this implementation discards the matcher's `wrappedLoadSpec` and `fingerprint` and calls `replicateSegment` with the original `DataSegment`. Any configured `loadPartial*` rule whose matcher resolves will therefore load the full segment, not the selected projections, so the new public rule types silently violate the partial-load contract until the action path can pass the wrapped load spec through.

---

clintropolis:
>server/src/main/java/org/apache/druid/server/coordinator/rules/PartialLoadRule.java:94 

I guess this is a comment for wrong PR?

PR #19378, opened 2026-04-26
feat: CBA idle re-modeling and separate scale-up / scale-down task-count boundaries

Fly-Style:
This PR updates the seekable-stream cost-based autoscaler to make task-count decisions more stable and easier to reason about.

The main behavioral change is replacing the previous linear idle cost with a U-shaped idle cost centered around an ideal idle ratio. This penalizes both under-provisioning, where tasks have too little idle headroom, and over-provisioning, where tasks spend too much time idle. The `predictedIdleRatio` clamp at 0 made the U-shape's under-provisioning penalty saturate, so smaller `taskCount` always won the cost race — the optimizer scaled down a   
busy, lag-free cluster. The goal is to keep ingestion tasks near a practical operating point instead of treating all additional idle time as uniformly bad.

This also separates task-count boundary controls for scale-up and scale-down. Scale-up remains unbounded by default so the autoscaler can react aggressively to lag, while scale-down is bounded by default to avoid large drops in task count. Candidate task counts are still generated from valid partitions-per-task ratios, but the optimizer can now limit which candidates are evaluated depending on the configured scale direction boundary.

### What Changed

  - Added U-shaped idle cost in `WeightedCostFunction`, with an ideal idle ratio and asymmetric penalties for under- and over-provisioning.
  - Updated default cost weights to favor a more balanced lag/idle tradeoff.
  - Split the single `useTaskCountBoundaries` setting into separate for scale up/scale down
  - Added `CostResult.INFINITE_COST` so skipped candidates can still be represented safely in cost tables.
  - Updated supervisor docs for the cost-based autoscaler behavior and config options.
  - Expanded tests around U-shaped idle cost, config serialization/defaults, valid task-count generation, and bounded vs unbounded task-count jumps.
  - Removed `highLagThreshold` and `useTaskCountBoundaries` parameter, as obsolete, but kept in ctor for b/w compatibility.
  - lag amplification multiplier was increased due to changes in idle calculation formula from a point of view where scale-up/scale-down decisions are 'normal' in terms of normal distribution near `0.5/0.5` weights. `0.4` is a good amplification multiplier that was picked after a series of tests and calculations. `0.4/0.6` as default weights was picked from a conservative point of view. A Python script with computations is available by request.

#### Updated scaleup scenery, visualized.

<details>

Task boundaries are disabled, lag = 50k, current `taskCount` is 1. Plot contains *p** as partitions count, and idle as current `poll-idle-avg-ratio` metric.
 
scaleup

</details>

#### Scaledown scenery, visualized.
 
<details>

Task boundaries for scale-down are disabled, `taskCount` = `partitionCount` = 128, lag = 0.

cost_based_scaledown_compare

</details>

This PR has:

- [ ] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -97,9 +105,12 @@ public CostBasedAutoScalerConfig(
     // Cost function weights with defaults
     this.lagWeight = Configs.valueOrDefault(lagWeight, DEFAULT_LAG_WEIGHT);
     this.idleWeight = Configs.valueOrDefault(idleWeight, DEFAULT_IDLE_WEIGHT);
-    this.useTaskCountBoundaries = Configs.valueOrDefault(useTaskCountBoundaries, false);
-    this.highLagThreshold = Configs.valueOrDefault(highLagThreshold, -1);
-    this.minScaleUpDelay = Configs.valueOrDefault(minScaleUpDelay, Duration.millis(this.minTriggerScaleActionFrequencyMillis));
+    this.useTaskCountBoundariesOnScaleUp = Configs.valueOrDefault(useTaskCountBoundariesOnScaleUp, false);
```
FrankChen021, line 111: [P2] Legacy boundary setting is accepted but ignored

The constructor still accepts the legacy `useTaskCountBoundaries` property, but the new scale-up/down fields are initialized only from `useTaskCountBoundariesOnScaleUp` and `useTaskCountBoundariesOnScaleDown`. Existing supervisor specs with `useTaskCountBoundaries: true` will silently lose the scale-up boundary after upgrade, allowing unbounded jumps to any candidate task count. Map the legacy value to the new fields when the new fields are absent, or reject/document a breaking config change.

Fly-Style: This autoscaler was in experimental mode, but I will log on warn level and document the breaking change.

Fly-Style: Done in : https://github.com/apache/druid/pull/19378/changes/5b61b6dba06f08086e8bbc26f7dd652474877050

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -64,21 +66,19 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
...
-  static final int MAX_INCREASE_IN_PARTITIONS_PER_TASK = 2;
-  static final int MAX_DECREASE_IN_PARTITIONS_PER_TASK = MAX_INCREASE_IN_PARTITIONS_PER_TASK * 2;
-
   /**
-   * If average partition lag crosses this value and the processing rate is
-   * still zero, scaling actions are skipped and an alert is raised.
+   * Maximum number of candidate task counts to evaluate above or below the current task count
+   * when scale-up or scale-down boundaries are enabled.
+   * <p>
+   * The misspelling is preserved to avoid unnecessary churn in this package-private constant.
```
gianm: I don't understand this comment.  The constant is new in this patch. Please fix the spelling.

docs/ingestion/supervisor.md
```diff
@@ -193,29 +193,30 @@ The following example shows a supervisor spec with `lagBased` autoscaler:
...
+The cost-based autoscaler picks the number of ingestion tasks that minimizes a combined cost score. The score has two components:
 
-**It is experimental and the implementation details as well as cost function parameters are subject to change.**
+- **Lag cost** — how long it would take to drain the current backlog at the observed processing rate. More tasks reduce this cost.
+- **Idle cost** — how far the predicted idle ratio is from the target of ~25%. Tasks that are too busy (under-provisioned) or too idle (over-provisioned) both drive the score up. 
+The sweet spot is roughly 25% idle, giving headroom to absorb traffic spikes without wasting resources.
+
+At every evaluation interval, Druid computes the score for each candidate task count and picks the one with the lowest total cost.
 
 Note: Kinesis is not supported yet, support is in progress.
```
gianm, line 206: Is it really in progress?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -24,17 +24,23 @@
...
 public class WeightedCostFunction
 {
   private static final Logger log = new Logger(WeightedCostFunction.class);
+
   /**
    * Multiplier for a lag amplification factor; it was carefully chosen
    * during extensive testing as the most balanced multiplier for high-lag recovery.
    */
-  static final double LAG_AMPLIFICATION_MULTIPLIER = 0.05;
+  static final double LAG_AMPLIFICATION_MULTIPLIER = 0.4;
```
gianm, line 43: Why this change?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
```diff
@@ -152,8 +152,8 @@ public void test_autoScaler_computesOptimalTaskCountAndProducesScaleUp()
         .taskCountStart(lowInitialTaskCount)
         .scaleActionPeriodMillis(500)
         .minTriggerScaleActionFrequencyMillis(1000)
-        .lagWeight(0.2)
-        .idleWeight(0.8)
+        .lagWeight(0.8)
```
gianm, line 155: What will be the effect of the change to lag and idle weights?

gianm (commented):
Something seems off with the "Scaledown scenery, visualized" plot. It says the conditions are:

> Task boundaries for scale-down are enabled, taskCount = partitionCount = 128, lag = 0.

It shows that when current idle ratio is < 0.15 or so, the optimal task count becomes ~40. Scaling down 3x when current idle ratio is low will likely lead to the new set of tasks being overloaded.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
Fly-Style: It was passing without any problems in normal circumstances. The main idea of the change is to reduce the potential of not scaling over the timeout due to CI CPU pressure.

Fly-Style:
> It shows that when current idle ratio is < 0.15 or so, the optimal task count becomes ~40. Scaling down 3x when current idle ratio is low will likely lead to the new set of tasks being overloaded.

Oh, this is a lag in my head, apologies. 🤦🏻

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: I will note it in the patch notes. Generally, the intention was to find a point where scale-up/scale-down decisions are 'normal' in terms of normal distribution near `0.5/0.5` weights. `0.4` is a good amplification multiplier. `0.4/0.6` as default weights was picked from conservativity point of view.

docs/ingestion/supervisor.md
Fly-Style: I need to verify if anybody have a Kinesis workload with CBA working. If you want, we can remove that part.

FrankChen021 (approved):

docs/ingestion/supervisor.md
gianm: I'd rather remove it. My feeling is the documentation should reference things that exist currently rather than speculating about future development.

gianm (approved):

Issue #4373, opened 2017-06-06
[Proposal] Add batching support in kafka-indexing-service


himanshug:
It would benefit some users if kafka-indexing service supported the capability to let users put multiple druid InputRow inside single kafka record.
This allows users to do batching while still using kafka sync producer which allows only one kafka record at a time.

I would imagine adding following method to `InputRowParser` interface.
```
  default List<InputRow> parseBatch(T input)
  {
    return ImmutableList.of(parse(input));
  }
```

current `InputRow parse(T input)` method would be deprecated and all of druid code would be adjusted to use `parseBatch(input)` instead.

kafka-indexing service will need to persist <offset,row-number-in-record> pair instead of just offset to support exactly once. I believe, this will end up being the biggest change really.

This approach would add batching support in general and not just for kafka-indexing .

gianm:
See also #3439.

himanshug:
We had few conversations internally and it appears tracking record-offset within kafka record introduces additional metadata to be tracked and increases code complexity for the Kafka Indexing Service related code. So, for now the proposal is to not track record-offset within kafka record and that means if `InputRowParser.parseBatch(kafka-record-bytes)` returns more than one `InputRow` then `exactly-once` behavior is not guaranteed but best effort. This behavior could be changed in future if necessary.
So, for now, a user who requires `exactly-once` guarantee would use batching support provided by kafka so that each kafka record stays single Druid row.

@gianm does that sound ok?

pjain1:
Also, we are planning to put this parse batch thing in KafkaIndexTask behind a config, which when set explicitly will enable parsing of batch otherwise the task will fail. This config can be documented with the warning that when enabled task does not support exactly once and can also be printed as warning message in the task logs.

gianm:
@himanshug In principle, it sounds ok to me to have some non-exactly-once-features as long as they're disabled by default and clearly marked. I think by default we should be exactly once and users should need to change configs somehow if they want to break that.

However there is a difficulty with dropping exactly-once behavior: the segment-allocate stuff (pending segments etc) assumes that replica tasks will do exactly the same thing given the same set of inputs. If they don't do exactly the same thing then they can get out of sync with regard to which messages go in which segments, and that can cause the segment sets not to line up, and that can cause allocation errors.

[...]

PR #19379, opened 2026-04-27
feat: Add `scan` transform to unnest arrays for realtime ingestion

### LINKED ISSUE #8446: flatten json to multiple rows
### Description

if i have an object like:

```js
{
  host: "myhost"
  metrics: [
    { name: "metric1", value: 100},
    { name: "metric2", value: 200}
  ]
}
```

flatten it to 2 rows like:

```
host: "myhost", metric: "metric1", value: 100
host: "myhost", metric: "metric2", value: 200
```

### Motivation

benefits of this approach:

1. Ability to transfer multiple rows of data in a single atomic unit

currently in order to achieve this, you need to use kafka transactional topics, which add even more overhead. with the mentioned approach, a single piece of json if transferred successfully results in 'all or nothing' for multiple rows.

2. Less data transfer

If a piece of data like this is currently received, it needs to be converted to multiple rows, transferred to kafka, then read in multiple separate rows by druid.
this results in memory/cpu overhead processsing all the redundant columns of data across those multiple rows. 
With the approach defined here, they are just defined once, resulting in an overall smaller payload.

3. decreased complexity

currently you need to have code in a transformer to translate the json above into multiple rows. all that code will no longer be required.

abhishekrb19:
Fixes https://github.com/apache/druid/issues/8446. Relates to https://github.com/apache/druid/issues/4373. 

This PR adds scan query to Druid's transformSpec that can unnest array-valued columns into individual rows during streaming ingestion and possibly do other things that are feasible with scan query semantics. This is similar to the UNNEST support in SQL with MSQ batch ingestion, but for realtime  ingestion.

-  New ScanTransform (type: "scan") — a multi-row transform that wraps an embedded scan query and reuses the existing ScanQueryEngine, UnnestDataSource and UnnestCursorFactory machinery to explode arrays at ingest time. Each input row is wrapped
   in a temporary single-row segment, the configured scan query runs against it, and the resulting rows are emitted downstream.

  - The scan query's data source uses "__input__" as the base table, with an unnest data source to specify which column to explode and an optional unnestFilter to filter array elements
  - Supports string arrays, object arrays, and nested arrays - objects are preserved through unnest

  This simplifies queries by avoiding UNNEST at query time and improves query performance since unnesting is done once at ingest time rather than repeatedly at query time.

 This can also potentially reduce Kafka I/O and storage, since batching multiple events into a single producer record gets better compression than sending them individually spanning multiple records.

Benchmarks can be found in https://github.com/apache/druid/pull/19379#issuecomment-5594130705

### Release note

 Scan transform for ingestion-time array unnesting. Added a new "scan" transform type in transformSpec that unnests array-valued columns into individual rows during streaming ingestion (similar to existing UNNEST functionality with Druid SQL  and the MSQ engine). The scan transform wraps a scan query with an unnest data source to explode arrays at ingest time.

This PR has:

- [x] been self-reviewed.


processing/src/main/java/org/apache/druid/segment/transform/ScanTransform.java
```diff
@@ -0,0 +1,268 @@
...
+  private final DimFilter unnestFilter;
+
+  @JsonCreator
+  public ScanTransform(
+      @JsonProperty("name") final String name,
+      @JsonProperty("unnestColumn") final VirtualColumn unnestColumn,
+      @JsonProperty("unnestFilter") @Nullable final DimFilter unnestFilter
+  )
+  {
+    this.name = name;
```
FrankChen021: [P2] Validate scan transform output name

ScanTransform stores `name` separately from `unnestColumn.getOutputName()`, but the transform framework treats `Transform.getName()` as the generated field name for collision checks and input-column pruning. If a spec sets `name: "tag"` but the virtual column outputs `"elt"`, Druid accepts it, writes `elt`, and silently leaves references to `tag` unresolved while also missing collisions on `elt`. Reject mismatches or derive the transform name from the virtual column output.

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
```diff
@@ -0,0 +1,285 @@
...
+      } else {
+        event.put(col, factory.makeColumnValueSelector(col).getObject());
+      }
+    }
+    return new MapBasedInputRow(inputRow.getTimestampFromEpoch(), dimensionColumns, event);
+  }
+
+  private List<String> resolveColumnsForRow(final InputRow inputRow)
+  {
+    final Set<String> columns = new LinkedHashSet<>();
```
FrankChen021, line 191: [P1] Preserve non-dimension raw columns in scan output

resolveColumnsForRow only emits __time, inputRow.getDimensions(), scan virtual columns, and unnest output columns. In ingestion, fields used by metricsSpec are commonly excluded from dimensions, so after a scan transform expands a row those raw metric fields disappear from the MapBasedInputRow event and downstream aggregators read null/zero instead of the original value. The previous implementation explicitly included MapBasedInputRow event keys, and the incremental diff removes the test that covered this case. Include all raw event keys, where available, when constructing scan output rows.

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
```diff
@@ -0,0 +1,285 @@
...
+  {
+    baseCursorFactory.set(inputRow);
+
+    if (cursor == null) {
+      cursor = cursorHolder.asCursor();
+    } else {
+      cursor.reset();
+    }
+
+    if (cursor == null || cursor.isDone()) {
```
FrankChen021, line 168: [P2] Do not pass through rows rejected by scan filters

When the scan cursor is done, the transformer always returns buildPassthroughRow(inputRow). That makes a scan query filter unable to drop rows: a row rejected by query.getFilter() is still ingested once with unnest outputs set to null, as the updated test now asserts. This diverges from normal TransformSpec filtering and from scan/unnest filtering semantics, and it inflates row counts for specs that use scan filters to exclude input or unnested rows. Distinguish empty/null unnest passthrough from filter rejection and return an empty list when the filter rejects the row.

jtuglu1:
Have we done performance profiles on this? Generally I've found transforms to really slow down tput on streaming ingests.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
```diff
@@ -0,0 +1,272 @@
...
+  private List<String> resolveColumnsForRow(final InputRow inputRow)
+  {
+    final Set<String> columns = new LinkedHashSet<>();
+    columns.add(ColumnHolder.TIME_COLUMN_NAME);
+    columns.addAll(inputRow.getDimensions());
+    // Include raw event fields that aren't in getDimensions() — e.g. metric inputs that DataSchema added
+    // to dimensionExclusions. Without this, fixed-dimension ingestions with metrics would read null for
+    // their metric source fields in expanded rows.
+    if (inputRow instanceof MapBasedInputRow) {
+      columns.addAll(((MapBasedInputRow) inputRow).getEvent().keySet());
```
FrankChen021, line 198: [P2] Keep metric source fields out of dimensions

Adding every MapBasedInputRow event key to the resolved scan columns also feeds those keys into resolveDimensionColumns, which promotes metric input fields such as bytes_sent into row.getDimensions() after expansion. IncrementalIndex treats row.getDimensions() as authoritative and will discover/store those metric source fields as dimensions even though DataSchema excluded them for metricsSpec. Preserve the raw field in the event map for aggregators, but do not add non-dimension event fields to the output dimension list; add a test that verifies the metric source is not ingested as a dimension.

abhishekrb19: Addressed in the latest patch.

abhishekrb19:
Benchmark 1 — transform mode comparison (unnestArraySize=5, 5×10s warmup + 10×10s measurement, 1 fork, 10k rows/op)

```
Benchmark                                       (mode)  (unnestArraySize)  Mode  Cnt     Score    Error  Units
TransformSpecBenchmark.transformRows              NONE                  5  avgt   10     0.542 ±  0.006  ns/op
TransformSpecBenchmark.transformRows        EXPRESSION                  5  avgt   10     5.985 ±  0.113  ns/op
TransformSpecBenchmark.transformRows  SCAN_PASSTHROUGH                  5  avgt   10   369.312 ±  9.751  ns/op
TransformSpecBenchmark.transformRows           SCAN_VC                  5  avgt   10  1851.823 ± 61.218  ns/op
TransformSpecBenchmark.transformRows       SCAN_UNNEST                  5  avgt   10  1061.146 ±  7.973  ns/op
```

Benchmark 2 — SCAN_UNNEST batch-size scaling (mode=SCAN_UNNEST, 5×10s warmup + 10×10s measurement, 1 fork, 10k rows/op)

```
Benchmark                                  (mode)  (unnestArraySize)  Mode  Cnt      Score      Error  Units
TransformSpecBenchmark.transformRows  SCAN_UNNEST                  5  avgt   10    963.466 ±   11.975  ns/op
TransformSpecBenchmark.transformRows  SCAN_UNNEST                 20  avgt   10   2237.936 ±   17.684  ns/op
TransformSpecBenchmark.transformRows  SCAN_UNNEST                100  avgt   10   9156.740 ±  148.981  ns/op
TransformSpecBenchmark.transformRows  SCAN_UNNEST               1000  avgt   10  84573.609 ± 4139.941  ns/op
```


Relative ordering (fastest → slowest): NONE < EXPRESSION < SCAN_PASSTHROUGH < SCAN_UNNEST < SCAN_VC.
The classic EXPRESSION transform is cheaper than any scan-based path: even a bare scan pipeline with no virtual columns or unnest (SCAN_PASSTHROUGH) costs more than EXPRESSION doing real work, and running that same expression through a scan virtual column (SCAN_VC) is the most expensive mode measured - more than SCAN_UNNEST. There's a real integration penalty from routing through the scan-query engine that likely has room to be optimized.

For SCAN_UNNEST specifically, per-input-row overhead drops as the unnest array grows, since the fixed per-row cost (cursor reset, selector setup) amortizes over more output rows - wider arrays are cheaper per output row than narrow ones.

Note: this benchmark isolates the transform stage only - it excludes Kafka fetch/poll, JSON parsing, and downstream indexing/persist, which are likely larger fixed costs per record in a real ingestion task. So these numbers are an upper bound on the transform stage's overhead, not a prediction of overall end-to-end throughput impact. The real-world impact may in fact be smaller, since Kafka I/O and other per-record costs are incurred regardless of transform mode and would dilute the relative overhead once folded into total per-record cost.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

The supplied concern about raw `MapBasedInputRow` metric fields being promoted to dimensions is resolved at the current head; the new logic preserves those fields in the expanded event map while excluding them from `getDimensions()`. Two related P2 findings remain in `ScanTransformer`: generated scan outputs used only as metrics are still added as dimensions, and list-backed input rows do not preserve non-dimension metric fields.



---

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
```diff
@@ -0,0 +1,293 @@
...
+  private static List<String> resolveDimensionColumns(
+      final InputRow inputRow,
+      @Nullable final List<String> resultColumns,
+      final Set<String> nonDimensionEventFields
+  )
+  {
+    final LinkedHashSet<String> dims = new LinkedHashSet<>(inputRow.getDimensions());
+    if (resultColumns != null) {
+      for (final String col : resultColumns) {
+        if (!ColumnHolder.TIME_COLUMN_NAME.equals(col) && !nonDimensionEventFields.contains(col)) {
```
FrankChen021: P2 Keep generated metric fields out of dimensions

`nonDimensionEventFields` contains only keys from the pre-scan `MapBasedInputRow`. A scan virtual-column output or unnested output used as a metric is generated after that step, so it is absent from this set and this loop adds it to `dimensionColumns`. For example, with a virtual column `rate` used by `longSum(total, rate)` but not declared as a dimension, every expanded row still reports `rate` from `getDimensions()`, causing the metric input to be stored as a discovered dimension. Track metric-only generated outputs when deriving dimensions while retaining them in the event map.

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  /**
+   * Returns the raw event fields present on {@code inputRow} that are not in {@link InputRow#getDimensions()}
+   * — e.g. metric inputs that {@code DataSchema} added to dimensionExclusions. These must still be read
+   * into the expanded rows' event maps (for aggregators), but must not be promoted to dimensions.
+   */
+  private static Set<String> resolveNonDimensionEventFields(final InputRow inputRow)
+  {
+    if (!(inputRow instanceof MapBasedInputRow)) {
```
FrankChen021: P2 Preserve metric fields for list-backed rows

This guard returns an empty set for every `ListBasedInputRow`, even though its `RowSignature` and data contain fields that `InputRow.getDimensions()` intentionally excludes, such as metric inputs from CSV or delimited ingestion. Since `ScanTransformSpec` is wired through the generic `DataSchema`/`InputSourceReader` path, a list-backed row with a metric field and scan expansion omits that field from the expanded event map; the aggregator then reads null/zero. Enumerate non-dimension fields from the list-backed row signature (or pass the input schema) as well as from `MapBasedInputRow`.

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
FrankChen021: Confirmed at head `072990954936d6755dd94460a90d4e088b40732a`: the raw `MapBasedInputRow` case is addressed. The transformer now retains those event keys for selectors/aggregators while excluding fields absent from `inputRow.getDimensions()` from the output dimension list, and the added unit and embedded coverage checks both sides of that behavior. I reviewed all 23 changed files; I recorded two related cases separately where generated metric outputs and list-backed metric fields still need the same treatment.

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
abhishekrb19: Fixed in the latest commit.

FrankChen021 (commented):
## 🟡 Changes recommended

Current changed files reviewed: 25 of 25.

I started from the prior review thread and verified that its two metric/dimension findings are fixed at head `865b0272c5f8a03ddec40869c84fb66e5d8f51f3`. Static review then covered the full merge-base diff, the incremental diff, and relevant surrounding scan, unnest, ingestion, sampler, and batch-input code. The current review has two P2 findings attached inline. Validation: `git diff --check 91fe02f0f256c55c9f2a7eeaa288f61b23163477...HEAD` passed; no builds or broad tests were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |



---

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
```diff
@@ -0,0 +1,317 @@
...
+    // Selectors are lazy views over the cursor's current position — create them once per column
+    // here, then re-read via getObject() as the cursor advances, rather than reallocating a selector
+    // for every (output-row x column) pair.
+    final ColumnValueSelector<?>[] selectors = new ColumnValueSelector<?>[columns.size()];
+    for (int i = 0; i < columns.size(); i++) {
+      selectors[i] = selectorFactory.makeColumnValueSelector(columns.get(i));
+    }
+
+    final List<InputRow> result = new ArrayList<>();
+    while (!cursor.isDone()) {
```
FrankChen021, line 192: P2 Apply ScanQuery pagination during expansion

`ScanTransformSpec` accepts a full `ScanQuery`, but this loop (and `SettableRowCursorFactory`) never uses `getScanRowsOffset()` or `getScanRowsLimit()`. Since the normal scan query toolchest applies those outside `ScanQueryEngine`, an embedded query such as `limit: 1` on a row with three unnested elements still emits all three rows into ingestion. Apply the query's offset/limit here, or reject non-default pagination so valid scan-query fields cannot silently change meaning.

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
```diff
@@ -0,0 +1,317 @@
...
+   */
+  private static Set<String> resolveNonDimensionEventFields(final InputRow inputRow)
+  {
+    final Set<String> allFields;
+    if (inputRow instanceof MapBasedInputRow) {
+      allFields = ((MapBasedInputRow) inputRow).getEvent().keySet();
+    } else if (inputRow instanceof ListBasedInputRow) {
+      allFields = ((ListBasedInputRow) inputRow).asMap().keySet();
+    } else {
+      return Set.of();
```
FrankChen021, line 222: P2 Preserve metric fields for non-map input rows

The new batch integration applies `ScanTransformSpec` to every `InputSourceReader`, but this fallback drops every raw field not listed in `getDimensions()` for any non-`MapBasedInputRow`/`ListBasedInputRow`. `DeltaInputSource` emits `DeltaInputRow`; with a metric source such as `bytes`, `ScanTransformer` therefore omits `bytes` from each expanded event and the aggregator reads null/0. Enumerate the required non-dimension fields (or reject unsupported row types) so batch inputs retain metric values.

processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
```diff
@@ -0,0 +1,339 @@
...
+      );
+    }
+    final Set<String> nonDimensionFields = new LinkedHashSet<>(allFields);
+    nonDimensionFields.removeAll(inputRow.getDimensions());
+    return nonDimensionFields;
+  }
+
+  private List<String> resolveColumnsForRow(final InputRow inputRow, final Set<String> nonDimensionEventFields)
+  {
+    final Set<String> columns = new LinkedHashSet<>();
```
FrankChen021, line 253: P2 [P2] Honor ScanQuery.columns projection

**Finding:** ScanTransformSpec accepts an embedded ScanQuery, whose columns property is the documented projection, but resolveColumnsForRow unconditionally adds every input dimension, every non-dimension event field, every query virtual column, and every unnest output. A query with columns: ["tag"] therefore still emits fields such as user, tags, and unrelated metrics, silently diverging from ScanQuery semantics and potentially storing fields the query explicitly excluded.

**Suggestion:** Apply query.getColumns() when constructing the output column set, while explicitly retaining any raw fields that ingestion still requires for filters or aggregators, or reject non-empty projections until that distinction is implemented. Add coverage for a non-empty columns list.

FrankChen021 (commented):
## 🟡 Changes recommended

Current changed files reviewed: 25 of 25.

I started from the incremental diff, then reviewed the full current merge-base diff and relevant surrounding scan, unnest, ingestion, sampler, streaming, batch-input, and query-engine code. The two P2 findings from the prior review—generated metric outputs being promoted to dimensions and list-backed metric fields being dropped—are fixed at the current head; the new unit and embedded coverage exercises both cases. The current review has one P2 finding attached inline. Validation: `git diff --check 13ae31afc23baf6102fd6d88d0bd813a33b174aa...a67749979b5fcb240d4e256903a3506efb7bab64` passed; no builds or broad tests were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |



---

benchmarks/src/test/java/org/apache/druid/benchmark/TransformSpecBenchmark.java
```diff
@@ -0,0 +1,229 @@
...
+ *                       work to EXPRESSION but via the scan pipeline.
+ *   SCAN_UNNEST       — ScanTransformSpec that unnests an array of {@code unnestArraySize} elements.
+ *
+ * Output is nanoseconds per input row. For SCAN_UNNEST, use {@link #totalOutputRowsPerInput()} to
+ * translate to ns/output-row.
+ *
+ * The input row is a {@link MapBasedInputRow} matching what Kafka JSON ingestion produces.
+ */
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.AverageTime)
```
FrankChen021, line 79: [P1] Resolve the shadowed JMH Mode type

**Finding:** The class imports JMH's org.openjdk.jmh.annotations.Mode but also declares a nested enum named Mode. In @BenchmarkMode(Mode.AverageTime), the nested enum wins name resolution and has no AverageTime constant, so this new benchmark source does not compile and blocks the benchmarks module/build.

**Suggestion:** Rename the benchmark parameter enum (for example, to BenchmarkMode) or fully qualify org.openjdk.jmh.annotations.Mode in the annotation so the JMH mode type is selected.

FrankChen021 (commented):
## 🟡 Changes recommended

Reviewed prepared PR https://github.com/apache/druid/pull/19379 at head e1642ff3a4866454a14b7cf30577e62cc3e04aa5.

Review scope and coverage: started with the incremental diff, then reviewed the full current diff, prior findings, and surrounding code. Reviewed all 25 of 25 current changed files:

- benchmarks/src/test/java/org/apache/druid/benchmark/TransformSpecBenchmark.java
- docs/ingestion/ingestion-spec.md
- embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaScanTransformTest.java
- indexing-service/src/main/java/org/apache/druid/indexing/common/task/AbstractBatchIndexTask.java
- indexing-service/src/main/java/org/apache/druid/indexing/input/InputRowSchemas.java
- indexing-service/src/main/java/org/apache/druid/indexing/overlord/sampler/InputSourceSampler.java
- indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SettableByteEntityReader.java
- indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/StreamChunkReader.java
- processing/src/main/java/org/apache/druid/segment/transform/BaseTransformSpec.java
- processing/src/main/java/org/apache/druid/segment/transform/BaseTransformer.java
- processing/src/main/java/org/apache/druid/segment/transform/CompactionTransformSpec.java
- processing/src/main/java/org/apache/druid/segment/transform/ScanTransformSpec.java
- processing/src/main/java/org/apache/druid/segment/transform/ScanTransformer.java
- processing/src/main/java/org/apache/druid/segment/transform/SettableRowCursorFactory.java
- processing/src/main/java/org/apache/druid/segment/transform/Transform.java
- processing/src/main/java/org/apache/druid/segment/transform/TransformSpec.java
- processing/src/main/java/org/apache/druid/segment/transform/Transformer.java
- processing/src/main/java/org/apache/druid/segment/transform/TransformingInputEntityReader.java
- processing/src/main/java/org/apache/druid/segment/transform/TransformingInputSourceReader.java
- processing/src/test/java/org/apache/druid/segment/IndexBuilder.java
- processing/src/test/java/org/apache/druid/segment/generator/SegmentGenerator.java
- processing/src/test/java/org/apache/druid/segment/transform/ScanTransformTest.java
- processing/src/test/java/org/apache/druid/segment/transform/TransformerTest.java
- server/src/main/java/org/apache/druid/segment/indexing/DataSchema.java
- website/.spelling

One new P1 finding is inline on TransformSpecBenchmark.java: the nested Mode enum shadows JMH's imported Mode, so @BenchmarkMode(Mode.AverageTime) refers to a nonexistent enum constant and the benchmark source cannot compile. The prior scan-transform findings reviewed in the previous pass, including pagination, raw/list-backed metric preservation, filter handling, metric-field dimension exclusions, and query-column projection handling, are addressed in the current head.

Validation: git diff --check passed for the current PR range. No builds or broad tests were run, per review scope.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |



---

benchmarks/src/test/java/org/apache/druid/benchmark/TransformSpecBenchmark.java
abhishekrb19: >Finding: The class imports JMH's org.openjdk.jmh.annotations.Mode but also declares a nested enum named Mode. In @BenchmarkMode(Mode.AverageTime), the nested enum wins name resolution and has no AverageTime constant, so this new benchmark source does not compile and blocks the benchmarks module/build.

I just ran the benchmark locally and CI succeeded, so the code actually compiles fine. The LLM probably got tripped up by a shadowing rule.

FrankChen021 (commented):
## 🟡 Changes recommended

The supplied P1 on `benchmarks/src/test/java/org/apache/druid/benchmark/TransformSpecBenchmark.java:79` is not valid. The `@BenchmarkMode` annotation is on the top-level class declaration, outside the class body, so the nested `TransformSpecBenchmark.Mode` member is not in scope there; the imported JMH `Mode` resolves and the benchmark compiles.

I also re-reviewed the full current change and found one separate issue:

**[P2] Include nested unnest-filter fields in the required input columns** (`processing/src/main/java/org/apache/druid/segment/transform/ScanTransformSpec.java:120`)

`ScanTransformSpec.getRequiredColumns()` collects each unnest virtual column and the top-level scan filter, but it does not collect `UnnestDataSource.getUnnestFilter().getRequiredColumns()` at each nesting level. If an `unnestFilter` references a raw field that is not otherwise a dimension, metric, or transform input, `InputRowSchemas.createColumnsFilter()` can omit that field before `ScanTransformer` runs. The base filter then sees a missing/null value and can incorrectly reject rows. Include every nested unnest filter's required columns and add coverage for a filter on a raw non-dimension field.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |



---

PR #19382, opened 2026-04-27
fix: hang in PeriodGranularity when using sub-day compound periods with imprecise timezone days

benhopp:
**This is my first real PR, and was largely assisted by AI, so please scrutinize heavily.**

Fixes a bug where timestamp_floor with compound time-only periods falls back to an infinite loop when the timezone has daylight saving time.

### Description
The issue is caused by how Druid's PeriodGranularity attempts to bucket timestamps when timezones and daylight saving time (DST) are involved. 

When trying to bucket timestamps using a compound period like PT1M1S, Druid first tries a fast-path (`truncateMillisPeriod`). It checks if the timezone has precise days and hours. Because timezones with DST (like `America/New_York`) observe daylight saving time, Druid incorrectly flags the timezone as imprecise for all period durations. It then falls back to a slow-path (`truncateCompoundPeriod`), which literally runs a while loop, adding the period duration starting from January 1st, 1970, until it reaches the target timestamp. This means it loops over 28 million times for every single row in the query, causing it to hang.

This PR updates the `truncateMillisPeriod` logic to check if the period contains any inherently imprecise components like years, months, weeks, or days. If the period only contains hours, minutes, seconds, or milliseconds, it will safely convert the period to milliseconds and use the fast-path modulo math, avoiding the infinite while loop completely.

A regression test `PeriodGranularityBugTest.java` is included to validate this fix.

### DIFF
--- a/processing/src/main/java/org/apache/druid/java/util/common/granularity/PeriodGranularity.java
+++ b/processing/src/main/java/org/apache/druid/java/util/common/granularity/PeriodGranularity.java
@@ -600,6 +600,16 @@ public class PeriodGranularity extends Granularity implements JsonSerializable
         offset += millis;
       }
       return t - offset;
+    } else if (period.getYears() == 0 && period.getMonths() == 0 && period.getWeeks() == 0 && period.getDays() == 0) {
+      // If the period only contains hours, minutes, seconds, and millis,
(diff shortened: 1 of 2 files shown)


FrankChen021 (approved):

PR #19383, opened 2026-04-28
feat: add PartialQueryableIndex for loading columns on demand

clintropolis:
### Description

changes:
* adds `PartialQueryableIndex` which uses a `PartialSegmentFileMapperV10` to expose a `QueryableIndex` that supports loading columns on demand
* adds `ColumnDescriptor.toColumnType` to provide a way to collect `ColumnType` information from `ColumnPartSerde` so we can expose schema type information without downloading files
* adds `PartialQueryableIndexTest` to cover a variety of scenarios of partial load


FrankChen021 (approved):

FrankChen021 (commented):

---

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndex.java
```diff
@@ -0,0 +1,427 @@
...
+  {
+    return ordering;
+  }
+
+  @Nullable
+  @Override
+  public BaseColumnHolder getColumnHolder(String columnName)
+  {
+    final Supplier<BaseColumnHolder> supplier = baseColumns.get(columnName);
+    return supplier == null ? null : supplier.get();
```
gianm, line 241: This line triggers a download, right? If I read it right, I think this defeats the background-downloading stuff we have in `ReadableInputQueue`. I believe it's going to start processing once the segment metadata is downloaded (rather than the entire segment), and then we'll block a processing thread on this line to fetch the piece of the segment that we need.

Do you have thoughts on how to make background fetching work with partial downloads? Maybe we can add a `ListenableFuture<CursorHolder> makeCursorHolder` in `CursorFactory`, which would fetch all the stuff needed by the `CursorBuildSpec`.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndex.java
```diff
@@ -0,0 +1,427 @@
...
+  }
+
+  /**
+   * Answers from metadata without triggering column downloads. The default implementation in {@link QueryableIndex}
+   * calls {@link #getColumnHolder(String)}, which would force a base table load.
+   * <p>
+   * Only {@link ColumnCapabilities#getType()} and {@link ColumnCapabilities#hasMultipleValues()} are populated from
+   * the metadata; richer fields ({@code isDictionaryEncoded}, {@code hasBitmapIndexes}, {@code hasNulls}, etc.) keep
+   * their default/UNKNOWN values. This is intentional:
```
gianm: This is a very sketchy detail to only exist on an implementation of the interface. Callers generally aren't going to know what implementation they're dealing with. These caveats should move to the `QueryableIndex` interface directly so callers know that if they need these richer fields, they need to fetch the `ColumnHolder`.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndex.java
```diff
@@ -0,0 +1,427 @@
...
+        continue;
+      }
+      if (projectionSpec.getSchema() instanceof AggregateProjectionSchema) {
+        aggProjections.add(
+            new AggregateProjectionMetadata(
+                (AggregateProjectionSchema) projectionSpec.getSchema(),
+                projectionSpec.getNumRows()
+            )
+        );
+        projectionSpecs.put(projectionSpec.getSchema().getName(), projectionSpec);
```
gianm, line 157: What happens to non-`AggregateProjectionSchema`? Are we ignoring them or do they get picked up somewhere else?

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndex.java
clintropolis: >This line triggers a download, right? 
yes indeed, 

>Do you have thoughts on how to make background fetching work with partial downloads? Maybe we can add a ListenableFuture<CursorHolder> makeCursorHolder in CursorFactory, which would fetch all the stuff needed by the CursorBuildSpec.

yes, this is more or less exactly what is in another branch I am working on that I will be opening probably today or tomorrow that will make the cursor factory have a future based way to get the holder, which we will integrate with the `ReturnOrAwait` stuff to not block processing.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndex.java
clintropolis: I agree, i want to remove this method from `QueryableIndex` and probably replace it with a `RowSignature` or something that can be used to get the type information that most callers need instead, but this PR didn't seem to be the best place to do that.

I can move this javadoc to the base interface until I do that refactor in a follow-up.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndex.java
clintropolis: this is an artifact of the way things currently are where there are essentially only aggregate projections that exist as projections today (since the base table projection defines the top level columns and stuff of the `QueryableIndex`). It is same-ish as the logic here https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/IndexIO.java#L986, I guess I could add the same defensive check we have there for now. This will also be changing a bit from some other work I have in flight but isn't a PR yet.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndex.java
gianm: Moving it to the interface for now sounds good. Although consider simplifying it a bit. It probably doesn't need to mention all the know callers like this.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndex.java
gianm: Some kind of acknowledgement would be good. Either a comment, or warning log, or defensive exception. Whichever one of those seems best.

gianm (approved):

PR #19386, opened 2026-04-28
feat: Add `now()` to troubleshoot pipeline latencies

### LINKED ISSUE #18208: Add possibility of having current date-time in Transformations (Druid ingestion)
### Description
Please add now() or a similarly named function to the transform capabilities in Druid Ingestion.
This function would read the current timestamp.

### Motivation
This is particularly useful in systems where the data processing time and the time of data insertion into Druid have significant differences.
Having such a function would enable analysis and locating the causes of delays.

abhishekrb19:
Fixes https://github.com/apache/druid/issues/18208

  Adds a `now()` expression macro to Druid's native expression. It returns the current system timestamp in milliseconds since epoch, evaluated per row at processing time. Intended primarily for ingestion-time troubleshooting - e.g.,
  capturing `now() - __time` as an ingestion_lag_ms dimension to measure end-to-end pipeline delay. This expression is only allowed for non-`__time` columns.

  Release note

  Added `now()` expression function that returns the current system timestamp in milliseconds since epoch. Useful at ingestion time for troubleshooting pipeline delays (e.g., now() - __time). Note: `now()` is non-deterministic as it evaluates for every row, so it can break idempotency. This can only be added to columns that are not `__time`.

This PR has:

- [x] been self-reviewed.


docs/querying/math-expr.md
```diff
@@ -110,6 +110,7 @@ The following built-in functions are available.
 
 |name|description|
 |----|-----------|
+|now|now() returns the current system timestamp in milliseconds since epoch (1970-01-01 00:00:00 UTC). This function is evaluated for each row at processing time. It's recommended to use this only for troubleshooting issues - see [Using now()in ingestion](#using-now-in-ingestion).|
```
aho135:
```suggestion
|now|now() returns the current system timestamp in milliseconds since epoch (1970-01-01 00:00:00 UTC). This function is evaluated for each row at processing time. It's recommended to use this only for troubleshooting issues - see [Using now() in ingestion](#using-now-in-ingestion).|
```

aho135:
Fixes https://github.com/apache/druid/issues/18208

aho135:
@abhishekrb19 Curious if you have any thoughts on this old issue: https://github.com/apache/druid/issues/6513 

Seems like the reservations were mainly for using `now()` for __time which you already mentioned should not be used. Maybe worth adding some validation/enforcement for this

docs/querying/math-expr.md
```diff
@@ -119,6 +120,26 @@ The following built-in functions are available.
...
 |timestamp_format|timestamp_format(expr, \[pattern, \[timezone\]\]) formats a timestamp as a string with a given [Joda DateTimeFormat pattern](http://www.joda.org/joda-time/apidocs/org/joda/time/format/DateTimeFormat), or ISO8601 if the pattern is not provided. The time zone, if provided, should be a time zone name like "America/Los_Angeles" or offset like "-08:00". Pattern and time zone must be literals.|
 
+### Using `now()` in ingestion
+
+:::warning
+`now()` is non-deterministic: replicated streaming tasks and task replays evaluate it to different
+values, producing inconsistent results across replicas. Do not use `now()` to overwrite `__time`.
+For Kafka, prefer [`kafka.timestamp`](../ingestion/data-formats.md#kafka) as the
+`__time` source.
```
aho135: I'm not sure if it makes sense to mention Kafka here. The timestamps aren't equivalent as one is the time the event lands on Kafka and the other is the time the event is ingested in Druid

docs/querying/math-expr.md
```diff
@@ -110,6 +110,7 @@ The following built-in functions are available.
 
 |name|description|
 |----|-----------|
+|now|now() returns the current system timestamp in milliseconds since epoch (1970-01-01 00:00:00 UTC). This function is evaluated for each row at processing time. It's recommended to use this only for troubleshooting issues - see [Using now() in ingestion](#using-now-in-ingestion).|
 |timestamp|timestamp(expr[,format-string]) parses string expr into date then returns milliseconds from java epoch. without 'format-string' it's regarded as ISO datetime format |
 |unix_timestamp|same with 'timestamp' function but returns seconds instead |
```
aho135:
```suggestion
|unix_timestamp|unix_timestamp(expr[,format-string]) same with 'timestamp' function but returns seconds instead |
```
Just for consistency with other functions

processing/src/main/java/org/apache/druid/math/expr/BuiltInExprMacros.java
```diff
@@ -214,4 +214,95 @@ public Object getLiteralValue()
...
+      {
+        return shuttle.visit(this);
+      }
+
+      @Override
+      public BindingAnalysis analyzeInputs()
+      {
+        // Return analysis indicating this is NOT constant
+        // by pretending we have a free variable
+        return new BindingAnalysis();
```
FrankChen021: [P2] Prevent now() from being planned as constant

Returning an empty BindingAnalysis makes ExpressionPlanner mark now() and expressions that only depend on now() as CONSTANT. Query-time expression selectors then build a ConstantExprEvalSelector and evaluate now() once when the selector is created, and expression filters/virtual columns also get stable cache keys, so native queries can reuse stale timestamps instead of evaluating at processing time. Add explicit non-deterministic handling, or otherwise keep now() out of constant selector/index/cache optimizations.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

processing/src/main/java/org/apache/druid/math/expr/BuiltInExprMacros.java
abhishekrb19: yes, good catch, this is fixed in the latest commit.

abhishekrb19:
>@abhishekrb19 Curious if you have any thoughts on this old issue: https://github.com/apache/druid/issues/6513
Seems like the reservations were mainly for using now() for __time which you already mentioned should not be used. Maybe worth adding some validation/enforcement for this

@aho135 @FrankChen021 - thanks for the reviews and for pointing me to the previous issues. I’ve added validation to restrict `now()` usage for non-`__time` columns and to throw an error when it’s used with the `__time` column.

processing/src/main/java/org/apache/druid/math/expr/BuiltInExprMacros.java
FrankChen021: Thanks, this fixes the `ExpressionPlan` constant trait, but I think the selector path can still fold `now()` once. `NowExpression.analyzeInputs()` now reports only the synthetic `__dummy__` binding; for normal segments that column has no capabilities and its selector is nil, so `ExpressionSelectors.createBindings` drops it and returns `InputBindings.nilBindings()`. `makeExprEvalSelector` then treats those nil bindings as a constant expression and returns `new ConstantExprEvalSelector(plan.getExpression().eval(bindings))`, so `now()` is still evaluated when the selector is created instead of at row/query processing time. The new tests cover the planner only; could you add selector-level coverage and avoid using a missing dummy input as the non-determinism marker?

abhishekrb19: Ack, updated the code with your non-determinism suggestion

FrankChen021: Thanks, the selector constant-folding path I called out is fixed now: `now()` carries a non-deterministic analysis flag and `ExpressionSelectors` skips `ConstantExprEvalSelector` for it. I think two related constant/cache paths still need the same treatment, though. `Expr.asBitmapColumnIndex` still treats any expression with no required bindings as a constant bitmap index, so an expression filter around `now()` can still be evaluated once while building the bitmap. Also, `ExpressionDimFilter` and `ExpressionVirtualColumn` cache keys still serialize the stable expression cache key (`now()` stringifies the same way every time), so cached query results can be reused across wall-clock changes. Could you either make non-deterministic expressions non-cacheable for these query components, or include a non-stable cache key / disable the affected optimizations when `analysis.isNonDeterministic()` is true?

FrankChen021 (commented):

---

processing/src/test/java/org/apache/druid/math/expr/ExprTest.java
```diff
@@ -221,6 +221,27 @@ public void testEqualsContractForNullDoubleExpr()
...
+  {
+    Expr now = Parser.parse("now()", ExprMacroTable.nil());
+    byte[] k1 = now.getCacheKey();
+    byte[] k2 = now.getCacheKey();
+    Assertions.assertFalse(java.util.Arrays.equals(k1, k2), "bare now() cache key must change across calls");
+
+    // Verify the same instability propagates when now() is nested in a parent expression.
+    Expr nested = Parser.parse("now() > 0", ExprMacroTable.nil());
+    Assertions.assertFalse(
+        java.util.Arrays.equals(nested.getCacheKey(), nested.getCacheKey()),
```
aho135:
```suggestion
        Arrays.equals(nested.getCacheKey(), nested.getCacheKey()),
```
Can be imported

processing/src/main/java/org/apache/druid/segment/transform/ExpressionTransform.java
```diff
@@ -57,6 +60,13 @@ public ExpressionTransform(
     this.parsedExpression = Suppliers.memoize(
         () -> Parser.parse(expression, Preconditions.checkNotNull(this.macroTable, "macroTable"))
     )::get;
+
+    if (ColumnHolder.TIME_COLUMN_NAME.equals(name)
+        && BuiltInExprMacros.NowExprMacro.containsNow(parsedExpression.get())) {
```
aho135: Would it make sense to just check isNonDeterministic() instead of containsNow?

processing/src/test/java/org/apache/druid/math/expr/ExprTest.java
```diff
@@ -221,6 +221,27 @@ public void testEqualsContractForNullDoubleExpr()
                   .verify();
   }
 
+  /**
+   * now() depends on wall-clock time, so its cache key must not be stable across calls — otherwise queries containing
+   * now() can hit a stale cached result. NowExpression overrides {@link Expr#decorateCacheKeyBuilder} to mix in
+   * nanoTime, which makes any containing Expr produce a fresh key on every {@link Expr#getCacheKey()} call.
```
aho135: Stale comment, cacheKey is monotonically increasing long

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
```diff
@@ -427,16 +427,15 @@ private boolean isDirectAccess(final ColumnInspector inspector)
 
   private Supplier<byte[]> makeCacheKeySupplier()
   {
-    return Suppliers.memoize(() -> {
+    return () -> {
```
aho135: Should memoize be conditional depending on whether the expression isNonDeterministic?

processing/src/main/java/org/apache/druid/segment/transform/ExpressionTransform.java
abhishekrb19: Yeah, good point, updated. I think we can make this be the case for all non-deterministic expressions (`now()` being the only one currently)

processing/src/test/java/org/apache/druid/math/expr/ExprTest.java
abhishekrb19: Removed this test doc altogether

processing/src/main/java/org/apache/druid/segment/virtual/ExpressionVirtualColumn.java
abhishekrb19: It shouldn't be needed bc non-deterministic expressions like `NowExpr.decorateCacheKeyBuilder()` appends an incrementing nonce, which makes the cache key unique every time it's built.

aho135: I see that this was reverted in the latest commit. Was that intentional?

abhishekrb19: That's right, it was reverted to the version in master to align with `decorateCacheKeyBuilder()`

aho135 (approved):
LGTM, thanks @abhishekrb19!

PR #19389, opened 2026-04-28
fix: Fix bug in reindexCascade timeline generation where too many search intervals are skipped

capistrant:
### Description

The original implementation is not incorrect in its application of skip offsets, but it is overly conservative. The way the original code works is that it:
1. generates granularity aligned search intervals for all existing rules from the rule provider
2. iterates these search intervals
    a. If the interval end falls after the calculated skip offset threshold (if skipOffsetFromNow or skipOffsetFromLatest is set), do not create jobs for the search interval

2a. is overly conservative. It is completely valid to take such an interval and truncate its end to `granularity.bucketStart(skipOffsetThreshold)` where granularity is the segment granularity of the search interval, and skipOffsetThreshold is the calculated threshold that we must skip creating compaction jobs for any data after. If the interval created by the original start and new end is valid, we can create compaction jobs for it.

#### Release note

Be more precise in applying skipOffsetFromNow and skipOffsetFromLatest when using reindexCascade compaction supervisor. Previously more search intervals than necessary could be skipped when generating compaction jobs if you were using one of the two skip offset styles.

<hr>

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/compact/CascadingReindexingTemplate.java
@@ -354,7 +354,8 @@ public class CascadingReindexingTemplate implements CompactionJobTemplate, DataS
       return Collections.emptyList();
     }
 
-    for (IntervalPartitioningInfo intervalInfo : searchIntervals) {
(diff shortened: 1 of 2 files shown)


FrankChen021 (commented):

---

gianm (approved):

PR #19393, opened 2026-04-30
fix: fix prolonged query timeout bug in SetAndVerifyContextQueryRunner

jtuglu1:
### Description

There's a bug in the `SetAndVerifyContextQueryRunner` where it unconditionally overwrites `QUERY_FAIL_TIME` with `startTimeMillis` + timeout, where `startTimeMillis` is when that runner instance was created — not when the original query started.

This can happen when a SQL query generates inner sub-queries first (e.g. CTEs), then outer native queries after those complete. The outer queries re-enter the broker's query pipeline and get a fresh `SetAndVerifyContextQueryRunner` created some ms offset into execution. For example, with timeout=300000ms inherited from context and a 49000ms offset, their deadline becomes now + 300s = original_start + 349s.

This can cause queries to run past their allotted query timeout, hogging resources and breaking operator/user assumptions. 

#### Release note

Fix prolonged query timeout bug in SetAndVerifyContextQueryRunner that can occur during generated sub-queries.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/SetAndVerifyContextQueryRunner.java
+++ b/server/src/main/java/org/apache/druid/server/SetAndVerifyContextQueryRunner.java
@@ -19,7 +19,6 @@
 
 package org.apache.druid.server;
 
-import com.google.common.collect.ImmutableMap;
 import org.apache.druid.client.DirectDruidClient;
 import org.apache.druid.java.util.common.guava.Sequence;
 import org.apache.druid.query.Queries;
@@ -29,6 +28,8 @@ import org.apache.druid.query.QueryPlus;
 import org.apache.druid.query.QueryRunner;
 import org.apache.druid.query.context.ResponseContext;
 import org.apache.druid.server.initialization.ServerConfig;
+
+import java.util.Map;
 
 /**
  * Use this QueryRunner to set and verify Query contexts.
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/druid/server/SetAndVerifyContextQueryRunner.java
```diff
@@ -70,13 +70,19 @@ public Query<T> withTimeoutAndMaxScatterGatherBytes(Query<T> query, ServerConfig
...
-    final long failTime;
     final QueryContext context = newQuery.context();
+    final long existingFailTime = context.getLong(DirectDruidClient.QUERY_FAIL_TIME, -1L);
+    final long computedFailTime;
     if (context.hasTimeout()) {
-      failTime = this.startTimeMillis + context.getTimeout();
+      computedFailTime = this.startTimeMillis + context.getTimeout();
     } else {
-      failTime = this.startTimeMillis + serverConfig.getMaxQueryTimeout();
+      computedFailTime = this.startTimeMillis + serverConfig.getMaxQueryTimeout();
```
maytasm, line 80: nit: this can overflow to negative (for example when serverConfig.getMaxQueryTimeout() is default value of Long.MAX_VALUE). This would then resulted in Math.min returning the overflow computedFailTime instead of existingFailTime. Not sure if this would matter since existingFailTime would already be Long.MAX_VALUE (has context.hasTimeout() is false). Might be worth adding a guard

maytasm (approved):

server/src/main/java/org/apache/druid/server/SetAndVerifyContextQueryRunner.java
```diff
@@ -70,13 +70,19 @@ public Query<T> withTimeoutAndMaxScatterGatherBytes(Query<T> query, ServerConfig
...
     }
+    // Never extend an existing deadline. This prevents nested native queries (e.g. those generated by the SQL
+    // planner mid-execution) and queued historicals from pushing the effective timeout past the original deadline.
+    final long failTime = (existingFailTime > 0)
```
FrankChen021: [P2] Saturate computed deadline before min

If context.getTimeout() or serverConfig.getMaxQueryTimeout() is Long.MAX_VALUE, startTimeMillis + timeout overflows negative; Math.min(existingFailTime, computedFailTime) then selects that negative value and the final write normalizes it to Long.MAX_VALUE, extending a finite existing deadline. Normalize or saturate the computed deadline before comparing with the existing QUERY_FAIL_TIME.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/server/SetAndVerifyContextQueryRunner.java
jtuglu1: Yep – added an overflow check + test.

jtuglu1:
There is another potential bug I have noticed with inconsistent `QUERY_FAIL_TIME=-1` handling between `JsonParserIterator` and `DirectDruidClient`. The latter treats `-1` as timed out [here](https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/client/DirectDruidClient.java#L429) whereas the former treats `-1` as [no timeout](https://github.com/apache/druid/blob/master/server/src/main/java/org/apache/druid/client/JsonParserIterator.java#L86). I will fix this inconsistency in a separate PR. cc @clintropolis @gianm not sure if there's a historical reason for this.

jtuglu1:
Might be worth considering deprecating this `QUERY_FAIL_TIME` field and moving to a "millis left" relative notation – one that will be less prone to overflow and other things like clock skew between nodes.

clintropolis (approved):

PR #19394, opened 2026-04-30
perf: Streamline s3 backend

jtuglu1:
### Description

S3 achieve strong read-after-write consistency in [2020](https://aws.amazon.com/about-aws/whats-new/2020/12/amazon-s3-now-delivers-strong-read-after-write-consistency-automatically-for-all-applications/). The current s3 backend architecture assumes a prior consistency model and therefore does some redundant calls which are both slow and costly.

Some other things to look into in a separate PR is parallel download requests using byte ranges for a single file (currently we use a single TCP connection which isn't the fastest and is subject to S3 single-connection bandwidth limitations).

#### High-level list of changes

1. Removed `isObjectInBucket` guards before zip and gzip downloads (and the now-dead private method). The 404 from GetObject propagates as a `SegmentLoadingException`.
2. Replaced the `doesObjectExist` + `listObjectsV2` two-call sequence with a single `getObjectMetadata` request in `S3DataSegmentMover`.

#### Release note

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentMover.java
+++ b/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentMover.java
@@ -24,8 +24,6 @@ import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.inject.Inject;
-import org.apache.druid.java.util.common.IOE;
-import org.apache.druid.java.util.common.ISE;
 import org.apache.druid.java.util.common.MapUtils;
 import org.apache.druid.java.util.common.RetryUtils;
 import org.apache.druid.java.util.common.StringUtils;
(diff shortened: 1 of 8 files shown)


extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentMover.java
```diff
@@ -185,90 +183,75 @@ private void safeMove(
...
-            .destinationKey(targetS3Path);
-        if (!config.getDisableAcl()) {
-          final String headerValue = S3Utils.grantFullControlHeaderValue(
-              S3Utils.grantFullControlToBucketOwner(s3Client, targetS3Bucket)
+    final HeadObjectResponse sourceMetadata;
+    try {
+      sourceMetadata = s3Client.getObjectMetadata(s3Bucket, s3Path);
+    }
+    catch (S3Exception e) {
+      if (e.statusCode() == 404 && "NoSuchKey".equals(S3Utils.getS3ErrorCode(e))) {
```
FrankChen021: [P2] Treat any HEAD 404 as a missing source

The idempotent already-moved path now only runs when headObject fails with status 404 and error code NoSuchKey. S3 HEAD failures for absent objects can surface as a generic 404/NotFound response, and other Druid S3 code handles missing HEAD results by status alone. In that case this code rethrows before checking whether the target object already exists, so a retry or competing move where the source was deleted after a successful copy can incorrectly fail instead of returning the moved segment. Please key this fallback off statusCode() == 404 rather than requiring NoSuchKey.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

FrankChen021 (commented):
I have reviewed the updated code for correctness, edge cases, concurrency, lifecycle, and integration risks; no issues found.

---

gianm (approved):

PR #19397, opened 2026-05-01
feat: add async CursorFactory API and migrate MSQ frame processors to use it

clintropolis:
### Description
This PR adds an async variant of `CursorFactory.makeCursorHolder` which returns a new `AsyncCursorHolder` type that provides lifecycle management of the async load process and migrates MSQ frame processors to use it, so that future cursor factories backed by partial / lazy-loaded storage can perform I/O without blocking the processing thread. This PR introduces no partial-loading behavior on its own, instead it just establishes the integration shape for more upcoming partial-segment work that will follow this PR. Every existing cursor factory remains synchronous via the default implementation which just calls `AsyncCursorHolder.completed(makeCursorHolder(spec))`.

Worth noting, I am planning some changes in a separate PR for V10 segments so that a partial segment can provide a `TimeBoundaryInspector` for usage by `GroupByPreShuffleFrameProcessor` without needing to download any column data so we can avoid making an async variant of it.

changes:
* add `CursorFactory.makeCursorHolderAsync(CursorBuildSpec)` for cursor factories backed by partial downloads can do I/O without blocking processing threads, with a default implementation returning `AsyncCursorHolder.completed(makeCursorHolder(spec))` so existing implementations remain async-correct without changes
* add `GroupingEngine.makeCursorHolderAsync` returning `AsyncCursorHolder`, and extracting shared `processWithCursorHolder` helper from `GroupingEngine.process()`, so that a caller which can yield and then resume can wait for the `CursorHolder` to be ready and later process it
* migrate `ScanQueryFrameProcessor.runWithSegment` to call `makeCursorHolderAsync` and yield via `ReturnOrAwait` while the load is pending
* migrate `GroupByPreShuffleFrameProcessor.runWithSegment` cursor path to call `GroupingEngine.makeCursorHolderAsync` and yield via `ReturnOrAwait` while loading


processing/src/main/java/org/apache/druid/query/groupby/GroupingEngine.java
```diff
@@ -482,8 +484,46 @@ public Sequence<ResultRow> process(
...
+    final CursorBuildSpec buildSpec = makeCursorBuildSpec(query, groupByQueryMetrics);
+    final CursorHolder cursorHolder = cursorFactory.makeCursorHolder(buildSpec);
+    return processWithCursorHolder(query, cursorFactory, cursorHolder, timeBoundaryInspector, bufferPool, buildSpec);
+  }
+
+  /**
+   * Asynchronous variant of {@link #process} that obtains the {@link CursorHolder} from
+   * {@link CursorFactory#makeCursorHolderAsync} so callers running on threads that must not block on I/O
+   * (e.g. MSQ frame processors) can yield via {@link org.apache.druid.frame.processor.ReturnOrAwait#awaitAllFutures}
+   * until the cursor holder is ready.
```
gianm: This javadoc feels a bit too specific to me. It's enough to say that this is an asynchronous variant of `process` that uses `CursorFactory#makeCursorHolderAsync` to avoid blocking on acquisition of `CursorHolder`.

processing/src/main/java/org/apache/druid/query/groupby/GroupingEngine.java
```diff
@@ -482,8 +484,46 @@ public Sequence<ResultRow> process(
...
+      @Nullable TimeBoundaryInspector timeBoundaryInspector,
+      NonBlockingPool<ByteBuffer> bufferPool,
+      @Nullable GroupByQueryMetrics groupByQueryMetrics
+  )
+  {
+    validateForProcess(query, cursorFactory);
+    final CursorBuildSpec buildSpec = makeCursorBuildSpec(query, groupByQueryMetrics);
+    return FutureUtils.transform(
+        cursorFactory.makeCursorHolderAsync(buildSpec),
+        cursorHolder -> processWithCursorHolder(
```
gianm: Won't this end up doing the main processing in whatever thread happened to resolve the `makeCursorHolderAsync` future? (Because `FutureUtils.transform` uses a direct executor.) Possibly that'd be in a virtual storage loader thread.

I think we'll need to either adjust this method to accept a `ListenableExecutorService` that will be used to run `processWithCursorHolder`, or break it up so callers first call `groupingEngine.makeCursorHolderAsync` and then call `groupingEngine.processCursorHolder`.

processing/src/main/java/org/apache/druid/query/groupby/GroupingEngine.java
```diff
@@ -494,21 +534,38 @@ public Sequence<ResultRow> process(
...
+  )
+  {
+    final GroupByQueryConfig querySpecificConfig = configSupplier.get().withOverrides(query);
+
+    final ResourceHolder<ByteBuffer> bufferHolder;
+    try {
+      bufferHolder = bufferPool.take();
+    }
+    catch (Throwable e) {
+      CloseableUtils.closeAndWrapExceptions(cursorHolder);
```
gianm: `throw CloseableUtils.closeAndWrapInCatch(e, cursorHolder)` will properly retain exceptions from closing `cursorHolder` as suppressed exceptions on `e`.

Although, there's probably some way of structuring this code to use the `Closer` to handle this better. Like, create the `Closer` first, register `cursorHolder` right after it's created, start the main `try`, then register `bufferHolder` after it's acquired. If the buffer fails to be acquired then the `catch` will close the `Closer` and release the `cursorHolder`.

multi-stage-query/src/main/java/org/apache/druid/msq/querykit/scan/ScanQueryFrameProcessor.java
```diff
@@ -297,21 +304,29 @@ protected ReturnOrAwait<SegmentsInputSlice> runWithDataServerQuery(final DataSer
...
+      if (cursorHolderFuture == null) {
+        final Segment segment = mapSegment(segmentHolder, closer);
+        final CursorFactory cursorFactory = segment.as(CursorFactory.class);
+        if (cursorFactory == null) {
+          throw DruidException.defensive(
+              "Null cursor factory found. Probably trying to issue a query against a segment being memory unmapped."
+          );
+        }
+
+        cursorHolderFuture = cursorFactory.makeCursorHolderAsync(
```
gianm: Handling futures that return closeable things is tricky. Maybe we can improve it by changing the return of `makeCursorHolderAsync` from `ListenableFuture<CursorHolder>` to `AsyncCursorHolder` that is closeable and has methods `get()` (blocks if not ready), `close()` (closes the resource no matter where it is in its lifecycle), and `addReadyCallback(Runnable)` (used by nonblocking callers to learn when `get` is ready).

The problem with the future approach is that once this call site gets the `cursorHolderFuture`, it's responsible for monitoring the future and closing `cursorHolder` if the future resolves successfully. This has to be done even if the processor is canceled before it has a chance to run through normally. It requires extra carefulness and is easy to mess up.

One way it can be handled is by attaching a callback in `cleanup` that closes the holder in `onSuccess`, like:

```
if (cursorHolderFuture != null) {
  Futures.addCallback(
    cursorHolderFuture,
    new FutureCallback<>() {
      void onSuccess(CursorHolder holder) { holder.close(); }
      void onFailure(Throwable t) { /* nothing */ }
    }
  );
}
```

But even with this structure, it's important watch out for pitfalls. A big one is that you can never cancel a future that returns a closeable thing. Cancellation of the future can cause the object to be orphaned and eventually GCed without being closed (if the object is created before cancellation has a chance to interrupt whatever was creating it).

If we can avoid these problems by returning something *directly* closeable (like this `AsyncCursorHolder` idea) rather than future-of-closeable, then the caller code becomes simpler.

processing/src/main/java/org/apache/druid/query/groupby/GroupingEngine.java
```diff
@@ -494,21 +534,38 @@ public Sequence<ResultRow> process(
...
+  private Sequence<ResultRow> processWithCursorHolder(
+      GroupByQuery query,
+      CursorFactory cursorFactory,
+      CursorHolder cursorHolder,
+      @Nullable TimeBoundaryInspector timeBoundaryInspector,
+      NonBlockingPool<ByteBuffer> bufferPool,
+      CursorBuildSpec buildSpec
+  )
+  {
+    final GroupByQueryConfig querySpecificConfig = configSupplier.get().withOverrides(query);
```
FrankChen021: P2 Close cursorHolder if query config override throws

processWithCursorHolder now receives an already-created CursorHolder, but computes querySpecificConfig before entering any cleanup-protected block. If withOverrides throws, for example due to an invalid groupBy query context value, the CursorHolder returned by makeCursorHolderAsync/makeCursorHolder is never closed. Move this config resolution before acquiring the holder, or wrap it so cursorHolder is closed on every pre-sequence failure.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

FrankChen021 (commented):

---

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,194 @@
...
+import java.io.Closeable;
+
+/**
+ * Closeable wrapper around an asynchronously-loaded {@link CursorHolder}, returned by
+ * {@link CursorFactory#makeCursorHolderAsync}. Designed to make lifecycle management safe even when the holder is
+ * still loading: callers receive a single Closeable handle and can register it once with their cleanup machinery,
+ * regardless of where the underlying load is in its lifecycle.
+ * <p>
+ * The hazard this exists to avoid: returning a {@code ListenableFuture<CursorHolder>} (or similar future-of-Closeable)
+ * makes correct cleanup error-prone, where cancelling the future or letting a caller fail before consuming the future
```
gianm: canceling (spelling)

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,194 @@
...
+ * can orphan the produced holder, leaking the underlying resources. By exposing a Closeable that internally tracks the
+ * load and disposes whatever has materialized, callers don't have to write that bookkeeping themselves.
+ * <p>
+ * Typical usage from a non-blocking caller (e.g. an MSQ frame processor):
```
gianm: No reason to include `(e.g. an MSQ frame processor)` here. I would also edit the code sample to be less MSQ specific.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,194 @@
...
+   * Already-ready {@link AsyncCursorHolder} backed by an immediately-available {@link CursorHolder}. The default
+   * synchronous {@link CursorFactory#makeCursorHolderAsync} implementation uses this.
+   */
+  public static AsyncCursorHolder completed(CursorHolder holder)
+  {
+    return new AsyncCursorHolder(Futures.immediateFuture(holder));
+  }
+
+  /**
+   * Wraps a {@link ListenableFuture}{@code <CursorHolder>}. Implementations doing real I/O should produce the future
```
gianm: The `@link` and `@code` seem weird here. Also the "Implementations doing real I/O…" bit doesn't seem necessary.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,194 @@
...
+   * (e.g. by submitting to an executor) and pass it here.
+   */
+  public static AsyncCursorHolder fromFuture(ListenableFuture<CursorHolder> future)
+  {
+    return new AsyncCursorHolder(future);
+  }
+
+  private final ListenableFuture<CursorHolder> future;
+  private boolean closed = false;
```
gianm: use `@GuardedBy` on `closed` and `disposed`, since synchronization is crucial given how they're used.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * Close the holder. Safe at any lifecycle point and idempotent:
+   * <ul>
+   *   <li>Already-loaded: closes the underlying {@link CursorHolder} immediately.</li>
+   *   <li>Loading in progress: marks for disposal; the produced holder is closed when the load completes.</li>
+   *   <li>Load failed: no-op (nothing to dispose).</li>
+   *   <li>Already closed: no-op.</li>
+   * </ul>
+   * Note that closing does NOT cancel an in-flight load — cancelling a future-of-Closeable is the exact lifecycle
+   * hazard this class exists to prevent. The load completes normally and the resulting holder is closed promptly.
```
gianm: Really we should structure this to be capable of canceling the in-flight load. I think to make it work we'll want to not base this class on futures at all. Something like:

- Accept a `Runnable canceler` in the constructor
- Replace the `future` field with `Either<Throwable, T>`
- Expose `set` and `setException` methods that return `boolean` (true for set accepted, false for set not accepted)
- In `close`, call `canceler` if the set methods haven't been called yet
- Whatever provides the actual `CursorHolder` should close it if `set` returns false

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,194 @@
...
+      return;
+    }
+    if (future.isDone() && !future.isCancelled()) {
+      try {
+        final CursorHolder holder = FutureUtils.getUncheckedImmediately(future);
+        disposed = true;
+        holder.close();
+      }
+      catch (Throwable t) {
+        // Future completed with a failure, no holder to close.
```
gianm: Comment isn't entirely right. The `holder.close()` might also have failed. I don't think we want to suppress that.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,194 @@
...
+    // If close() lands before the future completes we still need to dispose the eventually-produced holder. Register
+    // a callback up-front so the holder is closed whenever it arrives if we've been closed in the meantime.
+    Futures.addCallback(
+        future,
+        new FutureCallback<>()
+        {
+          @Override
+          public void onSuccess(CursorHolder holder)
+          {
+            synchronized (AsyncCursorHolder.this) {
```
gianm: We can avoid having two different `holder.close()` places by attaching the callback in `close` itself. Although, if we avoid futures entirely as suggested in another comment, this comment will become irrelevant.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,194 @@
...
+  {
+    return future.isDone();
+  }
+
+  /**
+   * Take ownership of the underlying {@link CursorHolder}. After this returns, {@link #close()} on this
+   * {@code AsyncCursorHolder} is a no-op; the caller is responsible for closing the returned holder. Useful when
+   * passing the holder to another component (e.g. a cursor-lifecycle manager) that takes ownership of it.
+   * <p>
+   * Throws {@link IllegalStateException} if the holder is not yet ready, has already been released, or this wrapper
```
gianm: `DruidException.defensive` seems more standard.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
clintropolis: This is nicer, though I ended up not using `Either` for the field since it didn't save me much and was kind of awkward since internally we are always dealing with one or the other (i did use it to make it easier to share a set method internally though)

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
clintropolis: The example code got a bit weird trying to get rid of `ReturnOrAwait`, so instead of tried to frame it that the example is using 'ReturnOrAwait' to show how you're supposed to use this thing, since the important part is the yield-then-resume when ready pattern, which i think makes it feel less like this is a thing for MSQ than the last iteration did?

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  private static void fireCallbacks(List<Runnable> callbacks)
+  {
+    for (Runnable cb : callbacks) {
+      try {
+        cb.run();
+      }
+      catch (Throwable ignored) {
+        // Best-effort; one bad callback shouldn't break others.
```
gianm, line 309: `log.warn` something, otherwise a callback failure will be completely silent, and this could make debugging something difficult down the road. Especially since the likely outcome of an readiness callback failing would be that things just get stalled out.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,298 @@
...
+  {
+    if (closed) {
+      throw DruidException.defensive("AsyncCursorHolder is already closed");
+    }
+    if (disposed) {
+      throw DruidException.defensive("AsyncCursorHolder has already been released");
+    }
+    if (error != null) {
+      // pass through as is
+      if (error instanceof RuntimeException runtime) {
```
gianm, line 193: or `Error`?

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,298 @@
...
+
+  /**
+   * Allows producer to mark the load successful with the given holder. Returns {@code true} if accepted, {@code false}
+   * if this wrapper has already been closed, in which case the producer is responsible for closing {@link CursorHolder}
+   * itself. Throws {@link DruidException} if the load was already completed (from prior calls to this method or
+   * {@link #setException}).
+   * <p>
+   * Callbacks registered via {@link #addReadyCallback} fire outside the lock to avoid re-entrancy deadlocks.
+   */
+  public boolean set(CursorHolder holder)
```
gianm, line 123: This method should reject null `holder` with a defensive exception.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,298 @@
...
+    this.canceler = canceler;
+  }
+
+  /**
+   * Allows producer to mark the load successful with the given holder. Returns {@code true} if accepted, {@code false}
+   * if this wrapper has already been closed, in which case the producer is responsible for closing {@link CursorHolder}
+   * itself. Throws {@link DruidException} if the load was already completed (from prior calls to this method or
+   * {@link #setException}).
+   * <p>
+   * Callbacks registered via {@link #addReadyCallback} fire outside the lock to avoid re-entrancy deadlocks.
```
gianm: What does this comment mean? Seems like there's only one lock (`this`) and it should be fine to re-enter it.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,298 @@
...
+        throw runtime;
+      }
+      throw DruidException.forPersona(DruidException.Persona.DEVELOPER)
+                          .ofCategory(DruidException.Category.UNCATEGORIZED)
+                          .build(error, error.getMessage());
+    }
+    if (result == null) {
+      throw DruidException.defensive("AsyncCursorHolder is not ready yet");
+    }
+    disposed = true;
```
gianm, line 208: set `result` to `null` here to allow GC? The AsyncCursorHolder can live longer than the reference from `release()` if the AsyncCursorHolder is added to a high-level `Closer` or `withBaggage` or something like that.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,298 @@
...
+   * </ul>
+   */
+  @Override
+  public void close()
+  {
+    final CursorHolder holderToClose;
+    final Runnable cancelerToRun;
+    synchronized (this) {
+      if (closed) {
+        return;
```
gianm: Instead of idempotent close, how about throwing a defensive error here? Double-close can be indicative of resource management problems, and an exception would help us identify those potential issues.

If you do this then the javadoc should be updated too.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
```diff
@@ -0,0 +1,298 @@
...
+      }
+      closed = true;
+      if (disposed) {
+        // Ownership was already transferred via release(); the caller manages the holder lifecycle.
+        return;
+      }
+      if (result != null) {
+        // Result is here and no one has released it; we close it.
```
gianm, line 261: I think this logic would end up being simpler if you had nulled out `result` in `release`. It would then just be: always close `result` if it is nonnull, because if it's nonnull, we own it.

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
clintropolis: oh, i guess bad wording; like its more like avoiding problems like if the callback needs to hold some other lock we don't have any weird ordering issues like if that lock is some lock shared between callback thread and producing thread, will try to clarify

processing/src/main/java/org/apache/druid/segment/AsyncCursorHolder.java
clintropolis: I added the nulling out of `result` in `release()`, however i still have the `disposed` flag because otherwise calling release would allow something to be able to call `set`/`setException` again which seems unchill, so I've kept this short-circuit on `disposed` here as is and haven't really changed this part.

We do essentially always call close on the result if we get to this part where we check not null, stuff maybe looks a bit more funny because we actually call close and the canceler outside of the synchronized block?

gianm (approved):

PR #19398, opened 2026-05-02
feat: persist __time min/max in V10 ProjectionMetadata

clintropolis:
### Description

changes:
* add nullable `minTime`/`maxTime` Long fields to `ProjectionMetadata` alongside the existing `numRows`
* track min/max __time in `IndexMergerBase.mergeIndexesAndWriteColumns` via `timestampSelector.getLong()` on each row; surface via `IndexMergeResult` (null when zero rows are walked)
* `IndexMergerV10` wires the merge result's values into the base-table `ProjectionMetadata`; aggregate projections leave the new fields null for now, though we can easily add this in future work if useful
* tracking is done regardless of segment sort order, non-time-sorted segments (`DimensionsSpec.forceSegmentSortByTime = false`) also store these values accurately (just with less utility since we can't use the information to skip rows, though we could use them to know if there are no rows within the time range)
* write-side only: fields are persisted but not yet consumed. A follow-up will add a partial V10-aware `TimeBoundaryInspector` that reads them

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/projections/ProjectionMetadata.java
+++ b/processing/src/main/java/org/apache/druid/segment/projections/ProjectionMetadata.java
@@ -20,11 +20,13 @@
 package org.apache.druid.segment.projections;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.Interner;
 import com.google.common.collect.Interners;
 import org.apache.druid.segment.Metadata;
 
+import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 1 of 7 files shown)


FrankChen021 (commented):

---

FrankChen021 (commented):

---

processing/src/main/java/org/apache/druid/segment/projections/ProjectionMetadata.java
```diff
@@ -35,28 +37,60 @@ public class ProjectionMetadata
...
       schema = TableProjectionSchema.fromMetadata(dims, metadata);
     }
-    return new ProjectionMetadata(numRows, schema);
+    return new ProjectionMetadata(numRows, schema, minTime, maxTime);
   }
 
   private final int numRows;
   private final ProjectionSchema schema;
+  /**
+   * Minimum {@code __time} value across all rows in this projection, or {@code null} if the writer didn't supply one
```
gianm, line 60: Comment seems overly verbose.

gianm (approved):

PR #19404, opened 2026-05-04
perf: Use Set instead of List for druid.request.logging.mutedQueryTypes

abhishekrb19:
Update `druid.request.logging.mutedQueryTypes` to use a `Set` instead of a `List`.
Set lookups are faster than list-based lookups, which can matter at high QPS when filtered logging with multiple muted types is enabled on servers.

This PR has:

- [x] been self-reviewed.


GWphua (approved):
LGTM

PR #19405, opened 2026-05-04
fix:  Fix unhealthy supervisor when a priority-assigned task dies before discovery

abhishekrb19:
### Description

  With https://github.com/apache/druid/pull/19040, when `ioConfig.serverPriorityToReplicas` is set, a supervisor can get stuck throwing
  `DruidException` from `computeUnassignedServerPriorities`, preventing any new
  task replicas with destined replicas from being created until the old tasks rollover. The 
exception is as follows:
  ```
Found unassignedServerPriorities[[]] of size[0] < total replicas[1] for taskGroupId[0].
  Task server priorities[[1, 0]] have already been assigned to tasks[[foo]].
```

The supervisor can remain in this unhealthy state unable to create additional task replicas until tasks eventually rollover. This can happen when a task spuriously fails after it's created but before it's discovered in the `runInternal()` loop.

  ### Root cause
  `createTasksForGroup` was an *additional* writer of `group.taskIdToServerPriority`,
  recording the priority at task-submission time — before the task was ever observed
  in `group.tasks`. The group's other invariant (`tasks` and `taskIdToServerPriority`
  stay in sync) is enforced by `discoverTasks` (additions) and `TaskGroup.removeTask`
  (removals), both of which operate on `group.tasks`.

  If a task died after submission but before the next supervisor run picked it up from
  the overlord, it never entered `group.tasks`, so `removeTask` never fired — the
  priority entry was orphaned and made the next `computeUnassignedServerPriorities`
  call throw.

 ### Fix

  This patch removes the additional writer, leaving `discoverTasks` and `removeTask`
  as the sole mutators. `group.taskIdToServerPriority` should now stay in sync with `group.tasks`.
  The added unit tests fail on master without the fix.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
+++ b/extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorTest.java
@@ -1727,24 +1727,12 @@ public class KafkaSupervisorTest extends EasyMockSupport
   }
 
   /**
-   * Regression test for a bug where {@code taskIdToServerPriority} could retain a stale entry for a task that
-   * was submitted in run 1 but died before the next supervisor run, causing
-   * {@link SeekableStreamSupervisor#computeUnassignedServerPriorities} to see a fully-assigned priority set and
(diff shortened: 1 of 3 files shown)


aho135 (approved):
Thanks for the quick fix on this one @abhishekrb19!

PR #19409, opened 2026-05-05
feat: Wire format and data model for partial-projection load rules

clintropolis:
### Description
Follow-up to #19374, this PR lays the wire-format groundwork for partial load rules. Coordinators can now send wrapped load specs to historicals carrying scheme-specific partial-load metadata, historicals can advertise whether they understand the wrapper family, and a generic `PartialLoadProfile` data model threads the metadata through the load queue. The actual coordinator assignment logic (fingerprint-aware replica counting, load-queue plumbing, inventory tracking, and the historical-side partial-load path) is intentionally deferred to follow-up PRs.

changes:
* add `PartialProjectionLoadSpec`, a `LoadSpec` wrapper for partial-projection loads with lazy delegate materialization
* add `PartialLoadSpecModule` and register it in `CoreInjectorBuilder` so historicals always have the wrapper
* add `supportsPartialLoad` capability flag to `SegmentLoadingCapabilities`; historicals (will) advertise true, coordinator defaults to false for unknown servers
* add optional `fingerprint` and `loadedBytes` wire fields to `SegmentChangeRequestLoad` for historical partial-load announcements (NON_NULL include)
* add `PartialLoadProfile` generic data model (wrappedLoadSpec, fingerprint, loadedBytes) with `forRequest` / `forLoaded` / `forFullFallback` factories
* add `SegmentActionHandler.replicateSegmentPartially` default throwing `UnsupportedOperationException`
* wire `PartialLoadRule.run()` to route through `replicateSegmentPartially` when the matcher resolves; fall back to `replicateSegment` for FULL_LOAD-on-cannot-match


server/src/main/java/org/apache/druid/server/coordinator/rules/PartialLoadRule.java
```diff
@@ -89,9 +90,21 @@ public boolean appliesTo(DataSegment segment, DateTime referenceTimestamp)
...
   public void run(DataSegment segment, SegmentActionHandler handler)
   {
-    // Partial plumbing is added in future work. For now, a partial rule that applies to a segment full-loads it,
-    // identical behavior to the corresponding non-partial rule
-    handler.replicateSegment(segment, getTieredReplicants());
+    final PartialLoadMatcher.MatchResult result = matcher.match(segment, segment.getLoadSpec());
+    if (result != null) {
+      // Matcher resolved: route through the partial-load handler. The wrappedLoadSpec map carries scheme-specific
+      // data that the historical-side wrapper deserializes.
+      handler.replicateSegmentPartially(
```
FrankChen021, line 97: [P1] Matched partial rules now call an unimplemented production handler

This changes matched partial-load rules from the prior full-load behavior to `handler.replicateSegmentPartially(...)`, but the production `StrategicSegmentAssigner` does not override that method; the only implementation is the new `SegmentActionHandler` default, which throws `UnsupportedOperationException`. Since `RunRules` passes a `StrategicSegmentAssigner` directly and does not catch this per rule, any configured partial-load rule whose matcher resolves for a segment can abort the coordinator rules duty instead of loading or falling back. Either keep the previous full-load path until assignment is implemented, or add a production override that degrades safely.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/server/coordinator/rules/PartialLoadRule.java
clintropolis: oops, I was trying to keep things so that partial load rule stuff works as I do these PRs, but missed this. I think its probably ok since I have a follow-up to this PR which fills out the rest of the coordinator side stuff and will fix this problem, but I suppose I can fix it up here if anyone feels strongly about it. Its probably low risk because this stuff is all new.

capistrant: not ideal but since I don't have other comments requiring fixup and you plan to fast follow, I think this is ok

capistrant (approved):

server/src/main/java/org/apache/druid/server/coordinator/rules/PartialLoadRule.java
FrankChen021: Acknowledged. Since this remains guarded by the still-in-progress partial-load path and there is a planned fast follow-up for the production coordinator handler, I am fine leaving this as a known interim gap for this PR.

clintropolis: I adjusted things to set `supportsPartialLoad` to false until all of this stuff is in place and working, not sure why I wasn't doing that previously 😅. It doesn't really change much in this PR since stuff isn't fully wired yet, but its still the right move to make until everything is actually working.

PR #19410, opened 2026-05-05
refactor: add ColumnFormat implementation for string columns

jaykanakiya:
### Description

Addresses a follow up from https://github.com/apache/druid/pull/19258

Adds StringDictionaryEncodedColumnFormat that persists StringColumnFormatSpec in DictionaryEncodedColumnPartSerde segment metadata.

<hr>

 * `StringDictionaryEncodedColumnFormat`
 * `DictionaryEncodedColumnPartSerde`
 * `StringDimensionIndexer`
 * `StringDimensionHandler`

<hr>

This PR has:

- [x] been self-reviewed.


processing/src/main/java/org/apache/druid/segment/column/StringDictionaryEncodedColumnFormat.java
```diff
@@ -0,0 +1,138 @@
...
+                                 .setHasSpatialIndexes(hasSpatialIndexes);
+  }
+
+  @Override
+  public DimensionHandler getColumnHandler(String columnName)
+  {
+    Integer maxStringLength = columnFormatSpec != null ? columnFormatSpec.getMaxStringLength() : null;
+    return new StringDimensionHandler(
+        columnName,
+        MultiValueHandling.ofDefault(),
```
FrankChen021: [P1] Persisted string format drops multi-value handling when rebuilding handlers

getColumnHandler reconstructs the StringDimensionHandler with MultiValueHandling.ofDefault() even when columnFormatSpec contains an explicit multiValueHandling. IndexMergerBase and IncrementalIndex.loadDimensionIterable call getColumnHandler directly, so compacting or loading persisted string columns with ARRAY or SORTED_SET can silently switch back to the default behavior and reorder or deduplicate multi-value strings. Derive the handler's multiValueHandling from columnFormatSpec when present, matching StringDimensionSchema.getDimensionHandler.

processing/src/main/java/org/apache/druid/segment/column/CapabilitiesBasedFormat.java
```diff
@@ -102,6 +102,10 @@ public ColumnFormat merge(@Nullable ColumnFormat otherFormat)
       return this;
     }
 
+    if (otherFormat instanceof StringDictionaryEncodedColumnFormat) {
```
FrankChen021, line 105: [P1] String format merge bypasses type compatibility checks

This delegation happens before CapabilitiesBasedFormat's existing type checks, and StringDictionaryEncodedColumnFormat.merge accepts any CapabilitiesBasedFormat without validating its logical type. If one segment has a string-specific format for a column and another legacy segment reports the same column as LONG, DOUBLE, COMPLEX, etc., the merge now produces a STRING format instead of rejecting the incompatible schemas. Restrict delegation to compatible string capabilities or perform the same logical/element/complex type validation in StringDictionaryEncodedColumnFormat.merge before merging.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 0 |
| P3 | 0 |
| Total | 2 |

---

processing/src/main/java/org/apache/druid/segment/column/StringDictionaryEncodedColumnFormat.java
jaykanakiya: Thanks for taking a look @FrankChen021. Updated to derive mvh from columnFormatSpec and use default as fallback.

processing/src/main/java/org/apache/druid/segment/column/CapabilitiesBasedFormat.java
jaykanakiya: Added type validation.

FrankChen021 (commented):
I reviewed the follow-up replies and current code. Both prior findings appear addressed, so no inline reply is needed.

---

jaykanakiya:
@FrankChen021 @clintropolis Let me know your thoughts on these changes.

processing/src/main/java/org/apache/druid/segment/StringDimensionIndexer.java
```diff
@@ -77,12 +82,39 @@ public StringDimensionIndexer(
...
+    this.columnFormatSpec = columnFormatSpec;
+  }
+
+  @Override
+  public ColumnFormat getFormat()
+  {
+    if (columnFormatSpec != null) {
+      return new StringDictionaryEncodedColumnFormat(
+          hasMultipleValues,
+          false,
```
gianm: We should be able to provide a proper value for `hasNulls` here, using the same check that `getColumnCapabilities` does: `dimLookup.getIdForNull() != DimensionDictionary.ABSENT_VALUE_ID`.

gianm (commented):
Generally looks good. Please take a look at the comment about `hasNulls` and please add a test for the new `StringDimensionIndexer#getFormat` implementation in StringDimensionIndexerTest.

processing/src/main/java/org/apache/druid/segment/StringDimensionIndexer.java
jaykanakiya: Thanks for taking a look @gianm. I added the null check and test for getFormat.

gianm (approved):
Looks good, thanks!

jaykanakiya:
Thanks for the approval @gianm! I don't think I have merge permissions. Could you help merge this when you get a chance?

gianm:
> Thanks for the approval @gianm! I don't think I have merge permissions. Could you help merge this when you get a chance?

Yes, I'm just waiting for the tests to pass.

PR #19412, opened 2026-05-05
feat: MSQ supervisors using mostFragmentedFirst policy and minor compaction can scale taskCount down for minor compactions

capistrant:
### Description

Minor compactions my not require the same parallelism that a full compaction does for a datasource. This change allows the compaction supervisor to add a taskContext parameter `minorCompactionTaskPercent` that scales `maxTaskCount` downwards for minor compaction tasks. For example

```JSON
...
  "taskContext": {
    "maxTaskCount": 100,
    "minorCompactionTaskPercent": 20
  },
...
```

full compactions will use 100 tasks and minor compactions will use 20.

by default this behavior is disabled and each supervisor who wants to use it must be using the MSQ compaction engine + opt in by modifying `minorCompactionTaskPercent` explicitly.

#### Release note

If you are using MSQ task engine with compaction supervisors for automatic compaction plus MostFragmentedFirst policy with minor compactions enabled, you can now set a per supervisor configuration to use fewer MSQ workers for minor compactions compared to full compactions by setting "minorCompactionTaskPercent" to an int between 1 and 99. minor compactions will use `FLOOR(2, maxTaskCount * minorCompactionTaskPercent)` for minor compaction task parallelism.

<hr>

 * `CompactSegments.java`

<hr>

This PR has:

- [ ] been self-reviewed.


server/src/main/java/org/apache/druid/server/coordinator/duty/CompactSegments.java
```diff
@@ -448,6 +462,52 @@ public Map<String, AutoCompactionSnapshot> getAutoCompactionSnapshot()
...
+      CompactionMode compactionMode,
+      ClientCompactionRunnerInfo compactionRunner
+  )
+  {
+    if (compactionMode != CompactionMode.UNCOMPACTED_SEGMENTS_ONLY
+        || !CompactionEngine.MSQ.equals(compactionRunner.getType())) {
+      return;
+    }
+
+    final int percent = QueryContext.of(context).getInt(
```
FrankChen021, line 488: [P2] Validate minorCompactionTaskPercent when accepting the compaction config

The new context key is only parsed and range-checked while creating a minor MSQ compaction task. Supervisor config validation still accepts values like 0, 200, or a non-numeric string, so the API can persist an invalid supervisor and it will later fail job creation repeatedly once a minor compaction candidate appears. Please add validation alongside the existing MSQ maxNumTasks validation paths, including CascadingReindexingTemplate if applicable, so bad configs are rejected before scheduling.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

cecemei (commented):
i felt having a percent variable is not really necessary, especially we're just trying to multiple two variables to get a task count for minor compaction, this mental math is not really necessary. 

what do you think if we just add another `minNumTasks`?

capistrant:
> i felt having a percent variable is not really necessary, especially we're just trying to multiple two variables to get a task count for minor compaction, this mental math is not really necessary.
> 
> what do you think if we just add another `minNumTasks`?

how would this new variable be used?

I choose a percent so someone can just make an estimate and then not maintain it as they change their max task counts for full compaction, they will just keep the same ratio for minor:full when it comes to task counts

FrankChen021 (commented):

---

server/src/main/java/org/apache/druid/client/indexing/ClientCompactionRunnerInfo.java
```diff
@@ -241,6 +242,37 @@ public static CompactionConfigValidationResult validateMaxNumTasksForMSQ(Map<Str
...
+   * config cannot be persisted and then fail every subsequent minor compaction job.
+   */
+  public static CompactionConfigValidationResult validateMinorCompactionTaskPercentForMSQ(Map<String, Object> context)
+  {
+    if (context == null || !context.containsKey(ClientMSQContext.CTX_MINOR_COMPACTION_TASK_PERCENT)) {
+      return CompactionConfigValidationResult.success();
+    }
+    final int percent;
+    try {
+      percent = QueryContext.of(context).getInt(ClientMSQContext.CTX_MINOR_COMPACTION_TASK_PERCENT, 0);
```
FrankChen021, line 257: [P2] Reject fractional minorCompactionTaskPercent values

This validation uses `QueryContext.getInt`, which accepts any `Number` by calling `intValue()`. JSON values like `50.9` deserialize as `Double` and are silently truncated to `50`, so malformed configs can be persisted despite the new contract requiring an integer between 1 and 100. Please reject non-integral `Number` values, or parse with exact integer conversion, before accepting the config.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

PR #19413, opened 2026-05-05
feat: use DruidNode.version field to support blue/green deployments

jtuglu1:
#### Druid Deployment Overview

Currently, deployment in Druid is geared towards "rolling" deployments, which, while potentially cheaper/faster are not the safest deployment mechanisms due to the lack of isolation during new cluster bring-up.

A [red/black](https://octopus.com/blog/blue-green-red-black) (a.k.a blue/green) deployment is better suited for cases where you want to bring another Druid cluster up in isolation of the existing one (but in same ZK/K8s discoverability namespace) in order to perform safety/performance checks before cutting over to the new deployment. The Overlord already supports a similar concept of worker "versioning" where it will only schedule peons on MMs/Indexers that are running the version it itself is configured with (or higher), allowing the cluster to eventually drain the ASGs with older Druid versions.

However, this functionality gets us part of the way to supporting what's effectively a zero-downtime (both query + ingest) deployment. To achieve a fully isolated (with the exception of master nodes: {coordinator, overlord}) deployment environment where we can mirror queries, observe state, etc. we also need to support version-based routing of queries.

#### Requirements

To do this, we need two things:

1. Data Isolation

Users cannot be impacted by data loading/unavailability issues when a node is rolled. This is especially pertinent in cases where there is no replica for a segment in a given tier, meaning rolling an instance requires either cloning() the historical first or taking downtime. This applies for both historical and realtime segments: we should be avoiding any/all data availability/freshness problems during deployment.

https://github.com/apache/druid/commit/681cbdee15d279acf976a4e851da3e7a03bbba81 provided support for tier aliases (so duplicate historical tier deployments can be brought up transparently to the user/operator). 

2. Query Isolation

Since Druid queries are generally read-only, isolating the new cluster from user traffic until it is deemed "healthy" is critical to ensure no regressions are deployed. This is also particularly helpful in performing performance/load tests. This PR provides the query routing support to allow user queries to route to strictly `old` nodes (router/broker/historical/peon), and any traffic sent to `new` nodes to route to `new` router/broker/historical/peon.

**NB:** we allow queries from both versions to hit old/new peons to allow for convenient roll-back and easier data availability properties when forcing task group hand-off. This behavior can be toggled via `druid.broker.segment.strictRealtimeDeploymentGroupFilter`, where setting it to `true` would imply realtime servers with non-matching versions would be excluded from query planning.

#### Deployment Steps

The combination of these 2 changes support the following deployment process:

1. Deploy `new` Druid ASGs: router, broker, historical, coordinator, overlord, MM, etc.
4. Configure Coordinator dynamic config to set up tier aliases for new/existing Druid historical tiers (so same set of segments are loaded in parallel onto equivalent tiers across the 2 versions). 
5. Wait for segments to load on the new Druid ASGs
6. Switch coordinator leader to `new` Druid version coordinator 
7. Optionally mirror traffic to the new ASGs (`new` router/broker will be able to query only historicals of their same version; peons are by default queryable by all versions).
8. Switch leader overlord to newer version (using generous timeouts/retries to avoid ingest task RPC failure)
9. Finally, cutover to `new` router/broker/historicals.

This deployment method combines the traditional red/black deployment with Druid's rolling deployment, providing zero ingest downtime as well as zero query downtime for users (both in terms of availability and data freshness). It also provides ample time to experiment/canary changes without impacting user traffic.

#### Release note

<hr>

This PR has:

- [ ] been self-reviewed.


services/src/main/java/org/apache/druid/server/router/TieredBrokerHostSelector.java
```diff
@@ -266,7 +284,26 @@ private Pair<String, Server> getServerPair(String brokerServiceName)
       nodesHolder = servers.get(tierConfig.getDefaultBrokerServiceName());
     }
 
-    return new Pair<>(brokerServiceName, nodesHolder.pick());
+    Server picked = nodesHolder.pick();
```
FrankChen021, line 298: [P2] Deployment-group filtering can be bypassed by router backup routing

When the filter removes all brokers for a service, this returns a Pair with a null server, but `QueryHostFinder.findServerInner` then falls back to `serverBackup` for the same/default service. After an acceptable broker has been cached and later re-announces with a non-matching `deploymentGroup`, or is removed while the same host remains reachable, queries can still be routed to that cached broker despite the configured `acceptableDeploymentGroups`. This breaks the intended fail-closed red/black isolation; the filtered/no-match case should avoid backup fallback or validate/clear backup entries against the deployment-group filter.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/server/http/DataSourcesResource.java
```diff
@@ -1010,4 +1039,49 @@ static boolean isSegmentLoaded(Iterable<ImmutableSegmentLoadInfo> servedSegments
...
+      Iterable<ImmutableSegmentLoadInfo> servedSegments,
+      SegmentDescriptor descriptor,
+      String group
+  )
+  {
+    for (ImmutableSegmentLoadInfo segmentLoadInfo : servedSegments) {
+      if (segmentLoadInfo.getSegment().getInterval().contains(descriptor.getInterval())
+          && segmentLoadInfo.getSegment().getShardSpec().getPartitionNum() == descriptor.getPartitionNumber()
+          && segmentLoadInfo.getSegment().getVersion().compareTo(descriptor.getVersion()) >= 0
+          && segmentLoadInfo.getServers().stream().anyMatch(
```
FrankChen021, line 1088: [P2] Handoff group check ignores the required tier

The per-group handoff check iterates each tier from the matching load rule, but isSegmentLoadedForDeploymentGroup only checks that some replication-target server in the group serves the segment. If the same deploymentGroup name exists in multiple tiers, a segment served by red in tier2 can satisfy the red check for tier1, causing /handoffComplete to return true before the segment is loaded in the rule-required tier. Pass the tier into this helper and require the serving server tier to match the tierEntry being evaluated.

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
```diff
@@ -264,34 +293,56 @@ public void replicateSegment(DataSegment segment, Map<String, Integer> tierToRep
     // Update replicas in every tier
     int dropsQueued = 0;
     for (String tier : allTiersInCluster) {
+      final int requiredReplicas = effectiveTierToReplicaCount.getOrDefault(tier, 0);
+      final Set<String> groupsInTier = tierToActiveGroups.get(tier);
+      if (groupsInTier != null) {
```
FrankChen021: [P2] Ungrouped replicas are skipped once a coordinated group exists

When coordinatingVersions is non-empty and a tier has at least one coordinated deployment group, this branch only updates those group-scoped keys and then continues. Replicas counted under the tier-wide key, such as servers with null deploymentGroup or groups not in coordinatingVersions, are never visited for drops or cancelled loads. During mixed deployments or after disabling a group, segments on those servers can remain loaded even when the rule requires zero replicas. The update loop should also process the tier-wide scope for uncoordinated servers, or explicitly drop/exclude those replicas.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

---

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Reviewed 33 of 33 changed files.

---

server/src/main/java/org/apache/druid/server/coordinator/duty/BalanceSegments.java
```diff
@@ -60,9 +63,20 @@ public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
...
-    );
+    final Set<String> coordinatingVersions = params.getCoordinatorDynamicConfig().getCoordinatingVersions();
+    params.getDruidCluster().getManagedHistoricals().forEach((tier, servers) -> {
+      if (coordinatingVersions.isEmpty()) {
+        new TierSegmentBalancer(tier, servers, maxSegmentsToMove, params).run();
+      } else {
+        // Partition tier servers by deployment group so segments never move across groups.
+        // Servers with no deploymentGroup form their own partition keyed under the empty string.
+        final Map<String, Set<ServerHolder>> serversByGroup = partitionByDeploymentGroup(servers);
+        final int perGroupMax = Math.max(1, maxSegmentsToMove / serversByGroup.size());
```
FrankChen021: P2 Per-group balancing can exceed the configured move cap

When coordinatingVersions is non-empty, maxSegmentsToMove is divided per deployment group but then clamped with Math.max(1, ...). If the configured cap is lower than the number of groups, each group can move one segment, so a tier with three groups and maxSegmentsToMove=1 can move up to three segments in one coordinator run. This violates the documented cap and can cause more segment movement/load pressure than operators configured. Consider distributing the cap so the sum across groups never exceeds maxSegmentsToMove, or skipping groups once the tier budget is exhausted.

server/src/main/java/org/apache/druid/client/BrokerServerView.java
```diff
@@ -246,6 +251,30 @@ private void validateSegmentWatcherConfig(BrokerSegmentWatcherConfig watcherConf
...
+  private boolean isDeploymentGroupAllowed(DruidServerMetadata server)
+  {
+    final boolean isRealtime =
+        server.getType() == ServerType.INDEXER_EXECUTOR || server.getType() == ServerType.REALTIME;
+    if (isRealtime && !segmentWatcherConfig.isStrictRealtimeDeploymentGroupFilter()) {
+      return true;
+    }
+
+    final Set<String> watched = segmentWatcherConfig.getWatchedDeploymentGroups();
+    if (watched != null && !watched.contains(server.getDeploymentGroup())) {
```
FrankChen021: [P1] Propagate deploymentGroup through inventory DruidServer construction

This filter depends on DruidServerMetadata.getDeploymentGroup(), but the normal discovery inventory path still builds DruidServer instances through DruidServer constructors that do not pass DruidNode.getDeploymentGroup() into DruidServerMetadata. In a real cluster, Historicals therefore arrive here with a null deploymentGroup, so setting druid.broker.segment.watchedDeploymentGroups filters out all Historicals; the same lost metadata also prevents coordinator resources from seeing active deployment groups. Please preserve the DruidNode deploymentGroup when constructing inventory DruidServer metadata.

server/src/main/java/org/apache/druid/server/coordinator/duty/BalanceSegments.java
```diff
@@ -60,9 +63,27 @@ public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
...
+        // Servers with no deploymentGroup form their own partition keyed under the empty string.
+        final Map<String, Set<ServerHolder>> serversByGroup = partitionByDeploymentGroup(servers);
+        int remainingGroups = serversByGroup.size();
+        int remainingSegmentsToMove = maxSegmentsToMove;
+        for (final Set<ServerHolder> groupServers : serversByGroup.values()) {
+          final int groupMaxSegmentsToMove =
+              (remainingSegmentsToMove + remainingGroups - 1) / remainingGroups;
+          if (groupMaxSegmentsToMove > 0) {
+            new TierSegmentBalancer(tier, groupServers, groupMaxSegmentsToMove, params).run();
+            remainingSegmentsToMove -= groupMaxSegmentsToMove;
```
FrankChen021: [P2] Do not consume unused balance budget

The per-group loop subtracts groupMaxSegmentsToMove even when TierSegmentBalancer moves zero segments. With maxSegmentsToMove lower than the number of groups, a first group that is already balanced can consume the whole budget every coordinator run, leaving later skewed groups with groupMaxSegmentsToMove == 0 indefinitely. This still honors the numeric cap, but can permanently prevent balancing in later deployment groups; base the budget on actual moves or rotate/fairly allocate across runs.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

Reviewed 35 of 35 changed files.

---

services/src/main/java/org/apache/druid/server/router/TieredBrokerHostSelector.java
```diff
@@ -266,7 +295,26 @@ private Pair<String, Server> getServerPair(String brokerServiceName)
...
+    return new Pair<>(brokerServiceName, picked);
+  }
+
+  private boolean isVersionAllowed(DiscoveryDruidNode node)
+  {
+    final Set<String> routable = tierConfig.getRoutableVersions();
+    if (routable == null || routable.isEmpty()) {
+      return true;
+    }
+    return routable.contains(node.getDruidNode().getVersion());
```
FrankChen021, line 317: [P1] Version filters use the local version after discovery serde

This check relies on the discovered node's announced `DruidNode.version`, but `DiscoveryDruidNode` is read from ZooKeeper with `DefaultObjectMapper`, and `DruidNode.version` is a final field that is not accepted by the `@JsonCreator` while `ALLOW_FINAL_FIELDS_AS_MUTATORS` is disabled. After deserialization, `getVersion()` is therefore the observer process's package version rather than the announcing broker's version. In a mixed red/black deployment, routers can include or exclude brokers based on their own version instead of the broker version, breaking the new `routableVersions` isolation. Make version a deserializable creator property and add a discovery serde test with a non-local version.

server/src/main/java/org/apache/druid/server/http/DataSourcesResource.java
```diff
@@ -980,11 +988,32 @@ public Response isHandOffComplete(
...
+      }
+
+      // When coordinatingVersions is configured, additionally verify per-version coverage.
+      final Set<String> coordinatingVersions = coordinatorConfigManager == null
+          ? Set.of()
+          : coordinatorConfigManager.getCurrentDynamicConfig().getCoordinatingVersions();
+      if (!coordinatingVersions.isEmpty() && matchingLoadRule != null) {
+        final Map<String, Set<String>> activeVersionsByTier =
+            computeActiveVersionsByTier(coordinatingVersions);
+        for (Map.Entry<String, Integer> tierEntry : matchingLoadRule.getTieredReplicants().entrySet()) {
```
FrankChen021, line 1002: [P2] Expand alias tiers before version handoff checks

The coordinator assignment path expands `historicalTierAliases` before applying load rules, but this handoff check iterates the raw `LoadRule` tiers. When a rule targets a virtual alias tier such as `hot -> [hot_1, hot_2]`, `activeVersionsByTier` has entries only for the physical tiers, so the per-version loop for `hot` is skipped and handoff can return true after the segment is loaded anywhere, before all required physical tiers and versions have it. Expand aliases here using the current dynamic config before checking active versions.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

---

PR #19416, opened 2026-05-06
fix: Flaky test runIndexTask_andKillData

GWphua:
### Description

This PR fixes a flaky assertion in `KubernetesClusterDockerTest#test_runIndexTask_andKillData`.

#### Failing Assertion

The failing assertion was caused by the test waiting on `segment/schemaCache/dataSource/removed` before querying `sys.segments`. That metric only proves that the broker schema cache removed the datasource; it does not prove that the broker `MetadataSegmentView`, which backs `sys.segments`, has polled the coordinator after the segments were marked unused.

This PR keeps the production behavior unchanged and updates the test to wait on the cache that actually serves `sys.segments`. The added wait covers the coordinator metadata cache sync and the broker metadata segment view sync before asserting that `sys.segments` is empty.

The `segment/schemaCache/dataSource/removed` metric is still valid for its own cache. The issue was the test using it as a proxy for a different cache’s freshness.

#### Changes
* Add an explicit wait after broker schema-cache datasource removal.
* Wait for the coordinator metadata cache to sync after the unused-segment update.
* Wait for the broker metadata segment view to sync before asserting that `sys.segments` is empty.

This keeps the test aligned with the async cache behavior instead of assuming that schema-cache removal also implies `sys.segments` metadata has refreshed.

<hr>

 * `IngestionSmokeTest`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/IngestionSmokeTest.java
@@ -189,6 +189,9 @@ public class IngestionSmokeTest extends EmbeddedClusterTestBase
                       .hasService("druid/broker")
     );
 
+    waitForNextCoordinatorCacheSync();
+    waitForNextBrokerCacheSync();
+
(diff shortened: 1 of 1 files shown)


FrankChen021 (commented):

---

FrankChen021 (approved):

PR #19417, opened 2026-05-06
fix: Precedence of taskCount, taskCountStart, taskCountMin.

gianm:
PR #18745 included a [discussion](https://github.com/apache/druid/pull/18745#issuecomment-3633901364) of desired behavior of taskCount, taskCountStart, and taskCountMin, but this has not been fully implemented. There are two missed cases:

1) On OL / supervisor restart, the persisted taskCount should be used. Currently, the persisted taskCountStart is used instead if set.

2) On supervisor POST, the previous spec's taskCount should be used if the user provides neither taskCount nor taskCountStart. Currently, in this case, the user's provided taskCountMin is used instead.

In addition, and related to (1), currently deserialization of the ioConfig causes the taskCount to be overwritten by the taskCountStart. This causes the get supervisor and supervisor history APIs to return inaccurate taskCounts.

This patch fixes these issues by doing two things:

1) Update SeekableStreamSupervisorIOConfig to always retain taskCount from the constructor (i.e. when deserializing) if set.

2) Update SeekableStreamSupervisorSpec#merge to handle the special case that taskCountStart has priority on POST.


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -90,16 +99,19 @@ public SeekableStreamSupervisorIOConfig(
     this.autoScalerConfig = autoScalerConfig;
     boolean isAutoScalerAvailable = autoScalerConfig != null;
     this.autoScalerEnabled = isAutoScalerAvailable && autoScalerConfig.getEnableTaskAutoScaler();
-    if (autoScalerEnabled) {
-      // Priority: taskCountStart > taskCount > taskCountMin
+    this.taskCountExplicit = taskCount != null;
+    if (taskCount != null) {
```
Fly-Style, line 103: nit: may use `this.taskCountExplicit` for condition

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -90,16 +99,19 @@ public SeekableStreamSupervisorIOConfig(
...
       this.taskCount = Configs.valueOrDefault(
           autoScalerConfig.getTaskCountStart(),
-          Configs.valueOrDefault(taskCount, autoScalerConfig.getTaskCountMin())
+          autoScalerConfig.getTaskCountMin()
       );
-    } else if (isAutoScalerAvailable) {
-      this.taskCount = Configs.valueOrDefault(taskCount, autoScalerConfig.getTaskCountMin());
     } else {
-      this.taskCount = Configs.valueOrDefault(taskCount, 1);
+      this.taskCount = 1;
```
Fly-Style, line 114: Why this change? Autoscaler also may be _disabled_, but spec may not be null (available).
Isn't it better to assign 1 only as a last resort?
AFAIR, this LOC in this verrsion led to some inaccurate behaviour some time ago, but do not remember exactly what it was. Anyway it is covered by case above.

gianm: The question here is that if `taskCount` is not provided, and an autoscaler is provided yet disabled, what should we use for the initial `taskCount`? This only matters on the very first submission of a supervisor, since if there's an existing supervisor, we'll use its persisted `taskCount` (which will always be there).

Anyway, the old code used `taskCountMin` (ignoring `taskCountStart` if present). I thought that was strange: why use the autoscaler configs in a slightly different way if the autoscaler is not enabled? This PR changes it to use 1, on the grounds that if the autoscaler is submitted in a disabled state then its configuration should not apply.

Fly-Style: Agree, thanks.

Fly-Style (approved):

FrankChen021 (commented):

---

PR #19418, opened 2026-05-06
fix: Use FileSystem on getPaths() instead of mapreduce.Job

### LINKED ISSUE #19411: HdfsInputSource with index_parallel unnecessarily requires KMS delegation token when using Kerberized HDFS
### Affected Version

Druid 36.0.0 (likely affects all versions using `druid-hdfs-storage` extension with Kerberos)

### Description

When using `index_parallel` with `HdfsInputSource` on a Kerberized HDFS cluster where the NameNode has KMS configured, the ingestion task attempts to acquire a KMS delegation token even though native ingestion does not need one. Druid authenticates directly via the Kerberos TGT and never interacts with KMS.

**Root Cause:**

`HdfsInputSource.getPaths()` uses `FileInputFormat.getSplits()` solely for HDFS glob/path expansion. As a side effect, `FileInputFormat.listStatus()` calls `TokenCache.obtainTokensForNamenodes()`, which calls `fs.addDelegationTokens()`. This cascades into `KMSClientProvider.getDelegationToken()` via `DelegationTokenIssuer.collectDelegationTokens()`. The acquired delegation tokens are never actually used by Druid's native ingestion.

The KMS URI is discovered from the NameNode via RPC at runtime, so client-side configuration overrides (`hadoop.security.key.provider.path`, `dfs.encryption.key.provider.uri` set to empty) do not prevent the KMS contact attempt, because Hadoop treats empty strings as "not set" and falls back to querying the NameNode.

**Proposed fix:**

Replace the `FileInputFormat`/`Job` usage in `HdfsInputSource.getPaths()` with direct `FileSystem.globStatus()` calls. This achieves the same glob expansion without entering the MapReduce `TokenCache` code path at all. For example:

```java
public static Collection<Path> getPaths(List<String> inputPaths, Configuration configuration) throws IOException {
    if (inputPaths.isEmpty()) {
        return Collections.emptySet();
    }
    Set<Path> paths = new LinkedHashSet<>();
    for (String inputPath : inputPaths) {
        Path p = new Path(inputPath);
        FileSystem fs = p.getFileSystem(configuration);
        FileStatus[] statuses = fs.globStatus(p);
        if (statuses != null) {
            for (FileStatus status : statuses) {
                if (status.getLen() > 0) {
                    paths.add(status.getPath());
                }
            }
        }
    }
    return paths;
}
```

**Call chain leading to unnecessary KMS contact:**

```
HdfsInputSource.getPaths()
  → FileInputFormat.getSplits()
    → FileInputFormat.listStatus()
      → TokenCache.obtainTokensForNamenodes()
        → fs.addDelegationTokens()
          → DelegationTokenIssuer.collectDelegationTokens()
            → KMSClientProvider.getDelegationToken()  ← unnecessary KMS contact
```

**Relevant source:**
- [`HdfsInputSource.java` (getPaths method)](https://github.com/apache/druid/blob/druid-36.0.0/extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java#L133-L165)
- [`FileInputFormat.java` (listStatus calling TokenCache)](https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/lib/input/FileInputFormat.java#L257)
- [`TokenCache.java` (obtainTokensForNamenodes)](https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/security/TokenCache.java)

JWuCines:
Fixes #19411.

### Description

When using `index_parallel` with `HdfsInputSource` on a Kerberized HDFS cluster where the NameNode has KMS configured, the ingestion task unnecessarily attempts to acquire a KMS delegation token. This happens because `HdfsInputSource.getPaths()` uses `FileInputFormat.getSplits()` for path/glob expansion, which internally calls `TokenCache.obtainTokensForNamenodes()`, cascading into `KMSClientProvider.getDelegationToken()`. Druid's native ingestion authenticates directly via Kerberos TGT and never needs these delegation tokens.

#### Replaced FileInputFormat with direct FileSystem.globStatus() calls

Replaced the `FileInputFormat`/`Job`-based path expansion in `HdfsInputSource.getPaths()` with direct `FileSystem.globStatus()` calls. This achieves the same HDFS glob expansion without entering the MapReduce `TokenCache` code path, eliminating the unnecessary KMS contact.

The inner `HdfsFileInputFormat` helper class and all `org.apache.hadoop.mapreduce` imports have been removed. No other file in the `druid-hdfs-storage` module references the MapReduce API.

#### Preserved FileInputFormat filtering semantics

- **Hidden file filter**: Files and directories whose names start with `_` or `.` are excluded, matching Hadoop's `FileInputFormat.hiddenFileFilter`.
- **Non-recursive directory listing**: When a path points to a directory, only immediate files are listed; subdirectories are not traversed. This matches `FileInputFormat`'s default behavior when `mapreduce.input.fileinputformat.input.dir.recursive` is not set.
- **Comma-separated path splitting**: Input path strings are split using `org.apache.hadoop.util.StringUtils.split()`, preserving the same comma-separation and escape behavior as `FileInputFormat.addInputPaths()`.

#### Updated documentation

Updated the `paths` property description in `docs/ingestion/input-sources.md` to document the non-recursive directory traversal behavior, hidden file filtering, and the use of glob patterns (e.g., `**/*.json`) for ingesting files from nested directories.

#### Added unit tests for getPaths() edge cases

Added a new `GetPathsTest` inner class to `HdfsInputSourceTest` with seven tests:
- `testGetPathsWithGlobMatchingNoFiles` — glob matching no files returns an empty collection
- `testGetPathsFiltersZeroLengthFiles` — zero-length files are excluded, non-empty files are included
- `testGetPathsWithMultipleInputPaths` — multiple distinct glob patterns are resolved correctly
- `testGetPathsWithCommaSeparatedString` — comma-separated path strings are split and resolved
- `testGetPathsFiltersHiddenFiles` — files starting with `_` or `.` are excluded from glob results
- `testGetPathsDirectoryListsFilesNonRecursively` — subdirectories and hidden files within a directory are skipped
- `testGetPathsSkipsHiddenDirectories` — hidden directories matched by a glob are not descended into

#### Release note

Fixed an issue where `HdfsInputSource` with `index_parallel` unnecessarily contacted KMS when using Kerberized HDFS, causing task failures if KMS was unreachable. The fix replaces the internal use of Hadoop MapReduce `FileInputFormat` for path expansion with direct `FileSystem.globStatus()` calls, while preserving hidden file filtering and non-recursive directory listing semantics.

<hr>

 * `HdfsInputSource`
 * `HdfsInputSourceTest`
 * `docs/ingestion/input-sources.md`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
+++ b/extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
@@ -48,13 +48,9 @@ import org.apache.druid.java.util.common.StringUtils;
 import org.apache.druid.storage.hdfs.HdfsStorageDruidModule;
 import org.apache.druid.utils.Streams;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.mapreduce.Job;
-import org.apache.hadoop.mapreduce.JobContext;
-import org.apache.hadoop.mapreduce.RecordReader;
-import org.apache.hadoop.mapreduce.TaskAttemptContext;
-import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
-import org.apache.hadoop.mapreduce.lib.input.FileSplit;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -66,6 +62,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -153,39 +150,31 @@ public class HdfsInputSource
       return Collections.emptySet();
     }
 
-    // Use FileInputFormat to read splits. To do this, we need to make a fake Job.
-    Job job = Job.getInstance(configuration);
-
-    // Add paths to the fake JobContext.
-    for (String inputPath : inputPaths) {
-      FileInputFormat.addInputPaths(job, inputPath);
+    final Set<Path> paths = new LinkedHashSet<>();
+    for (final String inputPath : inputPaths) {
+      final Path p = new Path(inputPath);
+      final FileSystem fs = p.getFileSystem(configuration);
+      final FileStatus[] statuses = fs.globStatus(p);
+      if (statuses != null) {
+        for (final FileStatus status : statuses) {
+          addFilesRecursively(fs, status, paths);
+        }
+      }
     }
-
-    return new HdfsFileInputFormat().getSplits(job)
-                                    .stream()
-                                    .filter(split -> ((FileSplit) split).getLength() > 0)
-                                    .map(split -> ((FileSplit) split).getPath())
-                                    .collect(Collectors.toSet());
+    return paths;
   }
 
-  /**
-   * Helper for leveraging hadoop code to interpret HDFS paths with globs
-   */
-  private static class HdfsFileInputFormat extends FileInputFormat<Object, Object>
+  private static void addFilesRecursively(FileSystem fs, FileStatus status, Set<Path> paths) throws IOException
   {
-    @Override
-    public RecordReader<Object, Object> createRecordReader(
-        org.apache.hadoop.mapreduce.InputSplit inputSplit,
-        TaskAttemptContext taskAttemptContext
-    )
-    {
-      throw new UnsupportedOperationException();
-    }
-
-    @Override
-    protected boolean isSplitable(JobContext context, Path filename)
-    {
-      return false;  // prevent generating extra paths
+    if (status.isDirectory()) {
+      final FileStatus[] children = fs.listStatus(status.getPath());
+      if (children != null) {
+        for (final FileStatus child : children) {
+          addFilesRecursively(fs, child, paths);
+        }
+      }
+    } else if (status.getLen() > 0) {
+      paths.add(status.getPath());
     }
   }
 
--- a/extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
+++ b/extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
@@ -144,6 +144,15 @@ public class HdfsInputSource
     }
   }
 
+  /**
+   * Matches Hadoop's FileInputFormat hidden-file filter: rejects paths whose name starts with '_' or '.'.
+   */
+  private static boolean isHiddenPath(Path path)
+  {
+    final String name = path.getName();
+    return name.startsWith("_") || name.startsWith(".");
+  }
+
   public static Collection<Path> getPaths(List<String> inputPaths, Configuration configuration) throws IOException
   {
     if (inputPaths.isEmpty()) {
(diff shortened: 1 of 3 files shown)


extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
```diff
@@ -153,39 +150,31 @@ public static Collection<Path> getPaths(List<String> inputPaths, Configuration c
...
 
-    // Use FileInputFormat to read splits. To do this, we need to make a fake Job.
-    Job job = Job.getInstance(configuration);
-
-    // Add paths to the fake JobContext.
-    for (String inputPath : inputPaths) {
-      FileInputFormat.addInputPaths(job, inputPath);
+    final Set<Path> paths = new LinkedHashSet<>();
+    for (final String inputPath : inputPaths) {
+      final Path p = new Path(inputPath);
```
FrankChen021: [P1] Comma-separated paths string no longer works

The documented HDFS `paths` property can be a comma-separated string, and the previous `FileInputFormat.addInputPaths(job, inputPath)` split that string using Hadoop's path parser. The replacement constructs one `Path` from the entire string, so a spec like `"hdfs://nn/a.json,hdfs://nn/b.json"` is now treated as a single path/glob and will fail or match nothing. Split comma-separated string inputs before creating `Path`s, preserving Hadoop's escaping/brace behavior where possible.

extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
```diff
@@ -153,39 +150,31 @@ public static Collection<Path> getPaths(List<String> inputPaths, Configuration c
...
-    {
-      throw new UnsupportedOperationException();
-    }
-
-    @Override
-    protected boolean isSplitable(JobContext context, Path filename)
-    {
-      return false;  // prevent generating extra paths
+    if (status.isDirectory()) {
+      final FileStatus[] children = fs.listStatus(status.getPath());
```
FrankChen021: [P2] FileInputFormat filtering semantics were dropped

The old `FileInputFormat` listing applied Hadoop's hidden-file filter, excluding path names starting with `_` or `.`, and only recursed into directories when `mapreduce.input.fileinputformat.input.dir.recursive` was enabled. The new direct `listStatus` traversal always recurses and never filters hidden files, so existing directory inputs can start ingesting nested data and non-empty marker/metadata files that were previously skipped. Reintroduce the same path filter and recursion behavior while avoiding the MapReduce token path.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

---

extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
JWuCines: Fixed using `org.apache.hadoop.util.StringUtils.split()` method, same as FileInputFormat (https://github.com/apache/druid/pull/19418/changes/ac8479ef1752eb29b9e0cd2c4f3973853f89d6a6#diff-9eeaffd33bf4ce833ed88c0a9d8bbed94ce80e1eda01ec78fd5157a98dfbc6b8R164)

extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
JWuCines: Added a method to filter out hidden-file filter: https://github.com/apache/druid/pull/19418/changes/ac8479ef1752eb29b9e0cd2c4f3973853f89d6a6#diff-9eeaffd33bf4ce833ed88c0a9d8bbed94ce80e1eda01ec78fd5157a98dfbc6b8R150

Made the change to not recursively list all the files: `mapreduce.input.fileinputformat.input.dir.recursive` will not be taken in account. Changed the documentation to explicitly inform about it and to suggest to use glob instead

extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
FrankChen021: Thanks, this looks addressed in the current code: `getPaths()` now uses Hadoop `StringUtils.split()` before constructing each `Path`, and the added comma-separated test covers the regression case.

extensions-core/hdfs-storage/src/main/java/org/apache/druid/inputsource/hdfs/HdfsInputSource.java
FrankChen021: Thanks, this addresses my concern: hidden names are filtered, directory inputs are listed non-recursively by default, and the docs now point users to explicit glob patterns for nested ingestion.

FrankChen021 (commented):

---

FrankChen021 (approved):

JWuCines:
Hey @FrankChen021!
Thank you for the approval!
Can you also merge the PR as reviewer and committer? I'm not able since it is my first PR on Druid project.
Thank you!

PR #19419, opened 2026-05-06
feat: podTemplateSelectionKey context for picking pod templates from task specs

rmn7:
### Description
To use different pod templates for different tasks, users today have to setup an `executionconfig` with a `selectorBased` strategy, setup equivalent tags on this config and pass these tags to the task context. This change allows the `DynamicConfigPodTemplateSelector` to use a new `podTemplateSelectionKey` context key from the task to directly pick the pod template based on its `selectionKey` in the runtime properties. I chose to fail task creation instead of falling back to the existing selection strategy since users would only set this context param intentionally and a missing pod template indicates misconfiguration. The behavior is gated on a new cluster-level runtime property, `druid.indexer.runner.allowTaskPodTemplateSelection`, that defaults to `false` so operators must opt-in.

#### Release note
You can now set podTemplateSelectionKey in a task's context to pick a specific configured pod template (e.g. `druid.indexer.runner.k8s.podTemplate.<selectionKey>`) without configuring a selectorBased strategy. The override takes precedence over both the default and selectorBased strategies. The feature is gated on the new runtime property `druid.indexer.runner.allowTaskPodTemplateSelection`, which defaults to `false`. If the named template isn't configured, the task fails to launch.

<hr>

 * `DynamicConfigPodTemplateSelector`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/taskadapter/DynamicConfigPodTemplateSelector.java
+++ b/extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/taskadapter/DynamicConfigPodTemplateSelector.java
@@ -26,6 +26,7 @@ import io.fabric8.kubernetes.client.utils.Serialization;
 import org.apache.druid.guice.IndexingServiceModuleHelper;
 import org.apache.druid.indexing.common.task.Task;
 import org.apache.druid.java.util.common.IAE;
+import org.apache.druid.java.util.common.logger.Logger;
 import org.apache.druid.k8s.overlord.KubernetesTaskRunnerEffectiveConfig;
 import org.apache.druid.k8s.overlord.common.DruidK8sConstants;
 
@@ -38,6 +39,8 @@ import java.util.Set;
 
 public class DynamicConfigPodTemplateSelector implements PodTemplateSelector
 {
+  private static final Logger log = new Logger(DynamicConfigPodTemplateSelector.class);
+
   private static final String TASK_PROPERTY = IndexingServiceModuleHelper.INDEXER_RUNNER_PROPERTY_PREFIX
                                               + ".k8s.podTemplate.";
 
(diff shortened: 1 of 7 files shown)


capistrant (changes requested):
cool new feature. I think we need a cluster level runtime property that enables/disables the ability for tasks to override the pod spec. We should default to disabled, thus forcing operators to acknowledge this and opt in to letting whoever has task submit privs control their spec. I think at this point we'd be able to merge as an mvp for druid 38. There may be value for some in getting more granular with what identities are able to specify this context key, and even allowing/disallowing specs based on role (like you have to be an admin to opt into some super expensive spec, etc. etc.). But for the purposes of the first cut, having an operator either allow or disallow this behavior cluster wide is fine

FrankChen021 (commented):

---

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/taskadapter/DynamicConfigPodTemplateSelector.java
```diff
@@ -119,16 +119,18 @@ private void validateTemplateSupplier(Supplier<PodTemplate> templateSupplier) th
...
-    String requested = task.getContextValue(DruidK8sConstants.TASK_CONTEXT_POD_TEMPLATE_SELECTION_KEY);
-    if (requested != null) {
-      Supplier<PodTemplate> supplier = podTemplates.get(requested);
-      if (supplier == null) {
-        throw new IAE(
-            "Task [%s] requested pod template [%s] via context key, but no such template is configured.",
-            task.getId(), requested
-        );
+    if (effectiveConfig.isAllowTaskPodTemplateSelection()) {
+      String requested = task.getContextValue(DruidK8sConstants.TASK_CONTEXT_POD_TEMPLATE_SELECTION_KEY);
```
capistrant: nit: for observability it might be nice to log a warn if this is not null + `effectiveConfig.isAllowTaskPodTemplateSelection()` is false. that way the owner of the spec has at least something to find if they are confused. something like -

`task[<insert_task_id>] set <DruidK8sConstants.TASK_CONTEXT_POD_TEMPLATE_SELECTION_KEY> in task context. Not applying because cluster config does not allow overriding pod spec at runtime.`

extensions-core/kubernetes-overlord-extensions/src/main/java/org/apache/druid/k8s/overlord/taskadapter/DynamicConfigPodTemplateSelector.java
```diff
@@ -119,16 +119,18 @@ private void validateTemplateSupplier(Supplier<PodTemplate> templateSupplier) th
...
+      if (requested != null) {
+        Supplier<PodTemplate> supplier = podTemplates.get(requested);
+        if (supplier == null) {
+          throw new IAE(
+              "Task [%s] requested pod template [%s] via context key, but no such template is configured.",
+              task.getId(), requested
+          );
+        }
+        return Optional.of(new PodTemplateWithName(requested, supplier.get()));
```
capistrant: debug level log stating the task is overriding the pod spec could be helpful down the road too

capistrant (commented):
looks good! I think a few logs to help with observability and it'll be ready to roll

capistrant (approved):

PR #19422, opened 2026-05-06
fix: adjust diskNormalized strategy to scale cost exponentially with disk utilization

jtuglu1:
### Description

The existing linear penalization factor is still ineffective in large skew scenarios where the CostBalancerStrategy's cost forces a move/load (even with the utilization-based penalty). This switches the penalty to scale exponentially with the disk utilization, ensuring that near-full historicals are penalized.

This is also particularly helpful when the size of segments on the cluster vary wildly.

##### Before (strategy=cost, roundRobinSegmentLoading=true)

Screenshot 2026-05-29 at 8 36 30 PM

##### After (strategy=diskNormalized, roundRobinSegmentLoading=true, 5% threshold)
Screenshot 2026-05-29 at 8 35 37 PM

#### Release note

Adjust diskNormalized segment balancer strategy to scale cost exponentially with disk utilization. Marks this strategy ready for production usage.

<hr>

This PR has:

- [ ] been self-reviewed.


server/src/main/java/org/apache/druid/server/coordinator/balancer/DiskNormalizedCostBalancerStrategy.java
```diff
@@ -85,17 +85,16 @@ protected double computePlacementCost(
...
     final long maxSize = server.getMaxSize();
     if (maxSize <= 0) {
       return cost;
     }
 
-    double usageRatio = (double) server.getSizeUsed() / maxSize;
-    double normalizedCost = cost * usageRatio;
+    final double usageRatio = (double) server.getSizeUsed() / maxSize;
+    final double headroom = Math.max(EPSILON, 1.0 - usageRatio);
+    double normalizedCost = cost / headroom;
```
FrankChen021, line 99: [P1] Existing threshold test now fails

Changing normalization to `cost / headroom` makes the existing `testThresholdBlocksMarginalMove` scenario choose DEST: source is roughly `38K / 0.20 * 0.95 = 180.5K`, while dest is `40K / 0.26 = 153.8K`. The test still asserts null, so the server test suite should fail unless the threshold scenario or algorithm is adjusted.

server/src/main/java/org/apache/druid/server/coordinator/balancer/DiskNormalizedCostBalancerStrategy.java
```diff
@@ -26,40 +26,40 @@
 
 /**
  * A {@link BalancerStrategy} which normalizes the cost of placing a segment on a
- * server as calculated by {@link CostBalancerStrategy} by multiplying it by the
- * server's disk usage ratio.
+ * server as calculated by {@link CostBalancerStrategy} by dividing by the
```
FrankChen021, line 29: [P2] Public docs still describe the old formula

The implementation and Javadoc now divide by available headroom, but `docs/design/coordinator.md` and `docs/configuration/index.md` still say `diskNormalized` multiplies cost by `diskUsed / maxSize`. That leaves user-facing behavior documentation incorrect for this config option.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

---

gianm (commented):
Have you tried this out in a real cluster? I wonder if we should have a property to restore the old behavior in case something undesirable happens.

I also struggle with the naming. The term "normalized" makes me think that costs will be scaled somehow into some bounded range. The adjusted cost function in the PR is doing something different from normalization, it's blowing up the cost as servers fill up.

I wonder if this could all be cleaner by having the cost adjustments be implemented as options to the `cost` balancer rather than entirely distinct balancer strategies. What are your thoughts on this?

Btw, the docs would need to be updated with this formula change, since the "multiplies the cost of placing a segment on a server by the server's disk usage ratio (diskUsed / maxSize)" is no longer what happens.

jtuglu1:
> Have you tried this out in a real cluster? I wonder if we should have a property to restore the old behavior in case something undesirable happens.
> 
> I also struggle with the naming. The term "normalized" makes me think that costs will be scaled somehow into some bounded range. The adjusted cost function in the PR is doing something different from normalization, it's blowing up the cost as servers fill up.
> 
> I wonder if this could all be cleaner by having the cost adjustments be implemented as options to the `cost` balancer rather than entirely distinct balancer strategies. What are your thoughts on this?
> 
> Btw, the docs would need to be updated with this formula change, since the "multiplies the cost of placing a segment on a server by the server's disk usage ratio (diskUsed / maxSize)" is no longer what happens.

Yes. @gianm old behavior was already broken (docs recommended it not be used in production), so no real Druid release exists with this strategy recommended for use in production. I'm attempting to make this strategy actually usable in production.

> I wonder if this could all be cleaner by having the cost adjustments be implemented as options to the cost balancer rather than entirely distinct balancer strategies. What are your thoughts on this?

I think this is fine, but I'd rather fix it in isolation for backwards compatibility first. We can then see if we can unify both `cost` and `diskNormalized`. Main part of the problem is the former doesn't really account for segment size as much as it does segment count.

server/src/main/java/org/apache/druid/server/coordinator/balancer/DiskNormalizedCostBalancerStrategy.java
jtuglu1: Fixed. `docs/design/coordinator.md` and `docs/configuration/index.md` now describe the projected-headroom formula instead of the old `cost * (diskUsed / maxSize)` multiplier.

server/src/main/java/org/apache/druid/server/coordinator/balancer/DiskNormalizedCostBalancerStrategy.java
jtuglu1: fixed

FrankChen021 (commented):
I have reviewed the follow-up replies and current code/docs/tests around the prior comments; the discussed issues are handled.

---

maytasm (approved):
I agree we don't need a rollback property. The old behavior was already broken — so a toggle to restore that behavior adds config surface for no real benefit. If something goes wrong in production, operators can switch back to the plain cost strategy, which is the actual production-proven fallback. A toggle between two variants of diskNormalized (one of which was never recommended for production) isn't worth the complexity imo.

On naming — Names like diskWeighted or diskPenalized would be more accurate. Since the old behavior was never recommended for production, we can also rename before we make this strategy supported/GA. We could also accept both names during deserialization and deprecate the old one. That said, the change still does what the docs describes "weights the cost of placing a segment on a server with the disk usage ratio of the server" — the penalization is still fundamentally based on the usage ratio, just with a steeper curve.

On unifying into cost balancer options — I think this sounds fine, but I think fixing the formula in isolation first is the pragmatic call.

docs/design/coordinator.md
```diff
@@ -88,7 +88,7 @@ But in a tier with several Historicals (or a low replication factor), segment re
...
+- `diskWeighted`: A derivative of the `cost` strategy that divides the cost of placing a segment on a server by the server's projected available disk headroom. The projected usage ratio is `(diskUsed + segmentSizeIfNotAlreadyProjected) / maxSize`, so the disk-adjusted cost is `cost / max(EPSILON, 1 - projectedUsageRatio)`. This strongly penalizes servers that would be nearly full after placement and drives disk utilization to equalize across the tier, which is useful when historicals within a tier hold segments of widely varying sizes. To prevent oscillation when servers have similar headroom, a segment that is already placed on a server receives a cost discount; a move only fires when the destination saves at least `druid.coordinator.balancer.diskWeighted.moveCostSavingsThreshold` (default `0.05`, i.e. 5%) of the source's cost.
```
abhishekrb19: IMO, it’d be best to leave the rename out since it would break operators’ configurations since it's technically a breaking change that I'm not sure is worth doing at this point. Even though the docs say it’s not recommended for production use, it has been documented for a while and doesn’t carry an experimental tag.

Two thoughts:
1. Introduce a new balancer strategy 
2. Or use jackson to handle both types like @maytasm may have noted `diskNormalized` / `diskWeighted` and deprecate `diskNormalized`

jtuglu1: Sure, I'll just revert to diskNormalized. I'd rather not add more complexity considering this strategy was never meant for production usage anyways. I would argue this change makes it strictly better performing than the implementation in prior releases.

abhishekrb19 (approved):

PR #19423, opened 2026-05-07
perf: Optimize aggregators for groupBy queries.

gianm:
Previously aggregators were only optimized for timeseries and topN. This patch adds optimization for groupBy too. It also adds a context parameter "optimizeAggregators" that can be used to switch off optimization, in case this is useful for debugging purposes.

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/querykit/groupby/GroupByPreShuffleFrameProcessor.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/querykit/groupby/GroupByPreShuffleFrameProcessor.java
@@ -49,6 +49,7 @@ import org.apache.druid.msq.input.table.SegmentsInputSlice;
 import org.apache.druid.msq.querykit.BaseLeafFrameProcessor;
 import org.apache.druid.msq.querykit.ReadableInput;
 import org.apache.druid.msq.querykit.SegmentReferenceHolder;
+import org.apache.druid.query.PerSegmentQueryOptimizationContext;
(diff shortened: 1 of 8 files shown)


multi-stage-query/src/main/java/org/apache/druid/msq/querykit/groupby/GroupByPreShuffleFrameProcessor.java
```diff
@@ -193,8 +194,11 @@ protected ReturnOrAwait<Unit> runWithSegment(final SegmentReferenceHolder segmen
         rowSequence = Sequences.simple(List.of(GroupByTimeBoundaryUtils.computeTimeBoundaryResult(query, tbi)));
       } else {
         // Resolve this query using a cursor.
+        final GroupByQuery segmentQuery = (GroupByQuery) query
+            .withQuerySegmentSpec(new SpecificSegmentSpec(segmentHolder.getDescriptor()))
+            .optimizeForSegment(new PerSegmentQueryOptimizationContext(segmentHolder.getDescriptor()));
```
clintropolis, line 199: maybe withQuerySegmentSpec should just set this too?

clintropolis (approved):
:metal:

multi-stage-query/src/main/java/org/apache/druid/msq/querykit/groupby/GroupByPreShuffleFrameProcessor.java
gianm: I don't mind them being separate calls.

PR #19424, opened 2026-05-07
build: Add lock to dependency cache for Dockerfile

GWphua:
This PR updates the Docker image build to use locked BuildKit cache mounts for the Maven local repository.

The Dockerfile already uses `RUN --mount=type=cache,target=/root/.m2` to reuse Maven dependencies across image builds. By default, BuildKit cache mounts are shared, which allows multiple concurrent builds to write to the same Maven cache at the same time. Maven’s local repository is not a good fit for concurrent writers, so parallel Docker builds can corrupt or partially write cached artifacts.

This PR changes both Maven cache mounts to use `sharing=locked`, causing concurrent builds that use the same `/root/.m2` cache to wait for the active writer to finish before reusing the cache.

<hr>

 * `distribution/docker/Dockerfile`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/distribution/docker/Dockerfile
+++ b/distribution/docker/Dockerfile
@@ -35,7 +35,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
 
 COPY . /src
 WORKDIR /src
-RUN --mount=type=cache,target=/root/.m2 if [ "$BUILD_FROM_SOURCE" = "true" ]; then \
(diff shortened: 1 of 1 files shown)


FrankChen021 (commented):

---

FrankChen021 (approved):

PR #19425, opened 2026-05-07
fix: Address issue where Broker upgrade before Overlord causes NPE

GWphua:
### Description

This PR fixes a rolling-upgrade compatibility issue when Brokers are upgraded before Overlords. In my case, I faced the following issue when upgrading from v27 -> v37.

Newer Brokers expect `SupervisorStatus` responses to include the `dataSource` field. Older Overlords do not include that field, so a newer Broker can deserialize an old Overlord response with a null datasource and later hit a `NullPointerException`.

This PR preserves compatibility by falling back to the supervisor `id` when `dataSource` is absent. This matches the old response shape where the supervisor id is the datasource identifier for this status object.

A regression test was added for deserializing an old Overlord response without `dataSource`.

#### Release note

Fixes a rolling-upgrade issue where upgrading Brokers before Overlords could cause a `NullPointerException` while reading supervisor status responses from older Overlords.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorStatus.java
+++ b/server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorStatus.java
@@ -57,7 +57,7 @@ public class SupervisorStatus
   )
   {
     this.id = Preconditions.checkNotNull(builder.id, "id");
-    this.dataSource = builder.dataSource;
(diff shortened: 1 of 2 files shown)


FrankChen021 (commented):

---

FrankChen021 (approved):

PR #19426, opened 2026-05-07
perf: Avoid creating processing buffers beyond what is needed.

gianm:
In Dart, processing buffers are sliced up from the merge buffer. For stages that do not use all processing threads -- perhaps because they do not have enough inputs -- we can be more efficient with memory by slicing the merge buffer based on the actual number of processors, not the number of processing threads.

This patch addresses it by deferring the choice of how many buffers are needed until the stage actually starts executing. At that point, it knows how many processors it will create.

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartProcessingBuffersProvider.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartProcessingBuffersProvider.java
@@ -109,6 +109,10 @@ public class DartProcessingBuffersProvider implements ProcessingBuffersProvider
         );
       }
 
+      if (requestedSlices < 1) {
+        throw DruidException.defensive("requestedSlices[%d] must be positive", requestedSlices);
+      }
+
       final int sliceSize = chunk.capacity() / requestedSlices;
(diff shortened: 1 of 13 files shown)


FrankChen021 (commented):

---

FrankChen021 (commented):

---

multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartProcessingBuffersProvider.java
```diff
@@ -67,27 +68,55 @@ public ResourceHolder<ProcessingBuffersSet> acquire(final int poolSize, final lo
...
+    {
+      if (requestedSlices > maxSlices) {
+        throw DruidException.defensive(
+            "requestedSlices[%d] too large for maxSlices[%d]",
+            requestedSlices,
+            maxSlices
+        );
+      }
+
+      final int sliceSize = chunk.capacity() / requestedSlices;
```
capistrant, line 116: defensive check that requested slices > 0 or do we trust the upstream callers always request at least one slice?

gianm: Added a defensive check.

capistrant (approved):

PR #19428, opened 2026-05-07
test: Write embedded test task logs to the main logger.

gianm:
In production, Indexers write task logs to their own dedicated files in order to make them separately uploadable and viewable. In embedded tests, it's better to write task logs to the main log file, to keep all logging output for the test run in one place.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/ThreadingTaskRunner.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/ThreadingTaskRunner.java
@@ -84,8 +84,8 @@ import java.util.concurrent.TimeoutException;
  *   shutdown on the Task objects. Only one shutdown per-task can be running at a given time,
  *   so we allocate one control thread per worker slot.
  *
- * Note that separate task logs are not currently supported, all task log entries will be written to the Indexer
- * process log instead.
+ * By default, the log output of each task is routed to a separate per-task file via Log4j thread context. This
+ * behavior can be disabled by setting {@code druid.worker.useSeparateTaskLogFiles=false}.
  */
 public class ThreadingTaskRunner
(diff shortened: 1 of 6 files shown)


imply-cheddar (approved):

clintropolis (approved):

indexing-service/src/main/java/org/apache/druid/indexing/overlord/ThreadingTaskRunner.java
```diff
@@ -191,7 +191,8 @@ public TaskStatus call()
 
                             final File taskFile = new File(taskDir, "task.json");
                             final File reportsFile = new File(attemptDir, "report.json");
-                            final File logFile = new File(taskDir, "log");
+                            final File logFile =
```
FrankChen021, line 194: [P2] Disabling task log files does not stop appenderator log routing

When druid.worker.useSeparateTaskLogFiles=false, the runner skips the task log file here, but ingestion tasks that create appenderators can still install per-task log routing through UnifiedIndexerAppenderatorsManager, whose override unconditionally calls Appenderators.setTaskThreadContextForIndexers(taskId, taskDirectory.getTaskLogFile(taskId)). BatchAppenderator and StreamAppenderator invoke that override on persist/push/background executor paths, so those logs can still be routed to per-task files instead of only the Indexer process log. The new flag needs to be honored by the appenderator manager/override as well.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

FrankChen021 (commented):
Reviewed 6 of 6 changed files for correctness, edge cases, concurrency, and integration risks; no issues found.

---

PR #19431, opened 2026-05-08
perf: batch reset invalid stream offsets in stream supervisor

jtuglu1:
#### Problem

I've noticed that recent Druid can be extremely slow to recover from a lagging supervisor (where offsets are invalid and `tuningConfig.resetOffsetAutomatically=true`). I have observed the following issue occurs:

1. Kafka supervisor falls behind, offsets are invalid
2. Supervisor starts [creating a new task group](https://github.com/apache/druid/blob/-/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java#L4104)
3. Supervisor [runs into](https://github.com/apache/druid/blob/master/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java#L4287) an invalid offset for a particular partition, throwing an exception. This exception propagates up to the supervisor, forcing us to wait until another round of `runInternal()`, which can take O(XXs) depending on the supervisor run loop configuration. Since these partitions are evaluated serially, if multiple partitions are invalid offsets, this means we keep looping for every invalid offset for every partition for every taskGroup, leading to many minutes of downtime. This issue is exacerbated when ingesting from large Kafka topics with many partitions.

#### Solution

For each taskGroup, identify + reset all invalid partition offsets in one go. Once we have established the set of invalid offsets, perform an internal reset, then throw the exception. This reduces the time-to-recovery of a fatally-lagged supervisor from N `runInternal()` calls to 1 `runInternal()` call, where N is the # of invalid partition offsets.

#### Release note

Batch reset invalid stream offsets in stream supervisor to speed up stream supervisor recovery.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
@@ -4192,8 +4192,8 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
       resetInternal(createDataSourceMetaDataForReset(ioConfig.getStream(), partitionsToReset));
       throw new StreamException(
           new ISE(
-              "Previous sequenceNumbers for partitions %s are no longer available - automatically resetting sequences",
-              partitionsToReset.keySet()
+              "Previous sequenceNumbers %s are no longer available - automatically resetting sequences",
+              partitionsToReset
           )
       );
     }
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
@@ -4099,6 +4099,12 @@ public abstract class SeekableStreamSupervisor<PartitionIdType, SequenceOffsetTy
     );
 
     // check that there is a current task group for each group of partitions in [partitionGroups]
+    // Fetch metadata offsets once and collect all stale partitions across all groups before committing
+    // any state changes. New TaskGroups are staged locally and only written to activelyReadingTaskGroups
+    // if no reset is required, keeping state consistent: either all new groups are committed or none are.
+    final Map<PartitionIdType, SequenceOffsetType> metadataOffsets = getOffsetsFromMetadataStorage();
+    final Map<PartitionIdType, SequenceOffsetType> partitionsToReset = new HashMap<>();
+    final Map<Integer, TaskGroup> newTaskGroups = new HashMap<>();
     for (Integer groupId : partitionGroups.keySet()) {
       if (!activelyReadingTaskGroups.containsKey(groupId)) {
         log.info("Creating new taskGroup[%d] for partitions[%s].", groupId, partitionGroups.get(groupId));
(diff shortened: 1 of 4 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4099,6 +4099,12 @@ private void createNewTasks() throws JsonProcessingException
     );
 
     // check that there is a current task group for each group of partitions in [partitionGroups]
+    // Fetch metadata offsets once and collect all stale partitions across all groups before committing
+    // any state changes. New TaskGroups are staged locally and only written to activelyReadingTaskGroups
+    // if no reset is required, keeping state consistent: either all new groups are committed or none are.
+    final Map<PartitionIdType, SequenceOffsetType> metadataOffsets = getOffsetsFromMetadataStorage();
```
jtuglu1, line 4105: LMK what you think about this – I think this shouldn't impact task groups without invalid offsets as currently they'd just be added to the activelyReadingTaskGroups then be delayed in processing (see few lines down) since generateStartingSequencesForPartitionGroup would just throw an exception (and force a new runInternal()).

FrankChen021 (commented):

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -4279,17 +4318,8 @@ private OrderedSequenceNumber<SequenceOffsetType> getOffsetFromStorageForPartiti
...
-            );
-            throw new StreamException(
-                new ISE(
-                    "Previous sequenceNumber [%s] is no longer available for partition [%s] - automatically resetting"
-                    + " sequence",
-                    sequence,
-                    partition
-                )
-            );
+            partitionsToReset.put(partition, sequence);
```
maytasm, line 4321: Can we still log which sequenceNumber / which partition is no longer available and that we are automatically resetting

maytasm (approved):

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
maytasm: Ignore this. I missed it above

jtuglu1: We do – that's the point of this change. Throw one big exception instead of N smaller ones that delay the progress of the supervisor. The big exception currently omits sequences (only includes partition #s). I generally never need to look at the partition #s unless supervisor doesn't have auto-reset on, but we can make sure those #s are logged as well.

maytasm: Yep. Missed it in the big exception. sequence number is less useful so fine as-is.

jtuglu1: I updated to throw with sequence #s to avoid loss of information

PR #19432, opened 2026-05-08
feat: Add reverse group lookup for LDAP servers that do not return memberOf

JWuCines:
### Description

When using LDAP servers that do not return the `memberOf` attribute in user search results, group-based authorization denies all requests because Druid cannot determine the user's group memberships. This is a known issue with LDAP servers such as OpenLDAP, where `memberOf` is either not enabled by default (it requires the `memberof` overlay module) or is stored as an operational attribute that Java JNDI cannot retrieve — even when `ldapsearch` returns it correctly.

This PR adds an optional reverse group lookup mechanism to `LDAPCredentialsValidator`. When configured, if the user search does not return a `memberOf` attribute, Druid searches group entries to find which groups contain the user's DN and injects the resolved group DNs as synthetic `memberOf` values into the `SearchResult`. The existing `LDAPRoleProvider` (authorizer) then processes these groups as usual, requiring no changes.

The feature is disabled by default. It activates only when both `groupBaseDn` and `groupSearch` are configured and the user search result lacks `memberOf`.

#### Added reverse group lookup to LDAPCredentialsValidator

Added `enrichWithGroupSearch()`, which performs a reverse LDAP search (e.g., `(uniqueMember=<userDN>)`) against the configured group base DN and puts the resolved group DNs into a synthetic `memberOf` attribute on the user's `SearchResult`. The method:
- Uses `setReturningAttributes(new String[]{"1.1"})` to request only the DN, minimizing response size.
- Escapes the user DN with `encodeForLDAP(userDn, true)` to prevent LDAP filter injection.
- Catches `NamingException` internally and logs an error, so a misconfigured group search does not block user authentication — the user proceeds without group memberships.

Added `hasMemberOfAttribute(SearchResult)` so the reverse lookup runs only when the user search did not already return `memberOf`.

The Jackson constructor accepts the two new `groupBaseDn` and `groupSearch` properties and passes them to `BasicAuthLDAPConfig`. Both are optional, so existing configurations are unaffected.

#### Added group search configuration to BasicAuthLDAPConfig

Added two optional fields (`groupBaseDn`, `groupSearch`) with a backward-compatible constructor that delegates to the new constructor with `null` defaults, plus an `isGroupSearchConfigured()` convenience method that returns true only when both are set.

The constructor fails closed with an `InvalidInput` `DruidException` if `groupSearch` does not contain an unescaped `%s` user DN placeholder. Without this validation, a filter such as `(uniqueMember=*)` (or one using the escaped `%%s`, which `String.format` emits literally) would match every group under `groupBaseDn` and grant all mapped group roles to every authenticated user.

#### Added embedded integration test for reverse group lookup

- `LdapReverseGroupLookupAuthResource`: Extends `LdapAuthResource` to add `groupBaseDn` and `groupSearch` properties to the LDAP credentials validator configuration.
- `BasicAuthLdapReverseGroupLookupTest`: Extends `BasicAuthLdapConfigurationTest` to run the full LDAP auth test suite with reverse group lookup enabled. Verifies that group-based authorization works on OpenLDAP (which does not return `memberOf` by default).
- `ldap-configs/bootstrap.ldif`: The `uid=stateOnlyNoLdapGroup` entry carried the `uid`/`cn`/`sn` of `druid` and relied on OpenLDAP implicitly adding the RDN attribute. It now uses its own `uid`, so that the user search filter `(uid=%s)` unambiguously resolves this user.

#### Updated documentation

- `docs/development/extensions-core/druid-basic-security.md`: Added property reference entries for `groupBaseDn` and `groupSearch`.
- `docs/operations/auth-ldap.md`: Added "Group search reverse lookup configuration" section explaining the problem and configuration. Updated the existing info box to reference both manual role mapping and the new reverse lookup as alternatives.

#### Release note

Added support for LDAP group-based authorization on servers that do not return the `memberOf` attribute (e.g., OpenLDAP). Configure `groupBaseDn` and `groupSearch` on the LDAP credentials validator to enable a reverse group lookup that resolves group memberships automatically.

<hr>

* `BasicAuthLDAPConfig`
* `LDAPCredentialsValidator`
* `LDAPCredentialsValidatorTest`
* `LdapReverseGroupLookupAuthResource`
* `BasicAuthLdapReverseGroupLookupTest`

##### Unit tests added to `LDAPCredentialsValidatorTest`
* `testValidateCredentialsWithGroupSearch`: verifies `memberOf` is populated from the reverse group search when the user search does not return it.
* `testGroupSearchWithoutUserDnPlaceholderIsRejected` and `testGroupSearchWithUserDnPlaceholderIsAccepted`: cover the `groupSearch` filter validation, including escaped `%%s` sequences.

<hr>

This PR has:

- [x] been self-reviewed.


extensions-core/druid-basic-security/src/main/java/org/apache/druid/security/basic/authentication/validator/LDAPCredentialsValidator.java
```diff
@@ -237,22 +243,92 @@ SearchResult getLdapUserObject(BasicAuthLDAPConfig ldapConfig, DirContext contex
...
+        userResult = results.next();
       }
       finally {
         results.close();
       }
+
+      // Some LDAP servers do not return memberOf in search results. When memberOf is
+      // absent and group search configuration is provided, fall back to a reverse
+      // group lookup.
+      if (ldapConfig.isGroupSearchConfigured() && !hasMemberOfAttribute(userResult)) {
```
FrankChen021: [P2] Reverse group lookup runs before password verification

getLdapUserObject now performs the configured subtree group search before validatePassword is called. With reverse lookup enabled, any unauthenticated request for an existing user whose search result lacks memberOf can force this extra LDAP group search even when the password is wrong, amplifying brute-force attempts into expensive LDAP/Druid load. Please do the reverse group lookup only after the user bind succeeds, or split lookup/enrichment so failed credentials do not trigger group enumeration.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

extensions-core/druid-basic-security/src/main/java/org/apache/druid/security/basic/authentication/validator/LDAPCredentialsValidator.java
JWuCines: Moved the reverse group lookup after the password verification: https://github.com/apache/druid/pull/19432/changes/a81ccf777cc8586da25a72cf9119adb566928da6#diff-472b992d587115d723b9c3027bcf078d0ee88bea6d389bb0ec2724b1a0ee8b0fR209

FrankChen021 (commented):
Reviewed 8 of 8 changed files. The reverse group lookup now runs only after password verification, and I found no remaining PR-caused correctness issues in the LDAP reverse lookup changes.

---

JWuCines:
Hey @FrankChen021,
Can you have a review and merge to this PR too?
I would like to have it as part of milestone/version 38.0.0
Thank you!

FrankChen021 (commented):
I have reviewed the updated PR for correctness, edge cases, concurrency, and integration risks; no issues found.

---

JWuCines:
@FrankChen021 can you review and approve it? Can you also merge the PR as reviewer and committer?
Thank you!

JWuCines:
Hi @clintropolis, @gianm and @kfaraz,
I saw your names in PRs and I see you are Member of Apache Druid project.
Can you review and merge this PR, when you have time?
Thank you!

FrankChen021 (commented):
I have reviewed the updated code for correctness, edge cases, concurrency, and integration risks; no issues found. The rebase introduced no PR behavior changes since the previous reviewed SHA.

---

extensions-core/druid-basic-security/src/main/java/org/apache/druid/security/basic/authentication/validator/LDAPCredentialsValidator.java
```diff
@@ -237,22 +247,90 @@ SearchResult getLdapUserObject(BasicAuthLDAPConfig ldapConfig, DirContext contex
...
+    try {
+      Thread.currentThread().setContextClassLoader(this.getClass().getClassLoader());
+      dirContext = new InitialDirContext(bindProperties(this.ldapConfig));
+
+      final String userDn = userResult.getNameInNamespace();
+      final SearchControls sc = new SearchControls();
+      sc.setSearchScope(SearchControls.SUBTREE_SCOPE);
+      sc.setReturningAttributes(new String[]{"1.1"});
+
+      final String filter = StringUtils.format(this.ldapConfig.getGroupSearch(), encodeForLDAP(userDn, true));
```
FrankChen021, line 288: [P1] Require a user-dependent LDAP filter

StringUtils.format leaves the extra argument unused when groupSearch has no %s placeholder. A valid filter such as (uniqueMember=*) therefore searches every group below groupBaseDn and adds all group DNs to memberOf, granting every user all mapped group roles. Validate that groupSearch depends on the user DN or fail closed before performing the lookup.

extensions-core/druid-basic-security/src/main/java/org/apache/druid/security/basic/authentication/validator/LDAPCredentialsValidator.java
```diff
@@ -77,7 +79,9 @@ public LDAPCredentialsValidator(
       @JsonProperty("credentialIterations") Integer credentialIterations,
       @JsonProperty("credentialVerifyDuration") Integer credentialVerifyDuration,
       @JsonProperty("credentialMaxDuration") Integer credentialMaxDuration,
-      @JsonProperty("credentialCacheSize") Integer credentialCacheSize
+      @JsonProperty("credentialCacheSize") Integer credentialCacheSize,
+      @JsonProperty("groupBaseDn") String groupBaseDn,
```
FrankChen021: [P2] Preserve the existing public constructor

Adding groupBaseDn and groupSearch to the only public URL-based constructor removes the previous 10-argument signature. Direct Java callers will fail to compile, and already-compiled integrations can fail with NoSuchMethodError even when reverse group lookup is unused. Keep a delegating overload for the existing signature.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

Reviewed 8 of 8 changed files in the full merge-base diff. The prior review SHA was unavailable, so the full diff was audited.

---

FrankChen021 (changes requested):
AI reports two problems, after my review, I removed one, and left one for you to resolve.

extensions-core/druid-basic-security/src/main/java/org/apache/druid/security/basic/authentication/validator/LDAPCredentialsValidator.java
JWuCines: Added a fix to enforce "%s" placeholder in groupSearch

FrankChen021: Thanks for the follow-up. The constructor now rejects `groupSearch` unless it contains an unescaped `%s` placeholder while allowing escaped `%%`, and the added tests cover both invalid and valid forms. This prevents the previously reported all-groups lookup when the placeholder is omitted. I reviewed all 9 changed files and have no remaining PR-caused findings.

FrankChen021 (commented):
Follow-up handled: the new validation rejects `groupSearch` filters without an unescaped `%s` placeholder, and the added tests cover escaped and valid forms. I reviewed all 9 changed files and found no remaining PR-caused issues.

---

FrankChen021 (approved):

FrankChen021 (commented):

---

PR #19434, opened 2026-05-08
fix: use short name from authToLocal rules as Kerberos authenticated identity

JWuCines:
### Description

The Kerberos authenticator was using `token.getName()` (the full Kerberos principal, e.g., `user@EXAMPLE.COM`) instead of `token.getUserName()` (the short name, e.g., `user`) when constructing the `AuthenticationResult` identity. This caused the authenticated identity to bypass `authToLocal` name mapping rules, resulting in users being unable to authenticate when authorizer rules reference short names.

#### Fixed the identity used in AuthenticationResult

Changed `KerberosAuthenticator` to use `token.getUserName()` instead of `token.getName()` when setting the `AuthenticationResult` identity. This is consistent with:
- How `getRemoteUser()` already returns `token.getUserName()` in the same filter
- How `DruidKerberosAuthenticationHandler` creates the token with `new AuthenticationToken(userName, clientPrincipal, getType())` where `userName` is the result of `KerberosName.getShortName()` (i.e., the `authToLocal`-mapped name)
- The purpose of the `authToLocal` configuration property, which exists to map Kerberos principals to local user names for authorization

#### Added unit tests

Added `KerberosAuthenticatorFilterTest` (4 tests) and `DruidKerberosAuthenticationHandlerTokenTest` (6 tests) to verify that:
- `AuthenticationToken.getUserName()` returns the short name and `getName()` returns the full principal
- `AuthenticationResult` identity uses the short name across simple principals, service principals, and custom `authToLocal` rules

#### Updated documentation

Added a "User identity and `authToLocal` rules" section to the Kerberos extension documentation explaining that the short name is used for authorization, with an upgrade note for users who may have authorizer rules referencing full Kerberos principals.

#### Release note

The Kerberos authenticator now correctly uses the short name derived from `authToLocal` rules as the authenticated user identity, instead of the full Kerberos principal. If you have authorizer rules that reference full Kerberos principals (e.g., `user@EXAMPLE.COM`), update them to use the short name (e.g., `user`) after upgrading.

<hr>

 * `KerberosAuthenticator`
 * `KerberosAuthenticatorFilterTest`
 * `DruidKerberosAuthenticationHandlerTokenTest`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/docs/development/extensions-core/druid-kerberos.md
+++ b/docs/development/extensions-core/druid-kerberos.md
@@ -121,5 +121,19 @@ To access Coordinator/Overlord console from browser you will need to configure y
     1. Configure trusted websites to include `"druid-coordinator-hostname"` and `"druid-overlord-hostname"`
     2. Allow negotiation for the UI website.
 
+## User identity and `authToLocal` rules
+
+The Kerberos authenticator uses the short name produced by `authToLocal` rules as the authenticated user identity.
+For example, if the Kerberos principal is `user@EXAMPLE.COM` and the `authToLocal` rule maps it to `user`, then the
+identity used for authorization is `user`, not the full principal `user@EXAMPLE.COM`.
+
+When configuring authorizer permissions, use the short (local) user name rather than the full Kerberos principal.
+
+:::info
+In previous versions, the authenticated identity was incorrectly set to the full Kerberos principal
+(e.g., `user@EXAMPLE.COM`) instead of the short name (e.g., `user`). If you have authorizer rules that
+reference full Kerberos principals, update them to use the short name after upgrading.
+:::
+
 ## Sending Queries programmatically
 Many HTTP client libraries, such as Apache Commons [HttpComponents](https://hc.apache.org/), already have support for performing SPNEGO authentication. You can use any of the available HTTP client library to communicate with druid cluster.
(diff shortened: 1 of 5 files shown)


FrankChen021 (commented):

---

FrankChen021 (commented):
Reviewed 5 of 5 changed files for correctness, edge cases, security, and integration risks; no issues found.

---

gianm (commented):
Given that the current logic has been in place for so long, and people have built their authorizer rules around it, the new behavior should be opt-in. I suggest a property like `useShortNameIdentity`, which would be specified as `druid.auth.authenticator.<name>.useShortNameIdentity = true`. The default should be `false` so existing behavior isn't changed.

Another thing to look at: the KerberosEscalator `createEscalatedAuthenticationResult` method still unconditionally uses the `internalClientPrincipal`. When `useShortNameIdentity` is set this should possibly also use the short name. Otherwise people will need to create authorization rules for the long name if it's used in an escalator.

With regard to docs, as long as you're updating those, it would be good to also clarify the meaning of `DEFAULT` for `authToLocal`. It's the default behavior but our docs don't make it clear what it means. I gather from [Hadoop docs](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/SecureMode.html#Mapping_from_Kerberos_principals_to_OS_user_accounts) that `DEFAULT` strips the realm from the principal if it is the system's default Kerberos realm, and leaves the principal intact otherwise.

Finally, an embedded test would be useful to validate that everything works end to end. You could look at `BasicAuthLdapConfigurationTest` for an example of a test that uses an OpenLDAP server in a Docker container to test LDAP authentication. Something similar might work with a Kerberos server.

PR #19449, opened 2026-05-11
Column projection added along with test function

Arijit6258:
Fixes #19267 .

### Description

Added column projection to determine which columns to read from iceberg table. This change will help greatly to improve read efficiency for use cases where whole table scan is not intended.

<hr>

 * `IcebergCatalog`
 * `IcebergInputSource`
 * `IcebergInputSourceTest`

<hr>

This PR has:

- [x] been self-reviewed.


extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -88,15 +92,27 @@ public List<String> extractSnapshotDataFiles(
...
+
+      if (columnsFilter != null) {
+        List<String> projectedColumns = table
+            .schema()
+            .columns()
+            .stream()
+            .map(Types.NestedField::name)
+            .filter(columnsFilter::apply)
+            .collect(Collectors.toList());
+        tableScan = tableScan.select(new ArrayList<>(projectedColumns));
```
FrankChen021, line 114: [P2] Projection is discarded before data is read

This selects projected columns on the Iceberg TableScan, but the method only returns task.file().location() afterward. The projected FileScanTask schema is discarded, and IcebergInputSource builds the warehouse delegate from the same raw file paths, so Druid's Parquet reader still opens the original files without the Iceberg projection. The new test also manually projects with Parquet.read(...).project(...), so it would pass even if this select had no effect. To make column projection work, the projected schema/split information needs to be carried into the reader path or pruning needs to happen in the delegate input format.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
gianm: This seems like a real concern. The method only returns a list of files.

FrankChen021 (commented):
Reviewed 3 of 3 changed files. The latest reply acknowledges the original projection concern, and the current code still matches that discussion: the Iceberg TableScan projection is reduced back to raw file locations before the delegate reader path. I did not find any additional high-confidence PR-caused issues beyond the existing thread.

---

PR #19451, opened 2026-05-11
feat: Metrics tracking storage load begin and end.

gianm:
Prior to this patch, the `storage/load/bytes` and `storage/virtual/load/bytes` metrics were measured from the time the space is reserved, which can obscure timings of the actual load. This patch makes two changes:

1) Modify `storage/load/bytes` and `storage/virtual/load/bytes` to measure once the load is complete, rather than when it starts.

2) Introduce `storage/load/begin/bytes` and `storage/virtual/load/begin/bytes` to match what the `storage/load/bytes` and `storage/virtual/load/bytes` were previously defined as.

The same changes are made to the corresponding `count` metrics.

### DIFF
--- a/server/src/main/java/org/apache/druid/segment/loading/StorageLocationStats.java
+++ b/server/src/main/java/org/apache/druid/segment/loading/StorageLocationStats.java
@@ -27,12 +27,25 @@ public interface StorageLocationStats
   long getUsedBytes();
 
   /**
-   * Number of load operations during the measurement period
+   * Number of load operations that were started (space reserved) during the measurement period. This is incremented
+   * when the load begins, regardless of whether it ultimately completes.
+   */
+  long getLoadBeginCount();
+
+  /**
+   * Number of bytes for which a load was started (space reserved) during the measurement period. This is incremented
+   * when the load begins, regardless of whether it ultimately completes.
+   */
+  long getLoadBeginBytes();
+
+  /**
+   * Number of load operations that finished (segment downloaded, deserialized, and made queryable) during the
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/druid/segment/loading/StorageLocationStats.java
```diff
@@ -27,12 +27,25 @@ public interface StorageLocationStats
   long getUsedBytes();
 
   /**
-   * Number of load operations during the measurement period
+   * Number of load operations that were started (space reserved) during the measurement period. This is incremented
+   * when the load begins, regardless of whether it ultimately completes.
+   */
+  long getLoadBeginCount();
```
FrankChen021, line 33: [P2] Keep storage stats interfaces source compatible

This adds new abstract methods to a public stats interface returned by SegmentCacheManager.getStorageStats(). Any extension or custom SegmentCacheManager that provides its own StorageLocationStats implementation will now fail to compile, or can hit AbstractMethodError when StorageMonitor invokes the new methods. Since implementations without begin-metric data can report zero, make these default methods returning 0 and apply the same compatibility treatment to VirtualStorageLocationStats.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/segment/loading/StorageLocationStats.java
gianm: This isn't meant to be a stable interface.

clintropolis (approved):

PR #19454, opened 2026-05-12
feat: configurable transport compression strategy

jtuglu1:
Fixes #XXXX.

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/client/DirectDruidClient.java
+++ b/server/src/main/java/org/apache/druid/client/DirectDruidClient.java
@@ -23,10 +23,12 @@ import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.smile.SmileFactory;
 import com.fasterxml.jackson.jaxrs.smile.SmileMediaTypes;
+import com.github.luben.zstd.ZstdInputStream;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import net.jpountz.lz4.LZ4BlockInputStream;
 import org.apache.druid.java.util.common.RE;
 import org.apache.druid.java.util.common.StringUtils;
 import org.apache.druid.java.util.common.concurrent.Execs;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/druid/client/DirectDruidClient.java
```diff
@@ -516,7 +521,7 @@ public JsonParserIterator<T> make()
           {
             return new JsonParserIterator<>(
                 queryResultType,
-                future,
+                wrapFutureWithDecompressor(future, responseContentEncoding.get()),
```
FrankChen021, line 524: [P1] Decompressor choice races the response headers

The decompressor is selected from responseContentEncoding.get() when the Sequence iterator is created, which can happen immediately after run() returns and before the async HTTP handler has seen the response headers. In that common path this passes an empty encoding and returns the original future, so a later zstd/x-lz4 response is handed to JsonParserIterator still compressed and parsing fails. Defer the encoding lookup until the future completes, e.g. always transform with a function that reads the AtomicReference after handleResponse has set it.

server/src/main/java/org/apache/druid/server/QueryResultPusher.java
```diff
@@ -385,6 +387,36 @@ public interface Writer extends Closeable
...
+   * Picks the best compression encoding the client accepts. Prefers zstd over x-lz4.
+   * Returns null if no supported encoding is advertised.
+   */
+  @Nullable
+  private static String negotiateContentEncoding(@Nullable String acceptEncoding)
+  {
+    if (acceptEncoding == null) {
+      return null;
+    }
+    if (acceptEncoding.contains("zstd")) {
```
FrankChen021, line 400: [P2] Honor q=0 when negotiating encodings

This substring check treats headers like `gzip, zstd;q=0` or `x-lz4;q=0` as accepting the encoding, then sends a compressed response that the client explicitly marked unacceptable. QueryResultPusher handles user-facing query responses too, so this can break standards-compliant clients. Parse Accept-Encoding as tokens with q-values and only select zstd/x-lz4 when q is absent or greater than zero.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

---

PR #19457, opened 2026-05-12
refactor: Make QueryBlocklistRule an interface for extensibility

abhishekrb19:
Summary

  - Converts `QueryBlocklistRule` from a concrete class to an interface with two methods: `getRuleName()` and `matches(Query<?>)`.
  - Moves the existing implementation to `DefaultQueryBlocklistRule` with a default impl on the interface `@JsonTypeInfo(defaultImpl = DefaultQueryBlocklistRule.class)` so existing JSON configs without a "type" field continue to deserialize correctly.
  - Extensions can register new rule types as Jackson subtypes via SimpleModule.registerSubtypes(...) and we can even introduce new union rule types with different logic w/o polluting existing default implementation.

  **Motivation**

  The current blocklist logic is fixed to datasource/queryType/context matching. Making it an interface allows adding new rule types (e.g., regex-based, time-window-based, user-based) without modifying the existing code.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/docs/api-reference/dynamic-configuration-api.md
+++ b/docs/api-reference/dynamic-configuration-api.md
@@ -534,7 +534,9 @@ The following table shows the dynamic configuration properties for the Broker.
 
 Query blocklist rules allow you to block specific queries based on datasource, query type, and/or query context parameters. This feature is useful for preventing expensive or problematic queries from impacting cluster performance.
 
-Each rule in the `queryBlocklist` array is a JSON object. The `type` field selects the rule implementation. If omitted, it defaults to `"default"`.
+Each rule in the `queryBlocklist` array is a JSON object. The `type` field selects the rule implementation. If omitted, it defaults to `"default"`. A query is blocked if it matches ANY rule in the blocklist (OR logic between rules).
+
+> **Note:** The `"type"` field is not required for `"default"` rules (existing configs without it continue to work), but including it explicitly is recommended for clarity.
 
 ##### `default` type
 
@@ -548,6 +550,13 @@ The built-in rule type. Blocks queries by matching on datasource, query type, an
 |`queryTypes`|List of query types to match (e.g., `scan`, `timeseries`, `groupBy`, `topN`). A query matches if its type is in this list.|No|Matches all query types|
 |`contextMatches`|Map of query context parameter key-value pairs to match. A query matches if all specified context parameters match the provided values (case-sensitive string comparison).|No|Matches all contexts|
 
+**Matching behavior:**
+
+- A query must match ALL specified criteria within a rule (AND logic) to be blocked by that rule
+- If any criterion is omitted, empty, or null, it matches everything (e.g., omitting `queryTypes` or setting it to null matches all query types)
+- For context matching: if a rule specifies context parameters, queries with missing or null values for those keys will not match
+- At least one criterion must be specified per rule to prevent accidentally blocking all queries
+
 ##### Custom types (extensions)
 
 Extensions can register additional rule types by adding Jackson subtypes to the `QueryBlocklistRule` interface. A custom rule is selected by setting `"type"` to its registered name:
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/druid/server/QueryBlocklistRule.java
```diff
@@ -19,158 +19,32 @@
...
+ *
+ * <p>Rules with no {@code "type"} field in JSON deserialize as {@link DefaultQueryBlocklistRule}
+ * for backwards compatibility. Extensions can register additional implementations as Jackson
+ * subtypes via {@code SimpleModule.registerSubtypes(...)}.
+ *
+ * <p>Implementations must define {@code equals} and {@code hashCode} so that
+ * {@link org.apache.druid.server.broker.BrokerDynamicConfig} can detect changes correctly.
  */
-public class QueryBlocklistRule
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type", defaultImpl = DefaultQueryBlocklistRule.class)
```
FrankChen021: [P2] Fail unknown blocklist rule types instead of defaulting

Using `defaultImpl = DefaultQueryBlocklistRule.class` makes Jackson use the default rule not only when `type` is absent, but also when a non-empty type id cannot be resolved. For the new extension point, a broker or coordinator that lacks a registered custom rule subtype can silently deserialize that rule as `DefaultQueryBlocklistRule`; with Druid's unknown-property-tolerant mapper, extension-specific fields are ignored and any standard fields such as `dataSources` can turn into a broader block than intended. Please distinguish the legacy missing-type case from an unrecognized explicit `type`, so unknown extension rules fail fast rather than being interpreted as default rules.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/server/QueryBlocklistRule.java
abhishekrb19: Addressed in this [commit](https://github.com/apache/druid/pull/19457/commits/10e753c9bff9217fe0c4a406e9b46f2681c2338f) by adding a `StrictTypeIdResolver` similar to `DimensionSchema`

jtuglu1:
@mshahid6 want to take a look here?

FrankChen021 (commented):
I rechecked the follow-up on `QueryBlocklistRule` type-id handling. The latest change uses `StrictTypeIdResolver`, preserves legacy missing-type defaulting, and adds coverage for missing, explicit default, and unknown explicit types. This addresses the previous concern; no inline reply is needed.

---

mshahid6:
LGTM, do we need to update the docs too?

abhishekrb19:
Thanks @mshahid6! Given that this is already backward [compatible](https://druid.apache.org/docs/latest/api-reference/dynamic-configuration-api/#query-blocklist-rules) without type, I  was thinking we can leave the docs as-is for now and include some notes about the `default` type when we introduce more block list rule types. What do you think?

abhishekrb19:
I went ahead and added some doc updates for completeness - ptal

docs/api-reference/dynamic-configuration-api.md
```diff
@@ -529,15 +534,34 @@ The following table shows the dynamic configuration properties for the Broker.
...
 |--------|-----------|--------|-------|
+|`type`|Rule type identifier. Not required — rules without a `"type"` field are treated as `"default"` for backwards compatibility with existing configurations.|No|`"default"`|
 |`ruleName`|Unique name identifying this blocklist rule. Used in error messages when queries are blocked.|Yes|N/A|
 |`dataSources`|List of datasource names to match. A query matches if it references any datasource in this list.|No|Matches all datasources|
 |`queryTypes`|List of query types to match (e.g., `scan`, `timeseries`, `groupBy`, `topN`). A query matches if its type is in this list.|No|Matches all query types|
 |`contextMatches`|Map of query context parameter key-value pairs to match. A query matches if all specified context parameters match the provided values (case-sensitive string comparison).|No|Matches all contexts|
 
+##### Custom types (extensions)
+
+Extensions can register additional rule types by adding Jackson subtypes to the `QueryBlocklistRule` interface. A custom rule is selected by setting `"type"` to its registered name:
```
FrankChen021, line 562: [P3] Scope the matching semantics to the default rule type

After introducing custom `QueryBlocklistRule` types, the following `Rule matching behavior` section still reads as if every blocklist rule must use datasource/query-type/context criteria with AND semantics and at least one of those criteria. The interface now lets extensions define arbitrary `matches` logic and fields, so this is only guaranteed for the built-in `default` rule. Please scope those bullets to the `default` type or add wording that custom types define their own matching behavior.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

---

docs/api-reference/dynamic-configuration-api.md
abhishekrb19: Thanks - updated the docs to clarify this in the latest commit.

FrankChen021 (commented):
I rechecked the follow-up on the docs scoping comment. The latest docs move the default-rule matching bullets under the `default` type and explicitly state that custom types define their own matching semantics, so no inline reply is needed.

---

jtuglu1 (approved):
LGTM

PR #19459, opened 2026-05-13
fix: Make `sys.server_properties` table filterable and resilient to unreachable servers

abhishekrb19:
Summary:
  1. Made `sys.server_properties` filterable, similar to a few other sys tables like `SegmentsTable` — changed from `ScannableTable` to `ProjectableFilterableTable`, enabling Calcite to push down filters (e.g., WHERE server = '...') and projections to the table scan.
  2. Graceful error handling — Previously, if any server was unreachable or returned an HTTP error, the entire query threw an exception and failed. Now the table returns a row for that server with property/value as null and the new
  error_message column populated with what went wrong.
  3. New `error_message` column — this column indicates if properties couldn't be fetched (e.g., "Connection refused", "HTTP 503: Service Unavailable"). Null on success.

Release note:

Added `error_message` column to `sys.server_properties table` and made the table resilient to unreachable servers. Previously, the entire query would fail if any server was unreachable; now a row is returned with `error_message` populated. The table also now supports filter and projection pushdown.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/docs/querying/sql-metadata-tables.md
+++ b/docs/querying/sql-metadata-tables.md
@@ -322,7 +322,7 @@ SELECT * FROM sys.supervisors WHERE healthy=0;
 
 ### SERVER_PROPERTIES table
 
-The `server_properties` table exposes the runtime properties configured on for each Druid server. Each row represents a single property key-value pair associated with a specific server.
+The `server_properties` table exposes the runtime properties configured on each Druid server. Each row represents a single property key-value pair associated with a specific server. This table supports filter and projection pushdown for efficient querying. If a server is unreachable, the table still returns a row for that server with the `error` column populated instead of failing the entire query.
 
 |Column|Type|Notes|
 |------|-----|-----|
@@ -331,6 +331,7 @@ The `server_properties` table exposes the runtime properties configured on for e
 |node_roles|VARCHAR|Comma-separated list of roles that the server performs. For example, `[coordinator,overlord]` if the server functions as both a Coordinator and an Overlord.|
 |property|VARCHAR|Name of the property|
 |value|VARCHAR|Value of the property|
+|error_message|VARCHAR|Describes why properties could not be retrieved from the server (e.g., connection refused, HTTP error). Null when properties were fetched successfully.|
 
 For example, to retrieve properties for a specific server, use the query
 
--- a/docs/querying/sql-metadata-tables.md
+++ b/docs/querying/sql-metadata-tables.md
@@ -322,7 +322,7 @@ SELECT * FROM sys.supervisors WHERE healthy=0;
 
 ### SERVER_PROPERTIES table
 
-The `server_properties` table exposes the runtime properties configured on each Druid server. Each row represents a single property key-value pair associated with a specific server. This table supports filter and projection pushdown for efficient querying. If a server is unreachable, the table still returns a row for that server with the `error` column populated instead of failing the entire query.
+The `server_properties` table exposes the runtime properties configured on each Druid server. Each row represents a single property key-value pair associated with a specific server. This table supports filter and projection pushdown for efficient querying. If a server is unreachable, the table still returns a row for that server with the `error_message` column populated instead of failing the entire query.
 
 |Column|Type|Notes|
 |------|-----|-----|
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/schema/SystemServerPropertiesTableTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/schema/SystemServerPropertiesTableTest.java
@@ -119,14 +119,14 @@ public class SystemServerPropertiesTableTest extends EmbeddedClusterTestBase
     );
 
     Assertions.assertEquals(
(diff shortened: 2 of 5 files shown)


docs/querying/sql-metadata-tables.md
```diff
@@ -322,7 +322,7 @@ SELECT * FROM sys.supervisors WHERE healthy=0;
 
 ### SERVER_PROPERTIES table
 
-The `server_properties` table exposes the runtime properties configured on for each Druid server. Each row represents a single property key-value pair associated with a specific server.
+The `server_properties` table exposes the runtime properties configured on each Druid server. Each row represents a single property key-value pair associated with a specific server. This table supports filter and projection pushdown for efficient querying. If a server is unreachable, the table still returns a row for that server with the `error` column populated instead of failing the entire query.
```
GWphua:
```suggestion
The `server_properties` table exposes the runtime properties configured on each Druid server. Each row represents a single property key-value pair associated with a specific server. This table supports filter and projection pushdown for efficient querying. If a server is unreachable, the table still returns a row for that server with the `error_message` column populated instead of failing the entire query.
```

embedded-tests/src/test/java/org/apache/druid/testing/embedded/schema/SystemServerPropertiesTableTest.java
```diff
@@ -146,6 +146,13 @@ public void test_serverPropertiesTable_hiddenProperties()
     Assertions.assertFalse(brokerProps.containsKey("password"));
   }
 
+  @Test
+  public void test_serverPropertiesTable_errorMessageColumnExists()
+  {
+    final String result = cluster.runSql("SELECT COUNT(*) FROM sys.server_properties WHERE error_message IS NULL");
+    Assertions.assertFalse(result.isEmpty(), "Should return count of servers with null error_message");
```
GWphua: nit: I think this test just checks if the query did not fail. Can we make the test stronger? For example, asserting a return of a known value (e.g. 0,1,etc.)

GWphua (approved):
LGTM with non-blocking comments.

GWphua:
If this is not urgent, we can wait a day for the AI-assisted review to trigger.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemServerPropertiesTable.java
```diff
@@ -108,42 +121,62 @@ public Schema.TableType getJdbcTableType()
   }
 
   @Override
-  public Enumerable<Object[]> scan(DataContext root)
+  public Enumerable<Object[]> scan(
+      final DataContext root,
+      final List<RexNode> filters,
```
FrankChen021, line 132: [P2] Actually apply pushed filters before fetching every server

The new ProjectableFilterableTable scan receives Calcite filters, but this method never inspects or removes any accepted filters. Calcite will apply the remaining filters after the scan, so queries such as `WHERE server = 'host:port'` still enumerate all discovered servers and call `/status/properties` for each one. That means an unrelated slow or unreachable node can still delay a query that should have been scoped to one server, and the advertised filter pushdown is not actually implemented. Please parse and consume supported filters, at least `server` and likely `service_name`/`property`, before fetching properties.

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemServerPropertiesTable.java
```diff
@@ -154,20 +187,34 @@ private Map<String, String> getProperties(DruidNode druidNode)
...
       }
-      return jsonMapper.readValue(
-          response.getContent(),
-          new TypeReference<>(){}
+      return new PropertiesResult(
+          jsonMapper.readValue(response.getContent(), new TypeReference<>(){}),
+          null
       );
     }
     catch (Exception e) {
```
FrankChen021, line 275: [P2] Do not turn interrupts into error rows

`Future.get()` can throw `InterruptedException`, but the broad catch now converts it into a synthetic `error_message` row and clears the interrupt status. If the SQL query is cancelled or the execution thread is interrupted during a properties fetch, the scan can keep iterating and return partial table data instead of aborting. Handle `InterruptedException` separately by re-interrupting the thread and propagating cancellation rather than treating it as an unreachable Druid server.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

---

sql/src/main/java/org/apache/druid/sql/calcite/schema/SystemServerPropertiesTable.java
abhishekrb19: Yeah, the initial version had some filter pushdown that was slightly complex. Currently, it supports equality filters on the `server` and `service_name` columns. It falls back to querying all servers if it detects any other filter predicate or column.

FrankChen021 (commented):
I have reviewed the follow-up on the `sys.server_properties` filter pushdown thread. The current code now applies `server` and `service_name` equality filters before fetching properties, so the original concern is handled.

---

PR #17214, opened 2024-10-01
Projections prototype

clintropolis:
### Description
#17117 + some refactors + projections persisted segments = possibly usable prototype

_todo_

```
SELECT string2, APPROX_COUNT_DISTINCT_DS_HLL("string5") FROM "druid"."projections" GROUP BY 1 ORDER BY 2
SELECT string2, SUM(long4) FROM "druid"."projections" GROUP BY 1 ORDER BY 2
```

realtime segments:
```
Benchmark                         (complexCompression)  (query)  (rowsPerSegment)  (schema)  (storageType)  (stringEncoding)  (useProjections)  (vectorize)  Mode  Cnt   Score   Error  Units
SqlProjectionsBenchmark.querySql                  none        0            150000  explicit    INCREMENTAL              UTF8              true        false  avgt    5   6.805 ± 0.928  ms/op
SqlProjectionsBenchmark.querySql                  none        0            150000  explicit    INCREMENTAL              UTF8             false        false  avgt    5  25.313 ± 2.134  ms/op
SqlProjectionsBenchmark.querySql                  none        1            150000  explicit    INCREMENTAL              UTF8              true        false  avgt    5   2.058 ± 0.429  ms/op
SqlProjectionsBenchmark.querySql                  none        1            150000  explicit    INCREMENTAL              UTF8             false        false  avgt    5  22.030 ± 2.828  ms/op
```

historical segments:
```
Benchmark                         (complexCompression)  (query)  (rowsPerSegment)  (schema)  (storageType)  (stringEncoding)  (useProjections)  (vectorize)  Mode  Cnt    Score    Error  Units
SqlProjectionsBenchmark.querySql                   lz4        0           1500000  explicit           MMAP              UTF8              true        false  avgt    5   18.742 ±  0.959  ms/op
SqlProjectionsBenchmark.querySql                   lz4        0           1500000  explicit           MMAP              UTF8              true        force  avgt    5   18.638 ±  0.775  ms/op
SqlProjectionsBenchmark.querySql                   lz4        0           1500000  explicit           MMAP              UTF8             false        false  avgt    5  180.287 ± 11.416  ms/op
SqlProjectionsBenchmark.querySql                   lz4        0           1500000  explicit           MMAP              UTF8             false        force  avgt    5  138.844 ± 10.884  ms/op
SqlProjectionsBenchmark.querySql                   lz4        1           1500000  explicit           MMAP              UTF8              true        false  avgt    5    1.764 ±  0.271  ms/op
[...]

PR #19460, opened 2026-05-13
feat: clustered segments pt.1

clintropolis:
### Description
This PR builds on the foundations laid by projections (#17214) and the v10 segment format (#18880) to introduce 'clustered' segments, to give operators the option to push a `CLUSTERED BY` clause _inside of a segment_, as a companion to partitioning data is distributed between segments in this manner. Internally, the 'base' table is decmposed into separate cluster groups, which are combined together to form the 'complete' view of all rows stored in the segment via concatenation. This optimizes for use cases where the typical queries are filtering down to a small subset of the cluster groups (ideally a single grouping), which like the effect from using aggregate projections, can greatly reduce the number of rows to be scanned. The expected use cases are things like multi-tenant-with-shared datasource clusters, metrics use cases which typically filter to a single type of service, etc

This PR contains only the read side, to get feedback on the internal segment metadata shapes and query engine integration. The write side (ingestion support) will come in a follow-up PR, so this PR uses some test fixtures to exercise the read paths until segment building is actually in place.

Since this is an experimental/mostly new feature, the most important part for reviewers is the new internal segment metadata, `ClusteredValueGroupsBaseTableSchema` and its internal new stuff like `TableClusterGroupSpec` and `ClusteringDictionaries`, so that we can ensure the metadata we will be storing in the segment is "good" since changing it after segments have been written is very hard.

_todo: elaborate on design_


processing/src/main/java/org/apache/druid/segment/QueryableIndexCursorFactory.java
```diff
@@ -110,9 +143,328 @@ public List<AggregatorFactory> getAggregatorsForPreAggregated()
...
+  {
+    final QueryableIndex groupIndex = index.getClusterGroupQueryableIndex(valueGroup);
+    if (groupIndex == null) {
+      throw DruidException.defensive(
+          "No cluster-group sub-index resolvable for clustering values "
+          + Arrays.toString(valueGroup.lookupClusteringValues())
+      );
+    }
+
+    return new QueryableIndexCursorHolder(
```
FrankChen021, line 206: [P1] Cluster-column filters are reapplied against sub-indexes that do not contain them

After pruneClusterGroups keeps a matching group, the original CursorBuildSpec is passed unchanged into a QueryableIndexCursorHolder for the group sub-index. That sub-index intentionally does not contain the clustering columns, so EqualityFilter.getBitmapColumnIndex sees a missing column and builds an exact all-false/all-unknown bitmap before the ClusteringColumnSelectorFactory wrapper can provide the constant value to a matcher. Queries such as tenant = 'acme' therefore prune to the acme group and then scan zero rows. The multi-group supplier path repeats the same pattern, so clustering predicates need to be stripped/residualized after pruning or the group holder needs a cluster-aware ColumnIndexSelector/filter-bundle path.

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -530,6 +548,241 @@ public static String getProjectionSegmentInternalFilePrefix(ProjectionSchema pro
...
+    if (filter instanceof EqualityFilter eq) {
+      final int idx = clusteringColumns.indexOf(eq.getColumn());
+      if (idx < 0) {
+        return DruidPredicateMatch.UNKNOWN;
+      }
+      // EqualityFilter doesn't match nulls; constructor also rejects null match values.
+      if (clusteringValues[idx] == null) {
+        return DruidPredicateMatch.FALSE;
+      }
+      return DruidPredicateMatch.of(Objects.equals(clusteringValues[idx], eq.getMatchValue()));
```
FrankChen021: [P1] Pruning ignores EqualityFilter type coercion semantics

The pruner compares the stored clustering value directly with eq.getMatchValue(), but EqualityFilter normal matcher casts the literal to the input column type. For example, a LONG clustering value 5 with an EqualityFilter typed as STRING value "5" would match when evaluated by the real matcher, but Objects.equals(5L, "5") returns false here and the only live cluster group is removed before scanning. TypedInFilter has the same risk below with direct Objects.equals over the raw sorted values. The pruning check should use the same predicate/cast semantics as the filter matcher, or stay UNKNOWN when the literal type does not exactly match the clustering column type.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 0 |
| P3 | 0 |
| Total | 2 |

---

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 1 |
| P3 | 0 |
| Total | 3 |

Reviewed 27 of 27 changed files.

---

processing/src/main/java/org/apache/druid/segment/IndexIO.java
```diff
@@ -1080,6 +1126,63 @@ private Map<String, Supplier<BaseColumnHolder>> readProjectionColumns(
...
+    {
+      final TableClusterGroupSpec spec = summary.getClusterGroups().get(groupIndex);
+      final List<Integer> clusteringValueIds = spec.getClusteringValueIds();
+      final String timeColumnName = summary.getTimeColumnName();
+      final boolean renameTime = !ColumnHolder.TIME_COLUMN_NAME.equals(timeColumnName);
+      final Map<String, Supplier<BaseColumnHolder>> groupColumns = new LinkedHashMap<>();
+
+      for (String column : summary.getGroupColumnNames()) {
+        final String smooshName = Projections.getClusterGroupSegmentInternalFileName(clusteringValueIds, column);
+        final ByteBuffer colBuffer = segmentFileMapper.mapFile(smooshName);
```
FrankChen021: [P1] Check descriptor before mapping sparse group column

The new clustered-group reader maps the smoosh file before checking whether the column descriptor exists. For sparse clustered groups, a logical column can be present in the summary but absent from an individual group, and the existing null-descriptor check suggests that case is expected. As written, `mapFile(smooshName)` fails before the code can skip the missing column, so loading such clustered segments throws instead of exposing the column as absent/null for that group.

processing/src/main/java/org/apache/druid/segment/QueryableIndexCursorFactory.java
```diff
@@ -110,9 +145,354 @@ public List<AggregatorFactory> getAggregatorsForPreAggregated()
...
+  {
+    static final EmptyClusteredCursorHolder INSTANCE = new EmptyClusteredCursorHolder();
+
+    @Override
+    public Cursor asCursor()
+    {
+      return new Cursor()
+      {
+        @Override
+        public ColumnSelectorFactory getColumnSelectorFactory()
```
FrankChen021, line 449: [P1] Empty pruned cursor must expose selectors

When all cluster groups are pruned, this holder returns a done cursor whose `getColumnSelectorFactory()` always throws. Query engines commonly obtain selector factories immediately after creating a non-null cursor, even when the cursor is already done, so filters such as `tenant = 'missing'` can fail the query instead of returning an empty result. The empty cursor should either return null from `asCursor()` or provide a harmless selector factory.

processing/src/main/java/org/apache/druid/segment/QueryableIndexCursorFactory.java
```diff
@@ -110,9 +145,354 @@ public List<AggregatorFactory> getAggregatorsForPreAggregated()
...
+    );
+    final ConcatenatingVectorCursor vectorCursor = new ConcatenatingVectorCursor(
+        holderSuppliers,
+        clusteringValuesByGroup,
+        vectorWrapperFactory
+    );
+
+    // canVectorize() is determined by the per-group holders. Probe the first one (lazily, `Suppliers.memoize`
+    // means this opens it once and is reused by ConcatenatingVectorCursor).
+    final boolean canVectorize = holderSuppliers.get(0).get().canVectorize();
```
FrankChen021: [P2] Vectorization decision only checks first group

The multi-group clustered holder reports `canVectorize()` based only on the first per-group holder, but each group gets a different rewritten filter. A first group whose filter folds to `TrueFilter` can vectorize while a later group's residual filter cannot; the engine will choose the vector path and `ConcatenatingVectorCursor` will later call `asVectorCursor()` on the non-vectorizable holder, which throws. This should require all surviving group holders to vectorize, or otherwise disable vectorization for the combined holder.

processing/src/main/java/org/apache/druid/segment/ConcatenatingCursor.java
```diff
@@ -0,0 +1,152 @@
...
+  {
+    if (initialized) {
+      return;
+    }
+    initialized = true;
+    advanceToNextNonEmptyGroup();
+  }
+
+  /**
+   * Open the next group whose cursor has at least one row. Sets {@code currentCursor = null} when all groups are
```
cecemei, line 81: nit: advance to the next group sounds slightly better and more consistent than open the next group.
should this reset delegate as well if all groups are exhausted?

processing/src/main/java/org/apache/druid/segment/ConcatenatingCursor.java
```diff
@@ -0,0 +1,152 @@
...
+    return wrapperFactory;
+  }
+
+  @Override
+  public void advance()
+  {
+    initializeIfNeeded();
+    if (currentCursor == null) {
+      return;
+    }
```
cecemei: nit: can reuse `isDone()`

processing/src/main/java/org/apache/druid/segment/ConcatenatingCursor.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  @Override
+  public void advanceUninterruptibly()
+  {
+    initializeIfNeeded();
+    if (currentCursor == null) {
+      return;
+    }
```
cecemei: same here, can reuse `isDone()`

processing/src/main/java/org/apache/druid/segment/projections/ClusteringColumnSelectorFactory.java
```diff
@@ -0,0 +1,566 @@
...
+    this.clusteringColumns = clusteringColumns;
+    setDelegate(delegate, clusteringValues);
+  }
+
+  /**
+   * Update the underlying factory and the constant values for the current cluster group. Called by a multi-group
+   * concatenating cursor on each group transition. Selectors previously returned by this factory will, on their next
+   * invocation, observe the updated state; see the per-call indirection in the inner selector classes.
+   */
+  public void setDelegate(ColumnSelectorFactory delegate, Object[] clusteringValues)
```
cecemei, line 75: I find how `setDelegate` and `generation` usage hard to reason. It felt like we should have all `CursorHolder` in this class and let it manage the advance of cursor, generation, ColumnSelectorFactory, along with the ValueMatcher all together in this class. and `ClusteringColumnSelectorFactory` can implement `Iterable<Cursor>`, WDYT?

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -530,6 +547,243 @@ public static String getProjectionSegmentInternalFilePrefix(ProjectionSchema pro
...
+    // clusteringColumns + groupVcs are loop-invariant, only the per-group clustering tuple changes.
+    final ClusteredValueGroupsBaseTableSchema summary = groups.getFirst().getSummary();
+    final RowSignature clusteringColumns = summary.getClusteringColumns();
+    final VirtualColumns groupVcs = summary.getVirtualColumns();
+
+    // Single walk per group: produces the rewritten filter, and a top-level FalseFilter means the group prunes.
+    // Cache the rewrite for every group (including pruned ones, where it's FalseFilter) so rewriteFor doesn't
+    // re-walk for either the cursor factory or callers that want to inspect a pruned group's outcome directly.
+    final List<TableClusterGroupSpec> kept = new ArrayList<>(groups.size());
+    final IdentityHashMap<TableClusterGroupSpec, Filter> rewriteCache = new IdentityHashMap<>();
```
cecemei, line 620: it felt like `ClusterGroupQueryPlan` doesnt have much more than this hash map, is there any other reason that we want to have a class wrap it?

processing/src/main/java/org/apache/druid/segment/QueryableIndexCursorFactory.java
```diff
@@ -110,9 +145,354 @@ public List<AggregatorFactory> getAggregatorsForPreAggregated()
...
+    // Initial wrapper state uses the first group's clustering values + a throwing placeholder delegate. The
+    // ConcatenatingCursor immediately calls setDelegate on init (before any selector is exposed). The vector
+    // wrapper carries the query-level max vector size from the build spec, the placeholder delegate can't be
+    // queried for sizing, and the value is constant across groups anyway.
+    final int vectorSize = spec.getQueryContext().getVectorSize();
+    final ClusteringColumnSelectorFactory wrapperFactory = new ClusteringColumnSelectorFactory(
+        UNINITIALIZED_DELEGATE,
+        clusteringColumns,
+        clusteringValuesByGroup.get(0)
+    );
```
cecemei, line 291: i find it confusing when the delegate is not initialized but clusteringValues are.... maybe we could set `ClusteringColumnSelectorFactory` as nullable in `ConcatenatingCursor`

processing/src/main/java/org/apache/druid/segment/IndexIO.java
```diff
@@ -983,31 +1009,51 @@ public QueryableIndex load(File inDir, ObjectMapper mapper, boolean lazy, Segmen
...
+        aggProjections.add(
+            new AggregateProjectionMetadata(
+                (AggregateProjectionSchema) projectionSpec.getSchema(),
+                projectionSpec.getNumRows()
+            )
+        );
+      }
+
+      final List<Map<String, Supplier<BaseColumnHolder>>> clusterGroupColumnsList;
+      if (isClusteredSummary) {
```
cecemei, line 1021: nit: maybe we could just check null for clusteredBaseSummary, and it would not require this additional boolean variable

processing/src/main/java/org/apache/druid/segment/projections/ClusteredValueGroupsBaseTableSchema.java
```diff
@@ -0,0 +1,410 @@
...
+    // deferring all summary-dependent state on the spec to setSummary, which we invoke here once the summary's
+    // own state is populated.
+    for (TableClusterGroupSpec spec : this.clusterGroups) {
+      spec.setSummary(this);
+    }
+  }
+
+  @JsonIgnore
+  @Override
+  public List<String> getColumnNames()
```
cecemei, line 186: the `Override` doesn't seem necessary? also this function name is very similar to getColumns() except this one has the metrics, maybe we should make them a bit more distinguishable?

processing/src/main/java/org/apache/druid/segment/projections/ClusteredValueGroupsBaseTableSchema.java
```diff
@@ -0,0 +1,410 @@
...
+  private final List<String> sharedColumns;
+  private final ClusteringDictionaries clusteringDictionaries;
+  private final List<TableClusterGroupSpec> clusterGroups;
+
+  // computed
+  private final int timeColumnPosition;
+  private final Granularity effectiveGranularity;
+
+  @JsonCreator
+  public ClusteredValueGroupsBaseTableSchema(
```
cecemei, line 75: while reading this file, i find myself mumbling words like clustering/ group too much, some personal preference for naming just for inspiration:

clusteringColumns -> groupKeySchema
clusteringDictionaries → groupKeyDictionaries
clusterGroups -> groupSpecs
getGroupColumnNames -> getGroupInternalColumnNames
getGroupDimensionNames -> getGroupInternalDimensionNames
getGroupOrdering -> getGroupInternalOrdering

processing/src/main/java/org/apache/druid/segment/IndexIO.java
```diff
@@ -952,16 +954,33 @@ public QueryableIndex load(File inDir, ObjectMapper mapper, boolean lazy, Segmen
...
-      );
+
+      // For clustered base tables the columns are always accessed through cluster groups, skip reading any base
+      // columns so we don't try to map files that don't exist
+      final boolean isClusteredSummary = baseSchema instanceof ClusteredValueGroupsBaseTableSchema;
+      final ClusteredValueGroupsBaseTableSchema clusteredBaseSummary;
+      final Map<String, Supplier<BaseColumnHolder>> baseColumns;
+      if (isClusteredSummary) {
+        clusteredBaseSummary = (ClusteredValueGroupsBaseTableSchema) baseSchema;
+        if (clusteredBaseSummary.getSharedColumns().isEmpty()) {
```
cecemei, line 965: i dont see how shared columns could be used in this PR, maybe we dont include it?

processing/src/main/java/org/apache/druid/segment/ConcatenatingCursor.java
clintropolis: eh, i feel like 'open' is more honest to what is happening though (at least on the first pass) since the supplier is computing the cursor holders on the fly on the first get as we advance through. with regards to the delegate, we don't want to nuke it because it still needs to answer stuff like column capabilities that things could be using even after a cursor is done (since those metadata methods just delegate to whatever the current group is even though they are largely identical across groups)

processing/src/main/java/org/apache/druid/segment/projections/ClusteringColumnSelectorFactory.java
clintropolis: hmm, all of those things 'belong' to the outer `ConcatenatingCursor` that provides this selector factory though, so what you're suggesting sounds strange to me since i think its basically that we have something that is both ColumnSelectorFactory and Cursor at the same time (even if a thin wrapper is moving the iterable of cursors forward or whatever as the actual cursor). To me at least, that seems a lot more complicated, since it would mean there is a lot going on in the selector factory. We would still need these methods i think anyway, because the cursor is what is driven by the engine so it would need to propagate to the selector factory what the current group is, even if it was providing the cursor the iterable instead of the iterable just providing it with the current thing.

The way it is currently `ClusteringColumnSelectorFactory` is a type of 'settable' column selector factory, so that callers can get a thing and something else can drive what happens underneath (the cursor in this case). We use similar patterns in quite a few other places. It seemed fairly straightforward to me at least, `setDelegate` updates the backing delegate factory to be the current group, and generation is just a counter for everytime we call set to indicate to the things the selector factory produces that we are on a new group. The selectors and matchers for this factory are themselves settable delegate selector/matcher too, so the pattern repeats (swap out delegate factory and also delegate matchers/selectors from those delegate factories on group transition)

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
clintropolis: to me it generally seems nicer to have a result object than a map since it gives us a bit more flexibility and is easier to document stuff about the result.

processing/src/main/java/org/apache/druid/segment/QueryableIndexCursorFactory.java
clintropolis: as mentioned in other comment this is because the delegate also does metadata stuff, so kind of always needs to work.

processing/src/main/java/org/apache/druid/segment/IndexIO.java
clintropolis: there are a handful of places we use this, so its a bit clearer to convert to boolean so i left this as is

processing/src/main/java/org/apache/druid/segment/projections/ClusteredValueGroupsBaseTableSchema.java
clintropolis: its part of the `ProjectionSchema` interface, so it needs implemented to be a base table projection. `getColumns` is the json property for the serde that writes this base table spec into the segment metadata (similar to equivalents on other base table schemas). So, they are serving different purposes

processing/src/main/java/org/apache/druid/segment/projections/ClusteredValueGroupsBaseTableSchema.java
clintropolis: I'd prefer to keep terminology aligned with MSQ 'clustered by' since this is basically pushing that operation inside of the segment. Additionally, 'group' feels slightly ambiguous when considering the 'rollup' base table projection schema, since it defines its schema as grouping columns and aggregators, https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/segment/projections/RollupTableProjectionSchema.java#L74, so I've mostly left things as-is.

processing/src/main/java/org/apache/druid/segment/IndexIO.java
clintropolis: yea, they are not wired up and cannot be used in this PR, this is just scaffolding and a forceful check to fix it once they get wired up "soon". I suppose i can remove until i add it later, but i was planning on this to be a near term follow-up so was just going ahead and putting all of the metadata in place to be ready later

processing/src/main/java/org/apache/druid/segment/IndexIO.java
clintropolis: this didn't really matter because mapFile just returns null if a file doesn't exist and the descriptor would be null too so we hit the continue block, however the code reads smoother if we just get the descriptor first and bail before calling mapFile so i went ahead and changed it

FrankChen021 (commented):
I have reviewed the follow-up on the clustered group column descriptor ordering. The current code now checks the column descriptor before mapping the clustered group file, so the sparse-group concern from the original inline comment is handled. I did not find any new actionable issues.

Reviewed 28 of 28 changed files.

---

processing/src/main/java/org/apache/druid/segment/projections/ClusteringColumnSelectorFactory.java
cecemei: i see what you mean, but also it's more than just a `settable` factory because it keeps a state of which group index the cursor is, and the deletegate/ clusteringValues class is self responsible for requesting new delegate/ values if the generation doesn't match. 

the generation counter still feels redundant given that setDelegate already passes in a new delegate or clusteringValues reference on every group transition, maybe each class can just cache the delegate and transition only when they see the delegate changes. 

That avoids introducing generation as a separate concept entirely.

cecemei (approved):

processing/src/main/java/org/apache/druid/segment/projections/ClusteringColumnSelectorFactory.java
clintropolis: generation isn't an index though currently, it is just a unique identifier for the current group, and only increases. Its use is similar to like how we use `ReadableVectorInspector.getId`. Currently on reset the concatenating cursor just creates fresh delegate cursors instead of individually reseting all of the delegate cursors so there isn't really even anything that could be cached since the delegate selector factories are tied to the delegate cursors (concat cursor doesn't currently save the delegate cursors anywhere either which it would need to if we were trying to reset them).

Since this is all entirely new stuff that mostly doesnt impact any existing code paths I am mainly going for correct behavior in this PR and plan to continue to iterate on these implementations in follow-up PRs.

PR #19461, opened 2026-05-13
feat: coordinator-side reconciliation for partial-load rules

clintropolis:
### Description
Follow-up to #19409, this PR enhances the coordintor side stuff so that partial replication is wired up and managed, taking the partial load size and rule fingerprints into consideration when doing assignment and balancing. A lot of the new code is similar to existing functionality of `StrategicSegmentAssigner`, but separate to minimize the chance of impacting existing code paths.

changes:
* adds selective inventory wrapping with `DataSegmentAndLoadProfile` which extends `DataSegment`; full-load slots in `DruidDataSource` / `ImmutableDruidDataSource` stay bare, only partial-loaded slots allocate a wrapper
* `PartialLoadProfile` records routed through a weak interner so identical profiles across replicas share by reference (the `wrappedLoadSpec` map is the bulk of per-replica heap).
* `SegmentChangeRequestLoad.forAnnouncement(segment)` detects a `partialProjection` wrapper on the segment's load spec and stamps the wrapper's fingerprint plus full segment size as `loadedBytes`; the coordinator reads this as a full-fallback profile, so partial-load rules work end-to-end (modulo no real partial loading yet) without thrashing
* `BatchDataSegmentAnnouncer` calls new method `forAnnouncement` on all three announcement paths to create the proper announcement depending on partial load profile
* fingerprint-aware partial reconciler in `StrategicSegmentAssigner`; classifies replicas (matching/stale, loaded/in-flight) via new `PartialSegmentStatusInTier`, applies the load-then-drop swap (stale stays serving until matching has actually loaded), cancels stale in-flight loads and re-targets the freed slots, allows additive reload on stale-loaded servers to mitigate the no-spare-server stuck state
* `ImmutableDruidServer.getPartialLoadProfile`, `DruidServer.addDataSegment(segment, profile)` with `loadedBytes`-aware `currSize` accounting, `HttpServerInventoryView` threads announcement-side profile through
* in-flight profile tracking on `ServerHolder.startOperation(action, segment, profile)` and `SegmentHolder.getProfile()` so the reconciler can read in-flight fingerprints.
* new coordinator stats - `PARTIAL_ASSIGNED`, `PARTIAL_STALE_DROPPED`, `PARTIAL_STALE_CANCELLED`
* removed `SegmentLoadingCapabilities.supportsPartialLoad` (now unused; per-replica `isFullFallback` is the better signal); coordinator always issues partial-load requests, historical handles via full-fallback until real partial loading lands
* tests: `StrategicSegmentAssignerPartialTest` (reconciler scenarios), `DruidServerPartialLoadTest` (size accounting), `PartialLoadProfileTest` (interning), `SegmentChangeRequestLoadTest` (forAnnouncement), updates across affected fixtures

### DIFF
--- a/server/src/main/java/org/apache/druid/client/HttpServerInventoryView.java
+++ b/server/src/main/java/org/apache/druid/client/HttpServerInventoryView.java
@@ -698,12 +698,7 @@ public class HttpServerInventoryView implements ServerInventoryView, FilteredSer
       }
       final DataSegment segment = loadRequest.getSegment();
       final Map<String, Object> loadSpec = segment.getLoadSpec();
-      if (loadedBytes < segment.getSize()
-          && loadSpec != null
-          && loadSpec.get("type") instanceof String
-          && ((String) loadSpec.get("type")).startsWith(PartialLoadSpec.TYPE_PREFIX)
-          && loadSpec.get("fingerprint") instanceof String
-          && loadSpec.get("delegate") instanceof Map) {
+      if (loadedBytes < segment.getSize() && PartialLoadSpec.detectPartialLoadSpec(loadSpec)) {
         return PartialLoadProfile.forLoaded(loadSpec, fingerprint, loadedBytes);
       }
       return PartialLoadProfile.forFullFallback(fingerprint, loadedBytes);
--- a/server/src/main/java/org/apache/druid/client/HttpServerInventoryView.java
+++ b/server/src/main/java/org/apache/druid/client/HttpServerInventoryView.java
@@ -48,6 +48,7 @@ import org.apache.druid.java.util.emitter.EmittingLogger;
 import org.apache.druid.java.util.emitter.service.ServiceEmitter;
 import org.apache.druid.java.util.emitter.service.ServiceMetricEvent;
 import org.apache.druid.java.util.http.client.HttpClient;
+import org.apache.druid.segment.loading.PartialLoadSpec;
 import org.apache.druid.server.DruidNode;
 import org.apache.druid.server.coordination.ChangeRequestHttpSyncer;
 import org.apache.druid.server.coordination.ChangeRequestsSnapshot;
@@ -593,13 +594,12 @@ public class HttpServerInventoryView implements ServerInventoryView, FilteredSer
           druidServer.iterateAllSegments().forEach(segment -> toRemove.put(segment.getId(), segment));
 
           for (DataSegmentChangeRequest request : changes) {
-            if (request instanceof SegmentChangeRequestLoad) {
-              SegmentChangeRequestLoad loadRequest = (SegmentChangeRequestLoad) request;
+            if (request instanceof SegmentChangeRequestLoad loadRequest) {
               DataSegment segment = loadRequest.getSegment();
               toRemove.remove(segment.getId());
               addSegment(segment, partialLoadProfileFor(loadRequest), true);
-            } else if (request instanceof SegmentSchemasChangeRequest) {
-              runSegmentCallbacks(input -> input.segmentSchemasAnnounced(((SegmentSchemasChangeRequest) request).getSegmentSchemas()));
+            } else if (request instanceof SegmentSchemasChangeRequest changeRequest) {
+              runSegmentCallbacks(input -> input.segmentSchemasAnnounced((changeRequest).getSegmentSchemas()));
             } else {
               log.error(
                   "Server[%s] gave a non-load dataSegmentChangeRequest[%s]., Ignored.",
@@ -666,7 +666,20 @@ public class HttpServerInventoryView implements ServerInventoryView, FilteredSer
 
     /**
      * Builds a {@link PartialLoadProfile} from a load announcement when the historical populated the partial-load
-     * fields ({@code fingerprint} + {@code loadedBytes}). Returns null for full-load announcements.
+     * wire fields ({@code fingerprint} + {@code loadedBytes}). Returns {@code null} for plain full-load
+     * announcements.
+     * <p>
+     * Picks between {@link PartialLoadProfile#forLoaded forLoaded} and
+     * {@link PartialLoadProfile#forFullFallback forFullFallback} based on what the historical actually realized:
+     * <ul>
+     *   <li>If {@code loadedBytes} is strictly less than the segment's full size <em>and</em> the announced segment
+     *       carries a {@link PartialLoadSpec} wrapper on its load spec, the historical honored the partial request —
+     *       preserve the wrapped load spec on the profile so per-replica observability (e.g.
+     *       {@link PartialLoadProfile#isFullFallback}) reports correctly.</li>
+     *   <li>Otherwise the historical either fell back to full or doesn't yet implement scheme-specific partial
+     *       loading; build a full-fallback profile (the wire fingerprint still satisfies the rule, so the
+     *       coordinator's reconciler counts it as matching without reload thrash).</li>
+     * </ul>
      */
     @Nullable
     private static PartialLoadProfile partialLoadProfileFor(SegmentChangeRequestLoad loadRequest)
@@ -676,9 +689,16 @@ public class HttpServerInventoryView implements ServerInventoryView, FilteredSer
       if (fingerprint == null || loadedBytes == null) {
         return null;
       }
-      // The wrapped LoadSpec is on the segment itself when a partial load was requested by the coordinator; we don't
-      // duplicate it on the inventory profile; the announcement uses the full-fallback factory which carries the
-      // fingerprint and the realized loadedBytes only.
+      final DataSegment segment = loadRequest.getSegment();
+      final Map<String, Object> loadSpec = segment.getLoadSpec();
+      if (loadedBytes < segment.getSize()
+          && loadSpec != null
+          && loadSpec.get("type") instanceof String
+          && ((String) loadSpec.get("type")).startsWith(PartialLoadSpec.TYPE_PREFIX)
+          && loadSpec.get("fingerprint") instanceof String
+          && loadSpec.get("delegate") instanceof Map) {
+        return PartialLoadProfile.forLoaded(loadSpec, fingerprint, loadedBytes);
+      }
       return PartialLoadProfile.forFullFallback(fingerprint, loadedBytes);
     }
 
--- a/server/src/main/java/org/apache/druid/client/HttpServerInventoryView.java
+++ b/server/src/main/java/org/apache/druid/client/HttpServerInventoryView.java
@@ -71,6 +71,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -653,6 +654,12 @@ public class HttpServerInventoryView implements ServerInventoryView, FilteredSer
                 }
               }
           );
+        } else if (!Objects.equals(druidServer.getPartialLoadProfile(segment.getId()), profile)) {
+          // Already present, but the per-server partial-load profile changed (e.g. the historical re-announced after
+          // an additive reload swapped the wrapper / fingerprint). Update the inventory state in place; the segment
+          // is still on this server, only the profile metadata changed, so don't fire segmentAdded/segmentRemoved
+          // callbacks
+          druidServer.updateDataSegmentProfile(segment, profile);
         } else if (!fullSync) {
           // duplicates can happen when doing a full sync from a 'reset', so only warn for dupes on delta changes
           log.warn(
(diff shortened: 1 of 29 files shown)


server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentHolder.java
```diff
@@ -90,18 +92,54 @@ public SegmentHolder(
...
   {
     this.segment = segment;
     this.action = action;
-    this.changeRequest = (action == SegmentAction.DROP)
-                         ? new SegmentChangeRequestDrop(segment)
-                         : new SegmentChangeRequestLoad(segment);
+    this.profile = profile;
+    if (action == SegmentAction.DROP) {
+      this.changeRequest = new SegmentChangeRequestDrop(segment);
+    } else if (profile != null && profile.wrappedLoadSpec() != null) {
```
capistrant, line 122: non-null profile with a null wrapped load spec silently falls back to a normal request. is that ok or would we want to either do something exceptional or log a warn of some kind to make sure it is addressed

server/src/main/java/org/apache/druid/server/coordination/SegmentChangeRequestLoad.java
```diff
@@ -38,6 +40,29 @@
...
+   * <p>
+   * For segments loaded without a partial-load wrapper (the common case), this returns a bare load request with no
+   * fingerprint or loadedBytes, equivalent to {@link #SegmentChangeRequestLoad(DataSegment)}.
+   */
+  public static SegmentChangeRequestLoad forAnnouncement(DataSegment segment)
+  {
+    final Map<String, Object> loadSpec = segment.getLoadSpec();
+    if (loadSpec != null && PartialProjectionLoadSpec.TYPE.equals(loadSpec.get("type"))) {
+      final Object fingerprint = loadSpec.get("fingerprint");
+      if (fingerprint instanceof String) {
```
capistrant: warn if fingerprint is not string type? maybe/hopefully impossible but silently falling back to a full load request if this did fail seems hard to troubleshoot. Unless non-string fingerprints would blow some other part of the code up

server/src/main/java/org/apache/druid/server/coordinator/ServerHolder.java
```diff
@@ -387,6 +400,16 @@ public int getNumQueuedSegments()
   }
 
   public boolean startOperation(SegmentAction action, DataSegment segment)
+  {
+    return startOperation(action, segment, null);
+  }
+
+  /**
+   * Like {@link #startOperation(SegmentAction, DataSegment)}, but additionally records the {@link PartialLoadProfile}
```
capistrant: nit: wording feels weird to say it is like `startOperation(SegmentAction, DataSegment)` when all that method does is call this method. I think this would be better off just describing what it does with a note about profile handling

server/src/main/java/org/apache/druid/server/http/DataSourcesResource.java
```diff
@@ -755,7 +755,7 @@ private ImmutableDruidDataSource getQueryableDataSource(final String dataSourceN
       }
     }
 
-    return new ImmutableDruidDataSource(dataSourceName, Collections.emptyMap(), segmentMap);
+    return new ImmutableDruidDataSource(dataSourceName, Collections.emptyMap(), segmentMap.values());
```
capistrant, line 765: If constructor choice here is material to the functioning of the platform, I think that the choice should be explained here in a short comment to avoid regression

server/src/main/java/org/apache/druid/server/coordinator/loading/PartialSegmentStatusInTier.java
```diff
@@ -0,0 +1,161 @@
...
+      if (inFlight != null && Objects.equals(inFlight.fingerprint(), requestedFingerprint)) {
+        matchingInFlight.add(server);
+      } else {
+        staleInFlight.add(server);
+      }
+    } else if (action == null && server.canLoadSegment(segment)) {
+      eligibleForFreshLoad.add(server);
+    }
+    // Other actions (DROP, MOVE_TO, MOVE_FROM) are intentionally not classified here, they're handled by the
+    // existing replica-counting paths.
```
capistrant: Potentially confusing comment. I assume you mean in extension to the !isLoaded case we intentionally do not classify, since in the case where teh segment is loaded we don't check action at all. Can this be dropped and replaced by a javadoc that kind of sets the full expectation of what is classified and what is purposely ignored?

server/src/main/java/org/apache/druid/server/coordinator/loading/PartialSegmentStatusInTier.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  /**
+   * A stale-loaded server is reload-eligible iff it's not decommissioning, has no other action queued for the segment,
```
capistrant: do we need something more to satisfy `has no other action queued for the segment`? In conjunction with `classify` it seems like we are not checking for the stale loaded but already queued additive load scenario.

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
```diff
@@ -273,6 +274,273 @@ public void replicateSegment(DataSegment segment, Map<String, Integer> tierToRep
...
+    }
+
+    int dropsQueuedOnTier = 0;
+    int dropBudget = maxReplicasToDrop;
+
+    // Surplus matching: drop excess matching replicas. Same shape as the full-load surplus path.
+    if (matchingProjected > requiredReplicas) {
+      final int surplus = matchingProjected - requiredReplicas;
+      final List<ServerHolder> cancelledMatching = new ArrayList<>();
+      cancelLoadsOnServers(segment, status.getMatchingInFlight(), surplus, cancelledMatching);
```
capistrant: appear to be missing incrementStat if the cancel takes some action

server/src/main/java/org/apache/druid/client/HttpServerInventoryView.java
```diff
@@ -659,6 +664,24 @@ public CallbackAction apply(SegmentCallback input)
...
+    {
+      final String fingerprint = loadRequest.getFingerprint();
+      final Long loadedBytes = loadRequest.getLoadedBytes();
+      if (fingerprint == null || loadedBytes == null) {
+        return null;
+      }
+      // The wrapped LoadSpec is on the segment itself when a partial load was requested by the coordinator; we don't
+      // duplicate it on the inventory profile; the announcement uses the full-fallback factory which carries the
+      // fingerprint and the realized loadedBytes only.
+      return PartialLoadProfile.forFullFallback(fingerprint, loadedBytes);
```
capistrant: Is it weird to be doing this? You justify it in the comment, but it still feels like we should either generalize the factory name/javadoc or create a new one that does the same, but is not named and documented as a full load fallback mechanism for the historical

capistrant (commented):
mostly minor comments on first pass. Still have to review DruidServerPartialLoadTest and StrategicSegmentAssignerPartialTest in more detail

server/src/main/java/org/apache/druid/server/coordinator/ServerHolder.java
clintropolis: oops, refactored some things while iterating on this stuff to consolidate and forgot to update 😅 

I did another pass on the javadoc and think this (and a bunch of other places) should be better now

server/src/main/java/org/apache/druid/server/http/DataSourcesResource.java
clintropolis: added comment with rationale

server/src/main/java/org/apache/druid/server/coordinator/loading/PartialSegmentStatusInTier.java
clintropolis: had a pass on javadocs, trying to steer things towards `StrategicSegmentAssignerupdateReplicasInTierPartial` which tries to describe the whole flow, since this thing is mainly a helper to collect information about the segment to help decide what to do

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
clintropolis: added comment

server/src/main/java/org/apache/druid/client/HttpServerInventoryView.java
clintropolis: reworked this a bit to make this a bit clearer/cleaner

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
```diff
@@ -273,6 +274,311 @@ public void replicateSegment(DataSegment segment, Map<String, Integer> tierToRep
...
+    }
+
+    final List<ServerHolder> destinations = new ArrayList<>(
+        status.getEligibleForFreshLoad().size()
+        + canceledStaleServers.size()
+        + status.getEligibleForAdditiveReload().size()
+    );
+    destinations.addAll(status.getEligibleForFreshLoad());
+    destinations.addAll(canceledStaleServers);
+    destinations.addAll(status.getEligibleForAdditiveReload());
```
FrankChen021, line 509: [P1] Additive reloads cannot update stale loaded replicas

The new fallback queues partial loads back onto stale-loaded servers, but same-id loads do not produce a fresh inventory profile: BatchDataSegmentAnnouncer skips announceSegment when the segment is already announced, and HttpServerInventoryView ignores addSegment for an id already present. In a saturated tier with only stale replicas, the load can report success while the coordinator never sees the new fingerprint, so the replica remains stale and is requeued every run. Either make same-server reloads update the announcement/inventory profile, or avoid this additive path until the historical/inventory lifecycle can actually replace the profile.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

Reviewed 28 of 28 changed files.

---

server/src/main/java/org/apache/druid/server/http/SegmentLoadingCapabilities.java
```diff
@@ -30,27 +30,15 @@ public class SegmentLoadingCapabilities
 {
   private final int numLoadingThreads;
   private final int numTurboLoadingThreads;
-  private final boolean supportsPartialLoad;
 
   @JsonCreator
   public SegmentLoadingCapabilities(
```
FrankChen021, line 35: [P2] Keep a capability gate for older historicals

After removing supportsPartialLoad, the coordinator has no way to distinguish historicals that can deserialize partialProjection load specs from older nodes during a rolling upgrade. replicateSegmentPartially now sends wrapped load specs to every eligible server, so an older historical will fail the load instead of doing the intended full fallback. Please preserve a capability bit that defaults false for old/missing endpoints and degrade those destinations to regular full loads until support is advertised.

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
capistrant: ^ seems like a legitimate concern

server/src/main/java/org/apache/druid/server/http/SegmentLoadingCapabilities.java
capistrant: I believe the PR description calls out why this is being removed with a proper justification.

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
clintropolis: this part will come as an enhancement when partial loads are actually wired up on the historical, we will check the fingerprint of the updated request against the existing loaded fingerprint and load the missing parts. There are no partial loads right now so we always fall back to full loads, which i think should never be stale.

server/src/main/java/org/apache/druid/server/http/SegmentLoadingCapabilities.java
clintropolis: To elaborate a bit more on why i removed `supportsPartialLoad `, with it existing and the coordinator considering it for assignment, when actually in this scenario it creates hot spots whenever you roll between versions; like if only a single server in a tier supports partial loads then all partial loads are assigned to the single server. This seemed like more problems than benefits.

With regards to older versions of historicals not understanding the new partial load specs, without the flag existing I plan to just document this as new functionality that should not be used until an operator is certain that there is no need to roll back, the same we do with certain types of other new features (like segment format stuff).

capistrant (approved):

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
clintropolis: Oh i see, yea, i guess there will be a temporary problem. I think I can put a fix in until the actual functionality is there.

capistrant (approved):
latest changes look good to me too. good catch

FrankChen021 (commented):
Reviewed 29 of 29 changed files for this follow-up. The additive reload/profile update concern appears handled by the current re-announcement path and in-place inventory profile update; I did not find new high-confidence correctness issues in the current PR head.

---

PR #19462, opened 2026-05-13
feat: clustered segments DataSegment changes and PartialLoadMatcher

clintropolis:
### Description
Companion PR to #19460 that adds adds the `DataSegment` changes so that clustered segments are visible to coordinators/brokers and a `PartialLoadMatcher` implementation that can use wildcards to match cluster groups for partial loads.


processing/src/main/java/org/apache/druid/timeline/ClusterGroupTuples.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  @JsonProperty
+  public RowSignature getClusteringColumns()
+  {
+    return clusteringColumns;
+  }
+
+  @JsonProperty
+  public List<List<Object>> getTuples()
```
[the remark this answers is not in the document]
clintropolis: Hmm, both inner and outers lists are wrapped in `Collections.unmodifiableList`, not sure what this is complaining about exactly

clintropolis:
if this PR goes in after #19461 it needs adjusted to use the `PartialLoadSpec` base type introduced there (likewise if this is merged first that PR needs to update)

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Reviewed 19 of 19 changed files.

---

processing/src/main/java/org/apache/druid/timeline/DataSegment.java
```diff
@@ -247,6 +273,7 @@ public DataSegment(
       @Nullable List<String> dimensions,
       @Nullable List<String> metrics,
       @Nullable List<String> projections,
+      @Nullable ClusterGroupTuples clusterGroups,
```
FrankChen021, line 316: [P2] Keep the existing public constructor signature

This inserts clusterGroups into the public constructor that previously took (... projections, ShardSpec, CompactionState, ...). DataSegment is annotated @PublicApi, and extension or application code can construct projection-aware segments with that old signature; after this change those callers either fail to compile or, if already compiled, hit NoSuchMethodError even though they do not use cluster groups. Please keep the old overload and delegate with clusterGroups = null instead of replacing the existing public signature.

clintropolis: fine, I can add it... `DataSegment` is a bit of a graveyard of deprecated constructors and builder creators though... we need to delete these at some point and just call it out in release notes. I would expect creating `DataSegment` directly is kind of uncommon in extensions (outside of tests at least).

capistrant (approved):
+1 with or without the DataSegment constructor business

capistrant (approved):

FrankChen021 (commented):
I reviewed the follow-up on the DataSegment constructor compatibility thread and the current incremental update. The previous public constructor overload has been restored and delegates with clusterGroups = null, so the original concern is handled.

Reviewed 4 of 19 changed files.

---

PR #19468, opened 2026-05-17
minor: add fileGroup to V10 segment SegmentFileContainerMetadata, switch default metadata compression to zstd

clintropolis:
### Description
Adds a `fileGroup` field to `SegmentFileContainerMetadata` of V10 segments, which after #19375 associated containers with `startFileGroup` to place all internal files under the same group in the same container. This just adds that field to the container so on the read side we have an easy way to get a list of the internal file bundles present in a v10 segment, making some things easier for partial download. The segment file builder now also enforces that columns and files added during a file group are prefixed with the file group name.

While I was here, I also set the default metadata compression for v10 segments to use zstd instead of uncompressed. I can't really think of much downside to doing this, I just hadn't got around to switching it yet, so did it now since I had to update the expected size of a test anyway.


FrankChen021 (commented):

---

capistrant (approved):

PR #19477, opened 2026-05-18
feat: resetToLatestAndBackfill using bounded stream supervisor

aho135:
This change adds an endpoint called resetToLatestAndBackfill to SupervisorResource. This is a useful feature for operating Druid clusters where the most recent data is the most important (such as alerting use cases).

### Description
Adds an endpoint called resetToLatestAndBackfill to automatically ingest skipped data in the case where the offset is reset to latest. This requires useEarliestOffset=false, useConcurrentLocks=true because there can be conflicting time intervals between the backfill task and the main supervisor tasks and the Supervisor needs to be in a running state in order to call updatePartitionLagFromStream() to get the latest offsets

#### Release note

Adds an endpoint called `resetToLatestAndBackfill` for Stream Supervisors. This is a useful feature for operating Druid clusters where the most recent data is the most important 

<hr>

 * `SupervisorResource`
 * `SupervisorManager`

<hr>

This PR has:

- [X] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +399,117 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
...
+      throw new ISE("Skipping reset: Failed to get checkpointed offsets for supervisor[%s]", id);
+    }
+
+    log.info("Resetting supervisor[%s] metadata to latest offsets", id);
+    DataSourceMetadata resetMetadata = streamSupervisor.createDataSourceMetaDataForReset(
+        streamSupervisor.getIoConfig().getStream(),
+        endOffsets
+    );
+
+    streamSupervisor.resetOffsets(resetMetadata);
```
FrankChen021, line 451: [P1] Reset can skip data before backfill is guaranteed

`resetOffsets` only enqueues a `ResetOffsetsNotice`; it does not synchronously update metadata. This code queues the main supervisor reset before the bounded config is fully built and before `createOrUpdateAndStartSupervisor` succeeds. If any later step fails, the queued reset can still advance the live supervisor to latest offsets with no backfill supervisor, losing the skipped range this endpoint is meant to preserve.

extensions-core/kinesis-indexing-service/src/main/java/org/apache/druid/indexing/kinesis/supervisor/KinesisSupervisor.java
```diff
@@ -336,7 +336,7 @@ protected OrderedSequenceNumber<String> makeSequenceNumber(String seq, boolean i
   }
 
   @Override
-  protected void updatePartitionLagFromStream()
+  public void updatePartitionLagFromStream()
```
FrankChen021, line 339: [P2] Kinesis cannot provide backfill end offsets

The new manager path calls `updatePartitionLagFromStream()` and then `getLatestSequencesFromStream()`, but Kinesis only updates time lag here and does not override `getLatestSequencesFromStream()`, so it inherits the base empty map. Any Kinesis supervisor that passes the earlier checks will fail with empty latest offsets instead of starting a backfill.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +399,117 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
...
+    SeekableStreamSupervisorSpec streamSpec = (SeekableStreamSupervisorSpec) supervisorPair.rhs;
+
+    // Verify useEarliestOffset is false
+    if (streamSupervisor.getIoConfig().isUseEarliestSequenceNumber()) {
+      throw new IAE("Reset with skipped offsets is not supported when useEarliestOffset is true.");
+    }
+
+    // Verify useConcurrentLocks is enabled
+    final Map<String, Object> context = streamSpec.getContext();
+    if (context == null || !Boolean.TRUE.equals(context.get("useConcurrentLocks"))) {
```
FrankChen021: [P2] Concurrent-lock check rejects valid true contexts

This check only accepts a literal Boolean true under the hard-coded key. Other Druid paths in this class parse `Tasks.USE_CONCURRENT_LOCKS` with `QueryContexts.getAsBoolean`, which accepts values like string `true`. Supervisors whose tasks actually use concurrent locks can therefore be rejected by this endpoint.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 2 |
| P3 | 0 |
| Total | 3 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 2 |
| P3 | 0 |
| Total | 3 |

---

extensions-core/kinesis-indexing-service/src/main/java/org/apache/druid/indexing/kinesis/supervisor/KinesisSupervisor.java
aho135: Good callout, @jaykanakiya will be tackling Kinesis support in a separate PR

FrankChen021: Thanks for confirming. I reviewed all 11 changed files in the current head; since this PR still exposes `/resetOffsetsAndBackfill` to every `SeekableStreamSupervisor`, Kinesis still reaches `SupervisorManager.resetSupervisorAndBackfill()`, calls `updatePartitionLagFromStream()`, then sees the base empty `getLatestSequencesFromStream()` and returns a 500. If Kinesis support is intentionally deferred, can this PR explicitly reject or disable Kinesis and document the supported supervisor types, rather than leaving the new endpoint available but broken until the follow-up PR lands?

aho135: Thanks for the review @FrankChen021! I'm not able to do an `instanceof` check in SupervisorManager due to module dependencies. I could add something like `supportsResetAndBackfill` to `SeekableStreamSupervisor` but seems like overkill given that it'll be reverted after `getLatestSequencesFromStream()` is implemented for Kinesis. Given that the current response of 500 is accurate I think it's fine for now

Alternatively we can hold off on merging this until the Kinesis changes are in

docs/api-reference/supervisor-api.md
```diff
@@ -3539,6 +3539,101 @@ when the supervisor's tasks restart, they resume reading from `{"0": 100, "1": 1
...
 
+### Reset offsets and start a backfill supervisor
+
+Resets the supervisor to the latest available stream offsets and starts a new bounded backfill supervisor to ingest the data in the skipped range.
+
+This endpoint is useful when a supervisor has fallen behind and you want to catch it up to the latest offsets without losing the skipped data. The main supervisor resumes ingesting from the latest offsets, while the backfill supervisor processes the range from the previously checkpointed offsets up to the latest offsets at the time of the reset.
+
+The following requirements must be met before calling this endpoint:
+
+- The supervisor must be a `SeekableStreamSupervisor`.
```
abhishekrb19: We could point to the supervisor [docs](https://druid.apache.org/docs/latest/ingestion/supervisor/) - any implementation of the supervisor should suffice I think (instead of the java class name which is not documented elsewhere afaik)

docs/api-reference/supervisor-api.md
```diff
@@ -3539,6 +3539,101 @@ when the supervisor's tasks restart, they resume reading from `{"0": 100, "1": 1
...
+- The supervisor's `useEarliestSequenceNumber` property must be `false`.
```
abhishekrb19, line 3559: Is this requirement needed? With `useConcurrentLocks` and two parallel supervisors, I'd imagine this would still work if the offset ranges are overlapping

extensions-core/kinesis-indexing-service/src/main/java/org/apache/druid/indexing/kinesis/supervisor/KinesisSupervisor.java
abhishekrb19: Yeah, I think returning a 5xx that should be good enough

We could probably add a one-line limitation in the docs for completeness noting that the feature doesn’t apply to Kinesis supervisors ? That can be removed once that gap is addressed in a future patch depending on when it lands

indexing-service/src/test/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManagerTest.java
```diff
@@ -1068,6 +1079,205 @@ public void test_isAnotherTaskGroupPublishingToPartitions()
     );
   }
 
+  @Test
+  public void testResetSupervisorAndBackfill() throws Exception
```
abhishekrb19: Thanks for the tests @aho135! Do you think it’d be possible to extend the embedded test `KafkaBoundedSupervisorTest` you added to include some end-to-end coverage for this API as well?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +374,115 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
...
+    Map<?, ?> startOffsets = streamSupervisor.getOffsetsFromMetadataStorage();
+
+    // Validate that we successfully retrieved offsets
+    if (endOffsets == null || endOffsets.isEmpty()) {
+      throw new ISE("Skipping reset: Failed to get latest offsets from stream for supervisor[%s]", id);
+    }
+    if (startOffsets == null || startOffsets.isEmpty()) {
+      throw new ISE("Skipping reset: Failed to get checkpointed offsets for supervisor[%s]", id);
+    }
+
```
abhishekrb19, line 423: nit: It’ll likely be cleaner to move these validations into their own function

abhishekrb19 (commented):
Thanks @aho135! Left some comments

docs/api-reference/supervisor-api.md
aho135: Yeah this is still a requirement even with separate supervisors. This is essentially multi-supervisor support which requires useConcurrentLocks, as the two supervisors can be writing to the same interval

extensions-core/kinesis-indexing-service/src/main/java/org/apache/druid/indexing/kinesis/supervisor/KinesisSupervisor.java
FrankChen021: Thanks, that sounds reasonable if Kinesis support is intentionally deferred. The current docs still say the endpoint applies to a streaming supervisor, and the same page describes streaming supervisors as Kafka or Kinesis, so Kinesis users would still be directed to an endpoint that is expected to 500. Reviewed 11 of 11 changed files; I think this just needs the explicit Kinesis limitation in the new docs section.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +374,120 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
...
+    log.info("Started backfill supervisor[%s] for supervisor[%s]", backfillSupervisorId, id);
+
+    log.info("Resetting supervisor[%s] metadata to latest offsets", id);
+    DataSourceMetadata resetMetadata = streamSupervisor.createDataSourceMetaDataForReset(
+        streamSupervisor.getIoConfig().getStream(),
+        endOffsets
+    );
+
+    streamSupervisor.resetOffsets(resetMetadata);
```
FrankChen021, line 451: [P1] Avoid resetting the main supervisor with stale offsets

The main supervisor keeps ingesting while this method captures `endOffsets`/`startOffsets`, creates the backfill supervisor, and only then calls `resetOffsets`. If a running task checkpoints after the captured `endOffsets` but before this call, `resetOffsetsInternal` builds `currentMetadata.plus(resetMetadata)`, and `plus` overwrites the newer partition value with the stale captured end offset. That rolls the main supervisor checkpoint backward and can reingest rows that were already published. The reset should be conditional on metadata still matching the captured range, retried with fresh offsets, or performed while the main supervisor is quiesced.

aho135: Thanks @FrankChen021 This should be a very rare scenario since I expect this endpoint only to be used if there is some lag built up. I did push a change to prefer the checkpointed offset if it is greater than the requested reset offset to handle this edge case

aho135: The issue still stands that there can be duplicate ingestion with the backfill Supervisor if the main Supervisor checkpoints after `startOffsets` is captured. To completely remove this race I think we'd need to suspend the supervisor and wait for all tasks to gracefully shutdown. This is something that I was trying to avoid as then we would need to deal with timeouts and such. Let me know your thoughts on this one though

FrankChen021: Thanks for adding `resetOffsetsForwardOnly`; it handles the common case where the checkpoint has already advanced by the time the reset notice reads metadata. I still think the rollback race is not fully closed, though: `resetOffsetsInternal` reads current metadata, computes the max map, then calls `resetDataSourceMetadata`, and that storage method does an unconditional `UPDATE ... WHERE dataSource = :dataSource`. If a task publishes/checkpoints after the read but before that update, the max is computed from stale metadata and the reset can still overwrite the newer checkpoint. So I would keep this open unless the reset write is CAS/retried against the metadata it read, or the main supervisor is quiesced before capturing/resetting. I agree the backfill-overlap duplicate case is separate and requires either accepting/documenting possible duplicates or suspending the main supervisor. Reviewed 17 of 17 changed files.

aho135: Thanks for the review @FrankChen021 

I am leaning towards reverting the `resetOffsetsForwardOnly` changes. My reasoning is that "backfill-overlap duplicate case" is not solvable without suspending and task shutdown. I expect that this duplication scenario is more common than the one `resetOffsetsForwardOnly` addresses, since it would only make sense to use this endpoint in the case where there is some lag built up. If we're not solving that scenario and accepting that this endpoint has the potential for duplicate ingestion then I think it would make sense to just accept the limitation and document it rather than adding additional complexity to solve the even more rare duplication scenario. 

Since many data streaming pipelines follow at least once ingestion and the duplication scenario is rare and bounded I think this is a reasonable trade-off to not have to deal with suspension and task shutdown

FrankChen021: Thanks, that trade-off makes sense for the separate backfill-overlap duplicate case. I would still keep the reset metadata issue distinct, though. The current `resetOffsetsForwardOnly` code only protects against checkpoints that are already visible when the reset notice reads metadata; after that it computes the max offsets and calls `resetDataSourceMetadata`, which still performs an unconditional update. A task checkpoint between that read and update can still be overwritten. If the endpoint intentionally accepts that as bounded at-least-once duplicate risk, the docs should say the main supervisor is not quiesced and duplicate ingestion can happen both from backfill overlap and from this reset race. Otherwise this still needs CAS/retry or suspend/drain before the reset.

aho135: @FrankChen021 this makes sense. I went ahead and reverted the `resetOffsetsForwardOnly` changes and updated the docs explaining the two duplication scenarios. Appreciate the review on this one!

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

The reset/backfill duplicate-ingestion follow-up is handled: the forward-only reset helper was reverted and the docs now describe both the backfill-overlap and reset-race duplicate windows.

---

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +374,120 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
...
+
+    return ImmutableMap.of(
+        "id", id,
+        "backfillSupervisorId", backfillSupervisorId
+    );
+  }
+
+  private void validateResetAndBackfill(String id, Pair<Supervisor, SupervisorSpec> supervisorPair)
+  {
+    if (!(supervisorPair.lhs instanceof SeekableStreamSupervisor)) {
```
FrankChen021: [P2] Reject unsupported Kinesis supervisors before reset/backfill

The new endpoint is documented as unsupported for Amazon Kinesis, but this guard accepts any `SeekableStreamSupervisor`. `KinesisSupervisor` extends that base class and does not override `getLatestSequencesFromStream`, so a Kinesis request that otherwise passes the useEarliest/useConcurrent/RUNNING checks reaches the empty latest-offset check and is returned as a 500 instead of a clear unsupported-supervisor 400. Please either reject Kinesis/supervisors that cannot provide latest offsets here, or implement the latest-offset capture needed for Kinesis.

aho135: From our earlier [conversation](https://github.com/apache/druid/pull/19477#discussion_r3281515095) I will leave this one as is. We have the Kinesis implementation planned for within the next two weeks

FrankChen021 (commented):

Follow-up checked: the Kinesis reset/backfill concern has been acknowledged as a known gap with a planned implementation, so no inline reply is useful here.

---

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -640,6 +640,50 @@ private Response handleResetRequest(
     );
   }
 
+  @POST
+  @Path("/{id}/resetOffsetsAndBackfill")
```
abhishekrb19: Thinking about it, the existing `/resetOffsets` API takes offsets and the `/reset` offsets simply does a hard reset. I think this API aligns with the latter, what do you think about renaming this to `resetToLatestAndBackfill`?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +374,120 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
...
+  public Map<String, Object> resetSupervisorAndBackfill(String id, @Nullable Integer backfillTaskCount)
+  {
+    Preconditions.checkState(started, "SupervisorManager not started");
+    Preconditions.checkNotNull(id, "id");
+
+    Pair<Supervisor, SupervisorSpec> supervisorPair = supervisors.get(id);
+    validateResetAndBackfill(id, supervisorPair);
+
+    SeekableStreamSupervisor streamSupervisor = (SeekableStreamSupervisor) supervisorPair.lhs;
+    SeekableStreamSupervisorSpec streamSpec = (SeekableStreamSupervisorSpec) supervisorPair.rhs;
```
abhishekrb19: nit: maybe get the `streamSupervisor` and `streamSpec` once with the type casted and pass it down to `validateResetAndBackfill` as well since it requires it

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +374,120 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
...
+  {
+    ObjectNode specNode = jsonMapper.valueToTree(sourceSpec);
+    specNode.put("id", backfillSupervisorId);
+    ObjectNode ioConfigNode = (ObjectNode) specNode.path("spec").path("ioConfig");
+    ioConfigNode.set("boundedStreamConfig", jsonMapper.valueToTree(boundedStreamConfig));
+    if (backfillTaskCount != null) {
+      ioConfigNode.put("taskCount", backfillTaskCount);
+    }
+    return jsonMapper.treeToValue(specNode, SupervisorSpec.class);
+  }
```
abhishekrb19: Instead of parsing the json tree here, what do you think about making this an abstract method on the `SeekableStreamSupervisorSpec` itself that would create a new spec based on the `sourceSpec` with the `boundedStreamConfig` populated? (similar to `createSuspendedSpec()` and `createRunningSpec()`)

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
abhishekrb19: nit: maybe make this invoke `requireStreamSupervisor()` for consistency with the other manager operations

docs/api-reference/supervisor-api.md
abhishekrb19: Ah yeah, I was thinking about this requirement: `The supervisor's useEarliestSequenceNumber property must be false`

Thinking about it more, ingesting from earliest + backfill doesn’t make much sense for the generic reset case, so this seems like a useful guardrail.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +374,120 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
...
+      SupervisorSpec backfillSpec = createBackfillSpec(streamSpec, backfillSupervisorId, boundedStreamConfig, backfillTaskCount);
+      createOrUpdateAndStartSupervisor(backfillSpec);
+    }
+    catch (JsonProcessingException e) {
+      throw new ISE(e, "Failed to create backfill supervisor spec for supervisor[%s]", id);
+    }
+
+    log.info("Started backfill supervisor[%s] for supervisor[%s]", backfillSupervisorId, id);
+
+    log.info("Resetting supervisor[%s] metadata to latest offsets", id);
```
abhishekrb19: I know this can be determined from the supervisor spec directly, but I think it’d be nice if we could also log the specific start/end offsets computed for the backfill supervisor so folks can audit things more easily if needed?

Looks like there are logs for the reset path already: `log.info("Posting ResetOffsetsNotice with reset dataSource metadata[%s]", resetDataSourceMetadata);`

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -631,4 +726,29 @@ private SupervisorSpec getSpec(String id)
...
+
+  /**
+   * Returns true if the spec's context enables concurrent (append) locks, accepting both
+   * {@code useConcurrentLocks: true} (or any truthy string) and {@code taskLockType: APPEND}.
+   */
+  private static boolean specHasConcurrentLocks(SeekableStreamSupervisorSpec spec)
+  {
+    Map<String, Object> context = spec.getContext();
+    if (context == null) {
+      return false;
```
abhishekrb19: Looks like the old code was effectively using this variable, which is currently set to false:

```suggestion
      return Tasks.DEFAULT_USE_CONCURRENT_LOCKS;
```

If the default changes someday, the different code flows would automatically pick it up

docs/api-reference/supervisor-api.md
```diff
@@ -3539,6 +3539,109 @@ when the supervisor's tasks restart, they resume reading from `{"0": 100, "1": 1
...
+
+This endpoint is useful when a supervisor has fallen behind and you want to catch it up to the latest offsets without losing the skipped data. The main supervisor resumes ingesting from the latest offsets, while the backfill supervisor processes the range from the previously checkpointed offsets up to the latest offsets at the time of the reset.
+
+**Duplicate ingestion notice:** The main supervisor is not quiesced before the reset. This means duplicate data can occur in two ways:
+- **Backfill overlap:** Any tasks that were in-flight at the time of the reset may publish segments covering part of the backfill range before being shut down.
+- **Reset race:** If a task checkpoint is written to the metadata store between when this endpoint captures the current offsets and when it applies the reset, that checkpoint can be overwritten, causing the main supervisor to re-ingest already-processed data.
+
+Both windows are narrow in practice, but cannot be fully eliminated without manually suspending the main supervisor before calling this endpoint and waiting for all pending tasks to complete.
```
abhishekrb19, line 3554: This seems reasonable to me. If an exactly-once guarantee is truly required, operators can still perform these steps manually, right?

I’m not sure what it would take to fully bake suspend + handoff semantics into this API, but that’s something we can evolve in the future if needed. In the interim, as the docs call out, operators can still:

suspend supervisor → wait for tasks to checkpoint and complete → kick off backfill supervisor with checkpoints & reset main supervisor

(Fwiw, `resetOffsetsAutomatically` and hard resets have similar caveats around exactly-once semantics)

docs/api-reference/supervisor-api.md
```diff
@@ -3539,6 +3539,109 @@ when the supervisor's tasks restart, they resume reading from `{"0": 100, "1": 1
...
+*Failed to retrieve stream offsets or serialize the backfill spec*
+
+</TabItem>
+</Tabs>
+
+---
+
+#### Sample request
+
+The following example resets a supervisor named `social_media` and starts a backfill supervisor with 2 tasks.
```
abhishekrb19, line 3609: For follow-up, what do you think about wiring this API to the web-console?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
aho135: Ahh yeah good point. I think `resetToLatestAndBackfill` is a more descriptive name for this. I'll make that update

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
aho135: Yeah, that seems more clean than parsing the tree

docs/api-reference/supervisor-api.md
aho135: The current code requires the Supervisor to be in RUNNING state. To get this working with SUSPENDED we can drop that requirement. The usage is a bit clunky though. It would look something like

1. Suspend the supervisor and wait for all tasks to complete
2. Call resetToLatestAndBackfill. Since the backfill supervisor reuses the main supervisor spec it also gets created in SUSPENDED state
3. Manually resume both supervisors

Some minor modifications also need to be made to Kafka's `updatePartitionLagFromStream`. This function doesn't work in isolation as it requires first calling `recordSupplier.assign(partitions)` otherwise `recordSupplier.seekToLatest(partitions)` will throw an `IllegalStateException`.  Currently this is handled by `assignRecordSupplierToPartitionIds()` in SeekableStreamSupervisor. Similar tweaks may be needed for Rabbit/Kinesis. I thought to leave this as a follow up if it's needed in the future

docs/api-reference/supervisor-api.md
aho135: Yeah good call, will do this in a follow up!

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
aho135: On second thought, I'm thinking to leave this as is since in this case we'd be discarding the returned value and the exception handling differs slightly

FrankChen021 (commented):

---

docs/api-reference/supervisor-api.md
```diff
@@ -3539,6 +3539,109 @@ when the supervisor's tasks restart, they resume reading from `{"0": 100, "1": 1
...
+
+#### URL
+
+`POST` `/druid/indexer/v1/supervisor/{supervisorId}/resetToLatestAndBackfill`
+
+#### Query parameters
+
+| Parameter | Type | Description | Default |
+|---------|---------|---------|---------|
+| `backfillTaskCount` | Integer | Number of parallel tasks for the backfill supervisor. If not specified, inherits `taskCount` from the source supervisor. | None |
```
abhishekrb19: Since there's an implicit default:
```suggestion
| `backfillTaskCount` | Integer | Number of parallel tasks for the backfill supervisor. | Defaults to `taskCount` from the source supervisor if not specified |
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -640,6 +640,50 @@ private Response handleResetRequest(
...
+  @ResourceFilters(SupervisorResourceFilter.class)
+  public Response resetToLatestAndBackfill(
+      @PathParam("id") final String id,
+      @QueryParam("backfillTaskCount") @Nullable final Integer backfillTaskCount
+  )
+  {
+    return handleResetAndBackfill(id, backfillTaskCount);
+  }
+
+  private Response handleResetAndBackfill(final String id, @Nullable final Integer backfillTaskCount)
```
abhishekrb19: Perhaps also rename all references of `handleResetAndBackfill` to `handleResetToLatestAndBackfill`?

docs/api-reference/supervisor-api.md
```diff
@@ -3539,6 +3539,109 @@ when the supervisor's tasks restart, they resume reading from `{"0": 100, "1": 1
   ```
 </details>
 
+### Reset offsets and start a backfill supervisor
+
+This endpoint is supported for Apache Kafka and RabbitMQ Stream supervisors. Amazon Kinesis is not supported yet.
+
+Resets the supervisor to the latest available stream offsets and starts a new bounded backfill supervisor to ingest the data in the skipped range.
+
+This endpoint is useful when a supervisor has fallen behind and you want to catch it up to the latest offsets without losing the skipped data. The main supervisor resumes ingesting from the latest offsets, while the backfill supervisor processes the range from the previously checkpointed offsets up to the latest offsets at the time of the reset.
```
abhishekrb19, line 3548: I think the this section's title could also be:
`Reset offset to latest and start a backfill supervisor`

server/src/main/java/org/apache/druid/rpc/indexing/OverlordClient.java
```diff
@@ -197,6 +197,15 @@ ListenableFuture<CloseableIterator<TaskStatusPlus>> taskStatuses(
...
   ListenableFuture<Map<String, String>> terminateSupervisor(String supervisorId);
 
+  /**
+   * Resets a supervisor to the latest stream offsets and starts a bounded backfill supervisor.
+   * <p>
+   * API: {@code POST /druid/indexer/v1/supervisor/<id>/resetToLatestAndBackfill}
+   *
+   * @return Map containing "id" and "backfillSupervisorId"
+   */
+  ListenableFuture<Map<String, Object>> resetSupervisorAndBackfill(String supervisorId);
```
abhishekrb19: Same here & all implementations - it'd be good to align on the naming (reset vs resetToLatest)

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +373,112 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
     return true;
   }
 
+  /**
+   * Resets a supervisor to the latest stream offsets and starts a bounded backfill supervisor to
+   * process the skipped range from the previously checkpointed offsets up to the latest offsets.
```
abhishekrb19, line 378: Do you think it’s worth leaving the caveat as a comment here in the javadoc or before the reset operation is invoked that for correctness the supervisor need to be suspended/tasks checkpoint?

In case someone wants to take a stab at it in the future, they’d have a reference point on where to adjust things

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -393,6 +373,112 @@ public boolean resetSupervisor(String id, @Nullable DataSourceMetadata resetData
...
+   *                                  or if the backfill spec cannot be serialized
+   */
+  public Map<String, Object> resetSupervisorAndBackfill(String id, @Nullable Integer backfillTaskCount)
+  {
+    Preconditions.checkState(started, "SupervisorManager not started");
+    Preconditions.checkNotNull(id, "id");
+
+    Pair<Supervisor, SupervisorSpec> supervisorPair = supervisors.get(id);
+
+    if (!(supervisorPair.lhs instanceof SeekableStreamSupervisor)) {
```
abhishekrb19: Claude mentioned that this could NPE without a null check after calling `supervisors.get(id)`.

I don’t think it can happen in practice bc `!manager.getSupervisorIds().contains(id)` in `SupervisorResource `would return earlier, but for safety, could we add a null check in the manager itself? Looks like the other methods in the manager have similar checks as well

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
```diff
@@ -292,6 +292,47 @@ public void test_boundedSupervisor_doesNotSilentlyCompleteWhenStaleOffsetExceeds
...
+    waitForSupervisorDetailedState(supervisor.getId(), "RUNNING");
+
+    final int totalRecords = publish1kRecords(topic, false);
+    waitUntilPublishedRecordsAreIngested(totalRecords);
+
+    // Reset the main supervisor and spin up a backfill supervisor.
+    // Since all records are already ingested before the call, the backfill
+    // supervisor will complete immediately without ingesting anything.
+    final Map<String, Object> result = cluster.callApi().resetSupervisorAndBackfill(supervisor.getId());
+    final String backfillSupervisorId = (String) result.get("backfillSupervisorId");
```
abhishekrb19, line 322: For completeness, also validate that the primary supervisor ID is also included in the response of this API?

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
```diff
@@ -292,6 +292,47 @@ public void test_boundedSupervisor_doesNotSilentlyCompleteWhenStaleOffsetExceeds
...
+
+    waitForSupervisorDetailedState(supervisor.getId(), "RUNNING");
+
+    final int totalRecords = publish1kRecords(topic, false);
+    waitUntilPublishedRecordsAreIngested(totalRecords);
+
+    // Reset the main supervisor and spin up a backfill supervisor.
+    // Since all records are already ingested before the call, the backfill
+    // supervisor will complete immediately without ingesting anything.
+    final Map<String, Object> result = cluster.callApi().resetSupervisorAndBackfill(supervisor.getId());
```
abhishekrb19: Yeah, I’m not sure how to induce lag in this setup easily :) 

In addition to assertions on the supervisor state, it’d be nice if we could also assert on the records ingested/offset ranges or something similar even if the backfill ends up ingesting 0 records in this case.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorSpec.java
```diff
@@ -173,6 +174,59 @@ protected KafkaSupervisorSpec toggleSuspend(boolean suspend)
     );
   }
 
+  @Override
+  public KafkaSupervisorSpec createBackfillSpec(
```
abhishekrb19, line 178: This looks much cleaner - thanks for adding this!

abhishekrb19 (approved):
Left a final set of minor comments, but otherwise this looks good to me - thanks @aho135!

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/KafkaBoundedSupervisorTest.java
aho135: Yeah, I would have like to do that but in this case there is no actual task spun up since the Supervisor just immediately detects that no work needs to be done

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
aho135: I think that since the docs already call out the race and the functions are now all aligned with the endpoint name it should be easy to find the adjustment points

PR #19484, opened 2026-05-20
fix: Peon metrics emit same value

GWphua:
### Description

This PR fixes incorrect Peon sink-level query metric emission in SinkQuerySegmentWalker. This bug exists since v32, introduced by #17170 

#### The Bug
Currently, the following Peon metrics will be emitted with misleading values, making the three metrics appear identical:
- query/segment/time
- query/segmentAndCache/time
- query/wait/time

image-20260519165853832

#### Cause
The existing code iterated over METRICS_TO_REPORT and used a switch without break statements. Because reportMetric.getValue() is bound to the current metric reporter, fallthrough caused later accumulator values to be reported through the wrong reporter. For example, the query/segment/time reporter could be called with segment time, then wait time, then segment-and-cache time before emit(). Since DefaultQueryMetrics stores metric values by metric name before emission, the last value overwrote earlier ones.

#### Release note
Fix incorrect reporting of `query/segment/time`, `query/wait/time` and `query/segmentAndCache/time` on Peon.

<hr>

 * `SinkQuerySegmentWalker`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
+++ b/server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
@@ -35,6 +35,7 @@ import org.apache.druid.java.util.common.Pair;
 import org.apache.druid.java.util.common.granularity.Granularities;
 import org.apache.druid.java.util.metrics.StubServiceEmitter;
 import org.apache.druid.metadata.PendingSegmentRecord;
+import org.apache.druid.query.DefaultQueryMetrics;
 import org.apache.druid.query.Druids;
 import org.apache.druid.query.Order;
 import org.apache.druid.query.QueryPlus;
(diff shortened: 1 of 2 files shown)


jtuglu1 (approved):
Thanks. Will this test change catch any regressions as well?

GWphua:
Unfortunately, no. 

Will find some time to add that

GWphua:
Hi @jtuglu1 

I looked at trying to test the regression, but this may break abstraction. I am looking to bring these 3 lines of code into a function `reportSegmentMetrics(QueryMetrics<?>, SegmentMetrics)`. 

```
queryMetrics.reportSegmentTime(segmentMetrics.getSegmentTime());
queryMetrics.reportWaitTime(segmentMetrics.getWaitTime());
queryMetrics.reportSegmentAndCacheTime(segmentMetrics.getSegmentAndCacheTime());
```

However, the `SegmentMetrics` class is nested within the private query runner. 

Given that the fix removes the fall-through switch and replaces it with direct reporter-to-value calls, I think that it is ok to omit the tests. What do you think?

server/src/test/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorTest.java
```diff
@@ -2296,14 +2297,14 @@ public void testQueryBySegments() throws Exception
   private void verifySinkMetrics(StubServiceEmitter emitter, Set<String> segmentIds)
   {
     int segments = segmentIds.size();
-    emitter.verifyEmitted("query/cpu/time", 1);
-    Assert.assertEquals(segments, emitter.getMetricEvents("query/segment/time").size());
-    Assert.assertEquals(segments, emitter.getMetricEvents("query/segmentAndCache/time").size());
-    Assert.assertEquals(segments, emitter.getMetricEvents("query/wait/time").size());
+    emitter.verifyEmitted(DefaultQueryMetrics.QUERY_CPU_TIME, 1);
+    Assert.assertEquals(segments, emitter.getMetricEvents(DefaultQueryMetrics.QUERY_SEGMENT_TIME).size());
```
FrankChen021, line 2301: [P3] Test helper does not cover the fixed metric-value regression

The modified helper still only checks that each metric name was emitted for each segment. The old fall-through bug also emitted query/segment/time, query/wait/time, and query/segmentAndCache/time with the right segment dimensions, but overwrote their values through the wrong reporter, so this test would still pass with the buggy implementation. Please add coverage that verifies the reporter-to-accumulator mapping, otherwise the exact regression fixed in SinkQuerySegmentWalker can return unnoticed.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

---

PR #19486, opened 2026-05-20
feat: add realtimeSegmentsMode query context param

jtuglu1:
### Description

A prior [commit](https://github.com/apache/druid/commit/96bd4ebf70b9cbac81efc96ce653b63bdced6888) added the `realtimeSegmentsOnly` query context flag, which wasn't exhaustive in the 3 possible options users might want (include realtime segments, exclude realtime segments, exclusively query realtime segments). This adds the query context parameter `realtimeSegmentsMode` and deprecates `realtimeSegmentsOnly`. `realtimeSegmentsOnly=true` maps to `realtimeSegmentsOnly=exclusive` and `realtimeSegmentsOnly=false` maps to `realtimeSegmentsOnly=include`.

In addition to just being able to query realtime segments, this is useful when performing things like blue/green deployments and you only want to query new historical replica ASGs and not touch any "live" nodes (neither realtime nor historical). 

#### Release note

This adds the query context parameter `realtimeSegmentsMode` and deprecates `realtimeSegmentsOnly`. realtimeSegmentsOnly=true maps to realtimeSegmentsOnly=exclusive and realtimeSegmentsOnly=false maps to realtimeSegmentsOnly=include.

This is useful when performing things like blue/green deployments and you only want to query new historical replica ASGs and not touch any "live" nodes (neither realtime nor historical).

<hr>

This PR has:

- [ ] been self-reviewed.


FrankChen021 (commented):

---

docs/querying/query-context-reference.md
```diff
@@ -71,7 +71,8 @@ Unless otherwise noted, the following parameters apply to all query types, and t
...
-|`realtimeSegmentsOnly` |`false`| When set to true, only query realtime segments. Historical segments are excluded. |
+|`realtimeSegmentsMode` |`include`| Controls which segments are queried, classified by whether a historical replica exists. `include` queries all segments. `exclusive` queries only segments served solely by realtime servers; any segment with at least one historical replica (including segments mid-handoff) is excluded. `exclude` is the inverse: segments served solely by realtime servers are skipped, but segments mid-handoff that have both a realtime and a historical replica are still queried. |
```
abhishekrb19: >`exclude` is the inverse: segments served solely by realtime servers are skipped, but segments mid-handoff that have both a realtime and a historical replica are still queried.


I don't know what mid-handoff is. Wouldn't this be the case where the segment is either fully loaded on the realtime server or the historical, it's not and Broker would deterministically choose one or the other as it sees in its server inventory?

processing/src/main/java/org/apache/druid/query/QueryContexts.java
```diff
@@ -233,6 +244,39 @@ public String toString()
...
+        return null;
+      }
+      return RealtimeSegmentsMode.valueOf(StringUtils.toUpperCase(str));
+    }
+
+    @Override
+    @JsonValue
+    public String toString()
+    {
+      return StringUtils.toLowerCase(name());
```
abhishekrb19, line 274: For consistency with `fromString()`:
```suggestion
      return StringUtils.toUpperCase(name());
```

processing/src/test/java/org/apache/druid/query/QueryContextTest.java
```diff
@@ -440,6 +441,66 @@ public void testIsRealtimeSegmentsOnly()
...
+
+  @Test
+  public void testGetRealtimeSegmentsModeInvalidValue()
+  {
+    BadQueryContextException e = assertThrows(
+        BadQueryContextException.class,
+        () -> QueryContext.of(ImmutableMap.of(QueryContexts.REALTIME_SEGMENTS_MODE, "badvalue"))
+                         .getRealtimeSegmentsMode()
+    );
+    assertTrue(e.getMessage().contains("realtimeSegmentsMode"));
```
abhishekrb19: nit: can we assert the exception msg beyond `contains()` (or catch & throw a `DruidException` if its appropriate)

processing/src/main/java/org/apache/druid/query/QueryContext.java
```diff
@@ -781,8 +782,34 @@ public boolean isPrePlanned()
     return getBoolean(QueryContexts.CTX_PREPLANNED, QueryContexts.DEFAULT_PREPLANNED);
   }
 
+  /**
+   * Returns the realtime segments mode for this query. If {@code realtimeSegmentsMode} is absent
```
abhishekrb19: Perhaps use `@link` for `realtimeSegmentsMode` variable instead of `@code` in the javadocs so it's clear how these things are referenced & used by callers.

processing/src/main/java/org/apache/druid/query/QueryContext.java
```diff
@@ -781,8 +782,34 @@ public boolean isPrePlanned()
...
+   * or null, falls back to the deprecated {@code realtimeSegmentsOnly} boolean: {@code true} maps
+   * to {@link RealtimeSegmentsMode#EXCLUSIVE}; otherwise returns {@link RealtimeSegmentsMode#INCLUDE}.
+   */
+  public RealtimeSegmentsMode getRealtimeSegmentsMode()
+  {
+    RealtimeSegmentsMode mode = getEnum(
+        QueryContexts.REALTIME_SEGMENTS_MODE,
+        RealtimeSegmentsMode.class,
+        null
+    );
```
abhishekrb19, line 797: Should we just pass this default enum value here as-is?
```suggestion
    RealtimeSegmentsMode mode = getEnum(
        QueryContexts.REALTIME_SEGMENTS_MODE,
        RealtimeSegmentsMode.class,
        QueryContexts.DEFAULT_REALTIME_SEGMENTS_MODE
    );
```

IMO the backward-compatible behavior here can be to error out if both the new field and the old deprecated field are set, instead of introducing precedence rules, since that can become confusing (and users would also have to explicitly opt in to the new property anyway)

abhishekrb19 (commented):
This makes sense. Left a few comments. It’d be good for @aho135 to take a look at this as well.

aho135:
Thanks for the update @jtuglu1 This makes sense to me

Shall we go ahead and update the [auto-completion](https://github.com/apache/druid/blob/master/web-console/src/dialogs/edit-context-dialog/query-context-completions.ts#L75-L78) as well?

docs/querying/query-context-reference.md
```diff
@@ -71,7 +71,8 @@ Unless otherwise noted, the following parameters apply to all query types, and t
...
-|`realtimeSegmentsOnly` |`false`| When set to true, only query realtime segments. Historical segments are excluded. |
+|`realtimeSegmentsMode` |`include`| Controls whether realtime segments are queried. `include` queries all segments, including realtime. `exclude` skips realtime segments. `exclusive` queries only realtime segments. |
+|`realtimeSegmentsOnly` |`false`| **Deprecated.** Use `realtimeSegmentsMode=exclusive` instead. When set to `true`, equivalent to `realtimeSegmentsMode=exclusive`. |
```
aho135:
```suggestion
|`realtimeSegmentsOnly` |`false`| **Deprecated.** Use `realtimeSegmentsMode=exclusive` instead. When set to `true`, equivalent to `realtimeSegmentsMode=exclusive`. When set to `false`, equivalent to `realtimeSegmentsMode=include`.|
```
Just for completeness

docs/querying/query-context-reference.md
jtuglu1: Yeah this was overly verbose – I've trimmed it down.

web-console/src/dialogs/edit-context-dialog/query-context-completions.ts
```diff
@@ -72,9 +72,13 @@ export const QUERY_CONTEXT_COMPLETIONS: JsonCompletionRule[] = [
         value: 'vectorize',
         documentation: 'Enable vectorized query execution',
       },
+      {
+        value: 'realtimeSegmentsMode',
+        documentation: 'Controls whether realtime segments are queried',
+      },
       {
         value: 'realtimeSegmentsOnly',
```
aho135: What do you think about removing `realtimeSegmentsOnly` from auto-complete to discourage usage?

jtuglu1: sure

docs/querying/query-context-reference.md
```diff
@@ -71,7 +71,8 @@ Unless otherwise noted, the following parameters apply to all query types, and t
...
-|`realtimeSegmentsOnly` |`false`| When set to true, only query realtime segments. Historical segments are excluded. |
+|`realtimeSegmentsMode` |`include`| Controls whether realtime segments are queried. `include` queries all segments, including realtime. `exclude` skips realtime segments. `exclusive` queries only realtime segments. |
+|`realtimeSegmentsOnly` |`false`| **Deprecated.** Use `realtimeSegmentsMode=exclusive` instead. When set to `true`, equivalent to `realtimeSegmentsMode=exclusive`. When set to `false`, equivalent to `realtimeSegmentsMode=include`.|
```
abhishekrb19:
```suggestion
|`realtimeSegmentsOnly` |`false`| **Deprecated.** Use `realtimeSegmentsMode=exclusive` instead. When set to `true`, this is equivalent to `realtimeSegmentsMode=exclusive`. When set to `false`, this is equivalent to `realtimeSegmentsMode=include`.|
```

abhishekrb19 (approved):
Lgtm - thanks!

processing/src/main/java/org/apache/druid/query/QueryContexts.java
jtuglu1: I want to keep this lowercase as it's more inline with the docs and prior query context enum values.

PR #19494, opened 2026-05-21
minor: Get StorageMonitor metrics from StorageLocations.

gianm:
This patch changes StorageMonitor to get metrics from StorageLocations directly, rather than through SegmentLocalCacheManager. This simplifies the logic by removing an unnecessary layer. It also ensures that metrics are reported properly no matter how the StorageLocations are accessed.


server/src/main/java/org/apache/druid/segment/loading/VirtualStorageLocationStats.java
```diff
@@ -86,4 +86,23 @@ public interface VirtualStorageLocationStats
...
   long getRejectCount();
+
+  /**
+   * Whether any stats are nonzero.
+   */
+  default boolean hasStats()
+  {
+    return getUsedBytes() != 0
+           || getHitCount() != 0
+           || getHoldBytes() != 0
```
FrankChen021, line 97: [P2] Include hold count when deciding whether to emit virtual stats

StorageMonitor now skips all virtual-storage emissions unless `weakStats.hasStats()` returns true, but this predicate checks `getHitCount()` twice and never checks `getHoldCount()`. If a virtual cache location has active holds with zero held bytes, which can happen for zero-sized/tombstone segments, `storage/virtual/hold/count` is nonzero but the whole virtual metric block is suppressed. Include `getHoldCount()` in this predicate so the hold-count metric is emitted whenever it is nonzero.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -473,6 +471,10 @@ public AcquireSegmentAction acquireSegment(final DataSegment dataSegment)
       return acquireExisting;
     }
 
+    if (!config.isVirtualStorage()) {
+      return AcquireSegmentAction.missingSegment();
+    }
```
capistrant: this feels off topic, can you provide some context? this also seems to make lines 486 to 488 below dead code, if it is indeed safe to hoist this block up out of the synchronized block

capistrant (commented):
metrics parts +1. left non-stats question about the diff in SegmentLocalCacheManager

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
gianm: Right, this isn't related to the main change, it's an opportunistic change to one of the related files. I'll pull it out and do it in a separate PR.

gianm: https://github.com/apache/druid/pull/19503

clintropolis (approved):

PR #19500, opened 2026-05-21
refactor: remove deprecated zookeeper-based task runner

clintropolis:
### Description
The zk-based `RemoteTaskRunner` (`druid.indexer.runner.type=remote`) is deprecated and off by default with `HttpRemoteTaskRunner` (`druid.indexer.runner.type=httpRemote`) has been the default for some time. This PR finishes the migration and removes it and related supporting stuff. Together with the recently merged #19377, this leaves ZooKeeper responsible for only:

1. Coordinator / Overlord leader election
2. Service announcement and discovery (`DruidNodeAnnouncer` / internal-discovery)

Default behavior does not change. Operators currently using `druid.indexer.runner.type=remote` must switch to `httpRemote` or some other runner such as the `k8s` extension.

`StartupInjectorBuilder.PropertiesValidator` now rejects the stale value at startup with a clear pointer to the upgrade notes.

#### ZooKeeper-based task runner removed

The ZooKeeper-based `RemoteTaskRunner` (`druid.indexer.runner.type=remote`) has been removed. The HTTP-based `httpRemote` runner has been the default since Druid 25.0.0 and is now the only supported distributed task runner. `local` (in-process) remains supported for single-process testing, and the `k8s` runner from the Kubernetes extension is unaffected.

If your configuration sets `druid.indexer.runner.type=remote`, startup will now fail with a clear error message. Remove the property (or set it to `httpRemote`, which is the default) to proceed.

The following configuration properties are no longer recognized and should be removed from `common.runtime.properties`:

* `druid.indexer.runner.maxZnodeBytes`
* `druid.indexer.runner.taskShutdownLinkTimeout`
* `druid.indexer.runner.compressZnodes`
* `druid.zk.paths.indexer.base`
* `druid.zk.paths.indexer.announcementsPath`
* `druid.zk.paths.indexer.tasksPath`
* `druid.zk.paths.indexer.statusPath`

ZooKeeper is still used for Coordinator/Overlord leader election and service (node) announcement and discovery.


jtuglu1 (approved):
Pre-approving since I did a portion of this already when doing some http remote task runner changes.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

Reviewed 60 of 60 changed files.

---

- docs/development/extensions-core/k8s-jobs.md:1022 - [P3] Remove stale remote workerType from k8s docs. This table still advertises `remote` as a possible value for `druid.indexer.runner.k8sAndWorker.runnerStrategy.workerType`, but this PR changes `KubernetesAndWorkerTaskRunnerConfig` to reject every value except `httpRemote`. Operators following this doc will now configure a value that fails Overlord startup, so the possible values should be updated to only list `httpRemote` and, if useful, mention that the ZooKeeper-backed `remote` worker type was removed.

PR #19501, opened 2026-05-22
feat: add etag conditional write support to dynamic config

jtuglu1:
### Description

Currently, competing writes to Druid dynamic configs can race with each other. This introduces the concept of conditional writes with etags, allowing clients some concurrency control over writes to Druid's dynamic config.

Callers can optionally provide an E-Tag header `If-Match: <ETag>` in dynamic config payloads which the server will match against and reject with a 412 code (preconditions failed) instead of potentially returning 200 success (racy overwrite). Clients can retrieve a valid E-Tag value from issuing GETs on the corresponding dynamic config endpoint (read-modify-write).

#### Release note

Add E-Tag conditional write support to dynamic config APIs

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/common/config/ConfigEtag.java
+++ b/processing/src/main/java/org/apache/druid/common/config/ConfigEtag.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.common.config;
+
+import javax.annotation.Nullable;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.Base64;
+
+/**
+ * Computes and compares ETags for stored dynamic config payloads. The ETag is a
+ * pure function of the payload bytes — used for {@code If-Match} preconditions.
+ */
+public final class ConfigEtag
+{
+  private static final int ETAG_HASH_BYTES = 16;
+
+  private ConfigEtag()
+  {
+  }
+
+  /**
+   * Quoted ETag for the given payload bytes, or {@code null} if {@code bytes}
+   * is {@code null}. SHA-256 truncated to {@value #ETAG_HASH_BYTES} bytes,
+   * base64url-encoded (no padding), wrapped in double quotes per RFC 7232.
+   */
+  @Nullable
+  public static String compute(@Nullable byte[] bytes)
+  {
+    if (bytes == null) {
+      return null;
+    }
+    try {
+      MessageDigest md = MessageDigest.getInstance("SHA-256");
+      byte[] full = md.digest(bytes);
+      byte[] truncated = Arrays.copyOf(full, ETAG_HASH_BYTES);
+      return "\"" + Base64.getUrlEncoder().withoutPadding().encodeToString(truncated) + "\"";
+    }
+    catch (NoSuchAlgorithmException e) {
+      // SHA-256 is required by every JRE.
+      throw new IllegalStateException("SHA-256 not available", e);
+    }
+  }
+
+  /**
+   * Whether {@code ifMatchHeader} matches the ETag of {@code currentBytes}.
+   * Wildcard {@code *} matches any existing value. A comma-separated list is
+   * satisfied if any element matches.
+   */
+  public static boolean matches(String ifMatchHeader, @Nullable byte[] currentBytes)
+  {
+    if (ifMatchHeader == null) {
+      return true;
+    }
+    final String trimmed = ifMatchHeader.trim();
+    if ("*".equals(trimmed)) {
+      return currentBytes != null;
+    }
+    final String currentEtag = compute(currentBytes);
+    if (currentEtag == null) {
+      return false;
+    }
+    for (String candidate : trimmed.split(",")) {
+      if (currentEtag.equals(candidate.trim())) {
+        return true;
+      }
+    }
+    return false;
+  }
+}
--- a/processing/src/main/java/org/apache/druid/common/config/ConfigEtag.java
+++ b/processing/src/main/java/org/apache/druid/common/config/ConfigEtag.java
@@ -32,6 +32,14 @@ import java.util.Base64;
 public final class ConfigEtag
 {
   private static final int ETAG_HASH_BYTES = 16;
+  private static final ThreadLocal<MessageDigest> SHA_256 = ThreadLocal.withInitial(() -> {
+    try {
+      return MessageDigest.getInstance("SHA-256");
+    }
+    catch (NoSuchAlgorithmException e) {
+      throw new IllegalStateException("SHA-256 not available", e);
+    }
+  });
 
   private ConfigEtag()
   {
@@ -48,16 +56,11 @@ public final class ConfigEtag
     if (bytes == null) {
       return null;
     }
-    try {
-      MessageDigest md = MessageDigest.getInstance("SHA-256");
-      byte[] full = md.digest(bytes);
-      byte[] truncated = Arrays.copyOf(full, ETAG_HASH_BYTES);
-      return "\"" + Base64.getUrlEncoder().withoutPadding().encodeToString(truncated) + "\"";
-    }
-    catch (NoSuchAlgorithmException e) {
-      // SHA-256 is required by every JRE.
-      throw new IllegalStateException("SHA-256 not available", e);
-    }
+    final MessageDigest md = SHA_256.get();
+    md.reset();
+    final byte[] full = md.digest(bytes);
+    final byte[] truncated = Arrays.copyOf(full, ETAG_HASH_BYTES);
+    return "\"" + Base64.getUrlEncoder().withoutPadding().encodeToString(truncated) + "\"";
   }
 
   /**
@@ -65,7 +68,7 @@ public final class ConfigEtag
    * Wildcard {@code *} matches any existing value. A comma-separated list is
    * satisfied if any element matches.
    */
-  public static boolean matches(String ifMatchHeader, @Nullable byte[] currentBytes)
+  public static boolean matches(@Nullable String ifMatchHeader, @Nullable byte[] currentBytes)
   {
     if (ifMatchHeader == null) {
       return true;
--- a/processing/src/main/java/org/apache/druid/common/config/ConfigManager.java
+++ b/processing/src/main/java/org/apache/druid/common/config/ConfigManager.java
@@ -181,6 +181,30 @@ public class ConfigManager
     return set(key, serde, null, obj);
   }
 
+  /**
+   * Returns the raw payload bytes currently stored for {@code key}, or
+   * {@code null} if none. Prefers the in-memory watched copy; falls back to the
+   * metadata store. Returns a defensive copy — the cached array backs equality
+   * checks and CAS oldValue payloads.
+   */
+  @Nullable
+  public byte[] getCurrentBytes(final String key)
+  {
+    final ConfigHolder<?> holder = watchedConfigs.get(key);
+    if (holder != null) {
+      final byte[] cached = holder.rawBytes.get();
+      if (cached != null) {
+        return cached.clone();
+      }
+    }
+    return dbConnector.lookup(configTable, "name", "payload", key);
+  }
+
+  public boolean isCompareAndSwapEnabled()
+  {
+    return config.get().isEnableCompareAndSwap();
+  }
+
   public <T> SetResult set(final String key, final ConfigSerde<T> serde, @Nullable final byte[] oldValue, final T newObject)
   {
     if (newObject == null || !started) {
@@ -259,10 +283,11 @@ public class ConfigManager
 
   public static class SetResult
   {
-    private static final SetResult SUCCESS = new SetResult(null, false);
+    private static final SetResult SUCCESS = new SetResult(null, false, false);
     private final Exception exception;
 
     private final boolean retryableException;
+    private final boolean preconditionFailed;
 
     public static SetResult ok()
     {
@@ -271,18 +296,25 @@ public class ConfigManager
 
     public static SetResult failure(Exception e)
     {
-      return new SetResult(e, false);
+      return new SetResult(e, false, false);
     }
 
     public static SetResult retryableFailure(Exception e)
     {
-      return new SetResult(e, true);
+      return new SetResult(e, true, false);
     }
 
-    private SetResult(@Nullable Exception exception, boolean retryableException)
+    /** Client-supplied precondition (e.g. {@code If-Match}) failed; maps to HTTP 412. */
+    public static SetResult preconditionFailed(Exception e)
+    {
+      return new SetResult(e, false, true);
+    }
+
+    private SetResult(@Nullable Exception exception, boolean retryableException, boolean preconditionFailed)
     {
       this.exception = exception;
       this.retryableException = retryableException;
+      this.preconditionFailed = preconditionFailed;
     }
 
     public boolean isOk()
@@ -293,6 +325,11 @@ public class ConfigManager
     public boolean isRetryable()
     {
       return retryableException;
+    }
+
+    public boolean isPreconditionFailed()
+    {
+      return preconditionFailed;
     }
 
     public Exception getException()
--- a/processing/src/main/java/org/apache/druid/common/config/ConfigManager.java
+++ b/processing/src/main/java/org/apache/druid/common/config/ConfigManager.java
@@ -283,11 +283,27 @@ public class ConfigManager
 
   public static class SetResult
   {
-    private static final SetResult SUCCESS = new SetResult(null, false, false);
-    private final Exception exception;
+    /**
+     * Outcome of a {@link #set} attempt. Mutually exclusive, so callers never
+     * have to reconcile overlapping flags.
+     */
+    private enum Status
+    {
+      /** The write committed. */
+      OK,
+      /** A concurrent writer won the CAS; the caller may re-read and retry. */
+      RETRYABLE,
+      /** A client-supplied precondition (e.g. {@code If-Match}) failed; maps to HTTP 412. */
+      PRECONDITION_FAILED,
+      /** A non-retryable failure (bad input, manager not started, etc.). */
+      FAILURE
+    }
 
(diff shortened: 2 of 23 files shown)


FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

---

processing/src/main/java/org/apache/druid/common/config/JacksonConfigManager.java
```diff
@@ -127,6 +127,54 @@ public <T> SetResult set(
...
+    if (ifMatchEtag == null) {
+      return set(key, newValue, auditInfo);
+    }
+    final byte[] currentBytes = configManager.getCurrentBytes(key);
+    if (!ConfigEtag.matches(ifMatchEtag, currentBytes)) {
+      return SetResult.preconditionFailed(
+          new IllegalStateException("If-Match precondition failed for key[" + key + "]")
+      );
+    }
+    final SetResult result = set(key, currentBytes, newValue, auditInfo);
```
FrankChen021: [P1] Conditional writes can overwrite newer configs when CAS is disabled

`setIfMatch` validates `If-Match` against `currentBytes`, but then relies on `ConfigManager.set(..., currentBytes, ...)` to enforce that value at commit time. `ConfigManager.set` ignores `oldValue` whenever `druid.manager.config.enableCompareAndSwap` is false, so in that supported mode a writer with a stale but locally matching ETag can still `insertOrUpdate` over a concurrent change and receive 200. That violates the advertised `If-Match` contract and can silently lose dynamic-config updates; conditional requests should fail when CAS is disabled or use a storage-level compare-and-swap before returning success.

server/src/main/java/org/apache/druid/server/http/DynamicConfigEtagHelper.java
```diff
@@ -0,0 +1,83 @@
...
+  {
+    if (req == null) {
+      return null;
+    }
+    final String header = req.getHeader(HttpHeaders.IF_MATCH);
+    if (header == null) {
+      return null;
+    }
+    if (header.trim().isEmpty()) {
+      throw InvalidInput.exception("If-Match header must not be blank");
```
FrankChen021, line 62: [P2] Blank If-Match escapes as an unhandled DruidException

`getIfMatch` documents a 400 for a blank `If-Match` header, but it throws `InvalidInput.exception` directly. Several new callers invoke it outside `ServletResourceUtils.buildReadResponse` and only catch `IllegalArgumentException` or do not catch at all, so a request with a present-but-blank `If-Match` escapes the resource method instead of producing the documented 400 response. Catch `DruidException` at the call sites or make the helper return an error response/result that callers handle explicitly.

processing/src/main/java/org/apache/druid/common/config/JacksonConfigManager.java
jtuglu1: Yeah – this is minor IMO. I don't think anyone is running with `enableCompareAndSwap=false`, given it's not even a publicly documented configuration. If they are, I would expect they don't require any sort of correctness under concurrency. I suppose we can throw an exception if an ETag is provided but `enableCompareAndSwap=false`.

FrankChen021: Thanks, the new guard fixes the `JacksonConfigManager.setIfMatch` callers, but this still looks open for compaction config writes. `CoordinatorConfigManager.getAndUpdateCompactionConfig` does its own `If-Match` check and then calls `jacksonConfigManager.set(..., currentBytes, updated, ...)`; when `enableCompareAndSwap=false`, `ConfigManager.set` still ignores that `oldValue` and performs `insertOrUpdate`, so the compaction endpoints that advertise `If-Match` can still overwrite a concurrent update and return success. Can we apply the same disabled-CAS precondition failure to that path as well? Reviewed 23 of 23 changed files.

jtuglu1: Updated – can I get a review?

server/src/main/java/org/apache/druid/server/http/CoordinatorCompactionConfigsResource.java
```diff
@@ -60,9 +60,16 @@ public CoordinatorCompactionConfigsResource(
...
   public Response getCompactionConfig()
   {
-    return ServletResourceUtils.buildReadResponse(
+    final Response baseResponse = ServletResourceUtils.buildReadResponse(
         configManager::getCurrentCompactionConfig
     );
+    if (baseResponse.getStatus() != Response.Status.OK.getStatusCode()) {
+      return baseResponse;
+    }
+    return DynamicConfigEtagHelper.withEtag(
```
FrankChen021: [P1] ETag can describe a different config than the response body

The ETagged GET path builds the response entity from `getCurrentCompactionConfig()` and then separately reads `getCurrentCompactionConfigBytes()` for the ETag. If another config update lands between those two reads, the response can contain the old config body with the new stored bytes' ETag. A client can then edit that stale body and submit it with `If-Match` using the fresh ETag, so the precondition passes and the unseen concurrent update is overwritten. The same body/bytes split appears in the other ETagged dynamic-config GET handlers; these should return an atomic bytes+decoded-config snapshot, or derive both body and ETag from the same byte array.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Follow-up checked: the disabled-CAS compaction path now returns a precondition failure for conditional writes, so the original issue is handled.

---

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Follow-up checked: the disabled-CAS compaction path now returns a precondition failure for conditional writes, and ETagged GET responses now derive the body and ETag from the same byte snapshot. I found one remaining concurrency issue in conditional partial updates.

---

processing/src/main/java/org/apache/druid/common/config/JacksonConfigManager.java
```diff
@@ -127,6 +127,66 @@ public <T> SetResult set(
...
+      return set(key, newValue, auditInfo);
+    }
+    if (!configManager.isCompareAndSwapEnabled()) {
+      return SetResult.preconditionFailed(
+          new IllegalStateException(
+              "If-Match requires druid.manager.config.enableCompareAndSwap to be enabled for key[" + key + "]"
+          )
+      );
+    }
+    final byte[] currentBytes = configManager.getCurrentBytes(key);
```
FrankChen021, line 174: [P1] Build conditional partial updates from the CAS snapshot

setIfMatch reads the bytes used for If-Match validation and CAS after callers have already built newValue. That is safe for full replacements, but the new Coordinator/Broker/K8s partial-update endpoints merge request fields with a watched config object before calling this method. If the watcher advances between that merge and this read, an If-Match for snapshot B can pass and CAS B -> newValue(A + patch), dropping fields introduced in B while returning success. Conditional partial writes need a transform-style path that reads bytes, validates the ETag, deserializes those same bytes, applies the patch, and CASes the same byte snapshot.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

I found one conditional-write gap: the unified datasource compaction POST/DELETE paths still ignore `If-Match` when they fall back to dynamic config writes.

- indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordCompactionResource.java:284 - [P1] Honor If-Match on unified datasource compaction writes. When compaction supervisors are disabled, the unified datasource POST and DELETE paths still mutate the same DruidCompactionConfig document but call the configManager overloads without passing DynamicConfigEtagHelper.getIfMatch(...). A client can read an ETag for the compaction document, another writer can change it, and then the stale client can POST or DELETE /druid/indexer/v1/compaction/config/datasources/... with If-Match; the header is ignored and the write succeeds instead of returning 412. Thread the If-Match value through these datasource update/delete calls, or explicitly reject unsupported conditional writes, so conditional requests do not silently degrade to last-writer-wins.

---

processing/src/main/java/org/apache/druid/common/config/ConfigEtag.java
```diff
@@ -0,0 +1,88 @@
...
+   * Wildcard {@code *} matches any existing value. A comma-separated list is
+   * satisfied if any element matches.
+   */
+  public static boolean matches(String ifMatchHeader, @Nullable byte[] currentBytes)
+  {
+    if (ifMatchHeader == null) {
+      return true;
+    }
+    final String trimmed = ifMatchHeader.trim();
+    if ("*".equals(trimmed)) {
```
cecemei, line 77: why does it need * for match? instead of not setting if match header.

processing/src/main/java/org/apache/druid/common/config/ConfigEtag.java
```diff
@@ -0,0 +1,88 @@
...
+      throw new IllegalStateException("SHA-256 not available", e);
+    }
+  }
+
+  /**
+   * Whether {@code ifMatchHeader} matches the ETag of {@code currentBytes}.
+   * Wildcard {@code *} matches any existing value. A comma-separated list is
+   * satisfied if any element matches.
+   */
+  public static boolean matches(String ifMatchHeader, @Nullable byte[] currentBytes)
```
cecemei: nit: `@Nullable String ifMatchHeader`

processing/src/main/java/org/apache/druid/common/config/ConfigManager.java
```diff
@@ -259,10 +283,11 @@ private MetadataCASUpdate createMetadataCASUpdate(
 
   public static class SetResult
   {
-    private static final SetResult SUCCESS = new SetResult(null, false);
+    private static final SetResult SUCCESS = new SetResult(null, false, false);
     private final Exception exception;
 
     private final boolean retryableException;
+    private final boolean preconditionFailed;
```
cecemei: The two-boolean design has an implicit invalid state (retryableException=true, preconditionFailed=true) and the semantics overlap in a confusing way, maybe consider use enum to represent the failure state. 
```
- RETRYABLE_FAILURE — caller can re-read the current value and retry automatically (CAS conflict, concurrent writer)
- PRECONDITION_FAILED — caller needs external input before retrying (stale ETag, requires client to re-fetch and resubmit — HTTP 412)
- FAILURE — neither
  ```

File: processing/src/main/java/org/apache/druid/common/config/JacksonConfigManager.java [hunk 8]
```diff
@@ -127,6 +128,112 @@ public <T> SetResult set(
...
+      );
+    }
+    final SetResult result = set(key, currentBytes, newValue, auditInfo);
+    // Retryable CAS failure here = concurrent writer between our read and CAS;
+    // surface as precondition failed since the caller asked us to reject that.
+    if (!result.isOk() && result.isRetryable()) {
+      return SetResult.preconditionFailed(
+          new IllegalStateException("If-Match precondition failed (concurrent update) for key[" + key + "]")
+      );
+    }
```
cecemei: This seems hacky, we should have that check/logic in `set(key, currentBytes, newValue, auditInfo)`, maybe pass a retryableIfCurrentBytesNotMatch boolean, or maybe return a stale value failure and let caller decides what to do.

server/src/main/java/org/apache/druid/server/coordinator/CoordinatorConfigManager.java
```diff
@@ -87,15 +88,62 @@ public CoordinatorDynamicConfig getCurrentDynamicConfig()
     return Preconditions.checkNotNull(dynamicConfig, "Got null config from watcher?!");
   }
 
+  public CoordinatorDynamicConfig convertBytesToDynamicConfig(@Nullable byte[] bytes)
+  {
+    return jacksonConfigManager.convertByteToConfig(
+        bytes,
+        CoordinatorDynamicConfig.class,
+        CoordinatorDynamicConfig.builder().build()
```
cecemei, line 96: is it necessary to pass in a default value here? would it be used to compute for ETag?

server/src/main/java/org/apache/druid/server/coordinator/CoordinatorConfigManager.java
```diff
@@ -87,15 +88,62 @@ public CoordinatorDynamicConfig getCurrentDynamicConfig()
...
+    );
+  }
+
   public ConfigManager.SetResult setDynamicConfig(CoordinatorDynamicConfig config, AuditInfo auditInfo)
   {
-    return jacksonConfigManager.set(
+    return setDynamicConfig(config, null, auditInfo);
+  }
+
+  public ConfigManager.SetResult setDynamicConfig(
```
cecemei: there's no usage on `setDynamicConfig` any more with changes in this pr?

server/src/main/java/org/apache/druid/server/coordinator/CoordinatorConfigManager.java
```diff
@@ -173,16 +254,25 @@ public boolean updateCompactionTaskSlots(
       return current.withClusterConfig(updatedClusterConfig);
     };
 
-    return updateConfigHelper(operator, auditInfo);
+    return updateConfigHelper(operator, ifMatchEtag, auditInfo);
+  }
+
+  public boolean updateClusterCompactionConfig(
```
cecemei, line 251: there's no usage in this function after this pr.

server/src/main/java/org/apache/druid/server/coordinator/CoordinatorConfigManager.java
```diff
@@ -268,14 +376,17 @@ public List<DataSourceCompactionConfigAuditEntry> getCompactionConfigHistory(
...
   private boolean updateConfigHelper(
       UnaryOperator<DruidCompactionConfig> configUpdateOperator,
+      @Nullable String ifMatchEtag,
       AuditInfo auditInfo
   )
   {
     int attemps = 0;
     ConfigManager.SetResult setResult = null;
+    // When the caller has supplied an If-Match precondition, do not retry on CAS failure.
+    final int maxAttempts = ifMatchEtag != null ? 1 : MAX_UPDATE_RETRIES;
```
cecemei: This feels strange, update could fail for other reasons too, maybe we dont want it to retry in these cases. the logic seems strange if Etag doesn't match but in line390 setResult is retryable.

processing/src/main/java/org/apache/druid/common/config/ConfigEtag.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Quoted ETag for the given payload bytes, or {@code null} if {@code bytes}
+   * is {@code null}. SHA-256 truncated to {@value #ETAG_HASH_BYTES} bytes,
+   * base64url-encoded (no padding), wrapped in double quotes per RFC 7232.
+   */
+  @Nullable
+  public static String compute(@Nullable byte[] bytes)
+  {
+    if (bytes == null) {
```
cecemei, line 56: imho it'd nice to use non null value maybe `__NULL__ `to represent that, WDYT?

jtuglu1: I think `null` for a null byte payload is fine. Any other sort of identifier would be more confusing IMO.

processing/src/main/java/org/apache/druid/common/config/ConfigManager.java
jtuglu1: Agreed – I'll update.

processing/src/main/java/org/apache/druid/common/config/ConfigEtag.java
jtuglu1: This behavior is in accordance with the [ETag](https://datatracker.ietf.org/doc/html/rfc7232#section-3.1) RFC. Not too relevant in common case, but still keeps things spec-correct.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

- indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordCompactionResource.java:284 - [P1] Honor If-Match on unified datasource compaction writes. When compaction supervisors are disabled, the unified datasource POST and DELETE fallback paths still mutate the same DruidCompactionConfig document through overloads that do not receive DynamicConfigEtagHelper.getIfMatch(...). A client can read the compaction ETag, race with another config write, then POST or DELETE /druid/indexer/v1/compaction/config/datasources/... with a stale If-Match header and still get a successful write instead of 412. Thread the If-Match value through both fallback calls, or explicitly reject conditional requests there. This appears unchanged from the prior review.

---

jtuglu1:
> Severity	Findings
> P0	0
> P1	1
> P2	0
> P3	0
> Total	1
> Reviewed 23 of 23 changed files.
> 
> Findings that could not be attached inline:
> 
> * indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordCompactionResource.java:284 - [P1] Honor If-Match on unified datasource compaction writes. When compaction supervisors are disabled, the unified datasource POST and DELETE fallback paths still mutate the same DruidCompactionConfig document through overloads that do not receive DynamicConfigEtagHelper.getIfMatch(...). A client can read the compaction ETag, race with another config write, then POST or DELETE /druid/indexer/v1/compaction/config/datasources/... with a stale If-Match header and still get a successful write instead of 412. Thread the If-Match value through both fallback calls, or explicitly reject conditional requests there. This appears unchanged from the prior review.
> 
> This is an automated review by Codex GPT-5.5

@kfaraz @cecemei @FrankChen021 how do you folks want to handle these internal updates? ETag with retries seems like the best option, although I'm not sure allowing possibility for mutations to be completely blocked is acceptable.

processing/src/main/java/org/apache/druid/common/config/JacksonConfigManager.java
jtuglu1: Done

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

- indexing-service/src/main/java/org/apache/druid/indexing/overlord/http/OverlordCompactionResource.java:289 - [P1] Honor If-Match on unified datasource compaction writes. The previous If-Match gap is still present for the unified datasource compaction POST/DELETE fallback paths. When compaction supervisors are disabled, these branches still call the overloads that do not receive DynamicConfigEtagHelper.getIfMatch(...), so a client can read the compaction ETag, race with another config write, then POST or DELETE /druid/indexer/v1/compaction/config/datasources/... with a stale If-Match header and still get a successful write instead of 412. Thread the If-Match value through both fallback calls, or explicitly reject conditional requests there.

kfaraz:
Removing from Druid 38 milestone since this PR is not fully reviewed yet.

Issue #19456, opened 2026-05-12
[Proposal] Native, vectorised, zero-copy execution path for Druid


Shekharrajak:
### Motivation

Native, vectorised, zero-copy execution path for Druid via FFM and Arrow — staged from segment reader → filter/project → DataFusion (follow-up to #19039 item 1)

Druid today does the entire query hot path on the JVM. Each row passes through a ColumnSelector interface, materialises into Java objects, gets compared/aggregated by JIT-compiled byte code - means : 

* No SIMD on filter / project / aggregate hot loops.
*  GC pressure from per-block wrapper objects on cold-cache scans.
* 2 GiB ceiling on off-heap buffers (see #19025).
* Bottleneck on MSQ joins / aggregates

### Proposed changes

Phase 0 — Spike: native segment column reader via FFM
Phase 1 — Layer0  production-shape: native column reader for all common types
Phase 2 — L1: vectorised native filter + project over Arrow batches
Phase 3 — L2: DataFusion as candidate aggregate/join executor in MSQ stages

### Operational impact

Performance 2–3× speedup on 

* speedup on cold-cache scan + filter for typical Druid datasources.
* speedup on MSQ fact-fact joins.
* speedup on external Parquet ingestion.
* lower p99 for high-cardinality GroupBy.
* faster execution under memory pressure (DataFusion graceful spill).

Operational

* 2 GiB ByteBuffer ceiling lifted on the native path (aligns with #19025).
*   Reduced GC pressure → fewer p99 latency spikes.

jtuglu1:
Thanks for raising this. I was waiting until Druid v38 was cut (so compiler can be switched to JDK 25). This way, we can use latest version of FFM.

jtuglu1:
I think using Datafusion would be a good idea (much better than the current query planner/execution engine set up), but one thing to call out here is that Datafusion always materializes rows into ArrowBatch between operators, which is a bit different in some cases compared to Druid. This materialization cost can be expensive.

jtuglu1:
To be clear, I would *much* rather rely on Datafusion to implement the core physical operators (since they do it much better, and we get future performance improvements for free), the only caveat being we will need to ensure we can get the same or better read amplification than the current native JDK Druid code does, e.g. where we can rely on column stats, row count, bitmap/dictionary indices to pushdown filters/projections and avoid decompressing/scanning more than we need to.

Shekharrajak:
Thanks @jtuglu1 for checking. 

[...]

PR #19510, opened 2026-05-23
feat: Adds an opt-in vectorized read path to the Iceberg input source using iceberg-arrow.

### LINKED ISSUE #19498: Iceberg source: adopt Iceberg's native Arrow reader stack for forward-compatibility with Iceberg spec evolution and performance improvements
### Description

Replace (initial both readers will be available) Druid's hand-written Iceberg reader path with an opt-in path that delegates reading, delete application, and type handling to Iceberg's official iceberg-arrow library. This stops shipping a custom Druid-side fork of Iceberg's reader semantics and lets Druid automatically inherit every Iceberg spec evolution (V2 deletes → V3 deletion vectors / row lineage → V4 and beyond), reader optimisation (pushdown, statistics, vectorisation), and format coverage (Parquet/ORC/Avro and future formats) the moment we bump the Iceberg dependency.

#### Current

* IcebergNativeRecordReader is a Druid-maintained reader 
* Every Iceberg spec improvement (new delete encodings, partition statistics, manifest changes, deletion vectors in V3, row lineage, etc) requires bespoke Druid implementation work .

#### After changes

* New IcebergArrowReader activated by useArrowReader: true in the input spec; defaults to false initially.
* Druid converts the resulting Arrow VectorSchemaRoot batches into MapBasedInputRow via one small adapter; InputRow remains the firewall and nothing else in Druid sees Arrow.
* Iceberg dependency bumps automatically deliver new spec features and optimisations to Druid users with no Druid code change.

### Motivation

This will be first step (foundation step) towards arrow integration https://github.com/apache/druid/issues/19456 and seeing druid + arrow working.

Shekharrajak:
Fixes https://github.com/apache/druid/issues/19498

### Description

Adds an opt-in vectorized read path to the Iceberg input source, backed by iceberg-arrow.

IcebergInputSource ("type": "iceberg") is now a facade over two internal reader-mode delegates, selected by the useArrowReader property:

- StandardDelegate — the existing behavior: resolve the snapshot to a list of live data file paths and read them through warehouseSource with the configured inputFormat.
- ArrowDelegate — builds a filtered/projected TableScan and reads it with Iceberg's ArrowReader, converting ColumnarBatch vectors into Druid InputRows and tracking processed bytes from the Arrow buffers.

Default values: 

useArrowReader : false
arrowBatchSize: 1024

#### Release note

Apache Iceberg ingestion supports an opt-in vectorized reader backed by iceberg-arrow. Enable it with "useArrowReader": true; tune batching with  arrowBatchSize (default 1024). The Arrow path supports Parquet data files, dictionary-encoded values, schema evolution, time-travel snapshots, column  projection, and predicate pushdown. Snapshots containing positional or equality delete files are currently unsupported and are rejected with a clear  error; use the standard delete-aware reader for those snapshots. The Arrow path is not splittable, so keep it disabled where parallel batch throughput is  required. The default reader remains unchanged, and both paths coexist under the same Iceberg input-source type.

This PR has:

- [x] been self-reviewed.


Shekharrajak:
Benchmark : 

```
| numRows  | numCols | icebergArrowInputSourceReader    | icebergInputSourceReader         | Speedup |
|         |         |  ms/op         throughput        |  ms/op         throughput        | (vs baseline) |
+---------+---------+----------------------------------+----------------------------------+---------+
| 100,000 |   5     |   17.70   5,651,074 rows/s       |   51.35   1,947,542 rows/s       |  2.90x  |
| 100,000 |  15     |   44.11   2,266,881 rows/s       |   95.81   1,043,680 rows/s       |  2.17x  |
| 500,000 |   5     |   74.92   6,673,450 rows/s       |  187.73   2,663,420 rows/s       |  2.51x  |
| 500,000 |  15     |  197.97   2,525,637 rows/s       |  420.94   1,187,817 rows/s       |  2.13x  |
================================================================================
Summary:
  Best speedup:    2.90x  (numRows=100,000, numCols=5)
  Worst speedup:   2.13x  (numRows=500,000, numCols=15)
  Geomean speedup: 2.41x
```

Shekharrajak:
Looking into benchmark module which will be used throughout the arrow and datafusion integration to quickly check the improvements.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -76,6 +76,12 @@ public class IcebergInputSource implements SplittableInputSource<List<String>>
   @JsonProperty
   private final ResidualFilterMode residualFilterMode;
 
+  @JsonProperty
+  private final boolean useArrowReader;
+
+  @JsonProperty
+  private final int arrowBatchSize;
```
Shekharrajak, line 72: With a batch of 1024 values in a contiguous buffer ->  emit SIMD instructions that process 8-16 rows per CPU cycle.

Shekharrajak: Decompression amortization -> batch-read helps in decompress once and consume many row, since parquet compressed pages .

Shekharrajak:
ingestion time at 100k × 5

#### Read

Arrow implementation : 17ms 
Current implementaiton : 53ms

#### Index + Persist (= total − read)
~330ms
Similar time on both implementation 

#### total time 

Arrow implementation: 347 ms
current implementation: 410 ms

Index+persist does substantially more work per row than read. So even though **Arrow makes read ~3x faster**, that gain is dwarfed when amortized over the much-larger indexing cost.

Benchmark added https://github.com/Shekharrajak/druid/pull/1/changes

Shekharrajak:
looks like network error https://github.com/apache/druid/actions/runs/26333466916/job/77523245546?pr=19510 - please trigger the CI check again.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -113,6 +125,16 @@ public InputSourceReader reader(
       File temporaryDirectory
   )
   {
+    if (useArrowReader) {
```
FrankChen021: [P1] Arrow path bypasses residual FAIL handling

The useArrowReader branch returns an IcebergArrowInputSourceReader before retrieveIcebergDatafiles() runs, so it skips the residual detection that enforces residualFilterMode=FAIL. IcebergArrowInputSourceReaderTest documents that a non-partition equality filter returns all rows from the file, so a spec with useArrowReader=true and residualFilterMode=FAIL will silently ingest residual rows instead of rejecting the ingestion. Please run the same residual check for the Arrow scan or otherwise honor the fail mode before reading.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,383 @@
...
+  }
+
+  private TableScan buildScan()
+  {
+    TableScan scan = table.newScan().caseSensitive(caseSensitive);
+
+    // Push column projection into the scan planner — only requested columns are read from disk.
+    final List<String> configuredDims = schema.getDimensionsSpec().getDimensionNames();
+    final String timestampColumn = schema.getTimestampSpec().getTimestampColumn();
+    if (!configuredDims.isEmpty()) {
```
FrankChen021: [P1] Projection omits required non-dimension columns

When fixed dimensions are configured, this projection selects only those dimension names plus the timestamp column. Druid carries required transform inputs and aggregator source fields through InputRowSchema.getColumnsFilter(), not DimensionsSpec alone. For example, dimensions=[name] with an aggregator over value will scan only ts/name, so value is missing from the event and downstream metrics/transforms read null. Please drive projection from the full columnsFilter/required fields, or avoid pruning when those required columns cannot be represented safely.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
FrankChen021: [P2] Parallel ingestion bypasses the Arrow reader

This special case only affects direct reader() calls. The input source still reports itself as splittable, and Druid's parallel task runners call createSplits() and then withSplit(); withSplit() returns the delegate input source, so subtasks read raw files through the old delegate path instead of IcebergArrowInputSourceReader. With useArrowReader=true, behavior therefore changes based on maxNumConcurrentSubTasks and can lose the Arrow path's Iceberg semantics. Please make the Arrow mode non-splittable or return split-aware Iceberg input sources that preserve useArrowReader.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 1 |
| P3 | 0 |
| Total | 3 |

---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: Updated.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
Shekharrajak: added test and fix  testResidualFilterModeFailWithArrowReader

Shekharrajak: updated https://github.com/apache/druid/pull/19510/changes/498daa6d476e69f15d79605aef938520a2171dd4 thanks!

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -134,6 +177,9 @@ public Stream<InputSplit<List<String>>> createSplits(
   @Override
   public int estimateNumSplits(InputFormat inputFormat, @Nullable SplitHintSpec splitHintSpec) throws IOException
   {
+    if (useArrowReader) {
+      return 1;
```
Shekharrajak: Future PR: Enable parallel ingestion (maxNumConcurrentSubTasks > 1) when useArrowReader: true by adding split-coordination plumbing to IcebergInputSource. Currently Arrow mode forces a single subtask to guarantee correctness;

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
FrankChen021: Thanks, this now enforces `FAIL` for Arrow before constructing the reader. One remaining parity gap: the enforcement scan does not apply `snapshotTime`, but `IcebergArrowInputSourceReader.buildScan()` later reads `asOfTime(snapshotTime)`. With `useArrowReader=true`, `residualFilterMode=FAIL`, and a historical snapshot, this can fail or pass based on the current snapshot instead of the snapshot being ingested. Can you apply `snapshotTime` to the `TableScan` before `enforceResidualMode`, or share the same scan-building path with the reader?

FrankChen021 (commented):
I reviewed the follow-up changes. The projection issue and Arrow split handling thread look resolved; I left a separate inline reply on the remaining residual-filter snapshot-time parity gap.

---

Shekharrajak:
@FrankChen021 @jtuglu1 - please have a look. CI checks are green.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,402 @@
...
+  public CloseableIterator<InputRow> read(final InputStats inputStats) throws IOException
+  {
+    final TableScan scan = buildScan();
+    final CloseableIterable<CombinedScanTask> tasks = TableScanUtil.planTasks(
+        scan.planFiles(),
+        scan.targetSplitSize(),
+        scan.splitLookback(),
+        scan.splitOpenFileCost()
+    );
+    final ArrowReader arrowReader = new ArrowReader(scan, batchSize, true);
```
Shekharrajak, line 132: Reads Iceberg data files as Arrow columnar batches, then iterates rows lazily as InputRow objects.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,402 @@
...
+ * columns and matching files are read from storage.
+ *
+ * Note: iceberg-arrow currently supports Parquet data files only. ORC and Avro files will throw
+ * {@link UnsupportedOperationException} at read time; use the standard delegate path for those.
+ */
+public class IcebergArrowInputSourceReader implements InputSourceReader
+{
+  // Pin Arrow to Unsafe allocator: Netty backend fails on JDK 25 (EmptyByteBuf.memoryAddress UnsupportedOperationException).
+  static {
+    if (System.getProperty("arrow.allocation.manager.type") == null) {
```
Shekharrajak, line 89: NettyAllocationManager.<clinit> throws UnsupportedOperationException on JDK 25 (incompatible with newer JDK module encapsulation). Pinning to "Unsafe" uses sun.misc.Unsafe which works on all supported JDKs (21, 25).

CI checks were failing .

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,402 @@
...
+ * Reads an Iceberg table via iceberg-arrow's {@link ArrowReader}, yielding {@link InputRow} objects.
+ *
+ * Delete application (V2 equality and positional deletes), type coercion, and schema evolution are
+ * handled entirely by the Iceberg library. Druid only consumes the resulting {@link ColumnarBatch}
+ * batches and maps them to {@link MapBasedInputRow}.
+ *
+ * Column projection and predicate push-down are applied at scan planning time so only requested
+ * columns and matching files are read from storage.
+ *
+ * Note: iceberg-arrow currently supports Parquet data files only. ORC and Avro files will throw
```
Shekharrajak, line 82: Supports only parquet data file

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -76,6 +81,12 @@ public class IcebergInputSource implements SplittableInputSource<List<String>>
   @JsonProperty
   private final ResidualFilterMode residualFilterMode;
 
+  @JsonProperty
+  private final boolean useArrowReader;
```
Shekharrajak, line 71: Adds useArrowReader opt-in flag

FrankChen021 (commented):
I reviewed the follow-up changes for correctness, edge cases, concurrency, and integration risks; no new issues found.

---

Shekharrajak:
Hi @jtuglu1 @clintropolis @FrankChen021  - Please have a look, I have benchmark PR drafted which we will use across all features to see the improvements we are getting through vectorization & arrow implementations.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
cecemei, line 1: The arrow based approach is so different from the file catalog based, it also doesn't really implement `SplittableInputSource` any more, have you considered adding it as a separate `InputSource`? maybe we can extract some shared logic into a separate abstract class.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,402 @@
...
+    this.table = table;
+    this.icebergFilter = icebergFilter;
+    this.snapshotTime = snapshotTime;
+    this.caseSensitive = caseSensitive;
+    this.schema = schema;
+    this.batchSize = batchSize;
+  }
+
+  @Override
+  public CloseableIterator<InputRow> read(final InputStats inputStats) throws IOException
```
cecemei: InputStats is nullable, since the default `read()` just pass in null.

cecemei (commented):
The benchmark results look good! I don't see any similar benchmark coverage in this area, but it's definitely worth adding.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
Shekharrajak: This make sense, Updated:  https://github.com/apache/druid/pull/19510/changes/bf71afb428ec50c1518cd5d68e4743fdc7714468 

AbstractIcebergInputSource — shared metadata fields  .

IcebergArrowInputSource extends AbstractIcebergInputSource implements InputSource — type "iceberg_arrow", non-splittable, needsFormat() = false, owns arrowBatchSize. Calls IcebergArrowInputSourceReader directly.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: Thanks for checking-  Fixed in d04657393b — read() now passes the existing NoopInputStats when inputStats is null

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSource.java
```diff
@@ -0,0 +1,101 @@
...
+      @Nullable InputFormat inputFormat,
+      File temporaryDirectory
+  )
+  {
+    final Table table = retrieveTable();
+    if (icebergFilter != null) {
+      final TableScan filteredScan = icebergFilter.filter(
+          table.newScan().caseSensitive(icebergCatalog.isCaseSensitive())
+      );
+      icebergCatalog.enforceResidualMode(filteredScan, getResidualFilterMode());
```
FrankChen021: [P2] Apply snapshotTime before residual enforcement

reader() checks residuals on a filtered scan built from table.newScan(), but snapshotTime is only applied later inside IcebergArrowInputSourceReader.buildScan(). With residualFilterMode=FAIL, historical ingestion can throw or pass based on the current snapshot rather than the snapshot actually being read. Apply asOfTime(snapshotTime.getMillis()) before enforceResidualMode, or move enforcement into the reader after the snapshot is selected.

extensions-contrib/druid-iceberg-extensions/pom.xml
```diff
@@ -760,6 +760,23 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
```
FrankChen021, line 814: [P2] Register new Arrow dependencies in licenses.yaml

This module adds iceberg-arrow, arrow-vector, and arrow-memory-unsafe, while the root pom also manages arrow-memory-netty, but licenses.yaml has no Arrow or Iceberg Arrow entries. dev/license.md says new library dependencies must be registered there, so binary license/notice generation is incomplete until these artifacts are added.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

---

extensions-contrib/druid-iceberg-extensions/pom.xml
Shekharrajak: Updated https://github.com/apache/druid/pull/19510/changes/7a467f81257505f0698aaf1e70e558aa2c9d12bc

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSource.java
Shekharrajak: Updated, thanks for checking.

Shekharrajak:
Flaky test found : https://github.com/apache/druid/issues/19563

FrankChen021 (commented):

The follow-up updates handle both prior review threads: the Arrow/Iceberg Arrow license metadata is now registered, with `arrow-memory-netty` excluded, and residual enforcement now applies `snapshotTime` before planning. I did not find any new high-confidence PR-caused issues in the updated diff.

---

FrankChen021 (commented):

---

Shekharrajak:
Thanks @FrankChen021 , can we go ahead and make this foundation merged in. I have next phases planned as part of vision https://github.com/apache/druid/issues/19456

kfaraz:
Removing from Druid 38 milestone since this PR has merge conflicts and needs an approval.

Shekharrajak:
Updated the branch with master.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

Reviewed 12 of 12 changed files. The upstream merge does not change the PR-authored Arrow path, but the current diff still corrupts dictionary-backed values and cannot support the delete files it claims to handle.

---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,407 @@
...
+    }
+    return filtered;
+  }
+
+  private InputRow batchRowToInputRow(final ColumnarBatch batch, final int rowIdx)
+  {
+    final int numCols = batch.numCols();
+    final Map<String, Object> event = Maps.newHashMapWithExpectedSize(numCols);
+    for (int col = 0; col < numCols; col++) {
+      final FieldVector vec = batch.column(col).getFieldVector();
```
FrankChen021, line 212: [P1] Decode dictionary-backed vectors before extraction

ColumnVector.getFieldVector() explicitly returns a potentially dictionary-encoded vector. For dictionary-backed strings or partition constants this is an IntVector, so extractValue ingests the dictionary ID instead of the actual value, silently corrupting dimensions. Use getArrowVector() or the typed ColumnVector accessors and assert actual string or partition values in tests; testWithEqualsFilter currently avoids this assertion.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,407 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+/**
+ * Reads an Iceberg table via iceberg-arrow's {@link ArrowReader}, yielding {@link InputRow} objects.
+ *
+ * Delete application (V2 equality and positional deletes), type coercion, and schema evolution are
```
FrankChen021, line 75: [P2] Do not claim unsupported delete-file handling

The class claims equality and positional deletes are applied, but Iceberg 1.10 documents that ArrowReader.open requires no delete files and throws UnsupportedOperationException otherwise. V2 snapshots containing deletes therefore fail ingestion. Detect and fall back to a capable path, or reject and document this limitation explicitly; add coverage for both delete types.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -88,16 +66,27 @@ public IcebergInputSource(
       @JsonProperty("icebergCatalog") IcebergCatalog icebergCatalog,
       @JsonProperty("warehouseSource") InputSourceFactory warehouseSource,
       @JsonProperty("snapshotTime") @Nullable DateTime snapshotTime,
-      @JsonProperty("residualFilterMode") @Nullable ResidualFilterMode residualFilterMode
+      @JsonProperty("residualFilterMode") @Nullable ResidualFilterMode residualFilterMode,
+      // deprecated: use type "iceberg_arrow" instead. retained for spec back-compat.
+      @JsonProperty("useArrowReader") @Nullable Boolean useArrowReader,
+      @JsonProperty("arrowBatchSize") @Nullable Integer arrowBatchSize
```
FrankChen021, line 89: why do we need useArrowReader and arrowBatchSize on the old input source?

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSource.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.iceberg.Table;
+import org.apache.iceberg.TableScan;
+import org.joda.time.DateTime;
+
+import javax.annotation.Nullable;
+import java.io.File;
+
+public class IcebergArrowInputSource extends AbstractIcebergInputSource implements InputSource
+{
+  public static final String TYPE_KEY = "iceberg_arrow";
```
FrankChen021: my only concern is whether we need to introduce a new input source type. from the name itself, input source defines where source files are, how we read files from the input source should be kept internally otherwise we may have lots of input sources combinations.

Can we keep your first revision that use useArrowReader on the IcebergeInputSource and use that property to control whether the arrow vectorized reader is used? I didn't trace your code change, although I know you have a reason to do this.

FrankChen021 (commented):
Suggested design: keep `IcebergInputSource` as the only public input-source type and make it a facade over two internal reader-mode delegates.

The current change leaves `useArrowReader` and `arrowBatchSize` on `IcebergInputSource`, but they are ignored. Instead, those properties could select the implementation explicitly while preserving the existing `iceberg` type:

```java
public class IcebergInputSource implements SplittableInputSource<List<String>>
{
  @JsonProperty
  private final boolean useArrowReader;

  @JsonProperty
  private final int arrowBatchSize;

  @Nullable
  @JsonProperty
  private final InputSourceFactory warehouseSource;

  private final InputSourceDelegate delegate;

  @JsonCreator
  public IcebergInputSource(
      @JsonProperty("tableName") String tableName,
      @JsonProperty("namespace") String namespace,
      @JsonProperty("icebergFilter") @Nullable IcebergFilter icebergFilter,
      @JsonProperty("icebergCatalog") IcebergCatalog icebergCatalog,
      @JsonProperty("warehouseSource") @Nullable InputSourceFactory warehouseSource,
      @JsonProperty("snapshotTime") @Nullable DateTime snapshotTime,
      @JsonProperty("residualFilterMode") @Nullable ResidualFilterMode residualFilterMode,
      @JsonProperty("useArrowReader") @Nullable Boolean useArrowReader,
      @JsonProperty("arrowBatchSize") @Nullable Integer arrowBatchSize
  )
  {
    this.useArrowReader = Boolean.TRUE.equals(useArrowReader);
    this.arrowBatchSize = arrowBatchSize != null && arrowBatchSize > 0
                          ? arrowBatchSize
                          : IcebergArrowInputSourceReader.DEFAULT_BATCH_SIZE;
    this.warehouseSource = warehouseSource;

    this.delegate = this.useArrowReader
                    ? new ArrowDelegate(
                        tableName,
                        namespace,
                        icebergFilter,
                        icebergCatalog,
                        snapshotTime,
                        residualFilterMode,
                        this.arrowBatchSize
                    )
                    : new StandardDelegate(
                        tableName,
                        namespace,
                        icebergFilter,
                        icebergCatalog,
                        Preconditions.checkNotNull(
                            warehouseSource,
                            "warehouseSource is required when Arrow is disabled"
                        ),
                        snapshotTime,
                        residualFilterMode
                    );
  }

  @Override
  public boolean needsFormat()
  {
    return delegate.needsFormat();
  }

  @Override
  public boolean isSplittable()
  {
    return delegate.isSplittable();
  }

  @Override
  public InputSourceReader reader(
      InputRowSchema inputRowSchema,
      @Nullable InputFormat inputFormat,
      File temporaryDirectory
  )
  {
    return delegate.reader(inputRowSchema, inputFormat, temporaryDirectory);
  }

  @Override
  public Stream<InputSplit<List<String>>> createSplits(
      InputFormat inputFormat,
      @Nullable SplitHintSpec splitHintSpec
  ) throws IOException
  {
    return delegate.createSplits(inputFormat, splitHintSpec);
  }

  @Override
  public int estimateNumSplits(
      InputFormat inputFormat,
      @Nullable SplitHintSpec splitHintSpec
  ) throws IOException
  {
    return delegate.estimateNumSplits(inputFormat, splitHintSpec);
  }

  @Override
  public InputSource withSplit(InputSplit<List<String>> split)
  {
    return delegate.withSplit(split);
  }
}
```

The private delegate contract can contain the mode-specific behavior:

```java
private interface InputSourceDelegate
{
  boolean needsFormat();

  boolean isSplittable();

  InputSourceReader reader(
      InputRowSchema inputRowSchema,
      @Nullable InputFormat inputFormat,
      File temporaryDirectory
  );

  Stream<InputSplit<List<String>>> createSplits(
      InputFormat inputFormat,
      @Nullable SplitHintSpec splitHintSpec
  ) throws IOException;

  int estimateNumSplits(
      InputFormat inputFormat,
      @Nullable SplitHintSpec splitHintSpec
  ) throws IOException;

  InputSource withSplit(InputSplit<List<String>> split);
}
```

`StandardDelegate` would contain the current warehouse/delegate implementation. `ArrowDelegate` would return `needsFormat() == false`, `isSplittable() == false`, and create `IcebergArrowInputSourceReader` from the Iceberg `TableScan`. It should validate `FileScanTask.file().format() == FileFormat.PARQUET` using Iceberg metadata, apply the filter and snapshot before validation, and return the owning `IcebergInputSource` from `withSplit()` if that method is reached unexpectedly.

This keeps old specs with no `useArrowReader` unchanged, gives the properties real meaning, avoids a second public input-source type, and makes the different format/splitting contracts explicit in the two delegates.

FrankChen021 (commented):
I reviewed the updated facade/refactor and documentation. No additional high-confidence issues were found in this update; the previously reported Arrow dictionary-vector corruption and delete-file support findings remain unchanged.

Validation: focused `git diff --check` passed; current CI checks pass. Builds and tests were not run locally.



---

FrankChen021 (approved):

FrankChen021:
@Shekharrajak Thanks for the contribution. I will keep this open for several days and merge it if there's no other concerns/comments from other committers

Shekharrajak:
@FrankChen021 gentle reminder .
I am also planning to present this in my Community Over Code presentation on Iceberg/Arrow/Rust .
Thanks!

FrankChen021:
> @FrankChen021 gentle reminder .
> 
> I am also planning to present this in my Community Over Code presentation on Iceberg/Arrow/Rust .
> 
> Thanks! 

Sure. If possible, can you share the link of slides here?

Shekharrajak:
> Sure. If possible, can you share the link of slides here?

The actual topic is "Native Iceberg Scans at Rust Speed: How DataFusion-Comet Achieves Faster Query Performance", I will share once I have draft ready (the previous iceberg summit talk slide is [here](https://shekharrajak.github.io/IcebergSummit2026/)).

I would like to add one slide to showcase how OLAP system adopting arrow reader to boost the reader performance + benchmark.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,407 @@
...
+ * Note: iceberg-arrow currently supports Parquet data files only. ORC and Avro files will throw
+ * {@link UnsupportedOperationException} at read time; use the standard delegate path for those.
+ */
+public class IcebergArrowInputSourceReader implements InputSourceReader
+{
+  // Pin Arrow to Unsafe allocator: Netty backend fails on JDK 25 (EmptyByteBuf.memoryAddress UnsupportedOperationException).
+  static {
+    if (System.getProperty("arrow.allocation.manager.type") == null) {
+      System.setProperty("arrow.allocation.manager.type", "Unsafe");
+    }
```
abhishekrb19: Is there a way to avoid getting and setting system properties? We could make this configurable via Druid runtime properties that operators can override, while defaulting to this `Unsafe` type. Alternatively, we could punt on the property altogether and use the default for now.

extensions-contrib/druid-iceberg-extensions/src/test/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReaderTest.java
```diff
@@ -0,0 +1,377 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+public class IcebergArrowInputSourceReaderTest
+{
+  private static final String NAMESPACE = "default";
```
abhishekrb19, line 63: Is it possible to write an embedded-test to test things more end-to-end? For example, there's an existing iceberg test here: https://github.com/apache/druid/blob/master/embedded-tests/src/test/java/org/apache/druid/testing/embedded/iceberg/IcebergRestCatalogIngestionTest.java

You could use that skeleton to test the arrow-based reader.

abhishekrb19 (commented):
A few comments

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: Without this we were getting `UnsupportedOperationException: EmptyByteBuf.memoryAddress` . Through some analysis : Arrow’s iceberg-arrow code creates its own RootAllocator . Let me check alternative.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/iceberg/IcebergRestCatalogIngestionTest.java
```diff
@@ -193,5 +197,12 @@ private void ingestFromIcebergRestCatalog(final String targetDataSource, final b
         + "2024-01-01T01:00:00.000Z,bob,200\n"
         + "2024-01-01T02:00:00.000Z,charlie,300"
     );
+
+    log.info(
```
Shekharrajak, line 201: Use this logger to analyze manually how much time both readers are taking. With these current updates, the approach looks highly promising 

```
 - Standard reader: 2289 ms
  - Arrow reader: 818 ms
approximately 2.8x faster 
```

extensions-contrib/druid-iceberg-extensions/src/test/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReaderTest.java
Shekharrajak: Thanks for review. Updated https://github.com/apache/druid/pull/19510/changes/62d719892752d78464c008c37b8b7d6ab66db031#r3954210250

embedded-tests/src/test/java/org/apache/druid/testing/embedded/iceberg/IcebergRestCatalogIngestionTest.java
```diff
@@ -151,30 +170,39 @@ public void testIngestFromIcebergRestCatalog()
...
         "2024-01-01T00:00:00.000Z,alice,100\n"
         + "2024-01-01T01:00:00.000Z,bob,200\n"
         + "2024-01-01T02:00:00.000Z,charlie,300"
     );
+
+    log.info(
+        "Iceberg REST catalog ingestion timing: useArrowReader[%s], dataSource[%s], elapsedMs[%d]",
+        useArrowReader,
+        targetDataSource,
+        TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNanos)
```
abhishekrb19: nit: use a `StopWatch`

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: Updated. CI checks all green.

Shekharrajak:
@abhishekrb19 @FrankChen021  - Please have a look. It is looking good to go.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,400 @@
...
+    return filtered;
+  }
+
+  private InputRow batchRowToInputRow(final ColumnarBatch batch, final int rowIdx)
+  {
+    final int numCols = batch.numCols();
+    final Map<String, Object> event = Maps.newHashMapWithExpectedSize(numCols);
+    for (int col = 0; col < numCols; col++) {
+      final FieldVector vec = batch.column(col).getFieldVector();
```
FrankChen021: [P1] Decode dictionary-backed vectors before extraction

ColumnVector.getFieldVector() is explicitly the potentially dictionary-encoded vector. For a Parquet dictionary-encoded string (including repeated values), Iceberg can expose an IntVector of dictionary IDs here; the IntVector branch at lines 242-244 then writes that ID into the Druid event instead of the string. The test even avoids asserting filtered string values for this reason. Use ColumnVector.getArrowVector() or the typed ColumnVector accessors before building the row, and add a repeated/partitioned-string assertion, otherwise useArrowReader=true silently corrupts dimensions.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,400 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+/**
+ * Reads an Iceberg table via iceberg-arrow's {@link ArrowReader}, yielding {@link InputRow} objects.
+ *
+ * Delete application (V2 equality and positional deletes), type coercion, and schema evolution are
```
FrankChen021: [P2] Delete-file snapshots are unsupported

Iceberg 1.11's ArrowReader.open rejects any CombinedScanTask with delete files by throwing UnsupportedOperationException; its API contract requires that there be no delete files. This class currently claims V2 equality and positional deletes are applied, and the Arrow delegate does not detect or route around them, so enabling useArrowReader on a valid Iceberg snapshot with deletes fails ingestion instead of applying deletes or giving a clear documented fallback. Detect delete tasks and use the standard path or reject them explicitly, and update this claim.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

This updated_since_review reassessment found that the previously reported dictionary-vector corruption and delete-file support issues remain actionable at the current head. The incremental test, timing-log, and allocator changes introduced no additional correctness findings.



---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -113,6 +117,22 @@ public CloseableIterator<InputRow> read(@Nullable final InputStats inputStats) t
...
 
+  private void validateNoDeleteFiles(final TableScan scan) throws IOException
+  {
+    try (CloseableIterable<FileScanTask> fileTasks = scan.planFiles()) {
+      for (FileScanTask fileTask : fileTasks) {
+        if (!fileTask.deletes().isEmpty()) {
+          throw DruidException.forPersona(DruidException.Persona.USER)
+                              .ofCategory(DruidException.Category.UNSUPPORTED)
+                              .build(
+                                  "Arrow reader does not support Iceberg snapshots with delete files. "
```
Shekharrajak, line 139: We can create a ticket to track "Add delete-file support for Iceberg Arrow reader"

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: updated

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: Thanks updated and added unit tests

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
FrankChen021: Verified at head `1779b3396d57be84715c4f45a9d73f3fb0ec3864`: this thread is addressed in the implementation. `validateNoDeleteFiles` walks the planned `FileScanTask.deletes()` entries before opening Arrow and returns a user-facing unsupported error, while the reader Javadoc now states that delete-file snapshots are rejected; the new regression test covers both positional and equality delete metadata. Please also update the PR release-note text, which still says the Arrow path applies V2 deletes.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
FrankChen021: Verified at head `1779b3396d57be84715c4f45a9d73f3fb0ec3864`: this thread is addressed. Row mapping now uses Iceberg `ColumnVector` null checks and typed accessors, including `getString`, so dictionary-backed values are decoded before they enter the Druid event; the repeated `alice`/`bob`/`alice` assertions cover the reported case.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,371 @@
...
+
+  private InputRow batchRowToInputRow(final ColumnarBatch batch, final int rowIdx)
+  {
+    final int numCols = batch.numCols();
+    final Map<String, Object> event = Maps.newHashMapWithExpectedSize(numCols);
+    for (int col = 0; col < numCols; col++) {
+      final ColumnVector column = batch.column(col);
+      final FieldVector vec = column.getFieldVector();
+      if (!column.isNullAt(rowIdx)) {
+        event.put(vec.getName(), extractValue(column, table.schema().findField(vec.getName()).type(), rowIdx));
```
FrankChen021: [P1] Use the selected snapshot schema for value extraction

`buildScan()` applies `snapshotTime`, and Iceberg's `DataTableScan` then projects the selected snapshot's schema, but this lookup uses `table.schema()` (the current schema). If a later schema update drops or renames a field that is present in the time-travel snapshot, `findField(vec.getName())` returns null and ingestion fails at `.type()`; if the field was type-promoted, the current type can select the wrong accessor. Retain the scan/read schema for this mapping and add a schema-evolution time-travel regression test (or reject/document the unsupported case).

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

The two supplied follow-ups were rechecked at head `1779b3396d57be84715c4f45a9d73f3fb0ec3864`: delete-file snapshots are now rejected explicitly, and dictionary-backed values are decoded through Iceberg `ColumnVector` accessors. I found one separate P1 in time-travel/schema-evolution mapping; see the inline finding.

---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: Thanks - updated : https://github.com/apache/druid/pull/19510/changes/ed415391a20865beeccc1f8a3ac1b4fda5d1181d

Shekharrajak: Thanks , update

FrankChen021: The original snapshot-schema issue is resolved: this head now passes the selected `scan.schema()` into value extraction, so the reported time-travel rename/drop case is addressed. I found four separate current-head issues in the Arrow path; the new COMMENT review records them.

---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -125,97 +195,248 @@ public Stream<InputSplit<List<String>>> createSplits(
...
+      final Table table = retrieveTable();
+      TableScan scan = table.newScan().caseSensitive(icebergCatalog.isCaseSensitive());
+      if (icebergFilter != null) {
+        scan = icebergFilter.filter(scan);
+      }
+      if (snapshotTime != null) {
+        scan = scan.asOfTime(snapshotTime.getMillis());
+      }
+      if (icebergFilter != null) {
+        icebergCatalog.enforceResidualMode(scan, residualFilterMode);
```
FrankChen021, line 371: [P1] Preserve IGNORE residual semantics

This only validates the residual mode; the reader below still builds the scan with the original filter. Iceberg 1.11's ArrowReader passes each FileScanTask.residual() to Parquet's record filter, so an unpartitioned equals filter removes non-matching rows. That contradicts this source's documented/default IGNORE behavior, where residual rows are ingested and may be filtered by Druid's transformSpec, and it also conflicts with the new test expecting all three rows. Disable the record filter for IGNORE (while retaining it for an explicitly supported mode), or change the mode's contract and tests.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,381 @@
...
+    final TableScan scan = buildScan();
+    validateNoDeleteFiles(scan);
+    final CloseableIterable<CombinedScanTask> tasks = TableScanUtil.planTasks(
+        scan.planFiles(),
+        scan.targetSplitSize(),
+        scan.splitLookback(),
+        scan.splitOpenFileCost()
+    );
+    final ArrowReader arrowReader = new ArrowReader(scan, batchSize, true);
+    final org.apache.iceberg.io.CloseableIterator<ColumnarBatch> batchIter = arrowReader.open(tasks);
```
FrankChen021: [P1] Keep the extension classloader for Arrow iteration

retrieveTable() restores the thread context classloader before this Arrow reader is opened. In Iceberg 1.11, ArrowReader lazily asks FormatModelRegistry to reflectively load org.apache.iceberg.arrow.vectorized.ArrowFormatModels through the thread context loader as the iterator advances. In a normal installed Druid extension that loader does not contain iceberg-arrow, so the first batch can fail to find/register the Parquet ColumnarBatch model even though classpath-based tests pass. Keep getClass().getClassLoader() active across Arrow reader creation and the returned iterator lifecycle, or explicitly register the model with the extension loader.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,381 @@
...
+      final ColumnarBatch batch,
+      final int rowIdx,
+      final Schema readSchema
+  )
+  {
+    final int numCols = batch.numCols();
+    final Map<String, Object> event = Maps.newHashMapWithExpectedSize(numCols);
+    for (int col = 0; col < numCols; col++) {
+      final ColumnVector column = batch.column(col);
+      final FieldVector vec = column.getFieldVector();
```
FrankChen021: [P1] Handle fields absent from older files

When an optional Iceberg field is added after an existing Parquet file, the per-file Arrow reader uses a dummy/null holder for that field. The holder has no FieldVector, but this line dereferences it before checking column.isNullAt; resolveDimensions() and estimateBatchBytes() make the same assumption. Reading a mixed old/new table therefore fails instead of producing null for the old rows (and Iceberg 1.11 may fail while constructing the batch for the dummy holder). Derive names from readSchema and handle dummy/constant columns, or reject/fallback for this schema evolution.

FrankChen021 (commented):
## 🟡 Changes recommended

The prior time-travel/schema-evolution finding is resolved at head `71d1cfbe8ee3ca913254930472f25c665801a181`: value extraction now uses the selected `scan.schema()`. The current-head review found four separate Arrow-path correctness issues covering residual-filter semantics, extension classloading, absent evolved columns, and legal primitive type promotion.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 4 |
| P2 | 0 |
| P3 | 0 |
| Total | 4 |



---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,381 @@
...
+      if (!column.isNullAt(rowIdx)) {
+        event.put(vec.getName(), extractValue(column, readSchema.findField(vec.getName()).type(), rowIdx));
```
FrankChen021: [P1] Guard against type promotion failures

readSchema is the logical table schema, but iceberg-arrow 1.11 allocates vectors from each Parquet file's physical/logical type and does not handle every legal promotion. For an existing int field promoted to long, an old Parquet INT32/INT(32) column can make ArrowReader cast a BigIntVector to IntVector while allocating the batch, before this accessor can read it; float-to-double has the same shape. A normal Iceberg type evolution therefore makes Arrow ingestion fail on the first old file. Detect promoted fields and use the standard reader/fail clearly, and do not claim schema evolution is fully delegated to Iceberg without covering this case.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,432 @@
...
+    if (snapshotTime != null) {
+      scan = scan.asOfTime(snapshotTime.getMillis());
+    }
+
+    final List<String> projection = projectedColumns(scan.schema());
+    if (projection != null) {
+      scan = scan.select(projection);
+    }
+    if (icebergFilter != null) {
+      scan = icebergFilter.filter(scan);
```
FrankChen021, line 187: [P1] Preserve IGNORE residual semantics

`ArrowReader.open` hard-codes `.filter(task.residual())` for every `FileScanTask`, and this scan never calls `ignoreResiduals()`. Consequently the default `residualFilterMode=IGNORE` still drops nonmatching rows before Druid sees them. The new `testWithEqualsFilter` expects all three rows for an unpartitioned `name = alice` filter, but the Arrow iterator returns only the two matches, so the behavior contradicts the documented standard mode and prevents a Druid `transformSpec` from handling residual rows. Pass an always-true residual/`ignoreResiduals()` for IGNORE while retaining FAIL validation, or change the mode contract and tests.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,432 @@
...
+        return TimeUnit.DAYS.toMillis(column.getInt(idx));
+      case TIME:
+        return TimeUnit.MICROSECONDS.toMillis(column.getLong(idx));
+      case TIMESTAMP:
+        return TimeUnit.MICROSECONDS.toMillis(column.getLong(idx));
+      case TIMESTAMP_NANO:
+        return TimeUnit.NANOSECONDS.toMillis(column.getLong(idx));
+      case DECIMAL:
+        final Types.DecimalType decimalType = (Types.DecimalType) type;
+        return column.getDecimal(idx, decimalType.precision(), decimalType.scale());
```
FrankChen021, line 282: [P1] Do not truncate high-precision decimals

`ColumnVector.getDecimal` in iceberg-arrow 1.11's Java decimal accessor converts the decoded value with `value.unscaledValue().longValue()` before constructing the returned `BigDecimal`. A valid DECIMAL(20,2) or DECIMAL(38,...) value whose unscaled value exceeds `long` is therefore silently truncated/wrapped before it reaches the Druid event. This path advertises scalar support but neither limits decimal precision nor tests it; the result is incorrect data rather than a clean unsupported-input failure. Preserve arbitrary-precision values or reject unsupported decimal schemas/values before selecting Arrow.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -34,76 +34,149 @@
...
       @JsonProperty("tableName") String tableName,
       @JsonProperty("namespace") String namespace,
       @JsonProperty("icebergFilter") @Nullable IcebergFilter icebergFilter,
       @JsonProperty("icebergCatalog") IcebergCatalog icebergCatalog,
-      @JsonProperty("warehouseSource") InputSourceFactory warehouseSource,
+      @JsonProperty("warehouseSource") @Nullable InputSourceFactory warehouseSource,
       @JsonProperty("snapshotTime") @Nullable DateTime snapshotTime,
-      @JsonProperty("residualFilterMode") @Nullable ResidualFilterMode residualFilterMode
+      @JsonProperty("residualFilterMode") @Nullable ResidualFilterMode residualFilterMode,
+      @JsonProperty("useArrowReader") @Nullable Boolean useArrowReader,
```
FrankChen021, line 88: [P2] Preserve the existing Java construction API

This replaces the only public seven-argument `IcebergInputSource` constructor with a nine-argument constructor, so existing Java callers and already-compiled clients that construct the source no longer compile or link even when Arrow is disabled. The refactor also removes the public `getDelegateInputSource()` and protected `retrieveIcebergDatafiles()` hooks. Keep a delegating seven-argument overload and preserve or deprecate the old hooks, or explicitly stage this as a breaking API change.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,432 @@
...
+        scan.targetSplitSize(),
+        scan.splitLookback(),
+        scan.splitOpenFileCost()
+    );
+    final ClassLoader extensionClassLoader = IcebergArrowInputSourceReader.class.getClassLoader();
+    final ClassLoader originalClassLoader = Thread.currentThread().getContextClassLoader();
+    try {
+      Thread.currentThread().setContextClassLoader(extensionClassLoader);
+      final ArrowReader arrowReader = new ArrowReader(scan, batchSize, true);
+      final org.apache.iceberg.io.CloseableIterator<ColumnarBatch> batchIter = arrowReader.open(tasks);
```
FrankChen021, line 116: [P2] Close scan tasks when reader setup fails

`TableScanUtil.planTasks` returns a closeable iterable that owns the underlying planned-file iterable, but it is only closed by `ArrowInputRowIterator.close()`. If `arrowReader.open(tasks)` throws while materializing an unsupported type, empty projection, bad delete state, or another setup error, no iterator is returned and neither `tasks` nor the partially-created reader is closed. Repeated failed Arrow ingestions can consequently retain scan resources; close the task iterable on every setup-failure path while transferring ownership only after `open` succeeds.

FrankChen021 (commented):
## 🟡 Changes recommended

Static updated-head assessment. I reviewed the incremental diff first, then the full merge-base diff and surrounding implementation/test code for the current head. The previous classloader, missing-evolved-field, snapshot-time ordering, and basic integer/float promotion concerns were rechecked; the first three are addressed, and the basic promotion concern is not carried forward because iceberg-arrow exposes widening typed accessors for those legal promotions.

Reviewed 11 of 11 changed files, including the documentation, embedded ingestion coverage, dependency/license changes, catalog/source refactor, Arrow reader, and all updated tests.

Validation: `git diff --check 61ed0a389d6af0a3e410b651f379d4f554bbcacb..f82f06fbf826f3e81f84dd1b2b4cfed2b7a79b7e` passed. Per request, no builds, tests, dependency installs, fetches, preparation, cleanup, or formatters were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 2 |
| P3 | 0 |
| Total | 4 |



---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
FrankChen021: AI is too strict on API changes, and I don't think this is true

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -158,6 +159,21 @@ private void validateNoDeleteFiles(final TableScan scan) throws IOException
...
+  private void validateDecimalPrecision(final TableScan scan)
+  {
+    for (final Types.NestedField field : scan.schema().columns()) {
+      if (field.type().typeId() == Type.TypeID.DECIMAL
+          && ((Types.DecimalType) field.type()).precision() > 18) {
+        throw DruidException.forPersona(DruidException.Persona.USER)
+                            .ofCategory(DruidException.Category.UNSUPPORTED)
+                            .build(
+                                "Arrow reader does not support decimal fields with precision greater than 18. "
+                                + "Use the standard Iceberg reader."
```
Shekharrajak, line 186: We will see this usecase and try to support these scenarios in future.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: Seems like it is considering very trivial cases as well but updated the PR.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: Updated.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
Shekharrajak: Updated https://github.com/apache/druid/pull/19510/changes/d78368da80c2fb411b1754ec9d460d4054dd1e94

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
```diff
@@ -0,0 +1,483 @@
...
+      }
+    }
+    final long timestamp = schema.getTimestampSpec().extractTimestamp(event).getMillis();
+    final List<String> dimensions = resolveDimensions(readSchema);
+    return new MapBasedInputRow(timestamp, dimensions, event);
+  }
+
+  private List<String> resolveDimensions(final Schema readSchema)
+  {
+    final List<String> configured = schema.getDimensionsSpec().getDimensionNames();
```
FrankChen021: [P1] Honor dynamic DimensionsSpec settings

When DimensionsSpec has explicit dimensions together with includeAllDimensions or useSchemaDiscovery, Druid's normal MapInputRowParser keeps those names and appends discovered fields; when no explicit dimensions are configured, it also removes dimensionExclusions. This method returns only configured names whenever the list is non-empty and otherwise adds every scanned field except the timestamp, so Arrow ingestion silently drops discovered dimensions and can re-add an excluded field when an Iceberg predicate causes that field to appear in scan.schema(). Build this list with the same DimensionsSpec rules as the standard reader.

FrankChen021 (commented):
## 🟡 Changes recommended

Static follow-up assessment. The three prior findings were rechecked at the current head: reader setup now closes task resources on failure, high-precision decimal schemas are rejected before Arrow accessors can truncate them, and IGNORE residual mode now calls `ignoreResiduals()` so residual rows are intentionally returned.

One remaining correctness issue is reported inline: Arrow dimension selection does not honor Druid's dynamic-dimension and exclusion semantics.

Reviewed 11 of 11 changed files, including the documentation, embedded ingestion coverage, dependency and license changes, catalog/source refactor, Arrow reader, and all updated tests.

Validation: `git diff --check 61ed0a389d6af0a3e410b651f379d4f554bbcacb..d78368da80c2fb411b1754ec9d460d4054dd1e94` passed. No build or test commands were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |



---

This is an automated review by Codex GPT-5.6 Luna(Max)

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
FrankChen021: @Shekharrajak  For this one, I think you can decide whether to fix it OR document it as this is not a core extension and exists in the opt-in path

Shekharrajak: Added tests

Shekharrajak:
```
Error: Can't get Docker image: RemoteDockerImage(imageName=minio/minio:latest, imagePullPolicy=DefaultPullPolicy(), imageNameSubstitutor=org.testcontainers.utility.ImageNameSubstitutor$LogWrappedImageNameSubstitutor@4b1ef6a)
  ⏩ Skipped creation of job summary

```
CI checks looking related to docker image pull.

FrankChen021:
> ```
> Error: Can't get Docker image: RemoteDockerImage(imageName=minio/minio:latest, imagePullPolicy=DefaultPullPolicy(), imageNameSubstitutor=org.testcontainers.utility.ImageNameSubstitutor$LogWrappedImageNameSubstitutor@4b1ef6a)
>   ⏩ Skipped creation of job summary
> ```
> 
> CI checks looking related to docker image pull.

Fixed in the latest master branch. you can merge back the master to your branch

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergArrowInputSourceReader.java
FrankChen021: ## Follow-up assessment

The prior P1 is resolved at current head `b82709bcbc97cc8fa9e5f7d2fa92f57bff98b96a`: `resolveDimensions` now delegates to Druid's canonical `MapInputRowParser.findDimensions`, so explicit dimensions, include-all/schema-discovery fields, exclusions, and the timestamp column follow the standard reader semantics. The added tests cover explicit dynamic dimensions and exclusions. I rechecked the full current-head change and found no additional PR-caused correctness, edge-case, concurrency/lifecycle, security, data-loss, API-compatibility, or missing-test issue.

Validation: `git diff --check 61ed0a389d6af0a3e410b651f379d4f554bbcacb..b82709bcbc97cc8fa9e5f7d2fa92f57bff98b96a` passed. No builds or tests were run.



---

FrankChen021 (commented):
## 🟢 Approval recommended

Static updated-head assessment. I started with the incremental diff from the previous reviewed head `d78368da80c2fb411b1754ec9d460d4054dd1e94`, then inspected the complete current diff and surrounding implementation and test code. The prior dynamic-dimension finding is resolved: Arrow dimension selection now delegates to Druid's canonical `MapInputRowParser.findDimensions` behavior. The previously reported residual-filter, extension-classloader, snapshot/schema-evolution, delete-file, decimal-precision, setup-cleanup, and licensing issues were rechecked at the current head and are not duplicated here.

No additional PR-caused correctness, edge-case, concurrency/lifecycle, security, data-loss, API-compatibility, or missing-test issue remains.

Reviewed 11 of 11 changed files, including the documentation, embedded ingestion coverage, dependency and license changes, catalog/source refactor, Arrow reader, and all updated tests.

Validation: `git diff --check 61ed0a389d6af0a3e410b651f379d4f554bbcacb..b82709bcbc97cc8fa9e5f7d2fa92f57bff98b96a` passed. No builds or tests were run.



---

Shekharrajak:
CI check error https://github.com/apache/druid/issues/20333

FrankChen021 (commented):
## 🟢 Approval recommended

Static updated-head assessment. I started with the incremental diff from previous reviewed head `b82709bcbc97cc8fa9e5f7d2fa92f57bff98b96a`; it is the merge commit’s upstream-master delta, so I then inspected the complete current merge-base diff and surrounding implementation and tests. The PR-authored Arrow path is unchanged.

The prior dictionary decoding, delete-file rejection, snapshot/schema-evolution, decimal precision, setup cleanup, residual-filter mode, extension-classloader, missing-field, basic type-promotion, and dynamic-dimension concerns were rechecked at the current head and are addressed or intentionally rejected with documented/tested behavior. I found no additional PR-caused correctness, edge-case, concurrency/lifecycle, security, data-loss, API-compatibility, integration, or missing-test issue.

Reviewed 11 of 11 current changed files: the documentation, embedded REST ingestion test, dependency/license metadata, catalog/source implementation, Arrow reader, and all updated tests.

Validation: `git diff --check bba74f6cb28eb091c41835b6e26d270b535a0c73..bac4e649c00ec291a3cd3c85645475c544bdb652` passed. No builds or tests were run.



---

Shekharrajak:
@FrankChen021 @abhishekrb19 - Please have a look. This is looking complete.

FrankChen021:
> @FrankChen021 @abhishekrb19 - Please have a look. This is looking complete.

No worry. It's ready to merge from my side.
I'm keeping this open for several days to see if there're new comments.

docs/development/extensions-contrib/iceberg.md
```diff
@@ -194,6 +194,34 @@ Example:
...
 When `residualFilterMode` is set to `fail` and a residual filter is detected, the job will fail with an error message indicating which filter expression produced the residual. This helps ensure data quality by preventing unintended rows from being ingested.
 
+## Arrow vectorized reader
+
+By default the Iceberg input source resolves the snapshot to a list of data file paths and reads them through the `warehouseSource`. Setting `useArrowReader` to `true` reads the table scan directly with Iceberg's vectorized Arrow reader instead, which avoids the per-file input format layer.
+
+| Property | Description | Default |
+|----------|-------------|---------|
+| `useArrowReader` | Read the table with Iceberg's vectorized Arrow reader. | `false` |
+| `arrowBatchSize` | Rows per Arrow batch. Only applies when `useArrowReader` is `true`. | `1024` |
```
abhishekrb19, line 204: Do we anticipate adding more arrow-specific configurable properties? If so, it'd be cleaner to have an `arrowConfig` container or so that'll encapsulate and group all the arrow reader props together.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/iceberg/IcebergRestCatalogIngestionTest.java
abhishekrb19, line 205: nit: use a `StopWatch`

embedded-tests/src/test/java/org/apache/druid/testing/embedded/iceberg/IcebergRestCatalogIngestionTest.java
```diff
@@ -151,30 +170,39 @@ public void testIngestFromIcebergRestCatalog()
...
         "SELECT __time, \"name\", \"value\" FROM %s ORDER BY __time",
-        dataSource,
+        targetDataSource,
         "2024-01-01T00:00:00.000Z,alice,100\n"
         + "2024-01-01T01:00:00.000Z,bob,200\n"
         + "2024-01-01T02:00:00.000Z,charlie,300"
     );
+
+    log.info(
+        "Iceberg REST catalog ingestion timing: useArrowReader[%s], dataSource[%s], elapsedMs[%d]",
```
abhishekrb19, line 202: Does the main code (outside of tests) actually log something like this? It seems generally useful to have a log and/or metric for this.

embedded-tests/src/test/java/org/apache/druid/testing/embedded/iceberg/IcebergRestCatalogIngestionTest.java
```diff
@@ -136,7 +139,23 @@ public void tearDownIceberg()
   @Test
   public void testIngestFromIcebergRestCatalog()
   {
+    ingestFromIcebergRestCatalog(dataSource, false);
+  }
+
+  @Test
+  public void testIngestFromIcebergRestCatalogWithArrowReader()
+  {
+    ingestFromIcebergRestCatalog(dataSource + "_arrow", true);
```
abhishekrb19, line 148: Thanks for adding these tests!

abhishekrb19 (approved):
Looks good to me overall. A few non-blocking comments.

Shekharrajak:
Thanks @abhishekrb19 for checking.  I will address the non blocking comments in next phase.

PR #19514, opened 2026-05-25
feat: Add Google RE2/J linear time regular expression as alternative to Java regex

### LINKED ISSUE #19513: CWE-1333: Inefficient Regular Expression Complexity
### Description

The product uses a regular expression with a worst-case computational complexity that is inefficient and possibly exponential.

## Summary
A DATASOURCE WRITE user can hang Overlord worker threads indefinitely with a single sampler request, degrading or denying service on the cluster control plane.

`processing/src/main/java/org/apache/druid/data/input/impl/RegexInputFormat.java:50-60`:

```java
public RegexInputFormat(
    @JsonProperty("pattern") String pattern,
    @JsonProperty("listDelimiter") @Nullable String listDelimiter,
    @JsonProperty("columns") @Nullable List<String> columns
)
{
  this.pattern = pattern;
  this.listDelimiter = listDelimiter;
  this.columns = columns;
  this.compiledPatternSupplier = Suppliers.memoize(() -> Pattern.compile(pattern));
}
```

RegexInputFormat compiles @JsonProperty pattern with no complexity/length limit and applies Matcher.matches() per line. The sampler runs in the Overlord JVM (CliOverlord.java:460); TimedShutoffInputSourceReader only checks the volatile closed flag at iterator boundaries (:89-101) and cannot interrupt an in-progress Matcher.matches(). Attacker also controls timeoutMs and can set it to 0.

**Exploit scenario (static hypothesis — unverified):**
DATASOURCE WRITE user POSTs sampler spec with InlineInputSource.data='aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaX' and RegexInputFormat.pattern='^(.*a){20}$', timeoutMs=0. Overlord thread enters catastrophic backtracking and never returns to the iterator boundary. A few concurrent requests exhaust the Jetty pool.

---

vivek807:
Fixes #19513.

### Description

Add Google RE2/J linear time regular expression as alternative to Java regex

```properties
druid.regex.engine=JAVA
```

Supported values:

| Value    | Description                                                              |
|----------|--------------------------------------------------------------------------|
| `JAVA`   | Uses Java's built-in `java.util.regex.Pattern` engine.                   |
| `RE2J`   | Uses `Google's RE2/J` regex engine with linear-time matching guarantees. |

Default value:

```properties
druid.regex.engine=JAVA
```

### RE2/J engine

Setting:

```properties
druid.regex.engine=RE2J
```

enables the RE2/J regex engine for ingestion task `regex` input formats.

RE2/J helps protect against catastrophic backtracking and Regular Expression Denial of Service (ReDoS) attacks by guaranteeing linear-time regex evaluation.

### Compatibility differences

RE2/J does not support all Java regex features.

Unsupported or partially supported features include:
- backreferences
- lookbehind assertions
- some advanced backtracking behavior

Patterns using unsupported constructs will fail during regex compilation.

### Example of catastrophic backtracking

The following Java regex may cause catastrophic backtracking:

```regex
^(.*a){20}$
```

against input such as:

```text
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaX
```

Using `RE2J` avoids this issue.

### Performance considerations

- `JAVA` may support more advanced regex syntax and behavior.
- `RE2J` provides safer and more predictable runtime characteristics.
- For trusted internal ingestion specs, `JAVA` may be preferred for compatibility.
- For externally supplied regex patterns, `RE2J` is recommended.

This PR has:

- [X] been self-reviewed.


processing/src/main/java/org/apache/druid/data/input/impl/RegexInputFormat.java
```diff
@@ -56,7 +60,7 @@ public RegexInputFormat(
     this.pattern = pattern;
     this.listDelimiter = listDelimiter;
     this.columns = columns;
-    this.compiledPatternSupplier = Suppliers.memoize(() -> Pattern.compile(pattern));
+    this.compiledPatternSupplier = Suppliers.memoize(() -> RegexPatternFactory.compile(regexConfig.getEngine(), pattern));
```
FrankChen021, line 63: [P2] Regex engine config is not wired into all ingestion runtimes

`RegexInputFormat` now depends on injected `RegexConfig`, but the module that binds the configured engine is not installed in every ingestion runtime that deserializes this type, such as standalone indexer and sampler/overlord paths. In those paths Guice can fall back to the default Java engine, so `druid.regex.engine=RE2J` is silently ignored and the ReDoS mitigation does not cover supported ingestion flows. Please install the regex engine module wherever input formats are materialized, or otherwise ensure this constructor receives the configured engine in those runtimes.

docs/ingestion/data-formats.md
```diff
@@ -897,7 +965,8 @@ This query returns:
 |---------------------|-----------------|
 | `1680795276351`     | `partition-1`   |
 
-## FlattenSpec
+## Flat
```
FrankChen021: [P3] Restore the FlattenSpec heading

This heading was split into `## Flat` and `tenSpec`, so the rendered docs no longer expose the `FlattenSpec` section title or the existing `#flattenspec` anchor. Please restore the single `## FlattenSpec` heading.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 1 |
| Total | 2 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 1 |
| Total | 2 |

---

docs/ingestion/data-formats.md
vivek807: Done

FrankChen021 (commented):
I reviewed the follow-up. The `FlattenSpec` heading has been restored, so no inline reply is needed.

Reviewed 24 of 24 changed files.

---

processing/src/main/java/org/apache/druid/data/input/impl/RegexInputFormat.java
vivek807: Done

FrankChen021 (commented):
I reviewed the follow-up. The regex engine wiring concern is handled: the module is now present in startup injection and the ingestion-facing service paths touched by the follow-up, including indexer, middle manager, overlord/sampler, peon, and coordinator standalone wiring.

Reviewed 24 of 24 changed files.

---

cecemei:
I'm slightly concerned about the multiple interfaces and classes introduced in this PR, I wonder maybe we could just add `RegexMatcher` interface and just have two anonymous implementation from re2j and java.  We could have another `FunctionalInterface` `RegexPattern` returns ` RegexMatcher` given a string, which can be used in `compiledPatternSupplier`.

We could have a system wide default engine, but this could be override in `RegexInputFormat`.

vivek807:
> I'm slightly concerned about the multiple interfaces and classes introduced in this PR, I wonder maybe we could just add `RegexMatcher` interface and just have two anonymous implementation from re2j and java. We could have another `FunctionalInterface` `RegexPattern` returns ` RegexMatcher` given a string, which can be used in `compiledPatternSupplier`.
> 
> We could have a system wide default engine, but this could be override in `RegexInputFormat`.

Thanks @cecemei for input.
@FrankChen021 Would you be also agree to refactor the current implementation to incorporate this?

About last point - We can add it later as separate PR - do not want to over complicate things here

FrankChen021:
> > I'm slightly concerned about the multiple interfaces and classes introduced in this PR, I wonder maybe we could just add `RegexMatcher` interface and just have two anonymous implementation from re2j and java. We could have another `FunctionalInterface` `RegexPattern` returns ` RegexMatcher` given a string, which can be used in `compiledPatternSupplier`.
> > We could have a system wide default engine, but this could be override in `RegexInputFormat`.
> 
> Thanks @cecemei for input. @FrankChen021 Would you be also agree to refactor the current implementation to incorporate this?
> 
> About last point - We can add it later as separate PR - do not want to over complicate things here

Thanks for submitting this PR. Indeed, the matcher and pattern interfaces confused me a little for the first glance of current implementation.

The comment left by @cecemei makes it more intuitive for other developers to understand that the matcher interface is the main implementation of different pattern lib providers.

vivek807:
> > > I'm slightly concerned about the multiple interfaces and classes introduced in this PR, I wonder maybe we could just add `RegexMatcher` interface and just have two anonymous implementation from re2j and java. We could have another `FunctionalInterface` `RegexPattern` returns ` RegexMatcher` given a string, which can be used in `compiledPatternSupplier`.
> > > We could have a system wide default engine, but this could be override in `RegexInputFormat`.
> > 
> > 
> > Thanks @cecemei for input. @FrankChen021 Would you be also agree to refactor the current implementation to incorporate this?
> > About last point - We can add it later as separate PR - do not want to over complicate things here
> 
> Thanks for submitting this PR. Indeed, the matcher and pattern interfaces confused me a little for the first glance of current implementation.
> 
> The comment left by @cecemei makes it more intuitive for other developers to understand that the matcher interface is the main implementation of different pattern lib providers.

Okay, in that case, I’ll refactor it. Thanks.

services/src/main/java/org/apache/druid/cli/CliCoordinator.java
```diff
@@ -178,6 +179,7 @@ protected List<? extends Module> getModules()
       modules.add(new QueryableModule());
     }
 
+    modules.add(new RegexEngineModule());
```
FrankChen021: [P1] Avoid installing RegexEngineModule twice for coordinator-as-overlord

CliCoordinator now always adds RegexEngineModule before the coordinator/overlord split, but when beOverlord is true it then appends CliOverlord.getModules(false), whose list also includes RegexEngineModule. Both modules call JsonConfigProvider.bind for RegexConfig, so coordinator-as-overlord startup gets duplicate explicit bindings, matching the adjacent LookupSerdeModule case this method already avoids, and fails before serving. Only add the regex module on the non-overlord branch, or move it to the shared core module list and remove the per-service copies.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Reviewed 20 of 20 changed files.

---

services/src/main/java/org/apache/druid/cli/CliCoordinator.java
vivek807: Done

FrankChen021 (commented):
I reviewed the follow-up. The duplicate `RegexEngineModule` installation concern is handled: `CliCoordinator` now adds the regex module only in the non-overlord path, so coordinator-as-overlord receives it only through `CliOverlord.getModules(false)`.

Reviewed 20 of 20 changed files.

---

vivek807:
@FrankChen021 Please let me know if there’s anything I can do to help move this PR forward.

FrankChen021 (approved):

PR #19525, opened 2026-05-27
feat: Include supervisorId in Kafka consumer metrics.

gianm:
This patch updates `KafkaConsumerMonitor` to accept the task's metric builder, which includes `supervisorId` as well as other dimensions from `IndexTaskUtils.setTaskDimensions`.


extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaConsumerMonitor.java
```diff
@@ -137,12 +136,23 @@ public class KafkaConsumerMonitor extends AbstractMonitor
...
+   * Supplies a new metric builder for each emitted metric.
+   */
+  @Nullable
+  private final Supplier<ServiceMetricEvent.Builder> metricBuilderSupplier;
+
   private final Map<MetricName, AtomicLong> counters = new HashMap<>();
   private final AtomicDouble pollIdleRatioAvg = new AtomicDouble(1.0d);
 
-  public KafkaConsumerMonitor(final KafkaConsumer<?, ?> consumer)
+  public KafkaConsumerMonitor(
```
FrankChen021, line 149: [P2] Preserve the existing public constructor

This replaces the existing public KafkaConsumerMonitor(KafkaConsumer) constructor with a two-argument signature. Any external extension or test code compiled against the Kafka indexing-service artifact that instantiates the monitor will fail to compile, or hit NoSuchMethodError when run with the updated jar, even though passing null preserves the old behavior. Please keep a one-argument overload that delegates to this constructor with a null supplier.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

---

- extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaRecordSupplier.java:92 - [P2] Keep the old KafkaRecordSupplier overloads. The public KafkaRecordSupplier constructors now require metricBuilderSupplier, removing the existing four-argument production constructor and two-argument test constructor signatures. Downstream code that creates a supplier without task metric dimensions will break at compile time or runtime despite null preserving the old behavior. Please retain overloads that delegate to the new constructors with a null supplier.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaConsumerMonitor.java
gianm: It's essentially an internal utility, so it's fine to modify the constructors.

FrankChen021 (commented):
Handled the follow-up on `KafkaConsumerMonitor` constructor compatibility. The author clarified this is effectively internal, so I do not think another inline reply is useful.

I also reviewed the incremental diff and the full prepared diff for new regressions. No new high-confidence findings.

---

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaSamplerSpec.java
```diff
@@ -69,8 +69,12 @@ protected KafkaRecordSupplier createRecordSupplier()
...
       KafkaSupervisorIOConfig kafkaSupervisorIOConfig = (KafkaSupervisorIOConfig) ioConfig;
 
-      return new KafkaRecordSupplier(props, objectMapper, kafkaSupervisorIOConfig.getConfigOverrides(),
-                                     kafkaSupervisorIOConfig.isMultiTopic()
+      return new KafkaRecordSupplier(
+          props,
+          objectMapper,
+          kafkaSupervisorIOConfig.getConfigOverrides(),
+          kafkaSupervisorIOConfig.isMultiTopic(),
+          null
```
jtuglu1, line 77: Do we even want to emit metrics for sampler specs? Can we treat `null` as don't actually emit anything?

cecemei (approved):

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/KafkaSamplerSpec.java
gianm: Metrics wouldn't be emitted here, since nothing is going to install the `recordSupplier.monitor()`.

PR #19528, opened 2026-05-28
build(deps): Bump jackson to `2.21.3`

amaechler:
### Description

This PR bumps `jackson` (`2.20.2` to `2.21.3`) and adds a missing license entry for `org.jspecify:jspecify` in `extensions-core/kubernetes-extensions`.

#### Bumped Jackson to 2.21.3

Jackson 2.21 changed the default resolution behaviour when `@JacksonInject` and `@JsonProperty` annotate the same parameter: the injected value now wins over the JSON value, where 2.20 treated the inject as a fallback used only when JSON did not supply one. This affects `DruidNode`'s `serviceName`, `port`, and `tlsPort` parameters, which carry both annotations with the expectation that JSON config values win when present; everywhere else the annotations are on distinct parameters and are unaffected. See _Most Wanted Feature: Injection-only `@JacksonInject`_ in the [release blog post](https://cowtowncoder.medium.com/jackson-2-21-released-ce0b6582ae67).

#### Restored JSON-input precedence on `DruidNode`

Added the explicit `useInput = OptBoolean.TRUE` to the three `@JacksonInject` annotations on `DruidNode`'s constructor parameters so that JSON wins over inject.

#### Release note

Dependency Bumps:
* jackson: 2.20.2 → 2.21.3

If external code has both `@JacksonInject` and `@JsonProperty` on the same parameter and relies on the JSON value winning when supplied, add the explicit `useInput = OptBoolean.TRUE` to the annotation (or stay on Jackson 2.20.x). All such sites in Druid itself have been updated.

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/DruidNode.java
+++ b/server/src/main/java/org/apache/druid/server/DruidNode.java
@@ -22,6 +22,7 @@ package org.apache.druid.server;
 import com.fasterxml.jackson.annotation.JacksonInject;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.OptBoolean;
 import com.google.common.base.Preconditions;
 import com.google.common.net.HostAndPort;
 import com.google.inject.name.Named;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/druid/server/DruidNode.java
```diff
@@ -130,12 +131,12 @@ public DruidNode(
    */
   @JsonCreator
   public DruidNode(
-      @JacksonInject @Named("serviceName") @JsonProperty("service") String serviceName,
+      @JacksonInject(useInput = OptBoolean.TRUE) @Named("serviceName") @JsonProperty("service") String serviceName,
```
FrankChen021, line 134: [P1] Preserve useInput through the Guice introspector

These annotations do not survive the production config mapper path. DruidSecondaryModule installs GuiceAnnotationIntrospector, whose findInjectableValue returns JacksonInject.Value.forId(id), rebuilding the JacksonInject.Value from the Guice Key and dropping the annotation's useInput flag. JsonConfigProvider uses that mapper for DruidNode from druid.* properties, so after the Jackson 2.21 bump configured service/port/tlsPort values can still lose to injected defaults despite these annotations. Please preserve the annotation's useInput value when replacing the injectable id, and add a Guice-configured DruidNode config test.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/server/DruidNode.java
amaechler: The static observation is correct: `findInjectableValue` calls `JacksonInject.Value.forId(id)`, which strips `useInput` from the Value. In the current setup the [`AnnotationIntrospectorPair.findInjectableValue` fallback](https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.21.3/src/main/java/com/fasterxml/jackson/databind/introspect/AnnotationIntrospectorPair.java#L289-L300) recovers the missing flag from the secondary (Jackson default) introspector, so visible runtime behavior is correct.

Instead of relying on that fallback though, I pushed a follow-up commit that preserves `useInput` and `optional` directly via `JacksonInject.Value.from(annotation).withId(id)`, removing the reliance.

FrankChen021 (commented):
I reviewed the follow-up. The author update preserves the `JacksonInject` flags directly, so the previous concern is handled and I found no new issues.

---

FrankChen021 (approved):

PR #19530, opened 2026-05-28
feat: add exclude sources clone query mode

jtuglu1:
### Description

This adds a case to the `cloneQueryMode` query context which only mirrors queries to cloned servers. While this can result in partial query results, this allows for things like mirroring traffic/load testing on cloned servers.

This also ensures differing `cloneQueryMode` values correctly skip the broker cache (since results returned by clones are not necessarily equal to that of excludeClones or preferClones). 

#### Release note

Add `excludeSources` option to `cloneQueryMode` query context parameter.

<hr>

This PR has:

- [ ] been self-reviewed.


FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

`excludeSource` queries can return cached data for segments that are not served by clone targets when broker segment cache has entries from prior normal queries.

- server/src/main/java/org/apache/druid/client/CachingClusteredClient.java:364 - [P1] excludeSource can reuse source-backed cache entries. With broker segment cache enabled, pruneSegmentsWithCachedResults runs before groupSegmentsByServer applies ServerSelector.pick(..., EXCLUDESOURCE). A prior non-excludeSource query can populate the same segment cache key for a segment that is not on any clone; the excludeSource query then removes that segment from remote querying and merges the cached rows, even though the new mode says segments missing from clones should be omitted. This defeats clone-only load tests and can turn expected partial results into full/source-backed results. Disable broker cache for EXCLUDESOURCE or make cache lookup/keying depend on clone availability/mode.

---

FrankChen021 (commented):

---

PR #19532, opened 2026-05-30
perf: optimize coordinator HistoricalManagement duty runtime

jtuglu1:
#### High-level changes

- `BalanceSegments` - computes cluster shape (historicals, segments) once per run and reuses it for both the segment-move budget and the summary log, instead of calling getNumHistoricalsAndSegments() twice.
- `MarkOvershadowedSegmentsAsUnused` - builds the set of datasources that actually have overshadowed segments first, then skips timeline construction entirely for every other datasource, both when scanning servers and when checking zero-replica segments.
- `MarkEternityTombstonesAsUnused` - groups overshadowed segments by datasource up front, then checks each candidate tombstone only against its own datasource's group, instead of scanning the full overshadowed-segment list per candidate. Also swaps the timeline-based findNonOvershadowedObjectsInInterval scan for a direct per-segment isOvershadowed check.
- `RunRules` - caches ruleHandler.getRulesWithDefault(datasource) per datasource in a map built during the segment loop, and reuses that same map for the later broadcast-datasource check. Previously rules were looked up once per segment and then a second time per datasource during the broadcast scan.
- `PrepareBalancerAndLoadQueues` - collectUsedSegmentStats now iterates ImmutableDruidDataSource objects directly, using getTotalSizeOfSegments() and getSegments().size(), instead of streaming and summing over every segment in each datasource's timeline.
- `SegmentReplicationStatus` - the constructor no longer makes two full ImmutableMap.copyOf defensive copies. It holds the caller's map directly (safe because it's rebuilt fresh each cycle and not mutated after this point) and computes the total-replica-count map in one pass instead of a separate forEach. Also adds computeSegmentStats(), which does unavailable, under-replicated, and deep-storage-only counting in one pass over used segments instead of three.
- `StrategicSegmentAssigner` - allTiersInCluster is now computed once in the constructor and stored as a field, instead of being rebuilt with Sets.newHashSet(cluster.getTierNames()) on every replicateSegment/replicateSegmentPartially call.

I've kept the logging changes in in-case we deem some of them are debug-level worthy. Otherwise, please ignore those diffs during the review as they will be removed before merge. Initial results have shown a decrease in total HistoricalManagement duty runtime of about 35% on a cluster with 4M segments.

#### Release note

Speed-up coordinator HistoricalManagementDuties duty group

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataQuery.java
+++ b/server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataQuery.java
@@ -1116,7 +1116,6 @@ public class SqlSegmentsMetadataQuery
     return intervals.stream().filter(Objects::nonNull).collect(Collectors.toList());
   }
 
-
   /**
    * Retrieves unused segments that exactly match the given interval.
    *
--- a/server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataQuery.java
+++ b/server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataQuery.java
@@ -1116,6 +1116,7 @@ public class SqlSegmentsMetadataQuery
     return intervals.stream().filter(Objects::nonNull).collect(Collectors.toList());
   }
 
+
   /**
    * Retrieves unused segments that exactly match the given interval.
    *
--- a/server/src/main/java/org/apache/druid/server/coordinator/DruidCoordinator.java
+++ b/server/src/main/java/org/apache/druid/server/coordinator/DruidCoordinator.java
@@ -39,6 +39,7 @@ import org.apache.druid.discovery.DruidLeaderSelector;
 import org.apache.druid.guice.ManageLifecycle;
 import org.apache.druid.indexer.CompactionEngine;
 import org.apache.druid.java.util.common.DateTimes;
+import org.apache.druid.java.util.common.Stopwatch;
 import org.apache.druid.java.util.common.concurrent.ScheduledExecutorFactory;
 import org.apache.druid.java.util.common.concurrent.ScheduledExecutors;
 import org.apache.druid.java.util.common.lifecycle.LifecycleStart;
@@ -782,11 +783,32 @@ public class DruidCoordinator
     @Override
     public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
     {
+      final Stopwatch totalTime = Stopwatch.createStarted();
+
+      final Stopwatch broadcastTime = Stopwatch.createStarted();
       broadcastSegments = params.getBroadcastSegments();
+      broadcastTime.stop();
+
+      final Stopwatch replStatusTime = Stopwatch.createStarted();
       segmentReplicationStatus = params.getSegmentReplicationStatus();
+      replStatusTime.stop();
+
+      final Stopwatch cacheUpdateTime = Stopwatch.createStarted();
       if (coordinatorSegmentMetadataCache != null) {
         coordinatorSegmentMetadataCache.updateSegmentReplicationStatus(segmentReplicationStatus);
       }
+      cacheUpdateTime.stop();
+
+      log.info(
+          "UpdateReplicationStatus summary: broadcastSegments[%,d], hasReplStatus[%s], hasMetaCache[%s];"
+          + " broadcastMs[%,d], replStatusMs[%,d], cacheUpdateMs[%,d], totalMs[%,d].",
+          broadcastSegments == null ? 0 : broadcastSegments.size(),
+          segmentReplicationStatus != null,
+          coordinatorSegmentMetadataCache != null,
+          broadcastTime.millisElapsed(), replStatusTime.millisElapsed(),
+          cacheUpdateTime.millisElapsed(), totalTime.millisElapsed()
+      );
+
       return params;
     }
   }
@@ -799,27 +821,66 @@ public class DruidCoordinator
     @Override
     public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
     {
+      final Stopwatch totalTime = Stopwatch.createStarted();
+
       // Collect stats for unavailable and under-replicated segments
       final CoordinatorRunStats stats = params.getCoordinatorStats();
-      getDatasourceToUnavailableSegmentCount().forEach(
-          (dataSource, numUnavailable) -> stats.add(
-              Stats.Segments.UNAVAILABLE,
-              RowKey.of(Dimension.DATASOURCE, dataSource),
-              numUnavailable
-          )
-      );
-      getTierToDatasourceToUnderReplicatedCount(false).forEach(
-          (tier, countsPerDatasource) -> countsPerDatasource.forEach(
-              (dataSource, underReplicatedCount) ->
-                  stats.addToSegmentStat(Stats.Segments.UNDER_REPLICATED, tier, dataSource, underReplicatedCount)
-          )
-      );
-      getDatasourceToDeepStorageQueryOnlySegmentCount().forEach(
-          (dataSource, numDeepStorageOnly) -> stats.add(
-              Stats.Segments.DEEP_STORAGE_ONLY,
-              RowKey.of(Dimension.DATASOURCE, dataSource),
-              numDeepStorageOnly
-          )
+
+      final Object2IntMap<String> dsToUnavailable;
+      final Map<String, Object2LongMap<String>> tierToDsToUnderRepl;
+      final Object2IntMap<String> dsToDeepStorageOnly;
+
+      if (segmentReplicationStatus == null) {
+        dsToUnavailable = Object2IntMaps.emptyMap();
+        tierToDsToUnderRepl = Collections.emptyMap();
+        dsToDeepStorageOnly = Object2IntMaps.emptyMap();
+      } else {
+        // Single fused pass replaces three independent full iterations over
+        // metadataManager.iterateAllUsedSegments(). ignoreMissingServers=true replicates
+        // the !false inversion inside computeUnderReplicated(segments, false).
+        final SegmentReplicationStatus.SegmentStatsSnapshot snapshot = segmentReplicationStatus.computeSegmentStats(
+            metadataManager.iterateAllUsedSegments(),
+            true
+        );
+        dsToUnavailable = snapshot.getDatasourceToUnavailableCount();
+        tierToDsToUnderRepl = snapshot.getTierToDatasourceToUnderReplicatedCount();
+        dsToDeepStorageOnly = snapshot.getDatasourceToDeepStorageOnlyCount();
+      }
+
+      long unavailableTotal = 0;
+      for (final Object2IntMap.Entry<String> e : dsToUnavailable.object2IntEntrySet()) {
+        unavailableTotal += e.getIntValue();
+        stats.add(Stats.Segments.UNAVAILABLE, RowKey.of(Dimension.DATASOURCE, e.getKey()), e.getIntValue());
+      }
+
+      long underReplTotal = 0;
+      int tierCount = 0;
+      for (final Map.Entry<String, Object2LongMap<String>> tierEntry : tierToDsToUnderRepl.entrySet()) {
+        tierCount++;
+        for (final Object2LongMap.Entry<String> dsEntry : tierEntry.getValue().object2LongEntrySet()) {
+          underReplTotal += dsEntry.getLongValue();
+          stats.addToSegmentStat(
+              Stats.Segments.UNDER_REPLICATED,
+              tierEntry.getKey(),
+              dsEntry.getKey(),
+              dsEntry.getLongValue()
+          );
+        }
+      }
+
+      long deepStorageTotal = 0;
+      for (final Object2IntMap.Entry<String> e : dsToDeepStorageOnly.object2IntEntrySet()) {
+        deepStorageTotal += e.getIntValue();
+        stats.add(Stats.Segments.DEEP_STORAGE_ONLY, RowKey.of(Dimension.DATASOURCE, e.getKey()), e.getIntValue());
+      }
+
+      log.info(
+          "CollectSegmentStats summary: unavailableTotal[%,d] across[%d] ds,"
+          + " underReplicatedTotal[%,d] across[%d] tiers, deepStorageOnly[%,d] across[%d] ds; totalMs[%,d].",
+          unavailableTotal, dsToUnavailable.size(),
+          underReplTotal, tierCount,
+          deepStorageTotal, dsToDeepStorageOnly.size(),
+          totalTime.millisElapsed()
       );
 
       return params;
--- a/server/src/main/java/org/apache/druid/server/coordinator/DruidCoordinator.java
+++ b/server/src/main/java/org/apache/druid/server/coordinator/DruidCoordinator.java
@@ -39,7 +39,6 @@ import org.apache.druid.discovery.DruidLeaderSelector;
 import org.apache.druid.guice.ManageLifecycle;
 import org.apache.druid.indexer.CompactionEngine;
 import org.apache.druid.java.util.common.DateTimes;
-import org.apache.druid.java.util.common.Stopwatch;
 import org.apache.druid.java.util.common.concurrent.ScheduledExecutorFactory;
 import org.apache.druid.java.util.common.concurrent.ScheduledExecutors;
 import org.apache.druid.java.util.common.lifecycle.LifecycleStart;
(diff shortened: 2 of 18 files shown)


FrankChen021 (commented):

Reviewed 15 of 15 changed files.

---

server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataQuery.java
```diff
@@ -1093,6 +1093,7 @@ public List<Interval> retrieveUnusedSegmentIntervals(String dataSource, int limi
     return intervals.stream().filter(Objects::nonNull).collect(Collectors.toList());
   }
 
+
```
kfaraz: nit: not required

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
```diff
@@ -86,6 +89,9 @@ public StrategicSegmentAssigner(
     this.cluster = cluster;
     this.strategy = strategy;
     this.loadQueueManager = loadQueueManager;
+    final Set<String> allTiersInCluster = new HashSet<>();
+    cluster.getTierNames().forEach(allTiersInCluster::add);
+    this.allTiersInCluster = allTiersInCluster;
```
kfaraz: to ensure immutability:

```suggestion
    this.allTiersInCluster = Set.copyOf(cluster.getTierNames());
```

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
```diff
@@ -64,6 +63,10 @@ public class StrategicSegmentAssigner implements SegmentActionHandler
   private final RoundRobinServerSelector serverSelector;
   private final BalancerStrategy strategy;
 
+  // Fixed for the lifetime of this assigner (a new instance is created each coordinator run),
+  // so it is computed once here rather than on every replicateSegment/replicateSegmentPartially call.
```
kfaraz: This comment can be omitted since we are already maintaining other similar fields which are valid for the current run such as `historicalTierAliases` and `tierToAliasName`.

FrankChen021 (commented):
I have reviewed the updated code for correctness, edge cases, concurrency, lifecycle, compatibility, performance semantics, and integration risks; no new issues found.

---

server/src/main/java/org/apache/druid/server/coordinator/duty/MarkEternityTombstonesAsUnused.java
```diff
@@ -132,43 +130,57 @@ public DruidCoordinatorRuntimeParams run(final DruidCoordinatorRuntimeParams par
...
 
-    dataSourcesSnapshot.getDataSourcesMap().keySet().forEach((datasource) -> {
+    for (final DataSegment overshadowedSegment : dataSourcesSnapshot.getOvershadowedSegments()) {
+      overshadowedSegmentsByDatasource
+          .computeIfAbsent(overshadowedSegment.getDataSource(), ds -> new HashSet<>())
+          .add(overshadowedSegment);
+    }
+
+    for (final ImmutableDruidDataSource dataSource : dataSourcesSnapshot.getDataSourcesWithAllUsedSegments()) {
+      final String datasource = dataSource.getName();
```
kfaraz: To avoid confusion with the other variable `dataSource`.
```suggestion
      final String datasourceName = dataSource.getName();
```

server/src/main/java/org/apache/druid/server/coordinator/duty/MarkOvershadowedSegmentsAsUnused.java
```diff
@@ -80,23 +80,33 @@ public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
       return params;
     }
 
+    // Identify the datasources that actually have overshadowed segments to check.
+    // Timelines only need to be built for these datasources.
+    final Set<String> relevantDatasources = new HashSet<>();
```
kfaraz:
```suggestion
    final Set<String> eligibleDatasources = new HashSet<>();
```

server/src/main/java/org/apache/druid/server/coordinator/duty/MarkOvershadowedSegmentsAsUnused.java
```diff
@@ -80,23 +80,33 @@ public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
...
+    for (DataSegment s : allOvershadowedSegments) {
+      relevantDatasources.add(s.getDataSource());
+    }
```
kfaraz: Nit: rename and make immutable
```suggestion
    final Set<String> eligibleDatasources = allOvershadowedSegments
        .stream()
        .map(DataSegment::getDataSource)
        .collect(Collectors.toSet());
```

server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCount.java
```diff
@@ -36,6 +36,25 @@ public class SegmentReplicaCount
...
   private int movingFrom;
 
+  SegmentReplicaCount()
+  {
+  }
+
+  /**
+   * Copies the counts out of {@code other} into a new, independent instance.
+   */
+  SegmentReplicaCount(SegmentReplicaCount other)
```
kfaraz: Instead of this, maybe add a `copy()` method.

server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicaCount.java
```diff
@@ -36,6 +36,25 @@ public class SegmentReplicaCount
   private int movingTo;
   private int movingFrom;
 
+  SegmentReplicaCount()
```
kfaraz: This shouldn't be needed.

server/src/main/java/org/apache/druid/server/coordinator/duty/MarkEternityTombstonesAsUnused.java
```diff
@@ -132,43 +130,57 @@ public DruidCoordinatorRuntimeParams run(final DruidCoordinatorRuntimeParams par
...
       final SegmentTimeline usedSegmentsTimeline
           = dataSourcesSnapshot.getUsedSegmentsTimelinesPerDataSource().get(datasource);
+      if (usedSegmentsTimeline == null) {
```
kfaraz: Can this ever happen? We are already iterating over the datasources provided by the snapshot itself, so the timeline must be non null.

server/src/main/java/org/apache/druid/server/coordinator/duty/MarkEternityTombstonesAsUnused.java
```diff
@@ -132,43 +130,57 @@ public DruidCoordinatorRuntimeParams run(final DruidCoordinatorRuntimeParams par
...
-            }
-          }
-        });
+      for (final DataSegment candidateSegment : dataSource.getSegments()) {
+        if (isNewGenerationEternityTombstone(candidateSegment)
+            && !usedSegmentsTimeline.isOvershadowed(candidateSegment)
+            && !overlapsAnyOvershadowedSegment(
+                candidateSegment,
+                overshadowedSegmentsByDatasource.get(candidateSegment.getDataSource())
+            )) {
```
kfaraz: Let's collapse these 3 conditions into a single new method for readability of this for loop:

```
shouldMarkAsUnused(candidateSegment, overshadowedSegmentsByDatasource, timeline)
```

the new method can check the 3 required conditions by invoking respective methods.

server/src/main/java/org/apache/druid/server/coordinator/duty/PrepareBalancerAndLoadQueues.java
```diff
@@ -203,14 +203,14 @@ private void collectHistoricalStats(
 
   private void collectUsedSegmentStats(DruidCoordinatorRuntimeParams params, CoordinatorRunStats stats)
   {
-    params.getUsedSegmentsTimelinesPerDataSource().forEach((dataSource, timeline) -> {
-      long totalSizeOfUsedSegments = timeline.iterateAllObjects().stream()
-                                             .mapToLong(DataSegment::getSize).sum();
+    for (final ImmutableDruidDataSource dataSource : params.getDataSourcesSnapshot().getDataSourcesWithAllUsedSegments()) {
+      final long totalSizeOfUsedSegments = dataSource.getTotalSizeOfSegments();
+      final int numUsedSegments = dataSource.getSegments().size();
```
kfaraz: We should also add a method `ImmutableDataSource.getNumSegments()` to avoid materializing the collection inside `ImmutableDataSource.getSegments()`.

server/src/main/java/org/apache/druid/server/coordinator/duty/RunRules.java
```diff
@@ -98,14 +99,15 @@ public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
 
     final DateTime now = DateTimes.nowUtc();
     final Object2IntOpenHashMap<String> datasourceToSegmentsWithNoRule = new Object2IntOpenHashMap<>();
+    final Map<String, List<Rule>> datasourceToRules = new HashMap<>();
```
kfaraz: Rules need not be cached here as they are already served from the cache in `SQLMetadataRuleManager` itself. The changes to this class may be reverted.

server/src/main/java/org/apache/druid/server/coordinator/DruidCoordinator.java
```diff
@@ -801,26 +803,46 @@ public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
...
-          (dataSource, numDeepStorageOnly) -> stats.add(
-              Stats.Segments.DEEP_STORAGE_ONLY,
-              RowKey.of(Dimension.DATASOURCE, dataSource),
-              numDeepStorageOnly
-          )
-      );
+
+      final Object2IntMap<String> dsToUnavailable;
+      final Map<String, Object2LongMap<String>> tierToDsToUnderRepl;
+      final Object2IntMap<String> dsToDeepStorageOnly;
```
kfaraz: Please use full names e.g. `datasourceToUnavailableCount`, `datasourceToUnderReplicatedCount`.

Do we need to keep 3 separate map references? Maybe just use an empty instance of `SegmentStatsSnapshot` and invoke getters on it where needed.

server/src/main/java/org/apache/druid/server/coordinator/DruidCoordinator.java
```diff
@@ -801,26 +803,46 @@ public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
...
+        dsToUnavailable = snapshot.getDatasourceToUnavailableCount();
+        tierToDsToUnderRepl = snapshot.getTierToDatasourceToUnderReplicatedCount();
+        dsToDeepStorageOnly = snapshot.getDatasourceToDeepStorageOnlyCount();
+      }
+
+      for (final Object2IntMap.Entry<String> e : dsToUnavailable.object2IntEntrySet()) {
+        stats.add(Stats.Segments.UNAVAILABLE, RowKey.of(Dimension.DATASOURCE, e.getKey()), e.getIntValue());
+      }
+
+      for (final Map.Entry<String, Object2LongMap<String>> tierEntry : tierToDsToUnderRepl.entrySet()) {
```
kfaraz: Using `forEach` and `fastForEach` instead of traditional for loops may make this more readable.

server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicationStatus.java
```diff
@@ -81,4 +96,84 @@ public Map<String, Object2LongMap<String>> getTierToDatasourceToUnderReplicated(
...
+      }
+    }
+
+    return new SegmentStatsSnapshot(datasourceToUnavailable, tierToUnderReplicated, datasourceToDeepStorageOnly);
+  }
+
+  /**
+   * Holder for the three segment-stat views computed together by {@link #computeSegmentStats}.
+   */
+  public static class SegmentStatsSnapshot
```
kfaraz: Please move out this class into a separate file as it seems completely independent.

server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicationStatus.java
```diff
@@ -40,15 +43,27 @@ public class SegmentReplicationStatus
 
   public SegmentReplicationStatus(Map<SegmentId, Map<String, SegmentReplicaCount>> replicaCountsInTier)
   {
-    this.replicaCountsInTier = ImmutableMap.copyOf(replicaCountsInTier);
-
-    final Map<SegmentId, SegmentReplicaCount> totalReplicaCounts = new HashMap<>();
-    replicaCountsInTier.forEach((segmentId, tierToReplicaCount) -> {
+    // replicaCountsInTier is the caller's live SegmentReplicaCountMap and is mutated further
+    // in the same coordinator cycle (e.g. by BalanceSegments), so we must snapshot both the
+    // structure and the per-tier SegmentReplicaCount values, not just alias the outer map.
```
kfaraz: Nit: simpler language
```suggestion
    // Make a deep copy of replicaCountsInTier as it may be mutated further
    // in the same coordinator cycle (e.g. by BalanceSegments)
```

server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicationStatus.java
```diff
@@ -81,4 +96,84 @@ public Map<String, Object2LongMap<String>> getTierToDatasourceToUnderReplicated(
...
+
+      final SegmentReplicaCount totalCount = totalReplicaCounts.get(segmentId);
+      if (totalCount != null && (totalCount.totalLoaded() > 0 || totalCount.required() == 0)) {
+        datasourceToUnavailable.addTo(datasource, 0);
+      } else {
+        datasourceToUnavailable.addTo(datasource, 1);
+      }
+      if (totalCount != null && totalCount.totalLoaded() == 0 && totalCount.required() == 0) {
+        datasourceToDeepStorageOnly.addTo(datasource, 1);
+      }
```
kfaraz: This logic is now duplicated with `DruidCoordinator`.
Please move unavailable logic and deep storage logic into new methods in this class so that `DruidCoordinator` may reuse it too.

```java
boolean isUnavailable(SegmentId segmentId);

boolean isUnderReplicated(SegmentId segmentId);
```

server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicationStatus.java
```diff
@@ -81,4 +96,84 @@ public Map<String, Object2LongMap<String>> getTierToDatasourceToUnderReplicated(
...
+      final Map<String, SegmentReplicaCount> tierToReplicaCount = replicaCountsInTier.get(segmentId);
+      if (tierToReplicaCount != null) {
+        tierToReplicaCount.forEach((tier, counts) -> {
+          final int underReplicated = ignoreMissingServers ? counts.missing() : counts.missingAndLoadable();
+          if (underReplicated >= 0) {
+            Object2LongOpenHashMap<String> datasourceToUnderReplicated = (Object2LongOpenHashMap<String>)
+                tierToUnderReplicated.computeIfAbsent(tier, ds -> new Object2LongOpenHashMap<>());
+            datasourceToUnderReplicated.addTo(datasource, underReplicated);
+          }
+        });
```
kfaraz, line 149: Can we replace this by a call to `getTierToDatasourceToUnderReplicated()` since it is the same logic?

kfaraz (commented):
Overall looks good, left some suggestions.

server/src/main/java/org/apache/druid/server/coordinator/DruidCoordinator.java
```diff
@@ -801,26 +803,46 @@ public DruidCoordinatorRuntimeParams run(DruidCoordinatorRuntimeParams params)
...
+      final Object2IntMap<String> dsToDeepStorageOnly;
+
+      if (segmentReplicationStatus == null) {
+        dsToUnavailable = Object2IntMaps.emptyMap();
+        tierToDsToUnderRepl = Collections.emptyMap();
+        dsToDeepStorageOnly = Object2IntMaps.emptyMap();
+      } else {
+        // Single fused pass replaces three independent full iterations over
+        // metadataManager.iterateAllUsedSegments(). ignoreMissingServers=true replicates
+        // the !false inversion inside computeUnderReplicated(segments, false).
```
kfaraz: simplify the comment. details may be found in the javadoc of `computeSegmentStats` itself.
```suggestion
        // Compute unavailable, under-replicated, deep-storage only counts in a single pass
```

kfaraz (approved):
LGTM 🚀

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
```diff
@@ -86,6 +87,7 @@ public StrategicSegmentAssigner(
     this.cluster = cluster;
     this.strategy = strategy;
     this.loadQueueManager = loadQueueManager;
+    this.allTiersInCluster = Set.copyOf(cluster.getTierNames());
```
FrankChen021, line 90: [P2] Preserve null tier handling in the cached set

`DruidCluster.Builder.addTier` and `DruidServerMetadata` both permit a null tier, and the previous `HashSet` copy preserved such a key. `Set.copyOf` rejects null elements, so one Historical with a null or omitted tier now throws here and aborts the HistoricalManagement duty before replica assignment. Use a null-tolerant copy, or normalize or reject null when constructing the cluster.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Found one coordinator robustness regression: caching tier names with `Set.copyOf` rejects a null or omitted Historical tier and aborts replica assignment.

---

server/src/main/java/org/apache/druid/server/coordinator/loading/StrategicSegmentAssigner.java
kfaraz: We can reject nulls when constructing the `DruidCluster` in a follow up PR.

server/src/main/java/org/apache/druid/server/coordinator/loading/SegmentReplicationStatus.java
kfaraz: @jtuglu1 , is this possible? If yes, then we can do it in a follow up.
Otherwise, please feel free to resolve this comment.

jtuglu1: The goal for this change was to avoid another full pass over usedSegments. I suppose we can do something like:

```java
final Map<...> tierToUnderReplicated =
    getTierToDatasourceToUnderReplicated(usedSegments, ignoreMissingServers);
```

as well to avoid any n^2 behavior during computeSegmentStats().

jtuglu1: I can run a few benchmarks and see if this has any material impact (the original goal was to avoid looping through 3x and instead fuse everything into a single pass).

PR #19534, opened 2026-06-01
feat: Support V2 Format for Iceberg Ingestion

a2l007:
Fixes #19190 

### Description
This PR fixes Iceberg delete semantics which was previously broken for v2 tables. No spec changes are required. The same ingestion spec works for V1 and V2 tables. Delete detection is automatic at planning time.

In the V2 path, each `FileScanTask` is encoded as an `IcebergFileTaskInputSource` , containing the data file path, format, size, delete file metadata (paths, content type, equality field IDs), and the table schema captured as JSON at planning time.

IcebergNativeRecordReader processes each split:

  1. Delete metadata is loaded upfront. For each associated delete file is processed based on whether it is a position delete file or an equality delete file.
    - Position delete files: Rows to skip are read and filtered to only positions matching the current data file path.
    - Equality delete files : Rows of key tuples are read per equality delete file.
  2. The data file is streamed lazily using Iceberg's Parquet.read() builder with GenericParquetReaders. 
  3. Deletes are applied per record: each record's position is checked against the position delete set; its field values are checked against each equality delete set. Matching records are skipped.
  4. Surviving records are converted to `MapBasedInputRow` via `IcebergRecordConverter`

  FileIO is resolved per path: local paths use `WarehouseFileIO`, non-local paths (HDFS, S3) fall back to `catalog.loadTable().io()` using the namespace and table name passed through the split.

#### Release note

The `druid-iceberg-extensions` input source now correctly handles Iceberg V2 tables that contain row-level delete files. Previously, position deletes and equality deletes were silently ignored, causing deleted rows to appear in ingested segments. Druid now auto-detects delete files at planning time and applies them during ingestion with no spec changes required.

<hr>

 * `IcebergCatalog`
 * `IcebergInputSource`

<hr>

This PR has:

- [x] been self-reviewed.


extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergNativeRecordReader.java
```diff
@@ -0,0 +1,417 @@
...
+ *       {@link IcebergRecordConverter}.</li>
+ * </ol>
+ *
+ * <p>Only Parquet format is supported for the initial implementation.
+ */
+public class IcebergNativeRecordReader implements InputSourceReader
+{
+  private static final Logger log = new Logger(IcebergNativeRecordReader.class);
+
+  private static final WarehouseFileIO WAREHOUSE_FILE_IO = new WarehouseFileIO();
```
FrankChen021, line 87: [P1] Undefined FileIO class breaks compilation

The new reader instantiates `WarehouseFileIO`, but no class with that name is defined in this package or elsewhere in the repo, and there is no import that could resolve it. Any build that compiles this extension will fail before the V2 ingestion path can run. Use an existing Iceberg `FileIO` implementation or add the missing implementation.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -121,8 +150,7 @@ public List<String> extractSnapshotDataFiles(
         }
       }
 
-      // Handle residual filter based on mode
-      if (detectedResidual != null) {
+      if (detectedResidual == null) {
```
FrankChen021, line 153: [P2] Residual-filter check is inverted

This condition now enters the residual-handling branch when `detectedResidual` is null, even though null means no residual was found. With `ResidualFilterMode.FAIL`, valid scans that have no residual, such as partition-pruned scans, will throw a residual-filter error, while scans that do have a residual skip the warning/failure path. Restore the check to run only when `detectedResidual != null`.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

---

FrankChen021 (commented):
## 🟡 Changes recommended

The current V2 reader has one blocking correctness issue and two additional behavior regressions: nested equality deletes are not resolved through Iceberg's nested field names, fixed-width binary equality keys cannot match because byte arrays use identity equality, and the native V2 path bypasses the configured Parquet input format. The two earlier MergeLens findings on the missing FileIO class and inverted residual-filter check are resolved in the current head.

Validation: static review of the incremental diff, full merge-base diff, current surrounding code, Druid input-source/task wiring, and the Iceberg 1.11.0 local sources; `git diff --check` passed. No builds, tests, dependency installs, or formatters were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 2 |
| P3 | 0 |
| Total | 3 |



---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergNativeRecordReader.java
```diff
@@ -0,0 +1,417 @@
...
+    {
+      this.fieldNames = fieldNames;
+      this.keys = keys;
+    }
+
+    boolean matches(Record dataRecord)
+    {
+      Map<String, Object> key = new HashMap<>();
+      for (String fieldName : fieldNames) {
+        key.put(fieldName, dataRecord.getField(fieldName));
```
FrankChen021: [P1] Nested equality deletes never match

**Finding:** For an equality delete on a nested field, Schema.findColumnName returns a dotted path such as `payload.id`. The projected equality schema contains the top-level `payload` struct, but GenericRecord.getField only resolves fields in that record's immediate schema, so both the delete key and each data-row key are built with null for the dotted field and the delete is never applied. Any nested equality delete therefore leaves rows that Iceberg says are deleted.

**Suggestion:** Extract equality values by field ID or recursively traverse the projected records so nested delete fields are compared with their corresponding nested data values.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -116,6 +139,34 @@ public InputSourceReader reader(
     if (!isLoaded) {
       retrieveIcebergDatafiles();
     }
+    if (hasDeleteFiles) {
+      // V2 path: create a combined reader across all file-scan tasks.
+      final List<IcebergNativeRecordReader> taskReaders = v2Tasks.stream()
```
FrankChen021: [P2] V2 reads ignore the configured input format

**Finding:** When delete files are present, the reader is constructed from only InputRowSchema and the supplied InputFormat is never passed to IcebergNativeRecordReader. The documented Iceberg ingestion specs use ParquetInputFormat, whose flattenSpec, binaryAsString, and related parsing behavior is applied on the V1 delegate path; V2 tables silently skip those settings and can produce different or missing dimensions compared with the same spec before deletes were added.

**Suggestion:** Preserve the configured input-format semantics on the V2 path, either by carrying the relevant Parquet options into the native reader or by applying deletes around the existing Druid input-format reader.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergNativeRecordReader.java
```diff
@@ -0,0 +1,417 @@
...
+      }
+      return keys.contains(key);
```
FrankChen021, line 440: [P2] Fixed binary equality keys use identity comparison

**Finding:** GenericParquetReaders.FixedReader returns a new byte[] for each fixed-width value. The delete key map and the data-row key map therefore contain distinct arrays, and HashMap.equals compares those arrays by identity rather than contents; equality deletes on Iceberg fixed fields never match and the deleted rows remain visible.

**Suggestion:** Canonicalize binary and fixed values to a content-based representation, or use deep/content equality when constructing and checking equality-delete keys.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -116,6 +140,34 @@ public InputSourceReader reader(
...
```
FrankChen021: [P2] V2 path drops configured input-format behavior

**Finding:** When any planned task has delete files, this branch routes every task through IcebergNativeRecordReader, but the reader is constructed with only InputRowSchema and the InputFormat argument is discarded. That changes the documented Parquet ingestion behavior for V2 tables: ParquetInputFormat options such as flattenSpec and binaryAsString are no longer applied, and even tasks without deletes are affected because the whole source switches to this branch.

**Suggestion:** Carry the configured Parquet input-format options into the native reader, or preserve the existing warehouse Parquet reader and layer the Iceberg delete filtering around its rows so the same InputFormat semantics are used on both paths.

FrankChen021 (commented):
## 🟡 Changes recommended

The current V2 delete-file path still has two concrete compatibility regressions: it bypasses the configured Parquet InputFormat, and its native row conversion does not apply Druid dimension discovery/include-all/exclusion semantics. Both can change ingestion results for otherwise valid configurations.

The earlier findings for nested equality deletes and fixed-width binary equality keys appear resolved in the current head. The prior FileIO and residual-filter findings also remain resolved.

Scope: updated-since-review incremental diff from 5fc1408bf0e6392dda861adf7e396b4164560e1b, the full merge-base diff, and relevant Druid/Iceberg surrounding code.

Validation: static review only; `git diff --check` passed. No builds, tests, dependency installs, or formatters were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |



---

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergRecordConverter.java
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * Convert a raw-value map plus an {@link InputRowSchema} into a {@link MapBasedInputRow}.
+   * If the dimensions list is empty (auto mode), all columns except the timestamp column are used.
+   */
+  public static InputRow toInputRow(Map<String, Object> row, InputRowSchema inputRowSchema)
+  {
+    DateTime timestamp = inputRowSchema.getTimestampSpec().extractTimestamp(row);
+
+    List<String> dimensions = inputRowSchema.getDimensionsSpec().getDimensionNames();
```
FrankChen021, line 76: [P2] Native conversion ignores dimension discovery rules

**Finding:** The native V2 conversion uses only getDimensionNames(), and when that list is empty it treats every non-timestamp column as a dimension. It never honors DimensionsSpec.useSchemaDiscovery(), isIncludeAllDimensions(), or getDimensionExclusions(). A V2 ingestion can therefore omit columns that schema discovery/include-all should add, or index excluded columns as dimensions, producing different rows from the existing V1 Parquet reader.

**Suggestion:** Build the dimension list with the same MapInputRowParser.findDimensions semantics, including schema discovery/include-all behavior and exclusions, before constructing MapBasedInputRow.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
```diff
@@ -283,6 +309,37 @@ public InputSourceReader reader(
         File temporaryDirectory
     )
     {
+      if (!isLoaded) {
+        retrieveIcebergDatafiles();
+      }
+      if (hasDeleteFiles) {
+        // V2 path: create a combined reader across all file-scan tasks.
+        final List<IcebergNativeRecordReader> taskReaders = v2Tasks.stream()
```
FrankChen021, line 317: [P2] V2 reads ignore the configured input format

**Finding:** When any planned task has delete files, this branch routes every task through IcebergNativeRecordReader, but the reader is constructed only from InputRowSchema and the supplied InputFormat is discarded. ParquetInputFormat options such as flattenSpec and binaryAsString are therefore skipped for V2 tables, including data files without deletes, so a valid ingestion spec can produce different or missing indexed values from the existing warehouse-reader path.

**Suggestion:** Preserve the configured Parquet input-format semantics on the delete-aware path, either by carrying the relevant options into the native reader or by applying delete filtering around the existing Druid input-format reader.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergRecordConverter.java
FrankChen021, line 76: [P2] Native conversion ignores dimension discovery rules

**Finding:** The native V2 conversion uses only DimensionsSpec.getDimensionNames(), and when that list is empty it treats every non-timestamp column as a dimension. It never applies DimensionsSpec.useSchemaDiscovery(), includeAllDimensions, or dimensionExclusions, so V2 ingestion can omit dimensions that discovery/include-all should add or index fields that the ingestion spec explicitly excludes.

**Suggestion:** Resolve dimensions with the same MapInputRowParser.findDimensions semantics as the normal input-format path, including discovery/include-all behavior and exclusions, before constructing MapBasedInputRow.

extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
```diff
@@ -114,42 +147,63 @@ public List<String> extractSnapshotDataFiles(
...
         tableScan = tableScan.asOfTime(snapshotTime.getMillis());
       }
-
       tableScan = tableScan.caseSensitive(isCaseSensitive());
-      enforceResidualMode(tableScan, residualFilterMode);
-      try (CloseableIterable<FileScanTask> tasks = tableScan.planFiles()) {
-        for (FileScanTask task : tasks) {
-          dataFilePaths.add(task.file().location());
+
+      CloseableIterable<FileScanTask> taskIterable = tableScan.planFiles();
```
FrankChen021, line 175: [P3] Close the planned scan iterable

**Finding:** extractFileScanTasksWithSchema obtains the CloseableIterable returned by TableScan.planFiles() and iterates it without closing it. Iceberg wraps scan planning in a close-completion callback that finalizes planning metrics and closes nested manifest iterables; the old path used try-with-resources. Normal scans therefore skip completion, and exception or early-failure paths can retain planning resources across repeated ingestion attempts.

**Suggestion:** Iterate the planned tasks inside try-with-resources and close the CloseableIterable on both normal completion and failure.

FrankChen021 (commented):
## 🟡 Changes recommended

The current head retains two correctness regressions in the Iceberg V2 delete-aware path: it bypasses configured Parquet InputFormat behavior, and its native row conversion does not honor Druid dimension discovery/include-all/exclusion semantics. I also found that the new scan-planning path leaves Iceberg's CloseableIterable unclosed, which skips scan completion and can leak planning resources on failure.

The earlier findings for nested equality deletes and fixed-width binary equality keys appear resolved in the current head. The prior FileIO and residual-filter findings also remain resolved.

Reviewed 10 of 10 changed files:

- docs/development/extensions-contrib/iceberg.md
- extensions-contrib/druid-iceberg-extensions/pom.xml
- extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/common/IcebergDruidModule.java
- extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergCatalog.java
- extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergFileTaskInputSource.java
- extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergInputSource.java
- extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergNativeRecordReader.java
- extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/IcebergRecordConverter.java
- extensions-contrib/druid-iceberg-extensions/src/main/java/org/apache/druid/iceberg/input/WarehouseFileIO.java
- extensions-contrib/druid-iceberg-extensions/src/test/java/org/apache/druid/iceberg/input/IcebergInputSourceTest.java

Scope: started from the prepared incremental diff, then reviewed the full merge-base diff, prior review findings, current-head code, Druid input-source/task wiring, and relevant Iceberg local sources.

Validation: static review only; git diff --check passed. No builds, tests, dependency installs, or formatters were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 1 |
| Total | 3 |



---

PR #19535, opened 2026-06-01
feat: on-demand partial segment loading in SegmentLocalCacheManager

clintropolis:
### Description
changes:
* adds `acquirePartialSegment` / `acquireCachedPartialSegment` to `SegmentCacheManager` and `SegmentManager` to allow callers to opt-in to async partial segments; MSQ `RegularLoadableSegment` uses the new partial path
* `PartialSegmentMetadataCacheEntry`, `PartialSegmentBundleCacheEntry` are now wired into `SegmentLocalCacheManager`, along with added `PartialBundleAcquirer` helper to pass things like download thread pool and the ability to acquire reference holds on cache entrie
* adds `PartialQueryableIndexSegment`, `PartialQueryableIndexCursorFactory`, `V10TimeBoundaryInspector` for references acquired from the now wired up metadata cache entry, implementing the async cursor holder contract
* moved `DirectoryBackedRangeReader` out of tests to use as the implementation for `LocalLoadSpec` range reader.
* add `FilteredCursorFactory`/`RestrictedCursorFactory` async cursor implementations
* adds config `druid.segmentCache.virtualStoragePartialDownloadsEnabled to enable feature, on iff druid.segmentCache.virtualStorage is on


FrankChen021 (commented):

Reviewed 37 of 37 changed files.

---

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexSegment.java
```diff
@@ -0,0 +1,110 @@
...
+  @Override
+  public void close()
+  {
+    CloseableUtils.closeAndWrapExceptions(onClose);
+  }
+
+  @SuppressWarnings("unchecked")
+  @Nullable
+  @Override
+  public <T> T as(@Nonnull Class<T> clazz)
```
gianm, line 93: Is it possible to include a `PhysicalSegmentInspector`? It would be useful to get the number of rows, which is used for incrementing certain counters.

server/src/main/java/org/apache/druid/server/SegmentManager.java
```diff
@@ -225,6 +224,32 @@ public AcquireSegmentAction acquireSegment(DataSegment dataSegment)
...
+   * still return a segment if it is present in cache and any async methods will function properly and return
+   * immediately.
+   */
+  public Optional<Segment> acquireCachedPartialSegment(SegmentId segmentId)
+  {
+    return cacheManager.acquireCachedPartialSegment(segmentId);
+  }
+
+  /**
+   * Partial-load variant of {@link #acquireSegment(DataSegment)}, returns an {@link AcquireSegmentAction} that
```
gianm: Is the segment returned by `acquireSegment` guaranteed to be fully-loaded? (Its javadocs do not seem to make that clear.)

server/src/main/java/org/apache/druid/server/SegmentManager.java
```diff
@@ -225,6 +224,32 @@ public AcquireSegmentAction acquireSegment(DataSegment dataSegment)
     return cacheManager.acquireSegment(dataSegment);
   }
 
+  /**
+   * Partial-load variant of {@link #acquireCachedSegment(SegmentId)}, returns a {@link Segment} when the cache holds
```
gianm: Is the segment returned by `acquireCachedSegment` guaranteed to be fully-loaded? (Its javadocs do not seem to make that clear.)

multi-stage-query/src/main/java/org/apache/druid/msq/input/RegularLoadableSegment.java
```diff
@@ -140,7 +140,7 @@ public synchronized Optional<Segment> acquireIfCached()
       throw DruidException.defensive("Segment with descriptor[%s] is already acquired", descriptor);
     }
 
-    final Optional<Segment> cachedSegment = segmentManager.acquireCachedSegment(segmentId);
+    final Optional<Segment> cachedSegment = segmentManager.acquireCachedPartialSegment(segmentId);
```
gianm: Make it an option whether `RegularLoadableSegment` does a partial fetch or full fetch? It may not be a safe assumption that all MSQ implementations of query processing are going to be able to handle partial segments / async stuff. Even if all existing ones do, there's also future ones, ones from extensions, etc.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -536,16 +897,30 @@ private AcquireSegmentAction acquireExistingSegment(SegmentCacheEntryIdentifier
             location.addWeakReservationHoldIfExists(identifier)
         );
         if (hold != null) {
-          if (hold.getEntry().isMounted()) {
+          if (!(hold.getEntry() instanceof CompleteSegmentCacheEntry complete)) {
+            // The eager (complete) acquire path found a non-complete entry under this id. This only arises if
+            // partial-load on-disk state survived a toggle of druid.segmentCache.virtualStoragePartialDownloadsEnabled
+            // to false (getCachedSegments reserves an on-disk partial layout regardless of the flag). The eager path
+            // cannot serve a partial layout; surface a clear operator error rather than a ClassCastException.
+            throw DruidException.forPersona(DruidException.Persona.OPERATOR)
```
gianm, line 882: Two concerns:

1. Won't this also be an issue on downgrade?
2. Is it possible to repair this rather than throwing an error? As is, it will be difficult to turn off partial downloads once they are enabled.

processing/src/main/java/org/apache/druid/segment/PartialBundleAcquirer.java
```diff
@@ -0,0 +1,48 @@
...
+ *   <li>The executor on which partial downloads should run.</li>
+ * </ol>
+ */
+public interface PartialBundleAcquirer
+{
+  /**
+   * Acquire a hold on the cache-layer bundle with the given name. The returned {@link Closeable} releases the hold
+   * when closed (the bundle itself stays in the cache for re-use by subsequent acquires).
+   * <p>
+   * Implementations should be safe to call concurrently for different bundle names.
```
gianm: "should be" or "must be"? Does the wording imply it's *not* safe to call concurrently with the same bundle name?

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
```diff
@@ -0,0 +1,279 @@
...
+  {
+    final QueryableProjection<QueryableIndex> matched = index.getProjection(spec);
+    final QueryableIndex rowSelector = matched != null ? matched.getRowSelector() : index;
+    final Set<String> requiredColumns = requiredColumns(rowSelector, matched, spec);
+    final String bundleName = matched != null ? matched.getName() : Projections.BASE_TABLE_PROJECTION_NAME;
+
+    // Mount the cache-layer bundle BEFORE submitting downloads. This sparse-allocates the bundle's container files
+    // and reserves the disk-usage accounting at bundle granularity. The returned Closeable releases the hold when
+    // the cursor closes, allowing the cache to later evict the bundle if needed.
+    // PartialBundleAcquirer.acquire() throws DruidException on failure.
```
gianm: This comment is too long

processing/src/main/java/org/apache/druid/segment/PartialBundleAcquirer.java
```diff
@@ -0,0 +1,48 @@
...
+ * Context object that bundles together pieces of cache-layer machinery needed to do on-demand partial segment loading:
+ * <ol>
+ *   <li>A hook for acquiring a hold on the cache-layer bundle.</li>
+ *   <li>The executor on which partial downloads should run.</li>
+ * </ol>
+ */
+public interface PartialBundleAcquirer
+{
+  /**
+   * Acquire a hold on the cache-layer bundle with the given name. The returned {@link Closeable} releases the hold
```
gianm, line 37: How are callers supposed to know what the bundle names are?

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
```diff
@@ -0,0 +1,279 @@
...
+      PartialBundleAcquirer bundleAcquirer
+  )
+  {
+    this.index = index;
+    this.delegate = new QueryableIndexCursorFactory(index, timeBoundaryInspector);
+    this.bundleAcquirer = bundleAcquirer;
+  }
+
+  @Override
+  public CursorHolder makeCursorHolder(CursorBuildSpec spec)
```
gianm, line 89: Given this is going to blow up, the default impl for `makeCursorHolderAsync` (which calls `makeCursorHolder`) isn't very useful, and seems likely to lead to delegating `CursorFactory` being implemented improperly. Consider deleting it, and perhaps replacing it with an abstract base class that can be used for leaf `CursorFactory` that don't have special logic for async cursor creation.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
```diff
@@ -0,0 +1,279 @@
...
+    return delegate.makeCursorHolder(spec);
+  }
+
+  @Override
+  public AsyncCursorHolder makeCursorHolderAsync(CursorBuildSpec spec)
+  {
+    final QueryableProjection<QueryableIndex> matched = index.getProjection(spec);
+    final QueryableIndex rowSelector = matched != null ? matched.getRowSelector() : index;
+    final Set<String> requiredColumns = requiredColumns(rowSelector, matched, spec);
+    final String bundleName = matched != null ? matched.getName() : Projections.BASE_TABLE_PROJECTION_NAME;
```
gianm, line 107: How did we know that the bundle name should be the projection name? Is that specified somewhere or just how things are currently implemented?

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
```diff
@@ -0,0 +1,279 @@
...
+        }));
+      }
+
+      // No canceler: downloads are allowed to complete even if the awaiter cancels. Once they complete and the inner
+      // CursorHolder is handed to asyncHolder.set(), it's owned by the asyncHolder until the awaiter takes one of
+      // two paths: asyncHolder.close() (cancel, closes the held holder) or asyncHolder.release() (success, transfers
+      // ownership to the caller). Either way the holder is drained and nothing leaks.
+      final AsyncCursorHolder asyncHolder = new AsyncCursorHolder(null);
+      Futures.addCallback(
+          Futures.allAsList(columnDownloads),
```
gianm: Seems like a good use case for `AsyncResource.collect(columnDownloads)` and `AsyncResource.transform` (from #19539)

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
```diff
@@ -0,0 +1,279 @@
...
+      final ListeningExecutorService downloadExec = bundleAcquirer.getDownloadExec();
+      final List<ListenableFuture<?>> columnDownloads = new ArrayList<>(requiredColumns.size());
+      for (String column : requiredColumns) {
+        columnDownloads.add(downloadExec.submit(() -> {
+          rowSelector.getColumnHolder(column);
+          return null;
+        }));
+      }
+
+      // No canceler: downloads are allowed to complete even if the awaiter cancels. Once they complete and the inner
```
gianm: why not cancel them? Is it because canceling them is awkward for some reason, or you feel it's better to not cancel them? I suppose there are pros and cons. The main pro I can think of (for canceling) is that no other query may need the files, so we're wasting time fetching them. The main con is that some other query _does_ need the files, so we're going to need to re-start the fetch if we cancel it.

Maybe the best behavior is to cancel it unless some other query is also waiting for the same thing.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
```diff
@@ -0,0 +1,279 @@
...
+      // index whose data extends past the query interval. That read happens after the cursor holder is handed back,
+      // so __time must be pre-fetched on the async path or it becomes a lazy deep-storage download on a processing
+      // thread. Predicting exactly when the holder reads __time would mean replicating its internals (fragile, and it
+      // reads __time in the common cases anyway), so always include it: __time is cheap, and it resolves to a
+      // no-download constant column for projections that don't carry a real time column.
+      required.add(ColumnHolder.TIME_COLUMN_NAME);
+      return required;
+    }
+    // Conservative fallback when physicalColumns isn't declared, fetch every column on the chosen row selector
+    // plus __time (which is special-cased and not enumerated by getColumnNames()).
```
gianm, line 220: Seems weird. Can you consider changing it in this PR, so `getColumnNames()` does include `__time`? Looking around at the usages of `getColumnNames()`, it doesn't seem like anything is super-dependent on them not including `__time`.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
```diff
@@ -0,0 +1,279 @@
...
+      // callback yet, so release it here. Already-submitted download tasks will complete with no callback wired,
+      // their captured row-selector refs drop naturally.
+      throw CloseableUtils.closeAndWrapInCatch(t, bundleHold);
+    }
+  }
+
+  /**
+   * Determine the set of physical column names to required from the chosen row selector given a {@link CursorBuildSpec}
+   */
+  private static Set<String> requiredColumns(
```
gianm, line 199: nit: the `private static` is typically placed after all non-`static` methods

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexSegment.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.Closeable;
+
+/**
+ * {@link Segment} wrapper around a {@link PartialQueryableIndex}. Mirrors {@link QueryableIndexSegment} but wires up
+ * the V10-specific {@link V10TimeBoundaryInspector} (which answers from
+ * {@link org.apache.druid.segment.projections.ProjectionMetadata} min/max fields without downloading the
+ * {@code __time} column) and the partial-aware {@link PartialQueryableIndexCursorFactory} (which downloads
+ * required files on the supplied download executor before handing back a cursor).
+ * <p>
+ * Lifecycle: this segment is intended to exist as a transient reference-hold scope over an externally-owned
```
gianm, line 37: What typically owns the `PartialQueryableIndex`?

server/src/main/java/org/apache/druid/segment/loading/StorageLocation.java
```diff
@@ -515,6 +515,45 @@ public void release(CacheEntry entry)
     }
   }
 
+  /**
+   * Remove a {@link #weakCacheEntries} entry that currently has no outstanding holds, unlinking it from the SIEVE
+   * queue and terminating its phaser (which fires the underlying {@link CacheEntry#unmount}). No-op when the entry
+   * is absent, is a {@link #staticCacheEntries} entry, or still has outstanding holds.
+   * <p>
+   * This exists for callers that register a weak entry <em>without</em> a {@link ReservationHold} (the bootstrap
+   * reserve path uses {@link #reserveWeak}) and need to clean it up after a failed mount. The normal runtime path
```
gianm, line 528: I think this will be useful in #19539: lingering entries on failed mount was a problem raised in review (https://github.com/apache/druid/pull/19539#discussion_r3341105641)

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -526,6 +548,345 @@ public AcquireSegmentAction acquireSegment(final DataSegment dataSegment)
...
+   * <p>
+   * Slow path: under the per-segment lock, {@link #findOrReservePartial} reuses an existing not-yet-usable entry or
+   * reserves a fresh weak one, and the action submits mount (+ optional ensureAllDownloaded) to
+   * {@link #virtualStorageLoadOnDemandExec} so callers that yield on the future never block a processing thread on
+   * deep-storage I/O.
+   */
+  private AcquireSegmentAction acquirePartialInternal(
+      DataSegment dataSegment,
+      SegmentRangeReader rangeReader,
+      boolean fullDownload
```
gianm, line 579: On the `fullDownload` path I don't see stuff that is creating a cache entry and hold on the `StorageLocation` to account for the storage space. The logic seems to be fetching things outside of a hold. Is that right or did I miss it?

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
```diff
@@ -0,0 +1,303 @@
...
+      // stops column downloads that haven't begun their deep-storage read yet: queued tasks are skipped, and tasks
+      // parked on the download executor's permit are interrupted out of the (interruptible) wait before doing any I/O.
+      // A download already in its read/write loop runs to completion. After canceling, release the bundle hold. Once
+      // the holder is produced and handed to set(), ownership transfers to the awaiter, which drains it via
+      // close() (cancel) or release() (success); the once-guard keeps the hold release safe across all of these paths.
+      final AsyncCursorHolder asyncHolder = new AsyncCursorHolder(() -> {
+        for (ListenableFuture<?> columnDownload : columnDownloads) {
+          columnDownload.cancel(true);
+        }
+        CloseableUtils.closeAndSuppressExceptions(releaseHoldOnce, ignored -> {});
```
FrankChen021: [P1] Keep bundle holds until running downloads finish

When an async cursor is closed after a column download has already entered rowSelector.getColumnHolder/mapFile, cancel(true) does not wait for that callable to stop; the code even documents that in-flight read/write loops may run to completion. The canceler still releases the bundle hold immediately, which drops the bundle reference and storage hold while mapper work may still be using the container. That allows SIEVE/ephemeral eviction or a pending unmount to reach PartialSegmentFileMapperV10.evictContainer despite mapFile/ensureFilesAvailable still being in flight, violating that mapper's concurrency contract and risking deletion/unmap of a container being read or written. Defer releasing the bundle hold/reference until submitted downloads have actually returned, or add per-download protection so eviction cannot run while a started download continues.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Reviewed 62 of 62 changed files.

- server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java:119 - [P2] Stop factory-owned loading pools. After moving the on-demand executor into StorageLoadingThreadPool, this shutdown path no longer stops virtualStorageLoadingThreadPool. That is OK for the lifecycle-managed singleton from StorageNodeModule, but SegmentCacheManagerFactory.manufacturate creates a StorageLoadingThreadPool manually for task/MSQ segment caches and callers only shut down the SegmentCacheManager/SegmentManager. Those per-task virtual-storage managers now leak their loading executor after task completion; with fixed platform threads this can keep task JVMs alive, and repeated tasks can accumulate idle pools. Please either have SegmentLocalCacheManager.shutdown stop factory-owned pools or arrange for the factory-created pool to be lifecycle-owned and stopped.

---

processing/src/main/java/org/apache/druid/segment/V10TimeBoundaryInspector.java
```diff
@@ -0,0 +1,82 @@
...
+ * The exact-bounds path is independent of segment sort order: the writer records the actual minimum and maximum
+ * timestamps across all walked rows even when the segment is not time-sorted. This is a strict improvement over
+ * the eager-load {@link QueryableIndexTimeBoundaryInspector}, which reports inexact bounds for non-time-sorted
+ * segments because its only option is to read positions 0 and N-1 of the {@code __time} column.
+ */
+public final class V10TimeBoundaryInspector implements TimeBoundaryInspector
+{
+  /**
+   * Build an inspector that reads from the given base projection metadata, falling back to {@code fallbackInterval}
+   * when the writer didn't persist min/max times (e.g. pre-Phase-1.5 segments).
```
capistrant: "pre-Phase-1.5" callout feels unnecessary with no added context

processing/src/test/java/org/apache/druid/segment/PartialQueryableIndexCursorFactoryTest.java
```diff
@@ -0,0 +1,806 @@
...
+      // cursor over the on-demand-loaded columns. Would throw here if the wrapper used the sync default.
+      try (AsyncCursorHolder asyncHolder = filtered.makeCursorHolderAsync(CursorBuildSpec.FULL_SCAN);
+           CursorHolder holder = asyncHolder.release()) {
+        Assertions.assertNotNull(holder);
+        Assertions.assertNotNull(holder.asCursor(), "async path should build a usable cursor without a sync download");
+      }
+    }
+  }
+
+  private File buildTimeOrderedProjectionSegment(String projectionName)
```
capistrant: nit: private helper method interleaved within tests

processing/src/test/java/org/apache/druid/segment/PartialQueryableIndexCursorFactoryTest.java
```diff
@@ -0,0 +1,806 @@
...
+  }
+
+  @Test
+  void testAsyncCloseBeforeDownloadCompletesClosesCursorHolderOnce() throws Exception
+  {
+    // Producer-doesn't-leak invariant: if the wrapper is closed before the producer's set(holder) fires, the producer
+    // is responsible for closing the holder. Use the real (single-threaded) executor with a gate that pauses the
+    // download task before it builds the cursor, then close the wrapper, then release the gate.
+    final CountingRangeReader rangeReader = new CountingRangeReader(segmentDir);
+    final AtomicReference<RuntimeException> bgErr = new AtomicReference<>();
```
capistrant: bgErr doesn't actually seem to be wired into the test at all. it is just created and used in get assertion below.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
capistrant: I am also interested in these ^

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
```diff
@@ -0,0 +1,360 @@
...
+          // holder, its inner, and the bundle hold don't leak.
+          holder.close();
+        }
+      });
+      return asyncHolder;
+    }
+    catch (Throwable t) {
+      // Failure between acquire and wiring up the downloads (submitDownload shut-down rejection, etc.). Ownership of
+      // the bundle hold hasn't transferred to the holder yet, so release it here.
+      throw CloseableUtils.closeAndWrapInCatch(t, holdRelease.releaser());
```
capistrant: why do we use `releaser()` here, which is documented as a "success-path" releaser, instead of requesting a release? also, do we need to worry about what is in `columnDownloads` at all at this point?

capistrant (commented):
nothing overly concerning jumping out at me so far. I am still processing some of the larger app file changes. but first glance didn't catch any glaring blockers

FrankChen021 (commented):

Reviewed 64 of 64 changed files.

---

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexSegment.java
clintropolis: #19577 split out `RowCountInspector` which is now wired up in this PR so we can answer the row count stuff which is much more widely used than the other parts of this interface (and answerable purely by metadata with v10 segments)

server/src/main/java/org/apache/druid/server/SegmentManager.java
clintropolis: the methods are now combined and accept an `AcquireMode` enum argument, i think it is a bit clearer what happens

server/src/main/java/org/apache/druid/server/SegmentManager.java
clintropolis: same thing with the enum `AcquireMode` argument, i think it should be more obvious now

multi-stage-query/src/main/java/org/apache/druid/msq/input/RegularLoadableSegment.java
clintropolis: this now accepts the `AcquireMode` enum argument to pass through, and the engine decides what mode it wants when creating the `ReadableInputQueue` (see `BaseLeafStageProcessor.makeReadableInputQueue`)

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
clintropolis: agree this was a wack failure mode, I have reworked stuff so that if we detect partials on disk during startup and partials are not enabled we now delete them instead of exploding at query time. I have also switched the config to disable this partial load functionality until we have some more time to mature things.

With regards to rollback to older versions, what happens is that on startup mount will fail with an exception because the path will not contain expected v9 or v10 segment files, and the handling on that exception will trigger deletion because the path exists _unless_ `druid.segmentCache.deleteOnRemove` is false (it is true by default). We could do something kind of tricky to ensure that they are always deleted, which is _always_ write the `DOWNLOAD_START_MARKER_FILE_NAME` for partial segments so it would trigger the 'corrupted segment' handling if we tried to mount as a regular segment, but I have not done that yet in this PR because it feels kind of strange.

processing/src/main/java/org/apache/druid/segment/PartialBundleAcquirer.java
clintropolis: clarified

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
clintropolis: I have left this as is for now because this path gets actually used when a partial segment is used in 'full load' mode (e.g. by the native engine instead of Dart/MSQ). These guards are mainly to ensure that the caller is using the correct `AcquireMode` to hold the segment during processing.

The default implementation is probably not very useful though and should maybe be an exception, I'll look into changing this (either in this PR or in a follow-up since I still need to wire up unnest and join cursor factories to have async implementations so could do it then too).

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
clintropolis: this is kind of fragile and magic I agree, I'll look into tightening this up, though maybe as a follow-up. Currently this is just baked into how v10 segments are written, we start a bundle every time we write a projection, but it would be nicer to like look this up in some way or push it into the filemapper or.. something, idk still thinking on it.

processing/src/main/java/org/apache/druid/segment/PartialBundleAcquirer.java
clintropolis: see https://github.com/apache/druid/pull/19535#discussion_r3415467548

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
clintropolis: switched this to use `collect`, not `transform` though yet since still trying to figure out how to make the transformed thing a closeable, safely. I think we would need a new `transformCloseable` or .. something to handle this case since the current transform only closes the source, not the output.

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
clintropolis: added cancel wiring, i didn't do it at first since was a bit complicated for book-keeping, but added a thing to handle it so it should be good now (the _actual_ download parts in the mapper still don't interrupt so an in flight pull will not stop since there is even more complicated cleanup that would need to be done to do that), but the at least all of the parts leading up to that can bail out now.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
clintropolis: yea, this was an oversight in a refactor, good catch; it should be handled correctly now

processing/src/test/java/org/apache/druid/segment/PartialQueryableIndexCursorFactoryTest.java
clintropolis: oops, this was stale test after refactoring some stuff, it no longer had a point, deleted

processing/src/main/java/org/apache/druid/segment/PartialQueryableIndexCursorFactory.java
clintropolis: this was a mistake actually, good catch

server/src/main/java/org/apache/druid/segment/loading/SegmentLoaderConfig.java
```diff
@@ -105,6 +105,15 @@ public class SegmentLoaderConfig
...
   private long virtualStorageMetadataReservationEstimate = 16L * 1024L * 1024L;
 
+  /**
+   * When true, partial-eligible V10 segments are mounted via the partial machinery and
+   * {@link SegmentCacheManager#acquireSegment} with {@link AcquireMode#PARTIAL} returns a metadata-anchored segment
+   * whose columns are downloaded on demand. When false (the default), {@link AcquireMode#PARTIAL} falls back to
+   * {@link AcquireMode#FULL} so the entire segment is downloaded up front (matching pre-partial-download behavior).
+   */
+  @JsonProperty("virtualStoragePartialDownloadsEnabled")
+  private boolean virtualStoragePartialDownloadsEnabled = false;
```
gianm, line 115: IMO it would be better to have this always enabled at the VSF layer, and to have the toggle be in the query context. There can be something like `QueryContexts#getAcquireMode`. (Something like `usePartialDownloads: true/false`). I like this approach because it makes testing much easier, and also it makes the code in the VSF layer easier to follow because it gets rid of conditionals. There would also be no reason to get rid of partially-downloaded directories during bootstrapping, because they're always considered OK.

Default should be `usePartialDownloads: false` for now.

multi-stage-query/src/main/java/org/apache/druid/msq/querykit/ReadableInputQueue.java
```diff
@@ -288,7 +298,7 @@ private ListenableFuture<ReadableInput> loadNextSegment()
         return null;
       }
 
-      final AcquireSegmentAction acquireSegmentAction = nextLoadableSegment.acquire();
+      final AcquireSegmentAction acquireSegmentAction = nextLoadableSegment.acquire(acquireMode);
```
gianm, line 301: It's been clear in testing that good overall performance on cold data is highly dependent on loadahead. With this change, we aren't fetching the entire segment as part of loadahead, which is going to lead to poorer download rates. It's ok to deal with this in a follow-up, but it does need to be dealt with somehow.

Either the query logic should declare what parts of the segment it wants to start with, or we should do something statistical: for each query, observe which parts of the segment are actually used, and fetch those during loadahead for future segments. The statistical approach only works for queries that fetch a large number of segments (so the earlier ones can be used to inform the later ones). But this is fine, since it's those queries where the performance concern lies.

gianm (approved):
Approved with two notes that could be done in this PR or could be done as follow-ups. I do think moving the partial download setting to context is a good idea.

PR #19537, opened 2026-06-01
feat: add support for specifying queryable historical tiers via query context

jtuglu1:
### Description

Support specifying specific historical tiers to be queried dynamically via query context. Tiers may either be physical or virtual (mapped to physical tiers via coordinator dynamic config).

#### Release note

Support specifying specific historical tiers to be queried dynamically via query context.

<hr>

This PR has:

- [ ] been self-reviewed.


FrankChen021 (commented):

---

FrankChen021 (commented):

---

FrankChen021 (commented):

---

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
```diff
@@ -604,10 +611,15 @@ private SortedMap<DruidServer, List<SegmentDescriptor>> groupSegmentsByServer(
         CloneQueryMode cloneQueryMode
     )
     {
+      final Set<String> queryableHistoricalTiers = query.context().getQueryableHistoricalTiers();
```
abhishekrb19: Similar to `cloneQueryMode`, it'll be good to get the historical tiers once and pass it to both `computeResultLevelCachingEtag()` and `pick()` to avoid getting the tiers from parsing the context unnecessarily

server/src/main/java/org/apache/druid/client/selector/ServerSelector.java
```diff
@@ -190,16 +191,123 @@ public List<DruidServerMetadata> getAllServers(CloneQueryMode cloneQueryMode)
...
+  {
+    if (queryableHistoricalTiers == null) {
+      return queryableServers;
+    }
+
+    final Int2ObjectRBTreeMap<Set<QueryableDruidServer>> filteredServers =
+        new Int2ObjectRBTreeMap<>(historicalTierStrategy.getComparator());
+    for (final int priority : queryableServers.keySet()) {
+      final Set<QueryableDruidServer> priorityServers = new HashSet<>();
+      for (final QueryableDruidServer server : queryableServers.get(priority)) {
```
abhishekrb19: Since this is a tree map, you could avoid the double lookup by iterating on the `queryableServers.int2ObjectEntrySet()` once as you go instead of getting the priority and then looking it up again with it

Would likely require a small restructure here.

abhishekrb19 (approved):
LGTM. Left some comments - this is on the query path, so it'll be good to remove cleanup any redundant ops - thanks.

server/src/main/java/org/apache/druid/client/selector/ServerSelector.java
```diff
@@ -190,16 +192,123 @@ public List<DruidServerMetadata> getAllServers(CloneQueryMode cloneQueryMode)
...
+    if (queryableHistoricalTiers == null) {
+      return queryableServers;
+    }
+
+    final Int2ObjectRBTreeMap<Set<QueryableDruidServer>> filteredServers =
+        new Int2ObjectRBTreeMap<>(historicalTierStrategy.getComparator());
+    for (final Int2ObjectMap.Entry<Set<QueryableDruidServer>> entry : queryableServers.int2ObjectEntrySet()) {
+      final Set<QueryableDruidServer> priorityServers = new HashSet<>();
+      for (final QueryableDruidServer server : entry.getValue()) {
+        if (queryableHistoricalTiers.contains(server.getServer().getTier())) {
```
FrankChen021, line 274: [P1] Resolve virtual tier aliases before filtering servers

This compares requested tiers directly with each server's physical tier, but the PR description says `queryableHistoricalTiers` supports virtual tiers from `historicalTierAliases`. Alias keys are never physical tiers: Coordinator expansion loads segments only onto their mapped tiers. Therefore, with `historicalTierAliases={"hot":["hot_1","hot_2"]}`, querying `queryableHistoricalTiers=["hot"]` matches no server and silently returns empty or partial results. The Broker's coordinator-config view should expand aliases before `hasQueryableHistoricalServer` and `getQueryableHistoricalServers` perform these checks, with coverage for virtual and mixed virtual/physical tier sets.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Found one correctness issue: virtual tier aliases are compared directly with physical server tiers and can silently select no servers.

---

docs/querying/query-context-reference.md
```diff
@@ -68,9 +68,10 @@ Unless otherwise noted, the following parameters apply to all query types, and t
...
 |`cloneQueryMode`|`excludeClones`| Indicates whether clone Historicals should be queried by brokers. Clone servers are created by the `cloneServers` Coordinator dynamic configuration. Possible values are `excludeClones`, `includeClones` and `preferClones`. `excludeClones` means that clone Historicals are not queried by the broker. `preferClones` indicates that when given a choice between the clone Historical and the original Historical which is being cloned, the broker chooses the clones. Historicals which are not involved in the cloning process will still be queried. `includeClones` means that broker queries any Historical without regarding clone status. This parameter only affects native queries. MSQ does not query Historicals directly.|
+|`queryableHistoricalTiers`|`null`|Set of Historical tier names that may be queried. When set, the Broker only queries Historical servers whose `druid.server.tier` is in this set. Segments without a replica on one of the listed tiers are skipped.|
```
kfaraz, line 74: Does this also need to mention that tier aliases may be used instead of actual tier names?

server/src/main/java/org/apache/druid/client/selector/ServerSelector.java
```diff
@@ -190,16 +192,123 @@ public List<DruidServerMetadata> getAllServers(CloneQueryMode cloneQueryMode)
...
       return realtimeTierStrategy.pick(query, realtimeServers, segment.get());
     }
   }
 
+  public <T> boolean hasQueryableServer(@Nullable final Query<T> query, final CloneQueryMode cloneQueryMode)
+  {
+    return hasQueryableServer(getQueryableHistoricalTiers(query), cloneQueryMode);
+  }
+
+  public boolean hasQueryableServer(
```
kfaraz, line 228: Please add a short javadoc.

server/src/main/java/org/apache/druid/client/selector/ServerSelector.java
```diff
@@ -190,16 +192,123 @@ public List<DruidServerMetadata> getAllServers(CloneQueryMode cloneQueryMode)
...
+  public <T> QueryableDruidServer pick(
+      @Nullable final Query<T> query,
+      final CloneQueryMode cloneQueryMode,
+      @Nullable final Set<String> queryableHistoricalTiers
+  )
   {
     synchronized (this) {
-      if (!historicalServers.isEmpty()) {
-        return historicalTierStrategy.pick(query, filter.getQueryableServers(historicalServers, cloneQueryMode), segment.get());
+      if (queryableHistoricalTiers != null && queryableHistoricalTiers.isEmpty()) {
```
kfaraz, line 208: This check can be done before the `synchronized` block.

server/src/main/java/org/apache/druid/client/selector/ServerSelector.java
```diff
@@ -190,16 +192,123 @@ public List<DruidServerMetadata> getAllServers(CloneQueryMode cloneQueryMode)
...
+        return null;
+      }
+      if (!historicalServers.isEmpty() || queryableHistoricalTiers != null) {
+        final Int2ObjectRBTreeMap<Set<QueryableDruidServer>> queryableHistoricalServers =
+            getQueryableHistoricalServers(
+                filter.getQueryableServers(historicalServers, cloneQueryMode),
+                queryableHistoricalTiers
```
kfaraz, line 215: It would probably make more sense to pass the list of queryable tiers into the `filter.getQueryableServers` since the function seems very closely related.
This would allow us to keep the server filtration logic in a single place, and will also provide an easy way for tier alias expansion in `BrokerViewOfCoordinatorConfig`, which is an impl of the `HistoricalFilter`.

server/src/main/java/org/apache/druid/client/selector/ServerSelector.java
```diff
@@ -190,16 +192,123 @@ public List<DruidServerMetadata> getAllServers(CloneQueryMode cloneQueryMode)
...
+    return hasQueryableServer(getQueryableHistoricalTiers(query), cloneQueryMode);
+  }
+
+  public boolean hasQueryableServer(
+      @Nullable final Set<String> queryableHistoricalTiers,
+      final CloneQueryMode cloneQueryMode
+  )
+  {
+    synchronized (this) {
+      if (queryableHistoricalTiers != null) {
```
kfaraz, line 234: This check can happen outside the `synchronized` block.

server/src/main/java/org/apache/druid/client/selector/ServerSelector.java
kfaraz: @jtuglu1 , this seems valid. It doesn't seem that we are doing tier alias expansion in the current patch.
If not, we can do it in a follow up and update the PR description/release note.

kfaraz:
As discussed with @jtuglu1 offline, this PR needs some more changes. We are leaving it out of Druid 38 for now.

server/src/main/java/org/apache/druid/client/selector/ServerSelector.java
FrankChen021: A follow-up is reasonable if this PR is intentionally narrowed to physical tier names only. Please update the description and release note before merge, since they still promise virtual tiers; the added query-context documentation already describes physical `druid.server.tier` matching. Reviewed 9 of 9 changed files.

PR #19538, opened 2026-06-01
feat: Wait for gracefulShutdownTimeout in DartWorkerRunner#stop.

gianm:
This patch adjusts DartWorkerRunner to wait for gracefulShutdownTimeout before canceling workers.

The default gracefulShutdownTimeout is changed from 0s to 30s to match the documentation, and also, it is believed, to match actual behavior.

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartWorkerRunner.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartWorkerRunner.java
@@ -44,6 +44,7 @@ import org.apache.druid.msq.exec.WorkerRunRef;
 import org.apache.druid.msq.rpc.ResourcePermissionMapper;
 import org.apache.druid.msq.rpc.WorkerResource;
 import org.apache.druid.query.QueryContext;
+import org.apache.druid.server.initialization.ServerConfig;
 import org.apache.druid.server.security.AuthorizerMapper;
 import org.apache.druid.utils.CloseableUtils;
 import org.joda.time.DateTime;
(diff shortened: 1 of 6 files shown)


multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartWorkerRunner.java
```diff
@@ -250,15 +275,47 @@ public void start()
...
+              "Canceling work for query[%s] due to timeout during stop (waited [%,d] ms)",
+              holder.workerContext.queryId(),
+              clock.millis() - waitStart.getMillis()
+          );
+
+          holder.runRef.cancel();
+          holder.runRef.awaitStop();
+        }
+      }
+      catch (InterruptedException e) {
```
FrankChen021, line 316: [P2] Cancel remaining workers when shutdown wait is interrupted

If the lifecycle stop thread is interrupted while waiting for a worker, awaitStop now throws and this catch exits without canceling the current or remaining runningWorkers. Since stopped is already true and workerMap has been cleared, a later stop() call returns immediately, leaving background worker threads running instead of forcing cancellation. The interruption path should cancel the copied workers before returning or rethrowing.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

PR #19539, opened 2026-06-01
feat: Background fetching of files for SQL EXTERN.

gianm:
This patch adds the setting `backgroundFetchExternalFiles`. When set, cloud storage files referenced by `ExternalSegment` (`EXTERN`) are fetched asynchronously into the task's storage locations. The setting defaults to true.

To support this, new infrastructure is added:

1) `VirtualStorageManager`, a layer on top of `StorageLocation` that provides a simple "fetch and cache a file" API. The main implementation is `StorageLocationVirtualStorageManager`.

2) `StorageLoadingThreadPool`, an extraction of the thread pool from `SegmentLocalCacheManager` so it can be shared with `VirtualStorageManager`.

3) `AsyncResource`, a Future-like utility that provides better tools for managing Closeable resources. It is used by `VirtualStorageManager` to provide the asynchronously-fetched file handles.

It is hoped that the infrastructure will be more broadly useful for other use cases.


jtuglu1:
Can this async behavior be extended to something like fetching an Iceberg table's schema (so you don't need to explicitly dictate it)?

gianm:
> Can this async behavior be extended to something like fetching an Iceberg table's schema (so you don't need to explicitly dictate it)?

I don't have a direct answer since I'm not super familiar with how the Iceberg loading works. But a roundabout response: in theory the changes to `AdaptedLoadableSegment` should make it possible to asynchronously load anything that might be required to create a usable `Segment`.

server/src/main/java/org/apache/druid/segment/loading/external/StorageLocationVirtualStorageManager.java
```diff
@@ -0,0 +1,308 @@
...
+              {
+                if (mounted.get()) {
+                  populationLocks.remove(identifier, lock);
+                }
+                super.unmount();
+              }
+            };
+
+            // Try to reserve weak space
+            if (!location.reserveWeak(entry)) {
```
FrankChen021: [P1] Hold weak entries before populating them

This calls reserveWeak(entry) and then populates the file before acquiring any ReservationHold. reserveWeak creates an unheld weak cache entry, so another concurrent reservation under storage pressure can reclaim/unmount it while the populator is still writing. If unmount runs before mountedFile is set, it also will not delete the just-written file, leaving untracked bytes or causing the first fetch to fail/fallback despite having reserved space. Please create the weak entry with a hold, or acquire a hold before population, and release it on failure.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Reviewed 93 of 93 changed files.

---

server/src/main/java/org/apache/druid/segment/loading/external/StorageLocationVirtualStorageManager.java
gianm: #19539 adds `StorageLocation#removeUnheldWeakEntry` which I believe can be used to solve this problem. If that PR is merged first I can update this to use it. If not then the update can be done in a follow-up PR.

processing/src/main/java/org/apache/druid/common/asyncresource/SettableAsyncResource.java
```diff
@@ -0,0 +1,326 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+
+/**
+ * Basic utility that allows creating {@link AsyncResource} wrappers. Analogous to JDK {@link CompletableFuture}
+ * or Guava {@link SettableFuture}. See {@link AsyncResource} for details about why you would want to use this
+ * instead of {@link Future}.
+ *
+ * <p>In addition to {@link #get()}, there is also {@link #release()}. Releasing is not allowed by instances
+ * of this class, but may be allowed by subclasses.
```
clintropolis: it feels worth elaborating a bit more about what the use-case is for release, which is about ownership transfer, mainly useful when the T is closeable and can be constructed to carry the closeable from set into the T that is released.

processing/src/main/java/org/apache/druid/common/asyncresource/SettableAsyncResource.java
```diff
@@ -0,0 +1,326 @@
...
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+
+/**
+ * Basic utility that allows creating {@link AsyncResource} wrappers. Analogous to JDK {@link CompletableFuture}
+ * or Guava {@link SettableFuture}. See {@link AsyncResource} for details about why you would want to use this
+ * instead of {@link Future}.
+ *
+ * <p>In addition to {@link #get()}, there is also {@link #release()}. Releasing is not allowed by instances
```
clintropolis, line 44: it feels worth perhaps elaborating on the use model for release, either here to describe it a bit to someone reading the main javadocs or i guess the method would be fine too;

I would describe it as a resource ownership transfer model mainly useful for callers whose `T` itself is `Closeable`, and whose producer that calls `set` constructs the `T` such that its close method will release all associated resources, so that it is self-contained. Or something like that.

server/src/main/java/org/apache/druid/segment/loading/external/DownloadableCacheEntry.java
```diff
@@ -0,0 +1,301 @@
...
+    return (T) extendMap.computeIfAbsent(clazz, k -> supplier.get());
+  }
+
+  private File getFileForIdentifier(File locationPath, String identifier)
+  {
+    String sanitized = sanitizePath(identifier);
+    return new File(locationPath, sanitized);
+  }
+
+  static String sanitizePath(final String originalPath)
```
clintropolis, line 209: this method feels worth javadocs since its got a lot going on, and looking at the tests seems rather unobvious what the behavior is

server/src/main/java/org/apache/druid/segment/loading/external/StorageLocationVirtualStorageManager.java
clintropolis: couldn't this call location.addWeakReservationHold which either returns a hold on an existing entry, or creates a new new entry under a hold if it doesn't exist, then this block and the later addWeakReservationHoldIfExists would be collapsed into the single call, and we only mount if we have a hold on the entry.

> #19539 adds StorageLocation#removeUnheldWeakEntry which I believe can be used to solve this problem. 

`StorageLocation#removeUnheldWeakEntry` is more for being able to remove the unmounted entry if mount fails so there isn't an unmounted entry left hanging out. In this case the comment is talking about, the entry was removed from the location by some other thread that inserted its own different entry and had to do an eviction; this entry is eligible since there isn't a hold on it. Later after mounting, when we try to acquire a hold for the existing entry it fails and we explode, leaving  the orphaned files on disk because the entry is no longer tracked and will never be unmounted (at least this is what I think happens).

I guess we could also unmount the entry if we cannot acquire a hold on the existing thing, but that seems less ideal since it means this query has to fail even though it was originally able to reserve the location.

I think its probably unlikely to happen in practice unless the eviction 'hand' happens to be close by the entry in the linked list, but is possible and should be guarded against.

server/src/main/java/org/apache/druid/segment/loading/external/StorageLocationVirtualStorageManager.java
```diff
@@ -0,0 +1,308 @@
...
+    log.info("Initialized VirtualStorageManager with [%d] storage locations", locations.size());
+  }
+
+  @Nullable
+  @Override
+  public CachedFile get(String identifier)
+  {
+    PopulationLock lock = populationLocks.get(identifier);
+    if (lock == null) {
+      return null;
```
clintropolis, line 83: feels worth a comment that we are just using the lock to get the location, at first glance it seems like perhaps it is a mistake that nothing holds the lock, but the problem solves itself if the acquireExistingHold fails

clintropolis (commented):
:metal:

server/src/main/java/org/apache/druid/segment/loading/external/StorageLocationVirtualStorageManager.java
FrankChen021: Yes, I agree that `addWeakReservationHold(...)` is the right primitive here. It creates or returns the weak entry with a hold already registered, so the entry cannot be reclaimed while `mount`/population is still in progress. `removeUnheldWeakEntry` would only help clean up an unheld entry after a failed mount; it does not close the race where another reservation evicts this entry before the post-mount `addWeakReservationHoldIfExists` call. Holding first should also avoid the `Entry [...] disappeared after mounting` runtime-failure path and the orphaned file case.

Reviewed 11 of 93 changed files.

gianm: Ah I think I read the comment from @FrankChen021 too quickly. I was thinking of a problem that was more of the "entry leaks on failed mount" variety. There was a comment in the code about it in a `catch`:

```
// Note: Even though the entry is unmounted, space reserved by location.reserveWeak(entry) remains
// reserved (albeit unheld). It should be reclaimed eventually if there is space pressure.
entry.unmount()
```

But I wasn't sure how to fix it. I tried restructuring things to use `addWeakReservationHold` and I believe it also makes this problem go away. Could you please take another look?

clintropolis (approved):

PR #19541, opened 2026-06-01
refactor: make supervisor diff+restart algorithm smarter on submission

jtuglu1:
### Description

Restarting a supervisor spec after submission sometimes isn't always needed and can be expensive for large supervisors. The current skip restart logic is very bare-bones (doing effectively a byte-level check between the 2 specs) instead of allowing each spec to override its own restart requirements, often causing some false positive restarts. This diff can also be affected by false-positive diffs like ioConfig.taskCount being updated while auto-scaling is configured, whitespace changes, etc.

This updates the restart-required checking logic by allowing supervisors to implement their own restart requirement checks (plus some sane defaults) along with proper comparators, and stop treating each supervisor spec as an opaque ball of text. Also adds builders to keep things clean.

Notably, this doesn't change the spec persistence behavior: byte-level differing spec updates are persisted into the DB, but may not necessarily invoke a supervisor restart.

#### Release note

Make supervisor diff+restart algorithm smarter on submission. Most notably, if auto-scaling is enabled, changes to ioConfig.taskCount do not trigger a supervisor restart, although they will be persisted.

<hr>

This PR has:

- [ ] been self-reviewed.


extensions-contrib/rabbit-stream-indexing-service/src/main/java/org/apache/druid/indexing/rabbitstream/supervisor/RabbitStreamSupervisorTuningConfig.java
```diff
@@ -185,6 +186,29 @@ public Duration getOffsetFetchPeriod()
...
+  public boolean equals(Object o)
+  {
+    if (this == o) {
+      return true;
+    }
+    if (!super.equals(o)) {
+      return false;
+    }
+    RabbitStreamSupervisorTuningConfig that = (RabbitStreamSupervisorTuningConfig) o;
+    return Objects.equals(workerThreads, that.workerThreads)
```
FrankChen021, line 199: [P1] Rabbit tuning changes can skip required restarts

This equality method now drives SeekableStreamSupervisorSpec.requireRestart(), but it only compares the supervisor-level fields after super.equals(). RabbitStreamIndexTaskTuningConfig declares recordBufferSize, recordBufferOfferTimeout, and maxRecordsPerPoll, and that intermediate class does not override equals/hashCode, so changes to those tuning knobs compare equal. With skipRestartIfUnmodified=true, such a changed spec can be persisted without recreating the Rabbit supervisor, leaving the running supervisor on the old tuning. Include the Rabbit task-tuning fields in equality/hashCode, or implement equality on RabbitStreamIndexTaskTuningConfig, and add a restart-decision test.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -166,6 +166,9 @@ public Response specPost(
           }
 
           if (Boolean.TRUE.equals(skipRestartIfUnmodified) && !manager.shouldUpdateSupervisor(spec)) {
+            // No restart needed, but still persist the spec if it actually changed (e.g. a taskCount
+            // change under autoscaling) so the metadata store reflects the latest submission.
+            manager.updateSupervisorSpecWithoutRestart(spec);
```
FrankChen021: [P2] Persisted no-restart updates are not audited

This branch can now mutate metadata by calling updateSupervisorSpecWithoutRestart, but it returns before the existing auditManager.doAudit block. A taskCount-only autoscaling update with skipRestartIfUnmodified=true is therefore persisted and exposed as the latest supervisor spec without any supervisor audit entry, unlike normal spec updates. Please audit when updateSupervisorSpecWithoutRestart returns true, or move the audit logic to cover both mutating paths.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,14 +186,69 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+  {
+    Preconditions.checkState(started, "SupervisorManager not started");
+    Preconditions.checkNotNull(spec, "spec");
+    Preconditions.checkNotNull(spec.getId(), "spec.getId()");
+    Preconditions.checkNotNull(spec.getDataSources(), "spec.getDatasources()");
+
+    synchronized (lock) {
+      Preconditions.checkState(started, "SupervisorManager not started");
+      final Pair<Supervisor, SupervisorSpec> current = supervisors.get(spec.getId());
+      if (current == null || current.rhs == null || !isSpecChangedAndValidate(spec)) {
```
FrankChen021: [P2] No-restart persistence can use a stale restart decision

updateSupervisorSpecWithoutRestart re-checks only whether the spec bytes changed, not whether the current supervisor still agrees that no restart is required. SupervisorResource calls shouldUpdateSupervisor and this method as two separate synchronized operations, so a concurrent POST can change the current spec between them; this method may then persist a spec that requires a restart relative to the new current supervisor. Re-check spec.requireRestart(current.rhs) under this lock, or expose a single atomic manager method for the no-restart path.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 2 |
| P3 | 0 |
| Total | 3 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 2 |
| P3 | 0 |
| Total | 3 |

Reviewed 34 of 34 changed files.

---

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

Reviewed 36 of 36 changed files.

---

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,14 +186,101 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+      Preconditions.checkState(started, "SupervisorManager not started");
+      final Pair<Supervisor, SupervisorSpec> current = supervisors.get(spec.getId());
+      final boolean isNew = current == null || current.rhs == null;
+      final boolean specChanged = isSpecChangedAndValidate(spec);
+
+      if (skipRestartIfUnmodified && !isNew) {
+        if (!specChanged) {
+          return SpecUpdateOutcome.UNCHANGED;
+        }
+        if (!spec.requireRestart(current.rhs)) {
```
FrankChen021: [P2] Compare restart requirements after merging carried-forward fields

The skip-restart path calls requireRestart before merge. For seekable-stream resubmissions where ioConfig.taskCount is omitted, merge would carry the current taskCount forward, making the effective spec unchanged, but requireRestart sees the constructor default first and can force an unnecessary stop/recreate. Compare against a merged copy, or teach requireRestart to ignore unset taskCount.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -303,4 +371,112 @@ public abstract SeekableStreamSupervisorSpec createBackfillSpec(
       @Nullable Integer taskCount
   );
 
+  /**
+   * Returns a builder pre-populated with this spec's values (including injected services), so callers
+   * can produce a modified copy without mutating this instance. Subclasses return their own builder.
+   */
+  public abstract Builder<?> toBuilder();
```
FrankChen021, line 359: [P2] Avoid making toBuilder abstract for extension subclasses

SeekableStreamSupervisorSpec is a public base for stream supervisor specs, and this new abstract method forces downstream extension subclasses to implement a new API. Existing extension jars can fail when requireRestart calls it. Keep the default behavior conservative by providing a non-abstract fallback that reports restart-required when no builder is available.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,14 +186,101 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
 
     synchronized (lock) {
       Preconditions.checkState(started, "SupervisorManager not started");
-      final boolean shouldUpdateSpec = shouldUpdateSupervisor(spec);
+      // Persist whenever the spec actually changed (or is new) — independent of whether a restart is
```
cecemei: this method is now `createOrUpdateAndStartSupervisor(spec, false)` ? also we dont need `shouldUpdateSupervisor` anymore?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,14 +186,101 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+
+  /**
+   * Decides whether the submitted spec needs a restart and applies it under a single lock, so the decision
+   * cannot go stale between deciding and acting (which would let a concurrent POST drop a write or persist a
+   * spec that the running supervisor needs to be recreated for). With {@code skipRestartIfUnmodified} set, an
+   * unchanged spec is a no-op and a changed spec whose {@link SupervisorSpec#requireRestart} is false (e.g. a
+   * taskCount change under autoscaling) is persisted without recreating the supervisor; otherwise the
+   * supervisor is stopped and recreated (the only behavior when the flag is false).
+   */
+  public SpecUpdateOutcome createOrUpdateAndStartSupervisor(SupervisorSpec spec, boolean skipRestartIfUnmodified)
```
cecemei: nit: maybe we need another `forceRestart`? since supervisor might not restart when spec is modified but it didnt require so?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -165,29 +165,39 @@ public Response specPost(
...
-                        .build()
-          );
+          // Audit any path that mutated the persisted spec; a no-op UNCHANGED submission is not audited.
+          if (outcome != SupervisorManager.SpecUpdateOutcome.UNCHANGED) {
+            auditSupervisorUpdate(spec, req);
+          }
 
-          return Response.ok(ImmutableMap.of("id", spec.getId(), "restarted", true)).build();
+          final boolean restarted = outcome == SupervisorManager.SpecUpdateOutcome.RESTARTED;
+          return Response.ok(ImmutableMap.of("id", spec.getId(), "restarted", restarted)).build();
```
cecemei: nit: my hunch is that maybe we should have a `modified` boolean field as well since sometimes it modifies but doesnt restart?

cecemei (commented):
overall looks good to me, left a few minor comments

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
jtuglu1: `skipRestartIfUnModified=false` -> should just mean we restart irrespective of semantic differences in the spec (e.g. current behavior)?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
aho135: `shouldUpdateSupervisor` is still used in [CompactionSupervisorManager](https://github.com/apache/druid/blob/master/indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/CompactionSupervisorManager.java#L77)

Perhaps we can remove the usage there and then remove `shouldUpdateSupervisor` entirely?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
aho135: I think this makes sense. There's currently no distinction between whether a supervisor was unchanged and restarted vs changed and restarted

jtuglu1: +1 – this makes sense.

FrankChen021 (commented):

---

server/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorSpec.java
```diff
@@ -139,4 +139,23 @@ default void merge(@Nullable SupervisorSpec existingSpec)
...
+   * enabled, since that field is overridden at runtime).
+   * <p>
+   * Only consulted once the two specs are known to differ (see
+   * {@link SupervisorManager#shouldUpdateSupervisor}); the conservative default treats any
+   * such difference as requiring a restart.
+   *
+   * @param old the currently-running supervisor spec
+   * @return true if the supervisor must be restarted to apply this spec
+   */
+  default boolean requireRestart(SupervisorSpec old)
```
kfaraz: It would make more sense to invoke this method on the currently running (i.e. old) supervisor spec and pass the new one as an argument, since the question we are asking is "should we restart the current supervisor?".
This would also align this method to `validateSpecUpdateTo()`, which takes the new proposed spec as an argument.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorIOConfig.java
```diff
@@ -316,4 +317,73 @@ public boolean isBounded()
...
+  /**
+   * Returns a builder pre-populated with this config's values, for producing a modified copy without
+   * mutating this instance. Subclasses <strong>must</strong> override to return their own builder: the
+   * default builder's {@code build()} yields a generic instance whose {@code getClass()} differs from any
+   * subclass, so a non-overriding subclass would never {@link #equals} its original and {@link
+   * SeekableStreamSupervisorSpec#requireRestart} would conservatively force a restart (safe, but unnecessary).
+   */
+  public SupervisorIOConfigBuilder<?, ?> toBuilder()
+  {
+    return new SupervisorIOConfigBuilder.DefaultSupervisorIOConfigBuilder().copyFromBase(this);
```
kfaraz: If sub-classes must always override this method, we should make it abstract.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/LagAggregator.java
```diff
@@ -61,5 +61,19 @@ public <PartitionIdType> LagStats aggregate(Map<PartitionIdType, Long> partition
       final long avgLag = partitionLags.isEmpty() ? 0 : totalLag / partitionLags.size();
       return new LagStats(maxLag, totalLag, avgLag);
     }
+
+    // Stateless: all instances are equal. Needed because Jackson creates fresh instances on
```
kfaraz: Alternatively, you can make the constructor private and add a static `@JsonCreator` method which just returns the `DEFAULT` instance.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -303,4 +377,114 @@ public abstract SeekableStreamSupervisorSpec createBackfillSpec(
       @Nullable Integer taskCount
   );
 
+  /**
+   * Returns a builder pre-populated with this spec's values (including injected services), so callers
+   * can produce a modified copy without mutating this instance. Subclasses return their own builder.
+   */
+  public Builder<?> toBuilder()
+  {
+    throw new UnsupportedOperationException("No builder is available for this supervisor spec.");
```
kfaraz: Can this method be abstract instead?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -303,4 +377,114 @@ public abstract SeekableStreamSupervisorSpec createBackfillSpec(
...
+      this.suspended = spec.suspended;
+      this.taskStorage = spec.taskStorage;
+      this.taskMaster = spec.taskMaster;
+      this.indexerMetadataStorageCoordinator = spec.indexerMetadataStorageCoordinator;
+      this.indexTaskClientFactory = spec.indexTaskClientFactory;
+      this.mapper = spec.mapper;
+      this.emitter = spec.emitter;
+      this.monitorSchedulerConfig = spec.monitorSchedulerConfig;
+      this.rowIngestionMetersFactory = spec.rowIngestionMetersFactory;
+      this.supervisorStateManagerConfig = spec.supervisorStateManagerConfig;
```
kfaraz: Will the builder ever use these dependencies?
The output of the `build()` method will only ever be used for an equality check which should not involve the services anyway. So the `build()` method could simply pass these as null, atleast for now.

kfaraz (commented):
Thanks for the PR, @jtuglu1 ! Overall makes sense, left some suggestions.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -209,24 +328,18 @@ public boolean shouldUpdateSupervisor(SupervisorSpec spec)
...
       }
-      catch (JsonProcessingException ex) {
-        log.warn("Failed to write spec as bytes for spec_id[%s]", spec.getId());
+      // Identical specs never require a restart. Otherwise (after validating the update is allowed)
+      // let the spec decide whether the differences actually warrant a restart — e.g. a taskCount-only
+      // change under autoscaling does not.
+      if (!isSpecChangedAndValidate(spec)) {
+        return false;
       }
+      return spec.requireRestart(currentSupervisor.rhs);
```
kfaraz: Maybe use this for readability:

```suggestion
      return isSpecChangedAndValidate(spec) && spec.requireRestart(currentSupervisor.rhs);
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,14 +186,133 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+      createAndStartSupervisorInternal(spec, specChanged);
+      return specChanged;
     }
   }
 
+  /**
+   * Result of applying a submitted supervisor spec. {@code modified} means the persisted spec changed;
+   * {@code restarted} means the running supervisor was stopped and recreated.
+   */
+  public static final class SpecUpdateResult
```
kfaraz: Please pull this into a separate file of its own named `SupervisorSpecUpdateResult`.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,14 +186,133 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+
+  /**
+   * Decides whether the submitted spec needs a restart and applies it under a single lock, so the decision
+   * cannot go stale between deciding and acting (which would let a concurrent POST drop a write or persist a
+   * spec that the running supervisor needs to be recreated for). With {@code skipRestartIfUnmodified} set, an
+   * unchanged spec is a no-op and a changed spec whose {@link SupervisorSpec#requireRestart} is false (e.g. a
+   * taskCount change under autoscaling) is persisted without recreating the supervisor; otherwise the
+   * supervisor is stopped and recreated (the only behavior when the flag is false).
+   */
+  public SpecUpdateResult createOrUpdateAndStartSupervisor(
```
kfaraz: It seems error prone to have 2 flavors of the `createOrUpdateAndStartSupervisor` method.
Please try to merge them and pass some appropriate value for the skip restart flag if needed.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -209,24 +328,18 @@ public boolean shouldUpdateSupervisor(SupervisorSpec spec)
     Preconditions.checkNotNull(spec.getDataSources(), "spec.getDatasources()");
     synchronized (lock) {
       Preconditions.checkState(started, "SupervisorManager not started");
-      try {
-        byte[] specAsBytes = jsonMapper.writeValueAsBytes(spec);
```
kfaraz, line 218: Maybe rename this method from `shouldUpdateSupervisor` to `shouldRestartSupervisor` to avoid ambiguity.

kfaraz: Alternatively, add a new method `shouldRestartSupervisor` and leave `shouldUpdateSupervisor` unchanged.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/LagAggregator.java
```diff
@@ -48,6 +49,18 @@ public interface LagAggregator
    */
   class DefaultLagAggregator implements LagAggregator
   {
+    private DefaultLagAggregator()
```
FrankChen021, line 52: [P2] Preserve the public default constructor

`DefaultLagAggregator` is a public nested class of a public interface, so before this PR its implicit no-arg constructor was public. Making that constructor private breaks source compatibility and can also break already-compiled extensions or callers that instantiate `new LagAggregator.DefaultLagAggregator()` after upgrade. The singleton `@JsonCreator` can coexist with a public constructor, or the equality concern can be handled without removing the constructor.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Reviewed 43 of 43 changed files.

---

FrankChen021 (commented):

Reviewed 43 of 43 changed files.

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/LagAggregator.java
```diff
@@ -48,6 +49,28 @@ public interface LagAggregator
    */
   class DefaultLagAggregator implements LagAggregator
   {
+    public DefaultLagAggregator()
```
kfaraz: This should be private since we want to use this class as a singleton.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,47 +188,81 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+      // merge() may carry forward omitted fields (e.g. taskCount); compare the effective spec.
+      spec.merge(current.rhs);
+      if (!isSpecChanged(spec, current.rhs)) {
+        return SupervisorSpecUpdateResult.of(false, false);
+      }
+
+      // The effective (merged) spec is what will be persisted, so validate that transition exactly once.
+      current.rhs.validateSpecUpdateTo(spec);
+
+      if (!current.rhs.requireRestart(spec)) {
```
FrankChen021: [P2] No-restart task-count updates leave the live supervisor stale

When an autoscaling supervisor is resubmitted with only an effective taskCount change, for example after autoscaling has moved the current ioConfig.taskCount to 5 and the new POST is merged back to taskCountStart 2, SeekableStreamSupervisorSpec.requireRestart masks taskCount differences whenever either side has autoscaling enabled. This branch then persists the new spec and swaps only the manager's stored spec, but it does not restart current.lhs or update the running SeekableStreamSupervisor's ioConfig. The API can return modified=true,restarted=false while ingestion continues using the old taskCount until a later autoscaler action or Overlord restart. Treat this reset as restart-required, or explicitly propagate the new taskCount into the running supervisor before taking the no-restart path.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Found 1 issue.

Reviewed 43 of 43 changed files.

---

FrankChen021 (commented):
I have reviewed the updated changes for correctness, edge cases, concurrency, and integration risks; no issues found.

Reviewed 44 of 44 changed files.

---

kfaraz:
@jtuglu1 , thanks for your patience! I have started a final review of this PR and plan to finish within a day.

jtuglu1:
@kfaraz one added benefit to this PR is could allow for auto-scaler to not bloat spec history when doing taskCount scaling updates. E.g. we'd only persist new spec json version when material change to spec has occurred.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -295,6 +297,56 @@ public void merge(@Nullable SupervisorSpec existingSpec)
...
+    final Builder<?> proposedCopy;
+    try {
+      proposedCopy = proposed.toBuilder();
+    }
+    catch (UnsupportedOperationException e) {
+      return true;
+    }
+
+    if (isAutoScalerEnabled() || proposed.isAutoScalerEnabled()) {
+      proposedCopy.taskCount(getIoConfig().getTaskCount());
```
kfaraz, line 313: I suppose this means that changes to `taskCount` will not trigger a restart.
Could you please add a short comment here explaining why it is okay to ignore this?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorSpec.java
```diff
@@ -303,4 +355,87 @@ public abstract SeekableStreamSupervisorSpec createBackfillSpec(
       @Nullable Integer taskCount
   );
 
+  /**
+   * Copy builder for restart comparison. Subclasses override; default requires restart on any change.
+   */
+  public Builder<?> toBuilder()
```
kfaraz: Please make this method abstract so that extensions do not forget to implement this and eventually encounter exception when trying to update a spec.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -165,29 +165,40 @@ public Response specPost(
...
-          auditManager.doAudit(
-              AuditEntry.builder()
-                        .key(spec.getId())
-                        .type("supervisor")
-                        .auditInfo(AuthorizationUtils.buildAuditInfo(req))
-                        .request(AuthorizationUtils.buildRequestInfo("overlord", req))
-                        .payload(auditPayload)
-                        .build()
-          );
+          if (updateResult.isModified() || updateResult.isRestarted()) {
```
kfaraz, line 171: Nit: Should restart in the absence of a modification also require an audit?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorResource.java
```diff
@@ -165,29 +165,40 @@ public Response specPost(
...
+                  "id", spec.getId(),
+                  "modified", updateResult.isModified(),
+                  "restarted", updateResult.isRestarted()
+              )
+          ).build();
         }
     );
   }
 
+  /** Audits supervisor spec submissions that changed or restarted the supervisor. */
```
kfaraz, line 186:
```suggestion
  /**
   * Audits supervisor spec submissions that changed or restarted the supervisor.
   */
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,47 +188,81 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+        spec.merge(existingSpec);
+        createAndStartSupervisorInternal(spec, specChanged);
+        return SupervisorSpecUpdateResult.of(specChanged, true);
+      }
+
+      final Pair<Supervisor, SupervisorSpec> current = supervisors.get(spec.getId());
+      final boolean isNew = current == null || current.rhs == null;
+
+      if (isNew) {
+        spec.merge(null);
```
kfaraz, line 211: This seems unnecessary and error-prone.

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,47 +188,81 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+
+      // Restart path: stop+recreate, persisting the changed spec.
+      possiblyStopAndRemoveSupervisorInternal(spec.getId(), false);
+      createAndStartSupervisorInternal(spec, true);
+      return SupervisorSpecUpdateResult.of(true, true);
     }
   }
 
+  /** Byte-level diff against the running spec; validates allowed updates when bytes differ. */
+  private boolean isSpecChangedAndValidate(SupervisorSpec spec)
```
kfaraz:
```suggestion
  private boolean isSpecChangedAndValidated(SupervisorSpec spec)
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,47 +188,81 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+  {
+    final Pair<Supervisor, SupervisorSpec> current = supervisors.get(spec.getId());
+    final SupervisorSpec currentSpec = current == null ? null : current.rhs;
+    if (!isSpecChanged(spec, currentSpec)) {
+      return false;
+    }
+    if (currentSpec != null) {
+      currentSpec.validateSpecUpdateTo(spec);
+    }
+    return true;
```
kfaraz: Please use a simpler if-else chain to clarify the flow here.

```suggestion
    if (currentSpec == null) {
        return true;
    } else if (!isSpecChanged(spec, currentSpec)) {
      return false;
    } else {
      return currentSpec.validateSpecUpdateTo(spec);
    }
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,47 +188,81 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
-   * Checks whether the submitted SupervisorSpec differs from the current spec in SupervisorManager's supervisor list.
-   * This is used in SupervisorResource specPost to determine whether the Supervisor needs to be restarted
-   *
-   * @param spec The spec submitted
-   * @return boolean - true only if the spec has been modified, false otherwise
+   * Byte-level diff of {@code spec} against {@code against}. A missing {@code against} (new supervisor)
+   * counts as changed. Does not validate the transition.
    */
-  public boolean shouldUpdateSupervisor(SupervisorSpec spec)
+  private boolean isSpecChanged(SupervisorSpec spec, @Nullable SupervisorSpec against)
```
kfaraz: Nit: Maybe use the arg names `currentSpec` and `proposedSpec` for clarity.

```suggestion
  private boolean areSpecBytesChanged(SupervisorSpec proposedSpec, @Nullable SupervisorSpec currentSpec)
```

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,47 +188,81 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+    return true;
+  }
+
   /**
-   * Checks whether the submitted SupervisorSpec differs from the current spec in SupervisorManager's supervisor list.
-   * This is used in SupervisorResource specPost to determine whether the Supervisor needs to be restarted
-   *
-   * @param spec The spec submitted
-   * @return boolean - true only if the spec has been modified, false otherwise
+   * Byte-level diff of {@code spec} against {@code against}. A missing {@code against} (new supervisor)
```
kfaraz: How can the new supervisor be missing?

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -186,47 +188,81 @@ public boolean createOrUpdateAndStartSupervisor(SupervisorSpec spec)
...
+      final Pair<Supervisor, SupervisorSpec> current = supervisors.get(spec.getId());
+      final boolean isNew = current == null || current.rhs == null;
+
+      if (isNew) {
+        spec.merge(null);
+        createAndStartSupervisorInternal(spec, true);
+        return SupervisorSpecUpdateResult.of(true, true);
+      }
+
+      if (!isSpecChanged(spec, current.rhs)) {
```
kfaraz: Assign currentSpec to a variable for readability:
```suggestion
      final SupervisorSpec currentSpec = current.rhs;
      if (!isSpecChanged(spec, currentSpec)) {
```

kfaraz:
> @kfaraz one added benefit to this PR is could allow for auto-scaler to not bloat spec history when doing taskCount scaling updates. E.g. we'd only persist new spec json version when material change to spec has occurred.

@jtuglu1 , do you mean this can be done in a follow up?
Also, isn't it desirable for the auto-scaler to persist updates made to `taskCount`?
Otherwise, if the Overlord crashes, there might be discrepancies when the new Overlord tries to determine the current task count.

kfaraz:
@jtuglu1 , the changes look good.
Please resolve the merge conflicts so that we can merge this one off.

kfaraz (approved):

kfaraz:
@jtuglu1 , could you please also update the release note to mention that if the auto-scaler is enabled, changes to `taskCount` alone do not trigger a restart.

PR #19548, opened 2026-06-03
fix: lookup cache retirement implemented

nozjkoitop:
#### Fixed a race condition in off-heap cached lookups

This PR fixes a race between lookup cache refresh and concurrent query execution for cached-global lookups, using off-heap mmap-backed storage.

Previously, query paths could obtain a `LookupExtractor` backed by the current cache version while a lookup refresh concurrently replaced and deleted that cache version. In off-heap mode, this could leave in-flight query work referencing disposed mmap-backed data.

This PR adds lookup cache retirement and retained lookup extractors. When a cache refresh replaces a lookup cache version, the old version is retired instead of closed immediately. Retired cache versions remain alive while retained references exist, and are disposed once those references are released or after a configured timeout.

#### Added retained lookup extractor lifecycle support

`LookupExtractorFactory` now has an optional `acquireRetainedLookupExtractor()` method. Factories that support retained backing resources can return a `RetainedLookupExtractor`, which wraps the actual `LookupExtractor` and closes the retained backing reference when query work is finished.

`RetainedLookupExtractor` supports explicit `close()` for query paths with a lifecycle hook, and also uses a `Cleaner` fallback for paths that hand lookup extractors to APIs without an explicit close hook.

Retained lookup extractors are now used across lookup query paths, including:

- registered lookup extraction functions
- lookup dimension specs
- lookup joins
- lookup segment iteration

#### Added cache retirement to cached-global lookups

`CacheScheduler.VersionedCache` now tracks cache lifecycle states: live, retired, and disposed. Cache versions can be retained by active query work. When a refresh swaps in a new cache version, the previous version is retired and tracked until it can be safely disposed.

To avoid unbounded retention, refreshes are skipped if the number of retained retired cache versions reaches the configured limit. Retired cache entries may also be forcibly disposed after a timeout to prevent abandoned references from blocking future refreshes indefinitely.

Two new cached-global lookup configuration properties were added:

| Property | Description | Default |
|---|---|---:|
| `druid.lookup.namespace.maxRetiredCacheEntries` | The maximum number of retired lookup cache versions to keep while they are still retained by in-flight query work. Further cache refreshes are skipped until the retired cache versions are released or time out. | `1` |
| `druid.lookup.namespace.retiredCacheEntryTimeoutMillis` | The amount of time after retirement before a retained cache version may be disposed even if its retained references have not been closed yet. This prevents abandoned retained references from blocking future cache refreshes indefinitely. | `900,000` |

#### Release note

Cached-global lookups now retain retired cache versions while in-flight queries are still using them, preventing queries from referencing disposed off-heap lookup cache data during concurrent lookup refreshes.

Two new configuration properties control this behavior: `druid.lookup.namespace.maxRetiredCacheEntries` and `druid.lookup.namespace.retiredCacheEntryTimeoutMillis`.

<hr>

 * `RetainedLookupExtractor`
 * `LookupExtractorFactory`
 * `NamespaceLookupExtractorFactory`
 * `CacheScheduler`
 * `NamespaceExtractionConfig`
 * `LookupDimensionSpec`
 * `RegisteredLookupExtractionFn`
 * `LookupSegment`
 * `LookupJoinable`
 * `LookupJoinableFactory`

<hr>

This PR has:

- [x] been self-reviewed.


FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 0 |
| P3 | 0 |
| Total | 2 |

Reviewed 20 of 20 changed files.

---

server/src/main/java/org/apache/druid/segment/join/LookupJoinableFactory.java
```diff
@@ -57,7 +57,7 @@ public Optional<Joinable> build(final DataSource dataSource, final JoinCondition
     if (condition.canHashJoin()) {
       final String lookupName = lookupDataSource.getLookupName();
       return lookupProvider.get(lookupName)
-                           .map(c -> LookupJoinable.wrap(c.getLookupExtractorFactory().get()));
+                           .map(c -> LookupJoinable.wrap(c.getLookupExtractorFactory()));
```
FrankChen021: [P1] Pin one lookup snapshot for each joinable

Wrapping the factory means a single LookupJoinable can observe different cache versions across getMatchableColumnValues/getCorrelatedColumnValues during join filter conversion or preanalysis, then makeJoinMatcher during execution. If a namespace refresh lands between those calls, pushed-down left filters can be based on version N while the matcher uses version N+1, which can drop rows that should match the execution-time lookup. The lookup version needs to be retained once for the query/joinable lifecycle and released through the existing reference path, not reacquired independently per method.

processing/src/main/java/org/apache/druid/query/dimension/LookupDimensionSpec.java
```diff
@@ -136,18 +139,31 @@ public String getName()
...
-          .get();
+      return this.lookup;
     }
 
+    final LookupExtractorFactory lookupExtractorFactory = lookupExtractorFactoryContainerProvider
+        .get(name)
+        .orElseThrow(() -> new ISE("Lookup [%s] not found", name))
+        .getLookupExtractorFactory();
+
+    final Optional<RetainedLookupExtractor> retainedLookupExtractor =
```
FrankChen021, line 160: [P1] Avoid retaining lookup caches for metadata-only calls

getExtractionFn now acquires a retained lookup, but callers such as preservesOrdering, ColumnProcessors.computeDimensionSpecCapabilities, and DimensionHandlerUtils call getExtractionFn just to inspect metadata and then drop the returned function without a close hook. Those retained references are released only by Cleaner/GC, so normal capability checks can keep retired cache versions referenced and block later namespace refreshes until GC or the 15 minute timeout. Metadata paths should avoid the retained acquisition or close it deterministically after inspection.

server/src/main/java/org/apache/druid/segment/join/LookupJoinableFactory.java
```diff
@@ -57,9 +61,20 @@ public Optional<Joinable> build(final DataSource dataSource, final JoinCondition
...
     } else {
       return Optional.empty();
     }
   }
+
+  private LookupJoinable buildLookupJoinable(final LookupExtractorFactoryContainer container)
+  {
+    final LookupExtractorFactory lookupExtractorFactory = container.getLookupExtractorFactory();
+    final Optional<RetainedLookupExtractor> retainedLookupExtractor =
+        lookupExtractorFactory.acquireRetainedLookupExtractor();
```
FrankChen021, line 73: [P1] Close retained lookup joinables deterministically

After the incremental change, buildLookupJoinable acquires a RetainedLookupExtractor once and stores it in the joinable, but LookupJoinable.acquireReference() now returns only a no-op closeable. JoinDataSource.createSegmentMapFunction has no close hook for the captured joinable list, so HashJoinSegment.close() never closes this retained extractor; it is released only by the Cleaner after GC. For cached namespace lookups, one retained retired version is enough to make CacheScheduler skip further refreshes with the default maxRetiredCacheEntries=1, so lookup-join queries can stall later lookup updates until GC or timeout. Please attach the retained handle to an explicit query/joinable lifecycle or use a real ref-counted closeable instead of the no-op reference.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Reviewed 22 of 22 changed files.

---

server/src/main/java/org/apache/druid/segment/join/LookupJoinableFactory.java
nozjkoitop: Yeah, `acquireReference()` is per mapped segment, but the pinned lookup lives at the joinable/query level, so closing it there is too narrow and can be wrong if the joinable is reused.

Deterministic cleanup would need a real close hook for the Joinable/JoinableClause lifecycle, not just wiring the retained handle into `LookupJoinable.acquireReference()`, which will mean the join api change and I'm not sure if we want it under this pull request, would it be reasonable to leave the Cleaner fallback here and handle deterministic cleanup as a follow up?

FrankChen021: I agree that wiring this into `LookupJoinable.acquireReference()` would be the wrong lifecycle, but I don't think the Cleaner-only path is reasonable for lookup joins as-is. The current code still retains the namespace cache in `LookupJoinableFactory` and the join path has no deterministic close for that retained extractor, while the scheduler intentionally skips further loads once a retired retained cache reaches the default max of 1. So a lookup-join query can keep later namespace updates from loading until GC or the retired-cache timeout. I'd either keep joins on the non-retained `get()` path until the joinable/query lifecycle can own a close hook, or include that lifecycle fix in this PR.

Reviewed 13 files total, including 12 of 22 changed files.

nozjkoitop: Fair, leaving it to GC isn’t ideal. But I also don’t want to leave joins without lookup retirement since that’s kinda the goal of this PR, so I added deterministic closure for those joinables and tried to keep the diff minimal, wdyt?

FrankChen021: Thanks, this addresses the joinable lifecycle issue I was worried about. The retained lookup is now owned by the query-scoped SegmentMapFunction, and the server, realtime, and MSQ callers I checked close that function after the mapped segments/segment references, so the join path no longer appears to rely on the Cleaner for normal completion.

Reviewed 32 of 32 changed files.

---

processing/src/main/java/org/apache/druid/query/dimension/LookupDimensionSpec.java
```diff
@@ -136,18 +139,51 @@ public String getName()
...
+    }
+
+    final LookupExtractorFactory lookupExtractorFactory = getLookupExtractorFactory();
 
+    final Optional<RetainedLookupExtractor> retainedLookupExtractor =
+        lookupExtractorFactory.acquireRetainedLookupExtractor();
+
+    // ExtractionFn has no close hook. The RetainedLookupExtractor cleaner releases this reference when the
+    // LookupExtractionFn that owns it becomes unreachable.
+    return retainedLookupExtractor.<LookupExtractor>map(retained -> retained).orElseGet(lookupExtractorFactory);
```
FrankChen021, line 165: [P1] Do not retain lookup dimension extraction functions without a close path

getExtractionFn() now acquires a RetainedLookupExtractor and returns it inside a plain LookupExtractionFn, but that API has no close hook. Several callers do not cache or close the returned function, for example TopNQueryQueryToolChest calls topNQuery.getDimensionSpec().getExtractionFn().apply(...) while post-processing each result value, and GroupByQueryQueryToolChest.extractionsToRewrite calls it just to inspect getExtractionType(). For named lookup dimensions, each call increments the namespace cache reference and then depends on Cleaner/GC to release it, so with the new default maxRetiredCacheEntries=1 a query can make later namespace refreshes skip until GC or the 15 minute timeout. Please avoid retaining from LookupDimensionSpec.getExtractionFn() unless the returned handle has a deterministic query/result lifecycle, or update these call sites to use a scoped retained wrapper that is explicitly closed.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Reviewed 32 of 32 changed files.

---

processing/src/main/java/org/apache/druid/query/dimension/LookupDimensionSpec.java
nozjkoitop: that makes sense, I moved the rest of the metadata-only checks to the non-retained path and changed TopN so it only retains once, but the other call sites should stay on the retiring flow so in-flight queries don’t lose the lookup they’re using

FrankChen021: I think one wrapper path still needs the same treatment. For a `ListFilteredDimensionSpec`, `RegexFilteredDimensionSpec`, or `PrefixFilteredDimensionSpec` around a named `LookupDimensionSpec`, `BaseFilteredDimensionSpec` inherits the default `getExtractionFnForMetadata()`, which calls `getExtractionFn()` and therefore still acquires/discards a retained lookup during the metadata-only checks you updated. `ListFilteredDimensionSpec.decorate` also calls `delegate.getExtractionFn() != null` only as an inspection, so that path has the same no-close behavior.

Can you add a `BaseFilteredDimensionSpec#getExtractionFnForMetadata()` override that delegates to `delegate.getExtractionFnForMetadata()`, and switch the `ListFilteredDimensionSpec.decorate` boolean check to the metadata accessor as well?

Reviewed 36 of 36 changed files.

nozjkoitop: fair, updated

FrankChen021 (commented):
Follow-up handled. The filtered dimension metadata path now avoids the retained lookup path: BaseFilteredDimensionSpec delegates getExtractionFnForMetadata() to its wrapped spec, and ListFilteredDimensionSpec.decorate uses the metadata accessor for its inspection-only check.

---

FrankChen021 (commented):

---

processing/src/main/java/org/apache/druid/query/dimension/LookupDimensionSpec.java
```diff
@@ -136,18 +139,51 @@ public String getName()
...
+
+  private LookupExtractor getLookupExtractor()
+  {
+    if (Strings.isNullOrEmpty(name)) {
+      return this.lookup;
+    }
+
+    final LookupExtractorFactory lookupExtractorFactory = getLookupExtractorFactory();
 
+    final Optional<RetainedLookupExtractor> retainedLookupExtractor =
```
FrankChen021, line 160: [P2] Avoid retaining a lookup for every joined row

LookupDimensionSpec.getExtractionFn() now acquires a retained cache reference, but LookupColumnSelectorFactory calls it twice from getValue() for every joined row (once for the null check and once for apply). The first RetainedLookupExtractor is immediately discarded and the second is never explicitly closed because the selector has no close hook. A lookup join that projects an extraction on k/v can therefore accumulate Cleaner-managed references, pin retired cache versions, and make CacheScheduler skip refreshes until the timeout. Hoist one extraction function per selector and give its retained resource an owner/lifetime; use the metadata accessor for the presence check.

processing/src/main/java/org/apache/druid/query/dimension/LookupDimensionSpec.java
```diff
@@ -136,18 +139,51 @@ public String getName()
   @Override
   public ExtractionFn getExtractionFn()
   {
-    final LookupExtractor lookupExtractor;
+    final LookupExtractor lookupExtractor = getLookupExtractor();
+    return makeLookupExtractionFn(lookupExtractor);
+  }
+
+  @Override
+  public ExtractionFn getExtractionFnForMetadata()
```
FrankChen021, line 147: [P2] Use the metadata accessor for presence checks

The new getExtractionFnForMetadata() contract exists specifically because capability checks discard the result without a close hook, but several existing metadata-only checks still call getExtractionFn(): ListFilteredVirtualColumn (including getIndexSupplier), ShimColumnSelectorFactory, and SortMergeJoinFrameProcessor. For a named cached lookup these calls now acquire and abandon a retained cache reference, even on paths that immediately reject the query, unnecessarily pinning retired versions. Switch those presence checks to getExtractionFnForMetadata().

processing/src/main/java/org/apache/druid/query/JoinDataSource.java
```diff
@@ -384,49 +386,66 @@ public SegmentMapFunction createSegmentMapFunction(Query query)
...
-      clausesToUse = conversionResult.rhs;
 
-    } else {
-      baseFilterToUse = baseFilter;
-      clausesToUse = joinableClauses.getJoinableClauses();
-    }
+    // joinables may retain resources (such as pinned lookup versions) that are owned by the returned function's
+    // close(), if anything below throws before the function takes ownership, release them here instead of leaving
+    // them to the Cleaner
+    try {
```
FrankChen021, line 393: [P2] Release joinables when clause construction fails

The cleanup try/catch starts after JoinableClauses.createClauses() returns. That method builds clauses incrementally, so if a later LookupJoinableFactory.build() or retained-reference acquisition fails, any earlier LookupJoinable is lost without close(). The new retained lookup reference can then block cache retirement and refreshes until Cleaner/timeout. Make clause construction close already-built closeable joinables on failure, or establish ownership before building the next clause.

FrankChen021 (commented):
## 🟡 Changes recommended

The PR is cleanly mergeable, but the retained lookup lifecycle is incomplete in several query and failure paths. These paths can abandon retained cache references, pin retired versions, and cause lookup refreshes to be skipped until timeout.

Validation: `git diff --check` passed; GitHub reported `mergeable=true` and `mergeable_state=clean`, and the prepared head was verified to have `master` as an ancestor. No build or tests were run.

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 4 |
| P3 | 0 |
| Total | 4 |



---

server/src/main/java/org/apache/druid/server/LocalQuerySegmentWalker.java
```diff
@@ -109,10 +109,15 @@ public <T> QueryRunner<T> getQueryRunnerForIntervals(final Query<T> query, final
                            .map(queryRunnerFactory::createRunner).iterator()
     );
 
+    // close the segment map function when the result sequence is fully consumed or closed, releasing resources it
+    // retains for the query, such as lookup versions pinned for joins
+    final QueryRunner<T> closingRunner =
+        (queryPlus, responseContext) -> baseRunner.run(queryPlus, responseContext).withBaggage(segmentMapFn);
```
FrankChen021, line 115: [P2] Close the segment map function on query errors

The segmentMapFn is created before runner setup, but it is only attached as sequence baggage after baseRunner.run() returns. If runner construction or execution throws, the resource-bearing SegmentMapFunction and its retained lookup joinables are not closed and must wait for Cleaner. Wrap setup/run failures with an explicit close, matching the cleanup already added to ServerManager and SinkQuerySegmentWalker.

PR #19549, opened 2026-06-03
feat: introduce intermediate valid task counts for big partition counts

Fly-Style:
**Description**
  
The cost-based autoscaler derives candidate task counts from possible partitions-per-task ratios. For large partition counts these candidates can be very far apart near the top of the assignment range - e.g. for 400 partitions, the candidates scales `200 -> 400`. Because the cost model only evaluates the generated candidates, it has no intermediate option to settle on, forcing coarse, all-or-nothing scaling decisions.
  
This PR adds deterministic intermediate candidates so the cost model has finer-grained options, without changing the cost model itself. 
  
**Introduced intermediate valid task counts for large gaps**
  
`CostBasedAutoScaler.computeValidTaskCounts` now post-processes the generated candidate list: after the base partitions-per-task candidates are produced and sorted, every adjacent pair whose gap exceeds `MAX_CANDIDATE_GAP` (100) is split with intermediate candidates at`INTERPOLATION_FRACTIONS = {0.33, 0.66}`, rounded to the nearest integer.

**Tuned the lag amplification multiplier:** 
`WeightedCostFunction.LAG_AMPLIFICATION_MULTIPLIER` is lowered from 0.4 to 0.35 based on further testing, for a slightly more balanced high-lag recovery response.

**Release note**
  
The cost-based supervisor autoscaler now considers intermediate task counts when the candidate task counts derived from partition assignment are far apart, enabling smoother scaling for streams with large partition counts.
 
  This PR has:

  - [x] been self-reviewed.
  - [x] added unit tests covering the new candidate-generation behaviour (no-gap, single-gap,
  multi-gap, and single-candidate cases) in CostBasedAutoScalerTest.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
@@ -78,6 +78,14 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
    */
   static final int MAX_IDLENESS_PARTITION_LAG = 10_000;
 
+  /**
+   * Adjacent candidate task counts whose gap exceeds this value are split with intermediate
+   * candidates so the cost model has finer-grained options to evaluate.
+   */
+  static final int MAX_CANDIDATE_GAP = 100;
+
+  static final double[] INTERPOLATION_FRACTIONS = new double[]{0.33, 0.66};
+
   private final String supervisorId;
   private final SeekableStreamSupervisor supervisor;
   private final ServiceEmitter emitter;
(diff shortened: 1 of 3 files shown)


FrankChen021 (commented):

---

kfaraz:
@Fly-Style , IIUC, the partition assignment is all-or-nothing by design. 

Could you please elaborate on the benefits of the intermediate scaling?
e.g. in the case of 400 partitions, what would be the benefit of scaling from 200 tasks to say 300 tasks (instead of directly to 400)?

With 300 tasks, 100 tasks would be working on 2 partitions each and 200 tasks would be working on 1 partition each (and we cannot control which partitions get assigned to which tasks). So, the tasks working on 2 partitions have effectively not changed and would continue to be the ingestion bottleneck.

Please note that in a non-uniform partition assignment, the idleness cost computation would be a little more involved, since the tasks working on fewer partitions would be more idle than others on average.

Fly-Style:
@kfaraz During our usage of this autoscaler, we saw that, let's say, 200 tasks to handle 400 partitions were under-provisioned, and 400 tasks were over-provisioned, and adjusting the weights didn't improve the situation much.

 This PR was introduced to provide better variants to the autoscaler in such a situation, and it acknowledges of margin of some tasks would handle 2 partitions instead of 1.

jtuglu1:
At some point I wonder how complicated we really want to make this thing. IMO, auto-scalers should be sufficiently "dumb" in that their behavior is predictable enough to debug on-call. Each "edge case" behavior adds more operator/code complexity which IMO isn't the direction we should be heading. 

If possible, I'd rather attack the core problem (that we cannot scale frequently without disruption) by potentially allowing "sticky" partitions to groupIds or somehow allowing tasks to dynamically get re-assigned to partitions (without shutting down). This would allow a "dumber" auto-scaler to run more frequently w/o the worry of disrupting high-throughput supervisors, as tasks could dynamically adjust to the suggested partition assignments.

Fly-Style:
@jtuglu1 partially done in https://github.com/apache/druid/pull/19562 :)

Also we have dynamic partition reassignment in mind, it is the next step.

Fly-Style:
@jtuglu1 returning to this discussion - one of the core concerns this PR tries to address is cost ($$$) effectiveness. Let's say, if you have 500 partitions and 250 tasks is still not worse than 500 tasks from the autoscaler perspective; I would sacrifice even partition distribution for the sake of cost reduction in that case.

kfaraz:
@Fly-Style , I agree with @jtuglu1 .
Let's keep this PR on hold.

There is not a strong case to be made for intermediate task counts because, with say 500 partitions, going from 250 tasks to 300 tasks would still leave several tasks processing 2 partitions each, thus not helping with the lag.
Allowing intermediate task counts in `costBased` auto-scaler would take us back in the direction of the old `lagBased` auto-scaler, which doesn't seem desirable.

In terms of cost, while 300 tasks would cost less than 500, there is no point in trying 300 if we know that they are not going to keep up. In fact, it would only delay the scale up to 500, which may also be a cost/SLA concern for the operator.

The only case where 300 tasks would help is if there is skew in the volume of data for each partition. But the correct solution for that would be to identify the slow tasks/partitions and try to focus the scale up only on those.

PR #19552, opened 2026-06-04
feat: add auth metrics emission

cecemei:
### Description

Introduces an opt-in mechanism to emit Druid service metrics for authorization events. When enabled, Druid emits metrics for access denials and authorization exceptions, making it easier to monitor and alert on security-relevant events.

**Changes**

New config flag (`AuthConfig`)
- Adds `emitAuthMetrics` boolean property (default: `false`)
- Enable via runtime config: `druid.auth.emitAuthMetrics=true`

`AuthorizerMapper`
- Accepts an optional `ServiceEmitter` to pass through to authorization utilities
- Defaults to null (no-op) when metrics are disabled

`AuthorizerMapperModule`
- Injects `ServiceEmitter` directly instead of via `Injector` (cleaner dependency graph)
- Wires the emitter into `AuthorizerMapper` only when `emitAuthMetrics` is enabled

`AuthorizationUtils`
- Adds private `emitAuthMetric()` helper that fires metrics with dimensions: `identity`, `authorizerName`, `resourceName`, `resourceType`, `action`, and `errorMessage`
- Emits `auth/forbidden` on denied resource access
- Emits `auth/exception` on internal errors (missing authorizer, duplicate policies, double-authorization check)

### Release Notes

Druid can now emit service metrics for authorization events. Set `druid.auth.emitAuthMetrics=true` to enable.

When enabled, two new metrics are available:

| Metric | Emitted when |
|--------|-------------|
| `auth/forbidden` | A request is denied due to insufficient permissions |
| `auth/exception` | An internal authorization error occurs (missing authorizer, duplicate policies, double-auth check) |

Both metrics include dimensions `identity`, `authorizerName`, `resourceName`, `resourceType`, `action`, and `errorMessage` (where applicable), enabling precise alerting on security events.

<hr>

This PR has:

- [x] been self-reviewed.


server/src/main/java/org/apache/druid/server/security/AuthorizationUtils.java
```diff
@@ -90,6 +96,13 @@ public static void verifyUnrestrictedAccessToDatasource(
     ResourceAction resourceAction = createDatasourceResourceAction(datasource, req);
     AuthorizationResult authResult = authorizeResourceAction(req, resourceAction, authorizerMapper);
     if (!authResult.allowAccessWithNoRestriction()) {
+      emitAuthMetric(
```
FrankChen021: [P2] Avoid double-counting datasource denial metrics

When a datasource request is denied, verifyUnrestrictedAccessToDatasource first calls authorizeResourceAction, which now emits auth/accessDenied from authorizeAllResourceActions when the authorizer returns a denial. The new emit here then records the same denial again before throwing ForbiddenException, so endpoints using this helper double-count access-denied metrics whenever druid.auth.emitAuthMetrics=true. Emit here only for the allowed-but-restricted case, or keep access-denied emission in one layer.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/server/security/AuthorizationUtils.java
```diff
@@ -90,6 +96,13 @@ public static void verifyUnrestrictedAccessToDatasource(
...
+          authorizerMapper.getServiceEmitter(),
+          authenticationResultFromRequest(req),
+          resourceAction,
+          METRIC_ACCESS_DENIED,
+          authResult.getErrorMessage()
+      );
       throw new ForbiddenException(authResult.getErrorMessage());
     }
```
capistrant, line 112: I think the comment from Frank is right. Could we just do a quick check on basic access and emit if basic access was granted but not if it wasn't, avoiding the double emit in that latter case?

```suggestion
 if (!authResult.allowAccessWithNoRestriction()) {
      if (authResult.allowBasicAccess()) {
        // Basic access was granted, but access was restricted by a policy.
        // This is checked to avoid double emitting the access denied metric if the basic access was denied,
        // since the authorizeResourceAction method already emits the metric in that case.
        emitAuthMetric(
            authorizerMapper.getServiceEmitter(),
            authenticationResultFromRequest(req),
            resourceAction,
            METRIC_ACCESS_DENIED,
            authResult.getErrorMessage()
        );
      }
      throw new ForbiddenException(authResult.getErrorMessage());
  }
```

capistrant (changes requested):
looks good other than Franks concern about the double emit when using the restricted access helper. I think that is legit problem. please check validity of comment and see my proposed mitigation to avoid double emit (if there is better solution you can think of please do ignore my suggestion)

edit: can you also add the metric and config to docs + consolidate a release note section in description?

server/src/main/java/org/apache/druid/server/security/AuthorizationUtils.java
```diff
@@ -50,6 +53,9 @@ public class AuthorizationUtils
       ResourceType.DATASOURCE
   );
 
+  private static final String METRIC_ACCESS_DENIED = "auth/accessDenied";
```
capistrant: I might consider `auth/forbidden` but that feels like personal preference so just sharing in case you are interested

cecemei: updated to auth/forbidden

server/src/test/java/org/apache/druid/server/security/AuthorizationUtilsTest.java
```diff
@@ -21,6 +21,8 @@
```
capistrant, line 21: metric names out of date with forbidden change

capistrant (changes requested):
just some metric name stuff pending in test but other than that looks good and I will approve

capistrant (approved):
+1 on green CI

PR #19553, opened 2026-06-04
fix: Retain exception that triggered Dart query cancellation.

gianm:
In some cases, cancellation is triggered by an exception (rather than a non-exceptional reason, like timeout or user request). This patch retains the exception and includes it in the query report.

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/sql/DartQueryMaker.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/dart/controller/sql/DartQueryMaker.java
@@ -279,7 +279,7 @@ public class DartQueryMaker implements QueryMaker
       return Sequences.simple(List.<Object[]>of(new Object[]{reportMap}));
     }
     catch (InterruptedException e) {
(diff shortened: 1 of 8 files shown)


FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

- multi-stage-query/src/main/java/org/apache/druid/msq/exec/ControllerHolder.java:249 - [P2] Preserve cancellation causes before the controller starts. cancel(reason, cause) only records cancelReason while the holder is still ACCEPTED. If the Dart result sequence closes or throws before the run executor starts the controller, the later runAsync path skips controller.stop(...) and builds makeCanceledReport(cancelReason), so the exception passed from DartQueryMaker is dropped and the report is still a bare UNKNOWN cancellation. Store the cause in the holder and use it when building the pre-run canceled report.

---

capistrant (approved):
Franks auto review comment looks accurate, but it feels like an improvement that could be a follow if you don't want to work on implementing the idea now

PR #19557, opened 2026-06-04
feat: post-pass shard-group reconciliation for asymmetric partial-load matchers

clintropolis:
### Description
This PR adds a post-processing step to `RunRules` so that partial-load matchers that resolve differently across segments of a shard group (e.g. `ClusterGroupPartialLoadMatcher` over range-partitioned segments) don't leave the broker with an incomplete `PartitionHolder`. `Rule.run` now returns a new `RuleRunResult` object that `RunRules` can use to 'do stuff', though most implementations today do `RuleRunResult.OK`. For these asymmetric partial matchers, there is a `ShardGroupFollowup` implementation of `RuleRunResult` that `RunRules` can use to check the siblings of a shard group to perform an 'empty' partial load to ensure that the group is fully available. I am unsure if there is enough of a pattern to make `RunRules` more generically handle `RuleRunResult`, i was planning to wait and see if any other use cases pop up before trying to make this handling more generic.

Doing an empty load like this seemed cheaper than trying to figure out how to allow the timeline to sometimes allow incomplete groups to appear as complete, since the empty partial load should be cheap for the target historicals.

changes:
* adds `PartialLoadMatcher.emptyMatch` (default null) as an opt-in way to do a zero-content "load" for matchers that can resolve asymmetrically.
* `Rule.run` returns a `RuleRunResult` (`OK` constant for most rules). `PartialLoadRule` returns a `ShardGroupFollowup` on a positive match, but only when numCorePartitions > 0.
* `RunRules` streams followups in a buffer keyed by (dataSource, interval, version), flushing at iteration boundaries (NEWEST_SEGMENT_FIRST groups segments contiguously by that triple). Flush dispatches emptyMatch loads to unmatched core siblings; siblings not part of the core partition group are skipped.
* adds `TimelineLookup.findChunks` which returns a defensive copy via `PartitionHolder.copyWithOnlyVisibleChunks` so iteration is safe outside the timeline lock.
* `PartialClusterGroupLoadSpec` allows an empty `clusterGroupIndices` list as the wire form of an empty match.

### DIFF
--- a/server/src/main/java/org/apache/druid/server/coordinator/rules/PartialLoadRule.java
+++ b/server/src/main/java/org/apache/druid/server/coordinator/rules/PartialLoadRule.java
@@ -88,23 +88,29 @@ public abstract class PartialLoadRule extends LoadRule
   }
 
   @Override
-  public void run(DataSegment segment, SegmentActionHandler handler)
+  public RuleRunResult run(DataSegment segment, SegmentActionHandler handler)
   {
     final PartialLoadMatcher.MatchResult result = matcher.match(segment, segment.getLoadSpec());
     if (result != null) {
-      // Matcher resolved: route through the partial-load handler. The wrappedLoadSpec map carries scheme-specific
-      // data that the historical-side wrapper deserializes.
       handler.replicateSegmentPartially(
           segment,
           PartialLoadProfile.forRequest(result.wrappedLoadSpec(), result.fingerprint()),
           getTieredReplicants()
       );
-    } else {
-      // Matcher does not apply, but the rule still applies because onCannotMatch == FULL_LOAD (FALL_THROUGH would
-      // have caused appliesTo to return false, so run wouldn't be invoked). Route through the regular full-load
-      // handler.
-      handler.replicateSegment(segment, getTieredReplicants());
+      // Flag the shard group for a post-pass: if the matcher resolves asymmetrically across siblings, RunRules will
+      // dispatch emptyMatch loads to the unmatched ones so the broker's PartitionHolder.isComplete() check still
+      // treats the group as queryable. Matchers without an emptyMatch (the default) make the post-pass a no-op.
+      // Only the core partition group has an atomic-replace completeness requirement on the broker, segments with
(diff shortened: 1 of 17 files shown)


server/src/main/java/org/apache/druid/server/coordinator/rules/PartialLoadRule.java
```diff
@@ -88,23 +88,29 @@ public boolean appliesTo(DataSegment segment, DateTime referenceTimestamp)
...
-      // have caused appliesTo to return false, so run wouldn't be invoked). Route through the regular full-load
-      // handler.
-      handler.replicateSegment(segment, getTieredReplicants());
+      // Flag the shard group for a post-pass: if the matcher resolves asymmetrically across siblings, RunRules will
+      // dispatch emptyMatch loads to the unmatched ones so the broker's PartitionHolder.isComplete() check still
+      // treats the group as queryable. Matchers without an emptyMatch (the default) make the post-pass a no-op.
+      // Only the core partition group has an atomic-replace completeness requirement on the broker, segments with
+      // no core partitions (e.g. append-only streaming) don't need sibling reconciliation.
+      if (segment.getShardSpec().getNumCorePartitions() > 0) {
+        return new ShardGroupFollowup(segment, matcher, getTieredReplicants());
```
capistrant, line 106: I'm a bit confused about how this behaves when using`onCannotMatch == FULL_LOAD` like won't this cause empty loads of other segments that fell through to full load?

capistrant (commented):
pretty cool idea! Biggest concern is behavior when `onCannotMatch = FULL_LOAD` won't we (potentially) create follow ups for segments that actually loaded?

server/src/main/java/org/apache/druid/server/coordinator/rules/PartialLoadRule.java
capistrant: will this create core partition set follow ups even when segments who are outside of the core are the only ones who matched? do we want that?

clintropolis:
^ closed in favor of #19565

PR #19558, opened 2026-06-04
fix: retry on transient AWS credential resolution failures

jtuglu1:
### Description

S3 segment pushes that use the AWS SDK v2 transfer manager can resolve credentials on the async upload path. If a file-session credential refresh, container credential lookup, or IMDS lookup is temporarily unavailable, the SDK reports an SdkClientException such as 'Unable to load credentials from any of the providers in the chain'.

Druid's S3 push path already wraps uploads in retryS3Operation, but these credential-provider failures were not classified as recoverable after the SDK v2 migration. That made an intermittent credential miss fail the task immediately instead of using the existing retry budget.

#### Release note

Retry on transient AWS credential resolution failures

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/cloud/aws-common/src/main/java/org/apache/druid/common/aws/AWSClientUtil.java
+++ b/cloud/aws-common/src/main/java/org/apache/druid/common/aws/AWSClientUtil.java
@@ -62,6 +62,12 @@ public class AWSClientUtil
       "Throttling"
   );
 
+  private static final String UNABLE_TO_LOAD_CREDENTIALS_FROM_PROVIDER_CHAIN =
+      "Unable to load credentials from any of the providers in the chain";
+  private static final String FAILED_TO_LOAD_CREDENTIALS_FROM_IMDS = "Failed to load credentials from IMDS";
+  private static final String CANNOT_REFRESH_AWS_CREDENTIALS = "cannot refresh AWS credentials";
+  private static final String CANNOT_FETCH_CREDENTIALS_FROM_CONTAINER = "Cannot fetch credentials from container";
+
   /**
    * Checks whether an exception can be retried or not for AWS SDK v2.
    */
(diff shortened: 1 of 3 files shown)


cloud/aws-common/src/main/java/org/apache/druid/common/aws/AWSClientUtil.java
```diff
@@ -62,6 +62,12 @@ public class AWSClientUtil
       "Throttling"
   );
 
+  private static final String UNABLE_TO_LOAD_CREDENTIALS_FROM_PROVIDER_CHAIN =
```
jtuglu1, line 65: ideally it'd be better to catch a particular expired creds exception type but these messages are all generic AwsSdkException class, so parsing the message is needed.

cloud/aws-common/src/main/java/org/apache/druid/common/aws/AWSClientUtil.java
```diff
@@ -62,6 +62,12 @@ public class AWSClientUtil
...
+      "Unable to load credentials from any of the providers in the chain";
+  private static final String FAILED_TO_LOAD_CREDENTIALS_FROM_IMDS = "Failed to load credentials from IMDS";
+  private static final String CANNOT_REFRESH_AWS_CREDENTIALS = "cannot refresh AWS credentials";
```
abhishekrb19, line 68: It seems like `cannot refresh AWS credentials` is not present in the v2 SDK: https://github.com/aws/aws-sdk-java-v2

Maybe this was a legacy one from v1 SDK that isn't relevant anymore and can be removed?

abhishekrb19 (approved):
One comment, lgtm otherwise

cloud/aws-common/src/main/java/org/apache/druid/common/aws/AWSClientUtil.java
jtuglu1: Possibly, I think it's fine to keep though; would rather retry on a false positive than risk killing a task in-case it surfaces somehow. I can push a follow-up to delete if we need to.

PR #19559, opened 2026-06-04
feat: Make segmentLoadAheadCount able to be configured at worker task level in addition to context

capistrant:
### Description

Allow configuring `segmentLoadAheadCount` for dart workers via local runtime property. This unlocks flexibility in situations where you do not want to rely on a controller provided context value to control this property. It can be powerful in situations where you deploy multiple tiers of historicals on disperate hardware, and would like to customize the value to maximize a given hardware spec's capabilities. unset by default, which maintains behavior that already exists - the controller provided context provides the value OR a default `2 * threads` is used.

#### Release note

Add dart worker runtime property `druid.msq.dart.worker.segmentLoadAheadCount` to allow worker local configuration of `segmentLoadAheadCount`. If configured to be `> 0` for a worker, this becomes the default `segmentLoadAheadCount` value for the worker.  If a query controller also sends `segmentLoadAheadCount` in the context for a query, the context value wins for that query.. If not set, the worker will either use the controller provided context value or the default `2 * threads` if nothing is set in context.

<hr>

 * `DartWorkerConfig`
 * `DartWorkerContext`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/dart/guice/DartWorkerConfig.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/dart/guice/DartWorkerConfig.java
@@ -48,10 +48,9 @@ public class DartWorkerConfig
    * Worker-local value for the segment load-ahead count used to size segment prefetch in {@link ReadableInputQueue}.
    * <p>
    * Defaults to null (unset), which leaves the value to query context (client or controller-default supplied) and the
-   * built-in {@code 2 * threadCount} fallback. When set to a positive value, it acts as a worker-local floor: the
-   * effective count becomes the larger of this value and any value present in the query context, so a query can still
-   * request more, but never less than this floor. Non-positive values are treated as unset.
-   *
+   * built-in {@code 2 * threadCount} fallback. When set to a positive value, it acts as a worker-local default used
+   * only when the query context does not supply a value.
+   * <p>
    * This is per-worker-process configuration, set independently on each worker. It lets workers with different
    * hardware (for example, separate tiers with more or less memory and storage bandwidth) tune their own prefetch
    * depth, rather than relying solely on a single cluster-wide query-context default.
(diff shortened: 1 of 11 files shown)


multi-stage-query/src/main/java/org/apache/druid/msq/dart/guice/DartWorkerConfig.java
```diff
@@ -41,6 +44,22 @@ public class DartWorkerConfig
   @JsonProperty("heapFraction")
   private double heapFraction = DEFAULT_HEAP_FRACTION;
 
+  /**
+   * Worker-local value for the segment load-ahead count used to size segment prefetch in {@link ReadableInputQueue}.
+   * <p>
+   * Defaults to null (unset), which leaves the value to query context (client or controller-default supplied) and the
+   * built-in {@code 2 * threadCount} fallback. When set to a positive value, it acts as a worker-local floor: the
+   * effective count becomes the larger of this value and any value present in the query context, so a query can still
```
gianm: I think it would be better to use the value from the query context as-is, if it is provided. That way the query context can more easily be used for experimentation and temporary tuning, in either direction.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/worker/DartWorkerContext.java
```diff
@@ -148,6 +157,9 @@ public class DartWorkerContext implements WorkerContext
     final int baseThreadCount = processingConfig.getNumThreads();
     final Integer maxThreads = MultiStageQueryContext.getMaxThreads(queryContext);
     this.threadCount = (maxThreads != null && maxThreads > 0) ? Math.min(baseThreadCount, maxThreads) : baseThreadCount;
+
+    // Worker-local segment load-ahead config, read once from this worker's DartWorkerConfig.
+    this.segmentLoadAheadCountConfig = injector.getInstance(DartWorkerConfig.class).getSegmentLoadAheadCount();
```
gianm: Probably better to not use the injector so explicitly here. It would be more aligned to do it like the other stuff in this constructor, which is injected into `DartWorkerContextFactoryImpl` and then passed through.

multi-stage-query/src/main/java/org/apache/druid/msq/dart/guice/DartWorkerConfig.java
capistrant: fair point, I can make this switch to the selection precedence

gianm (approved):

FrankChen021 (commented):

---

PR #19560, opened 2026-06-05
feat: S3 SegmentRangeReader for partial segment downloads

clintropolis:
### Description
This PR implements `SegmentRangeReader` for S3 so V10 segments stored unzipped (after this PR) can be partially loaded via range read requests. This is achieved by setting a new `rangeable` boolean on `S3LoadSpec` at publish time, so this will only apply to segments written after this change. Any existing V10 segments will fall back to full download for S3 at least; I think this is ok since it is an experimental feature and off by default, else we would need to do some sort of ugly existence check in a place where a check should be cheap and non-blocking.

changes:
* adds new `S3SegmentRangeReader` that wraps `ServerSideEncryptingAmazonS3` + bucket + key prefix and issues closed-range `GetObjectRequests` against `keyPrefix + filename`. Returned stream is wrapped in a `RetryingInputStream` with the `S3Utils.S3RETRY` predicate (the same retry policy `S3DataSegmentPuller` uses for full-segment downloads) so a transient mid-stream error reopens at the byte offset where it failed and resumes with a fresh range request for the remaining bytes, rather than restarting the whole read.
* New `rangeable` boolean on `S3LoadSpec` stamped by the pusher at write time. `S3LoadSpec.openRangeReader()` returns a reader iff the flag is true and the key isn't .zip
* `S3DataSegmentPusher.pushNoZip` stamps rangeable=true when binaryVersion is `V10_VERSION`, false otherwise. `pushZip` omits the field

### DIFF
--- a/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3SegmentRangeReader.java
+++ b/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3SegmentRangeReader.java
@@ -1,0 +1,147 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.druid.storage.s3;
+
+import com.google.common.base.Preconditions;
+import org.apache.druid.data.input.impl.RetryingInputStream;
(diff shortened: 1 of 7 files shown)


extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3SegmentRangeReader.java
```diff
@@ -0,0 +1,147 @@
...
+    this.bucket = Preconditions.checkNotNull(bucket, "bucket");
+    this.keyPrefix = Preconditions.checkNotNull(keyPrefix, "keyPrefix");
+  }
+
+  @Override
+  public InputStream readRange(String filename, long offset, long length) throws IOException
+  {
+    Preconditions.checkNotNull(filename, "filename");
+    Preconditions.checkArgument(offset >= 0, "offset must be non-negative, got [%s]", offset);
+    Preconditions.checkArgument(length > 0, "length must be positive, got [%s]", length);
```
capistrant: This seemed curious to me. say there is a defined dimension with all nulls.. couldn't that create a 0 byte internal metadata file that would blow up here?

capistrant (commented):
looks good, one open precondition question in the range reader impl

extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3SegmentRangeReader.java
clintropolis: ah, i guess `SegmentFileBuilderV10` technically allows empty add operations (though it would be very unlikely). I've added a short-circuit here to return an empty string if this does happen (since the actual API wouldn't really allow it)

capistrant (approved):

FrankChen021 (commented):

---

PR #19562, opened 2026-06-05
feat: enhance U-shape idle prediction for scale-down scenarios

Fly-Style:
**Description**

The cost-based supervisor autoscaler wouldn't scale down a healthy, over-provisioned supervisor - one above the ideal idle ratio with low lag stayed pinned at its current task count.

**Root cause.** The idle projection was linear:

```rawIdle = 1.0 - busyFraction / taskRatio;   // taskRatio = proposed / current```

This assumes busy time is fully conserved when work moves onto fewer tasks, so a reasonable consolidation projects negative idle `(e.g. 1 − 0.6/0.5 =−0.2)`. That clamps to 0 (the worst point of the U-shaped idle cost) and turns an overrun into phantom virtual lag — pinning the task count even at ~0 real lag. In reality, busy grows sublinearly (an observed 2× consolidation raised busy ~1.25×, not 2×).

**Fix.** Redistribute busy sublinearly:
```
projectedBusy = busyFraction * (currentTaskCount / proposedTaskCount) ^ IDLE_SUBLINEARITY_EXPONENT;  // 0.32
rawIdle = 1.0 - projectedBusy;
```
`IDLE_SUBLINEARITY_EXPONENT = 0.32 (≈ log₂(1.25))` is a tuned constant based on careful testing and theoretical math application.

A healthy consolidation now lands near the ideal idle ratio instead of going negative, so the supervisor scales down; the exponent stays > 0, so extreme over-consolidation still diverges and is broken. 

**Validation (plots under hood)**

<details>
Optimal task count vs. observed poll-idle ratio, across realistic configs (rate = total cluster throughput, split per-task):

cost_based_scaledown_medium_7Mpm

Old version stays pinned at 128 until idle ~0.55, while new version consolidates from ~0.32.

cost_based_scaledown_large_30Mpm

Safe under load: new version consolidates earlier on the high-idle side, but at low idle both still jump to max — lag-driven scale-up is unaffected.

cost_based_v1_vs_v2_large_30Mpm_amp0 35

The existing version is flat (pinned at max by the phantom overrun); new version consolidates and holds more tasks as lag weight rises.

</details>

**Release note**

Fixed an issue where the cost-based supervisor autoscaler would not scale down an over-provisioned supervisor running above its ideal idle ratio with low lag.

  - [x] self-reviewed.
  - [x] added comments explaining the "why".
  - [x] added/updated unit tests.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
@@ -43,6 +43,12 @@ public class WeightedCostFunction
   static final double LAG_AMPLIFICATION_MULTIPLIER = 0.4;
 
   /**
+   * Exponent (&lt; 1) for sublinear busy redistribution in the idle projection: busy grows as
+   * {@code (currentTaskCount / proposedTaskCount)^EXPONENT}, not linearly. Calibrated as log2(1.25) ~= 0.32.
+   */
+  static final double IDLE_SUBLINEARITY_EXPONENT = 0.32;
+
+  /**
    * Minimum rate of processing for any task in records per second. This is used
    * as a placeholder if avg rate is not available to ensure that cost computations
    * do not return infinitely large lag recovery times.
@@ -109,10 +115,10 @@ public class WeightedCostFunction
       lagRecoveryTime = metrics.getAggregateLag() * amplification / (proposedTaskCount * adjustedProcessingRate);
     }
 
-    // Capacity-based idle prediction. When the proposed count would oversaturate the cluster
-    // (busy work exceeds available capacity), the unmet demand becomes a virtual lag-recovery
-    // time on the same axis as real lag — so the optimizer treats predicted saturation as
-    // predicted lag, not as "perfect utilization".
+    // Capacity-based idle prediction with sublinear busy redistribution. When the proposed count
+    // would oversaturate the cluster (busy work exceeds available capacity), the unmet demand
+    // becomes a virtual lag-recovery time on the same axis as real lag — so the optimizer treats
+    // predicted saturation as predicted lag, not as "perfect utilization".
     final double currentPollIdleRatio = metrics.getPollIdleRatio();
     final int currentTaskCount = metrics.getCurrentTaskCount();
     final double predictedIdleRatio;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by FrankChen021 — COMMENTED** [review]

---

File: indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java [hunk 1]
```diff
@@ -42,6 +42,12 @@ public class WeightedCostFunction
    */
   static final double LAG_AMPLIFICATION_MULTIPLIER = 0.4;
 
+  /**
+   * Exponent (&lt; 1) for sublinear busy redistribution in the idle projection: busy grows as
+   * {@code (currentTaskCount / proposedTaskCount)^EXPONENT}, not linearly. Calibrated as log2(1.25) ~= 0.32.
+   */
+  static final double IDLE_SUBLINEARITY_EXPONENT = 0.32;
```
kfaraz, line 51: Why this specific value?
I feel that these constants somehow make the behaviour of the auto-scaler more difficult to reason about as well as control via the weights.

We should probably not be trying to predict the idleness in this manner anyway.

Instead, as an alternative, we should probably consider using the `avg. processing rate * task count * idle ratio` as a measure of total work that needs to be done. Assuming that the processing rate remains the same, we can find the new idle ratio for the new task count.

kfaraz (commented):
Left a suggestion.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: > Why this specific value? 

This is already explained in PR description: we want to redistribute projected busyness in a sub-linear way, basically making it U-shaped (instead of V-shaped, like it was previously). It is also proved by theoretical math behind it (see plots).

Regarding an alternative -- I'd encourage you to elaborate more, please.

kfaraz: Ah, thanks!

I guess you mean,

> In reality, busy grows sublinearly (an observed 2× consolidation raised busy ~1.25×, not 2×).

Do you mean that when task count is halved, idle ratio increases by 1.25x?

kfaraz: For the alternative, I meant something along these lines

```
current workload
= total records per second
= current processing rate * current task count * idle ratio

assumptions:
target workload = current workload
predicted processing rate = current processing rate

predicted idle ratio
= target workload / (target task count * predicted processing rate)
= current workload / (target task count * current processing rate)
```

Let me know if that makes sense.

kfaraz (approved):
Approving this for now since the experiments suggest that this helps with idle prediction.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -42,6 +42,12 @@ public class WeightedCostFunction
    */
   static final double LAG_AMPLIFICATION_MULTIPLIER = 0.4;
 
+  /**
+   * Exponent (&lt; 1) for sublinear busy redistribution in the idle projection: busy grows as
+   * {@code (currentTaskCount / proposedTaskCount)^EXPONENT}, not linearly. Calibrated as log2(1.25) ~= 0.32.
```
kfaraz: Please clarify here that this value was determined empirically and it was seen that when task count is halved, idle ratio increases roughly by 1.25.

Fly-Style: Done

PR #19565, opened 2026-06-06
fix: empty loads for asymmetric cluster-group partial-load matchers

clintropolis:
### Description
This PR fixes a flaw with `ClusterGroupPartialLoadMatcher.match()` when it asymmetrically matches clustered segments within a core-partition group, resulting in only some of the core-partition segments in the group being loaded and being 'incomplete' in the timeline.

The fix is to allow `PartialLoadMatcher` to perform 'empty' loads, where if the rule matcher is processing a clustered segment but none of the cluster groups match. Coupled with this, `RunRules` guards against totally empty groups by using a new `EmptyDeferringHandler` wrapper `SegmentActionHandler` that buffers these empty requests and 'flushes' them at shard-group borders, only performing empty loads when some part of a core partition group was a non-empty match.

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/loading/PartialClusterGroupLoadSpec.java
+++ b/processing/src/main/java/org/apache/druid/segment/loading/PartialClusterGroupLoadSpec.java
@@ -25,7 +25,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Preconditions;
-import org.apache.druid.utils.CollectionUtils;
 
 import java.util.List;
 import java.util.Map;
@@ -72,10 +71,9 @@ public class PartialClusterGroupLoadSpec extends PartialLoadSpec
   )
   {
     super(delegate, fingerprint, jsonMapper);
-    Preconditions.checkArgument(
-        !CollectionUtils.isNullOrEmpty(clusterGroupIndices),
-        "clusterGroupIndices must not be null or empty"
-    );
+    // An empty index list is used when a cluster-group matcher applies to a (clustered) segment but no configured
+    // pattern matches its cluster groups. The historical-side partial loader honors this by loading nothing.
+    Preconditions.checkNotNull(clusterGroupIndices, "clusterGroupIndices");
     this.clusterGroupIndices = List.copyOf(clusterGroupIndices);
   }
 
--- a/processing/src/main/java/org/apache/druid/segment/loading/PartialClusterGroupLoadSpec.java
+++ b/processing/src/main/java/org/apache/druid/segment/loading/PartialClusterGroupLoadSpec.java
@@ -33,7 +33,7 @@ import java.util.Objects;
 /**
  * A {@link PartialLoadSpec} that requests partial loading of a clustered segment's cluster groups. The base class
  * carries the common {@code fingerprint} and {@code delegate} wire fields; this subtype adds the resolved
- * {@code clusterGroupIndices} (positions into {@link org.apache.druid.timeline.ClusterGroupTuples#getTuples()}) that
+ * {@code clusterGroupIndices} (positions into {@link org.apache.druid.timeline.ClusterGroupTuples#tuples()}) that
  * the historical should range-read into the local segment.
  */
 @JsonTypeName(PartialClusterGroupLoadSpec.TYPE)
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/druid/server/coordinator/rules/ClusterGroupPartialLoadMatcher.java
```diff
@@ -59,10 +78,13 @@ public MatchResult match(DataSegment segment, Map<String, Object> baseLoadSpec)
...
     final List<Integer> resolved = resolveClusterGroupIndices(segment);
-    if (resolved.isEmpty()) {
+    final ShardSpec shardSpec = segment.getShardSpec();
+    if (resolved.isEmpty() && shardSpec.getPartitionNum() >= shardSpec.getNumCorePartitions()) {
+      // No patterns match and this segment isn't part of a core partition group, so no need for an empty load. Fall
+      // through to the cannot-match handling.
       return null;
     }
-    final String fingerprint = computeFingerprint(resolved);
+    final String fingerprint = resolved.isEmpty() ? EMPTY_LOAD_FINGERPRINT : computeFingerprint(resolved);
```
FrankChen021, line 99: [P1] Empty matches bypass FALL_THROUGH for fully unmatched groups

Returning a non-null empty MatchResult for every core segment means PartialLoadRule.appliesTo returns true even when the rule is configured with onCannotMatch=FALL_THROUGH. If a core shard group has clusterGroups but none of the segments match the configured patterns, RunRules stops on this partial rule for every segment, then EmptyDeferringHandler discards all buffered empties because no positive was seen. Later fallback rules never run, so segments that previously fell through to a full-load rule can be left with no assignment. The group-level discard path needs to preserve the configured cannot-match behavior.

processing/src/main/java/org/apache/druid/segment/loading/PartialClusterGroupLoadSpec.java
```diff
@@ -72,10 +71,9 @@ public PartialClusterGroupLoadSpec(
   )
   {
     super(delegate, fingerprint, jsonMapper);
-    Preconditions.checkArgument(
-        !CollectionUtils.isNullOrEmpty(clusterGroupIndices),
-        "clusterGroupIndices must not be null or empty"
-    );
+    // An empty index list is used when a cluster-group matcher applies to a (clustered) segment but no configured
+    // pattern matches its cluster groups. The historical-side partial loader honors this by loading nothing.
+    Preconditions.checkNotNull(clusterGroupIndices, "clusterGroupIndices");
```
FrankChen021, line 76: [P2] Empty loads still delegate to a full segment load

The constructor now accepts an empty clusterGroupIndices list for the empty-load sentinel, but PartialClusterGroupLoadSpec still inherits PartialLoadSpec.loadSegment(), which delegates directly to the inner load spec. A historical receiving this supposedly empty request will download the full segment and announce a matching full-fallback profile, so sparse matches can full-load every empty sibling instead of loading nothing. This needs an explicit empty-load implementation or another guard before emitting empty cluster-group load specs.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

---

server/src/main/java/org/apache/druid/server/coordinator/rules/ClusterGroupPartialLoadMatcher.java
clintropolis: ah, so this is intentional behavior to not fall-through, however this comment made me realize that it was a bit too aggressive with the empty matches since it was applying to _any_ clustered segments.

I only intended for empty matches to occur when the matcher is 'compatible' with the segment clustering, e.g. there is some overlap between what clustering columns the matcher is looking at and what columns the segment is actually clustered on so that we don't incorrect do an empty load for segments that are clustered completely differently from what the matcher is looking for.

I've updated the check to ensure that empty matches only occur when things are compatible, which was the intended behavior

processing/src/main/java/org/apache/druid/segment/loading/PartialClusterGroupLoadSpec.java
clintropolis: yea, this will be the case until partial load stuff is fully merged

FrankChen021 (commented):
Handled follow-up for this PR. Reviewed 8 of 8 changed files. The compatibility update now returns null for incompatible clustering schemes, so the prior FALL_THROUGH concern is addressed. The remaining empty-load loader delegation point was acknowledged as expected until the broader partial-load implementation lands. No inline reply is needed.

---

capistrant (approved):

PR #12032, opened 2021-12-07
Migrate Druid HttpClient from Netty 3 to 4

xvrl:
- HttpResponse no longer includes request content, this changes some
  HttpReponseHandler implementations to no longer assume content is
  generated as part of the handleResponse call, and instead expect all
  request content as part of the handleChunk

- HttpResponseHandler changes uncovered a long-standing bug in
  AppendableByteArrayInputStream where calling exceptionCaught before
  appending any bytes would not throw the exception to the reader

- Netty HTTP decoder now stores decoder errors in the message instead of
  triggering exceptionCaught, so we need to check the status of the
  decoder.

- Netty Messages and underlying buffers are reference counted, so they
  need to be explicitly released after handling. As part of this
  ChannelBuffer became ByteBuf and are reference counted instead of
  garbage collected.

- To support back-pressure, Channels need to disable AUTO_READ and call
  read() explicitly when needed instead of using setReadable(true/false)

- ClientBootstrap no longer uses separate bosspool and workerpool sizes,
  instead relying on the workerpool.

- Netty HTTP codec now supports resuming handling of HTTP messages after
  handling CONNECT, removing the need to manually add the codec back to
  the pipeline.

- Remove deprecated Netty method calls and rename classes to align with
  Netty class name changes

- Remove deprecated use of ExpectedException in favor of
  Assert.assertThrows for all tests that had to be touched.

PR #19567, opened 2026-06-07
refactor: migrate Druid HTTP client from Netty 3 to 4

clintropolis:
### Description
Third time's a charm? Previous attempts #12032 and #14479 afaict both got stuck trying to resolve failures in the integration tests (the old docker based integration tests). I started fresh for this attempt, but credit to prior attempts which were used as reference for comparison, and our embedded-tests which made things easier/quicker to test/debug than our old integration test framework.

#### Allocator and direct memory
The HTTP client previously used Netty 3's `HeapChannelBufferFactory` which is unpooled heap buffers. Netty 4.2's default is `AdaptiveByteBufAllocator` (pooled internally via `AdaptivePoolingAllocator`, direct by default). This is a substantive change in memory characteristics:

- **Direct memory**: the HTTP client now consumes some direct memory. Operators
  running close to their `-XX:MaxDirectMemorySize` budget may need to bump it.
- **Pooled vs unpooled**: buffers are now pooled. Resident memory may show
  higher than under Netty 3 even when idle, normal pooled-allocator behavior.

This PR is using Netty 4.2's default allocator (`adaptive`) as our default, rather than pinning to the `PooledByteBufAllocator` that was the Netty 4 default until 4.2 (or the unpooled heap behavior of Netty 3). Since Druid operators are coming from heap-unpooled (Netty 3), neither Netty 4 choice represents a "known good baseline"; both are new characteristics for this workload. However, using pooled off-heap buffers of some kind should offer a performance improvement so it seems worth the risk.

A few new Druid configs are wired up to control some of this stuff to control the http clients that use netty:

|Property|Description|Default|
|`druid.global.http.connectTimeout`|Connect timeout for the HTTP client used for most direct RPC between Druid services. This covers, among other things, Overlord-to-task and supervisor-to-task calls in the indexing service, Coordinator lookup management, dynamic config sync between services, MSQ tasks reading from data servers, and general Coordinator/Overlord/Broker service clients. Does not affect Broker-to-Historical query dispatch (see `druid.broker.http.connectTimeout`) or request forwarding (see `clientConnectTimeout`).|`PT10S`|
|`druid.global.http.allocator`|Netty ByteBuf allocator used by the direct-RPC HTTP client. Accepts `adaptive` (adaptive between pooled and unpooled based on load), `pooled` (always pooled), or `unpooled` (always freshly allocated).|`adaptive`|

and broker -> historical connections:

|Property|Description|Default|
|`druid.broker.http.connectTimeout`|Connect timeout for the HTTP client the Broker uses to dispatch queries to Historical and real-time processes.|`PT10S`|
|`druid.broker.http.allocator`|Netty ByteBuf allocator used by the direct-RPC HTTP client. Accepts `adaptive` (adaptive between pooled and unpooled based on load), `pooled` (always pooled), or `unpooled` (always freshly allocated).|`adaptive`|

There are some other possibly useful knobs that I did not wire into explicit druid configs, but can be set directly with netty configs: 

  | Property | Default | Purpose |
  |---|---|---|
  | `io.netty.noPreferDirect` | `false` | `true` ⇒ pooled but heap rather than direct |
  | `io.netty.allocator.maxOrder` | 9 | Smaller value ⇒ smaller chunk size, lower memory floor |
  | `io.netty.maxDirectMemory` | derived from `-XX:MaxDirectMemorySize` | Netty's direct-memory cap |
  | `io.netty.leakDetection.level` | `simple` | `disabled` / `simple` / `advanced` / `paranoid` |

#### Other stuff
`InputStreamHolder.fromChannelBuffer(ByteBuf, long)` now copies bytes synchronously instead of wrapping a `ByteBufInputStream` over a pooled buffer that `SimpleChannelInboundHandler` releases after `channelRead0` returns.


processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
```diff
@@ -131,29 +123,30 @@ public <Intermediate, Final> ListenableFuture<Final> go(
...
+      channel.config().setAutoRead(true);
     }
     final String urlFile = StringUtils.nullToEmptyNonDruidDataString(url.getFile());
-    final HttpRequest httpRequest = new DefaultHttpRequest(
+    final FullHttpRequest httpRequest = new DefaultFullHttpRequest(
         HttpVersion.HTTP_1_1,
         method,
-        urlFile.isEmpty() ? "/" : urlFile
+        urlFile.isEmpty() ? "/" : urlFile,
+        request.hasContent() ? request.getContent() : Unpooled.EMPTY_BUFFER
```
FrankChen021: [P1] Do not hand the request-owned ByteBuf to Netty

This passes the Request's stored ByteBuf directly into a Netty 4 FullHttpRequest. Netty's HTTP encoder releases FullHttpMessages after encoding, and may also advance the content reader index, so after the first send request.getContent() can be released or consumed. Kerberos' 401 retry path calls request.copy() after the first response, which can then fail or resend an empty body for authenticated POSTs. Please build the outbound request from a retained duplicate/copy, or make Request store bytes and create a fresh ByteBuf per send.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Reviewed 164 of 164 changed files.

---

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Reviewed 52 of 52 changed files.

- processing/src/main/java/org/apache/druid/segment/QueryableIndexCursorFactory.java:91 - [P1] Projection matching bypasses clustered column capabilities. The clustered index path now depends on clustered-aware column capabilities, but makeCursorHolder tries index.getProjection(spec) before checking index.getClusteredBaseSummary(). For clustered segments, the top-level physical index does not expose the clustered logical columns, so aggregate projection matching can treat clustered filter/group columns as absent from the base table and select a projection that does not carry them. That returns a normal QueryableIndexCursorHolder over the projection instead of entering clustered dispatch, which can silently produce empty or incorrect aggregate results. Projection selection needs to be clustered-aware, or aggregate projection matching should be deferred/skipped until the clustered schema can validate the query.

---

FrankChen021 (commented):
I reviewed the updated Netty 3 to 4 HTTP client migration paths, including async lifecycle, streaming response handling, request reuse, TLS/proxy setup, headers, timeout behavior, and Druid call-site compatibility. I found no new correctness issues to report.

Reviewed 164 of 164 changed files.

---

processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
```diff
@@ -131,29 +123,32 @@ public <Intermediate, Final> ListenableFuture<Final> go(
...
         HttpVersion.HTTP_1_1,
         method,
-        urlFile.isEmpty() ? "/" : urlFile
+        urlFile.isEmpty() ? "/" : urlFile,
+        request.hasContent() ? request.getContent().retainedDuplicate() : Unpooled.EMPTY_BUFFER
     );
 
-    if (!headers.containsKey(HttpHeaders.Names.HOST)) {
-      httpRequest.headers().add(HttpHeaders.Names.HOST, getHost(url));
+    if (!headers.containsKey(HttpHeaderNames.HOST.toString())) {
```
FrankChen021: [P2] Check default headers case-insensitively

Request stores headers in a case-sensitive Multimap, but the migration now checks only the lower-case Netty 4 names before adding defaults. A caller that already set "Host" or "Accept-Encoding" using conventional HTTP casing is not detected here, so the client adds a default and then copies the caller header later, producing duplicate Host headers or multiple Accept-Encoding values. This regresses the old Netty 3 behavior for callers overriding those headers and can break virtual-host/proxy requests. Please make these checks case-insensitive or normalize Request header names before adding defaults.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Reviewed 100 of 100 changed files.

I found one correctness issue in the Netty 4 migration: caller-supplied Host or Accept-Encoding headers with conventional casing can be duplicated because the new default-header checks are case-sensitive.

---

FrankChen021 (commented):
I reviewed the changes since the previous reviewed commit and found no new issues. The previously posted case-sensitive default-header finding remains unchanged and is not duplicated here.

Reviewed 100 of 100 changed files.

---

FrankChen021 (commented):
I reviewed the updated HTTP client migration for correctness, edge cases, concurrency, resource lifecycle, and integration risks; no new issues found. The previously reported case-sensitive default-header issue is fixed with case-insensitive Netty header checks and regression coverage.

Reviewed 100 of 100 changed files.

---

FrankChen021 (commented):
I have reviewed the updated head for correctness, edge cases, concurrency, and integration risks; no new issues found. Range-diff confirmed all six PR commits are unchanged, and the prior case-insensitive header fix remains intact.

Reviewed 100 of 100 changed files.

---

FrankChen021:
Cross-reference for the overlapping Netty 3 → 4 migration: [full implementation comparison with #19754](https://github.com/apache/druid/pull/19754#issuecomment-5138134196).

Short conclusion: use #19567 as the consolidation base, then selectively port #19754's failed-decoder propagation and bounded proxy `CONNECT` hardening. The report covers the dependency-line difference, explicit-read semantics, buffer ownership, timeouts/lifecycle, header handling, tests, and current CI state.

Disclosure: this comment was generated by GPT-5.6-Sol.

embedded-tests/pom.xml
```diff
@@ -912,6 +912,14 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
           <excludedGroups>docker-test,perf</excludedGroups>
+          <!-- Enable Netty 4 PARANOID leak detection in the embedded tests because this is where
+               real cross-node HTTP traffic exercises the migrated HttpClient code paths (chunked
+               responses, long-poll syncs, TLS, ByteBuf-bearing handlers). The Stage 1 CI workers
+               that happen to run embedded-tests classes pay a small overhead per class; in exchange
+               we catch ref-count regressions immediately rather than as a flaky off-heap leak. -->
+          <systemPropertyVariables>
+            <io.netty.leakDetection.level>PARANOID</io.netty.leakDetection.level>
```
gianm, line 921: I think that `PARANOID` with the builtin leak detector logs but doesn't throw an exception, which probably won't be noticed. Maybe there's some way to install a custom detector.

processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
```diff
@@ -254,37 +254,46 @@ public void abort()
 
                 assert currentChunkNum == 0;
                 possiblySuspendReads(response);
+                // Servers (notably Druid's QueryResource on the historical) often flush the response
+                // status line + headers in one TCP write and the chunked body in subsequent writes.
+                // Netty 4's AUTO_READ chains reads automatically after a complete read cycle, but in
+                // practice the body chunks for these split writes are not picked up without an explicit
+                // ctx.read(), leaving the caller blocked on an InputStream that never gets bytes.
+                // Force a read here to drain the body chunks for these multi-write responses.
+                ctx.read();
```
gianm: How was the stuff in the comment determined? Is there a test that would fail without this line?

Also: do we really want to do the `ctx.read()` if the `possiblySuspendReads` call above decided to suspend reads? Seems like no.

Also: the PR description has the related claim that "the root problem with prior attempts that resulted in the observed hanging is that Netty 4's `setAutoRead(true)` is just a config flag". It doesn't seem true though. Netty 4's `DefaultChannelConfig#setAutoRead` has some logic that looks like:

```
if (autoRead && !oldAutoRead) {
  channel.read();
}
```

I'm asking this stuff because I am hoping to better understand why this PR isn't succumbing to the hanging problems that plagued the earlier ones.

FrankChen021 (commented):
I have reviewed the updated head for correctness, edge cases, concurrency, and integration risks; no new issues found.

Reviewed 164 of 164 changed files.

The prior case-insensitive header issue remains fixed.



---

processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
clintropolis: oops, forgot to update here, but to update in this PR's comments I dug into it a bit, and it turned out that this extra `ctx.read()` wasn't actually needed, and indeed wack if it should be suspended. Rather claude likely stuck it in there while fixing one of the other bugs which was causing a hang in the tests and it didn't hurt anything tests covered and so got counted as an important part of fixing things. I probably should have checked a bit closer before trusting things 😅

Removed this.

processing/src/main/java/org/apache/druid/java/util/http/client/pool/ChannelResourceFactory.java
```diff
@@ -123,75 +122,80 @@ public ChannelFuture generate(final String hostname)
...
-                )
-            );
-          }
+              }
+          );
+          // Bound the wait for the CONNECT response. Bootstrap's CONNECT_TIMEOUT_MILLIS only covers the TCP
+          // handshake, and the per-request readTimeout in NettyHttpClient does not apply yet because this
+          // channel is not in the pool. A proxy that accepts TCP but never sends a CONNECT response would
+          // otherwise hang overallConnectPromise (and every future waiter on the pool) indefinitely.
+          final ScheduledFuture<?> connectTimeoutTask = channel.eventLoop().schedule(
```
FrankChen021, line 169: [P2] Close the channel when CONNECT times out

When this newly scheduled timeout wins, it only fails `overallConnectPromise`; it never closes `channel`. `NettyHttpClient.go()` then returns the failed future to `ResourcePool`, whose `giveBack` retains it and only calls `factory.close` on a later `take()` for that key or pool shutdown. Repeated silent or unresponsive proxy connections for different pool keys can therefore leave open sockets until shutdown. Close the channel on this failure path and keep the completion handling idempotent.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

The incremental review focused on the new proxy CONNECT timeout and its interaction with pooled channel lifecycle.

Reviewed 168 of 168 changed files.

The timeout now bounds silent proxy waits, but its failure path needs to close the underlying channel before returning the failed future to the pool.



---

processing/src/main/java/org/apache/druid/java/util/http/client/response/StringFullResponseHandler.java
```diff
@@ -57,7 +57,7 @@ public ClientResponse<StringFullResponseHolder> handleChunk(
       return ClientResponse.finished(null);
     }
 
-    holder.addChunk(chunk.getContent().toString(charset));
+    holder.addChunk(chunk.content().toString(charset));
```
gianm: This shouldn't be doing `toString` on each chunk individually, because nothing guarantees that chunks are cut on UTF-8 character boundaries. (If a chunk is cut in the middle of a character, it will become garbled.) I'm not sure how this worked in netty3 but it would be good to fix it now.

Similar comment for `StatusResponseHandler`.

In both cases we should be able to use `CharsetDecoder` to do a correct streaming decode.

processing/src/main/java/org/apache/druid/java/util/http/client/response/StringFullResponseHolder.java
```diff
@@ -19,21 +19,23 @@
...
 import java.nio.charset.Charset;
 
 public class StringFullResponseHolder extends FullResponseHolder<String>
 {
   private final StringBuilder builder;
 
+  @SuppressWarnings("unused")
   public StringFullResponseHolder(
       HttpResponse response,
       Charset charset
```
gianm, line 34: If we don't need it, it's better to remove `charset` rather than adding `@SuppressWarnings("unused")`.

processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
```diff
@@ -125,47 +117,61 @@ public <Intermediate, Final> ListenableFuture<Final> go(
     final Channel channel;
     final String hostKey = getPoolKey(url);
     final ResourceContainer<ChannelFuture> channelResourceContainer = pool.take(hostKey);
```
gianm, line 120: There's a lot of code between here and the error-handling `try` below, where an exception would cause the `channelResourceContainer` to never be returned via `returnResource`. If enough leak then connections will start to hang.

Netty 3 and 4 have some differences here that make this interesting: Netty 4 does additional validations in `new DefaultFullHttpRequest` that Netty 3 does not do in `new DefaultHttpRequest`, so the odds of seeing an exception in this span of code are higher with Netty 4.

processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
```diff
@@ -370,13 +393,7 @@ private void handleExceptionAndCloseChannel(final Throwable t, final boolean clo
             }
 
             if (!retVal.isDone()) {
-              if (t instanceof ReadTimeoutException) {
-                // ReadTimeoutException thrown by ReadTimeoutHandler is a singleton with a misleading stack trace.
```
gianm, line 374: Netty 4 still seems to throw `ReadTimeoutException.INSTANCE` so I would keep this block. It was there to make the timeout exception more meaningful.

processing/src/main/java/org/apache/druid/java/util/http/client/HttpClientInit.java
```diff
@@ -125,37 +92,23 @@ public static SSLContext sslContextWithTrustedKeyStore(final String keyStorePath
...
 
-    final ClientBootstrap bootstrap = new ClientBootstrap(new NioClientSocketChannelFactory(bossPool, workerPool));
-
-    bootstrap.setOption("keepAlive", true);
-    bootstrap.setPipelineFactory(new HttpClientPipelineFactory());
+    final Bootstrap bootstrap = new Bootstrap();
+    bootstrap.group(eventLoopGroup)
+             .channel(NioSocketChannel.class)
+             .handler(new HttpClientPipelineFactory())
+             .option(ChannelOption.SO_KEEPALIVE, true);
```
gianm: The default `connectTimeoutMillis` (which we aren't setting) changed from 10s to 30s from Netty 3 -> 4. We should set it explicitly to 10s and consider making it configurable.

processing/src/main/java/org/apache/druid/java/util/http/client/HttpClientInit.java
```diff
@@ -51,41 +47,13 @@ public class HttpClientInit
...
-            {
-              timer.stop();
-            }
-          }
-      );
       return lifecycle.addMaybeStartManagedInstance(
           new NettyHttpClient(
               new ResourcePool<>(
                   new ChannelResourceFactory(
-                      createBootstrap(lifecycle, timer, config.getBossPoolSize(), config.getWorkerPoolSize()),
```
gianm, line 85: `HttpClientConfig.getBossPoolSize` is unused after this change, please remove it.

processing/src/main/java/org/apache/druid/java/util/http/client/netty/HttpClientPipelineFactory.java
```diff
@@ -19,24 +19,21 @@
...
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelPipeline;
+import io.netty.handler.codec.http.HttpClientCodec;
+import io.netty.handler.codec.http.HttpContentDecompressor;
 
 /**
  */
-public class HttpClientPipelineFactory implements ChannelPipelineFactory
+public class HttpClientPipelineFactory extends ChannelInitializer<Channel>
```
gianm: Rename to `HttpClientChannelInitializer`

processing/src/main/java/org/apache/druid/java/util/http/client/pool/ChannelResourceFactory.java
```diff
@@ -208,62 +222,40 @@ public void operationComplete(ChannelFuture f2)
...
-                        StringUtils.format("Failed to connect to host[%s]", hostname),
-                        f.getCause()
-                    )
-                );
-              }
+      final SslHandler sslHandler = new SslHandler(sslEngine);
+      sslHandler.setHandshakeTimeoutMillis(sslHandshakeTimeout);
+
+      final ChannelPromise handshakePromise = connectFuture.channel().newPromise();
+      connectFuture.channel().pipeline().addLast(ERROR_HANDLER_NAME, new ConnectionErrorHandler(handshakePromise));
```
gianm: Apparently in Netty 4, channel bootstrap from the `ChannelInitializer` happens in the event loop rather than the calling thread, so to be properly ordered this handler addition has to happen in the event loop too. You should be able to make that happen by adding it deferredly in `channel.eventLoop()`. Make sure removal of the handler is done on the event loop too.

processing/src/main/java/org/apache/druid/java/util/http/client/pool/ChannelResourceFactory.java
gianm: It looks like the fix to this issue is there now, a few lines down. However it possibly needs to call `overallConnectPromise.tryFailure` rather than `overallConnectPromise.setFailure` to make multiple failure signals really OK.

processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
```diff
@@ -371,9 +401,13 @@ private void handleExceptionAndCloseChannel(final Throwable t, final boolean clo
...
             if (!retVal.isDone()) {
               if (t instanceof ReadTimeoutException) {
-                // ReadTimeoutException thrown by ReadTimeoutHandler is a singleton with a misleading stack trace.
-                // No point including it: instead, we replace it with a fresh exception.
-                retVal.setException(new ReadTimeoutException(StringUtils.format("[%s] Read timed out", requestDesc)));
+                // ReadTimeoutHandler fires ReadTimeoutException.INSTANCE, a shared singleton whose stack
+                // trace was captured once at class load and points at whichever code first touched Netty's
+                // class initializer. Emit a fresh instance so the stack trace reflects this actual timeout.
+                if (log.isDebugEnabled()) {
+                  log.debug("[%s] Read timed out", requestDesc);
```
gianm: There is no need to use `log.isDebugEnabled()` here, it's only valuable if it saves computation. The same check is in `log.debug`.

processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
```diff
@@ -371,9 +401,13 @@ private void handleExceptionAndCloseChannel(final Throwable t, final boolean clo
...
+                }
+                retVal.setException(new ReadTimeoutException());
```
gianm: Hmm. It seems like `ReadTimeoutException` in Netty 4 doesn't have a stack trace (it inherits a no-op `fillInStackTrace` from `TimeoutException`). I guess it means we should either stick with the builtin `ReadTimeoutException` from `ReadTimeoutHandler` or create a new one that has a good message.

If you go with a solution that uses `new ReadTimeoutException()` or `ReadTimeoutException.INSTANCE` (i.e. no nice message) then at least update the `HttpResponseHandler` in `DirectDruidClient` to interpolate `e` rather than `e.getMessage()` into `"Query[%s] url[%s] failed with exception msg [%s]"`. Maybe only if `e.getMessage()` is null, just to avoid chaanging too many messages. That way intead of seeing `exception msg [null]` for a timeout it'd be `exception msg [io.netty.handler.timeout.ReadTimeoutException: null]` (I think).

processing/src/main/java/org/apache/druid/java/util/http/client/pool/ChannelResourceFactory.java
```diff
@@ -208,62 +222,46 @@ public void operationComplete(ChannelFuture f2)
...
+      sslChannel.eventLoop().execute(
+          () -> sslChannel.pipeline().addLast(ERROR_HANDLER_NAME, new ConnectionErrorHandler(handshakePromise))
       );
+      connectFuture.addListener((ChannelFuture f) -> {
+        if (f.isSuccess()) {
+          final ChannelPipeline pipeline = f.channel().pipeline();
+          pipeline.addFirst("ssl", sslHandler);
+          sslHandler.handshakeFuture().addListener(f2 -> {
+            if (f2.isSuccess()) {
+              handshakePromise.setSuccess();
```
gianm: This `setSuccess` and the following two `setFailure` should I think all be `trySuccess` / `tryFailure` to avoid blowing up on an already-completed promise.

processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
```diff
@@ -125,49 +119,28 @@ public <Intermediate, Final> ListenableFuture<Final> go(
     final Channel channel;
     final String hostKey = getPoolKey(url);
     final ResourceContainer<ChannelFuture> channelResourceContainer = pool.take(hostKey);
+    // ResourcePool.take is documented as blocking, but a null return is possible on pool exhaustion or shutdown;
+    // fail fast rather than NPE on the awaitUninterruptibly() below.
+    if (channelResourceContainer == null) {
+      return Futures.immediateFailedFuture(
+          new ChannelException(StringUtils.format("Connection pool exhausted or timed out for host[%s]", hostKey))
```
gianm: This message is inaccurate: `pool.take(hostKey)` is only null if `pool` is closed. (Which shouldn't happen unless the client itself is closed, and would lead to an `ERROR` being logged anyway.)

Although, I think the container can have a null inside, i.e., `channelResourceContainer.get()` can be null in some other cases, such as an interrupt happening on the thread. It may be worth checking for that (one line down).

processing/src/main/java/org/apache/druid/java/util/http/client/pool/ChannelResourceFactory.java
FrankChen021: ## Follow-up assessment

The current head addresses the follow-up concern. The timeout callback and the proxy CONNECT response, write, and failure paths all call `overallConnectPromise.tryFailure`, so competing failure signals are idempotent; the promise listener also cancels the timeout and closes the channel on failure. I reviewed 175 of 175 changed files and found no additional current-head issues.



---

FrankChen021 (commented):
## 🟢 Approval recommended

No actionable issues found in this review. The updated Netty 4 client, proxy CONNECT lifecycle, response streaming/decoding, and Druid RPC paths are consistent at the current head; the previously reported failed-proxy-channel leak is fixed.

Reviewed 175 of 175 changed files.

Validation: `git diff --check` passed. Narrow `javap` checks against Netty 4.2.17.Final confirmed the migrated legacy header/status aliases and the explicit decompressor constructor's default behavior. No builds or broad tests were run.

---

This is an automated review by Codex GPT-5.6-Luna

processing/src/main/java/org/apache/druid/java/util/http/client/HttpClientInit.java
```diff
@@ -125,37 +93,24 @@ public static SSLContext sslContextWithTrustedKeyStore(final String keyStorePath
     }
   }
 
-  private static ClientBootstrap createBootstrap(Lifecycle lifecycle, Timer timer, int bossPoolSize, int workerPoolSize)
+  private static Bootstrap createBootstrap(Lifecycle lifecycle, int workerPoolSize, Duration connectTimeout)
```
gianm: Let's explicitly configure an allocator here, just so we know what we're getting.

processing/src/main/java/org/apache/druid/java/util/http/client/NettyHttpClient.java
```diff
@@ -125,49 +118,43 @@ public <Intermediate, Final> ListenableFuture<Final> go(
...
+    // just fail the request.
+    final ChannelFuture rawChannelFuture = channelResourceContainer.get();
+    if (rawChannelFuture == null) {
+      return Futures.immediateFailedFuture(
+          new ChannelException(
+              StringUtils.format("Interrupted or pool closed while waiting for a channel to host[%s]", hostKey)
+          )
+      );
+    }
+    final ChannelFuture channelFuture = rawChannelFuture.awaitUninterruptibly();
```
gianm: If this fails, or if `channel.config().setAutoRead(true)` below fails, then the resource is leaked. I'm not sure if this can happen but it would be nice to guard against it.

server/src/main/java/org/apache/druid/client/DirectDruidClient.java
```diff
@@ -423,11 +425,16 @@ public ClientResponse<InputStream> done(ClientResponse<InputStream> clientRespon
         @Override
         public void exceptionCaught(final ClientResponse<InputStream> clientResponse, final Throwable e)
         {
+          // Fall back to Throwable.toString() when the exception carries no message, so a timeout
+          // (Netty's ReadTimeoutException is a stackless, messageless singleton) does not render as
+          // "exception msg [null]" but as "exception msg [io.netty.handler.timeout.ReadTimeoutException]"
+          // instead. Behaviour for exceptions that DO have a message is unchanged.
+          final String exceptionDetail = e.getMessage() != null ? e.getMessage() : e.toString();
```
gianm, line 432: Please do a similar thing in `DataServerResponseHandler`.

docs/configuration/index.md
```diff
@@ -2331,4 +2333,5 @@ Supported query contexts:
 |`druid.router.http.numMaxThreads`|Maximum number of worker threads to handle HTTP requests and responses|`(number of cores) * 3 / 2 + 1`|
 |`druid.router.http.numRequestsQueued`|Maximum number of requests that may be queued to a destination|`1024`|
 |`druid.router.http.requestBuffersize`|Size of the content buffer for receiving requests. These buffers are only used for active connections that have requests with bodies that will not fit within the header buffer|`8 * 1024`|
-|`druid.router.http.clientConnectTimeout`|The timeout (in milliseconds) for establishing client connections.|500|
+|`druid.router.http.clientConnectTimeout`|Connect timeout (in milliseconds) for the HTTP client the Router uses to forward incoming queries and management requests to Brokers and other Druid services. Does not affect the Router's direct RPC connections; see `connectTimeout` for those.|500|
+|`druid.router.http.connectTimeout`|Connect timeout for the HTTP client the Router uses for direct RPC to Brokers (for example, service-status polling).|`PT10S`|
```
gianm: I believe this config may be fake (because we don't use the NettyHttpClient for Router -> Broker communications). Please double check it.

docs/configuration/index.md
```diff
@@ -1813,7 +1814,8 @@ client has the following configuration options.
...
 |`druid.broker.http.maxQueuedBytes`|Maximum number of bytes queued per query before exerting [backpressure](../operations/basic-cluster-tuning.md#broker-backpressure) on channels to the data servers.<br /><br />Similar to `druid.server.http.maxScatterGatherBytes`, except that `maxQueuedBytes` triggers [backpressure](../operations/basic-cluster-tuning.md#broker-backpressure) instead of query failure. Set to zero to disable. You can override this setting by using the [`maxQueuedBytes` query context parameter](../querying/query-context-reference.md). Druid supports [human-readable](human-readable-byte.md) format. |25 MB or 2% of maximum Broker heap size, whichever is greater.|
 |`druid.broker.http.numMaxThreads`|`Maximum number of I/O worker threads|(number of cores) * 3 / 2 + 1`|
-|`druid.broker.http.clientConnectTimeout`|The timeout (in milliseconds) for establishing client connections.|500|
+|`druid.broker.http.clientConnectTimeout`|Connect timeout (in milliseconds) for the Broker's request-forwarding HTTP client. Does not affect the Broker's direct connections to data servers; see `connectTimeout` for those.|500|
```
gianm: I don't think the Broker has a request-forwarding HTTP client? Please double check this.

docs/configuration/index.md
```diff
@@ -678,7 +678,8 @@ All Druid components can communicate with each other over HTTP.
 |`druid.global.http.readTimeout`|The timeout for data reads.|`PT15M`|
 |`druid.global.http.unusedConnectionTimeout`|The timeout for idle connections in connection pool. The connection in the pool will be closed after this timeout and a new one will be established. This timeout should be less than `druid.global.http.readTimeout`. Set this timeout = ~90% of `druid.global.http.readTimeout`|`PT4M`|
 |`druid.global.http.numMaxThreads`|Maximum number of I/O worker threads|`(number of cores) * 3 / 2 + 1`|
-|`druid.global.http.clientConnectTimeout`|The timeout (in milliseconds) for establishing client connections.|500|
+|`druid.global.http.clientConnectTimeout`|Connect timeout (in milliseconds) for the HTTP client used to forward requests between Druid services (for example, when the Router proxies queries to Brokers, or when management API calls are forwarded to the Coordinator or Overlord). Does not affect direct RPC connections between services; see `connectTimeout` for those.|500|
```
gianm: I believe this overstates what this is used for. Isn't the `druid.global.http` client used for other stuff beyond forwarding requests? Like, I thought it was also used for most internal RPCs generally via the `@EscalatedGlobal` client.

server/src/main/java/org/apache/druid/guice/http/HttpClientModule.java
```diff
@@ -159,4 +165,35 @@ public <Intermediate, Final> ListenableFuture<Final> go(
       }
     }
   }
+
+  /**
+   * Single shared adaptive allocator, instantiated eagerly. {@link AdaptiveByteBufAllocator} has no {@code DEFAULT}
+   * singleton of its own (the inherited {@code ByteBufAllocator.DEFAULT} constant routes through
+   * {@code -Dio.netty.allocator.type}). Instantiate one here so the "adaptive" config value really is adaptive.
+   */
+  private static final ByteBufAllocator ADAPTIVE_ALLOCATOR_INSTANCE = new AdaptiveByteBufAllocator();
```
gianm: Strange place to put a `private static final`. Generally the static fields should be at the top of the file.

processing/src/main/java/org/apache/druid/java/util/http/client/HttpClientConfig.java
```diff
@@ -76,6 +78,10 @@ public String getEncodingString()
   // aggressive default so a slow or dead peer surfaces sooner. Callers can override via the builder.
   private static final Duration DEFAULT_CONNECT_TIMEOUT_DURATION = new Period("PT10S").toStandardDuration();
 
+  // Set an explicit allocator so callers know exactly what they're getting rather than depending on whatever
+  // ByteBufAllocator.DEFAULT happens to resolve to
+  private static final ByteBufAllocator DEFAULT_BYTE_BUF_ALLOCATOR = new AdaptiveByteBufAllocator();
```
gianm: Why do we need both this one and the one in `HttpClientModule`?

server/src/main/java/org/apache/druid/discovery/DataServerResponseHandler.java
```diff
@@ -211,10 +211,15 @@ public ClientResponse<InputStream> done(ClientResponse<InputStream> clientRespon
   @Override
   public void exceptionCaught(ClientResponse<InputStream> clientResponse, Throwable e)
   {
+    // Fall back to Throwable.toString() when the exception carries no message, so a timeout
+    // (Netty's ReadTimeoutException is a stackless, messageless singleton) does not render as
+    // "exception msg [null]" but as "exception msg [io.netty.handler.timeout.ReadTimeoutException]"
+    // instead. Behaviour for exceptions that DO have a message is unchanged.
```
gianm: Behavior (spelling). Although it'd be better to cut this last sentence entirely. Generally comments should avoid referring to prior states of the same code block.

gianm (approved):

FrankChen021 (commented):
## 🟢 Approval recommended

Updated-head assessment for [apache/druid#19567](https://github.com/apache/druid/pull/19567): no new actionable findings.

Reviewed the incremental diff from `e45b396c7dadf216a3980b69a12d310638d5dd49` to `55ae9486f34124610137c0c0251dea8999864f4f`, then the full merge-base diff and surrounding code across all 175 changed files. Rechecked the earlier case-insensitive `Host`/`Accept-Encoding` default-header issue and failed-proxy-channel lifecycle issue; both remain fixed. The Netty 4 response framing and buffer-copying paths, proxy CONNECT timeout and cleanup, TLS setup, allocator and connect-timeout wiring, and Druid RPC call sites are consistent at the current head.

Reviewed 175 of 175 changed files.

Validation: `git diff --check fb8bb5fafd08ea7ec2a5e83704e4c287ef5bede9..HEAD` passed. Narrow `javap` checks against local Netty 4.2.17.Final confirmed the status aliases, `HttpClientCodec`/response framing behavior, leak-detector constructor/API, and `HttpContentDecompressor(false, 0)` default. No builds or broad tests were run.



---

PR #19568, opened 2026-06-08
feat: introduce stopGracefullyOnNewSpec for `Supervisor`s

Fly-Style:
Generally, any supervisor should not terminate their tasks during spec update, but historically it happened that stream supervisors became bound to that logic. This PR introduces the correct behaviour - it supervisor stops gracefully, but stream supervisors do not via `stopGracefullyOnNewSpec()` mechanism.
It was done to enhance UX for existing and future supervisor kinds, and there are plans to enhance stream supervisors to stop gracefully too in a follow-up PR.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
@@ -281,7 +281,7 @@ public class SupervisorManager implements SupervisorStatsProvider
     Preconditions.checkState(started, "SupervisorManager not started");
     List<ListenableFuture<Void>> stopFutures = new ArrayList<>();
     synchronized (lock) {
-      log.info("Stopping [%d] supervisors", supervisors.keySet().size());
+      log.info("Stopping [%d] supervisors", supervisors.size());
       for (String id : supervisors.keySet()) {
         try {
           stopFutures.add(supervisors.get(id).lhs.stopAsync());
--- a/indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
@@ -377,7 +377,7 @@ public class SupervisorManager implements SupervisorStatsProvider
    * Resets a supervisor to the latest stream offsets and starts a bounded backfill supervisor to
    * process the skipped range from the previously checkpointed offsets up to the latest offsets.
    *
-   * @param id               supervisor ID
+   * @param id                supervisor ID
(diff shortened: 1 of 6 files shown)


jtuglu1:
This is related to: https://github.com/apache/druid/pull/19541

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
```diff
@@ -615,12 +625,10 @@ private SupervisorSpec possiblyStopAndRemoveSupervisorInternal(String id, boolea
...
     if (writeTombstone) {
-      metadataSupervisorManager.insert(
-          id,
-          new NoopSupervisorSpec(null, pair.rhs.getDataSources())
-      ); // where NoopSupervisorSpec is a tombstone
+      // NoopSupervisorSpec is a tombstone
+      metadataSupervisorManager.insert(id, new NoopSupervisorSpec(null, pair.rhs.getDataSources()));
     }
-    pair.lhs.stop(true);
+    pair.lhs.stop(pair.lhs.stopGracefullyOnNewSpec());
```
FrankChen021: [P2] Do not use the new-spec stop policy for terminate

This helper is also called by stopAndRemoveSupervisor(id) with writeTombstone=true, so explicit supervisor terminate now passes stopGracefullyOnNewSpec() into stop(). The new default returns false, and an implementation may return false specifically to avoid killing tasks during spec replacement; on the public terminate path that means stop(false), which the Supervisor contract says leaves running tasks as-is even though the terminate API promises to stop associated tasks and publish segments. Please keep tombstone/terminate removal graceful, or split the replacement policy from the removal policy.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

indexing-service/src/main/java/org/apache/druid/indexing/overlord/supervisor/SupervisorManager.java
Fly-Style: Addressed, thanks!

FrankChen021 (commented):
Follow-up checked: the current head restores graceful termination by calling `stop(writeTombstone || stopGracefullyOnNewSpec())`, so the original inline concern is addressed.

---

kfaraz:
@Fly-Style , I believe that for now we should close off this PR in favor of #19541 since it proposes a broader solution which is based on the `SupervisorSpec` (which is the entity being modified) rather than the `Supervisor`.

In the future, we might want to decide whether to restart or not based on the current running state of the supervisor tasks (and not just the update made to the spec). When that is needed, we can add the method proposed in this PR.

cc: @jtuglu1

Fly-Style:
Agree @kfaraz

Issue #12929, opened 2022-08-20
Add Kafka hash partition type to improve query performance


599166320:
### Motivation

We use druid to store a large amount of monitoring data. By default, the Kafka index service is used to ingest data, and the query performance is very poor.

After analysis, by default, the monitoring data will be randomly written into each partition of Kafka. The data will be randomly consumed by Kafka peon, and the index will be established to generate segments.

When querying, the broker can only filter segments by time, and cannot further filter according to query conditions. It needs to scan a large number of real-time nodes, historical nodes, and a large number of segments, resulting in poor performance.

### Proposed changes

1. Kafka real-time index service supports a new partition type, adding `KafkaPartitionNumberedShardSpec` class, which extents from `NumberedShardSpec` ,Override the `possibleInDomain` function in the `KafkaPartitionNumberedShardSpec` class, implement hash filtering, and add the following core fields:`type = "kafka_partition", kafkapartitionids, partitiondimensions`.

2. Modify the `KafkaIndexTask` class to support the `KafkaPartitionNumberedShardSpec` type in `kafkaindextask.newdriver`.

3. Add the `partitionfunction` field in the configuration of Kafka ingression spec, and configure all fields that need to be hashed in the `partitionfunction`.

4.  Add a simple hash code to the data production side:

```
int p = hash(dim1,dim2,...)
ProducerRecord<byte[], byte[]>(topic, p,event)
```

### Rationale
When writing data, partition the data, which is beneficial to compression and sorting. During the query, the scanning range is further trimmed by the user-defined filter conditions to optimize the performance of concurrent queries.

### Performance comparison

Before optimization
![FQrxkz62dz](https://user-images.githubusercontent.com/3204398/185734896-9555b9ae-5e08-44b6-8524-3da42cf35936.jpg)

After optimization
![A93J4pk1Qm](https://user-images.githubusercontent.com/3204398/185734907-d3c111d0-fb45-4bd7-9442-16e6e786e001.jpg)

gianm:
This kind of approach is definitely great for performance. The main reason we haven't done it yet is that it can be too fiddly for most users. A key problem is: what happens if the producer doesn't correctly do the hash codes? What if we want to add partitions to the kafka topic? We would want to build something that the average user could use easily, and correctly.

[...]

PR #19571, opened 2026-06-09
feat: Prunable shard specs for streaming published segments

abhishekrb19:
Relates to https://github.com/apache/druid/issues/12929

  Streaming-published segments currently have numbered shard specs, which aren't prunable by design. Compaction must reindex the data with range or hashed partition strategy once the data is handed off — even if the topic is partitioned, which
  is easy to do with multiple supervisors. For multi-tenant datasources this means every tenant-filtered query hits every recent segment regardless of the partitioning strategy for numbered shard specs.

  This PR lets streaming tasks record, per published segment, the distinct values observed for a configured set of dimensions, and declare them on a new shard spec so the broker can prune near-realtime data without waiting for compaction to
  reindex handed off segments. Concurrent compaction cannot always keep up with the incoming data and additionally the compaction process itself takes time to reindex; so the benefits of range or hashed shard specs may not be fully realized for
  however long it takes to reindex (30-45 minutes in our case), and doesn't help with high concurrent query workloads that are only querying more recent data.

  So this PR adds a way to publish prunable shards right off the bat when they're handed off by streaming tasks, if configured. This functionality is opt-in, Kafka-only, and disabled by default.

  **Design**

  - `DimensionValueSetShardSpec` (type: `"dim_value_set"`) — extends `NumberedShardSpec` (behaves as a normal append segment) plus a partitionDimensionValues map (dimension → observed values). possibleInDomain() prunes a segment when the query
  constrains a declared dimension and none of its values intersect the domain; a dimension not in partitionDimensionValues is never pruned on. Set-based (not min/max), so it prunes precisely for sparse values and tolerates overlapping value
  sets across tasks/restarts.
  - Because partition values are observed at ingestion rather than hardcoded, incorrect or abruptly-changing partitioning never breaks correctness — at worst it yields non-prunable shard specs (similar to the default numbered shard) or bloated
  shard specs (this caveat can be addressed with a guard rail noted below).
  - Ingestion (SeekableStreamIndexTaskRunner) — when `partitionDimensions` is set, the task accumulates observed values per segment and stamps each at publish.

  **Configuration**

  New optional `tuningConfig` field `streamingPartitionsSpec` on the Kafka supervisor/task (default null), with a `partitionDimensions` list. When unset, behavior is unchanged. Documented in docs/ingestion/kafka-ingestion.md.

  **Compatibility**

  Backward-compatible and opt-in. But dim_value_set is a new core ShardSpec type with no defaultImpl fallback, so it is not forward-compatible: upgrade all services before enabling streamingPartitionsSpec, and note that once dim_value_set
  segments are published, downgrade isn't supported until they're compacted away.

  **Results**

  Tested in a cluster, where I saw up to ~40% reduction in segment scans on the historicals for a few low to medium cardinal partition dimensions. In a follow-up, I want to extend this to also prune tasks, for reduced peon buffers and better
  query performance at the task layer.

  **Caveats**

  There's currently no limit on the number of observed values stamped into a segment's partitionDimensionValues. It may make sense, in a follow-up, to add a configurable guardrail that falls back to NumberedShardSpec when the count exceeds a
  threshold, so shard specs don't get bloated.

  Release note

  Kafka ingestion can now publish segments that the broker prunes at query time, without waiting for compaction. Set tuningConfig.streamingPartitionsSpec.partitionDimensions to a list of low-to-medium cardinality dimensions; each task records
  the distinct values it observes per dimension and stamps them onto a new dim_value_set shard spec. Queries that filter on a declared dimension then skip segments whose values can't match. The feature is opt-in, Kafka-only, and disabled by
  default; when unset, behavior is unchanged.

  Compatibility: `dim_value_set` is a new core shard spec type with no fallback, so it is not forward-compatible. Upgrade all services before enabling streamingPartitionsSpec. Once dim_value_set segments are published, downgrade is unsupported
  until they are compacted away or streamingPartitionsSpec is removed.

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/docs/ingestion/kafka-ingestion.md
+++ b/docs/ingestion/kafka-ingestion.md
@@ -265,7 +265,7 @@ The following example shows a supervisor spec with idle configuration enabled:
 
 #### Streaming partitions spec
 
-When you set `streamingPartitionsSpec.partitionDimensions` in the tuning config, the supervisor tracks the distinct values observed for each listed dimension during ingestion. At segment publish time, each segment is annotated with only the values it actually ingested. The broker then uses these annotations to skip segments at query time when the query filter doesn't intersect the segment's declared values.
+When you set `streamingPartitionsSpec.partitionDimensions` in the tuning config, the supervisor tracks the distinct values observed for each listed dimension during ingestion. At segment publish time, each segment is annotated with only the values it actually ingested by publishing it with a `dim_value_set` shard spec, which records the observed values per tracked dimension. The broker then uses these annotations to skip segments at query time when the query filter doesn't intersect the segment's declared values.
 
 This enables segment pruning for streaming-ingested data without waiting for compaction to produce hash or range-partitioned segments. The `partitionDimensions` should be kept in sync with the compaction config's `partitionDimensions` for the same datasource.
 
--- a/docs/ingestion/kafka-ingestion.md
+++ b/docs/ingestion/kafka-ingestion.md
@@ -126,7 +126,6 @@ For configuration properties shared across all streaming ingestion methods, refe
 |`pollTimeout`|Long|The length of time to wait for the Kafka consumer to poll records, in milliseconds.|No|100|
 |`useEarliestOffset`|Boolean|If a supervisor is managing a datasource for the first time, it obtains a set of starting offsets from Kafka. This flag determines whether the supervisor retrieves the earliest or latest offsets in Kafka. Under normal circumstances, subsequent tasks start from where the previous segments ended so this flag is only used on the first run.|No|`false`|
 |`idleConfig`|Object|Defines how and when the Kafka supervisor can become idle. See [Idle configuration](#idle-configuration) for more details.|No|null|
-|`partitionFilterDimensions`|List of String|Dimensions to track for query-time segment pruning. See [Partition filter dimensions](#partition-filter-dimensions) for details.|No|null|
 
 #### Ingest from multiple topics
 
@@ -264,17 +263,18 @@ The following example shows a supervisor spec with idle configuration enabled:
 ```
 </details>
 
-#### Partition filter dimensions
+#### Streaming partitions spec
 
-When you set `partitionFilterDimensions` in the IO config, the supervisor tracks the distinct values observed for each listed dimension during ingestion. At segment publish time, each segment is annotated with only the values it actually ingested. The broker then uses these annotations to skip segments at query time when the query filter doesn't intersect the segment's declared values.
+When you set `streamingPartitionsSpec.partitionDimensions` in the tuning config, the supervisor tracks the distinct values observed for each listed dimension during ingestion. At segment publish time, each segment is annotated with only the values it actually ingested. The broker then uses these annotations to skip segments at query time when the query filter doesn't intersect the segment's declared values.
 
-This enables segment pruning for streaming-ingested data without waiting for compaction to produce hash or range-partitioned segments.
+This enables segment pruning for streaming-ingested data without waiting for compaction to produce hash or range-partitioned segments. The `partitionDimensions` should be kept in sync with the compaction config's `partitionDimensions` for the same datasource.
 
 **Usage guidelines:**
 
 - Use only low-to-medium cardinality dimensions (for example, `tenant_id`, `region`, `environment`). High-cardinality dimensions bloat segment metadata with no pruning benefit.
 - Most effective when Kafka partitions are keyed by the tracked dimension (for example, using tenant ID as the message key). Each task naturally sees a subset of values, and segments get tight filter annotations.
 - Also works with multiple supervisors reading from separate topics into one datasource.
+- Use a range or hashed compaction `partitionsSpec`, not the dynamic strategy: dynamic compaction does not partition by dimension, so it cannot preserve pruning after compaction.
 - After compaction, the `StreamRangeShardSpec` annotations are replaced by the compaction output's shard spec (hash or range partitioning), which provides its own pruning.
 
 The following example configures a supervisor to track the `tenant` dimension:
@@ -295,8 +295,11 @@ The following example configures a supervisor to track the `tenant` dimension:
       "consumerProperties": {"bootstrap.servers": "localhost:9092"},
       "inputFormat": {"type": "json"},
       "taskCount": 4,
-      "taskDuration": "PT1H",
-      "partitionFilterDimensions": ["tenant"]
+      "taskDuration": "PT1H"
+    },
+    "tuningConfig": {
+      "type": "kafka",
+      "streamingPartitionsSpec": {"partitionDimensions": ["tenant"]}
     }
   }
 }
@@ -481,6 +484,7 @@ For configuration properties shared across all streaming ingestion methods, refe
 |Property|Type|Description|Required|Default|
 |--------|----|-----------|--------|-------|
 |`numPersistThreads`|Integer|The number of threads to use to create and persist incremental segments on the disk. Higher ingestion data throughput results in a larger number of incremental segments, causing significant CPU time to be spent on the creation of the incremental segments on the disk. For datasources with number of columns running into hundreds or thousands, creation of the incremental segments may take up significant time, in the order of multiple seconds. In both of these scenarios, ingestion can stall or pause frequently, causing it to fall behind. You can use additional threads to parallelize the segment creation without blocking ingestion as long as there are sufficient CPU resources available.|No|1|
+|`streamingPartitionsSpec`|Object|Configures query-time segment pruning for streaming-ingested segments. Contains a single property, `partitionDimensions` (List of String), the dimensions whose observed values each segment records so the broker can skip segments that can't match a query filter. See [Streaming partitions spec](#streaming-partitions-spec) for details.|No|null|
 
 ## Deployment notes on Kafka partitions and Druid segments
 
--- a/docs/ingestion/kafka-ingestion.md
+++ b/docs/ingestion/kafka-ingestion.md
@@ -126,6 +126,7 @@ For configuration properties shared across all streaming ingestion methods, refe
 |`pollTimeout`|Long|The length of time to wait for the Kafka consumer to poll records, in milliseconds.|No|100|
 |`useEarliestOffset`|Boolean|If a supervisor is managing a datasource for the first time, it obtains a set of starting offsets from Kafka. This flag determines whether the supervisor retrieves the earliest or latest offsets in Kafka. Under normal circumstances, subsequent tasks start from where the previous segments ended so this flag is only used on the first run.|No|`false`|
 |`idleConfig`|Object|Defines how and when the Kafka supervisor can become idle. See [Idle configuration](#idle-configuration) for more details.|No|null|
+|`partitionFilterDimensions`|List of String|Dimensions to track for query-time segment pruning. See [Partition filter dimensions](#partition-filter-dimensions) for details.|No|null|
 
 #### Ingest from multiple topics
 
@@ -262,6 +263,46 @@ The following example shows a supervisor spec with idle configuration enabled:
 }
 ```
 </details>
+
+#### Partition filter dimensions
+
+When you set `partitionFilterDimensions` in the IO config, the supervisor tracks the distinct values observed for each listed dimension during ingestion. At segment publish time, each segment is annotated with only the values it actually ingested. The broker then uses these annotations to skip segments at query time when the query filter doesn't intersect the segment's declared values.
+
+This enables segment pruning for streaming-ingested data without waiting for compaction to produce hash or range-partitioned segments.
+
+**Usage guidelines:**
+
+- Use only low-to-medium cardinality dimensions (for example, `tenant_id`, `region`, `environment`). High-cardinality dimensions bloat segment metadata with no pruning benefit.
+- Most effective when Kafka partitions are keyed by the tracked dimension (for example, using tenant ID as the message key). Each task naturally sees a subset of values, and segments get tight filter annotations.
+- Also works with multiple supervisors reading from separate topics into one datasource.
+- After compaction, the `StreamRangeShardSpec` annotations are replaced by the compaction output's shard spec (hash or range partitioning), which provides its own pruning.
+
+The following example configures a supervisor to track the `tenant` dimension:
+
+```json
+{
+  "type": "kafka",
+  "spec": {
+    "dataSchema": {
+      "dataSource": "multi_tenant_events",
+      "timestampSpec": {"column": "timestamp", "format": "iso"},
+      "dimensionsSpec": {"dimensions": ["tenant", "region", "event_type"]},
+      "granularitySpec": {"type": "uniform", "segmentGranularity": "HOUR", "queryGranularity": "NONE"}
+    },
+    "ioConfig": {
+      "type": "kafka",
+      "topic": "events",
+      "consumerProperties": {"bootstrap.servers": "localhost:9092"},
+      "inputFormat": {"type": "json"},
+      "taskCount": 4,
+      "taskDuration": "PT1H",
+      "partitionFilterDimensions": ["tenant"]
+    }
+  }
+}
+```
+
+With this configuration, a query like `SELECT * FROM multi_tenant_events WHERE tenant = 'acme'` skips segments that contain no rows for `acme`, reducing the number of segments scanned.
 
 #### Data format
 
--- a/docs/ingestion/kafka-ingestion.md
+++ b/docs/ingestion/kafka-ingestion.md
@@ -271,6 +271,7 @@ This enables segment pruning for streaming-ingested data without waiting for com
 
 **Usage guidelines:**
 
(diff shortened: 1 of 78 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -999,15 +1057,86 @@ private void checkPublishAndHandoffFailure() throws ExecutionException, Interrup
...
+    );
+  }
+
   private void publishAndRegisterHandoff(SequenceMetadata<PartitionIdType, SequenceOffsetType> sequenceMetadata)
   {
     log.debug("Publishing segments for sequence [%s].", sequenceMetadata);
 
+    // annotateSegmentWithPartitionFilters is a no-op (returns the segment unchanged) when partition filters are not
+    // configured, so it is always safe to apply here.
+    final java.util.function.Function<Set<DataSegment>, Set<DataSegment>> shardSpecAnnotator =
```
aho135: nit: can add Function to imports

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -1042,6 +1171,12 @@ public void onSuccess(SegmentsAndCommitMetadata publishedSegmentsAndCommitMetada
             );
             log.infoSegments(publishedSegmentsAndCommitMetadata.getSegments(), "Published segments");
 
+            for (DataSegment segment : publishedSegmentsAndCommitMetadata.getSegments()) {
+              observedDimensionValuesBySegment.remove(
+                  SegmentIdWithShardSpec.fromDataSegment(segment).toString()
```
aho135: Should we also clean up `restartSpannedSegments.remove` here?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -999,15 +1057,86 @@ private void checkPublishAndHandoffFailure() throws ExecutionException, Interrup
...
+    final String lookupKey = SegmentIdWithShardSpec.fromDataSegment(s).toString();
+    if (restartSpannedSegments.contains(lookupKey)) {
+      return s;
+    }
+    final Map<String, Set<String>> segObserved = observedDimensionValuesBySegment.get(lookupKey);
+    if (segObserved == null || segObserved.isEmpty()) {
+      return s;
+    }
+    final Map<String, List<String>> snapshotFilters = new HashMap<>();
+    for (String dim : filterDims) {
```
aho135:
```
for (String dim : filterDims) {
    segObserved.computeIfPresent(dim, (k, vals) -> {
      synchronized (vals) {
        if (!vals.isEmpty()) {
          snapshotFilters.put(dim, new ArrayList<>(vals));
        }
      }
      return vals;  // Return unchanged - we're just reading
    });
  }
```

aho135: Claude recommendation for race condition

docs/ingestion/kafka-ingestion.md
```diff
@@ -126,6 +126,7 @@ For configuration properties shared across all streaming ingestion methods, refe
 |`pollTimeout`|Long|The length of time to wait for the Kafka consumer to poll records, in milliseconds.|No|100|
 |`useEarliestOffset`|Boolean|If a supervisor is managing a datasource for the first time, it obtains a set of starting offsets from Kafka. This flag determines whether the supervisor retrieves the earliest or latest offsets in Kafka. Under normal circumstances, subsequent tasks start from where the previous segments ended so this flag is only used on the first run.|No|`false`|
 |`idleConfig`|Object|Defines how and when the Kafka supervisor can become idle. See [Idle configuration](#idle-configuration) for more details.|No|null|
+|`partitionFilterDimensions`|List of String|Dimensions to track for query-time segment pruning. See [Partition filter dimensions](#partition-filter-dimensions) for details.|No|null|
```
aho135: What do you think about naming this `partitionDimensions` to align with the compaction config? That may make it more clear that those values should be in sync

docs/ingestion/kafka-ingestion.md
```diff
@@ -263,6 +264,46 @@ The following example shows a supervisor spec with idle configuration enabled:
...
+
+This enables segment pruning for streaming-ingested data without waiting for compaction to produce hash or range-partitioned segments.
+
+**Usage guidelines:**
+
+- Use only low-to-medium cardinality dimensions (for example, `tenant_id`, `region`, `environment`). High-cardinality dimensions bloat segment metadata with no pruning benefit.
+- Most effective when Kafka partitions are keyed by the tracked dimension (for example, using tenant ID as the message key). Each task naturally sees a subset of values, and segments get tight filter annotations.
+- Also works with multiple supervisors reading from separate topics into one datasource.
+- After compaction, the `StreamRangeShardSpec` annotations are replaced by the compaction output's shard spec (hash or range partitioning), which provides its own pruning.
```
aho135: Maybe worth mentioning that when using `partitionFilterDimensions`, dynamic compaction strategy should not be used

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -999,15 +1057,86 @@ private void checkPublishAndHandoffFailure() throws ExecutionException, Interrup
...
+   */
+  @VisibleForTesting
+  DataSegment annotateSegmentWithPartitionFilters(DataSegment s)
+  {
+    final List<String> filterDims = ioConfig.getPartitionFilterDimensions();
+    if (CollectionUtils.isNullOrEmpty(filterDims)) {
+      return s;
+    }
+    final String lookupKey = SegmentIdWithShardSpec.fromDataSegment(s).toString();
+    if (restartSpannedSegments.contains(lookupKey)) {
```
FrankChen021: P1 Mixed shard specs can fail publish after restart

Restart-spanned segments return unchanged as NumberedShardSpec, while new same-interval segments in the same publish batch can be annotated as StreamRangeShardSpec. TransactionalSegmentPublisher then runs SegmentPublisherHelper.annotateShardSpec, which rejects mixed shard-spec classes per interval, so a restarted task can fail publish/handoff. Make the fallback interval-wide, or stamp restored segments with a non-pruning stream_range spec.

extensions-core/kafka-indexing-service/src/main/java/org/apache/druid/indexing/kafka/supervisor/KafkaSupervisorIOConfig.java
```diff
@@ -55,6 +57,62 @@ public class KafkaSupervisorIOConfig extends SeekableStreamSupervisorIOConfig
...
+        lateMessageRejectionPeriod,
+        earlyMessageRejectionPeriod,
+        lateMessageRejectionStartDateTime,
+        configOverrides,
+        idleConfig,
+        stopTaskCount,
+        emitTimeLagMetrics,
+        serverPriorityToReplicas,
+        boundedStreamConfig,
+        null
```
FrankChen021: P3 Backfill specs drop partitionFilterDimensions

This compatibility constructor always forwards null for partitionFilterDimensions. KafkaSupervisorSpec.createBackfillSpec still uses this overload when deriving bounded backfill specs, so a supervisor configured with partitionFilterDimensions silently creates backfill tasks without the pruning annotations. Pass the existing dimension list through for backfill specs.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 1 |
| Total | 2 |

I found two issues: restart-spanning segments can mix shard spec classes within one publish interval and fail publishing, and Kafka backfill specs drop partitionFilterDimensions.

---

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

Reviewed 19 of 19 changed files.

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -999,15 +1060,80 @@ private void checkPublishAndHandoffFailure() throws ExecutionException, Interrup
...
+        if (vals == null) {
+          continue;
+        }
+        // vals is a synchronized set written by the run loop; copy it under its monitor to iterate safely.
+        final List<String> snapshot;
+        synchronized (vals) {
+          if (vals.isEmpty()) {
+            continue;
+          }
+          snapshot = new ArrayList<>(vals);
```
FrankChen021, line 1111: [P3] Sort observed partition values before publishing

The list stored in partitionDimensionValues is created directly from a HashSet, so its order is unspecified. The embedded test already asserts a concrete order for these values, and equivalent published segment metadata can vary by JVM or run even when the value set is identical. Sort the snapshot deterministically, with explicit null handling, before putting it into the shard spec.

Reviewed 19 of 19 changed files.

abhishekrb19: Updated

docs/ingestion/kafka-ingestion.md
abhishekrb19: Yes, good call - thanks for the suggestion! I also moved this into the `tuningConfig` for consistency. It's nested under a `streamingPartitionsSpec` container to avoid naming ambiguity and so we can extend it with additional properties and/or add new types in the future if needed.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
abhishekrb19: Thanks, this is fixed with some test coverage added.

docs/ingestion/kafka-ingestion.md
```diff
@@ -263,6 +263,50 @@ The following example shows a supervisor spec with idle configuration enabled:
...
+      "type": "kafka",
+      "topic": "events",
+      "consumerProperties": {"bootstrap.servers": "localhost:9092"},
+      "inputFormat": {"type": "json"},
+      "taskCount": 4,
+      "taskDuration": "PT1H"
+    },
+    "tuningConfig": {
+      "type": "kafka",
+      "streamingPartitionsSpec": {"partitionDimensions": ["tenant"]}
```
aho135, line 303: Nice! Makes sense to have this in tuningConfig, and in the future we can add cardinality guardrails into `streamingPartitionsSpec` as well

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaIndexTaskTuningConfigTest.java
```diff
@@ -128,6 +130,36 @@ public void testSerdeWithNonDefaults() throws Exception
...
+        TuningConfig.class
+    );
+
+    Assert.assertEquals(
+        new StreamingPartitionsSpec(List.of("tenant", "region")),
+        config.getStreamingPartitionsSpec()
+    );
+    Assert.assertEquals(List.of("tenant", "region"), config.getStreamingPartitionsSpec().getPartitionDimensions());
+  }
+
```
aho135, line 154: Might be worth having a test case which covers edge case configuration for partitionDimensions (e.g. null values, empty string, integers)

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -999,15 +1062,82 @@ private void checkPublishAndHandoffFailure() throws ExecutionException, Interrup
...
+        }
+        // Sort for deterministic published metadata; null (missing value) sorts first.
+        snapshot.sort(Comparator.nullsFirst(Comparator.naturalOrder()));
+        snapshotFilters.put(dim, snapshot);
+      }
+    }
```
aho135, line 1117: For a follow up, it may be useful to emit metric here for the cardinality of `observedPartitionDimValuesBySegment.get(segmentId)`

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -693,6 +735,27 @@ public void run()
...
+                      // Empty getDimension result means a null/missing value; record null so IS NULL is not pruned
+                      // (distinct from "", which getDimension returns as ["" ]).
+                      final List<String> dimValues = row.getDimension(dim);
+                      if (dimValues == null || dimValues.isEmpty()) {
+                        dimSet.add(null);
+                      } else {
+                        dimSet.addAll(dimValues);
+                      }
+                    }
+                  }
```
clintropolis: it would be cool if there was an interface to abstract this 'do something to collect information for the shard spec' operation (thinking about other useful kinds of shard-specs like this, such as more classic ranges where we just collect min/max values, or if we could build bloom filters or something for higher cardinality columns).

I guess this would need streaming partition spec to either be an interface or have a 'processor' thingy that could provide the appropriate interface implementation.

This comment is not necessarily a blocker for this PR, just thinking how we could make something like this a bit more flexible/less specific for future improvements.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorDriver.java
```diff
@@ -279,6 +279,16 @@ public ListenableFuture<SegmentsAndCommitMetadata> publish(
...
   )
+  {
+    return publish(publisher, committer, sequenceNames, java.util.function.Function.identity());
+  }
+
+  public ListenableFuture<SegmentsAndCommitMetadata> publish(
+      final TransactionalSegmentPublisher publisher,
+      final Committer committer,
+      final Collection<String> sequenceNames,
+      final java.util.function.Function<Set<DataSegment>, Set<DataSegment>> segmentAnnotateFunction
```
aho135, line 290: `java.util.function.Function` can also be imported since it's used twice

aho135 (approved):
Left a few minor comments, but otherwise LGTM. Thanks @abhishekrb19!

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderatorDriver.java
abhishekrb19: Yeah, I tried using Guava's `Function` here (which is used in the class below), but it doesn't seem to fit, hence the inline `java.util.function.Function`

FrankChen021 (commented):
Reviewed 5 of 20 changed files (6 files total inspected) for this reply-needed follow-up.

Confirmed the two existing review follow-ups are addressed in current code: observed partition values are now sorted deterministically with explicit null handling, and restart-spanned segments now use empty-filter `StreamRangeShardSpec` behavior with relevant task-runner and publisher-helper test coverage.

No inline reply looks necessary.

---

extensions-core/kafka-indexing-service/src/test/java/org/apache/druid/indexing/kafka/KafkaIndexTaskTuningConfigTest.java
abhishekrb19: >Might be worth having a test case which covers edge case configuration for partitionDimensions (e.g. null values, empty string

I've added some tests in this unit test and there's some existing end-to-end coverage in the embedded ones too for null, empty cases

Re numeric types, they're currently unsupported for the `range` shard spec as well, given how the filters work - some details here: https://github.com/apache/druid/pull/19415. For now it's just documented for this new shard spec, but it would be nice to expand that functionality generally so all shard specs benefit from it; if that's not trivial, we could block them at creation time at least for this new shard spec in the future

abhishekrb19:
Thanks for the reviews @aho135 @FrankChen021! I also went ahead and rename the shard spec type from `stream_range ` to `dim_value_set` as it better aligns with what it does.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
abhishekrb19: Yeah agreed, I think that should be doable. We could spec out the specifics of the interface contract sometime. Thanks for taking a look!

PR #19577, opened 2026-06-11
refactor: split out RowCountInspector, PhysicalSegmentColumnInspector from PhysicalSegmentInspector

clintropolis:
### Description
Small change to split out a new  `RowCountInspector` and `PhysicalSegmentColumnInspector` from `PhysicalSegmentInspector`, the latter of which takes all of the column specific methods (and the `ColumnInspector` part). These changes should simplify some things for the partial load functionality I have been working on, since the methods I have moved to `PhysicalSegmentColumnInspector` currently at least need the actual column present to correctly implement, however the remaining methods in `PhysicalSegmentInspector` can be answered purely with metadata. `PhysicalSegmentColumnInspector` is only used by `SegmentAnalyzer` for segment metadata queries for column analysis, row count is driven through `RowCountInspector`, and `Metadata` retrieval is now done directly from `Segment.as`. The existing interface is marked deprecated, and still wired up for QueryableIndex/IncrementalIndex, but is no longer used internally other than a test.

### DIFF
--- a/processing/src/main/java/org/apache/druid/segment/PhysicalSegmentColumnInspector.java
+++ b/processing/src/main/java/org/apache/druid/segment/PhysicalSegmentColumnInspector.java
@@ -38,7 +38,7 @@ public interface PhysicalSegmentColumnInspector extends ColumnInspector
   Comparable getMinValue(String column);
 
   /**
-   * Returns the minimum value of the provided column, if known through an index, dictionary, or cache. Returns null
+   * Returns the maximum value of the provided column, if known through an index, dictionary, or cache. Returns null
    * if not known. Does not scan the column to find the maximum value.
    */
   @Nullable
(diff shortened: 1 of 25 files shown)


FrankChen021 (commented):

---

processing/src/main/java/org/apache/druid/segment/PhysicalSegmentColumnInspector.java
```diff
@@ -0,0 +1,52 @@
...
+{
+  /**
+   * Returns the minimum value of the provided column, if known through an index, dictionary, or cache. Returns null
+   * if not known. Does not scan the column to find the minimum value.
+   */
+  @Nullable
+  Comparable getMinValue(String column);
+
+  /**
+   * Returns the minimum value of the provided column, if known through an index, dictionary, or cache. Returns null
```
capistrant:
```suggestion
   * Returns the maximum value of the provided column, if known through an index, dictionary, or cache. Returns null
```

capistrant (approved):

PR #19578, opened 2026-06-13
fix: zstd + v10 segment metadata decompression for large metadata

clintropolis:
### Description
Fixes an issue I ran into with v10 segment segment metadata decompression when using zstd (which uses heap buffers) when the target size is larger than the pooled buffer size, resulting in failure on decompression. The fix is to check if the target size is larger than the pooled size, and if so do a non-pooled allocation instead of exploding. LZF also had this problem, so fixed it too. Added test would run into this problem before the fixes in this PR.


capistrant (approved):
thanks for the fix!

FrankChen021 (commented):

---

PR #19579, opened 2026-06-13
feat: clustered segments pt.2 (write side support)

clintropolis:
### Description
Follow-up to #19460, this PR introduces the writer side stuff so that the segments can actually be created.

As part of this I'm experimenting with a new V10 oriented way to express `DataSchema`, where the ingest spec looks a lot more like how the V10 segment metadata is organized, as 1 or more projections. Using the wikipedia example, a clustered segment ingest spec looks something like this:

```
    "dataSchema": {
      "dataSource": "wikipedia_clustered",
      "segmentGranularitySpec": {
        "segmentGranularity": "day"
      },
      "timestampSpec": {
        "column": "timestamp",
        "format": "iso"
      },
      "baseTable": {
        "type": "clusteredValueGroups",
        "virtualColumns": [
          {
            "type": "expression",
            "name": "__virtualGranularity",
            "expression": "timestamp_floor(__time,'PT1H')",
            "outputType": "LONG"
          }
        ],
        "columns": [
          "channel",
          {
            "type": "long",
            "name": "__time"
          }
          "page",
          "comment"
...
        ],
        "clusteringColumns":[
          "channel"
        ]
      }
    }
```

`__time` must be specified in the `columns` list to indicate where the time ordering occurs (clustered segments always order by the clustering columns first and `__time` clustering is not yet supported). Like projections, it requires expressing the query granularity as a virtual column transformation of `__time` (though a bit more fragile than expressions since it requires the virtual column be named `__virtualGranularity` right now, thinking about changing this to 'resolve' it like we do for projections, or something else, not sure yet). To support this, when using `baseTable` there is also a `segmentGranularity` field which splits out the segment granularity parts of the existing `GranularitySpec`. To be less disruptive for now, these can be computed into a `GranularitySpec`, but over time i'd like to migrate to this model of expressing the schema.

Includes the fix part o #19578 since this is the branch where I ran into that problem.

changes:
* adds `BaseTableProjectionSpec` interface to capture the operator facing shape of V10 base table schemas
* adds `ClusteredValueGroupsBaseTableSchema` implementation for ingesting clustered segments
* adds `DataSchema.baseTable`, a `BaseTableProjectionSpec` which when set puts the DataSchema into a new mode where the majority of the schema is defined via the baseTable, rejecting other top level fields
* adds `DataSchema.segmentGranularity` to use when `baseTable` is set, which captures the segment granularity and intervals (query granularity is defined by the baseTable)
* adds `AdaptedBaseTableProjectionSpec` implementation for converting classic `DataSchema` fields to a `BaseTableProjectionSpec`
* adds `OnHeapClusteredBaseTable`, `OnHeapClusterGroup` used by `OnHeapIncrementalIndex` to build clustered segments
* adds `IndexMergerV10.makeClusteredIndexFiles` which merges and builds clustered v10 segments
* `Sink`/`BatchAppenderator`/`StreamAppenderator` wiring for clustered segments so that the cluster group tuples appear on the `DataSegment`
* known issues: unbounded, no aggregate projections, no compaction support, no time ordered cursor support


processing/src/main/java/org/apache/druid/data/input/impl/ClusteredValueGroupsBaseTableProjectionSpec.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+    this.clusteringColumns = Collections.unmodifiableList(new ArrayList<>(clusteringColumns));
+    this.virtualColumns = virtualColumns == null ? VirtualColumns.EMPTY : virtualColumns;
+    this.nonClusteringDimensions = nonClusteringDimensions == null
+                                   ? Collections.emptyList()
+                                   : Collections.unmodifiableList(new ArrayList<>(nonClusteringDimensions));
+    this.metrics = metrics == null ? new AggregatorFactory[0] : metrics;
+    validateNoOverlap(this.clusteringColumns, this.nonClusteringDimensions);
+    this.dimensionsSpec = computeDimensionsSpec(this.clusteringColumns, this.nonClusteringDimensions);
+    this.ordering = declaredOrdering != null
```
FrankChen021: [P2] Declared ordering can be advertised without being honored

The spec accepts any declaredOrdering here, but the clustered write path does not use it to build the actual per-group row comparator: OnHeapClusterGroup derives ordering from timePosition/DimensionsSpec, while IndexMergerV10 emits groups in ascending tuple order and then stores firstSchema.getOrdering() in metadata. A spec such as [tenant ASC, page ASC, __time ASC], or any DESC order, can therefore be persisted with advertised ordering that the data does not actually satisfy. Query engines may trust CursorHolder/DataSegment ordering and skip sorting, producing misordered results. Please reject unsupported ordering or wire the declared order through ingestion and merge.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Reviewed 53 of 53 changed files.

---

processing/src/main/java/org/apache/druid/data/input/impl/ClusteredValueGroupsBaseTableProjectionSpec.java
clintropolis: ah this is fair, i should probably make it purely computed for now like `AggregateProjectionSpec` since its a bit of work to actually wire this up to honor it

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Reviewed 32 of 55 changed files.

---

processing/src/main/java/org/apache/druid/data/input/impl/ClusteredValueGroupsBaseTableProjectionSpec.java
```diff
@@ -0,0 +1,345 @@
...
+      if (!seen.add(d.getName())) {
+        throw InvalidInput.exception("columns contains duplicate name [%s]", d.getName());
+      }
+    }
+
+    int timeIndex = -1;
+    boolean bothPresent = false;
+    for (int i = 0; i < columns.size(); i++) {
+      final String name = columns.get(i).getName();
+      if (ColumnHolder.TIME_COLUMN_NAME.equals(name) || Granularities.GRANULARITY_VIRTUAL_COLUMN_NAME.equals(name)) {
```
FrankChen021: [P2] Reject the virtual granularity time marker until it is wired through

This validation accepts a spec whose only time marker in `columns` is `Granularities.GRANULARITY_VIRTUAL_COLUMN_NAME`, but the clustered write/read metadata path still assumes `ColumnHolder.TIME_COLUMN_NAME`: `IncrementalIndex` discovers `timePosition` only from `__time`, `OnHeapClusterGroup` skips only `__time` when building per-group dimensions, `BaseTableProjectionSchema#getTimeColumnName` defaults to `__time`, and `ClusteredValueGroupsBaseTableSchema` throws unless ordering contains `__time`. A clustered base-table spec with `__virtualGranularity` in columns and no `__time` therefore passes this validation, then either treats the marker as an ordinary dimension or fails during metadata schema construction. Please require `__time` for now, or wire the virtual-granularity marker through the clustered schema/group writer consistently.

FrankChen021 (commented):
I rechecked the ordering follow-up. The clustered base-table spec now computes ordering from its columns and no longer accepts or serializes a declared ordering, so the original concern is handled.

Reviewed 32 of 55 changed files.

---

processing/src/main/java/org/apache/druid/data/input/impl/ClusteredValueGroupsBaseTableProjectionSpec.java
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  @Override
+  @JsonIgnore
+  public DimensionsSpec getDimensionsSpec()
+  {
+    return dimensionsSpec;
+  }
+
+  private static void validate(List<DimensionSchema> columns, List<String> clusteringColumns)
```
capistrant, line 173: could we eagerly validate clustering dimension types and throw InvalidInput exception using the DimensionSchema in `columns` for the relevant `clusteringColumns`?

processing/src/main/java/org/apache/druid/data/input/impl/ClusteredValueGroupsBaseTableProjectionSpec.java
```diff
@@ -193,40 +206,45 @@ private static void validate(List<DimensionSchema> columns, List<String> cluster
...
-          "clustering by %s / %s is not yet supported; the time column must be a non-clustering column",
-          ColumnHolder.TIME_COLUMN_NAME,
-          Granularities.GRANULARITY_VIRTUAL_COLUMN_NAME
+          "clustering by [%s] is not yet supported; the time column must be a non-clustering column",
+          ColumnHolder.TIME_COLUMN_NAME
       );
     }
   }
 
+  private static boolean isSupportedClusteringType(ColumnType type)
```
capistrant: `Projections.isAllowedClusteringType` is an existing public utility doing the same. I think consolidating would be helpful to make future changes to supported clustering types (if any) require less fixups

capistrant (approved):
Last comment is non-blocking. +1 from me

PR #19581, opened 2026-06-15
test: Try to fix row-count race in flaky `EmbeddedKafkaSupervisorTest`

amaechler:
`EmbeddedKafkaSupervisorTest.test_runKafkaSupervisor` is flaky in CI:

```
EmbeddedKafkaSupervisorTest.test_runKafkaSupervisor:129 expected: <10> but was: <8>
```

(observed values 7–9 across retries)

The test produces 10 records, waits only for the broker to discover the datasource (the first matching metric event), then immediately asserts `SELECT COUNT(*) == 10`. Under a loaded CI runner the query races ingestion and observes fewer than 10 rows.

This adds a wait on `ingest/events/processed` reaching the expected count before the row-count assertion identical to what `test_runSupervisor_withEmptyDimension` in this file already uses. 

<hr>

This PR has:

- [x] been self-reviewed.


kfaraz (approved):

abhishekrb19:
Thanks @amaechler. Saw this test flake a few times recently in some recent PRs as well.

PR #19592, opened 2026-06-17
fix(delta): drain all batches per scan file in DeltaInputSourceIterator

### LINKED ISSUE #18606: Druid Delta Table load Error - Unable to load full table, restricted to 1024 records per parquet file
Running Druid on an Openshift cluster, and using the Druid Delta Lake extension(https://github.com/apache/druid/tree/master/extensions-contrib/druid-deltalake-extensions) to connect and load Delta tables. 

Facing the following issue, - error while loading with delta connector: only 1024 records of each constituent parquet file(each partition of the delta table) is loaded

Also, there is an an error on the UI as soon as the load is over - ERROR: Request failed with status code 404. This may be unrelated to the issue with parquet data load

### Affected Version
34.0.0
33.0.0

### Description
- Running Druid on Openshift cluster and using Datainfra Druid operator version: 0.3.8
- Trying to load delta table from parquet, using MSQ load with the below query context,
{
  "finalizeAggregations": false,
  "groupByEnableMultiValueUnnesting": false,
  "arrayIngestMode": "array",
  "maxNumTasks": 11,
  "externalDataSampleRows": 0,
  "taskStatusCheckPeriodMs": 5000,
  "sqlInsertTaskNumSlots": "max"
}

- Query used for the load,

REPLACE INTO "table" OVERWRITE ALL
WITH "ext" AS (
  SELECT *
  FROM TABLE(
    EXTERN(
      '{"type":"delta","tablePath":"path"}',
      '{"type":"parquet"}'
    )
  ) EXTEND ("col1" VARCHAR, "col2" VARCHAR, "col3" VARCHAR, "col4" BIGINT, "col5" VARCHAR, "col6" VARCHAR, "col7" BIGINT, "col8" VARCHAR, "col9" VARCHAR)
)
SELECT
  MILLIS_TO_TIMESTAMP("dop" * 1000) AS "__time",
  "col1", "col2", "col3", "col4", "col5", "col6", "col7", "col8", "col9"
FROM "ext"
PARTITIONED BY DAY

rinchinov:
Fixes #18606.

#### Bug

`DeltaInputSourceIterator.hasNext()` used a **local variable** for the per-file
`CloseableIterator<FilteredColumnarBatch>`. When the method returned `true` after
finding the first non-empty batch of a file, that iterator went out of scope.
The next `hasNext()` call advanced to the **next file**, skipping all remaining
batches of the current file.

With the Delta kernel's default batch size of **1024 rows**, this produced exactly
`1024 × numFiles` rows regardless of actual file size — matching the symptom
reported in #18606.

#### Fix

Promoted `filteredBatchIterator` to a class field (`currentFileIterator`).
`hasNext()` now drains all batches of the current file before advancing to the
next one. Also fixed `close()` to close `currentFileIterator` and drain all
remaining file iterators (the original only closed one).

#### Regression test

Added `LargeRowGroupDeltaTable` (2 Parquet files × 2000 rows = 4000 total) and
`BatchDrainRegressionTests` inside `DeltaInputSourceTest`.

- **Without the fix**: `1024 × 2 = 2048` rows returned
- **With the fix**: `4000` rows returned

#### Release note

Fixed a bug in the Delta Lake input source where only 1024 rows per Parquet file
were ingested. Ingestion tasks now return all rows from each file.

---

- `DeltaInputSourceReader` (fix)
- `LargeRowGroupDeltaTable` (new test descriptor)
- `DeltaInputSourceTest` (new `BatchDrainRegressionTests` inner class)
- `src/test/resources/large-row-group-table` (new test Delta table)

---

This PR has:

- [x] been self-reviewed.


extensions-contrib/druid-deltalake-extensions/src/test/java/org/apache/druid/delta/input/LargeRowGroupDeltaTable.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public static final InputRowSchema SCHEMA = new InputRowSchema(
+      new TimestampSpec("id", "posix", null),
+      new DimensionsSpec(ImmutableList.of(
+          new LongDimensionSchema("id"),
+          new StringDimensionSchema("name")
+      )),
+      ColumnsFilter.all()
+  );
+}
```
FrankChen021: [P1] Add the missing final newline

This new Java file does not end with a newline; its last byte is the closing brace. Druid's Checkstyle configuration enables `NewlineAtEndOfFile`, so this should fail style/CI until a trailing newline is added.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

extensions-contrib/druid-deltalake-extensions/src/main/java/org/apache/druid/delta/input/DeltaInputSourceReader.java
```diff
@@ -111,20 +118,20 @@ public DeltaInputSourceIterator(
...
           if (currentBatch.hasNext()) {
             return true;
           }
         }
+
+        // Advance to the next file.
+        if (!filteredColumnarBatchIterators.hasNext()) {
+          return false;
+        }
+        currentFileIterator = filteredColumnarBatchIterators.next();
```
FrankChen021, line 148: [P1] Close exhausted file iterators before advancing

The latest commit only adds a newline, so this remains from the full PR diff: after draining currentFileIterator, hasNext() overwrites it with the next scan-file iterator without closing the exhausted one. These iterators come from Scan.transformPhysicalData(), whose close() closes the underlying Parquet physicalDataIter, so a fully read multi-file table leaks every completed file reader except the last one closed by DeltaInputSourceIterator.close(). Large Delta ingestions can therefore accumulate open Parquet readers/file handles. Close the exhausted currentFileIterator before assigning the next one, or otherwise ensure each per-file iterator is closed once it is drained.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

extensions-contrib/druid-deltalake-extensions/src/main/java/org/apache/druid/delta/input/DeltaInputSourceReader.java
rinchinov: thanks for mentioning it, applied closing iterator

abhishekrb19 (approved):
Good catch, thanks for the fix!

abhishekrb19:
@rinchinov, if you’re interested in adding an embedded test for the Delta lake extension which is an end-to-end integration test, that would be a good addition. There’s one for the Iceberg extension here: https://github.com/apache/druid/tree/master/embedded-tests/src/test/java/org/apache/druid/testing/embedded/iceberg

We could try something similar for the delta-lake extension as well.

rinchinov:
Added the embedded end-to-end ingestion test for the Delta Lake input source as a separate (stacked) draft PR: #19611. It reuses the 4000-row regression table from this PR and is green only with the fix here, so it's kept as a draft until this one merges.

PR #19597, opened 2026-06-17
feat: clustered segments + MSQ compaction

clintropolis:
### Description
Follow-up to #19579, this PR adds MSQ compaction support for clustered segments when using 'inline' or reindexing template based compaction.

changes:
* `CompactionTask` now can specify `baseTable` spec to create clustered segments
* `DataSourceMSQDestination` can now specify a `baseTable` so MSQ can generate clustered segments (or any other future baseTable spec)
* adds `baseTable` to 'inline' and reindexing template compaction configs to feed to compaction task for auto-compaction
* adds `baseTable`, `segmentGranularitySpec` to `CompactionState`, `CompactionStatus` is `baseTable` aware for checks
* guards to prevent `baseTable` from working with 'native' compaction and direct towards MSQ compaction

### DIFF
--- a/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/destination/SegmentGenerationUtils.java
+++ b/multi-stage-query/src/main/java/org/apache/druid/msq/indexing/destination/SegmentGenerationUtils.java
@@ -22,6 +22,7 @@ package org.apache.druid.msq.indexing.destination;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import it.unimi.dsi.fastutil.ints.IntList;
+import org.apache.druid.data.input.impl.BaseTableProjectionSpec;
 import org.apache.druid.data.input.impl.DimensionSchema;
 import org.apache.druid.data.input.impl.DimensionsSpec;
 import org.apache.druid.data.input.impl.LongDimensionSchema;
@@ -85,20 +86,6 @@ public final class SegmentGenerationUtils
   {
     final DataSourceMSQDestination destination = (DataSourceMSQDestination) querySpec.getDestination();
     final boolean isRollupQuery = isRollupQuery(query);
-    final boolean forceSegmentSortByTime =
-        MultiStageQueryContext.isForceSegmentSortByTime(querySpec.getContext());
-
-    final NonnullPair<DimensionsSpec, List<AggregatorFactory>> dimensionsAndAggregators =
-        makeDimensionsAndAggregatorsForIngestion(
-            querySignature,
-            queryClusterBy,
-            destination.getSegmentSortOrder(),
-            forceSegmentSortByTime,
-            columnMappings,
-            isRollupQuery,
-            query,
-            destination.getDimensionSchemas()
-        );
 
     final TransformSpec transformSpec;
     if (query.getFilter() != null) {
@@ -112,6 +99,34 @@ public final class SegmentGenerationUtils
     } else {
       transformSpec = null;
     }
+
+    // base-table mode, the destination supplies the base-table shape
+    if (destination.getBaseTable() != null) {
+      final Granularity queryGranularity =
+          query.context().getGranularity(DruidSqlInsert.SQL_INSERT_QUERY_GRANULARITY, jsonMapper);
+      final BaseTableProjectionSpec baseTable =
+          destination.getBaseTable().withQueryGranularity(queryGranularity);
+      return DataSchema.builder()
+                       .withDataSource(destination.getDataSource())
+                       .withTimestamp(new TimestampSpec(ColumnHolder.TIME_COLUMN_NAME, "millis", null))
+                       .withTransform(transformSpec)
+                       .withBaseTable(baseTable)
+                       .build();
+    }
+
+    final boolean forceSegmentSortByTime =
+        MultiStageQueryContext.isForceSegmentSortByTime(querySpec.getContext());
+    final NonnullPair<DimensionsSpec, List<AggregatorFactory>> dimensionsAndAggregators =
+        makeDimensionsAndAggregatorsForIngestion(
+            querySignature,
+            queryClusterBy,
+            destination.getSegmentSortOrder(),
+            forceSegmentSortByTime,
+            columnMappings,
+            isRollupQuery,
+            query,
+            destination.getDimensionSchemas()
+        );
     return DataSchema.builder()
                      .withDataSource(destination.getDataSource())
                      .withTimestamp(new TimestampSpec(ColumnHolder.TIME_COLUMN_NAME, "millis", null))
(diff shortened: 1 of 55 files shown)


server/src/main/java/org/apache/druid/client/indexing/ClientCompactionRunnerInfo.java
```diff
@@ -99,6 +99,11 @@ public static CompactionConfigValidationResult validateCompactionConfig(
...
     CompactionEngine compactionEngine = newConfig.getEngine() == null ? defaultCompactionEngine : newConfig.getEngine();
     if (compactionEngine == CompactionEngine.NATIVE) {
+      if (newConfig.getBaseTable() != null) {
+        return CompactionConfigValidationResult.failure(
+            "Compaction engine[native] does not support 'baseTable'; use the MSQ compaction engine."
+        );
+      }
       return CompactionConfigValidationResult.success();
     } else {
       return compactionConfigSupportedByMSQEngine(newConfig);
```
capistrant, line 109: what about adding a `baseTable != null` block to this MSQ validation method that blows up if the config is trying to set the top level data schema configs that are now handled by the base table. fuzzy to me on if that situation would be handled okay (base table wins or blow up later on), but blowing up eagerly feels safer and will prompt operator to consciously modify the config (or rules?) to make it play nice

capistrant (dismissed):
looks good. I re-read some of the paths before hitting submit here and I think it is a silent ignore of the other top level configs that base table now owns. which seems kind of sad but not destructive at least? I think it is worth forcing an operator to acknowledge what they are doing by cleaning up their config (or rules)

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/destination/SegmentGenerationUtils.java
```diff
@@ -112,6 +99,34 @@ public static DataSchema makeDataSchemaForIngestion(
...
+    if (destination.getBaseTable() != null) {
+      final Granularity queryGranularity =
+          query.context().getGranularity(DruidSqlInsert.SQL_INSERT_QUERY_GRANULARITY, jsonMapper);
+      final BaseTableProjectionSpec baseTable =
+          destination.getBaseTable().withQueryGranularity(queryGranularity);
+      return DataSchema.builder()
+                       .withDataSource(destination.getDataSource())
+                       .withTimestamp(new TimestampSpec(ColumnHolder.TIME_COLUMN_NAME, "millis", null))
+                       .withTransform(transformSpec)
+                       .withBaseTable(baseTable)
```
FrankChen021, line 113: [P1] Carry projections into base-table segment generation

The baseTable branch builds the worker DataSchema without withProjections(destination.getProjections()), unlike the legacy branch. MSQCompactionRunner passes projections through DataSourceMSQDestination, but they are dropped before segment generation, so baseTable compaction configured with projections writes segments without them and stores lastCompactionState.projections as null; CompactionStatus then keeps scheduling the same compaction. Add the projections to this builder path and cover baseTable+projections.

processing/src/main/java/org/apache/druid/data/input/impl/ClusteredValueGroupsBaseTableProjectionSpec.java
```diff
@@ -170,6 +172,73 @@ public DimensionsSpec getDimensionsSpec()
...
+   * Idempotent: if the spec already declares a query-granularity virtual column, that one is authoritative and this is
+   * a no-op. (The compaction path attaches the virtual column up front; the MSQ generation path then calls this again
+   * with the query-derived granularity, which must not double-add.)
+   */
+  @Override
+  public ClusteredValueGroupsBaseTableProjectionSpec withQueryGranularity(@Nullable Granularity queryGranularity)
+  {
+    if (queryGranularity == null
+        || Granularities.NONE.equals(queryGranularity)
+        || Granularities.ALL.equals(queryGranularity)
```
FrankChen021: [P1] Do not erase ALL query granularity

Treating Granularities.ALL as a no-op makes baseTable compaction report an effective query granularity of NONE. That bypasses MSQCompactionRunner's existing ALL handling, which assigns rows to the interval start timestamp, so queryGranularity=ALL produces incorrectly timestamped clustered segments; the later status check also compares configured ALL to stored NONE and can recompact forever. Preserve or reject ALL explicitly instead of silently normalizing it away.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 2 |
| P2 | 1 |
| P3 | 0 |
| Total | 3 |

Reviewed 55 of 55 changed files.

- server/src/main/java/org/apache/druid/client/indexing/ClientCompactionRunnerInfo.java:127 - [P2] Validate range partitions against baseTable columns. The affected call is existing code made incomplete by the new baseTable config field. compactionConfigSupportedByMSQEngine still passes dimensionSchemas from newConfig.getDimensionsSpec(), so baseTable configs with no legacy dimensionsSpec skip range-partition column validation at config time. The same config is later validated against dataSchema.getDimensionsSpec() inside MSQCompactionRunner and can fail only after the coordinator submits tasks. Use newConfig.getBaseTable().getDimensionsSpec()/virtual columns when baseTable is present.

---

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/destination/SegmentGenerationUtils.java
capistrant: oh this is a good catch.

processing/src/main/java/org/apache/druid/data/input/impl/ClusteredValueGroupsBaseTableProjectionSpec.java
capistrant: I think rejecting ALL with demand to use some other granularity makes more sense than going through effort to try and support it in granularity virtual column (if that is even possible).

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/destination/SegmentGenerationUtils.java
clintropolis: i was planning on wiring this up in a follow-up that enables projections support for clustered segments, but i guess it doesn't hurt to wire it up now

processing/src/main/java/org/apache/druid/data/input/impl/ClusteredValueGroupsBaseTableProjectionSpec.java
clintropolis: i'll change this to reject all for now. with projections we support all granularity on a projection by leaving out a time column, i considered the same thing for clustered segments but haven't wired it up yet.

multi-stage-query/src/main/java/org/apache/druid/msq/indexing/destination/SegmentGenerationUtils.java
capistrant: ah I see. but ya, one less thing to have to remember in the projections support PR

capistrant (approved):
responses to frank's comments look good. re-approve

PR #19598, opened 2026-06-18
Add advertisedPlaintextPort to DruidNode for sidecar proxy support

saithal-confluent:
## Description

Allow Druid nodes to advertise a different plaintext port for peer discovery while binding Jetty to the actual service port.

### Motivation

When running Druid with a sidecar proxy (e.g. Envoy for east-west mTLS), the port the proxy listens on differs from the port Jetty binds to. Without this change, peer discovery advertises the Jetty bind port, so other nodes try to connect directly — bypassing the sidecar — and mTLS fails.

### Changes

- Added `advertisedPlaintextPort` field to `DruidNode` (JSON property: `advertisedPlaintextPort`, `@Max(0xffff)`)
- When set to a positive value, `getHostAndPort()` returns the advertised port instead of `plaintextPort`
- When unset or ≤ 0, falls back to `plaintextPort` — no behaviour change for existing deployments
- Updated `ServiceLocation` to use the advertised port
- Added tests covering the new field and its fallback semantics

### Testing

- [ ] `mvn test -pl server -Dtest=DruidNodeTest`
- [ ] `mvn test -pl server` (full server module)

<sub>This is a backport of an internal change validated in production.</sub>

### DIFF
--- a/server/src/main/java/org/apache/druid/server/DruidNode.java
+++ b/server/src/main/java/org/apache/druid/server/DruidNode.java
@@ -100,6 +100,10 @@ public class DruidNode
   @JsonProperty
   private Map<String, String> labels;
 
+  @JsonProperty
+  @Max(0xffff)
+  private int advertisedPlaintextPort = -1;
+
   public DruidNode(
       String serviceName,
       String host,
@@ -110,7 +114,21 @@ public class DruidNode
       boolean enableTlsPort
   )
   {
-    this(serviceName, host, bindOnHost, plaintextPort, null, tlsPort, enablePlaintextPort, enableTlsPort, null);
+    this(serviceName, host, bindOnHost, plaintextPort, null, tlsPort, enablePlaintextPort, enableTlsPort, null, null);
+  }
+
+  public DruidNode(
+      String serviceName,
+      String host,
+      boolean bindOnHost,
+      Integer plaintextPort,
+      Integer port,
+      Integer tlsPort,
+      Boolean enablePlaintextPort,
+      boolean enableTlsPort
+  )
+  {
+    this(serviceName, host, bindOnHost, plaintextPort, port, tlsPort, enablePlaintextPort, enableTlsPort, null, null);
   }
 
   /**
@@ -139,7 +157,8 @@ public class DruidNode
       @JacksonInject(useInput = OptBoolean.TRUE) @Named("tlsServicePort") @JsonProperty("tlsPort") Integer tlsPort,
       @JsonProperty("enablePlaintextPort") Boolean enablePlaintextPort,
       @JsonProperty("enableTlsPort") boolean enableTlsPort,
-      @JsonProperty("labels") @Nullable Map<String, String> labels
+      @JsonProperty("labels") @Nullable Map<String, String> labels,
+      @JsonProperty("advertisedPlaintextPort") Integer advertisedPlaintextPort
   )
   {
     init(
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/druid/server/DruidNode.java
```diff
@@ -139,7 +157,8 @@ public DruidNode(
       @JacksonInject(useInput = OptBoolean.TRUE) @Named("tlsServicePort") @JsonProperty("tlsPort") Integer tlsPort,
       @JsonProperty("enablePlaintextPort") Boolean enablePlaintextPort,
       @JsonProperty("enableTlsPort") boolean enableTlsPort,
-      @JsonProperty("labels") @Nullable Map<String, String> labels
+      @JsonProperty("labels") @Nullable Map<String, String> labels,
+      @JsonProperty("advertisedPlaintextPort") Integer advertisedPlaintextPort
```
FrankChen021, line 176: [P1] Preserve the labels-taking DruidNode constructor

Adding advertisedPlaintextPort to the existing @JsonCreator constructor removes the previous nine-argument overload that accepted labels. Existing source in DruidNodeTest still calls new DruidNode(..., labels), for example the equality and serde tests, and those calls no longer match any constructor, so the server module will not compile. Please keep a nine-argument overload that delegates with a null advertisedPlaintextPort, or update all local call sites while preserving source compatibility intentionally.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/server/DruidNode.java
saithal-confluent: Restored the 9-arg overload so existing call sites keep compiling. The new field is reachable via the 10-arg constructor instead of 9-arg one as it would make existing null-literal calls ambiguous. 
Thanks for the review @FrankChen021 , could you please take a look again?

FrankChen021 (commented):
Follow-up reviewed. The constructor compatibility concern is addressed: the nine-argument labels overload is restored, and the advertised plaintext port remains on the ten-argument constructor path.

---

FrankChen021:
Thanks for adding this. After tracing the usage, I think the scope needs clarification and possibly additional changes.

The new property is applied indirectly through `DruidNode.getHostAndPort()` (and therefore `getHostAndPortToUse()` when Druid TLS is disabled), and directly through `ServiceLocation.fromDruidNode()`. However, `DruidNode.getPortToUse()` and `getUriToUse()` still return/use `plaintextPort`, and there are peer-facing paths that use those methods, including basic-security cache notifications, catalog synchronization, the system server-properties fetch, and Consul registration.

If the goal is for peer traffic to go through the sidecar, could you please review these call sites and either:

1. use the advertised port where appropriate, or
2. document the narrower set of discovery/RPC paths this property is intended to affect and explain why the remaining paths should continue using the bind port?

Could you also update the configuration documentation to include `druid.advertisedPlaintextPort` alongside `druid.plaintextPort`? The documentation should ideally state:

- `plaintextPort` remains the Jetty bind/listen port;
- `advertisedPlaintextPort` is the port published/used by the supported discovery and RPC paths;
- when unset or non-positive, it falls back to `plaintextPort`;
- it applies only when plaintext is enabled, and Druid's configured TLS port remains preferred when TLS is enabled.

Without this documentation, it is difficult for operators to discover the property or understand its interaction with the existing port settings.

FrankChen021 (changes requested):

FrankChen021:
@saithal-confluent can you solve the conflicts and resolve the problems I raised above?

PR #19599, opened 2026-06-18
feat: support aggregate projections on clustered segments

clintropolis:
### Description
This PR allows aggregate projections to be built on clustered segments. Also fixes a bug uncovered while adding an embedded-test for this stuff where grouping was incorrectly using dictionaryIds when using the concatenating cursors, resulting in wrong results. This PR fixes by just not allowing grouping on the dictionary ids, but will do a better fix in a follow-up PR to bring it back, but correctly, for perf reasons.

changes:
* allow `AggregateProjectionSpec` on a clustered base table; remove the build and merge time guards that rejected it
* persist projections in `IndexMergerV10.makeClusteredIndexFiles` via the shared `makeProjections(...)`
* fix a bug on a non-clustering dictionary column over the clustered base table which conflated values (per-group-local dictionary IDs reused across the `ConcatenatingCursor` causing values to inappropriately group together). Force value-based grouping by reporting non-clustering columns as non-dictionary-encoded, on capabilities and on the selector (cardinality/name-lookup)
* tests: build and query tests for projections for both incremental and persisted segments, and also added first E2E coverage for clustered segments (`ClusteredSegmentProjectionQueryTest`, native ingestion + projection-vs-noProjections queries)


processing/src/main/java/org/apache/druid/segment/projections/ClusteringVectorColumnSelectorFactory.java
```diff
@@ -181,7 +181,21 @@ public ColumnCapabilities getColumnCapabilities(String column)
...
+      // keys on the selector's (per-group-local) IDs when the column reports as dictionary-encoded
+      // (GroupByVectorColumnProcessorFactory#useDictionaryEncodedSelector), conflating distinct values across groups.
+      // Reporting non-dictionary-encoded routes it to the value-building vector selector, which is correct across
+      // groups.
+      final ColumnCapabilities delegateCapabilities = delegate.getColumnCapabilities(column);
+      if (delegateCapabilities == null) {
+        return null;
+      }
+      return ColumnCapabilitiesImpl.copyOf(delegateCapabilities)
+                                   .setDictionaryEncoded(false)
```
FrankChen021, line 192: [P2] Hide vector selector dictionaries too

This hides per-group dictionaries through ColumnCapabilities, but the delegating vector dimension selectors still return the current group's cardinality/name-lookup/idLookup. Any vectorized caller that directly uses makeSingleValueDimensionSelector or makeMultiValueDimensionSelector and caches ids across ConcatenatingVectorCursor group transitions can reproduce the same wrong-result class this PR fixes for scalar grouping. Please mirror the scalar DelegatingDimensionSelector behavior for vector dimension selectors by returning CARDINALITY_UNKNOWN, false, and null for non-clustering columns, with a vectorized multi-group clustered test.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

processing/src/main/java/org/apache/druid/segment/projections/ClusteringVectorColumnSelectorFactory.java
capistrant: hmm. tracing some code paths, I think that the capabilities fix will actually properly fix the bug for any actual query paths, but maybe this suggestion makes sense to remove any doubt or fragility since I think frank is right that we should mirror the column selector factory changes in this PR

capistrant (commented):
If Frank's comment is innacurate/false positive I can flip to approve. But I think makes sense. Other changes look good, cool to see some embedded tests coming online

processing/src/main/java/org/apache/druid/segment/projections/ClusteringVectorColumnSelectorFactory.java
clintropolis: cardinality unknown doesn't make sense for the vector engine, dictionary encoded selectors are only used when the columns are actually dictionary encoded, and callers must check the column capabilities (though i see now that there aren't really any javadocs specifying this contract).

I technically should remove the implementation since it will never be used, the main reason i left it is because later I'm going to have another pass at making this better and will likely need to bring an implementation back for that, but I guess I could just do it whenever I make the optimization instead of leaving dead code hanging around.

clintropolis: removed them for now, i can add them back later when i do some optimization

capistrant (approved):

PR #19600, opened 2026-06-19
feat: unnest and join makeCursorHolderAsync

clintropolis:
### Description
This PR adds `makeCursorHolderAsync` implementations to `UnnestCursorFactory` and `HashJoinSegmentCursorFactory`, and also switches the default implementation of `makeCursorHolderAsync` to explode with a defensive exception instead of call the sync method. A new `ResidentCursorFactory` interface has been added that can be used for any factory implementations that never have a reason to wait on stuff to avoid having to copy a default implementation everywhere.


FrankChen021 (commented):

---

processing/src/main/java/org/apache/druid/segment/CursorFactory.java
capistrant, line 1: if we want to make implementations of CursorFactory implement this method, why offer a default at all?

capistrant (approved):

processing/src/main/java/org/apache/druid/segment/CursorFactory.java
clintropolis: `CursorFactory` is sort indirectly a `@PublicApi` via `Segment#as` and also a sort of very low level extension point for custom segment implementations, so adding a new method with no default means extension writers would need to implement it and recompile for functionality that is currently at least off by default, so it seemed nicer to implement it so that they don't need to recompile immediately, making this async stuff more opt-in. I was thinking if something changes that makes it so that this path can be hit by default then I would change this to remove the default because at that point it really would need to be implemented

capistrant: Makes sense, thanks for the context!

PR #19603, opened 2026-06-19
fix: Fix invalid time comparison

### LINKED ISSUE #19602: SqlSegmentsMetadataManager: forceOrWaitOngoingDatabasePoll compares System.nanoTime() against converted System.currentTimeMillis(), always skipping fresh on-demand poll detection
SqlSegmentsMetadataManager: forceOrWaitOngoingDatabasePoll compares System.nanoTime() against converted System.currentTimeMillis(), always skipping fresh on-demand poll detection

## Bug

`forceOrWaitOngoingDatabasePoll` in `SqlSegmentsMetadataManager` contains an invalid clock comparison when checking whether a recent `OnDemandDatabasePoll` makes a new poll unnecessary.
            
## Affected version(s)
Reproducible on current `master`.
                                                    
## Root cause                   
`OnDemandDatabasePoll.initiationTimeNanos` is set via `System.nanoTime()` — a monotonic clock with an arbitrary JVM-internal reference point. Values are typically in the range of a few seconds to minutes in nanoseconds from JVM startup.     
                                                    
The staleness check constructs its comparison value as:                                            
                                                    
```java                                              
long checkStartTimeNanos = TimeUnit.MILLISECONDS.toNanos(checkStartTime);
// where checkStartTime = System.currentTimeMillis() 
```

`System.currentTimeMillis()` returns milliseconds since the Unix epoch.  Converting that to nanoseconds yields many orders of magnitude larger than any value returned by System.nanoTime().

As a result, the condition:                                                                                                                                              
                                                                                                                                                                           
  ```java
if (latestOnDemandPoll.initiationTimeNanos > checkStartTimeNanos)  
```

is always false. A concurrent on-demand poll that completed while the calling thread was waiting for the write lock is never recognised as sufficient, so forceOrWaitOngoingDatabasePoll always triggers an additional unnecessary poll.

Impact

Every call to forceOrWaitOngoingDatabasePoll that races with an in-progress OnDemandDatabasePoll performs a redundant full database poll instead of reusing the result of the concurrent one. Under high coordinator leadership-change frequency this can cause extra load on the metadata store.

### PR DESCRIPTION
**Pull request by vivek807**
Fixes #19602.

### Description

`forceOrWaitOngoingDatabasePoll` in `SqlSegmentsMetadataManager` checks whether a concurrent `OnDemandDatabasePoll` that completed while the calling thread waited for the write lock makes a new poll unnecessary. This check was always false due to a clock mismatch, causing an extra database poll on every contended call.

#### Fixed invalid clock comparison in `forceOrWaitOngoingDatabasePoll`

`OnDemandDatabasePoll.initiationTimeNanos` is assigned via `System.nanoTime()`, which is a monotonic clock with an arbitrary JVM-internal reference point.
The comparison value was constructed as:

```java
long checkStartTimeNanos = TimeUnit.MILLISECONDS.toNanos(checkStartTime);
// checkStartTime = System.currentTimeMillis()
```

System.currentTimeMillis() returns milliseconds since the Unix epoch (~1.7 × 10¹² ms). Converting to nanoseconds yields ~1.7 × 10²¹ ns, which is orders of magnitude larger than any System.nanoTime() value. The condition `initiationTimeNanos > checkStartTimeNanos` was therefore always false, so an in-progress on-demand poll was never recognised as sufficient.

This PR has:

- [x] been self-reviewed.


server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataManager.java
```diff
@@ -569,7 +570,6 @@ void forceOrWaitOngoingDatabasePoll()
         }
         // Verify if there was a on-demand poll completed while we were waiting for the lock
         if (latestDatabasePoll instanceof OnDemandDatabasePoll) {
-          long checkStartTimeNanos = TimeUnit.MILLISECONDS.toNanos(checkStartTime);
           OnDemandDatabasePoll latestOnDemandPoll = (OnDemandDatabasePoll) latestDatabasePoll;
           if (latestOnDemandPoll.initiationTimeNanos > checkStartTimeNanos) {
```
FrankChen021: [P2] Check poll success before reusing a contended on-demand poll

Now that this nanoTime comparison can be true, a caller queued behind another on-demand poll can return without checking whether that poll succeeded. If the first poll throws, doOnDemandPoll completes pollCompletionFuture exceptionally and releases the write lock; any caller whose checkStartTimeNanos was before that poll started will hit this branch and return the stale dataSourcesSnapshot instead of retrying or propagating the failure. The existing catch says unsuccessful latest polls should trigger a new poll, so this branch should verify pollCompletionFuture, for example with Futures.getUnchecked(latestOnDemandPoll.pollCompletionFuture), before returning.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

server/src/test/java/org/apache/druid/metadata/SqlSegmentsMetadataManagerTest.java
```diff
@@ -231,6 +233,51 @@ public void testPollPeriodicallyAndOnDemandInterleave() throws Exception
...
+    Thread caller = new Thread(() -> sqlSegmentsMetadataManager.forceOrWaitOngoingDatabasePoll());
+    caller.setDaemon(true);
+
+    SqlSegmentsMetadataManager.OnDemandDatabasePoll concurrentPoll;
+
+    rwLock.writeLock().lock();
+    try {
+      caller.start();
+      // Allow the caller to capture checkStartTimeNanos and block on the write lock.
+      Thread.sleep(100);
```
FrankChen021: [P3] Replace timing sleep with deterministic thread coordination

The new regression test depends on the caller thread being scheduled within this fixed 100 ms window so it captures checkStartTimeNanos before concurrentPoll is created. On a slow or overloaded CI worker, the caller can start after the sleep, making concurrentPoll.initiationTimeNanos <= checkStartTimeNanos; then the production branch under test is skipped and the caller may run its own poll or return early, causing a flaky failure or a false pass. Please coordinate on the lock queue/state, e.g. wait until the caller is queued on the write lock before creating concurrentPoll, instead of relying on sleep timing.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

---

maytasm:
@vivek807 
Thanks for flagging this issue. You are right, the values are not comparable. 
I'm thinking that maybe instead we should standardize the timestamps in OnDemandDatabasePoll and PeriodicDatabasePoll to use the same format (i.e. both use System.nanoTime() or both use System.currentTimeMillis()). This can avoid similar confusion / bug like this in the future. Wdyt? Thanks!

vivek807:
> @vivek807 Thanks for flagging this issue. You are right, the values are not comparable. I'm thinking that maybe instead we should standardize the timestamps in OnDemandDatabasePoll and PeriodicDatabasePoll to use the same format (i.e. both use System.nanoTime() or both use System.currentTimeMillis()). This can avoid similar confusion / bug like this in the future. Wdyt? Thanks!

Yes, this will avoid confusion in future. Let me refactor it to use same type of clock here

server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataManager.java
```diff
@@ -100,12 +100,12 @@ static class PeriodicDatabasePoll implements DatabasePoll
...
   {
-    final long initiationTimeNanos = System.nanoTime();
+    final long initiationTimeMillis = System.currentTimeMillis();
     final CompletableFuture<Void> pollCompletionFuture = new CompletableFuture<>();
 
-    long nanosElapsedFromInitiation()
+    long millisElapsedFromInitiation()
     {
-      return System.nanoTime() - initiationTimeNanos;
+      return System.currentTimeMillis() - initiationTimeMillis;
```
FrankChen021: [P2] Use monotonic time for freshness elapsed checks

`millisElapsedFromInitiation()` now computes elapsed time from `System.currentTimeMillis()`. If the wall clock moves backward after an on-demand poll starts, this value can become negative or too small, so `useLatestSnapshotIfWithinDelay()` can keep accepting a stale snapshot and the periodic poll task can sleep longer than `periodicPollDelay`. Keep a monotonic `nanoTime()` value for elapsed/freshness calculations and use a separate wall-clock timestamp only where this code compares against wall-clock poll start times.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataManager.java
```diff
@@ -567,11 +567,11 @@ void forceOrWaitOngoingDatabasePoll()
             && ((PeriodicDatabasePoll) latestDatabasePoll).lastPollStartTimestampInMs > checkStartTime) {
           return;
         }
-        // Verify if there was a on-demand poll completed while we were waiting for the lock
+        // Verify if there was an on-demand poll completed while we were waiting for the lock
         if (latestDatabasePoll instanceof OnDemandDatabasePoll) {
-          long checkStartTimeNanos = TimeUnit.MILLISECONDS.toNanos(checkStartTime);
```
kfaraz, line 572: We should change this line to simply be:

```java
long checkStartTimeNanos = System.nanoTime();
```

We should also change `PeriodicDatabasePoll.lastPollStartTimestampInMs` to be `lastPollStartTimestampInNanos`.

server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataManager.java
```diff
@@ -100,12 +100,12 @@ static class PeriodicDatabasePoll implements DatabasePoll
   @VisibleForTesting
   static class OnDemandDatabasePoll implements DatabasePoll
   {
-    final long initiationTimeNanos = System.nanoTime();
+    final long initiationTimeMillis = System.currentTimeMillis();
```
kfaraz: Let's stick to using nanos since it is more reliably monotonic.

server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataManager.java
```diff
@@ -567,11 +567,11 @@ void forceOrWaitOngoingDatabasePoll()
...
           return;
         }
-        // Verify if there was a on-demand poll completed while we were waiting for the lock
+        // Verify if there was an on-demand poll completed while we were waiting for the lock
         if (latestDatabasePoll instanceof OnDemandDatabasePoll) {
-          long checkStartTimeNanos = TimeUnit.MILLISECONDS.toNanos(checkStartTime);
           OnDemandDatabasePoll latestOnDemandPoll = (OnDemandDatabasePoll) latestDatabasePoll;
-          if (latestOnDemandPoll.initiationTimeNanos > checkStartTimeNanos) {
+          if (latestOnDemandPoll.initiationTimeMillis > checkStartTime) {
+            Futures.getUnchecked(latestOnDemandPoll.pollCompletionFuture);
```
kfaraz: Is this really needed?

kfaraz (commented):
Thanks for catching this @vivek807 !
I believe we should stick to using nanos, and just update the couple of places which use `System.currentTimeMillis()` to use `System.nanoTime()`.

In the future, we may just use `druid.java.util.common.Stopwatch` for simplicity but that change may not be worth the effort right now since this class will soon be deprecated in favor of using `SqlSegmentsMetadataManagerV2` which uses `SegmentMetadataCache`.

kfaraz (approved):

server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataManager.java
```diff
@@ -556,22 +556,22 @@ boolean useLatestSnapshotIfWithinDelay()
...
-            && ((PeriodicDatabasePoll) latestDatabasePoll).lastPollStartTimestampInMs > checkStartTime) {
+            && ((PeriodicDatabasePoll) latestDatabasePoll).lastPollStartTimestampInNanos > checkStartTimeNanos) {
           return;
         }
-        // Verify if there was a on-demand poll completed while we were waiting for the lock
+        // Verify if there was an on-demand poll completed while we were waiting for the lock
         if (latestDatabasePoll instanceof OnDemandDatabasePoll) {
-          long checkStartTimeNanos = TimeUnit.MILLISECONDS.toNanos(checkStartTime);
           OnDemandDatabasePoll latestOnDemandPoll = (OnDemandDatabasePoll) latestDatabasePoll;
           if (latestOnDemandPoll.initiationTimeNanos > checkStartTimeNanos) {
```
FrankChen021: [P2] Contended force calls still miss the poll they waited for

This still compares the existing on-demand poll's initiation time with the caller's check start. In the common contended path, thread A creates the OnDemandDatabasePoll and holds the write lock while polling; thread B enters forceOrWaitOngoingDatabasePoll after that initiation, waits for the same write lock until A's poll completes, then sees initiationTimeNanos <= checkStartTimeNanos and starts another database poll. That means the PR still serializes duplicate forced polls for callers that arrived during the actual poll, even though the previous poll completed after their call. Track and compare completion time, or otherwise record that the caller waited behind the in-progress on-demand poll, before returning/forcing a new poll.

server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataManager.java
```diff
@@ -90,7 +90,7 @@ static class PeriodicDatabasePoll implements DatabasePoll
      * leadership changes.
      */
     final CompletableFuture<Void> firstPollCompletionFuture = new CompletableFuture<>();
-    long lastPollStartTimestampInMs = -1;
+    long lastPollStartTimestampInNanos = -1;
```
FrankChen021: [P2] Do not use -1 as a nanoTime sentinel

This timestamp is now compared directly with System.nanoTime() readings, but System.nanoTime() has an arbitrary origin and is allowed to return negative values. If the manager has installed a new PeriodicDatabasePoll but its first task has not set this field yet, lastPollStartTimestampInNanos remains -1; on a JVM where the caller's checkStartTimeNanos is less than -1, forceOrWaitOngoingDatabasePoll will treat that unstarted poll as newer and return without polling or waiting. Use Long.MIN_VALUE, an explicit initialized flag, or a nullable/completion timestamp instead.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

Found 2 concurrency issues in the updated poll timing logic.

---

server/src/main/java/org/apache/druid/metadata/SqlSegmentsMetadataManager.java
kfaraz: Yeah, I suppose `null` would be a better choice for this now.

FrankChen021 (commented):
Handled the follow-up on the nanoTime sentinel discussion; no inline reply is needed at this point.

---

kfaraz (approved):

FrankChen021 (commented):

---

kfaraz:
Thanks for the changes, @vivek807 !

PR #19607, opened 2026-06-22
fix: Terminate connection on closing the stream from the DirectDruidClient

LakshSingla:
`DirectDruidClient` returns a `SequenceInputStream` consisting of the chunks fetched by the downstream data servers.  `close()` on the `SequenceInputStream` consumes the entire iteration, attempts to call close on the elements, before returning to the caller. This means that on attempting to `close()` the input stream returned by the DirectDruidClient, we end up fetching the entire response from the data servers. This fetching is wasteful, as we discard the results, and there's nothing to be done on the remaining chunks. 

This PR short-circuits the close() on the stream returned by the DirectDruidClient, so that close() exits immediately, and terminates the connection between the broker and the dataservers.

### DIFF
--- a/processing/src/main/java/org/apache/druid/java/util/http/client/response/HttpResponseHandler.java
+++ b/processing/src/main/java/org/apache/druid/java/util/http/client/response/HttpResponseHandler.java
@@ -98,5 +98,13 @@ public interface HttpResponseHandler<IntermediateType, FinalType>
      * @return time that backpressure was applied (channel was closed for reads)
      */
     long resume(long chunkNum);
+
+    /**
+     * Closes the underlying connection, abandoning any remaining response.
+     *
+     * Intended for callers that decide they no longer need the rest of the response (for example, because the
+     * consumer of the resulting stream has been closed early)
+     */
+    void abort();
   }
 }
--- a/server/src/main/java/org/apache/druid/client/DirectDruidClient.java
+++ b/server/src/main/java/org/apache/druid/client/DirectDruidClient.java
@@ -337,8 +337,6 @@ public class DirectDruidClient<T> implements QueryRunner<T>
                 {
                   final TrafficCop trafficCop;
                   synchronized (done) {
-                    // If the response already completed normally there is nothing to abandon, and the connection may
-                    // already be back in the pool, so don't race that cleanup by aborting it now.
                     if (done.get()) {
                       return;
                     }
--- a/server/src/main/java/org/apache/druid/client/DirectDruidClient.java
+++ b/server/src/main/java/org/apache/druid/client/DirectDruidClient.java
@@ -181,6 +181,9 @@ public class DirectDruidClient<T> implements QueryRunner<T>
         private final AtomicLong channelSuspendedTime = new AtomicLong(0);
         private final BlockingQueue<InputStreamHolder> queue = new LinkedBlockingQueue<>();
         private final AtomicBoolean done = new AtomicBoolean(false);
+        // Set when the consumer closes the result stream early (see the SequenceInputStream.close() override in
+        // handleResponse). Once set, incoming chunks are dropped rather than buffered.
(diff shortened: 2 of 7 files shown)


FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

---

server/src/main/java/org/apache/druid/client/DirectDruidClient.java
```diff
@@ -310,7 +325,33 @@ public InputStream nextElement()
...
+                 * Closing this stream means the caller no longer needs the response. The default
+                 * {@link SequenceInputStream#close()} would drain the entire remaining response off the wire first
+                 * we want to avoid. Instead, abandon the response and force-close the connection
+                 */
+                @Override
+                public void close()
+                {
+                  final TrafficCop trafficCop;
+                  synchronized (done) {
+                    if (done.get()) {
```
FrankChen021, line 340: [P2] Early close still needs to discard completed buffered response

When the server has already finished sending the response, `done` is true, so this new `close()` override returns before setting `discard`, clearing `queue`, or delegating to `SequenceInputStream.close()`. A caller that stops after the first rows while a fast historical has already filled the remaining queue will now retain all queued `InputStreamHolder`/`ChannelBuffer` chunks until the yielder and parser are GC'd; before this change, `SequenceInputStream.close()` drained and removed the queued elements once `done` was true. The close path should still clear/discard queued buffers for `done == true`; only the connection abort needs to be skipped once the response is complete.

processing/src/main/java/org/apache/druid/java/util/http/client/response/HttpResponseHandler.java
```diff
@@ -98,5 +98,13 @@ interface TrafficCop
...
      */
     long resume(long chunkNum);
+
+    /**
+     * Closes the underlying connection, abandoning any remaining response.
+     *
+     * Intended for callers that decide they no longer need the rest of the response (for example, because the
+     * consumer of the resulting stream has been closed early)
+     */
+    void abort();
```
FrankChen021, line 108: [P2] Keep TrafficCop source and binary compatible

Adding `abort()` as a second abstract method makes `TrafficCop` no longer a functional interface and forces every out-of-tree mock or custom `HttpClient` implementation to add the method, with existing compiled implementations at risk of `AbstractMethodError` if this hook is invoked. Since only the Netty implementation needs special behavior, make `abort()` a default no-op and let Netty override it; that preserves existing lambda/test implementations while still enabling this PR's connection-abort path.

LakshSingla: I don't want to add a default no-op for this modification. The callers must explicitly define what close() should do.

FrankChen021: Thanks for clarifying. I agree Netty should define the actual close behavior explicitly, and it does here. My concern is narrower: TrafficCop is a public nested interface that was previously SAM-compatible, so adding abort() as a required abstract method breaks source compatibility for lambdas/custom HttpClient fakes and can still throw AbstractMethodError for compiled implementations if this path is called. A default no-op does not stop Netty from overriding abort(); it just preserves the existing contract for implementations that cannot abort. If the intent is to require all implementations to make an explicit choice, I think this should be treated as an intentional API break and reflected accordingly. Reviewed 7 of 7 changed files.

cryptoe: I am not sure I get the concern. Traffic cop is not marked as an extension point so if extenders need to implement something they need to. Also its quite a core concept of druid internal networking so I am not sure if folks would be actually extending this.

cryptoe (approved):

PR #19609, opened 2026-06-22
feat: Handle MRAP URI

### LINKED ISSUE #19608: Add AWS S3 Multi-Region Access Point (MRAP) support to the S3 extension
### Description

Add support for AWS S3 Multi-Region Access Points (MRAPs) and S3 Access Point ARNs in Druid's S3 extension.

Currently, the bucket field in Druid's S3 configuration only accepts standard DNS-compliant bucket names. AWS Access Point ARNs (eg., `arn:aws:s3::123456789123:accesspoint:bucket.mrap`) are rejected at construction time in `CloudObjectLocation` because they fail the URL-encoding equality check used to enforce DNS naming rules. Additionally, some tools produce ARNs with a slash separator (accesspoint/alias) instead of the colon-delimited form (accesspoint:alias) expected by the AWS SDK, causing further failures downstream. 
 
This change: 
- Relaxes the bucket name validation in CloudObjectLocation to permit valid S3 Access Point ARNs alongside DNS-compliant names. 
- Adds S3Utils.normalizeBucketName() to canonicalize the slash-delimited form to the colon-delimited form at ingestion points (S3DataSegmentPusherConfig, S3LoadSpec). 
- Supports both regional Access Point ARNs (`arn:aws:s3:<region>:<account>:accesspoint:<name>`) and MRAP ARNs (`arn:aws:s3::<account>:accesspoint:<name>.mrap`). 
 
No API surface changes; the bucket configuration field continues to accept plain bucket names unchanged.

### Motivation

**Use case**

AWS Multi-Region Access Points provide a single global S3 endpoint that routes requests to the nearest healthy bucket replica across regions. Operators use MRAPs for:

- Active-active multi-region Druid deployments backed by S3 Cross-Region Replication (CRR).
- Disaster recovery setups where deep storage must remain accessible during a regional outage.
- Simplifying Druid configuration across regions — one ARN in druid.storage.bucket instead of per-region overrides.
- Access Point ARNs more broadly (single-region) are also used to enforce fine-grained IAM access controls on shared buckets without exposing the bucket name.

Why the current behavior blocks this

CloudObjectLocation enforces:
```java
Preconditions.checkArgument(
this.bucket.equals(StringUtils.urlEncode(this.bucket)),
"bucket must follow DNS-compliant naming conventions"
);
```

An ARN like `arn:aws:s3::123456789123:accesspoint:bucket.mrap` URL-encodes to `arn:aws:s3::123456789123:accesspoint:bucket.mrap`, so the check always fails. There is no escape hatch. Users who configure an MRAP ARN as the Druid storage bucket receive an IllegalArgumentException at startup with no workaround short of patching the code.

The AWS SDK for Java (v1 and v2) accepts ARN strings wherever a bucket name is expected, so no SDK-level changes are required. The fix is purely a validation relaxation and a normalization helper.

vivek807:
Fixes #19608.

Description

Adds support for AWS S3 Multi-Region Access Points (MRAPs) and S3 Access Point ARNs as the bucket value in Druid's S3 extension. Previously, any ARN passed as a bucket name was rejected at startup due to a strict DNS-naming validation check in CloudObjectLocation.

Relaxed bucket validation in `CloudObjectLocation`

The existing check enforces that a bucket name URL-encodes to itself, which is a proxy for DNS compliance. ARNs contain colons and fail that check unconditionally. The validation now accepts a bucket that either passes the existing DNS check or matches a valid S3 Access Point ARN pattern (regional and MRAP, across aws, aws-cn, and aws-us-gov partitions).

Added `S3Utils.isS3Arn()` and `S3Utils.normalizeBucketName()`

Some tooling produces Access Point ARNs with a slash separator (accesspoint/alias) rather than the colon-delimited form (accesspoint:alias) that the AWS SDK expects. `normalizeBucketName()` canonicalizes the slash form to the colon form. It is a no-op for plain bucket names. S3DataSegmentPusherConfig.setBucket() and S3LoadSpec constructor both call this at the point of construction so the rest of the code never sees the unnormalized form.

Release note

Druid's S3 extension now accepts AWS S3 Access Point ARNs and Multi-Region Access Point (MRAP) ARNs as the druid.storage.bucket value. This enables operators to route deep storage traffic through a single global MRAP endpoint for multi-region active-active deployments and regional failover scenarios. Plain bucket names are unaffected.

<hr> 

Key changed/added classes in this PR

- CloudObjectLocation — relaxed bucket validation to permit S3 Access Point ARNs
- S3Utils — added isS3Arn() and normalizeBucketName()
- S3DataSegmentPusherConfig — normalize bucket name on set
- S3LoadSpec — normalize bucket name on constructor

<hr>

This PR has:

This PR has: 

- [x] been self-reviewed. 
- [ ] ~~using the concurrency checklist~~ (no concurrency changes) 
- [ ] ~~added or updated version, license, or notice information in licenses.yaml~~ (no new dependencies) 

### DIFF
--- a/processing/src/main/java/org/apache/druid/data/input/impl/CloudObjectLocation.java
+++ b/processing/src/main/java/org/apache/druid/data/input/impl/CloudObjectLocation.java
@@ -27,6 +27,7 @@ import org.apache.druid.java.util.common.StringUtils;
 
 import java.net.URI;
 import java.util.Objects;
+import java.util.regex.Pattern;
 
 /**
  * Common type for 'bucket' and 'path' concept of cloud objects to allow code sharing between cloud specific
@@ -46,6 +47,8 @@ import java.util.Objects;
  */
 public class CloudObjectLocation
 {
+  private static final Pattern S3_ARN = Pattern.compile("^arn:(aws|aws-cn|aws-us-gov):s3:[a-z0-9-]*:\\d{12}:accesspoint[:/][A-Za-z0-9.-]+$");
+
   public static URI validateUriScheme(String scheme, URI uri)
   {
     if (!scheme.equalsIgnoreCase(uri.getScheme())) {
@@ -64,14 +67,20 @@ public class CloudObjectLocation
                  "bucket name cannot be null. Please verify if bucket name adheres to naming rules");
     this.path = Preconditions.checkNotNull(StringUtils.maybeRemoveLeadingSlash(path));
     Preconditions.checkArgument(
-        this.bucket.equals(StringUtils.urlEncode(this.bucket)),
-        "bucket must follow DNS-compliant naming conventions"
+      this.bucket.equals(StringUtils.urlEncode(this.bucket)) || isS3Arn(this.bucket),
+      "bucket must follow DNS-compliant naming conventions or be a valid S3 Access Point ARN"
+      + " or S3 Multi-Region Access Point ARN"
     );
   }
 
   public CloudObjectLocation(URI uri)
   {
     this(uri.getHost() != null ? uri.getHost() : uri.getAuthority(), uri.getPath());
+  }
+
+  private static boolean isS3Arn(String value)
+  {
+    return value != null && S3_ARN.matcher(value).matches();
   }
 
   /**
--- a/processing/src/main/java/org/apache/druid/data/input/impl/CloudObjectLocation.java
+++ b/processing/src/main/java/org/apache/druid/data/input/impl/CloudObjectLocation.java
@@ -47,7 +47,7 @@ import java.util.regex.Pattern;
  */
 public class CloudObjectLocation
 {
-  private static final Pattern S3_ARN = Pattern.compile("^arn:(aws|aws-cn|aws-us-gov):s3:[a-z0-9-]*:\\d{12}:accesspoint[:/][A-Za-z0-9.-]+$");
+  private static final Pattern S3_ARN = Pattern.compile("^arn:(aws|aws-cn|aws-us-gov):s3:[a-z0-9-]*:\\d{12}:accesspoint:[A-Za-z0-9.-]+$");
 
   public static URI validateUriScheme(String scheme, URI uri)
   {
@@ -69,7 +69,7 @@ public class CloudObjectLocation
     Preconditions.checkArgument(
       this.bucket.equals(StringUtils.urlEncode(this.bucket)) || isS3Arn(this.bucket),
       "bucket must follow DNS-compliant naming conventions or be a valid S3 Access Point ARN"
-      + " or S3 Multi-Region Access Point ARN"
+      + " or S3 Multi-Region Access Point ARN. Use colon form: arn:...:accesspoint:name"
     );
   }
 
--- a/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentPusherConfig.java
+++ b/extensions-core/s3-extensions/src/main/java/org/apache/druid/storage/s3/S3DataSegmentPusherConfig.java
@@ -45,7 +45,7 @@ public class S3DataSegmentPusherConfig
 
   public void setBucket(String bucket)
   {
-    this.bucket = bucket;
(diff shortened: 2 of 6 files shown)


processing/src/main/java/org/apache/druid/data/input/impl/CloudObjectLocation.java
```diff
@@ -64,8 +67,9 @@ public CloudObjectLocation(@JsonProperty("bucket") String bucket, @JsonProperty(
                  "bucket name cannot be null. Please verify if bucket name adheres to naming rules");
     this.path = Preconditions.checkNotNull(StringUtils.maybeRemoveLeadingSlash(path));
     Preconditions.checkArgument(
-        this.bucket.equals(StringUtils.urlEncode(this.bucket)),
-        "bucket must follow DNS-compliant naming conventions"
+      this.bucket.equals(StringUtils.urlEncode(this.bucket)) || isS3Arn(this.bucket),
```
FrankChen021, line 70: [P2] Normalize slash-form ARN buckets before accepting them

This now accepts both `accesspoint:` and `accesspoint/` ARN buckets, but stores the slash form unchanged. AWS documents access point/MRAP ARNs with the slash separator, and direct JSON `objects` or catalog bucket+path inputs can now construct `CloudObjectLocation` with that value. Later `toUri()` writes the raw bucket into `s3://%s/%s`; with `arn:...:accesspoint/bucket.mrap`, the access-point name becomes part of the URI path, so code that re-reads the URI via `getAuthority()` sees only `arn:...:accesspoint` as the bucket. The deep-storage config/loadSpec paths normalize this, but direct `CloudObjectLocation` construction does not. Please normalize slash-form S3 ARN buckets here, or reject slash-form ARNs from this class and require the internal colon form.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

processing/src/main/java/org/apache/druid/data/input/impl/CloudObjectLocation.java
```diff
@@ -60,20 +63,37 @@ public static URI validateUriScheme(String scheme, URI uri)
...
+      "bucket must follow DNS-compliant naming conventions or be a valid S3 Access Point ARN"
+      + " or S3 Multi-Region Access Point ARN"
     );
   }
 
+  private static String normalizeBucket(String bucket)
+  {
+    String trimmed = StringUtils.maybeRemoveTrailingSlash(bucket);
+
+    if (trimmed != null && isS3Arn(trimmed)) {
```
FrankChen021: [P2] Handle slash-form MRAP ARNs before URI splitting

This normalization only works when the full ARN is already passed as the bucket string. For a URI like `s3://arn:aws:s3::123456789123:accesspoint/bucket.mrap/path`, `java.net.URI` splits the value into authority `arn:aws:s3::123456789123:accesspoint` and path `/bucket.mrap/path`, so the bucket passed here does not match `S3_ARN` and the slash-form MRAP URI remains rejected or parsed with the access point name as part of the key. Please normalize this form before the URI constructor splits bucket/path, or add validation/tests that explicitly reject it.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

processing/src/main/java/org/apache/druid/data/input/impl/CloudObjectLocation.java
vivek807: Updated check

FrankChen021 (commented):
Follow-up handled: slash-form S3 Access Point and MRAP ARNs are now explicitly rejected by `CloudObjectLocation`, with coverage for both direct construction and URI parsing. Callers accepting slash form normalize it through `S3Utils.normalizeBucketName`.

---

PR #19613, opened 2026-06-22
minor:  add debug logs for diagnosing projection match failures

cecemei:
### Description

Debugging why `matchAggregateProjection` returns null is difficult today because there are 14 distinct rejection sites spread across the matching pipeline, and there's no visibility into which one fired without a debugger or global DEBUG logging.

This PR adds debug logging at INFO level if debug flag is on. When flag is unset, the same messages are emitted at DEBUG level.
                                                                                                                                                  
Rejection messages include the projection name and the specific reason (e.g., incompatible ordering, missing column, granularity mismatch), making it straightforward to identify which projection was rejected and why.                                                                                  
                                                                           

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/processing/src/main/java/org/apache/druid/query/QueryContexts.java
+++ b/processing/src/main/java/org/apache/druid/query/QueryContexts.java
@@ -126,7 +126,6 @@ public class QueryContexts
   public static final String NO_PROJECTIONS = "noProjections";
   public static final String FORCE_PROJECTION = "forceProjections";
   public static final String USE_PROJECTION = "useProjection";
-  public static final String PROJECTION_TRACE = "projectionTrace";
 
   // Unique identifier for the query, that is used to map the global shared resources (specifically merge buffers) to the
   // query's runtime
--- a/processing/src/main/java/org/apache/druid/query/QueryContexts.java
+++ b/processing/src/main/java/org/apache/druid/query/QueryContexts.java
@@ -126,6 +126,7 @@ public class QueryContexts
   public static final String NO_PROJECTIONS = "noProjections";
   public static final String FORCE_PROJECTION = "forceProjections";
   public static final String USE_PROJECTION = "useProjection";
+  public static final String PROJECTION_TRACE = "projectionTrace";
 
   // Unique identifier for the query, that is used to map the global shared resources (specifically merge buffers) to the
   // query's runtime
--- a/processing/src/main/java/org/apache/druid/segment/projections/Projections.java
+++ b/processing/src/main/java/org/apache/druid/segment/projections/Projections.java
@@ -26,6 +26,8 @@ import org.apache.druid.error.InvalidInput;
 import org.apache.druid.java.util.common.granularity.Granularities;
 import org.apache.druid.java.util.common.granularity.Granularity;
 import org.apache.druid.java.util.common.granularity.PeriodGranularity;
+import org.apache.druid.java.util.common.logger.Logger;
+import org.apache.druid.query.QueryContext;
 import org.apache.druid.query.QueryContexts;
 import org.apache.druid.query.aggregation.AggregatorFactory;
 import org.apache.druid.query.aggregation.FilteredAggregatorFactory;
@@ -67,6 +69,17 @@ import java.util.function.Function;
 
 public class Projections
 {
+  private static final Logger log = new Logger(Projections.class);
+
+  private static void logTrace(QueryContext context, String format, Object... args)
+  {
+    if (context.getBoolean(QueryContexts.PROJECTION_TRACE, false)) {
+      log.info(format, args);
+    } else {
+      log.debug(format, args);
+    }
+  }
+
   public static final String BASE_TABLE_PROJECTION_NAME = "__base";
 
   private static final String CLUSTER_GROUP_PREFIX = BASE_TABLE_PROJECTION_NAME + "$";
@@ -161,13 +174,16 @@ public class Projections
   )
   {
     if (!queryCursorBuildSpec.isCompatibleOrdering(projection.getOrderingWithTimeColumnSubstitution())) {
+      logTrace(queryCursorBuildSpec.getQueryContext(), "matchAggregateProjection: projection [%s] rejected — incompatible ordering", projection.getName());
       return null;
     }
     if (CollectionUtils.isNullOrEmpty(queryCursorBuildSpec.getPhysicalColumns())) {
+      logTrace(queryCursorBuildSpec.getQueryContext(), "matchAggregateProjection: projection [%s] rejected — no physical columns in query", projection.getName());
       return null;
     }
 
     if (isUnalignedInterval(projection, queryCursorBuildSpec, dataInterval)) {
+      logTrace(queryCursorBuildSpec.getQueryContext(), "matchAggregateProjection: projection [%s] rejected — unaligned interval", projection.getName());
       return null;
     }
     ProjectionMatchBuilder matchBuilder = new ProjectionMatchBuilder();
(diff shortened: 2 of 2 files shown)


processing/src/main/java/org/apache/druid/query/QueryContexts.java
```diff
@@ -126,6 +126,7 @@ public class QueryContexts
   public static final String NO_PROJECTIONS = "noProjections";
   public static final String FORCE_PROJECTION = "forceProjections";
   public static final String USE_PROJECTION = "useProjection";
+  public static final String PROJECTION_TRACE = "projectionTrace";
```
clintropolis: i wonder if instead of adding a new flag if this should just check `ENABLE_DEBUG` context flag

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -161,35 +174,42 @@ public static ProjectionMatch matchAggregateProjection(
   )
   {
     if (!queryCursorBuildSpec.isCompatibleOrdering(projection.getOrderingWithTimeColumnSubstitution())) {
+      logTrace(queryCursorBuildSpec.getQueryContext(), "matchAggregateProjection: projection [%s] rejected — incompatible ordering", projection.getName());
```
clintropolis: might be nice to print the ordering the `CursorBuildSpec` wanted and the ordering the projection provides

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -67,6 +69,17 @@
 
 public class Projections
 {
+  private static final Logger log = new Logger(Projections.class);
+
+  private static void logTrace(QueryContext context, String format, Object... args)
```
clintropolis, line 74: why not just pass the `CursorBuildSpec` into this method? it looks like all of the callers are getting it from there anyway

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
```diff
@@ -161,35 +174,42 @@ public static ProjectionMatch matchAggregateProjection(
...
     if (isUnalignedInterval(projection, queryCursorBuildSpec, dataInterval)) {
+      logTrace(queryCursorBuildSpec.getQueryContext(), "matchAggregateProjection: projection [%s] rejected — unaligned interval", projection.getName());
       return null;
     }
     ProjectionMatchBuilder matchBuilder = new ProjectionMatchBuilder();
 
     // match virtual columns first, which will populate the 'remapColumns' of the match builder
     matchBuilder = matchQueryVirtualColumns(projection, queryCursorBuildSpec, physicalColumnChecker, matchBuilder);
     if (matchBuilder == null) {
+      logTrace(queryCursorBuildSpec.getQueryContext(), "matchAggregateProjection: projection [%s] rejected — virtual column mismatch", projection.getName());
```
clintropolis, line 200: this and some of these other top level logs seem kind of redundant since the methods they call also log when returning nulls, but i guess we sort of need these here though because otherwise i'm not sure its obvious which caller of which match method was the failing one. 

This is probably ok for now since its just logs, but in the future we might consider collecting these failure reasons instead of logging them when they happen so that we can produce more concise messaging.

processing/src/main/java/org/apache/druid/query/QueryContexts.java
cecemei: i've thought about that as well, but decided not because debug=true already triggers a lot of things at once, e.x SQL planner rule, CalciteRulesManager in particular is very noisy when it logs planner rules. It'd be very noisy if someone is just trying to understand why a projection isn't matching.

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
cecemei: we could consider collecting failure reasons when forceProjection or useProjection is set in the future, and maybe add it as a system flag.

even though many messages are added here, only 1-2 lines would be logged so it should not be too verbose.

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
cecemei: a narrower input param would be more declarative? yes all callers are getting it from `CursorBuildSpec`.

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
cecemei: this is done

processing/src/main/java/org/apache/druid/query/QueryContexts.java
clintropolis: yea, i considered the noise briefly too, but then said to myself it was ok though because the noisy SQL logs are all on the broker, while these logs are only going to be on the historicals and realtime servers, per segment.

If you really want to add a new flag, i still think we should also do this logging if `debug` is set, so that way someone wanting "all the logs" doesn't have to set a bunch of different flags, so the `logTrace` method could check either `projectionTrace` or `debug`

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
clintropolis: i guess it was just jarring to see `logTrace(queryCursorBuildSpec.getQueryContext()...` like 20 times in the same file with all the stuff sourced from a `CursorBuildSpec`.

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
clintropolis: yea, i think its fine for now, though 'only 1-2 lines' is _per projection, per segment_ of the query, so its going to be pretty verbose anywhere that is touching more than a few segments if there are any projections at all.

processing/src/main/java/org/apache/druid/query/QueryContexts.java
cecemei: good point, i added the logging if debug is set

processing/src/main/java/org/apache/druid/segment/projections/Projections.java
cecemei: yea it would be suboptimal if logging for a lot of segments, maybe could be helpful in the case when only several segments out of all are missing proper projections, it gives some amount of visibility and operatibility to projection at least.

processing/src/main/java/org/apache/druid/query/QueryContexts.java
jtuglu1: > i wonder if instead of adding a new flag if this should just check ENABLE_DEBUG context flag

+1 to this. IMO, we should minimize the # of possible query context configuration combinations (there are frankly so many, some of which are documented and others not). My opinion: a user seeking debug information about a particular query should not need to go enable multiple query context flags to get the information they are looking for, _especially_ if they do not know what they are looking for (debugging). For debug, I'd rather get the kitchen sink and then filter out the noise myself manually.

jtuglu1 (commented):
While I like the idea, left a comment why I think we should just bake this behavior into the `debug` query context param.

FrankChen021 (commented):

---

processing/src/main/java/org/apache/druid/query/QueryContexts.java
cecemei: I guess if we filter by Projections class we could always reduce the noise. Updated this to use debug flag.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 1 |
| Total | 1 |

---

- processing/src/main/java/org/apache/druid/segment/projections/Projections.java:548 - [P3] Keep the existing public helper overload. This changes the public static matchQueryPhysicalColumn(...) helper from four arguments to five just to pass logging context. Downstream extensions or tests compiled against Projections.matchQueryPhysicalColumn(column, projection, checker, builder) will now fail even though the old behavior can be preserved by retaining the existing overload and delegating with QueryContext.empty() or similar. Please keep the old overload to avoid an avoidable source-compatibility break.

clintropolis (approved):

PR #19620, opened 2026-06-23
feat: partial load for clustered segments

clintropolis:
### Description
Wires up partial load support for clustered segments, which is mostly converting the existing non-partial clustered stuff to be on the partial stuff. I think as a follow-up I'll try to split out the 'clustered' implementations from the non-clustered stuff since it feels like a lot of 'if (clustered ...) do something' blocks and there is a bunch of stuff duplicated, but didn't seem appropriate to do in this PR.

### DIFF
--- a/server/src/test/java/org/apache/druid/segment/loading/SegmentLocalCacheManagerPartialAcquireTest.java
+++ b/server/src/test/java/org/apache/druid/segment/loading/SegmentLocalCacheManagerPartialAcquireTest.java
@@ -24,15 +24,19 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.jsontype.NamedType;
 import org.apache.druid.data.input.InputRow;
 import org.apache.druid.data.input.ListBasedInputRow;
+import org.apache.druid.data.input.MapBasedInputRow;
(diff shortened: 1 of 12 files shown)


server/src/test/java/org/apache/druid/segment/loading/SegmentLocalCacheManagerPartialAcquireTest.java
```diff
@@ -113,10 +118,17 @@ class SegmentLocalCacheManagerPartialAcquireTest
       new ListBasedInputRow(ROW_SIGNATURE, TIME.plusMinutes(3), ROW_SIGNATURE.getColumnNames(), Arrays.asList("b", 4L))
   );
 
+  // A second segment that is both clustered and carries an aggregate projection (the apache/druid#19599 combo): no
```
capistrant:
```suggestion
  // A second segment that is both clustered and carries an aggregate projection: no
```

capistrant (approved):
all I have is an ultra-nit. I don't know that anyone will care what apache PR was responsible here. since I didn't find anything else along the way either don't bother making the change or change it in some follow up.

capistrant (approved):

PR #19622, opened 2026-06-23
feat: introduce processing rate to max ratio as poll-idle-ratio replacement

Fly-Style:
**Summary**                                                                                                                                                                                              
                                                                                                                                                                                                       
Kafka's `poll-idle-ratio` is not the best possible metric to represent the idle cost for autoscaler - it reflects time spent polling, not whether the task has spare processing capacity.  Also, it is not supported for Kinesis, which is still in play in read-world productions.

This patch adds an alternative idle signal : `1 - (avgProcessingRate / maxObservedRate)`, gated behind a new opt-in flag `useUtilizationRatio` (default false, so existing deployments are unaffected).                                                   
                                                                                                                                                                                                       
- `CostBasedAutoScaler`: tracks a bounded watermark of the task's best-observed processing rate (`maxObservedRate`), feeding CostMetrics.                                                                
- `WeightedCostFunction`: when the flag is on, derives idle ratio from that utilization ratio instead of `pollIdleRatio`; falls back to `IDEAL_IDLE_RATIO` until a watermark sample exists (cold start).   
- `CostBasedAutoScalerConfig`: new useUtilizationRatio boolean, wired through builder/serde/equals/hashCode.                                                                                           
- `CostMetrics`: new nullable maxObservedRate field; old constructor kept (delegates with null) since three test call sites still use it.          

This PR has:
- [x] been self-reviewed.


embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
```diff
@@ -180,6 +180,54 @@ public void test_autoScaler_computesOptimalTaskCountAndProducesScaleUp()
     cluster.callApi().postSupervisor(kafkaSupervisorSpec.createSuspendedSpec());
   }
 
+  @Test
+  public void test_autoScaler_computesOptimalTaskCountAndProducesScaleUp_withUtilizationRatio()
+  {
+
+    final int lowInitialTaskCount = 1;
+    // This ensures tasks are busy processing (low idle ratio)
+    Executors.newSingleThreadExecutor().submit(() -> {
```
FrankChen021: [P2] Shut down the background publisher executor

The new test creates a single-thread ExecutorService and immediately drops it. That executor uses a non-daemon worker and is never shut down, so the embedded-test JVM can stay alive after the test returns; on timeout or failure it can also keep publishing into the topic while cleanup is running. Keep a reference and shut it down/cancel the Future in a finally block or use an existing managed executor.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
```diff
@@ -337,7 +398,9 @@ private int getCurrentTaskCount(String supervisorId)
     final String getSupervisorPath = StringUtils.format("/druid/indexer/v1/supervisor/%s", supervisorId);
     final KafkaSupervisorSpec supervisorSpec = cluster.callApi().serviceClient().onLeaderOverlord(
         mapper -> new RequestBuilder(HttpMethod.GET, getSupervisorPath),
-        new TypeReference<>(){}
+        new TypeReference<>()
+        {
+        }
```
kfaraz: Nit: extra change

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -103,6 +115,11 @@ public CostBasedAutoScaler(
     this.costFunction = new WeightedCostFunction();
     this.autoscalerExecutor = Execs.scheduledSingleThreaded("CostBasedAutoScaler-"
                                                             + StringUtils.encodeForFormat(spec.getId()));
+
+    // Bound the rate watermark to roughly one task's lifetime (in collectMetrics() ticks).
+    final long taskDurationMillis = supervisor.getIoConfig().getTaskDuration().getStandardSeconds() * 1000L;
```
kfaraz: If task duration is specified as less than 1s (e.g. in embedded tests), will this become 0? I think we should use `Duration.toMillis()` to be on the safe-side. Please double check this either way.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -103,6 +115,11 @@ public CostBasedAutoScaler(
...
+    this.lagGatedRateWindowSize = Math.max(1, (int) (taskDurationMillis / config.getScaleActionPeriodMillis()));
```
kfaraz: So, if task duration is 1 hour and scaling period is 15 mins, the window size would be only 4?
I feel like we should be looking at a much larger window if we are trying to determine the max, maybe even 100?
It probably doesn't have to be a function of the task duration, just gating the sample collection on the lag condition should be enough.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -85,6 +87,16 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
...
   private final WeightedCostFunction costFunction;
+  /**
+   * Recent {@code avgProcessingRate} readings, but only from ticks where lag was present.
+   * "Lag-gated" because lag is the condition that decides whether a reading counts as evidence
+   * of capacity at all: with no lag the task is only processing as fast as data arrives (an
+   * upstream supply number, not a capacity number); with lag present, demand isn't the
+   * bottleneck, so the rate observed is the task's own ceiling. The max of these samples
+   * becomes {@link CostMetrics#getMaxObservedRate()}.
+   */
+  private final EvictingQueue<Double> lagGatedRateSamples;
```
kfaraz:
```suggestion
  private final EvictingQueue<Double> lagGatedProcessingRateSamples;
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -85,6 +87,16 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
...
+  private final int lagGatedRateWindowSize;
```
kfaraz: This probably doesn't need to be a member field of this class.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -85,6 +87,16 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
   private final CostBasedAutoScalerConfig config;
   private final ScheduledExecutorService autoscalerExecutor;
   private final WeightedCostFunction costFunction;
+  /**
+   * Recent {@code avgProcessingRate} readings, but only from ticks where lag was present.
+   * "Lag-gated" because lag is the condition that decides whether a reading counts as evidence
+   * of capacity at all: with no lag the task is only processing as fast as data arrives (an
+   * upstream supply number, not a capacity number); with lag present, demand isn't the
+   * bottleneck, so the rate observed is the task's own ceiling. The max of these samples
+   * becomes {@link CostMetrics#getMaxObservedRate()}.
```
kfaraz: This can be simplified:

```suggestion
   * When there is no lag i.e. tasks are able to keep up with incoming data, the processing rate
   * is dictated by the rate of incoming data alone and does not denote the actual processing
   * capacity of the task.
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -91,7 +92,8 @@ public CostBasedAutoScalerConfig(
       @Nullable @JsonProperty("useTaskCountBoundariesOnScaleDown") Boolean useTaskCountBoundariesOnScaleDown,
       @Nullable @JsonProperty("minScaleUpDelay") Duration minScaleUpDelay,
       @Nullable @JsonProperty("minScaleDownDelay") Duration minScaleDownDelay,
-      @Nullable @JsonProperty("scaleDownDuringTaskRolloverOnly") Boolean scaleDownDuringTaskRolloverOnly
+      @Nullable @JsonProperty("scaleDownDuringTaskRolloverOnly") Boolean scaleDownDuringTaskRolloverOnly,
+      @Nullable @JsonProperty("useUtilizationRatio") Boolean useUtilizationRatio
```
kfaraz: We should invert the meaning of this flag and name it `usePollIdleRatio` (true by default) since that is a concept already being used in this auto-scaler. Adding another new terminology (utilization ratio) makes things more confusing.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -282,6 +285,15 @@ public boolean isScaleDownOnTaskRolloverOnly()
     return scaleDownDuringTaskRolloverOnly;
   }
 
+  /**
+   * Opt-in flag for replacing default idle-ratio with the rate-based utilization idle ratio.
+   */
+  @JsonProperty("useUtilizationRatio")
+  public boolean shouldUseUtilizationRatio()
```
kfaraz: Use the prefix `is` for boolean configs to align with the rest of the druid code. You can also skip the String literal inside the `@JsonProperty` annotation.

```suggestion
  public boolean isUseUtilizationRatio()
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -156,6 +157,19 @@ public CostResult computeCost(
     return new CostResult(cost, lagCost, weightedIdleCost);
   }
 
+  /**
+   * Derives the current idle ratio from measured utilization ({@code avgProcessingRate / maxObservedRate}).
+   */
+  static double estimateIdleFromUtilization(CostMetrics metrics)
```
kfaraz: Maybe this method should exist in `CostMetrics` itself?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -156,6 +157,19 @@ public CostResult computeCost(
...
+  {
+    final Double maxObservedRate = metrics.getMaxObservedRate();
+    if (maxObservedRate == null || maxObservedRate <= 0) {
+      return IDEAL_IDLE_RATIO;
```
kfaraz: So, if there has been no lag for a while, we will assume that we are already in the optimal usage band?
This would mean that scale down would not happen when, ideally, it should.

I wonder if it would be better to collect samples of `avgProcessingRate` even when there is no lag, and simply use the max to determine the max observed rate.

kfaraz (commented):
Conceptually, this makes sense. Haven't reviewed the tests yet.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
Fly-Style: I will change the prefix, but I may assume that you would agree that `is...` sounds awful here 😁

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: We're receiving new metrics once per `scaleActionPeriodMillis`, but yeah, we probably should increase this window size. 50 or 100 looks like a decent size.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: This is a slippery place, but after short consideration I'd agree with you. Reworking that part.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Removed because it has been massively simplified.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
Fly-Style: Renamed in other way due to new circumstances.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
kfaraz: Yes, I know 😛

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
kfaraz: Thanks! We can tune it further based on how it performs in real clusters.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
kfaraz: Please rename this method to `estimateIdleRatioFromProcessingRate()` and move it to the class `CostMetrics`.

kfaraz (approved):
One final non-blocking comment. Rest looks good.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -156,6 +157,16 @@ public CostResult computeCost(
     return new CostResult(cost, lagCost, weightedIdleCost);
   }
 
+  /**
+   * Derives the current idle ratio from measured utilization ({@code avgProcessingRate / maxObservedRate}).
+   */
+  static double estimateIdleFromUtilization(CostMetrics metrics)
+  {
+    final Double maxObservedRate = metrics.getMaxObservedRate();
+    final double utilization = Math.min(1.0, metrics.getAvgProcessingRate() / maxObservedRate);
```
FrankChen021: [P2] Restore fallback before dividing by maxObservedRate

When `usePollIdleRatio` is false, this method derives idle ratio from `maxObservedRate`, but `CostMetrics` still allows that value to be null and the added tests explicitly cover null and zero watermarks. Dividing by the boxed value now throws on null, and a zero watermark yields `Infinity`/`NaN` behavior instead of the intended `IDEAL_IDLE_RATIO` cold-start fallback. Please keep the `maxObservedRate == null || maxObservedRate <= 0` guard before this division.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostMetrics.java
```diff
@@ -136,6 +115,16 @@ public Double getMaxObservedRate()
     return maxObservedRate;
   }
 
+
+  /**
+   * Derives the current idle ratio from measured utilization ({@code avgProcessingRate / maxObservedRate}).
+   */
+  double estimateIdleRatioFromProcessingRate(CostMetrics metrics)
```
kfaraz: We don't need to pass the `CostMetrics` here.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -78,13 +80,21 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
    */
   static final int MAX_IDLENESS_PARTITION_LAG = 10_000;
 
+  /**
+   * The number of most recent processing rate samples to maintain for auto-scaling decisions
+   * in case when {@link CostBasedAutoScalerConfig#usePollIdleRatio} is disabled.
+   */
+  static final int RATE_WINDOW_SIZE = 50;
```
kfaraz:
```suggestion
  static final int PROCESSING_RATE_WINDOW_SIZE = 50;
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostMetrics.java
```diff
@@ -105,6 +108,26 @@ public double getAvgProcessingRate()
     return avgProcessingRate;
   }
 
+  public Double getMaxObservedRate()
```
kfaraz, line 114:
```suggestion
  public double getMaxObservedRate()
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostMetrics.java
```diff
@@ -105,6 +108,26 @@ public double getAvgProcessingRate()
...
+
+
+  /**
+   * Derives the current idle ratio from measured utilization ({@code avgProcessingRate / maxObservedRate}).
+   */
+  double estimateIdleRatioFromProcessingRate(CostMetrics metrics)
+  {
+    if (maxObservedRate <= 0) {
+      // No throughput baseline yet (e.g. only zero-rate samples at startup). Returning a negative
+      // value routes computeCost to its neutral idle branch instead of producing NaN from 0/0.
```
kfaraz: How the caller (i.e. computeCost) deals with a negative value should not be a concern of this class. We can remove that from the comment here.
The javadoc of this method should just call out the case when a negative value is returned.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostMetrics.java
```diff
@@ -120,7 +143,8 @@ public boolean equals(Object o)
            && partitionCount == that.partitionCount
            && Double.compare(that.pollIdleRatio, pollIdleRatio) == 0
            && taskDurationSeconds == that.taskDurationSeconds
-           && Double.compare(that.avgProcessingRate, avgProcessingRate) == 0;
+           && Double.compare(that.avgProcessingRate, avgProcessingRate) == 0
+           && Objects.equals(that.maxObservedRate, maxObservedRate);
```
kfaraz, line 150:
```suggestion
           && Double.compare(that.maxObservedRate, maxObservedRate) == 0;
```

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerTest.java
```diff
@@ -563,6 +571,78 @@ public void testScalingActionSkippedWhenMovingAverageRateUnavailable()
...
+    when(supervisor.getStats()).thenReturn(buildTaskStatsForRate(500.0));
+    CostBasedAutoScaler defaultScaler = new CostBasedAutoScaler(supervisor, defaultConfig, spec, emitter);
+    Assert.assertNull(
+        "With usePollIdleRatio=true (the default), samples are never collected",
+        defaultScaler.collectMetrics().getMaxObservedRate()
+    );
+
+    // Disabling usePollIdleRatio switches the idle cost to processing-rate utilization,
+    // which requires tracking a watermark of observed rates.
+    CostBasedAutoScalerConfig utilizationConfig = CostBasedAutoScalerConfig.builder()
```
kfaraz: Please remove usages of the word utilization as it is ambiguous.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerTest.java
```diff
@@ -563,6 +571,78 @@ public void testScalingActionSkippedWhenMovingAverageRateUnavailable()
...
+                                                                           .taskCountMin(1)
+                                                                           .enableTaskAutoScaler(true)
+                                                                           .usePollIdleRatio(false)
+                                                                           .build();
+    when(supervisor.getStats()).thenReturn(
+        buildTaskStatsForRate(500.0),
+        buildTaskStatsForRate(9000.0),
+        buildTaskStatsForRate(300.0)
+    );
+    CostBasedAutoScaler utilizationScaler = new CostBasedAutoScaler(supervisor, utilizationConfig, spec, emitter);
```
kfaraz: Maybe rename this to `autoScalerWithoutPollIdleRatio` or similar.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerTest.java
```diff
@@ -563,6 +571,78 @@ public void testScalingActionSkippedWhenMovingAverageRateUnavailable()
     );
   }
 
+  @Test
+  public void testCollectMetricsTracksRateWatermarkOnlyWhenPollIdleRatioDisabled()
```
kfaraz:
```suggestion
  public void testCollectMetricsTracksMaxProcessingRateOnlyWhenPollIdleRatioDisabled()
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -282,6 +285,16 @@ public boolean isScaleDownOnTaskRolloverOnly()
     return scaleDownDuringTaskRolloverOnly;
   }
 
+  /**
+   * If true (the default), idle cost is derived from the poll idle ratio reported by tasks.
+   * If false, idle cost is derived from processing-rate utilization instead.
+   */
+  @JsonProperty("usePollIdleRatio")
```
kfaraz: Not really needed
```suggestion
  @JsonProperty
```

Fly-Style:
@kfaraz all comments addressed; one note - I rolled back to `Double maxObservedRate`.

kfaraz (approved):
Looks good!

FrankChen021 (commented):
I have reviewed the updated code for correctness, edge cases, concurrency, lifecycle, security/data loss/API compatibility, and missing-test risks; no issues found.

---

PR #19624, opened 2026-06-24
feat: add query/segments/count metric to data nodes

jtuglu1:
### Description

Allows for quick queries to understand how many segments a given query issued scans for against a given data node. Better than counting the unique query/segment/time or similar, especially if you are sampling higher volume metrics like query/segment/time, but still want to have quick way to know stats about # of segments queried per node.

#### Release note

Add add query/segments/count metric metric to data nodes (peon + historical).

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/docs/operations/metrics.md
+++ b/docs/operations/metrics.md
@@ -64,7 +64,7 @@ Most metric values reset each emission period, as specified in `druid.monitoring
 |`query/failed/count`|Number of failed queries.|This metric is only available if the `QueryCountStatsMonitor` module is included.| |
 |`query/interrupted/count`|Number of queries interrupted due to cancellation.|This metric is only available if the `QueryCountStatsMonitor` module is included.| |
 |`query/timeout/count`|Number of timed out queries.|This metric is only available if the `QueryCountStatsMonitor` module is included.| |
-|`query/segments/count`|This metric is not enabled by default. See the `QueryMetrics` Interface for reference regarding enabling this metric. Number of segments that will be touched by the query. In the broker, it makes a plan to distribute the query to realtime tasks and historicals based on a snapshot of segment distribution state. If there are some segments moved after this snapshot is created, certain historicals and realtime tasks can report those segments as missing to the broker. The broker will resend the query to the new servers that serve those segments after move. In this case, those segments can be counted more than once in this metric.||Varies|
+|`query/segments/count`|Number of segments that will be touched by the query. The Broker makes a plan to distribute the query to realtime tasks and historicals based on a snapshot of segment distribution state. If there are some segments moved after this snapshot is created, certain historicals and realtime tasks can report those segments as missing to the Broker. The Broker will resend the query to the new servers that serve those segments after move. In this case, those segments can be counted more than once in this metric.||Varies|
 |`query/priority`|Assigned lane and priority, only if Laning strategy is enabled. Refer to [Laning strategies](../configuration/index.md#laning-strategies)|`lane`, `dataSource`, `type`|0|
 |`sqlQuery/time`|Milliseconds taken to complete a SQL query.|`id`, `nativeQueryIds`, `dataSource`, `remoteAddress`, `success`, `engine`, `statusCode`|< 1s|
 |`sqlQuery/planningTimeMs`|Milliseconds taken to plan a SQL to native query.|`id`, `nativeQueryIds`, `dataSource`, `remoteAddress`, `success`, `engine`| |
@@ -106,6 +106,7 @@ Most metric values reset each emission period, as specified in `druid.monitoring
 |Metric|Description|Dimensions|Normal value|
 |------|-----------|----------|------------|
 |`query/time`|Milliseconds taken to complete a query.|<p>Common: `dataSource`, `type`, `interval`, `hasFilters`, `duration`, `context`, `remoteAddress`, `id`, `statusCode`.</p><p> Aggregation Queries: `numMetrics`, `numComplexMetrics`.</p><p> GroupBy: `numDimensions`.</p><p> TopN: `threshold`, `dimension`.</p>|< 1s|
+|`query/segments/count`|Number of segments this Historical scans for a query.|<p>Common: `dataSource`, `type`, `interval`, `hasFilters`, `duration`, `context`, `remoteAddress`, `id`.</p><p> Aggregation Queries: `numMetrics`, `numComplexMetrics`.</p><p> GroupBy: `numDimensions`.</p><p> TopN: `threshold`, `dimension`.</p>|Varies|
 |`query/segment/time`|Milliseconds taken to query individual segment. Includes time to page in the segment from disk.|`id`, `status`, `segment`, `vectorized`.|several hundred milliseconds|
 |`query/wait/time`|Milliseconds spent waiting for a segment to be scanned.|`id`, `segment`|< several hundred milliseconds|
 |`segment/scan/pending`|Number of segments in queue waiting to be scanned.||Close to 0|
@@ -140,6 +141,7 @@ to represent the task ID are deprecated and will be removed in a future release.
 |Metric|Description|Dimensions|Normal value|
 |------|-----------|----------|------------|
 |`query/time`|Milliseconds taken to complete a query.|<p>Common: `dataSource`, `type`, `interval`, `hasFilters`, `duration`, `context`, `remoteAddress`, `id`, `statusCode`.</p><p> Aggregation Queries: `numMetrics`, `numComplexMetrics`.</p><p> GroupBy: `numDimensions`.</p><p> TopN: `threshold`, `dimension`.</p>|< 1s|
+|`query/segments/count`|Number of segments this Peon scans for a query.|<p>Common: `dataSource`, `type`, `interval`, `hasFilters`, `duration`, `context`, `remoteAddress`, `id`.</p><p> Aggregation Queries: `numMetrics`, `numComplexMetrics`.</p><p> GroupBy: `numDimensions`. </p><p>TopN: `threshold`, `dimension`.</p>|Varies|
 |`query/wait/time`|Milliseconds spent waiting for a segment to be scanned.|`id`, `segment`|several hundred milliseconds|
 |`segment/scan/pending`|Number of segments in queue waiting to be scanned.||Close to 0|
 |`segment/scan/active`|Number of segments currently scanned. This metric also indicates how many threads from `druid.processing.numThreads` are currently being used.||Close to `druid.processing.numThreads`|
--- a/processing/src/main/java/org/apache/druid/query/QueryMetrics.java
+++ b/processing/src/main/java/org/apache/druid/query/QueryMetrics.java
@@ -331,7 +331,7 @@ public interface QueryMetrics<QueryType extends Query<?>>
   /**
    * Registers the {@code query/segments/count} metric, the number of segments touched by the query.
    *
-   * Emitted once per query. The value's meaning depends on the emitting process:
+   * Emitted once per query. The meaning of the metric depends on the emitting process:
    * <ul>
    *   <li>On the Broker, it is the number of segments in the planned query distribution (a snapshot-based
    *       count that may double-count segments that moved and were re-fetched).</li>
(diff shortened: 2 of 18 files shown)


FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

Reviewed 14 of 14 changed files.

---

server/src/main/java/org/apache/druid/server/ServerManager.java
```diff
@@ -639,6 +639,7 @@ public Sequence<T> run(QueryPlus<T> queryPlus, ResponseContext responseContext)
             segmentsBundle,
             segmentMapFn
         );
+        queryPlus.getQueryMetrics().reportLocalSegmentCount(segmentReferences.size()).emit(emitter);
```
FrankChen021: [P2] Register segment references before emitting the new metric

This emits after getOrLoadBundleSegments returns but before closer.registerAll(segmentReferences). If reportLocalSegmentCount or the emitter throws, the catch block closes only the still-empty closer, so the acquired SegmentReference objects are leaked. Register the references with the closer before emitting this metric, or otherwise guarantee the references are closed on emission failure.

docs/operations/metrics.md
```diff
@@ -140,6 +141,7 @@ to represent the task ID are deprecated and will be removed in a future release.
 |Metric|Description|Dimensions|Normal value|
 |------|-----------|----------|------------|
 |`query/time`|Milliseconds taken to complete a query.|<p>Common: `dataSource`, `type`, `interval`, `hasFilters`, `duration`, `context`, `remoteAddress`, `id`, `statusCode`.</p><p> Aggregation Queries: `numMetrics`, `numComplexMetrics`.</p><p> GroupBy: `numDimensions`.</p><p> TopN: `threshold`, `dimension`.</p>|< 1s|
+|`query/segment/count`|Number of segments this Peon scans for a query. This is local to this data node; the Broker metric `query/segments/count` reports the distributed query plan.|<p>Common: `dataSource`, `type`, `interval`, `hasFilters`, `duration`, `context`, `remoteAddress`, `id`.</p><p> Aggregation Queries: `numMetrics`, `numComplexMetrics`.</p><p> GroupBy: `numDimensions`. </p><p>TopN: `threshold`, `dimension`.</p>|Varies|
```
FrankChen021: [P2] Peon metric is documented but not emitted

The PR documents query/segment/count for Peons, but the only new emission is in ServerManager. Peon/realtime queries route through PeonAppenderatorsManager or UnifiedIndexerAppenderatorsManager into SinkQuerySegmentWalker, which still only emits segment time, wait time, and segmentAndCache time. As a result the new metric is absent for Real-time/Peon data nodes despite the docs and emitter allowlists; add equivalent emission in the sink walker or remove the Peon entry.

docs/operations/metrics.md
```diff
@@ -106,6 +106,7 @@ Most metric values reset each emission period, as specified in `druid.monitoring
 |Metric|Description|Dimensions|Normal value|
 |------|-----------|----------|------------|
 |`query/time`|Milliseconds taken to complete a query.|<p>Common: `dataSource`, `type`, `interval`, `hasFilters`, `duration`, `context`, `remoteAddress`, `id`, `statusCode`.</p><p> Aggregation Queries: `numMetrics`, `numComplexMetrics`.</p><p> GroupBy: `numDimensions`.</p><p> TopN: `threshold`, `dimension`.</p>|< 1s|
+|`query/segment/count`|Number of segments this Historical scans for a query. This is local to this data node; the Broker metric `query/segments/count` reports the distributed query plan.|<p>Common: `dataSource`, `type`, `interval`, `hasFilters`, `duration`, `context`, `remoteAddress`, `id`.</p><p> Aggregation Queries: `numMetrics`, `numComplexMetrics`.</p><p> GroupBy: `numDimensions`.</p><p> TopN: `threshold`, `dimension`.</p>|Varies|
```
kfaraz: Ah, the similar metric names `query/segment/count` and `query/segments/count` (with an 'S') can be confusing. I wish the Broker metric name reflected that it was a plan-only metric.

Nit: For the use case in this PR though, I feel that `query/segments/count` probably makes more sense since `query/segment/time` reflects time taken to scan a single segment and not the total time.

kfaraz: Btw, do we even need to mention the Broker metric here? It seems unrelated despite the similar names. The disambiguation, if needed, may be done in Broker metrics section itself.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/SinkQuerySegmentWalker.java
```diff
@@ -470,6 +474,8 @@ private static class SinkMetricsEmittingQueryRunner<T> implements QueryRunner<T>
     private final Set<String> metricsToCompute;
     @Nullable
     private final String segmentId;
+    @Nullable
+    private final Integer localSegmentCount;
```
kfaraz: What does the prefix `local` signify?

server/src/main/java/org/apache/druid/segment/realtime/appenderator/SinkQuerySegmentWalker.java
```diff
@@ -452,7 +455,8 @@ public static String makeHydrantCacheIdentifier(final SegmentId segmentId, final
   }
 
   /**
-   * This class is responsible for emitting query/segment/time, query/wait/time and query/segmentAndCache/Time metrics for a Sink.
+   * This class is responsible for emitting query/segment/time, query/wait/time, query/segmentAndCache/Time and
+   * query/segment/count metrics for a Sink.
```
kfaraz: We should also call out that unlike other metrics emitted here, `query/segment/count` is emitted only when `segmentId` is non-null, since it is a total count.

server/src/test/java/org/apache/druid/server/coordination/ServerManagerTest.java
```diff
@@ -482,6 +483,29 @@ public void testGetQueryRunnerForIntervals_whenTimelineIsMissingReturningNoopQue
...
+
+    final Interval interval = Intervals.of("P2d/2011-04-02");
+    final SearchQuery query = searchQuery("test", interval, Granularities.DAY);
+
+    serverManager.getQueryRunnerForIntervals(query, ImmutableList.of(interval))
+                 .run(QueryPlus.wrap(query))
+                 .toList();
+
+    Assert.assertEquals(1, stubServiceEmitter.getMetricEventCount(DefaultQueryMetrics.QUERY_SEGMENT_COUNT));
+    Assert.assertEquals(2L, stubServiceEmitter.getLatestMetricEventValue(DefaultQueryMetrics.QUERY_SEGMENT_COUNT));
```
kfaraz: Nit: You may also use `stubServiceEmitter.verifyEmitted` and `verifyValue`.

kfaraz (commented):
Looks good, left some minor comments.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/SinkQuerySegmentWalker.java
jtuglu1: local to this data node

kfaraz: I feel like that is implicit, since the data node is not aware of segments from other nodes anyway.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/SinkQuerySegmentWalker.java
```diff
@@ -354,6 +354,8 @@ public <T> QueryRunner<T> getQueryRunnerForSegments(final Query<T> query, final
         );
       }
 
+      final int localSegmentCount = allSegmentReferences.size();
```
FrankChen021: [P2] Count realtime sinks instead of hydrants

This counts `allSegmentReferences`, but those references are hydrant-level: `Sink.acquireSegmentReferences` adds one `SinkSegmentReference` per `FireHydrant`. After a realtime segment has multiple hydrants, for example after persist/swap or with both persisted and current hydrants, a query over one Druid segment emits `query/segment/count` as the hydrant count instead of the segment/Sink count. That disagrees with the metric's documented segment-level meaning and with the existing per-segment metrics that aggregate by `sinkSegmentId`; please count distinct queried sinks/descriptors and add a multi-hydrant test.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

processing/src/main/java/org/apache/druid/query/QueryMetrics.java
```diff
@@ -333,6 +333,14 @@ default void filterBundle(FilterBundle.BundleInfo bundleInfo)
    */
   QueryMetrics<QueryType> reportQueriedSegmentCount(long segmentCount);
 
+  /**
+   * Registers the number of segments scanned by this data node for the query.
+   */
+  default QueryMetrics<QueryType> reportLocalSegmentCount(long segmentCount)
```
kfaraz: @jtuglu1 , could you please add some info in the javadoc of the existing `reportQueriedSegmentCount()` metric to clearly distinguish that method from `reportLocalSegmentCount()`?

processing/src/main/java/org/apache/druid/query/QueryMetrics.java
```diff
@@ -329,7 +329,16 @@ default void filterBundle(FilterBundle.BundleInfo bundleInfo)
   QueryMetrics<QueryType> reportQueryBytes(long byteCount);
 
   /**
-   * Registers "segments queried count" metric.
+   * Registers the {@code query/segments/count} metric, the number of segments touched by the query.
+   *
+   * Emitted once per query. The value's meaning depends on the emitting process:
```
kfaraz:
```suggestion
   * Emitted once per query. The meaning of the metric depends on the emitting process:
```

processing/src/main/java/org/apache/druid/query/QueryMetrics.java
```diff
@@ -329,7 +329,16 @@ default void filterBundle(FilterBundle.BundleInfo bundleInfo)
...
+   * <ul>
+   *   <li>On the Broker, it is the number of segments in the planned query distribution (a snapshot-based
+   *       count that may double-count segments that moved and were re-fetched).</li>
+   *   <li>On a data node (Historical, Peon/realtime), it is the number of segments that node actually scanned
+   *       for the query.</li>
+   * </ul>
+   * The two are disambiguated by the emitting service/host, not by the metric name.
```
kfaraz, line 341: Nit: Do we really want to overload the definition of this metric?
We may even take this opportunity to use a different metric name for the number of segments targeted by the query plan.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/SinkQuerySegmentWalker.java
```diff
@@ -544,6 +579,20 @@ public void after(boolean isDone, Throwable thrown)
                     log.error(e, "Failed to emit metrics for segment[%s]", segmentAndMetrics.getKey());
                   }
                 }
+
+                if (segmentCount != null) {
```
kfaraz, line 584:
```suggestion
                // Emit total segmentCount once per query
                if (segmentCount != null) {
```

Do we also want to add a validation on `segmentId == null` here to ensure that we don't accidentally call this more than once?

kfaraz (approved):
Looks good, left minor non-blocking suggestions.

@jtuglu1 , please avoid force pushing commits after a review has been done as it makes subsequent reviews difficult 🙂

processing/src/main/java/org/apache/druid/query/QueryMetrics.java
jtuglu1: Generally metric renaming is a pain since we need to break user's metric setups. I feel like they both are measuring the same thing for now, so we can potentially address in a follow-up to rename the broker-side ones.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/SinkQuerySegmentWalker.java
jtuglu1: The if (segmentCount != null) block is already in the else branch of the (segmentId != null guard) above

PR #19625, opened 2026-06-24
fix: fix broker segment metadata cache refresh

jtuglu1:
### Description

Brokers can maintain a schema cache via segment metadata queries. Currently, if any of these queries timeout, the remaining queries are aborted until the next refresh. If you have a huge datasource delta (think 500k+ segments being scanned), such a query can fail/timeout and cause other unrelated datasources' broker schema discovery to fail. Without centralized schema through coordinator, there is no intra-datasource atomicity guarantee w.r.t schema discovery (it is just ASAP), so decoupling this error dependency and instead emiting a metric per datasource when failures occur. 

Introduces `segment/schemaCache/refresh/failed` metric with a dataSource dimension, emitted when a refresh fails.  Can alternatively just aggregate and emit at the end. Also open to keeping this a warning/error log. 

#### Release note

Fix broker segment metadata cache refresh

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/segment/metadata/AbstractSegmentMetadataCache.java
+++ b/server/src/main/java/org/apache/druid/segment/metadata/AbstractSegmentMetadataCache.java
@@ -45,6 +45,7 @@ import org.apache.druid.java.util.emitter.service.ServiceEmitter;
 import org.apache.druid.java.util.emitter.service.ServiceMetricEvent;
 import org.apache.druid.query.DruidMetrics;
 import org.apache.druid.query.QueryContexts;
+import org.apache.druid.query.QueryInterruptedException;
 import org.apache.druid.query.TableDataSource;
 import org.apache.druid.query.metadata.metadata.AllColumnIncluderator;
 import org.apache.druid.query.metadata.metadata.ColumnAnalysis;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/druid/segment/metadata/AbstractSegmentMetadataCache.java
```diff
@@ -717,8 +718,29 @@ public Set<SegmentId> refreshSegments(final Set<SegmentId> segments) throws IOEx
...
 
+    // Refresh each dataSource independently so that a failure (e.g. a SegmentMetadataQuery timeout on
+    // one dataSource) does not abort the whole cycle and starve the remaining dataSources of schema.
     for (Map.Entry<String, TreeSet<SegmentId>> entry : segmentMap.entrySet()) {
-      updatedSegmentIds.addAll(refreshSegmentsForDataSource(entry.getKey(), entry.getValue()));
+      final String dataSource = entry.getKey();
+      try {
+        updatedSegmentIds.addAll(refreshSegmentsForDataSource(dataSource, entry.getValue()));
+      }
+      catch (QueryInterruptedException e) {
```
FrankChen021, line 728: [P2] Do not treat every QueryInterruptedException as shutdown

This catch rethrows all QueryInterruptedException instances before the per-datasource failure handler. In Druid, QueryInterruptedException is also used as a generic client-side wrapper for failed queries, including remote QueryExceptions with unknown/runtime failures, not only actual thread interruption. When one datasource's SegmentMetadataQuery fails through that wrapper, refreshSegments still aborts the whole cycle, later datasources are skipped, and no segment/schemaCache/refresh/failed metric is emitted for the offending datasource. Please distinguish real local interruption/cancellation from ordinary query failure, for example by checking the cause/errorCode, and let non-interruption query failures follow the isolated datasource path.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

server/src/main/java/org/apache/druid/segment/metadata/AbstractSegmentMetadataCache.java
```diff
@@ -717,13 +718,42 @@ public Set<SegmentId> refreshSegments(final Set<SegmentId> segments) throws IOEx
...
+        // QueryInterruptedException also wraps ordinary query failures, not just interruption
+        if (e.getCause() instanceof InterruptedException) {
+          Thread.currentThread().interrupt();
```
clintropolis, line 732: does this not record the failure on purpose? i guess to not emit a metric for shutdown? maybe worth a comment

clintropolis (approved):

server/src/main/java/org/apache/druid/segment/metadata/AbstractSegmentMetadataCache.java
jtuglu1: Yeah – I didn't want to emit a failure on broker shutdown

PR #19627, opened 2026-06-24
feat: Add mergeBuffer/maxSpillProximity metric for groupBy spill diagnosis

aho135:
### Description

When a groupBy query runs, `ConcurrentGrouper` divides the single acquired merge buffer into `druid.processing.numThreads` equal slices (`sliceSize = capacity / numThreads`) and gives one slice to each processing thread. A query spills to disk as soon as its **fullest single slice** fills — at roughly `sizeBytes / numThreads`, which can be far below the configured `druid.processing.buffer.sizeBytes`.

The existing metrics do not let an operator see this. `mergeBuffer/maxBytesUsed` is a per-query **sum across slices**, further discounted by the hash-table load factor, so it never approaches `sizeBytes` even while queries are actively spilling — making it impossible to compare against `druid.processing.buffer.sizeBytes` or to reason about spill pressure.

Concretely, an operator with `sizeBytes = 125 MiB` and `numThreads = 240` (slices ≈ 546 KiB) saw `groupBy/spilledQueries` climbing while `mergeBuffer/maxBytesUsed` sat around ~60 MB, which looks contradictory until you account for slicing.

#### Change

This PR adds **`mergeBuffer/maxSpillProximity`**, a dimensionless gauge in `[0.0, 1.0]` that reports how close a groupBy query came to spilling to disk:

```
maxSpillProximity = MAX over queries of
                      MAX over that query's slices of
                        peak( size / terminalRegrowthThreshold )     // clamped to [0, 1]
```

- It is computed **per-slice and maxed** (never summed), because the fullest slice is what actually triggers a spill.
- It is **bucket-count based**, measured on the exact condition a `BufferHashGrouper` spills on: it can no longer allocate a new bucket. The denominator is the **terminal-level `regrowthThreshold`** — the bucket-count ceiling of the largest table the slice can grow to before it must spill. Because the table grows by doubling and `size` only grows (preserved across rehash), the ratio is **monotonic** and rises smoothly to exactly **`1.0` at the real spill point** (a bucket allocation is rejected, or the table is parked at its terminal threshold). Dividing by the *current* level's threshold instead produces a sawtooth — it resets to ~0.5 on every doubling and climbs back to ~1.0 before the next — which would pin the lifetime max near `(N-1)/N` for any query that grows even once, regardless of remaining headroom.
- Being bucket-count based, the value is independent of bucket width, offset-list overhead, and integer truncation in the arena-size calculation.
- **Limit push-down queries read `0.0`.** A `LIMIT` query without an aggregator-dependent sort uses `LimitedBufferHashGrouper`, whose two-sub-buffer `AlternatingByteBufferHashTable` heap-trims to the top-`limit` entries and **never disk-spills** via that path. Its sub-buffer refills to the threshold on every swap, so it is deliberately excluded from fill-proximity recording (it would otherwise saturate near `(T-1)/T ≈ 1.0` and read as a perpetual near-spill). Only a genuine bucket rejection on that table would report `1.0`.

Operators can read `mergeBuffer/maxSpillProximity` alongside `groupBy/spilledQueries`: a value near `1.0` means slices are saturating, and the fix is to widen each slice by raising `druid.processing.buffer.sizeBytes` or lowering `druid.processing.numThreads`.

#### Changed files
- `ByteBufferHashTable` — track the lifetime-max `size / terminalRegrowthThreshold` ratio (`getMaxSpillProximity()`), pinned to `1.0` at the real spill trigger and preserved across `reset()`. `computeSpillRegrowthThreshold()` derives the fixed terminal denominator by replaying the arena geometry (no allocation, memoized); a `recordsFillProximity()` hook lets heap-trim tables opt out of fill recording.
- `LimitedBufferHashGrouper` — its `AlternatingByteBufferHashTable` overrides `recordsFillProximity()` to `false` so heap-trim swaps are not misreported as spills; a genuine bucket rejection still pins `1.0`.
- `AbstractBufferHashGrouper` — expose `getMaxSpillProximity()` (0.0 when the grouper was never initialized).
- `GroupByStatsProvider` — keep the per-slice max `getSpillProximity()` (clamped to `[0,1]`, `NaN` ignored), aggregated as a max across queries.
- `SpillingGrouper` — report each slice's peak proximity in `close()` (only when initialized).
- `GroupByStatsMonitor` — emit `mergeBuffer/maxSpillProximity`.
- `docs/operations/metrics.md` — document the new metric and clarify the slicing semantics of `mergeBuffer/bytesUsed` and `mergeBuffer/maxBytesUsed`.

<hr>

This PR has:

- [x] been self-reviewed.
- [x] a release note entry in the PR description. (New metric `mergeBuffer/maxSpillProximity`; no behavior or config changes.)
- [x] added integration tests. (N/A — metric is covered by unit tests.)


aho135:
Hey @GWphua! We were using the `mergeBuffer/maxBytesUsed` metric for tuning mergeBuffer size but found that it wasn't a very accurate indicator for spilling. Let me know if you have any thoughts on this. Thanks!

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
```diff
@@ -224,9 +266,24 @@ public void mergeBufferAcquisitionTime(long delay)
...
+    /**
+     * Records one slice's peak used bytes and its spill threshold ({@code sliceSize * maxLoadFactor}). Both are tracked
+     * as maxima, so after all slices close the pair describes the single fullest slice — the one that drives spilling.
+     * Used to compute {@code mergeBuffer/maxSpillProximity}.
+     */
+    public void sliceUsage(long usedBytes, long spillThresholdBytes)
     {
-      maxMergeBufferUsedBytes.addAndGet(bytes);
+      maxSliceUsedBytes.accumulateAndGet(usedBytes, Math::max);
+      sliceSpillThresholdBytes.accumulateAndGet(spillThresholdBytes, Math::max);
```
FrankChen021: [P2] Track spill proximity as a per-slice ratio

This stores the maximum used bytes and maximum threshold independently, which breaks when one query reports groupers with different thresholds. Nested/subtotal processing can pass the same PerQueryStats through a sliced ConcurrentGrouper and later full-buffer SpillingGroupers; if a small slice reaches its spill threshold, the larger full-buffer threshold can be retained here and `getSpillProximity()` will divide by that larger value, under-reporting the slice spill as roughly `1 / concurrencyHint`. Please track the maximum `usedBytes / spillThresholdBytes` per `sliceUsage` call, or otherwise keep the used/threshold pair together, so mixed grouper sizes still report the true max proximity.

aho135: Good catch — fixed in fc770ee. I replaced the two independent maxima (`maxSliceUsedBytes` and `sliceSpillThresholdBytes`) with a single `DoubleAccumulator` that maxes the per-slice ratio: `sliceUsage()` now computes `min(1.0, usedBytes / spillThresholdBytes)` at the point of measurement, so each slice's used bytes stay paired with its own threshold.

This closes exactly the case you described: when one `PerQueryStats` flows through both small `ConcurrentGrouper` slices and a full-buffer `SpillingGrouper` (subtotal/nested processing), a saturated small slice now reports the true `1.0` instead of being diluted to ~`1/concurrencyHint` by the larger threshold winning the separate max. Added `testSpillProximityKeepsPerSliceRatioWhenThresholdsDiffer` covering the mixed-threshold scenario.

FrankChen021 (commented):
I reviewed the follow-up to the prior per-slice spill-proximity comment. The fc770ee update now computes and maxes the per-slice ratio at measurement time, with a mixed-threshold regression test covering the original issue; no further inline reply looks useful.

---

FrankChen021 (commented):
I reviewed the follow-up to the prior per-slice spill-proximity comment. The current fc770ee code now computes and maxes the per-slice ratio at measurement time, and the mixed-threshold regression test covers the original issue; no further inline reply looks useful.

---

FrankChen021 (commented):
I reviewed the latest update for correctness, edge cases, concurrency, lifecycle, and monitor integration risks; no new issues found.

---

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Validation: `git diff --check` passed; builds and tests not run.



---

server/src/main/java/org/apache/druid/server/metrics/GroupByStatsMonitor.java
```diff
@@ -68,6 +68,7 @@ public boolean doMonitor(ServiceEmitter emitter)
       emitter.emit(builder.setMetric("mergeBuffer/maxAcquisitionTimeNs", statsContainer.getMaxMergeBufferAcquisitionTimeNs()));
       emitter.emit(builder.setMetric("mergeBuffer/bytesUsed", statsContainer.getTotalMergeBufferUsedBytes()));
       emitter.emit(builder.setMetric("mergeBuffer/maxBytesUsed", statsContainer.getMaxMergeBufferUsedBytes()));
+      emitter.emit(builder.setMetric("mergeBuffer/maxSpillProximity", statsContainer.getMaxSpillProximity()));
```
FrankChen021, line 71: [P1] Add the new metric to the default Prometheus mapping

The PR emits the new `mergeBuffer/maxSpillProximity` metric, but the default Prometheus mapping does not register it. With the default `dimensionMapPath`, Prometheus only exports metrics listed in `extensions-contrib/prometheus-emitter/src/main/resources/defaultMetrics.json`, so this event is silently dropped for the default configuration. Add `mergeBuffer/maxSpillProximity` as a no-dimension gauge to the default mapping and cover the mapping in a test.

aho135: Good catch — done in 4e00753. Added `mergeBuffer/maxSpillProximity` to `prometheus-emitter/src/main/resources/defaultMetrics.json` as a no-dimension gauge (matching the sibling `mergeBuffer/*` entries), and added `MetricsTest.testMergeBufferMaxSpillProximityRegisteredAsGauge` asserting the default mapping resolves it to a `Gauge` with only the standard service/host labels.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
```diff
@@ -442,19 +433,159 @@ public int getGrowthCount()
...
+  /**
+   * Replays the arena geometry to the terminal growth level — via the same {@link #initialTableStart} /
+   * {@link #nextGrowthLevel} primitives {@link #reset()} and {@link #adjustTableWhenFull()} use — and returns
+   * {@link #maxSizeForBuckets} of its bucket count. Allocates no buffers.
+   *
+   * <p>Valid only for the standard grow-by-doubling layout. Fixed-layout variants (the alternating limit-pushdown
+   * table) must never reach this — guaranteed by {@link #recordsFillProximity()} == false — and must override it if
+   * they ever need a spill denominator.
+   */
+  protected int computeSpillRegrowthThreshold()
```
abhishekrb19: nit:  `ccomputeSpillRegrowthThreshold()` and `getSpillRegrowthThreshold()` methods can be `private` scoped

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
```diff
@@ -224,9 +263,26 @@ public void mergeBufferAcquisitionTime(long delay)
...
+    /**
+     * Records one slice's peak fill ratio (1.0 iff it spilled), kept as a max across the query's slices; see
+     * {@link #maxSpillProximity}. Clamped to [0, 1]; NaN is ignored so a never-initialized grouper contributes nothing.
+     */
+    public void sliceUsage(double proximity)
+    {
+      if (Double.isNaN(proximity)) {
+        return;
+      }
+      final double clamped = proximity < 0.0 ? 0.0 : (proximity > 1.0 ? 1.0 : proximity);
```
abhishekrb19: nit: 
```suggestion
      final double clamped = proximity < 0.0 ? 0.0 : (Math.min(proximity, 1.0));
```

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
```diff
@@ -224,9 +263,26 @@ public void mergeBufferAcquisitionTime(long delay)
...
     {
-      maxMergeBufferUsedBytes.addAndGet(bytes);
+      mergeBufferUsedBytes.addAndGet(bytes);
+    }
+
+    /**
+     * Records one slice's peak fill ratio (1.0 iff it spilled), kept as a max across the query's slices; see
+     * {@link #maxSpillProximity}. Clamped to [0, 1]; NaN is ignored so a never-initialized grouper contributes nothing.
+     */
+    public void sliceUsage(double proximity)
```
abhishekrb19: Naming: what do you think of a different method name like `spillProximityUsage` so it doesn't seem like bytes usage per slice?

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/SpillingGrouper.java
```diff
@@ -249,7 +249,13 @@ public void reset()
   public void close()
   {
     perQueryStats.dictionarySize(getDictionarySizeEstimate());
-    perQueryStats.maxMergeBufferUsedBytes(getMaxMergeBufferUsedBytes());
+    final long sliceUsedBytes = getMaxMergeBufferUsedBytes();
+    perQueryStats.addMergeBufferUsedBytes(sliceUsedBytes);
+    if (grouper.isInitialized()) {
```
abhishekrb19: I see `getMaxMergeBufferUsedBytes()` also calls `grouper.isInitialized()`, the checks can perhaps be folded into one conditional block.

abhishekrb19 (approved):

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/ByteBufferHashTable.java
aho135: Done in d004d4d — both scoped to `private`. Neither is overridden (the alternating table opts out of proximity via `recordsFillProximity()` rather than overriding these), and they're only called within `ByteBufferHashTable`. Also dropped the now-inapplicable "must override it" note from the Javadoc.

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
aho135: Done in d1d2455.

processing/src/main/java/org/apache/druid/query/groupby/GroupByStatsProvider.java
aho135: Renamed to `spillProximity(double)` in be68bc4. Went with `spillProximity` over `spillProximityUsage` so it pairs with the existing `getSpillProximity()` getter and matches the noun-style accumulators (`spilledBytes`, `dictionarySize`) — and drops the "usage" association entirely.

processing/src/main/java/org/apache/druid/query/groupby/epinephelinae/SpillingGrouper.java
aho135: Done in 3aea492. Folded both the merge-buffer-usage and spill-proximity reporting under a single `grouper.isInitialized()` block and removed the now single-use private `getMaxMergeBufferUsedBytes()` helper. Behavior-preserving — the uninitialized path previously just did an inert `addMergeBufferUsedBytes(0L)`.

PR #19629, opened 2026-06-25
minor: bump Log4j dep to 2.26 and make JsonTemplateLayout available to operators in their custom  log4j configs

capistrant:
### Description

Using json formatted logs is not the default format shipped with Apache Druid stock log4j files, but many operators choose to customize their logging format. It is likely pretty common for them to use JSON formatted logs. As it stands, operators who do not add `log4j-layout-template-json` to their own druid bundle would be stuck with [JsonLayout](https://logging.apache.org/log4j/2.x/manual/layouts.html#JSONLayout) which is deprecated in favor of [JsonTemplateLayout](https://logging.apache.org/log4j/2.x/manual/json-template-layout.html). In addition to being the supported way to use JSON formatted logs, this layout provides superior performance to the legacy Layout. Also, `JsonLayout` is completely removed in log4j 3.x so we will need the new dependency eventually either way.

One of the main reasons I found myself down a rabbit hole in log4j was in doing some chaos testing of my configs for a virtual storage enabled data server. I was purposely stress testing with different configs and one such test put artificial load on the s3 connection pool by way oversubscribing segment read requests compared to the s3 client thread pool size. This led to a cascade of s3 related exceptions which end up logged with extended traces by RetryUtils. Sadly, the data server ended up deadlocked because all of the carrier threads for the virtual storage load virtual threads had pinned virtual threads due to native calls in their stacks that I traced back to log4j exception logging. While I don't know that what I was doing was a very realistic scenario for day to day Druid operations, I do think that it raised a real potential problem for both perf and operations in a cluster using JSON formatted logs.

#### Release note

Bump log4j dependency to 2.26.0. Add `log4j-layout-template-json` as a dependency to allow operators to switch from deprecated `JsonFormat` to `JsonTemplateFormat` if they are using Json formatted logs. Any operator wishing to migrate should refer to [JsonTemplateLayout](https://logging.apache.org/log4j/2.x/manual/json-template-layout.html). The jar ships with default Layouts that are defined in log4j docs, and you can also ship custom layouts if the defaults do not meet your needs.

<hr>

 * `pom.xml`
 * `processing/pom.xml`

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/licenses.yaml
+++ b/licenses.yaml
@@ -2040,13 +2040,14 @@ name: Apache Log4j
 license_category: binary
 module: java-core
 license_name: Apache License version 2.0
-version: 2.25.4
+version: 2.26.0
 libraries:
   - org.apache.logging.log4j: log4j-1.2-api
   - org.apache.logging.log4j: log4j-api
   - org.apache.logging.log4j: log4j-core
   - org.apache.logging.log4j: log4j-jul
   - org.apache.logging.log4j: log4j-slf4j-impl
+  - org.apache.logging.log4j: log4j-layout-template-json
 notices:
   - log4j-1.2-api: |
       Apache Log4j 1.x Compatibility API
@@ -2068,6 +2069,9 @@ notices:
       Copyright 1999-2015 Apache Software Foundation
   - log4j-slf4j-impl: |
       Apache Log4j SLF4J Binding
+      Copyright 1999-2015 Apache Software Foundation
+  - log4j-layout-template-json: |
+      Apache Log4j Layout Template JSON
       Copyright 1999-2015 Apache Software Foundation
 
 ---
--- a/licenses.yaml
+++ b/licenses.yaml
@@ -2071,8 +2071,8 @@ notices:
       Apache Log4j SLF4J Binding
       Copyright 1999-2015 Apache Software Foundation
   - log4j-layout-template-json: |
-      Apache Log4j Layout Template JSON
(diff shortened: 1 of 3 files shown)


licenses.yaml
```diff
@@ -2069,6 +2070,9 @@ notices:
   - log4j-slf4j-impl: |
       Apache Log4j SLF4J Binding
       Copyright 1999-2015 Apache Software Foundation
+  - log4j-layout-template-json: |
+      Apache Log4j Layout Template JSON
```
FrankChen021: [P2] Correct the NOTICE text for the new Log4j layout jar

The newly added log4j-layout-template-json notice does not match the 2.26.0 jar's META-INF/NOTICE: the artifact identifies itself as `Apache Log4j JSON Template Layout` and uses `Copyright 1999-2026 The Apache Software Foundation`, while this entry says `Apache Log4j Layout Template JSON` and keeps the older 1999-2015 copyright. `distribution/bin/generate-binary-notice.py` copies these notices into NOTICE.BINARY, so the binary release metadata for the newly bundled jar would be inaccurate.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

cryptoe (approved):

PR #19630, opened 2026-06-26
fix: partial segment bootstrap

clintropolis:
### Description
Fixes partial segment bootstrapping to use the actual range reader when restoring entries instead of a stub that was initially added to ease testing (this should have been done as part of #19535 when stuff got wired up). Without this fix, bootstrapped entries that were not fully resident in the cache would fail to load on demand later since they were using a range reader that wasn't really functional.


FrankChen021 (commented):
I have reviewed the code for correctness, edge cases, concurrency, lifecycle, security/data loss/API compatibility, and missing-test risks; no issues found.

---

capistrant (approved):
thanks for the catch and fix!

capistrant:
from jacoco report - the exceptional case of a null range reader during bootstrap seems like important test coverage to make sure it behaves as desired to not blow up the server going forward.

capistrant (approved):

PR #19632, opened 2026-06-26
minor: improve partial load metrics

clintropolis:
### Description
This PR fixes and improves metric tracking for partial segment loading, focusing on the `StorageMonitor`. Prior to this PR only the metadata header was tracked (and only partially, sometimes incorrectly).

To assist, a `PartialSegmentDownloadListener` interface has been introduced and passed to the file mapper so that callers can be notified whenever operations like a download happen, which is used to wire it into the `StorageLocation` metrics tracking (since it is no longer `SegmentLocalCacheManager` driving these things when partial loads are involved, instead it is the caller, e.g. query engines that retrieved segments).

Additionally, some new metrics:

* `storage/virtual/read/count` - number of deep-storage range read operations.
* `storage/virtual/read/bytes` - bytes pulled over the wire from range read operations.
* `storage/virtual/read/time` - total time spent in range reads (ms)

Finally, fixes flakiness in `QueryVirtualStorageTest` caused by the change in #19494 to not emit metrics when values are zeros, which this test was using as a sort of latch to indicate when activity on a StorageLocation had cleared out from a prior test. Instead, this is now done through new `tryWaitForNextEvent` and `awaitMetricQuiescent` methods on `LatchableEmitter` to watch for the monitor to either stabilize or stop emitting.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/query/QueryVirtualStorageTest.java
@@ -24,7 +24,6 @@ import org.apache.druid.data.input.impl.LocalInputSource;
 import org.apache.druid.indexer.TaskState;
 import org.apache.druid.java.util.common.HumanReadableBytes;
 import org.apache.druid.java.util.common.StringUtils;
-import org.apache.druid.java.util.emitter.service.ServiceMetricEvent;
 import org.apache.druid.msq.indexing.report.MSQTaskReport;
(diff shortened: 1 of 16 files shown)


FrankChen021 (commented):
I have reviewed the code for correctness, edge cases, concurrency, lifecycle, security/data loss/API compatibility, and missing-test risks; no issues found.

---

capistrant:
the missing coverage is not a concern to me. new tests validate the listener metrics are reported just fine IMO. re-running with skip label. if you feel compelled to make jacoco happy at later commit that is ok too

capistrant (approved):
+1 on green CI after adding jacoco skip label

PR #19644, opened 2026-06-30
perf: allow follower coordinators to pre-fill segment cache for faster leadership assumption via delta sync

jtuglu1:
### Description

Not currently in a reviewable state; however, open to ideas on any caveats here.

Currently, after a leadership failover/election the newly elected coordinator must wait to pull all data into its segment cache (if enabled) while it is in this LEADER_FIRST_SYNC state. This is because Coordinator treats the segment metadata cache as the authoritative source for snapshots. For large segment tables, this can cause significant downtime.

My proposed solution is allow follower Coordinator replicas to also fill their cache and rely on a delta sync upon becoming the leader. The delta sync is a much cheaper operation and allows the Coordinator to come up faster with a healthy cache. While this faster recovery/higher availability is at the expense of higher DB load, this is particularly useful for situations like rolling deployments where you may only have > 1 coordinators running for a short period (then back to 1).

#### Release note

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

<hr>

This PR has:

- [ ] been self-reviewed.


kfaraz:
> My proposed solution is allow follower Coordinator replicas to also fill their cache and rely on a delta sync upon becoming the leader. The delta sync is a much cheaper operation and allows the Coordinator to come up faster with a healthy cache. While this faster recovery/higher availability is at the expense of higher DB load, this is particularly useful for situations like rolling deployments where you may only have > 1 coordinators running for a short period (then back to 1).

@jtuglu1 , this is already the case. Follower coordinator/overlords already perform delta syncs in the background.
Only the first sync (that happens upon startup) is a full sync for both Leader and Follower.
This can be verified from logs as well as cache metrics.
The only requirement is that upon becoming Leader, the service needs to finish another delta sync (or full sync only if no sync has happened yet) before it starts to use the cache.

Please let me know if you have noticed otherwise.

PR #19646, opened 2026-07-01
feat: rebalance cost-based autoscaler for best throughput

Fly-Style:
This PR retunes the cost-based autoscaler scoring for better throughput-oriented decisions and exposes `idealIdleRatio` as a configurable knob for the U-shaped idle-cost function. 
It also updates autoscaler logging to show the effective idle ratio and max observed processing rate used during optimal task-count calculation.

<details>

**Picture 1**

image

- Small supervisor case: `48` partitions, starting from `6` tasks.
- Shows that the new cost function scales up gradually as lag becomes more important, while higher observed idle ratios keep the recommendation conservative.
- Useful sanity check for avoiding aggressive jumps on smaller supervisors.

**Picture 2**

image

- Medium supervisor case: `125` partitions, plot title shows `startTaskCount=25`.
- Shows that low-idle/high-pressure cases recommend materially more tasks, while moderate or high idle ratios stay well below the partition count.
- Useful for checking that the scoring does not blindly chase one-task-per-partition when idle capacity is already available.

**Picture 3**

image

- Large supervisor case: `500` partitions, starting at the maximum task count.
- Shows the strongest separation between low-idle and high-idle scenarios: low idle can keep the recommendation near max capacity, while higher idle ratios strongly favor scale-down.
- Useful for validating that the idle-cost side still prevents waste at large partition counts.

</details>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
@@ -289,7 +289,7 @@ public class CostBasedAutoScalerIntegrationTest extends StreamIndexTestBase
                       .hasDimension(DruidMetrics.SUPERVISOR_ID, supervisor.getId())
                       .hasValueMatching(Matchers.greaterThan(1L))
     );
-    Assertions.assertEquals(4, getCurrentTaskCount(supervisor.getId()));
+    Assertions.assertTrue(getCurrentTaskCount(supervisor.getId()) > 1);
     waitUntilPublishedRecordsAreIngested(totalRecords);
 
     // Let the tasks work through the lag.
@@ -398,7 +398,9 @@ public class CostBasedAutoScalerIntegrationTest extends StreamIndexTestBase
     final String getSupervisorPath = StringUtils.format("/druid/indexer/v1/supervisor/%s", supervisorId);
     final KafkaSupervisorSpec supervisorSpec = cluster.callApi().serviceClient().onLeaderOverlord(
         mapper -> new RequestBuilder(HttpMethod.GET, getSupervisorPath),
-        new TypeReference<>() {}
+        new TypeReference<>()
+        {
+        }
     );
     Assertions.assertNotNull(supervisorSpec);
     return supervisorSpec.getSpec().getIOConfig().getTaskCount();
--- a/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
+++ b/embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
@@ -398,9 +398,7 @@ public class CostBasedAutoScalerIntegrationTest extends StreamIndexTestBase
     final String getSupervisorPath = StringUtils.format("/druid/indexer/v1/supervisor/%s", supervisorId);
     final KafkaSupervisorSpec supervisorSpec = cluster.callApi().serviceClient().onLeaderOverlord(
         mapper -> new RequestBuilder(HttpMethod.GET, getSupervisorPath),
-        new TypeReference<>()
-        {
-        }
+        new TypeReference<>() {}
     );
     Assertions.assertNotNull(supervisorSpec);
     return supervisorSpec.getSpec().getIOConfig().getTaskCount();
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
@@ -250,13 +250,14 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
 
     log.info(
         "Computing optimal taskCount for supervisor[%s] with metrics:"
-        + " currentTaskCount[%d], avgPartitionLag[%.1f], avgProcessingRate[%.1f],"
-        + " pollIdleRatio[%.1f], lagWeight[%.1f], idleWeight[%.1f].",
+        + " currentTaskCount[%d], avgPartitionLag[%.1f], avgProcessingRate[%.1f], maxProcessingRate[%.1f]"
+        + " idleRatio[%.1f], lagWeight[%.1f], idleWeight[%.1f].",
         supervisorId,
         currentTaskCount,
         metrics.getAvgPartitionLag(),
         metrics.getAvgProcessingRate(),
-        metrics.getPollIdleRatio(),
+        metrics.getMaxObservedRate(),
+        config.isUsePollIdleRatio() ? metrics.getPollIdleRatio() : metrics.estimateIdleRatioFromProcessingRate(),
         config.getLagWeight(),
         config.getIdleWeight()
     );
--- a/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
@@ -251,13 +251,14 @@ public class CostBasedAutoScaler implements SupervisorTaskAutoScaler
     log.info(
         "Computing optimal taskCount for supervisor[%s] with metrics:"
         + " currentTaskCount[%d], avgPartitionLag[%.1f], avgProcessingRate[%.1f], maxProcessingRate[%.1f]"
-        + " idleRatio[%.1f], lagWeight[%.1f], idleWeight[%.1f].",
+        + " idleRatio[%.1f], pollIdleRatio[%.1f], lagWeight[%.2f], idleWeight[%.2f].",
         supervisorId,
         currentTaskCount,
         metrics.getAvgPartitionLag(),
         metrics.getAvgProcessingRate(),
         metrics.getMaxObservedRate(),
-        config.isUsePollIdleRatio() ? metrics.getPollIdleRatio() : metrics.estimateIdleRatioFromProcessingRate(),
+        metrics.estimateIdleRatioFromProcessingRate(),
+        metrics.getPollIdleRatio(),
         config.getLagWeight(),
         config.getIdleWeight()
     );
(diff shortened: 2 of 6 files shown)


embedded-tests/src/test/java/org/apache/druid/testing/embedded/indexing/autoscaler/CostBasedAutoScalerIntegrationTest.java
```diff
@@ -398,7 +398,9 @@ private int getCurrentTaskCount(String supervisorId)
     final String getSupervisorPath = StringUtils.format("/druid/indexer/v1/supervisor/%s", supervisorId);
     final KafkaSupervisorSpec supervisorSpec = cluster.callApi().serviceClient().onLeaderOverlord(
         mapper -> new RequestBuilder(HttpMethod.GET, getSupervisorPath),
-        new TypeReference<>() {}
+        new TypeReference<>()
```
kfaraz: Nit: Please revert this

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -250,13 +250,14 @@ int computeOptimalTaskCount(CostMetrics metrics)
...
-        + " pollIdleRatio[%.1f], lagWeight[%.1f], idleWeight[%.1f].",
+        + " currentTaskCount[%d], avgPartitionLag[%.1f], avgProcessingRate[%.1f], maxProcessingRate[%.1f]"
+        + " idleRatio[%.1f], lagWeight[%.1f], idleWeight[%.1f].",
         supervisorId,
         currentTaskCount,
         metrics.getAvgPartitionLag(),
         metrics.getAvgProcessingRate(),
-        metrics.getPollIdleRatio(),
+        metrics.getMaxObservedRate(),
+        config.isUsePollIdleRatio() ? metrics.getPollIdleRatio() : metrics.estimateIdleRatioFromProcessingRate(),
```
kfaraz: Let's log both the items separately, as that might help an operator choose one over the other.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -53,27 +53,28 @@ public class WeightedCostFunction
...
    */
-  static final double UNDER_PROVISIONING_PENALTY = 2.0;
+  static final double UNDER_PROVISIONING_PENALTY = 1.0;
 
   /**
    * Penalty magnitude applied when idle ratio is 1 (fully wasted capacity).
    * Controls the steepness of the U-shape on the over-provisioning side.
    */
-  static final double OVER_PROVISIONING_PENALTY = 1.0;
+  static final double OVER_PROVISIONING_PENALTY = 2.5;
```
kfaraz: Why change this to 2.5 instead of 2.0?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -53,27 +53,28 @@ public class WeightedCostFunction
   /**
    * Minimum rate of processing for any task in records per second. This is used
    * as a placeholder if avg rate is not available to ensure that cost computations
-   * do not return infinitely large lag recovery times.
+   * do not return infinitely large lag recovery times, at the expense of underestimating the lag cost.
    */
-  static final double MIN_PROCESSING_RATE = 1_000;
+  static final double MIN_PROCESSING_RATE = 5_000;
```
kfaraz: Why this change?
Technically, 1000 was also an arbitrary number but 5000 definitely seems to be on the higher side, especially when dealing with bulkier records with large (say JSON) column values. I would rather the user choose whether they want to prefer lag recovery or throughput by tweaking the weights.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -86,6 +87,7 @@ public CostBasedAutoScalerConfig(
       @Nullable @JsonProperty("scaleActionPeriodMillis") Long scaleActionPeriodMillis,
       @Nullable @JsonProperty("lagWeight") Double lagWeight,
       @Nullable @JsonProperty("idleWeight") Double idleWeight,
+      @Nullable @JsonProperty("idealIdleRatio") Double idealIdleRatio,
```
kfaraz: Maybe rename to `optimalIdleRatio` or `optimalTaskIdleRatio`.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -53,27 +53,28 @@ public class WeightedCostFunction
...
 
   /**
-   * Target idle ratio representing the optimal operating point for the U-shaped idle cost.
+   * Default target idle ratio representing the optimal operating point for the U-shaped idle cost.
    * At this ratio the idle cost is at its minimum; both lower (risk) and higher (waste) are penalized.
+   * Configurable via {@link CostBasedAutoScalerConfig#getIdealIdleRatio()}.
    */
   static final double IDEAL_IDLE_RATIO = 0.25;
```
kfaraz:
```suggestion
  static final double DEFAULT_TARGET_IDLE_RATIO = 0.25;
```

kfaraz (commented):
In conclusion, does the auto-scaler now prefer under-provisioning by default?

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: I decided to tweak it too and realized that `1000` is too permissible to scaleup during minimal lag. `5000` implies more strict behaviour in that critical state where we have not received metrics yet.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: I tried 2.0, but 2.5 looked better in simulation.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
FrankChen021: [P2] Measured low throughput is treated as 5k r/s

This new 5,000 r/s floor is applied with Math.max(avgProcessingRate, MIN_PROCESSING_RATE), so any real measured rate between the old 1,000 r/s floor and 5,000 r/s is now replaced with 5,000 when calculating lagRecoveryTime. Slow-but-running supervisors understate backlog recovery time by up to 5x, lowering lag cost and potentially preventing needed scale-up. The fallback should only mask missing/zero rates, or the floor should not override valid low measured rates.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Found 1 issue.

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
kfaraz: If we have not received metrics yet, auto-scaling would be skipped since `CostBasedAutoScaler.validateMetricsForScaling` would return an error.

I have personally seen tasks in prod clusters maxing out at 5000 records/sec when dealing with large records.
So, using a large `MIN_PROCESSING_RATE` would cause us to always under-estimate `lagRecoveryTime`, irrespective of the actual `avgProcessingRate`.
As such, let's keep a low value of `MIN_PROCESSING_RATE`(maybe even as low as 100), since it is meant to be a safe-side measure that kicks in only when `avgProcessingRate` is very low.

The penalty for scale-up is already driven by the optimal task idleness, and can be controlled using the weights.

kfaraz: > The fallback should only mask missing/zero rates, or the floor should not override valid low measured rates.

Yes, the fallback is meant to mask zero or unrealistically low rates.
Missing metrics would already lead to a no-op scaling decision.

Fly-Style: P.S. reverting

kfaraz: Thanks!

kfaraz: In a future PR, I think we can remove the `MIN_PROCESSING_RATE` altogether and maybe use the window `maxProcessingRate`, but I haven't fully thought it through yet. It might have some unforeseen side effects.

Fly-Style: Fixed

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
kfaraz: I feel 2x penalty on over-provisioning is already enough. Changing these constants very frequently will make it difficult to gather enough data to refine the constants empirically.

kfaraz: Let's use 2 for now and see how it does in real clusters. This PR is already interchanging the under provisioned penalty and the over provisioning penalty, so we will have enough stuff to validate anyway.

kfaraz (approved):
Left a non-blocking comment.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: Okay, let's use 2, but I can assure you 2.5 looks even better on simulation. We'll see 😺

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
Fly-Style: @kfaraz I forgot about Math.max(...) and then I reconsider my approach 😁

PR #19648, opened 2026-07-01
minor: Add segment debug logging to CachingClusteredClient

cecemei:
### Description
When debugging slow or incorrect queries on the broker, it's useful to know how many segments were found during timeline lookup and how they break down by interval, version, and server type (realtime vs historical).

<hr>

This PR has:

- [x] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
+++ b/server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
@@ -346,18 +346,17 @@ public class CachingClusteredClient implements QuerySegmentWalker
 
       final Set<SegmentServerSelector> segmentServers = computeSegmentsToQuery(timeline, specificSegments);
       log.info("Query[%s] found [%d] segments from timeline lookup.", query.getId(), segmentServers.size());
-      if (query.context().isDebug()) {
-        segmentServers.stream()
-                      .collect(Collectors.groupingBy(
-                          s -> s.getSegmentDescriptor().getInterval()
-                               + "_" + s.getSegmentDescriptor().getVersion()
-                               + "_" + (s.getServer() != null && s.getServer().isRealtimeSegment() ? "realtime" : "historical"),
-                          Collectors.counting()
-                      ))
-                      .forEach((key, count) ->
-                          log.info("Query[%s] interval/version/type[%s] partitions[%d]", query.getId(), key, count)
-                      );
-      }
+      segmentServers.stream()
+                    .collect(Collectors.groupingBy(
+                        s -> s.getSegmentDescriptor().getInterval()
+                             + "_" + s.getSegmentDescriptor().getVersion()
+                             + "_" + (s.getServer() != null && s.getServer().isRealtimeSegment() ? "realtime" : "historical"),
+                        Collectors.counting()
+                    ))
+                    .forEach((key, count) ->
+                        log.info("Query[%s] interval/version/type[%s] partitions[%d]", query.getId(), key, count)
+                    );
+
       final CloneQueryMode cloneQueryMode = query.context().getCloneQueryMode();
       @Nullable
       final byte[] queryCacheKey = cacheKeyManager.computeSegmentLevelQueryCacheKey();
--- a/server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
+++ b/server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
@@ -345,19 +345,18 @@ public class CachingClusteredClient implements QuerySegmentWalker
       }
 
       final Set<SegmentServerSelector> segmentServers = computeSegmentsToQuery(timeline, specificSegments);
-      log.info("Query[%s] found [%d] segments from timeline lookup.", query.getId(), segmentServers.size());
       if (query.context().isDebug()) {
-        final Map<String, Long> segmentCountByKey = segmentServers.stream().collect(
-            Collectors.groupingBy(
-                s -> s.getSegmentDescriptor().getInterval()
-                     + "_" + s.getSegmentDescriptor().getVersion()
-                     + "_" + (s.getServer() != null && s.getServer().isRealtimeSegment() ? "realtime" : "historical"),
-                Collectors.counting()
-            )
-        );
-        segmentCountByKey.forEach(
-            (key, count) ->
-                log.info("Query[%s] interval/version/type[%s] partitions[%d]", query.getId(), key, count)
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
```diff
@@ -345,6 +345,19 @@ ClusterQueryResult<T> run(
       }
 
       final Set<SegmentServerSelector> segmentServers = computeSegmentsToQuery(timeline, specificSegments);
+      log.info("Query[%s] found [%d] segments from timeline lookup.", query.getId(), segmentServers.size());
```
clintropolis: i think this info bit would be noisy and should be debug too, also maybe consider using `log.debugSegments` for the debug part since it chunks large lists of segments up into smaller log messages

kfaraz: +1

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
```diff
@@ -345,6 +345,19 @@ ClusterQueryResult<T> run(
...
+      if (query.context().isDebug()) {
+        segmentServers.stream()
+                      .collect(Collectors.groupingBy(
+                          s -> s.getSegmentDescriptor().getInterval()
+                               + "_" + s.getSegmentDescriptor().getVersion()
+                               + "_" + (s.getServer() != null && s.getServer().isRealtimeSegment() ? "realtime" : "historical"),
+                          Collectors.counting()
```
kfaraz: Instead of this, I think we might as well create a `SegmentId` and invoke `.toString()` on it.

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
```diff
@@ -345,6 +345,19 @@ ClusterQueryResult<T> run(
       }
 
       final Set<SegmentServerSelector> segmentServers = computeSegmentsToQuery(timeline, specificSegments);
+      log.info("Query[%s] found [%d] segments from timeline lookup.", query.getId(), segmentServers.size());
+      if (query.context().isDebug()) {
```
kfaraz, line 348: Since this will happen only when `debug` is set to true in the context, I assume we would do it only in a test environment or while trying to debug a production cluster.

In that case, could a query plan provide this info with the actual segment IDs that would be touched by the query?

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
cecemei: we're only logging the total size of segments not the whole set, we also don't have datasegment here, this is really cropped segment descriptor.

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
cecemei: this doesn't correspond to a segment, the interval is cropped

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
cecemei: yes it's only logging in debug setting to avoid verbose, i dont think a query plan has this info? or do you mean defer the logging to sometime later? i think here's better because it's before the caching/pruning stage so we get a whole picture of what segments are included from the timeline

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
kfaraz: > we're only logging the total size of segments

Even logging the total segment count per query will become very noisy at high QPS. It is also a little confusing since we don't log any other thing from the lifecycle of a query when `debug` is off.

kfaraz: > we also don't have datasegment here, this is really cropped segment descriptor.

The `SegmentDescriptor` is the same as the `SegmentId` minus the datasource.
We already have the datasource name here (`ev.getBaseTableDataSource()`) which is used to extract the timeline in preceding lines.

We should convert all descriptors to `SegmentId` and add a new method `Logger.debugSegmentIds` similar to `Logger.infoSegmentIds`, if needed or we can just use `infoSegmentIds` itself if we are gating this on `query.context().isDebug()`

FrankChen021 (commented):

---

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
cecemei: updated

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Found 1 issue: debug logging can emit synthetic segment IDs for sliced timeline lookups.

---

server/src/main/java/org/apache/druid/client/CachingClusteredClient.java
```diff
@@ -345,6 +345,20 @@ ClusterQueryResult<T> run(
       }
 
       final Set<SegmentServerSelector> segmentServers = computeSegmentsToQuery(timeline, specificSegments);
+      if (query.context().isDebug()) {
+        final String dataSource = ev.getBaseTableDataSource().getName();
+        log.infoSegmentIds(
+            segmentServers.stream().map(
+                s -> SegmentId.of(
+                    dataSource,
+                    s.getSegmentDescriptor().getInterval(),
```
FrankChen021: [P2] Log real segment IDs, not query slices

The debug log rebuilds a `SegmentId` from `SegmentDescriptor.getInterval()`, but `computeSegmentsToQuery` fills that descriptor from `TimelineObjectHolder.getInterval()`. `VersionedIntervalTimeline.lookup` trims the first or last holder to the requested interval, so a query over one day inside a week-long segment logs a synthetic segment id that never existed. Use `s.getServer().getSegment().getId()` for the segment id, and log the query interval separately if needed.

FrankChen021 (commented):
I reviewed the new commits since the previous review and the full current diff for correctness, edge cases, concurrency, and integration risks; no new issues found in the latest changes.

---

FrankChen021 (commented):
I reviewed the latest update and surrounding selector/logging paths. The current code logs the real selected segment IDs, and I found no high-confidence PR-caused issues.

---

PR #19650, opened 2026-07-02
fix: Perform cleanup of intermediary data safely

kfaraz:
### Description

#19187 introduced cleanup of intermediary data in `ParallelIndexSupervisorTask.cleanup()` step.
It invokes the `DataSegmentKiller` (bound on peons to `OmniDataSegmentKiller`) and cleans up all files under the directory specified by the supervisor task id. The `OmniDataSegmentKiller` tries to do a recursive kill on all possible implementations of `DataSegmentKiller`.

#### Bug
<details>

<summary>
This can lead to task failures in cases where an extension (say Azure or GCS) has been loaded on the cluster but is not configured to work as a deep store.
</summary>

```java
com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) druid.azure.container - must not be null
  at JsonConfigProvider.bind(JsonConfigProvider.java:151)
      \_ installed by: Modules$OverrideModule -> AzureStorageDruidModule
  at AzureDataSegmentKiller.<init>(AzureDataSegmentKiller.java:61)
      \_ for 1st parameter
  while locating AzureDataSegmentKiller
  at AzureStorageDruidModule.configure(AzureStorageDruidModule.java:103)
      \_ installed by: Modules$OverrideModule -> AzureStorageDruidModule
  while locating DataSegmentKiller annotated with @Element(setName=,uniqueId=447, type=MAPBINDER, keyType=String)

1 error

======================
Full classname legend:
======================
AzureDataSegmentKiller:  "org.apache.druid.storage.azure.AzureDataSegmentKiller"
AzureStorageDruidModule: "org.apache.druid.storage.azure.AzureStorageDruidModule"
DataSegmentKiller:       "org.apache.druid.segment.loading.DataSegmentKiller"
Element:                 "com.google.inject.internal.Element"
JsonConfigProvider:      "org.apache.druid.guice.JsonConfigProvider"
Modules$OverrideModule:  "com.google.inject.util.Modules$OverrideModule"
========================
End of classname legend:
========================

	at com.google.inject.internal.InternalProvisionException.toProvisionException(InternalProvisionException.java:251)
	at com.google.inject.internal.InjectorImpl$1.get(InjectorImpl.java:1151)
	at com.google.common.base.Suppliers$NonSerializableMemoizingSupplier.get(Suppliers.java:181)
	at org.apache.druid.segment.loading.OmniDataSegmentKiller.killRecursively(OmniDataSegmentKiller.java:109)
	at org.apache.druid.indexing.common.task.batch.parallel.ParallelIndexSupervisorTask.cleanUp(ParallelIndexSupervisorTask.java:1840)
```

</details>

#19647 was meant to mitigate this issue by ignoring such exceptions and allowing the task to succeed.
The exception would still show up as a warning in the task logs.

There is also an inherent risk with this approach since it is possible (however unlikely) that the same path prefix may be under use in other deep storage environments for completely different purposes.
This is especially a concern in cases where an extension (say Azure) is loaded only to be used as an input source.
For the same reason, `OmniDataSegmentKiller.killAll()` has not been implemented yet either.

### Changes

- Fix `OmniDataSegmentKiller.killRecursively()` to only invoke the currently bound deep store
- Use `IntermediaryDataManager` in `ParallelIndexSupervisorTask` instead of invoking `DataSegmentKiller` directly.
This allows use of a cleaner interface at the task level and also provides symmetry with `DataSegmentPusher` within `DeepStorageIntermediaryDataManager`.
- Implement `deletePartitions` and invoke `DataSegmentKiller.killRecursively()`
- Deprecate `DataSegmentKiller.killAll()` since it is not used anymore and is a potentially risky operation
- Add javadocs and comments
- Fix tests

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/indexing-service/src/main/java/org/apache/druid/indexing/worker/shuffle/DeepStorageIntermediaryDataManager.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/worker/shuffle/DeepStorageIntermediaryDataManager.java
@@ -24,6 +24,7 @@ import com.google.inject.Inject;
 import org.apache.druid.indexing.common.TaskToolbox;
 import org.apache.druid.indexing.common.task.batch.parallel.DeepStoragePartitionStat;
 import org.apache.druid.java.util.common.IAE;
+import org.apache.druid.segment.loading.DataSegmentKiller;
 import org.apache.druid.segment.loading.DataSegmentPusher;
 import org.apache.druid.timeline.DataSegment;
 import org.apache.druid.timeline.partition.BucketNumberedShardSpec;
@@ -38,6 +39,7 @@ public class DeepStorageIntermediaryDataManager implements IntermediaryDataManag
 {
   public static final String SHUFFLE_DATA_DIR_PREFIX = "shuffle-data";
   private final DataSegmentPusher dataSegmentPusher;
+  private final DataSegmentKiller dataSegmentKiller;
 
   /**
    * Deep storage path to the directory that holds all shuffle intermediate files for {@code supervisorTaskId},
@@ -50,9 +52,10 @@ public class DeepStorageIntermediaryDataManager implements IntermediaryDataManag
   }
 
   @Inject
-  public DeepStorageIntermediaryDataManager(DataSegmentPusher dataSegmentPusher)
+  public DeepStorageIntermediaryDataManager(DataSegmentPusher dataSegmentPusher, DataSegmentKiller dataSegmentKiller)
   {
     this.dataSegmentPusher = dataSegmentPusher;
+    this.dataSegmentKiller = dataSegmentKiller;
   }
 
   @Override
@@ -109,18 +112,16 @@ public class DeepStorageIntermediaryDataManager implements IntermediaryDataManag
    * Not implemented for deep storage mode. Unlike {@link LocalIntermediaryDataManager},
    * which can walk the local filesystem to find and delete files by supervisorTaskId,
    * this manager has no way to discover what files were pushed: it has no
-   * {@link org.apache.druid.segment.loading.DataSegmentKiller}, does not track pushed
+   * {@link DataSegmentKiller}, does not track pushed
    * paths, and runs on short-lived peon processes whose state is lost on exit.
    * <p>
    * Deep storage shuffle cleanup is handled in {@link org.apache.druid.indexing.common.task.batch.parallel.ParallelIndexSupervisorTask#cleanUp}
-   * via {@link org.apache.druid.segment.loading.DataSegmentKiller#killRecursively} on
+   * via {@link DataSegmentKiller#killRecursively} on
    * {@link #retrieveShuffleDataStoragePath(String)} (recursive delete of that directory).
    */
   @Override
-  public void deletePartitions(String supervisorTaskId)
+  public void deletePartitions(String supervisorTaskId) throws IOException
   {
-    throw new UnsupportedOperationException(
-        "Deep storage shuffle cleanup is handled by ParallelIndexSupervisorTask, not by the data manager"
-    );
+    dataSegmentKiller.killRecursively(retrieveShuffleDataStoragePath(supervisorTaskId));
   }
 }
--- a/indexing-service/src/main/java/org/apache/druid/indexing/worker/shuffle/DeepStorageIntermediaryDataManager.java
+++ b/indexing-service/src/main/java/org/apache/druid/indexing/worker/shuffle/DeepStorageIntermediaryDataManager.java
@@ -24,6 +24,7 @@ import com.google.inject.Inject;
 import org.apache.druid.indexing.common.TaskToolbox;
 import org.apache.druid.indexing.common.task.batch.parallel.DeepStoragePartitionStat;
 import org.apache.druid.java.util.common.IAE;
+import org.apache.druid.java.util.common.ISE;
 import org.apache.druid.segment.loading.DataSegmentKiller;
 import org.apache.druid.segment.loading.DataSegmentPusher;
 import org.apache.druid.timeline.DataSegment;
@@ -51,8 +52,19 @@ public class DeepStorageIntermediaryDataManager implements IntermediaryDataManag
     return SHUFFLE_DATA_DIR_PREFIX + "/" + supervisorTaskId;
   }
 
+  /**
+   * Used by Guice to create an instance of {@link DeepStorageIntermediaryDataManager}.
+   *
+   * @param dataSegmentPusher Always non-null
+   * @param dataSegmentKiller Can be null in certain cases such as on MiddleManagers
+   *                          when using druid.storage.type=s3 since the respective
+   *                          S3DataSegmentKiller uses scheme "s3_zip" instead of "s3"
+   */
   @Inject
-  public DeepStorageIntermediaryDataManager(DataSegmentPusher dataSegmentPusher, DataSegmentKiller dataSegmentKiller)
+  public DeepStorageIntermediaryDataManager(
+      DataSegmentPusher dataSegmentPusher,
+      @Nullable DataSegmentKiller dataSegmentKiller
+  )
   {
     this.dataSegmentPusher = dataSegmentPusher;
     this.dataSegmentKiller = dataSegmentKiller;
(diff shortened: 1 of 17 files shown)


indexing-service/src/main/java/org/apache/druid/indexing/worker/shuffle/DeepStorageIntermediaryDataManager.java
```diff
@@ -50,9 +52,10 @@ public static String retrieveShuffleDataStoragePath(String supervisorTaskId)
   }
 
   @Inject
-  public DeepStorageIntermediaryDataManager(DataSegmentPusher dataSegmentPusher)
+  public DeepStorageIntermediaryDataManager(DataSegmentPusher dataSegmentPusher, DataSegmentKiller dataSegmentKiller)
```
FrankChen021: [P1] S3 deepstore mode can no longer provision this manager

This constructor now requires a DataSegmentKiller anywhere DeepStorageIntermediaryDataManager is selected. On middleManager-style injectors that killer is PolyBind-selected by druid.storage.type, but S3 sets that to s3 while S3StorageDruidModule registers its killer under s3_zip. With druid.processing.intermediaryData.storage.type=deepstore on S3, Guice fails to provision this manager instead of starting the service.

server/src/main/java/org/apache/druid/segment/loading/OmniDataSegmentKiller.java
```diff
@@ -97,27 +103,26 @@ private DataSegmentKiller getKiller(DataSegment segment) throws SegmentLoadingEx
...
-          firstFailure.addSuppressed(e);
-        }
-      }
-    }
-    if (firstFailure != null) {
-      throw firstFailure;
+    // Delegate kill to the currently bound deep storage implementation only.
+    // It is unnecessary and also potentially unsafe to kill all files under the
+    // same folder name on inactive deep storage bindings provided by other extensions.
+    final Supplier<DataSegmentKiller> killer = killers.get(deepStorageType);
```
FrankChen021, line 117: [P2] Recursive cleanup looks up killers with the wrong key

The killers map is keyed by segment loadSpec killer type, not always by druid.storage.type. S3 and OSS use storage types s3 and oss but killer keys s3_zip and oss_zip, so this lookup throws during recursive cleanup for those configured stores instead of preserving the prior no-op behavior for unsupported recursive cleanup.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

Found 2 issues.

---

indexing-service/src/main/java/org/apache/druid/indexing/worker/shuffle/DeepStorageIntermediaryDataManager.java
kfaraz: Hmm, this seems to be true. I was under the assumption that `DataSegmentKiller` is always bound to `OmniDataSegmentKiller` at the top level. But that is applicable only for peons (via `CliPeon`) and Overlord. MMs do not bind to `OmniDataSegmentKiller` and use the extension provided `DataSegmentKiller` itself.
So, when using S3, MMs would not have any initialized binding for `DataSegmentKiller`.

However, MMs do use `IntermediaryDataManager`.

I will revert this change and add javadocs to avoid further confusion.

kfaraz: Trying out an optional dependency first.

server/src/main/java/org/apache/druid/segment/loading/OmniDataSegmentKiller.java
kfaraz: Ugh!

So this means that we are always forced to use the `OmniDataSegmentKiller` since `S3DataSegmentKiller` is never truly bound as a `DataSegmentKiller`.

The `S3DataSegmentKiller` is forced to use type `s3_zip` since the `S3LoadSpec` uses `s3_zip`. For backwards compat reading old segments, neither of these can change.

Workaround option would be to register `S3DataSegmentKiller` under the `s3` key as well, which should be okay since the `S3DataSegmentKiller` does not keep any state, so it is fine to end up with 2 `S3DataSegmentKiller`, one bound to key `s3`, the other to `s3_zip`.

kfaraz: I think we can live with this for now since the `killRecursively()` method is not implemented for S3 or OSS anyway. Going to add some comments to that effect.

indexing-service/src/main/java/org/apache/druid/indexing/worker/shuffle/DeepStorageIntermediaryDataManager.java
FrankChen021: Thanks for checking. I don't think `@Nullable` makes this dependency optional in this path: `DataSegmentKiller` is still bound through `LocalDataStorageDruidModule`'s `PolyBind.createChoice`, and `PolyBind.ConfiggedProvider#get()` throws a `ProvisionException` when `druid.storage.type=s3` cannot be found in the killer map, which still only has `s3_zip`. That happens while Guice resolves the constructor argument, before this constructor can receive `null`, so the MiddleManager deepstore provisioning failure should still reproduce.

indexing-service/src/main/java/org/apache/druid/indexing/worker/shuffle/DeepStorageIntermediaryDataManager.java
```diff
@@ -49,10 +52,22 @@ public static String retrieveShuffleDataStoragePath(String supervisorTaskId)
...
+   * @param dataSegmentPusher Always non-null
+   * @param dataSegmentKiller Can be null in certain cases such as on MiddleManagers
+   *                          when using druid.storage.type=s3 since the respective
+   *                          S3DataSegmentKiller uses scheme "s3_zip" instead of "s3"
+   */
   @Inject
-  public DeepStorageIntermediaryDataManager(DataSegmentPusher dataSegmentPusher)
+  public DeepStorageIntermediaryDataManager(
+      DataSegmentPusher dataSegmentPusher,
+      @Nullable DataSegmentKiller dataSegmentKiller
```
FrankChen021, line 66: [P1] @Nullable does not make the killer binding optional

`@Nullable` only permits an injected provider to return null; it does not turn a failing PolyBind choice into an optional dependency. On MiddleManager or Overlord with deep-store shuffle and `druid.storage.type=s3` or `oss`, the `DataSegmentKiller` choice requests `s3` or `oss`, while the registered keys are `s3_zip` or `oss_zip`, so `ConfiggedProvider` throws `ProvisionException` while constructing this manager, before the null guard in `deletePartitions`. This prevents the deep-store manager from being provisioned; use an actually optional binding/provider or add storage-type aliases.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Reviewed 17 of 17 changed files. The original recursive-cleanup concern remains valid, and the attempted nullable injection also leaves deep-store manager provisioning broken for S3 and OSS.

---

server/src/main/java/org/apache/druid/segment/loading/OmniDataSegmentKiller.java
FrankChen021: Thanks — the added comments explain the mismatch, but they do not avoid it. For `druid.storage.type=s3` or `oss`, this still looks up `killers.get("s3")` or `killers.get("oss")` while the map keys are `s3_zip` and `oss_zip`, so it throws before reaching those killers' default no-op `killRecursively()`. The supervisor catches the exception, but every cleanup will log a warning. Could this explicitly no-op unsupported storage types, or map/alias the configured storage type to the killer key?

PR #19651, opened 2026-07-02
minor: Add logs and metrics observability to the segment upgrade path that fires while using concurrent locking

capistrant:
### Description

Investigating potential issues with concurrent append and replace actions not being properly accepted and reflected on realtime ingestion tasks who should be upgrading their segments to reflect the concurrent append that happened on an interval they are ingesting to. This PR attempts to add some logs and metrics to the related code to help give observability into the process and smoke out if there are indeed issues going on. Biggest risk of these new mets and logs is probably chattiness, since they scale with the volume of upgrade segments that may be existing under a clusters normal operaions.

#### New Metrics

|-|-|-|-|
|`ingest/segmentUpgrade/count`|Number of pending segments that a concurrent replace (for example, compaction) upgraded to a new version and asked the supervisor to have running tasks announce under the new version. Emitted by the replace task only when streaming ingestion is running concurrently with replace on the same interval.|`dataSource`, `taskId`, `taskType`, `groupId`, `tags`|0 unless [concurrent append and replace](../ingestion/concurrent-append-replace.md) is in use.|
|`ingest/segmentUpgrade/notified`|Number of upgraded pending segments the supervisor successfully routed to at least one running task. Compare with `ingest/segmentUpgrade/count`: `count` should equal `notified` + `unmatched` over the same period and `dataSource`.|`supervisorId`, `dataSource`, `stream`, `tags`|Equal to `count` in a healthy cluster.|
|`ingest/segmentUpgrade/unmatched`|Number of upgraded pending segments the supervisor could not route to any running task. These are not announced under the new version until handoff, so the corresponding data may be briefly missing from queries.|`supervisorId`, `dataSource`, `stream`, `tags`|0. A non-zero value indicates a lost upgrade announcement.|
|`ingest/segmentUpgrade/sendFailed`|Number of upgrade requests that matched a running task but failed to reach it over the wire after retries.|`supervisorId`, `dataSource`, `stream`, `taskId`, `tags`|0|
|`ingest/segmentUpgrade/announced`|Number of upgraded segments a task announced under the new version. Emitted once per task, so it scales with the replica count. Do not subtract it directly from `ingest/segmentUpgrade/count`, which is per-segment rather than per-replica.|`dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Greater than 0 while concurrent replace occurs.|
|`ingest/segmentUpgrade/skipped`|Number of upgrade requests a task received but did not announce. The `reason` dimension is one of `unknownBase` (the request reached the wrong task), `noSink` (the base sink is no longer present), or `dropping` (the base sink is handing off, which is benign and covered by the durable publish path).|`dataSource`, `taskId`, `taskType`, `groupId`, `tags`, `reason`|0, excluding `reason=dropping`.|

#### Release note

TBD

<hr>

 TBD

<hr>

This PR has:

- [ ] been self-reviewed.


FrankChen021 (commented):
I have reviewed the code for correctness, edge cases, concurrency/lifecycle, security, data loss, API compatibility, and missing-test risks; no high-confidence issues found.

---

indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentUpgradeMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+ *   <li>the supervisor emits {@link #NOTIFIED}, {@link #UNMATCHED} and {@link #SEND_FAILED} as it fans requests out,</li>
+ *   <li>the streaming task emits {@link #ANNOUNCED} and {@link #SKIPPED} (with a {@code reason}) as it applies them.</li>
+ * </ul>
+ * Comparing {@link #COUNT} (created) against {@link #ANNOUNCED} (applied) per dataSource quantifies the visibility gap;
+ * {@link #UNMATCHED}, {@link #SEND_FAILED} and {@link #SKIPPED} attribute where a lost upgrade dropped out.
+ */
+public class SegmentUpgradeMetrics
+{
+  /** Number of upgraded pending segments a REPLACE commit created and handed to the supervisor. Task-action dims. */
+  public static final String COUNT = "ingest/segmentUpgrade/count";
```
kfaraz: Please include the prefix `realtime` in all the metrics to distinguish upgrade of pending segments from upgrade of appended segments.

```suggestion
  public static final String REALTIME_PERSISTED = "ingest/realtime/segmentUpgrade/persisted";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentUpgradeMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+  /** A task announced an upgraded segment under the new version. Task dims. */
+  public static final String ANNOUNCED = "ingest/segmentUpgrade/announced";
+
+  /** A task received an upgrade request but did not announce it; see the {@code reason} dimension. Task dims. */
+  public static final String SKIPPED = "ingest/segmentUpgrade/skipped";
+
+  // Values for the DruidMetrics.REASON dimension on SKIPPED.
+
+  /** The task holds no pending segment matching upgradedFromSegmentId (request targeted the wrong task). */
+  public static final String REASON_UNKNOWN_BASE = "unknownBase";
```
kfaraz: Nit: The reasons need not be camel cased, e.g. `unknown base sink` instead of `unknownBase`.

indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentUpgradeMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+  /** The base sink is gone even though this task once held it. */
+  public static final String REASON_NO_SINK = "noSink";
+  /** The base sink is being dropped (handoff in progress); the durable path re-announces at the new version. */
+  public static final String REASON_DROPPING = "dropping";
```
kfaraz:
```suggestion
  public static final String REASON_DROPPING = "dropping base sink";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentUpgradeMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /** A task received an upgrade request but did not announce it; see the {@code reason} dimension. Task dims. */
+  public static final String SKIPPED = "ingest/segmentUpgrade/skipped";
+
+  // Values for the DruidMetrics.REASON dimension on SKIPPED.
+
+  /** The task holds no pending segment matching upgradedFromSegmentId (request targeted the wrong task). */
+  public static final String REASON_UNKNOWN_BASE = "unknownBase";
+  /** The base sink is gone even though this task once held it. */
+  public static final String REASON_NO_SINK = "noSink";
```
kfaraz:
```suggestion
  public static final String REASON_NO_SINK = "base sink already dropped";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentUpgradeMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+ * Comparing {@link #COUNT} (created) against {@link #ANNOUNCED} (applied) per dataSource quantifies the visibility gap;
+ * {@link #UNMATCHED}, {@link #SEND_FAILED} and {@link #SKIPPED} attribute where a lost upgrade dropped out.
+ */
+public class SegmentUpgradeMetrics
+{
+  /** Number of upgraded pending segments a REPLACE commit created and handed to the supervisor. Task-action dims. */
+  public static final String COUNT = "ingest/segmentUpgrade/count";
+
+  /** A record was delivered to at least one running task. Supervisor dims. */
+  public static final String NOTIFIED = "ingest/segmentUpgrade/notified";
```
kfaraz:
```suggestion
  public static final String REALTIME_TASK_NOTIFIED = "ingest/realtime/segmentUpgrade/notified";
```

indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentUpgradeMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /** A record matched no running task and will not be re-announced until handoff. Supervisor dims. */
+  public static final String UNMATCHED = "ingest/segmentUpgrade/unmatched";
```
kfaraz: I wonder if we shouldn't just club this metric with `ingest/realtime/segmentUpgrade/skipped` and add a reason "no matching task"

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1419,22 +1421,89 @@ public void registerNewVersionOfPendingSegment(
...
+          + " supervisor[%s]; it will not be re-announced until handoff. Currently tracking [%d] activelyReading"
+          + " and [%d] pendingCompletion task group(s).",
+          pendingSegmentRecord.getId(),
+          pendingSegmentRecord.getUpgradedFromSegmentId(),
+          taskAllocatorId,
+          supervisorId,
+          activelyReadingTaskGroups.size(),
+          pendingCompletionTaskGroups.size()
+      );
+      emitter.emit(getMetricBuilder().setMetric(SegmentUpgradeMetrics.UNMATCHED, 1));
```
kfaraz: Let's include the segment version and interval in all the metrics.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1419,22 +1421,89 @@ public void registerNewVersionOfPendingSegment(
...
+    } else {
+      log.info(
```
kfaraz: Let's skip logging this for every pending segment that has been upgraded. We can just log a total count summary in the task action.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1419,22 +1421,89 @@ public void registerNewVersionOfPendingSegment(
...
           }
         }
       }
     }
+
+    if (notifiedTasks == 0) {
+      // No running task matched: the segment will not be re-announced until handoff. This is a potential silent-loss
+      // window where data will not be queryable until handoff.
+      log.warn(
+          "Upgraded pending segment[%s] (upgradedFrom[%s], taskAllocatorId[%s]) matched no running task on"
```
kfaraz: Nit: slight rephrase
```suggestion
          "Could not find any task matching taskAllocatorId[%s] in supervisor[%s] for upgraded pending segment[%s] (upgradedFrom[%s]) "
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1419,22 +1421,89 @@ public void registerNewVersionOfPendingSegment(
...
+      emitter.emit(getMetricBuilder().setMetric(SegmentUpgradeMetrics.UNMATCHED, 1));
+    } else {
+      log.info(
+          "Notified [%d] task(s) across [%d] task group(s) of upgraded pending segment[%s] (upgradedFrom[%s]).",
+          notifiedTasks,
+          matchedGroups,
+          pendingSegmentRecord.getId(),
+          pendingSegmentRecord.getUpgradedFromSegmentId()
+      );
+      emitter.emit(getMetricBuilder().setMetric(SegmentUpgradeMetrics.NOTIFIED, 1));
```
kfaraz: Let's emit the number of tasks notified instead of always emitting 1.
(I wonder if we should also emit the `taskId` that was notified)

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1419,22 +1421,89 @@ public void registerNewVersionOfPendingSegment(
...
+          {
+            // Successful delivery is captured by the notified metric; nothing to do here.
+          }
+
+          @Override
+          public void onFailure(Throwable t)
+          {
+            log.warn(
+                t,
+                "Failed to send upgraded pending segment[%s] to task[%s] on supervisor[%s].",
```
```suggestion
                "Failed to register upgraded pending segment[%s] on task[%s] of supervisor[%s].",
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1419,22 +1421,89 @@ public void registerNewVersionOfPendingSegment(
...
+  private void notifyTaskOfUpgradedPendingSegment(String taskId, PendingSegmentRecord pendingSegmentRecord)
+  {
+    Futures.addCallback(
+        taskClient.registerNewVersionOfPendingSegmentAsync(taskId, pendingSegmentRecord),
+        new FutureCallback<>()
+        {
+          @Override
+          public void onSuccess(Boolean result)
+          {
+            // Successful delivery is captured by the notified metric; nothing to do here.
```
kfaraz: Nit: I don't think this is entirely true since the `notified` metric only captures the fact that we have sent a notification, not that it has been acknowledged successfully.
But I agree that we can skip emitting a metric here. (since the lack of a `sendFailed` metric should suffice).

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -1865,6 +1869,33 @@ public Response registerUpgradedPendingSegment(
...
+    switch (outcome) {
+      case SKIPPED_UNKNOWN_BASE:
+        reason = SegmentUpgradeMetrics.REASON_UNKNOWN_BASE;
+        break;
+      case SKIPPED_NO_SINK:
+        reason = SegmentUpgradeMetrics.REASON_NO_SINK;
+        break;
+      case SKIPPED_DROPPING:
+        reason = SegmentUpgradeMetrics.REASON_DROPPING;
+        break;
```
kfaraz, line 1916: Can we capture the full reason string inside the enum itself?

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -1181,12 +1181,46 @@ private void unannounceSegment(DataSegment segment)
     }
   }
 
-  public void registerUpgradedPendingSegment(PendingSegmentRecord pendingSegmentRecord) throws IOException
+  /**
+   * Outcome of a {@link #registerUpgradedPendingSegment} call, returned so the caller (which owns the emitter) can
+   * emit the corresponding metric.
+   */
+  public enum UpgradeAnnouncementOutcome
```
kfaraz:
```suggestion
  public enum PendingSegmentUpgradeResult
```

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalReplaceAction.java
```diff
@@ -169,14 +171,33 @@ private void registerUpgradedPendingSegmentsOnSupervisor(
       List<PendingSegmentRecord> upgradedPendingSegments
   )
   {
+    // Emit the count of upgrades this commit produced regardless of whether a supervisor exists to
```
kfaraz:
```suggestion
    // Emit the count of upgrades successfully committed regardless of whether a supervisor exists to
```

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalReplaceAction.java
```diff
@@ -169,14 +171,33 @@ private void registerUpgradedPendingSegmentsOnSupervisor(
...
+      log.info("No active streaming supervisor for datasource[%s]; the [%d] upgraded pending segment(s) from task[%s]"
+               + " will become queryable when their tasks hand off.",
+                task.getDataSource(),
+                upgradedPendingSegments.size(),
+                task.getId()
+      );
       return;
     }
 
+    log.info("Registering [%d] upgraded pending segments created by task[%s] on supervisor[%s]",
```
kfaraz: Nit: for better formatting, start the message in a new line.
```suggestion
    log.info(
        "Registering [%d] upgraded pending segments created by task[%s] on supervisor[%s]",
```

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalReplaceAction.java
```diff
@@ -169,14 +171,33 @@ private void registerUpgradedPendingSegmentsOnSupervisor(
...
+             upgradedPendingSegments.size(),
+             task.getId(),
+             activeSupervisorIdWithAppendLock.get()
+    );
     upgradedPendingSegments.forEach(
         upgradedPendingSegment -> supervisorManager.registerUpgradedPendingSegmentOnSupervisor(
```
kfaraz: Maybe the `registerUpgradedPendingSegment` method should return the number of tasks that were notified for this pending segment. We could log a summary map of pending segment ID to number of tasks notified.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalReplaceAction.java
kfaraz: Maybe move this log to the end of this method and include a summary map from pending segment ID to number of tasks notified.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
```diff
@@ -2116,6 +2117,85 @@ public void testRegisterNewVersionOfPendingSegment()
...
+    EasyMock.expect(spec.isSuspended()).andReturn(false);
+    replayAll();
+
+    final TestSeekableStreamSupervisor supervisor = new TestSeekableStreamSupervisor();
+    supervisor.getIoConfig().setTaskCount(3);
+    supervisor.start();
+
+    supervisor.addTaskGroupToActivelyReadingTaskGroup(
+        supervisor.getTaskGroupIdForPartition("0"),
+        ImmutableMap.of("0", "5"),
```
kfaraz, line 2139: Nit: Use `Map.of` and `Set.of` for brevity.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -1181,12 +1181,46 @@ private void unannounceSegment(DataSegment segment)
     }
   }
 
-  public void registerUpgradedPendingSegment(PendingSegmentRecord pendingSegmentRecord) throws IOException
+  /**
+   * Outcome of a {@link #registerUpgradedPendingSegment} call, returned so the caller (which owns the emitter) can
+   * emit the corresponding metric.
```
kfaraz:
```suggestion
   * Result of a {@link #registerUpgradedPendingSegment} call.
```

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -1181,12 +1181,46 @@ private void unannounceSegment(DataSegment segment)
...
+  public UpgradeAnnouncementOutcome registerUpgradedPendingSegment(PendingSegmentRecord pendingSegmentRecord) throws IOException
   {
     SegmentIdWithShardSpec basePendingSegment = idToPendingSegment.get(pendingSegmentRecord.getUpgradedFromSegmentId());
     SegmentIdWithShardSpec upgradedPendingSegment = pendingSegmentRecord.getId();
-    if (!sinks.containsKey(basePendingSegment) || droppingSinks.contains(basePendingSegment)) {
-      return;
+    if (basePendingSegment == null || droppingSinks.contains(basePendingSegment) || !sinks.containsKey(basePendingSegment)) {
+      if (basePendingSegment == null) {
+        // This task never allocated a segment matching upgradedFromSegmentId, i.e. the request targeted the wrong task.
+        log.info(
```
kfaraz: Should be a warn, I think.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -1181,12 +1181,46 @@ private void unannounceSegment(DataSegment segment)
...
+        log.debug(
+            "Not announcing upgraded pending segment[%s] for base segment[%s] on task[%s] because the base sink is being dropped.",
+            upgradedPendingSegment, basePendingSegment, myId
+        );
+        return UpgradeAnnouncementOutcome.SKIPPED_DROPPING;
+      } else {
+        // Unexpected: the base sink is gone even though this task once held it.
+        log.info(
+            "Not announcing upgraded pending segment[%s] for base segment[%s] on task[%s] because the base sink is no longer present.",
+            upgradedPendingSegment, basePendingSegment, myId
```
kfaraz: I feel like these logs can be merged into one and can be logged by the caller itself using the result of this method.

kfaraz (commented):
Thanks for adding the metrics/logs, @capistrant ! These would be instrumental in debugging failures in concurrent compaction.

Overall looks good, but I have left some nitpicks.

indexing-service/src/test/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisorStateTest.java
capistrant: ack on Set. the method signature seems to require explicit ImmutableMap tho

kfaraz: Oh, I see. Thanks for the clarification!

indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentUpgradeMetrics.java
capistrant: I'm in favor of keeping right now. especially  comparing the deliveries by the supervisor to the tasks actions with those deliveries I think it is easier if they are separate

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1419,22 +1421,89 @@
...
+          "Could not find any task matching taskAllocatorId[%s] in supervisor[%s] for upgraded pending segment[%s] (upgradedFrom[%s]) "
+          + " supervisor[%s]; it will not be re-announced until handoff. Currently tracking [%d] activelyReading"
+          + " and [%d] pendingCompletion task group(s).",
+          pendingSegmentRecord.getId(),
+          pendingSegmentRecord.getUpgradedFromSegmentId(),
+          taskAllocatorId,
+          supervisorId,
+          activelyReadingTaskGroups.size(),
+          pendingCompletionTaskGroups.size()
+      );
```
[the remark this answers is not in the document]
capistrant: whoops

docs/operations/metrics.md
```diff
@@ -317,6 +317,12 @@ batch ingestion emit the following metrics. These metrics are deltas for each em
 |`ingest/notices/time`|Milliseconds taken to process a notice by the supervisor.|`supervisorId`, `dataSource`, `tags`| < 1s |
 |`ingest/pause/time`|Milliseconds spent by a task in a paused state without ingesting.|`dataSource`, `taskId`, `tags`| < 10 seconds|
 |`ingest/handoff/time`|Total number of milliseconds taken to handoff a set of segments.|`dataSource`, `taskId`, `taskType`, `groupId`, `tags`|Depends on the coordinator cycle time.|
+|`ingest/realtime/segmentUpgrade/count`|Number of pending segments that a concurrent replace (for example, compaction) upgraded to a new version and asked the supervisor to have running tasks announce under the new version. Emitted by the replace task only when streaming ingestion is running concurrently with replace on the same interval.|`dataSource`, `taskId`, `taskType`, `groupId`, `interval`, `version`, `tags`|0 unless [concurrent append and replace](../ingestion/concurrent-append-replace.md) is in use.|
```
kfaraz: We should probably rename this metric from `count` to `persisted` to signify that this is the count that was actually persisted to the metadata store.
Also, rephrased the text a bit.

```suggestion
|`ingest/realtime/segmentUpgrade/persisted`|Number of pending segments that were upgraded in the metadata store while publishing segments to an interval. Emitted by the Overlord only when a REPLACE task (e.g., compaction task with `useConcurrentLocks` set to `true`) commits segments to an interval already containing pending segments allocated to an APPEND task (e.g. streaming task with `useConcurrentLocks` set to `true`). |`dataSource`, `taskId`, `taskType`, `groupId`, `interval`, `version`, `tags`|0 unless [concurrent append and replace](../ingestion/concurrent-append-replace.md) is in use.|
```

indexing-service/src/main/java/org/apache/druid/indexing/common/SegmentUpgradeMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+   * Task dims.
+   */
+  public static final String SKIPPED = "ingest/realtime/segmentUpgrade/skipped";
+
+  /**
+   * Adds the upgraded pending segment's {@code interval} and {@code version} to a metric builder so that every
+   * segment-upgrade metric can be sliced by the specific segment being re-announced. Mirrors
+   * {@code IndexTaskUtils.setSegmentDimensions}, which serves the same purpose for {@code DataSegment}s.
+   */
+  public static ServiceMetricEvent.Builder setSegmentDimensions(
```
kfaraz: Nit: Maybe move this method to `IndexTaskUtils` itself, as it feels like a more generic place (rather than upgrade specific) allowing for reuse for this method.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalReplaceAction.java
```diff
@@ -169,19 +173,45 @@ private void registerUpgradedPendingSegmentsOnSupervisor(
...
     final Optional<String> activeSupervisorIdWithAppendLock =
         supervisorManager.getActiveSupervisorIdForDatasourceWithAppendLock(task.getDataSource());
 
     if (!activeSupervisorIdWithAppendLock.isPresent()) {
+      log.info("No active streaming supervisor for datasource[%s]; the [%d] upgraded pending segment(s) from task[%s]"
+               + " will become queryable when their tasks hand off.",
+                task.getDataSource(),
+                upgradedPendingSegments.size(),
+                task.getId()
+      );
```
kfaraz: minor rephrase
```suggestion
      log.info(
          "Could not find any active concurrent streaming supervisor for datasource[%s]."
          + " Ignoring registry of [%d] pending segment(s) upgraded by task[%s]. These will become queryable only after handoff."
          task.getDataSource(),
          upgradedPendingSegments.size(),
          task.getId()
      );
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1415,26 +1417,102 @@ public void resetOffsets(@Nonnull DataSourceMetadata resetDataSourceMetadata)
     addNotice(new ResetOffsetsNotice(resetDataSourceMetadata));
   }
 
-  public void registerNewVersionOfPendingSegment(
+  /**
+   * Notifies every running task in the matching task group(s) of an upgraded pending segment.
+   *
+   * @return the number of tasks notified, which the caller aggregates into a per-batch summary
```
```suggestion
   * @return the number of tasks notified
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1415,26 +1417,102 @@ public void resetOffsets(@Nonnull DataSourceMetadata resetDataSourceMetadata)
...
-            taskClient.registerNewVersionOfPendingSegmentAsync(taskId, pendingSegmentRecord);
+            notifyTaskOfUpgradedPendingSegment(taskId, pendingSegmentRecord);
+            notifiedTasks++;
           }
         }
       }
     }
+
+    // Only the exceptional no-match case is logged per segment; the notified count is returned to the task action,
+    // which summarizes the whole batch in one log line, and each notification is captured by the per-task metric.
```
kfaraz: We need not call this out. This method should be agnostic of the behaviour of its caller.
We choose not to log the successful case here to reduce verbosity, irrespective of the behaviour of the caller.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
```diff
@@ -1415,26 +1417,102 @@ public void resetOffsets(@Nonnull DataSourceMetadata resetDataSourceMetadata)
...
+    if (notifiedTasks == 0) {
+      // No running task matched: the segment will not be re-announced until handoff. This is a potential silent-loss
+      // window where data will not be queryable until handoff.
+      log.warn(
+          "Could not find any task matching taskAllocatorId[%s] in supervisor[%s] for upgraded pending segment[%s]"
+          + " (upgradedFrom[%s]); it will not be re-announced until handoff. Currently tracking [%d] activelyReading"
+          + " and [%d] pendingCompletion task group(s).",
```
kfaraz: Do you think logging the number of actively reading / pending completion groups would be useful for debugging? I feel we may omit these for now.

server/src/main/java/org/apache/druid/segment/realtime/appenderator/StreamAppenderator.java
```diff
@@ -1181,12 +1181,53 @@ private void unannounceSegment(DataSegment segment)
...
-  public void registerUpgradedPendingSegment(PendingSegmentRecord pendingSegmentRecord) throws IOException
+  /**
+   * Result of a {@link #registerUpgradedPendingSegment} call. Each skipped outcome carries the human-readable
+   * {@link #getReason() reason} emitted as the {@code reason} metric dimension.
+   */
+  public enum PendingSegmentUpgradeResult
+  {
+    ANNOUNCED(null),
+    /** The task holds no pending segment matching upgradedFromSegmentId (request targeted the wrong task). */
+    SKIPPED_UNKNOWN_BASE("unknown base"),
```
kfaraz:
```suggestion
    SKIPPED_UNKNOWN_BASE("unknown base sink"),
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -1865,6 +1869,61 @@ public Response registerUpgradedPendingSegment(
     }
   }
 
+  /**
+   * Logs and emits a metric for the outcome of a pending-segment upgrade request, keyed off the {@code outcome}
+   * returned by {@link StreamAppenderator#registerUpgradedPendingSegment}.
+   */
+  private void recordUpgradeOutcome(
```
kfaraz:
```suggestion
  private void recordUpgradeResult(
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/SeekableStreamIndexTaskRunner.java
```diff
@@ -1865,6 +1869,61 @@ public Response registerUpgradedPendingSegment(
...
+      PendingSegmentRecord upgradedPendingSegment,
+      StreamAppenderator.PendingSegmentUpgradeResult outcome
```
kfaraz:
```suggestion
      StreamAppenderator.PendingSegmentUpgradeResult result
```

kfaraz (approved):

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |
| P3 | 0 |
| Total | 2 |

---

extensions-contrib/prometheus-emitter/src/main/resources/defaultMetrics.json
```diff
@@ -134,6 +134,12 @@
   "ingest/notices/time" : { "dimensions" : ["dataSource"], "type" : "timer", "conversionFactor": 1000.0, "help": "Seconds taken to process a notice by the supervisor." },
   "ingest/pause/time" : { "dimensions" : ["dataSource"], "type" : "timer", "conversionFactor": 1000.0, "help": "Seconds spent by a task in a paused state without ingesting." },
   "ingest/handoff/time" : { "dimensions" : ["dataSource"], "type" : "timer", "conversionFactor": 1000.0, "help": "Total number of seconds taken to handoff a set of segments." },
+  "ingest/realtime/segmentUpgrade/count" : { "dimensions" : ["dataSource", "interval", "version"], "type" : "count", "help": "Number of pending segments upgraded by a concurrent replace." },
```
FrankChen021: [P1] Avoid unbounded per-version Prometheus series

The default exporter has no TTL unless `flushPeriod` is explicitly configured, and its collector retains every label tuple. `version`, `interval`, and `taskId` continually change across replacements and tasks, so recurring concurrent compaction makes these six default metrics accumulate series indefinitely on long-lived Overlord/supervisor processes, potentially exhausting heap. Keep high-cardinality identifiers out of the default mapping or enforce bounded expiry.

indexing-service/src/main/java/org/apache/druid/indexing/common/actions/SegmentTransactionalReplaceAction.java
```diff
@@ -169,19 +173,45 @@ private void registerUpgradedPendingSegmentsOnSupervisor(
...
+          .registerUpgradedPendingSegmentOnSupervisor(activeSupervisorIdWithAppendLock.get(), upgradedPendingSegment)
+          .ifPresent(notified -> notifiedTasksBySegment.put(upgradedPendingSegment.getId().toString(), notified));
+    }
+    log.info(
+        "Registered [%d] upgraded pending segment(s) created by task[%s] on supervisor[%s]; tasks notified per"
+        + " segment[%s].",
+        notifiedTasksBySegment.size(),
+        task.getId(),
+        activeSupervisorIdWithAppendLock.get(),
+        notifiedTasksBySegment
```
FrankChen021: [P2] Bound the per-segment INFO summary

The upgraded-segment list is not capped, yet this builds and formats every segment ID into one INFO message. Broad compaction can produce thousands of entries and megabyte-scale log lines on the task-action path. Log aggregate counts with a bounded sample, reserving full detail for debug output.

capistrant: fair point

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/SeekableStreamSupervisor.java
capistrant: hmm I guess probably not useful. was kinda just packing in bits of information wherever I could find them in this PR. but you're right I'm not sure what I do with this number in this log so I will remove

PR #19654, opened 2026-07-03
fix: fix query context precedence layer for datasource-level per-segment timeout

jtuglu1:
### Description

Currently, per-datasource, per-segment timeout can be not set if there's a hardcoded static default query context (`druid.query.default.context.perSegmentTimeout`). This ensures the evaluation precedence for the per-segment timeout is as follows (from highest to lowest):

1. Query-specified context values (passed in the query payload)
2. Dynamic configs (broker per-datasource per-segment timeout, broker dynamic config, etc.). Former beats the latter for this particular config.
3. Static defaults specified at process boot (`druid.query.default.context.perSegmentTimeout`)
4. Default values specified in the application code.

Also adds the perSegmentTimeout context key so it's easy to auto-complete in the UI.

#### Release note

Fix query context precedence layer for datasource-level per-segment timeout.

<hr>

This PR has:

- [ ] been self-reviewed.

### DIFF
--- a/server/src/main/java/org/apache/druid/client/BrokerViewOfBrokerConfig.java
+++ b/server/src/main/java/org/apache/druid/client/BrokerViewOfBrokerConfig.java
@@ -36,6 +36,7 @@ import org.apache.druid.rpc.ServiceClientFactory;
 import org.apache.druid.rpc.ServiceLocator;
 import org.apache.druid.rpc.StandardRetryPolicy;
 import org.apache.druid.server.broker.BrokerDynamicConfig;
+import org.apache.druid.server.broker.QueryConfigSnapshot;
 
 import javax.validation.constraints.NotNull;
 import java.util.Map;
@@ -126,5 +127,14 @@ public class BrokerViewOfBrokerConfig extends BaseBrokerViewOfConfig<BrokerDynam
   public Map<String, Object> getContext()
   {
     return resolvedDefaultQueryContext.asMap();
+  }
+
+  /**
+   * Captures the resolved default context and the dynamic config atomically, so a single query resolves its context
+   * and blocklist against one consistent snapshot rather than re-reading the live config.
+   */
+  public synchronized QueryConfigSnapshot snapshotForQuery()
+  {
+    return new QueryConfigSnapshot(getContext(), getDynamicConfig());
   }
 }
--- a/server/src/main/java/org/apache/druid/client/BrokerViewOfBrokerConfig.java
+++ b/server/src/main/java/org/apache/druid/client/BrokerViewOfBrokerConfig.java
@@ -55,11 +55,13 @@ public class BrokerViewOfBrokerConfig extends BaseBrokerViewOfConfig<BrokerDynam
   private final DefaultQueryConfig defaultQueryConfig;
 
   /**
-   * Pre-computed merge of {@link DefaultQueryConfig#getContext()} and
+   * The dynamic config plus the merge of {@link DefaultQueryConfig#getContext()} and
    * {@link BrokerDynamicConfig#getQueryContext()}, recomputed on each config sync.
-   * Dynamic config values override static defaults. {@link QueryContext} provides immutability.
+   *
+   * <p>volatile, not synchronized: read on the query hot path, see {@link BaseBrokerViewOfConfig}. Both halves
+   * live in one field so a query cannot observe them from different generations.
    */
-  private volatile QueryContext resolvedDefaultQueryContext;
+  private volatile QueryConfigSnapshot querySnapshot;
 
   @Inject
   public BrokerViewOfBrokerConfig(
@@ -70,7 +72,7 @@ public class BrokerViewOfBrokerConfig extends BaseBrokerViewOfConfig<BrokerDynam
   )
   {
     this.defaultQueryConfig = defaultQueryConfig;
-    this.resolvedDefaultQueryContext = QueryContext.of(defaultQueryConfig.getContext());
+    this.querySnapshot = new QueryConfigSnapshot(QueryContext.of(defaultQueryConfig.getContext()).asMap(), null);
     this.coordinatorClient =
         new CoordinatorClientImpl(
             clientFactory.makeClient(
@@ -90,7 +92,7 @@ public class BrokerViewOfBrokerConfig extends BaseBrokerViewOfConfig<BrokerDynam
   {
     this.coordinatorClient = coordinatorClient;
     this.defaultQueryConfig = defaultQueryConfig;
-    this.resolvedDefaultQueryContext = QueryContext.of(defaultQueryConfig.getContext());
+    this.querySnapshot = new QueryConfigSnapshot(QueryContext.of(defaultQueryConfig.getContext()).asMap(), null);
   }
 
   @Override
@@ -110,13 +112,16 @@ public class BrokerViewOfBrokerConfig extends BaseBrokerViewOfConfig<BrokerDynam
    * resolved default query context by merging static defaults with dynamic overrides.
    */
   @Override
-  public synchronized void setDynamicConfig(@NotNull BrokerDynamicConfig updatedConfig)
+  public void setDynamicConfig(@NotNull BrokerDynamicConfig updatedConfig)
   {
     super.setDynamicConfig(updatedConfig);
-    resolvedDefaultQueryContext = QueryContext.of(QueryContexts.override(
-        defaultQueryConfig.getContext(),
-        updatedConfig.getQueryContext().asMap()
-    ));
+    querySnapshot = new QueryConfigSnapshot(
+        QueryContext.of(QueryContexts.override(
+            defaultQueryConfig.getContext(),
+            updatedConfig.getQueryContext().asMap()
+        )).asMap(),
+        updatedConfig
+    );
   }
 
   /**
@@ -126,15 +131,23 @@ public class BrokerViewOfBrokerConfig extends BaseBrokerViewOfConfig<BrokerDynam
   @Override
   public Map<String, Object> getContext()
   {
-    return resolvedDefaultQueryContext.asMap();
+    return querySnapshot.getResolvedDefaultQueryContext();
   }
 
   /**
-   * Captures the resolved default context and the dynamic config atomically, so a single query resolves its context
-   * and blocklist against one consistent snapshot rather than re-reading the live config.
+   * Snapshot for a single query to resolve its context and blocklist against, instead of re-reading the live config.
    */
-  public synchronized QueryConfigSnapshot snapshotForQuery()
+  public QueryConfigSnapshot snapshotForQuery()
   {
-    return new QueryConfigSnapshot(getContext(), getDynamicConfig());
+    return querySnapshot;
+  }
+
+  /**
+   * Reads through {@link #querySnapshot} so this and {@link #snapshotForQuery()} always agree.
+   */
+  @Override
+  public BrokerDynamicConfig getDynamicConfig()
+  {
+    return querySnapshot.getDynamicConfig();
   }
 }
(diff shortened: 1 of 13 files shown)


FrankChen021 (commented):

---

server/src/main/java/org/apache/druid/server/QueryLifecycle.java
```diff
@@ -212,43 +226,64 @@ public void after(final boolean isDone, final Throwable thrown)
...
+   * pass the real user-provided keys.
+   *
+   * @throws DruidException if the current state is not NEW, which indicates a bug
+   */
+  public void initialize(final Query<?> baseQuery, @Nullable final Set<String> userProvidedContextKeys)
   {
     transition(State.NEW, State.INITIALIZED);
 
     userContextKeys = new HashSet<>(baseQuery.getContext().keySet());
+    final Set<String> effectiveUserKeys =
```
maytasm: nit: since it's not really clear from the variable names (`userContextKeys` vs `userProvidedContextKeys`), can you add a comment here on the difference between userContextKeys and userProvidedContextKeys

maytasm (approved):

FrankChen021 (commented):

---

server/src/main/java/org/apache/druid/server/QueryLifecycle.java
jtuglu1: Agreed. I've refactored since I didn't really want to leak this context override logic (per segment timeout, etc.) into the query handler code.

server/src/main/java/org/apache/druid/client/BrokerViewOfBrokerConfig.java
kfaraz: Does this method still need to be `synchronized`?

server/src/main/java/org/apache/druid/client/BrokerViewOfBrokerConfig.java
```diff
@@ -127,4 +128,13 @@ public Map<String, Object> getContext()
   {
     return resolvedDefaultQueryContext.asMap();
   }
+
+  /**
+   * Captures the resolved default context and the dynamic config atomically, so a single query resolves its context
+   * and blocklist against one consistent snapshot rather than re-reading the live config.
+   */
+  public synchronized QueryConfigSnapshot snapshotForQuery()
```
kfaraz: Probably need not be synchronized given the recent changes in #19787 .

server/src/main/java/org/apache/druid/server/broker/QueryConfigSnapshot.java
```diff
@@ -0,0 +1,72 @@
...
+  private final BrokerDynamicConfig dynamicConfig;
+
+  public QueryConfigSnapshot(Map<String, Object> defaultContext, @Nullable BrokerDynamicConfig dynamicConfig)
+  {
+    this.defaultContext = defaultContext;
+    this.dynamicConfig = dynamicConfig;
+  }
+
+  /**
+   * The final query context. Precedence high to low: a key the client set > per-query dynamic override > the query's
```
kfaraz: Please put the precedence list inside an `<ol>` tag with `<li>` items.

server/src/main/java/org/apache/druid/server/broker/BrokerDynamicConfig.java
```diff
@@ -91,6 +96,33 @@ public Map<String, PerSegmentTimeoutConfig> getPerSegmentTimeoutConfig()
     return perSegmentTimeoutConfig;
   }
 
+  /**
+   * Query-specific broker dynamic config query context overrides (e.g. per segment timeout).
+   */
+  public QueryContext getQuerySpecificContextOverrides(Query<?> query)
```
kfaraz:
```suggestion
  public QueryContext getContextOverridesForQuery(Query<?> query)
```

server/src/main/java/org/apache/druid/server/broker/BrokerDynamicConfig.java
```diff
@@ -91,6 +96,33 @@ public Map<String, PerSegmentTimeoutConfig> getPerSegmentTimeoutConfig()
...
+  {
+    if (perSegmentTimeoutConfig.isEmpty()) {
+      return QueryContext.empty();
+    }
+
+    for (String tableName : query.getDataSource().getTableNames()) {
+      PerSegmentTimeoutConfig dsConfig = perSegmentTimeoutConfig.get(tableName);
```
kfaraz: Please use the name `dataSourceTimeoutConfig` instead.

server/src/main/java/org/apache/druid/server/broker/BrokerDynamicConfig.java
```diff
@@ -91,6 +96,33 @@ public Map<String, PerSegmentTimeoutConfig> getPerSegmentTimeoutConfig()
...
+      if (dsConfig != null) {
+        if (dsConfig.isMonitorOnly()) {
+          log.debug(
+              "Per-segment timeout [%d ms] configured for datasource [%s] in monitorOnly mode (not enforced) for query [%s].",
```
kfaraz:
```suggestion
              "Per-segment timeout [%d ms] configured for datasource[%s] in monitorOnly mode (not enforced) for query[%s].",
```

server/src/main/java/org/apache/druid/server/broker/BrokerDynamicConfig.java
```diff
@@ -91,6 +96,33 @@ public Map<String, PerSegmentTimeoutConfig> getPerSegmentTimeoutConfig()
...
+  {
+    if (perSegmentTimeoutConfig.isEmpty()) {
+      return QueryContext.empty();
+    }
+
+    for (String tableName : query.getDataSource().getTableNames()) {
+      PerSegmentTimeoutConfig dsConfig = perSegmentTimeoutConfig.get(tableName);
+      if (dsConfig != null) {
+        if (dsConfig.isMonitorOnly()) {
+          log.debug(
```
kfaraz, line 114: Does this really warrant a log, even debug?

server/src/main/java/org/apache/druid/server/broker/QueryConfigSnapshot.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A snapshot of the broker's resolved default query context and dynamic config, captured once for a single query so
+ * that context resolution and the blocklist read one consistent view (rather than re-reading the live config at
+ * different points in the query lifecycle). The context-resolution logic lives here, keeping it out of
+ * {@code QueryLifecycle}. On non-broker nodes {@code dynamicConfig} is null (defaults only, no per-query overrides).
```
kfaraz: Nit: rephrase for simplicity:
```suggestion
 * A snapshot of the {@link BrokerDynamicConfig} (null on non-Broker nodes)
 * and the resolved default query context that is used for the entire {@code QueryLifecycle}
 * of a single query.
```

server/src/main/java/org/apache/druid/server/broker/QueryConfigSnapshot.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class QueryConfigSnapshot
+{
+  private final Map<String, Object> defaultContext;
```
kfaraz: Please rename this to `resolvedDefaultQueryContext` to avoid confusion and indicate that it has already been resolved with `BrokerDynamicConfig.getContext()`.

server/src/main/java/org/apache/druid/server/broker/QueryConfigSnapshot.java
kfaraz: Why is the query's own context the 3rd item in this list?

As the PR description mentions, I think the context precedence order should be:
query payload > dynamic config > runtime property > code defaults.

kfaraz: I suppose my confusion might be stemming from the fact that the `BrokerDynamicConfig` has a dedicated `context` field but also provides query-specific overrides.

server/src/main/java/org/apache/druid/server/QueryLifecycle.java
```diff
@@ -113,8 +108,11 @@ public class QueryLifecycle
 
   @MonotonicNonNull
   private Query<?> baseQuery;
+  /**
+   * Keys present on the query context as received; the candidate set fed to context-key authorization.
+   */
   @MonotonicNonNull
-  private Set<String> userContextKeys;
+  private Set<String> authorizationContextKeys;
```
kfaraz: Should this be called `authorizedUserContextKeys` or similar?

server/src/main/java/org/apache/druid/server/QueryLifecycle.java
```diff
@@ -212,61 +220,41 @@ public void after(final boolean isDone, final Throwable thrown)
...
+
     String queryId = baseQuery.getId();
     if (Strings.isNullOrEmpty(queryId)) {
       queryId = UUID.randomUUID().toString();
     }
 
-    // Start with system defaults, apply per-datasource override, then user context wins
-    Map<String, Object> contextWithDefaults = new HashMap<>(queryConfigProvider.getContext());
-    applyPerDatasourcePerSegmentTimeout(baseQuery, contextWithDefaults, queryId);
-    Map<String, Object> finalContext = QueryContexts.override(contextWithDefaults, baseQuery.getContext());
```
kfaraz, line 227: @jtuglu1 , I assume here was the bug that we are trying to fix.

IIUC, since we didn't have the exact set of `clientProvidedQueryContextKeys`, we assumed that everything in the `baseQuery` was set by the client and the per-datasource-per-segment timeout in the dynamic config would never get applied.

Is that understanding correct?

kfaraz (commented):
Minor nitpicks, otherwise looks good.

server/src/main/java/org/apache/druid/server/QueryLifecycle.java
jtuglu1: Correct.

jtuglu1: The dynamic config would never override the per-segment timeout context value (assuming it wasn't set directly by the client) if there was a static default per-segment timeout context value.

server/src/main/java/org/apache/druid/client/BrokerViewOfBrokerConfig.java
jtuglu1: Yeah – I removed this.

server/src/main/java/org/apache/druid/client/BrokerViewOfBrokerConfig.java
jtuglu1: Updated.

server/src/main/java/org/apache/druid/server/broker/QueryConfigSnapshot.java
jtuglu1: I think this comment was misleading. The order should be (and is) `query payload > dynamic config > runtime property > code defaults`, I think the last 2 just got baked into the "query's context." Updated the comment.

server/src/main/java/org/apache/druid/server/broker/QueryConfigSnapshot.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Set;
+
+/**
+ * A snapshot of the {@link BrokerDynamicConfig} (null on non-Broker nodes)
+ * and the resolved default query context that is used for the entire {@code QueryLifecycle}
+ * of a single query.
+ */
+public class QueryConfigSnapshot
+{
+  /** Already resolved against {@link BrokerDynamicConfig#getQueryContext()}. */
```
kfaraz, line 39: Nit: Please move or copy this to the field getter as well.

kfaraz (approved):
LGTM 🚀

server/src/main/java/org/apache/druid/server/broker/QueryConfigSnapshot.java
kfaraz: Merging the PR for now, this change can be included in some later related patch, if any.

PR #19655, opened 2026-07-03
feat: Introduce lag emergency option for cost-based autoscaler

Fly-Style:
## Summary

Adds a two-tier "critical lag" fast path to the cost-based Kafka/Kinesis supervisor autoscaler, so an operator-configured SLA threshold can trigger a faster scale-up reaction, when enabled.

`CostBasedAutoScalerConfig` gains an optional `criticalLagThreshold` (aggregate/global lag across all partitions, not per-partition). When set, `CostMetrics#getAggregateLag()` is compared against it on every evaluation:

- **Tier 1 (≥75% of threshold):** the lag-amplification multiplier used in the cost function's lag-recovery-time term maxes out at `6.0` (vs. the default `0.3`), and the scale-up candidate search bypasses its usual boundary cap (`useTaskCountBoundariesOnScaleUp`), letting the argmin search the full candidate range instead of being capped to +2 steps above the current task count.
- **Tier 2 (≥100% of threshold):** cost minimization is skipped entirely — the autoscaler jumps straight to the maximum valid task count, regardless of configured weights or idle ratio.

`criticalLagThreshold` is `null` by default, which leaves existing behavior unchanged. The `6.0` multiplier is configurable.

Meanwhile, default `LAG_AMPLIFICATION_MULTIPLIER` was set to 0.0 to prioritize better task utilization whenever possible.

<details>

**IMAGE 1**
Plot shows how the tier-1 fast path (amp=6.0) responds across a range of operating conditions, holding `partitionCount=500`, `aggregateLag=10M` (above the 75% tier-1 mark but below the 95% tier-2 mark, given `criticalLagThreshold=11.5M`), and `currentTaskCount=50` fixed:

- **X-axis** — `lagWeight`/`idleWeight` pairs (the two cost-function weights, always summing to 1.0), swept from lag-averse (`0.1/0.9`) to lag-aggressive (`0.9/0.1`).
- **Y-axis** — the resulting optimal task count chosen by the argmin search for each weight pair.
- **Lines** — one per `(idle, rate)` combination: color encodes current idle ratio (red=0.1, green=0.25), line style encodes processing rate (solid=50k/s, dashed=80k/s, dotted=100k/s).

The takeaway: even holding lag constant, the tier-1 boost isn't a flat override — task count still climbs smoothly as `lagWeight` increases, but low-idle/low-rate combinations (red solid) saturate at the task-count ceiling far earlier (~`0.6/0.4`) than high-idle/high-rate combinations (green dotted, ~`0.9/0.1`), because those still have slack the idle-cost term can trade off against. This demonstrates tier 1 amplifies urgency rather than short-circuiting the optimization the way tier 2 does.

image

**IMAGE 2**
Scale-down driven purely by idle ratio (`128 tasks, 128 partitions, aggregateLag≈1000, rate=25k, optimalTaskIdleRatio=0.25`).

image

</details>

This PR has:

- [ ] been self-reviewed.


indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -284,6 +321,12 @@ int computeOptimalTaskCount(CostMetrics metrics)
                    : startIndex;
     }
 
+    // Emergency (tier 2) lag skips the argmin search entirely: evaluate only the maximum valid task count.
+    if (emergencyLag) {
+      startIndex = validTaskCounts.length - 1;
```
FrankChen021, line 333: [P1] Make emergency lag choose the max unconditionally

The emergency path narrows the loop to only the maximum valid task count, but `optimalTaskCount` and `optimalCost` are still initialized from the current task count before this block. If the configured weights make the current-count cost lower than the max-count cost, for example `lagWeight=0` and `idleWeight=1`, the loop evaluates the max candidate but never updates `optimalTaskCount`, so tier-2 emergency lag returns the current count instead of jumping to the max. Set the emergency result to the max candidate unconditionally, or initialize the optimal candidate from `startIndex` after the emergency bounds are applied.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Found 1 issue: tier-2 emergency lag can still keep the current task count instead of forcing the configured maximum.

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -207,6 +207,25 @@ public CostBasedAutoScalerConfig getConfig()
     return config;
   }
 
+  private boolean isCriticalLag(CostMetrics metrics)
+  {
+    final Long criticalLagThreshold = config.getCriticalLagThreshold();
+    return metrics != null && criticalLagThreshold != null
+           && metrics.getAggregateLag() >= criticalLagThreshold * WeightedCostFunction.CRITICAL_LAG_TIER1_FRACTION;
```
FrankChen021: [P2] Use exact total lag for tier checks

`collectMetrics()` supplies only `lagStats.getAvgLag()`, while `CostMetrics#getAggregateLag()` reconstructs the total as `avgLag * partitionCount`. `DefaultLagAggregator` computes that average using integer division, so the reconstructed value can undercount by up to `partitionCount - 1`; for example, total lag 999 across 1000 partitions becomes aggregate lag 0. A valid `criticalLagThreshold` may therefore be crossed without engaging tier 1 or tier 2. With the updated normal multiplier set to 0, a missed tier also leaves lag unamplified and scale-up boundaries enabled. Carry `lagStats.getTotalLag()` into `CostMetrics` and use the exact total for these comparisons.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

---

FrankChen021 (commented):
I have reviewed the updated code for correctness, edge cases, concurrency, and integration risks; no new issues found.

---

docs/ingestion/supervisor.md
```diff
@@ -208,13 +208,13 @@ The following table outlines the configuration properties related to the `costBa
...
 |`idleWeight`|How much weight to give the idle cost relative to the lag cost. Higher values make the autoscaler more aggressive about removing over-provisioned tasks.|No| `0.6`                     |
 |`useTaskCountBoundariesOnScaleUp`|Limits scale-up to a small step relative to the current task count, preventing large jumps. Disable to allow the autoscaler to jump directly to any task count.|No| `false`                   |
 |`useTaskCountBoundariesOnScaleDown`|Limits scale-down to a small step relative to the current task count, preventing large drops. Disable to allow the autoscaler to drop directly to any task count.|No| `true`                    |
-|`minScaleUpDelay`|Minimum cooldown after a scale-up before the next scale-up is allowed. Specified as an ISO-8601 duration.|No| `scaleActionPeriodMillis` |
-|`minScaleDownDelay`|Minimum cooldown after a scale-down before the next scale-down is allowed. Specified as an ISO-8601 duration.|No| `PT30M`                   |
+|`minScaleUpDelay`|Minimum cooldown after a scale-up before the next scale-up is allowed. Specified as an ISO-8601 duration.|No| `PT15M`                   |
```
FrankChen021, line 218: [P2] Align the advertised cooldown defaults with the implementation

The updated table promises `PT15M` and `PT20M`, but the merged implementation retains `DEFAULT_MIN_SCALE_UP_DELAY = PT10M` and `DEFAULT_MIN_SCALE_DOWN_DELAY = PT30M`, which are applied when these fields are omitted. Consequently, operators receive a shorter scale-up cooldown and longer scale-down cooldown than documented. The pre-merge update changed the constants, so this appears to be a merge-resolution regression; restore those runtime defaults or correct the documentation and add literal-value assertions.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Found one merge-resolution regression: the documented autoscaler cooldown defaults disagree with the runtime constants.

---

FrankChen021 (commented):
I have reviewed the updated code for correctness, edge cases, concurrency, lifecycle, compatibility, and integration risks; no new issues found. The prior cooldown-default finding is addressed: runtime defaults now match the documented 15-minute scale-up and 20-minute scale-down values.

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -46,8 +46,9 @@ public class CostBasedAutoScalerConfig implements AutoScalerConfig
 {
   static final double DEFAULT_LAG_WEIGHT = 0.4;
   static final double DEFAULT_IDLE_WEIGHT = 0.6;
-  static final Duration DEFAULT_MIN_SCALE_UP_DELAY = Duration.standardMinutes(10);
-  static final Duration DEFAULT_MIN_SCALE_DOWN_DELAY = Duration.standardMinutes(30);
+  static final Duration DEFAULT_MIN_SCALE_UP_DELAY = Duration.standardMinutes(15);
+  static final Duration DEFAULT_MIN_SCALE_DOWN_DELAY = Duration.standardMinutes(20);
```
kfaraz: Let's keep scale down delay at 30 mins only.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -292,6 +329,12 @@ int computeOptimalTaskCount(CostMetrics metrics)
                    : startIndex;
     }
 
+    // Emergency (tier 2) lag skips the argmin search entirely: evaluate only the maximum valid task count.
+    if (emergencyLag) {
```
FrankChen021: [P1] Bypass the cost-drop gate during tier-2 emergencies

After this emergency branch restricts evaluation to the maximum task count, execution still reaches the existing `minCostDropPercentForScaling` check at lines 377-385. With both options configured, a tier-2 lag emergency can therefore return the current count when the maximum improves cost by less than the configured percentage, contradicting the stated jump-to-maximum behavior and potentially leaving an SLA-critical backlog under-provisioned. Tier-2 should bypass that gate or return the maximum directly.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 1 |
| P2 | 0 |
| P3 | 0 |
| Total | 1 |

Found one emergency-scaling path where the normal cost-drop gate can still prevent the documented jump to the maximum task count.

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -37,10 +37,30 @@ public class WeightedCostFunction
...
+   * and the scale-up candidate boundary is bypassed.
+   */
+  static final double CRITICAL_LAG_TIER1_FRACTION = 0.75;
+
+  /**
+   * Fraction of {@link CostBasedAutoScalerConfig#getCriticalLagThreshold()} at which tier 2 of the
+   * critical-lag fast path engages: the cost-minimization search is skipped entirely and the task
+   * count jumps straight to the maximum.
+   */
+  static final double CRITICAL_LAG_TIER2_FRACTION = 0.95;
```
kfaraz: Let's have tier 2 at 100% itself.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/WeightedCostFunction.java
```diff
@@ -37,10 +37,30 @@ public class WeightedCostFunction
...
+   * urgency amplification, so normal lag uses unamplified recovery time.
    */
-  static final double LAG_AMPLIFICATION_MULTIPLIER = 0.3;
+  static final double LAG_AMPLIFICATION_MULTIPLIER = 0.0;
+
+  /**
+   * Amplification multiplier used once aggregate lag crosses {@link #CRITICAL_LAG_TIER1_FRACTION} of
+   * {@link CostBasedAutoScalerConfig#getCriticalLagThreshold()} (tier 1 of the critical-lag fast path).
+   */
+  static final double CRITICAL_LAG_AMPLIFICATION_MULTIPLIER = 6.0;
```
kfaraz: Let's make this configurable, as discussed offline.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
kfaraz: Please double check this, @Fly-Style !

Fly-Style:
@kfaraz all addressed in https://github.com/apache/druid/pull/19655/commits/19c0acb01c65e2c3044012c1443ba9472e64dba5

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
```diff
@@ -214,6 +214,25 @@ public CostBasedAutoScalerConfig getConfig()
...
+    return metrics != null && criticalLagThreshold != null
+           && metrics.getAggregateLag() >= criticalLagThreshold * WeightedCostFunction.CRITICAL_LAG_TIER1_FRACTION;
+  }
+
+  /**
+   * Whether the last collected metrics crossed {@link WeightedCostFunction#CRITICAL_LAG_TIER2_FRACTION} of
+   * {@link CostBasedAutoScalerConfig#getCriticalLagThreshold()}, meaning the argmin search should be
+   * skipped entirely in favor of jumping straight to the maximum task count.
+   */
+  private boolean isEmergencyLag(CostMetrics metrics)
```
kfaraz: Let's call this state as `critical lag` and the other state as simply `high lag`. The terms critical and emergency can be confusing, especially since "emergency lag" currently corresponds to 100% of critical lag.

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
```diff
@@ -87,11 +90,13 @@ public CostBasedAutoScalerConfig(
       @Nullable @JsonProperty("minScaleDownDelay") Duration minScaleDownDelay,
       @Nullable @JsonProperty("scaleDownDuringTaskRolloverOnly") Boolean scaleDownDuringTaskRolloverOnly,
       @Nullable @JsonProperty("usePollIdleRatio") Boolean usePollIdleRatio,
-      @Nullable @JsonProperty("minCostDropPercentForScaling") Integer minCostDropPercentForScaling
+      @Nullable @JsonProperty("criticalLagThreshold") Long criticalLagThreshold,
+      @Nullable @JsonProperty("minCostDropPercentForScaling") Integer minCostDropPercentForScaling,
+      @Nullable @JsonProperty("criticalLagAmplificationMultiplier") Double criticalLagAmplificationMultiplier
```
kfaraz: Maybe use a simpler name.
```suggestion
      @Nullable @JsonProperty("highLagCostFactor") Double highLagCostFactor
```

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScalerConfig.java
kfaraz:
```suggestion
  static final Duration DEFAULT_MIN_SCALE_DOWN_DELAY = Duration.standardMinutes(30);
```

docs/ingestion/supervisor.md
```diff
@@ -208,13 +208,13 @@ The following table outlines the configuration properties related to the `costBa
...
+|`minScaleDownDelay`|Minimum cooldown after a scale-down before the next scale-down is allowed. Specified as an ISO-8601 duration.|No| `PT20M`                   |
```
FrankChen021: [P2] Document the new emergency controls

The cost-based autoscaler reference still omits both new operator-facing properties, `criticalLagThreshold` and `criticalLagAmplificationMultiplier`. Since the threshold defaults to null, readers cannot discover how to enable the headline feature or that it uses aggregate lag; the configurable multiplier also makes the fixed-6.0 wording in the Javadoc and PR description stale. Please document the properties, aggregate-lag units, 75%/100% tier behavior, and null/6.0 defaults.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 1 |
| P3 | 0 |
| Total | 1 |

Reviewed 9 of 9 changed files. The original P1 is fixed and is acknowledged in its existing inline thread.

---

indexing-service/src/main/java/org/apache/druid/indexing/seekablestream/supervisor/autoscaler/CostBasedAutoScaler.java
FrankChen021: Rechecked on the current head: this is addressed. Tier 2 now selects the maximum valid task count even when that candidate costs more, and it bypasses `minCostDropPercentForScaling`; the new regression test covers the higher-cost-maximum case. Reviewed 9 of 9 changed files.

kfaraz (approved):

docs/ingestion/supervisor.md
Fly-Style: Addressed in https://github.com/apache/druid/pull/19655/commits/685641a6ba3ceae77f4d9c436b9eeb4fbdaa64ec

PR #19657, opened 2026-07-06
feat: wire up partial load rules to segment cache

clintropolis:
### Description
This PR finishes the partial load rules by wiring the partial load specs up to the segment cache, allowing historicals in virtual storage mode to eagerly load and pin parts of segments to the cache, while allowing the rest of the segment to still be loaded on demand (and evicted).


server/src/test/java/org/apache/druid/segment/loading/PartialSegmentCacheBootstrapTest.java
capistrant, line 1: nit: tests are slapped in-between a bunch of helper methods. please hoist up to where tests are

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -872,6 +925,429 @@ private record ReservedPartial(
...
+    }
+    if (rangeReader == null) {
+      // Backend doesn't support range reads (e.g. zipped deep storage). The rule can't be honored as a partial load;
+      // fall through to the on-demand weak-full-load path at query time. But if a prior loadPartial installed a
+      // static entry with a different rule fingerprint, we cannot silently keep serving under the old rule so
+      // reconcile the static side so the coordinator's rule change takes effect, then return so the next query
+      // installs a weak entry via the on-demand path.
+      log.warn(
+          "Backend for segment[%s] does not support range reads; partial-load rule[fingerprint=%s] cannot be honored, "
+          + "falling back to weak full-load at query time",
```
capistrant: I agree with the decision to fallback here, but how will we help an operator get out of this state? do you think that will come in via documentation of the feature, guiding an operator on how to configure and common pitfalls (like zipped segments resulting in logs like these, and what to change to fix)?

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -872,6 +925,429 @@ private record ReservedPartial(
...
+   * with its data eagerly downloaded (non-selected bundles are downloaded on-demand at query time via the existing
+   * weak path).
+   * <p>
+   * The entire flow runs synchronously on {@link #virtualStorageLoadingThreadPool} (via submit + get) so it respects
+   * the pool's concurrency cap and virtual-thread runtime while giving
+   * {@link org.apache.druid.server.coordination.SegmentLoadDropHandler} a truthful success/failure signal. On any
+   * failure (mount, bundle reservation, download) the metadata + any partially-reserved bundles are cascade-released
+   * before the {@link SegmentLoadingException} propagates, so no orphan static reservation survives the load call.
+   */
+  private void loadPartial(DataSegment dataSegment) throws SegmentLoadingException
```
capistrant, line 940: this method feels long. It is just an observation at this point since it all seems like necessary code and the sequencing makes sense. but it was kind of exhausting to read from top to bottom. I hate to advocate for breaking out some stuff into additional private helpers with one call-site though.

server/src/test/java/org/apache/druid/segment/loading/SegmentLocalCacheManagerPartialRuleLoadTest.java
```diff
@@ -0,0 +1,830 @@
...
+          location.isReserved(id),
+          "no static partial entry should be installed alongside a held weak entry"
+      );
+    }
+    finally {
+      hold.close();
+    }
+  }
+
+  private static CacheEntry stubCacheEntry(CacheEntryIdentifier id, long size)
```
capistrant: nit: should go up with the other private helpers above the tests

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -872,6 +925,429 @@ private record ReservedPartial(
...
+              rangeReader,
+              jsonMapper,
+              virtualStorageLoadingThreadPool,
+              config.getVirtualStorageMetadataReservationEstimate()
+          );
+          if (!candidate.reserve(candidateEntry)) {
+            continue;
+          }
+          try {
+            FileUtils.mkdirp(partialDir);
```
capistrant: not seeing this cleaned up in the event of failure and rollback. does that need to cleaned up outside of the callable if it fails and rolls back?

capistrant (commented):
nothing I see as outright blocking. left a few comments and nits.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -872,6 +925,493 @@ private record ReservedPartial(
...
+            return;
+          }
+          dropStaleStatic = true;
+        }
+        if (dropStaleStatic) {
+          log.info(
+              "Dropping stale static partial entry for segment[%s]: current rule cannot be applied without range reads",
+              dataSegment.getId()
+          );
+          drop(dataSegment);
```
FrankChen021, line 1048: [P2] Preserve stale static partials while refs are active

The range-reader-null fallback drops a stale static partial entry and immediately deletes stale segment directories without the outstanding-reference precheck used by the normal reconciliation path. If a query still holds a metadata or bundle reference, `drop` only removes the entry from the static map while actual cleanup is deferred; deleting/reusing the directory before those refs close can break in-flight reads or let deferred cleanup delete files for the next entry. This path should fail/retry when the old partial would defer cleanup.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
```diff
@@ -1069,9 +1659,18 @@ public void drop(final DataSegment segment)
...
-        location.release(entry);
+      if (entry == null) {
+        continue;
       }
+      // Cascade-release linked bundle entries before the metadata. {@link StorageLocation#release} only operates on
+      // entries in {@code staticCacheEntries}, so this is a no-op for weak/on-demand bundles; for static rule-pinned
+      // bundles installed by {@link #loadPartial}, this is the only path that removes them from the static map.
+      if (entry instanceof PartialSegmentMetadataCacheEntry partial) {
+        for (PartialSegmentBundleCacheEntry bundle : partial.snapshotLinkedBundles()) {
+          location.release(bundle);
```
FrankChen021, line 1670: [P2] Evict weak linked bundles during partial drops

The new drop cascade calls `location.release` for each linked bundle, but `release` is a no-op for weak/on-demand bundles. A rule-loaded partial segment can still have non-selected bundles mounted through the weak path; those bundles hold metadata references for their lifetime, so dropping/reconciling the static metadata can proceed, delete/recreate the same segment directory, and leave old weak bundle entries to later evict containers through the old mapper. Drop/reconcile needs to remove unheld weak linked bundles too, or refuse to proceed while any linked weak bundle remains active.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

---

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
FrankChen021, line 1048: [P2] Range-reader fallback drops live partial entries

This unsupported-range-reader fallback calls `drop()` and then removes every stale per-segment directory without the `cascadeReleaseWouldDeferCleanup()` guard used by the main reconciliation path. If a query is still holding the old static partial metadata or bundle when a changed rule falls back because the delegate has no range reader, `drop()` can defer old cleanup while `removeStaleSegmentDirsOnAllLocations()` deletes or reuses the same cache directory underneath that query. Please apply the same outstanding-reference precheck here and retry later instead of dropping when cleanup would defer.

server/src/main/java/org/apache/druid/segment/loading/SegmentLocalCacheManager.java
FrankChen021, line 1670: [P2] Weak linked bundles survive partial drop

`location.release(bundle)` only removes entries from `staticCacheEntries`, so it is a no-op for weak/on-demand bundles linked to a static partial metadata entry. Those weak bundles still hold metadata references, but the reconciliation precheck subtracts all linked bundles as if drop will release them, after which the new code can nuke or reuse the segment directory while the weak bundles remain mounted. Drop/reconcile should either remove unheld weak linked bundles or treat their presence as a deferred-cleanup blocker.

FrankChen021 (commented):
| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

| Severity | Findings |
| --- | ---: |
| P0 | 0 |
| P1 | 0 |
| P2 | 2 |
| P3 | 0 |
| Total | 2 |

Found 2 lifecycle issues in the partial-load-rule reconciliation/drop paths.

---

clintropolis:
closed in favor of #19671

